PR #2, opened 2024-11-02
Add first Actions

ppraegla:
The following workflows are added:
- Build the docker image containing the dependencies (remove the newer
  gcc compiler because the docker build takes too long otherwise. There
  is limit of 6 hours.
- Checkcode which runs pre-commit (The codeowner check is left out as it
  is tied to Gitlab)
- A first build and test workflow. We only run the minimal test as we
  ran out of disk space when running all tests.

Once this is merged. We are able to test if the actions and triggers work as expected.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -1,0 +1,68 @@
+name: buildtest
+
+on:
+  pull_request:
+    branches:
+      - 'main'
+  push:
+    branches:
+      - 'main'
+
+env:
+  REGISTRY: ghcr.io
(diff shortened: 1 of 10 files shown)


sebproell:
@ppraegla Thanks for hanging in there and getting the basic workflow to run now for the first time! Do you want to squash all the CI commits into one but keep the two code-related fixes? This could help us keep track of what we need to reflect back to Gitlab.

ppraegla:
@sebproell I adjusted the workflow triggers and restructured the commits.

sebproell (approved):

PR #3, opened 2024-11-08
Don't run no-commit-to-branch in CI

amgebauer:
## Description and Context

Why does this check actually not fail on Gitlab?
@sebproell @ppraegla


sebproell:
The checkcode job is not run on main in Gitlab. Do we want to run it on Github?

sebproell (approved):

PR #4, opened 2024-11-09
Create reusable configure and build action

amgebauer:
## Description and Context

Create composite actions to reuse configure and build 4C.

### DIFF
--- a/.github/actions/build_4C/action.yml
+++ b/.github/actions/build_4C/action.yml
@@ -1,0 +1,20 @@
+name: Build 4C
+description: Build a specified 4C target
+inputs:
+  build-directory:
+    description: Path to the build directory
+    required: true
+  build-targets:
+    description: Targets to build
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Build 4C with targets ${{ inputs.build-targets }}
+      run: |
+        cd $BUILD_DIRECTORY
+        time cmake --build . --target $BUILD_TARGETS -- -j `nproc` 2>&1
+      shell: bash
+      env:
+        BUILD_DIRECTORY: ${{ inputs.build-directory }}
+        BUILD_TARGETS: ${{ inputs.build-targets }}
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -28,37 +28,29 @@ jobs:
     defaults:
       run:
         shell: bash
-    env:
-      CMAKE_PRESET: docker
-      BUILD_TARGETS: full
     steps:
       - run: echo ${{ github.repository }}
       # Due to a bug in runner action the variables $GITHUB_WORKSPACE and ${{ github.workspace }} are different inside a container. https://github.com/actions/runner/issues/2058
       # The repo gets cloned to `/__w/4C/4C` ($GITHUB_WORKSPACE) while ${{ github.workspace }} points to `/home/runner/work/4C/4C`.`
       # Use $GITHUB_WORKSPACE instead of ${{ github.workspace }}
       - uses: actions/checkout@v4
-      - name: Fix git
-        run: git config --global --add safe.directory $GITHUB_WORKSPACE
       - name: info
         run: |
           pwd; ls -l
-      - name: Build
-        run: |
-          echo ${{ github.workspace }}
-          echo $GITHUB_WORKSPACE
-          cmake --version; echo ${{ env.CMAKE_PRESET }}
-          mkdir -p $GITHUB_WORKSPACE/../build
-          cd $GITHUB_WORKSPACE
-          ./utilities/set_up_dev_env.sh
-          cd $GITHUB_WORKSPACE/../build
-          cmake $GITHUB_WORKSPACE --fresh --preset=${{ env.CMAKE_PRESET }}
-          echo Building the following targets ${{ env.BUILD_TARGETS }}
-          time cmake --build . --target ${{ env.BUILD_TARGETS }} -- -j `nproc` 2>&1
+      - uses: ./.github/actions/configure_4C
+        with:
+          cmake-preset: docker
+          build-directory: ${{ github.workspace }}/build
+      - uses: ./.github/actions/build_4C
+        with:
+          build-targets: full
+          build-directory: ${{ github.workspace }}/build
       - name: Test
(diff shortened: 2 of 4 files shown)


sebproell (approved):
LGTM, but I'll let @ppraegla decide if this makes sense in combination with the planned next steps.

amgebauer:
#5 will also add a similar action for build. I will combine these two into this PR.

amgebauer:
With the latest changes, it's now possible to reuse configure and build in other workflows. I did it in #5 for doxygen and readthedocs.

Not, this PR is ready for review.

.github/workflows/buildtest.yml
```diff
@@ -28,37 +28,29 @@ jobs:
...
+      - uses: ./.github/actions/build_4C
+        with:
+          build-targets: full
+          build-directory: ${{ github.workspace }}/../build
       - name: Test
         run: |
           cd $GITHUB_WORKSPACE/../build
           time ctest -L minimal -j `nproc` --output-on-failure --output-junit $GITHUB_WORKSPACE/junit_test_summary.xml
       - name: Upload build and test logs
+        if: ${{ always() }}
```
amgebauer, line 53: @ppraegla This was missing. I added it on the way.

.gitlab/CODEOWNERS
```diff
@@ -415,6 +415,7 @@ Gemfile.lock                                   @baci/team_infrastructure
 .github/ISSUE_TEMPLATE/bug_report.md                                         @baci/baci_developers
 .github/ISSUE_TEMPLATE/enhancement.md                                        @baci/baci_developers
 .github/workflows/                                                           @baci/team_infrastructure
+.github/actions/                                                             @baci/team_infrastructure
```
ppraegla, line 418:
```suggestion
.github/                                                             @baci/team_infrastructure
```
I think we could just name the `.github/` folder to save us the trouble to adapt in other PRs. The Gitlab codeowners file does not make sense anyway once we are fully on Github.

.github/actions/build_4C/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    description: Path to the log-file
+    required: false
+    default: ${{ github.workspace }}/build.log
+runs:
+  using: composite
+  steps:
+    - name: Build 4C with targets ${{ inputs.build-targets }}
+      run: |
+        cd $BUILD_DIRECTORY
+        time cmake --build . --target $BUILD_TARGETS -- -j `nproc` 2>&1 | tee $LOG_FILE
```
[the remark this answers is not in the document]
ppraegla: I think it is fine because when we set the shell to bash below. The option `pipefail` is added automatically. 
See `shell: bash --noprofile --norc -e -o pipefail {0}` from the workflow log.

But I agree that we might not even need this log. But we can keep it for now and remove it later.

.github/workflows/buildtest.yml
```diff
@@ -28,37 +28,29 @@ jobs:
...
-          cd $GITHUB_WORKSPACE
-          ./utilities/set_up_dev_env.sh
-          cd $GITHUB_WORKSPACE/../build
-          cmake $GITHUB_WORKSPACE --fresh --preset=${{ env.CMAKE_PRESET }}
-          echo Building the following targets ${{ env.BUILD_TARGETS }}
-          time cmake --build . --target ${{ env.BUILD_TARGETS }} -- -j `nproc` 2>&1
+      - uses: ./.github/actions/configure_4C
+        with:
+          cmake-preset: docker
+          build-directory: ${{ github.workspace }}/../build
```
ppraegla, line 43: I'm a bit confused whats happening here (not your code. But what the github runner is doing). You set the build directory to `${{ github.workspace }}/../build` which according to the [log](https://github.com/4C-multiphysics/4C/actions/runs/11763968417/job/32768579260#step:6:27) is `BUILD_DIRECTORY: /home/runner/work/4C/4C/../build`. At the end of the configuring the [log](https://github.com/4C-multiphysics/4C/actions/runs/11763968417/job/32768579260#step:6:522) says `Build files have been written to: /__w/4C/build`. The subsequent build works though. So it seems that`$GITHUB_WORKSPACE` and `${{ github.workspace }}` are after all the same directory.

ppraegla (approved):
Nice work. Only some minor comments. You can merge if you want.

sebproell (approved):

.github/actions/build_4C/action.yml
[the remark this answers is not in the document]
amgebauer: Let me remove it. We don't even need it for readthedocs.

.github/workflows/buildtest.yml
amgebauer: Yes, that also confused me.

sebproell: Slightly related: how about using a subfolder `build` inside the repo in the future? Actually, I think we should even ignore this folder in the `.gitignore` since this is how I nowadays recommend setting up a build.

amgebauer: good idea. I do that for a few other jobs to avoid unnecessary copying and avoiding relative links for artifacts. Only for clang-tidy, it might not work (or we might need to exclude the build-folder then)

sebproell: Why should it not work for clang-tidy? We are normally using the `-p` arg which tells clang-tidy where to look for compile command lines. I use that locally all the time.

.gitlab/CODEOWNERS
amgebauer: I left it for now like so as we also have the PR/Issue templates. We can change it later if we decide how we will proceed with codeowners.

.github/workflows/buildtest.yml
amgebauer: We then also get warnings/errors from our dependencies (ArborX, Google test, etc).

sebproell: Ah I see, it is the combination of `-p compile_commands.json` and the additional working directory. I am fine with leaving it as is.

sebproell (approved):

amgebauer:
I had to fix a path. The tests passed, but I need another approval.

sebproell (approved):

PR #6, opened 2024-11-10
Extend codecheck with clang-tidy and verify-headers

amgebauer:
## Description and Context

Extend codecheck workflow with clang-tidy and verify-headers

### DIFF
--- a/.github/actions/configure_4C/action.yml
+++ b/.github/actions/configure_4C/action.yml
@@ -28,7 +28,8 @@ runs:
         cd $GITHUB_WORKSPACE
         ./utilities/set_up_dev_env.sh
         cd $BUILD_DIRECTORY
-        cmake $GITHUB_WORKSPACE --fresh --preset=$CMAKE_PRESET $ADDITIONAL_CMAKE_FLAGS
+        # use xargs as $ADDITIONAL_CMAKE_FLAGS can contain multiple flags
+        echo $GITHUB_WORKSPACE --fresh --preset=$CMAKE_PRESET $ADDITIONAL_CMAKE_FLAGS | xargs cmake
         cd $GITHUB_WORKSPACE
       shell: bash
       env:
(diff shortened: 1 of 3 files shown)


.github/workflows/checkcode.yml
```diff
@@ -20,4 +20,41 @@ jobs:
...
+          . venv/bin/activate
+          pip install wheel
+          pip install -r ${GITHUB_WORKSPACE}/utilities/code_climate/requirements.txt
+          python ${GITHUB_WORKSPACE}/utilities/code_climate/export_clang_tidy_issues.py -j `nproc` --code-climate-file ${GITHUB_WORKSPACE}/codeclimate.json --source-folder ${GITHUB_WORKSPACE} ${GITHUB_WORKSPACE}/../clang-tidy-build/clang-tidy-issues/*.yaml
+      - name: Upload clang-tidy issues
+        uses: actions/upload-artifact@v4
+        with:
+          name: codeclimate-report
+          path: |
+            codeclimate.json
```
sebproell, line 58: Can Github use this code climate file the same way as Gitlab does? I couldn't find anything from a quick search.

.github/actions/configure_4C/action.yml
```diff
@@ -32,7 +32,7 @@ runs:
         cd $GITHUB_WORKSPACE
         ./utilities/set_up_dev_env.sh
         cd $BUILD_DIRECTORY
-        cmake $GITHUB_WORKSPACE --fresh --preset=$CMAKE_PRESET $ADDITIONAL_CMAKE_FLAGS | tee $LOG_FILE
+        echo $GITHUB_WORKSPACE --fresh --preset=$CMAKE_PRESET $ADDITIONAL_CMAKE_FLAGS | xargs cmake | tee $LOG_FILE
```
sebproell, line 35: What is the reason for this special invocation?

amgebauer: We need this so that `$ADDITIONAL_CMAKE_FLAGS` can contain multiple flags (and not just one).

sebproell: I would add a comment explaining this reason, otherwise somebody is going to "optimize" the call.

.github/workflows/checkcode.yml
amgebauer: No. There are automatic actions that add comments to the PR, but I don't think that is feasible here for us as we still have too many warnings. We could also remove the artifact.

sebproell: Just a curiosity. We should probably convert a lot more clang-tidy warnings into errors in the future. The report itself was never that useful to me on Gitlab anyway.

sebproell (approved):

PR #7, opened 2024-11-10
Separate build and test jobs

amgebauer:
## Description and Context

Test a simple separation of build and test. Probably, we want to do it a bit smarter.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -63,6 +63,11 @@ jobs:
       credentials:
         username: ${{ github.actor }}
         password: ${{ secrets.github_token }}
+    strategy:
+      fail-fast: false
+      matrix:
+        test-chunk: ["0,199", "200,399", "400,599", "600,799", "800,999", "1000,1199", "1200,1399", "1400,1599",
+          "1600,1799", "1800,1999", "2000,2199", "2200,2399", "2400,2599", "2600,2799", "2800,"]
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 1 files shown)


amgebauer:
This proof of concept works and we shift an artifact of around 300 MB from build to test. I think that should be acceptable.

sebproell:
> This proof of concept works and we shift an artifact of around 300 MB from build to test. I think that should be acceptable.

So the idea would be to download this artifact on a lot of parallel jobs all of which run a distinct set of tests, right? Do you want to try this here?

amgebauer:
>> This proof of concept works and we shift an artifact of around 300 MB from build to test. I think that should be acceptable.
>
> So the idea would be to download this artifact on a lot of parallel jobs all of which run a distinct set of tests, right? Do you > want to try this here?

Yes, that's the idea. Since tests are changing very quickly, I would try to do it on Gitlab first and will cherry-pick it here for testing. Otherwise, we will build up a lot of conflicts.

sebproell:
It is interesting that the dependent jobs do not yet appear in the report. Are we not using the feature `depends_on` in the intended way here?

amgebauer:
> It is interesting that the dependent jobs do not yet appear in the report. Are we not using the feature depends_on in the intended way here?

I have it from here: https://github.com/actions/upload-artifact?tab=readme-ov-file#example-output-between-jobs. They appear at least here: https://github.com/4C-multiphysics/4C/actions/runs/11766531115?pr=7

sebproell:
This almost works but apparently you are trying to overwrite the same log file: https://github.com/4C-multiphysics/4C/actions/runs/11767090752/job/32776462744?pr=7

PR #9, opened 2024-11-12
Create junit report for tests

amgebauer:
## Description and Context

Let's try to create a junit report of the tests.


amgebauer:
image

That's all that we get with this action, and it's even hard to find. Closing for now.

PR #10, opened 2024-11-12
Use ccache in build-stage

amgebauer:
Enable to use ccache in build stage between jobs with the same preset

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -18,6 +18,8 @@ permissions:
 
 jobs:
   gcc9_build:
+    env:
+      CMAKE_PRESET: docker
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/ppraegla/4c-dependencies:latest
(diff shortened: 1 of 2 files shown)


sebproell:
I think you need to actively use `ccache` during building. Currently, the cache is still empty after building.

sebproell (approved):
I like this version where both the CMake flag and the use of the ccache action are explixit in the workflow.

isteinbrecher:
Just so I understand, does this mean that we start from a clean `ccache` in the build stage and reuse this in later jobs such as Doxygen, or does this mean that pipelines reuse the `ccache` of previous pipelines?

amgebauer:
It means the latter (we reuse the cache from previous pipelines).

from https://ccache.dev/manual/4.8.2.html
> Ccache has been carefully written to always produce exactly the same compiler output that you would get without the cache. The only way you should be able to tell that you are using ccache is the speed

I don't expect a problem here.

isteinbrecher:
We did have problems with that in the past. Can the cache be cleaned somehow?

sebproell:
> We did have problems with that in the past. Can the cache be cleaned somehow?

@isteinbrecher Sure, the cache can be cleaned, but what is the point of caching if we do that? Also, could you be more specific about what kind of problems you encountered with `ccache`?

isteinbrecher:
We had the issue that even with a clean build folder 4C did not compile, only after a reset of the `ccache` cache. @sebproell you misread my question, of course it makes no sense to clean the cache each build. My question was more in the direction if we can "clean" the cache.

amgebauer:
The workflows on Github should work without manual intervention. If we encounter any problems in the CI because of ccache, we can (and need to) change the setup. But we should not do that if we are not sure whether the problems back then were really a problem with ccache. I rather suspect the cmake cache that caused the problems back then. Here, we don't share here between runs.

sebproell:
@isteinbrecher Do you maybe mean "clean the cache manually"? If yes, I don't know if we can do that, although I suspect there is an option to wipe caches.

> We had the issue that even with a clean build folder 4C did not compile, only after a reset of the ccache cache.

Personally, I have never experienced that problem with ccache. Could it be that you are referring to a point in time more than 2-3 years in the past? Back then, our configuration and build setup was a brittle mess, full of custom logic. Let's go for this PR and if we encounter problems, we fix them.

amgebauer:
Caching is based on this action https://github.com/actions/cache and is stored somewhere similar to artifacts but with the following additional logic:

> A repository can have up to 10GB of caches. Once the 10GB limit is reached, older caches will be evicted based on when the cache was last accessed. Caches that are not accessed within the last week will also be evicted.

isteinbrecher:
Thanks for the explanation!

isteinbrecher (approved):

amgebauer:
@isteinbrecher there is even a mechanism to remove caches in the GUI of Github, see https://github.com/4C-multiphysics/4C/actions/caches

PR #12, opened 2024-11-13
CI: use separate workflow for deployment

sebproell:
Try to make the overview of actions a bit more intuitive by not showing the skipped deployment in PRs.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,37 @@
+name: deploy
+
+on:
+  workflow_run:
+    workflows: [documentation]
+    types: [completed]
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  deploy:
+    runs-on: ubuntu-24.04
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    permissions:
+      pages: write
+      id-token: write
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - uses: actions/download-artifact@v4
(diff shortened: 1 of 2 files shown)


sebproell:
This achieves what I wanted in a PR. Should we merge and see if it does the correct thing on `main`?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  workflow_run:
+    workflows: [documentation.yml]
+    types: [completed]
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
```
amgebauer, line 12:
```suggestion
jobs:
 on-success:
```

I think we need this to ensure that we only deploy if build of doxygen and readthedocs was successful

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,36 @@
+name: deploy
+
+on:
+  workflow_run:
+    workflows: [documentation.yml]
```
ppraegla, line 5: Is this the correct syntax or does it need to be the workflow name?
```suggestion
    workflows: [documentation]
```

.github/workflows/deploy.yml
sebproell: I agree that we might need such a check but isn't `on-success` just the name of the job?

amgebauer: true, I read too fast over the docs

sebproell: Condition added.

amgebauer:
@sebproell merge?

PR #14, opened 2024-11-13
Fix ccache in pipelines

amgebauer:
ccache does currently not work as expected. Github does not find the cache in other jobs (only in restarted jobs).


amgebauer:
The cache works as it should, I just misunderstood it. The PRs don't share caches, they just have access to their own cache or the default-branch cache.

PR #15, opened 2024-11-13
Add concurrency to auto-cancel workflows

amgebauer:
To not waste resources, we need to cancel redundant workflows


amgebauer:
This PR is ready for approval. I tested it, and redundant pipelines were canceled.

ppraegla (approved):

PR #16, opened 2024-11-13
Only execute enforced checks in clang-tidy

amgebauer:


amgebauer:
clang-tidy now takes only 21 minutes. The big downside is that we duplicate a lot info from `.clang-tidy` in `utilities/code_climate/clang-tidy-config`. I hesitate to adapt `.clang-tidy` as the checks in there are still relevant during development.

Maybe that (https://github.com/llvm/llvm-project/issues/84508) could help to at least reduce duplicate information.

PR #17, opened 2024-11-13
Update docker build from branch inside repo

ppraegla:
Try out a PR from a branch within this repo to update the docker image because the workflow from a fork cannot write to the package registry.

### DIFF
--- a/.github/actions/compute-and-check-dependencies-hash/action.yml
+++ b/.github/actions/compute-and-check-dependencies-hash/action.yml
@@ -1,0 +1,31 @@
+name: dependencies hash
+description: Compute and check the docker dependencies hash
+inputs:
+  docker_image_hash:
+    description: 'The hash for the docker image'
+    required: true
+outputs:
+  computed_dependencies_hash:
+    description: 'The computed hash from the dependencies'
+    value: ${{ steps.compute-depencencies-hash.outputs.dependencies_hash }}
+runs:
+  using: composite
+  steps:
+    - id: compute-depencencies-hash
+      name: Compute the dependencies hash
+      run: |
+        cd $GITHUB_WORKSPACE
+        COMPUTED_DOCKER_DEPENDENCIES_HASH=`find dependencies docker -not -wholename '*/trilinos_develop/*' -not -name 'README.md' -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8`
+        echo "dependencies_hash=$COMPUTED_DOCKER_DEPENDENCIES_HASH" >> $GITHUB_OUTPUT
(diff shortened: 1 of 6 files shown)


ppraegla:
@sebproell @amgebauer I'm not sure how to proceed with this PR. 

To update the docker image, we have to push a branch directly to the 4C repo and open a PR. When opening a PR from a fork, the workflow does not have the permission to write to our package registry (makes sense such that not anybody can publish new packages). I don't know of a better way than just pushing the branch directly to the 4C repo and I think it is justified for an update to the docker image.

I would keep our strategy to use the hash of the dependencies and Dockerfile as tag for the image. However, I'm not sure how we should do the workflows during the update PR. My idea was to only start the docker workflow, compute and check the hash and build the docker image, if it does not already exist. Once this workflow is done, it should trigger the other workflows via `workflow_run`. The problem with `workflow_run` is that it only uses the workflow file on the default branch, i.e. the code state on main. This will not use our newly created docker image. According to [this](https://github.com/potherca-blog/github-actions-workflow_run-event), we could use the checkout action to specifically check out the state of the originating workflow. But this will still use the workflow file from main.

And lastly, how do we want to do the normal workflows for PRs from works? Do we also start with one workflow where we check if the docker image tag matches the dependencies and trigger the other workflows from that or do we add the check to every job like we do it on Gitlab?

Any ideas are welcome :)

amgebauer:
Thanks for investigating it. 

The good thing is that we don't have to upgrade the docker image too often, so I think it's acceptable if it comes with some extra effort. With all these downsides, I would also not clutter our jobs with the extra docker step and rather end up in a similar workflow that we currently have on Gitlab, i.e., verifying the hash after the checkout action.

I would also move the docker job into a workflow that we can trigger manually for a branch. Then we can hide all these details as 99% of the PRs won't touch our dependencies. I think this manual intervention is acceptable since PRs that change dependencies are anyway special as they need to be an internal branch.

Maybe, we could build Trilinos in our pipeline and rely on ccache (we should never need to rebuilt it from scratch once built). Then we can capture this case that happens regularly.

mrp089:
No idea if it helps, but here is svFSIplus's "docker landscape:"

Docker Hub: https://registry.hub.docker.com/u/simvascular
Dockerfile dependencies: https://github.com/SimVascular/svFSIplus/blob/main/Docker/ubuntu22/dockerfile
Dockerfile solver: https://github.com/SimVascular/svFSIplus/blob/main/Docker/solver/dockerfile
GitHub Actions: https://github.com/SimVascular/svFSIplus/blob/main/.github/workflows/docker_solver_build.yml

ppraegla:
Thanks @mrp089! I will have a look.

sebproell:
@ppraegla close?

ppraegla:
> @ppraegla close?

Yes, everything has been done in #19 and #24 .

PR #19, opened 2024-11-14
Allow manual build of the docker image

ppraegla:
This is the first part of the experiments in PR #17 . Once the `docker.yml` is on the default branch we can try to manually trigger the docker build and I will create another PR were we can try the docker update mechanism.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,24 +1,57 @@
 name: Docker
 
 on:
-  pull_request:
-    branches:
-      - 'main'
-    paths:
-      - '.github/workflows/docker.yaml'
-      - 'dependencies/current/'
-      - 'dependencies/testing/'
+  workflow_dispatch:
 
 env:
   REGISTRY: ghcr.io
   IMAGE_SUFFIX: dependencies
   FOUR_C_DOCKER_DEPENDENCIES_HASH: 7a6ad12e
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -1,24 +1,67 @@
...
-      - 'dependencies/current/'
-      - 'dependencies/testing/'
+  workflow_dispatch:
+  # pull_request:
+  #   branches:
+  #     - 'main'
+  #   paths:
+  #     - '.github/workflows/docker.yml'
+  #     - 'dependencies/current/'
+  #     - 'dependencies/testing/'
```
amgebauer, line 11:
```suggestion
```

.github/workflows/docker.yml
```diff
@@ -1,24 +1,67 @@
...
 
+# TODO find a way to share these variables across the workflows
```
amgebauer, line 13: I don't think there is a way currently.

```suggestion
```

amgebauer (commented):
I cannot tell much about the remaining todos. What should we do with them?

.github/workflows/docker.yml
ppraegla: I removed the TODOs. The docker.yml needs a final overhaul anyway.

amgebauer (approved):

PR #21, opened 2024-11-15
Add buildtest summary

amgebauer:
Add a job after buildtest that fails if one prior job failed.

This is the simplest way (I'm aware of) to tell Github that all jobs must succeed.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -110,3 +110,13 @@ jobs:
           path: |
             junit_test_summary-${{ matrix.test-chunk }}.xml
           retention-days: 1
+  ensure_all_tests_pass:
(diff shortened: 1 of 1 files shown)


.github/workflows/buildtest.yml
```diff
@@ -110,3 +110,13 @@ jobs:
           path: |
             junit_test_summary-${{ matrix.test-chunk }}.xml
           retention-days: 1
+  buildtest_summary:
```
sebproell, line 113:
```suggestion
  ensure_all_tests_pass:
```
Or something similar. If you call it "summary", I'd assume it is somehow writing the summary somewhere.

sebproell (approved):

PR #23, opened 2024-11-15
CI: better chunking of tests

sebproell:
Add a script to determine better test chunks. For now, I ran it manually and updated the settings in the workflow.


amgebauer (approved):
Nice to see how fast we can get if we choose smarter chunks.

I would not invest much more time in creating these groups automatically. I would rather like to see groups by actual physics modules and would accept slightly longer workflow runs for that.

PR #24, opened 2024-11-18
Update the docker image

ppraegla:
This PR updates the docker image. I removed some unnecessary files from the docker image such that we save 3 GB for the resulting image. It is still quite large with 8.7 GB. But for now this should get us around disk space problems on the runner when building the image.

This is a follow-up to #19 .  I pushed the branch directly into the 4c repo in order to be able to use a manual workflow to update the docker image. After that the pipeline triggered by this PR should work :crossed_fingers: 

The update process is quite annoying at the moment because we cannot use a variable for `container.image` such that you need to manually replace all occurrences when you update the docker hash. At the moment, this is quite brittle as the I only check that the pulled files inside the repo match the environment variable, defined in the workflow. So, it can still happen that you run the wrong container if you forget to adapt the image name.

In a follow-up PR I will adapt the check such that we check the docker hash against the actually used image and not the error-prone environment variable in the workflow.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -9,9 +9,7 @@ on:
       - 'main'
 
 env:
-  REGISTRY: ghcr.io
-  IMAGE_NAME: 4c-dependencies
-  FOUR_C_DOCKER_DEPENDENCIES_HASH: 7a6ad12e
+  FOUR_C_DOCKER_DEPENDENCIES_HASH: 40d8b295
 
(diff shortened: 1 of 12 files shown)


ppraegla:
The actions fail successfully. Hopefully, the succeed once the manually trigger docker actions succeeds https://github.com/4C-multiphysics/4C/actions/runs/11890363608

sebproell (approved):

PR #25, opened 2024-11-18
Docker add hash as metadata

ppraegla:
Add the dependencies hash as metadata (docker label + environment variable) to the docker image in order to be able to check that the actions use the correct image.
This prevents that we run on the wrong docker image if we forget to adapt one of the `container.image`.

Additionally, we make use of the docker/metadata-action to include more metadata (like repo url, created_date, commit_hash) into the docker image.

And, once the docker update PR is merged, a job runs on main that takes the current docker image with the dependencies hash and tags it as `latest` such that the current dependencies image can be pulled more conveniently.


ppraegla:
I adapted the workflows with the new dependencies hash and added some errors
- :white_check_mark: `clang-tidy`: Runs on the wrong (old) image and we use the old check against `env.FOUR_C_DOCKER_DEPENDENCIES_HASH`. The old check should pass and we run clang-tidy on the old image. See [job output](https://github.com/4C-multiphysics/4C/actions/runs/11899682760/job/33158846936?pr=25)
- :white_check_mark: `verify_headers`: Runs on the wrong (old) image, but we use the new check against the `DEPENDENCIES_HASH` inside the image (which does not exist in the old image and should therefore fail). See [job output](https://github.com/4C-multiphysics/4C/actions/runs/11899682760/job/33158847659?pr=25)
- :white_check_mark:  `gcc9_build`: Should be correctly adapted and succeed the check (and then fail due to our supplied docker image). See [job output](https://github.com/4C-multiphysics/4C/actions/runs/11899682731/job/33158846881?pr=25)

It seems that everything works as expected. The new check catches when we are running on the wrong image.

amgebauer:
I would add gcc14 and increase the timeout limit of the job :thinking:

amgebauer:
I'm fine. Start the job over night and we can have the discussion how we can integrate gcc14. Maybe it's also an option to provide a second image with ubuntu 24.04

.github/workflows/documentation.yml
```diff
@@ -49,9 +46,11 @@ jobs:
           retention-days: 1
 
   readthedocs:
+    env:
+      CMAKE_PRESET: docker
```
amgebauer, line 50: Could you use this also on line 74 at configuring 4C?

amgebauer (approved):

amgebauer (approved):

PR #29, opened 2024-11-21
Update docker image for testing to Ubuntu 24.04 noble

ppraegla:
Switch to Ubuntu 24.04 in order to make use of a newer gcc compiler. In a follow-up PR, I will add a nightly test that also runs on the Ubuntu 20.04 image we used before.


ppraegla:
FYI: @sebproell The `gcc` package in Ubuntu 24.04 only provides gcc 13.2.0 and not gcc 14.

amgebauer:
In ubuntu 24.04, gcc14 should at least be in the main repo: https://askubuntu.com/a/1163021

ppraegla:
True, but then we cannot easily re-use the same Dockerfile for Ubuntu 20.04 and 24.04 by just switching `focal` with `noble`.

Configuring failed anyway on Ubuntu 24.04. So, I will fix that first.

Why did we choose gcc-14 again instead of gcc-13 for the additional testing?

sebproell:
> Why did we choose gcc-14 again instead of gcc-13 for the additional testing?

The main webpage claimed that Ubuntu 24.04 would ship GCC14 as default. I guess 13 will also work for us.

sebproell (approved):

PR #31, opened 2024-11-22
Add full tests every night

amgebauer:
Adds the following jobs:

 * gcc9_build + test
 * clang18_build + test
 * gcc9_asan_build + test

Missing:
 * assertion-jobs

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -1,0 +1,186 @@
+name: nightly_tests
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


amgebauer:
The full tests took 3h18m: https://github.com/4C-multiphysics/4C/actions/runs/11973040839

Now ready for review.

amgebauer:
> Nice! Who will be notified when the nightly tests fail?

I think only me since I created the workflow. We may need to add a job that posts the summary of failures in Slack.

PR #32, opened 2024-11-27
Build Ubuntu 20.04 and 24.04 based images and use them for testing

amgebauer:
Add builds for Ubuntu 20.04 and 24.04 and use them for testing. Additionally, llvm is installed from the apt repository and not by manuall build.

This PR adds 4C dependencies images based on Ubuntu 20.04 (for gcc9) and adds tests on Ubuntu20.04 (gcc9) and Ubuntu24.04 (gcc13 and clang18).

LLVM is not now installed by a manual build but via apt-get directly, which saves us a lot time. Docker image builds now in around 2 h.

PR testing now additionally contains gcc9 (with minimal tests) and the gcc13 is now a build with assertions.


docker/Dockerfile
```diff
@@ -5,7 +5,9 @@
 #
 # SPDX-License-Identifier: LGPL-3.0-or-later
 
-FROM ubuntu:noble
+ARG UBUNTU_VERSION=20.04
+FROM ubuntu:${UBUNTU_VERSION}
+ARG UBUNTU_VERSION=20.04
 LABEL org.opencontainers.image.description="Image containing all the dependencies required for building and testing 4C"
```
sebproell, line 11: Why does the `UBUNTU_VERSION` arg appear twice?

docker/Dockerfile
```diff
@@ -36,7 +38,6 @@ RUN apt-get update && apt-get install -y \
       wget \
       && \
     apt-get update && apt-get install -y \
-      cxxtest \
```
sebproell, line 39: :laughing:

docker/Dockerfile
```diff
@@ -90,12 +95,6 @@ RUN /dependencies/current/backtrace/install.sh ${INSTALL_DIR}
 ENV FOUR_C_TESTING_DEPENDENCIES_DIR="/opt/4C-dependencies-testing/"
 RUN mkdir ${FOUR_C_TESTING_DEPENDENCIES_DIR}
 
-# Install LLVM (use standard install directory)
-RUN /dependencies/testing/llvm/install.sh /usr/local
-
-# Install newer GCC
-# RUN /dependencies/testing/gcc/install.sh ${FOUR_C_TESTING_DEPENDENCIES_DIR}
```
sebproell, line 97: Should we remove the respective scripts?

sebproell (commented):
Very nice!

docker/Dockerfile
amgebauer: ARGs are cleared after FROM. But you are rightm we don't need `UBUNTU_VERSION` in this particular case afterwards, so we can remove it here.

PR #33, opened 2024-11-27
Add pre-commit check for dependencies hash in our workflows

amgebauer:
We need to manually set the dependencies hash in our workflows and keep them up to date. If we forget to keep all of them up to date, we might see the failure only after merge. This PR adds a simple check that all dependencies hashes are up to date in all workflows.


utilities/four_c_utils/src/four_c_utils/check_workflow_dependencies_hash.py
```diff
@@ -0,0 +1,60 @@
...
+
+import subprocess
+import os
+import yaml
+import sys
+
+
+def compute_dependencies_hash():
+    result = subprocess.run(
+        r"find dependencies docker -not -wholename '*/trilinos_develop/*' -not -name 'README.md' -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8",
```
sebproell, line 16: We are duplicating this line in a few places already. Can we extract it somewhere into a script?

amgebauer: True. I'll do it once we have synchronized with Gitlab.

ppraegla:
@amgebauer 
>If we forget to keep all of them up to date, we might see the failure only after merge

Isn't this case already covered by the dependencies check in the Actions?

amgebauer:
>>If we forget to keep all of them up to date, we might see the failure only after merge
>
>Isn't this case already covered by the dependencies check in the Actions?

yes, but if you don't update it in the nightly_tests workflow, you will only realize it after merge.

ppraegla:
> > > If we forget to keep all of them up to date, we might see the failure only after merge
> > 
> > 
> > Isn't this case already covered by the dependencies check in the Actions?
> 
> yes, but if you don't update it in the nightly_tests workflow, you will only realize it after merge.

Yes, of course. I did not look at the new nightly tests yet. Thanks!

amgebauer:
Close, in favor of #80 (I forgot that this one here is still open :see_no_evil:).

PR #35, opened 2024-11-28
Optimize struct handling in MulitplicativeSplitDefgradElastHyper material

dragos-ana:
## Description and Context

The current management of the struct instance `kinematic_quantities` in `src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.cpp` is not optimal, as some of its variables are modified during the associated method calls, while others should remain `const` throughout.  Additionally, this struct also contains both elastic stiffness components and stress factors, which are not kinematic quantities. This PR addresses these issues, and  the following work has been performed:

- Removed the elastic stiffness components from `KinematicQuantities`.
- Removed the $\gamma_i$ and $\delta_i$ stress factors  from `KinematicQuantities` and added them to a separate `StressFactors` struct. 
- Aside from `evaluate_kin_quant_elast`, which evaluates all of its variables, the struct instance of `KinematicQuantities` is now only passed by const reference to the material methods. The same applies for the struct instance of `StressFactors`. Furthermore,  `evaluated_s_d_ifin` and `evaluate_additional_cmat` have been reformulated to return the computed objects directly.

### DIFF
--- a/unittests/mat/4C_multiplicative_split_defgrad_elasthyper_test.cpp
+++ b/unittests/mat/4C_multiplicative_split_defgrad_elasthyper_test.cpp
@@ -417,8 +417,7 @@ namespace
     const double concentration(44327.362);
     set_concentration_to_inelastic_material(concentration);
 
-    // actual call that is tested
-    Core::LinAlg::Matrix<6, 6> cMatAdd(true);
+
     // reference solution
     Core::LinAlg::Matrix<6, 6> cMatAdd_ref;
     cMatAdd_ref(0, 0) = -0.701208493301168;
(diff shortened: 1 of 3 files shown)


unittests/mat/4C_multiplicative_split_defgrad_elasthyper_test.cpp
```diff
@@ -417,8 +417,7 @@ namespace
     const double concentration(44327.362);
     set_concentration_to_inelastic_material(concentration);
 
-    // actual call that is tested
-    Core::LinAlg::Matrix<6, 6> cMatAdd(true);
+    ;
```
sebproell, line 420:
```suggestion
```

sebproell (commented):
Can you squash the meaningless second commit?

dragos-ana:
> Can you squash the meaningless second commit?

Done. Thanks for the review, @sebproell !

sebproell (approved):

c-p-schmidt (approved):
Looks good to me.

PR #40, opened 2024-11-28
Remove duplication of dependencies hash generator

amgebauer:
Add a small script that computes the hash. It is reused at all places where we need to do so.

Please carefully check whether I did everything correct, especially on those checks that didn't run for these changes.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -25,8 +25,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Check docker hash
         uses: ./.github/actions/compute-and-check-dependencies-hash
-        with:
-          docker_image_hash: $DEPENDENCIES_HASH
       - uses: ./.github/actions/build_4C
         with:
           cmake-preset: docker_assertions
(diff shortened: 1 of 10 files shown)


sebproell (approved):
LGTM

.github/workflows/buildtest.yml
```diff
@@ -16,17 +16,17 @@ jobs:
...
       run:
         shell: bash
     steps:
       - uses: actions/checkout@v4
       - name: Check docker hash
-        uses: ./.github/actions/compute-and-check-dependencies-hash
+        uses: ./.github/actions/compute-dependencies-hash
         with:
-          docker_image_hash: $DEPENDENCIES_HASH
+          assert_running_in_correct_image: 'true'
```
ppraegla, line 29: Does it make sense to flip the default to true? The input line can easily be forgotten which would mean that we do not check the hash. If the default is true, forgetting the line is just annoying but causes no harm. WDYT?

amgebauer: I just flipped the defaults because of the name. With the name `compute-dependencies-hash` I would not expect to check the hash. But I agree, checking by default is more safe. I will change the default (and maybe the name)

ppraegla (approved):

sebproell (approved):

PR #41, opened 2024-11-29
Fix a bug where multiscale mat is freed too late

sebproell:
Might fix https://gitlab.lrz.de/baci/baci/-/issues/1347

FYI @maxfirmbach


mayrmt (approved):

sebproell:
Who can take a second look?

PR #45, opened 2024-11-29
Draft: Enable contact support for poroscatra

vryy:
## Description and Context

Allow to use contact in `poroscatra` to compute scalar flux. Currently, only the partitioning algorithm is supported.

!2504 

## Checklist

- [x] These changes maintain backwards compatibility.


vryy:
@SchaefP1 could we add a simple test with two elements and interface flux?

SchaefP1:
> @SchaefP1 could we add a simple test with two elements and interface flux?

The problem is that in the current state some poro related things are not done correctly which can lead to it not converging. I localy have some changes that would fix that including test cases. I think it's best to add that once #184 is merged as it depends on each other.

PR #80, opened 2024-11-29
Add dependencies hash checker in pre-commit

amgebauer:
This check helps to always keep it up to date.

Note, this is mostly a convenience check for developers as the hash is already checked in all jobs.


sebproell (dismissed):
I assume you tested this locally with a wrong hash?

amgebauer:
I added the legacy check for the gitlab-ci file (to avoid the failures that we currently see on Gitlab).

This PR should also fix the pipelines on Gitlab.

ppraegla (approved):

sebproell (approved):

Issue #82, opened 2024-11-29
Replace Epetra_Comm with MPI_Comm


sebproell:
_Original issue: https://gitlab.lrz.de/baci/baci/-/issues/1368_

## Description

Currently, we use `Epetra_Comm` as an abstraction built on top of the `MPI_Comm` handle. Since Epetra will be deprecated, this type needs to be changed to something different.

## Possible Solution and Definition of Done

- [x] Provide functions for the methods that are implemented on Epetra_Comm (e.g. SumAll)
- [x] Replace Epetra_Comm with MPI_Comm

PR #89, opened 2024-11-29
Wrap more Epetra_Comm calls

sebproell:
Part of #82 

This PR replaces most functions of the `Epetra_MpiComm` with our own wrapper version and drops the use of `Epetra_SerialComm`. This change uncovered an existing memory bug in Xfluid which I also fixed here.

### DIFF
--- a/src/xfem/4C_xfem_xfluid_contact_communicator.cpp
+++ b/src/xfem/4C_xfem_xfluid_contact_communicator.cpp
@@ -1477,6 +1477,7 @@ void XFEM::XFluidContactComm::create_new_gmsh_files()
   plot_data_.resize(sections.size());
 #endif
 
+  sum_gps_.resize(5);
   std::vector<int> g_sum_gps(5);
   Core::Communication::sum_all(sum_gps_.data(), g_sum_gps.data(), 5, fluiddis_->get_comm());
   if (!Core::Communication::my_mpi_rank(fluiddis_->get_comm()))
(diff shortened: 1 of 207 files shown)


src/xfem/4C_xfem_xfluid_contact_communicator.cpp
```diff
@@ -1478,7 +1478,7 @@ void XFEM::XFluidContactComm::create_new_gmsh_files()
 #endif
 
   std::vector<int> g_sum_gps(5);
-  fluiddis_->get_comm().SumAll(sum_gps_.data(), g_sum_gps.data(), 5);
+  Core::Communication::sum_all(sum_gps_.data(), g_sum_gps.data(), 5, fluiddis_->get_comm());
```
sebproell, line 1481: There is no resize operation on `sum_gps_`, so its size is not going to be 5. This invokes UB.

sebproell: I now resized this vector appropriately. Maybe this fixes one of the famous random xfluid failures.

amgebauer (approved):

c-p-schmidt (approved):

mayrmt (approved):
@sebproell Looks good! 

Will there be a point in the near future, where one must never store a pointer to the communicator or its properties such as size or rank on any class? I believe we do have quite some `comm_`, `my_rank_`, ... class members.

sebproell:
@mayrmt 

> Will there be a point in the near future, where one must never store a pointer to the communicator or its properties such as size or rank on any class? I believe we do have quite some `comm_`, `my_rank_`, ... class members. 

The only change I mean to make with #82, is ro replace all Epetra_Comm occurrences with MPI_Comm. You can and should still store or pass the relevant communicator where needed. It's just a different and simpler type which is not part of Epetra. While we do this, we might as well change the interface of our MPI functions to more modern C++, but this is a side project for the Tpetra migration.

Issue #56, opened 2024-11-29
Further improve testing framework


davidrudlstorfer:
_Ported from Gitlab. Original issue [here](https://gitlab.lrz.de/baci/baci/-/issues/1363)_

## Description

The current testing framework has a few internal inconsistencies, design flaws and drawbacks. This issue aims to collect issues which will be resolved step-by-step to improve our testing strategy.

## Possible Solution and Definition of Done

- [x] Improve `four_c_test_result_file`, i.e., merge with `four_c_test` (see !2547) -> completed in !2584
- [x] Rename `TestingFramework.cmake` to more descriptive file name, i.e., `four_c_testing_functions.cmake`
- [x] Move `TestingFramework.cmake` to better place, i.e., `/cmake/functions/`
- [x] Rename `TestingFrameworkListOfTests.cmake` to more descriptive file name
- [x] ~~Possibly split `TestingFrameworkListOfTests.cmake` into regular .dat file tests and special tests such as cut tests, ...~~
- [x] Move `TestingFrameworkListofTests.cmake` to better place, i.e., somewhere in tests and closer to the actual .dat files
- [x] Alphabetically sort .dat file tests within in `TestingFrameworkListsofTests.cmake` for better overview, check if a pre-commit check would be helpful?
- [ ] Perform test cleanup prior to test run and not at the end of testing => keep test results after each run
- [x] Check if currently set labels are necessary, i.e., `muelu`, `amgnxn`, ...
- [x] Add .dat file ending to TEST_FILE for better understanding what files are used => !2636 
- [ ] Remove specific first names from tests, i.e., `cut_test_alex_39.cpp`
- [ ] TBD

## Interested Parties

@sebproell

davidrudlstorfer:
_Ported comment by @davidrudlstorfer_ 

After the work in https://gitlab.lrz.de/baci/baci/-/merge_requests/2588 the testing framework is cleaned-up even further.

In cooperation with @sebproell we decided on the following layout for a better and more flexible testing layout without changing the underlying infrastructure too much.

One of the main current drawbacks is the complicated and lengthy syntax necessary for each test definition. With the following concept we envision the following:

```
four_c_test(TEST_FILE elch_3D_hex8_line2_scl_2dof.dat NP 2 RETURN_AS elch_test_1)
_four_c_test_add_restart(BASED_ON elch_test_1 RESTART_STEP 123)
[...]

PR #96, opened 2024-12-01
Add docker image with prebuilt 4c

ppraegla:
This PR adds a new docker image where 4C is already built. The plan is to create the docker image once the nightly workflow succeeds. Since we do not have versioning yet, the image is always tagged as `latest`. This docker image can be used in Actions where a 4C build is required or for users who want to quickly check out 4C without needing to build it themselves.

~~This is a first draft as I wanted to get your feedback first, before rebuilding the `4c-dependencies` image because I moved a Dockerfile.~~

The Dockerfile for the dependencies image has been moved to a separate folder in #112. The Dockerfile to build 4c is added to the new folder `prebuilt_4C`.

## Additional info
The pre-built 4C docker image just uses the `4c-dependencies-ubuntu24.04` image out of convenience for now. Originally, I wanted to use a multi-stage build where I just copy the executable to the final image. But then the tests did not work because they need the dat file from the source. I think as a start, we can have the source and build in the final docker image. Later we can focus how we can reduce the size of the docker image because 7.4 GB is huge.

@isteinbrecher  @gilrrei  @davidrudlstorfer  I think this image is a good starting point for further discussions for your use cases in Queens and for the actions meshpy.

## How to use the 4c docker image
Start the docker container with the current directory mounted to `/workspace`
```
docker run --interactive --tty --mount type=bind,source=./,target=/workspace ghcr.io/ppraegla/4c:latest
```
Set the correct owner of the mount (Unfortunately, we need to do that because we are the user `user` inside the docker image and the mount point is owned by root)
```
sudo chown user:user -R /workspace/
```
Run 4C with an input file from the host system and the output is also written to the host system
```
mpirun -np 2 /home/user/4C/build/4C /workspace/particle_dem_1d_radius_from_input.dat /workspace/out
```

This whole process is still a little awkward to use. But may be a good starting point to work with.

### DIFF
--- a/.github/workflows/docker_prebuilt_4c.yml
+++ b/.github/workflows/docker_prebuilt_4c.yml
@@ -1,0 +1,62 @@
+name: 4C docker image
+
+on:
+  workflow_run:
+    workflows: [nightly_tests]
+    types: [completed]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ghcr.io/4c-multiphysics/4c
+
+jobs:
+  build_4c_image:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Log in to the Container registry
+        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      # This step uses [docker/metadata-action](https://github.com/docker/metadata-action#about) to extract tags and labels that will be applied to the specified image. The `id` "meta" allows the output of this step to be referenced in a subsequent step. The `images` value provides the base name for the tags and labels.
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81 # v5.5.1
+        with:
+          images: ${{ env.IMAGE_NAME }}
+          labels: |
+            org.opencontainers.image.description=Image containing the built 4C code
+      # This step uses the `docker/build-push-action` action to build the image, based on your repository's `Dockerfile`. If the build succeeds, it pushes the image to GitHub Packages.
+      # It uses the `context` parameter to define the build's context as the set of files located in the specified path. For more information, see "[Usage](https://github.com/docker/build-push-action#usage)" in the README of the `docker/build-push-action` repository.
+      # It uses the `tags` and `labels` parameters to tag and label the image with the output from the "meta" step.
+      - name: Build and push Docker image
+        id: push
+        uses: docker/build-push-action@4f58ea79222b3b9dc2c8bbdd6debcef730109a75 # v6.9.0
+        with:
+          context: .
+          file: docker/prebuilt_4C/Dockerfile
+          push: true
+          tags: ${{ env.IMAGE_NAME }}:latest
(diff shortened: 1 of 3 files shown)


sebproell:
> One could copy input files to the build directory during configuration. Then, the build and its tests are independent of the source location. Other projects do that.

@mayrmt Yes, we could do this. I normally configure all my necessary support files into the build tree to have a self-sufficient build folder after building. I think I already talked with @davidrudlstorfer about that in the context of #56.

@ppraegla The question would also be if we really need/want to configure and build all the tests in the docker image. If the image is only meant for users to get quick access to a pre-built 4C, the test suite is not going to be very interesting for them. This might depend on the envisioned use case, I just wanted to mention that we might not need to solve this problem at all.

docker/prebuilt_4C/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+FROM ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:latest
+LABEL org.opencontainers.image.description="Image containing the built 4C code"
+LABEL org.4c-multiphysics.project=4C
+
+RUN git clone https://github.com/4C-multiphysics/4C.git
+
+RUN mkdir 4C/build && cd 4C/build && \
+  cmake ../ --fresh --preset=docker && \
+  cmake --build . --target full -- -j `nproc`
```
sebproell, line 16: Once #44 works, we could install it here and remove the build artifacts.

sebproell (commented):
Nice! Amazing how little code and configuration this requires. Our work on the build system and CI infrastructure really pays off here.

ppraegla:
@sebproell @mayrmt I also thought about what makes sense to include in the docker image (only the plain executable or also the source with the ability to run tests). Since we still have all the testing dependencies in the image, I just included the source as well (This was also the easiest solution). But I think it would make sense to create a bare minimum image with only the exectuable.

sebproell:
@ppraegla I am also OK with the current state as a first working version. Feel free to build the image.

isteinbrecher:
@ppraegla thanks a lot for doing this! I will have a look on how to integrate this into the MeshPy testing. This will hopefully save a lot of manual work in the future.

isteinbrecher:
I just tried it myself, and I am amazed how easy it is to run 4C (compared to 6 years ago, when I first saw it). I am very much in favour of putting this high up on our "getting started with 4C" instructions. 🚀

ppraegla:
The `workflow_run` will only work once it is on the default branch. But I tested the `workflow_run` [here](https://github.com/ppraegla/github-actions-playground/actions/runs/12175810683) which triggered [this](https://github.com/ppraegla/github-actions-playground/actions/runs/12175813877) workflow. So, it should be good to go :crossed_fingers:

docker/prebuilt_4C/Dockerfile
ppraegla: Good point. We need to clean up the docker image anyway once we know what exactly we want.

sebproell (approved):

isteinbrecher (approved):
Thanks a lot again!

.github/workflows/docker_prebuilt_4c.yml
```diff
@@ -0,0 +1,62 @@
...
+          # tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+      # This step generates an artifact attestation for the image, which is an unforgeable statement about where and how it was built. It increases supply chain security for people who consume the image. For more information, see "[AUTOTITLE](/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds)."
+      - name: Generate artifact attestation
+        uses: actions/attest-build-provenance@ef244123eb79f2f7a7e75d99086184180e6d0018 # v1.1.4
+        with:
+          subject-name: ${{ env.IMAGE_NAME }}
+          subject-digest: ${{ steps.push.outputs.digest }}
+          push-to-registry: true
```
amgebauer, line 62: Can we generalize the action `build_dependencies` that we already have? We could call it `build_and_publish_docker_image` (or so) and pass the base image, Dockerfile, imagename etc.

For dependencies, we pass `ubuntu:2{0,4}.04` and for this one here, we can pass our dependencies image.

amgebauer (commented):
Nice work!

amgebauer (approved):

PR #99, opened 2024-12-02
Reduce Gitlab CI to minimum and remove utilities

amgebauer:
This PR reduces the Gitlab CI to a minimum (coverage report and trilinos develop) and removes some utility scripts only used by the Gitlab CI.


sebproell (approved):
Do we need to test this? I am fine with merging and fixing if things should fail on Gitlab.

mayrmt (approved):
I am ok with merging as is.

amgebauer:
I at least used the pipeline syntax check on Gitlab. Let's try and see whether there are any problems.

PR #102, opened 2024-12-02
Improve substepping implementation in InelasticDefgradTransvIsotropElastViscoplast

dragos-ana:
## Description and Context

The implementation of substepping in the time integration of InelasticDefgradTransvIsotropElastViscoplast currently uses try-catch mechanisms, with explicit throwing of exceptions to trigger a new substep. As discussed in [2646](https://gitlab.lrz.de/baci/baci/-/merge_requests/2646#note_3454413), it is desirable to avoid this type of mechanism. Moreover, simulations using the current implementation have shown various convergence problems and "random" overflow errors. This PR addresses the issues outlined herein. Specifically, the following work was performed:

- Categorized the types of errors that can be encountered in the time integration, and in the additional stiffness calculation (contribution to the global linearization).
- Removed the try-catch handling and improved the substepping management by using an error status integer indicating which of the possible errors were encountered. Thus, substepping is now triggered using this error status. This also applies for the perturbation-based linearization, which is triggered if an error is encountered in the analytical linearization. The perturbation-based linearization can now be found in a separate function `evaluate_additional_cmat_perturb_based`.
- The overflow error checking is now also performed for the logarithmic substepping, but with other numerical bounds since the computation of the update tensor is not required. The "magic" small value of the time step used previously by default for logarithmic substepping (1.0e-100) was removed.
- Also added overflow error checking for the derivative evaluation of the implemented Reformulated Johnson-Cook law, by specifically calculating the logarithm of both derivatives and checking their numerical values against set numerical bounds. This, together with the previous point, solved the problem of "random" overflow errors.
- Several other small changes to improve the code readability, such as the addition of the dedicated `check_predictor` method or the `SubstepParams` struct containing substepping params. 

[2646](https://gitlab.lrz.de/baci/baci/-/merge_requests/2646#note_3454413)

### DIFF
--- a/src/mat/4C_mat_inelastic_defgrad_factors.cpp
+++ b/src/mat/4C_mat_inelastic_defgrad_factors.cpp
@@ -26,6 +26,7 @@
 #include <Teuchos_ParameterList.hpp>
 #include <Teuchos_StandardParameterEntryValidators.hpp>
 
+#include <map>
 #include <memory>
 #include <stdexcept>
 #include <utility>
@@ -33,6 +34,53 @@
 FOUR_C_NAMESPACE_OPEN
 namespace
 {
+  // enum class for error types in InelasticDefgradTransvIsotropElastViscoplast, used for triggering
+  // the substepping procedures
+  enum class ErrorType
+  {
+    NegativePlasticStrain = 1,  // negative plastic strain which does not allow for evaluations
+                                // inside the viscoplasticity laws
+    OverflowError =
+        2,  // overflow error of the term \f$ \Delta t \dot{\varepsilon}^{\text{p}} \f$ (and \f$
+            // \mathsymbol{E}^{\text{p}}  = \exp(- \Delta t \dot{\varepsilon}^{\text{p}}
+            // \mathsymbol{N}^{\text{p}}) \f$) checked in the standard substepping procedure
+    NoPlasticIncompressibility = 3,  // no plastic incompressibility, meaning that our determinant
+                                     // of the inelastic defgrad is far from 1
+    FailedSolLinSystLNL =
+        4,                 // solution of the linear system in the Local Newton-Raphson Loop failed
+    NoConvergenceLNL = 5,  // the Local Newton Loop did not converge for the given loop settings
+    SingularJacobian = 6,  // singular Jacobian after converged LNL, which does not enable our
+                           // analytical evaluation of the linearization
+    FailedSolAnalytLinearization =
+        7,  // solution of the linear system in the analytical linearization failed
+  };
+
+  // map: error types to error messages in InelasticDefgradTransvIsotropElastViscoplast
+  std::map<ErrorType, std::string> ErrorMessages = {
+      {ErrorType::NegativePlasticStrain,
+          "Error in InelasticDefgradTransvIsotropElastViscoplast: negative plastic strain!"},
+      {ErrorType::OverflowError,
+          "Error in InelasticDefgradTransvIsotropElastViscoplast: overflow error related to the "
+          "evaluation of the plastic strain increment!"},
+      {ErrorType::NoPlasticIncompressibility,
+          "Error in InelasticDefgradTransvIsotropElastViscoplast: plastic incompressibility not "
+          "satisfied!"},
+      {ErrorType::FailedSolLinSystLNL,
+          "Error in InelasticDefgradTransvIsotropElastViscoplast: solution of the linear system in "
+          "the Local Newton Loop failed!"},
+      {ErrorType::NoConvergenceLNL,
+          "Error in InelasticDefgradTransvIsotropElastViscoplast: Local Newton Loop did not "
+          "converge for the given loop settings!"},
+      {ErrorType::SingularJacobian,
+          "Error in InelasticDefgradTransvIsotropElastViscoplast: singular Jacobian after "
+          "converged Local Newton Loop, which does not allow for the analytical evaluation of the "
+          "linearization!"},
+      {ErrorType::FailedSolAnalytLinearization,
+          "Error in InelasticDefgradTransvIsotropElastViscoplast: solution of the linear system in "
+          "the analytical linearization failed"},
+  };
+
+
   // struct: constant non-material tensors
   struct ConstNonMatTensors
   {
@@ -622,6 +670,7 @@ Mat::PAR::InelasticDefgradTransvIsotropElastViscoplast::
       bool_log_substepping_(matdata.parameters.get<bool>("LOG_SUBSTEP")),
       max_halve_number_(matdata.parameters.get<int>("MAX_HALVE_NUM_SUBSTEP"))
 {
+  if (max_halve_number_ < 0) FOUR_C_THROW("Parameter MAX_HALVE_NUM_SUBSTEP must be >= 0!");
 }
 
 
@@ -1716,7 +1765,7 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::calculate_gamma_delta(
 Mat::InelasticDefgradTransvIsotropElastViscoplast::StateQuantities
 Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
     const Core::LinAlg::Matrix<3, 3> &CM, const Core::LinAlg::Matrix<3, 3> &iFinM,
-    const double plastic_strain, const double int_dt, const double check_dt)
+    const double plastic_strain, int &err_status, const double int_dt, const double check_dt)
 {
   StateQuantities state_quantities{};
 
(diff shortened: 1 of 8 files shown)


src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -26,13 +26,61 @@
...
 #include <memory>
 #include <stdexcept>
 #include <utility>
 
 FOUR_C_NAMESPACE_OPEN
 namespace
 {
+  // enum class for error types in InelasticDefgradTransvIsotropElastViscoplast, used for triggering
+  // the substepping procedures
+  enum class ErrorType
```
c-p-schmidt, line 39: One question on the error handling. In other methods, you use an `int` to encode the error. Wouldn't it be nicer and more descriptive to also make use of such an enum?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2878,164 +2850,147 @@ Core::LinAlg::Matrix<10, 1> Mat::InelasticDefgradTransvIsotropElastViscoplast::l
...
-  unsigned int max_num_of_substeps = 1;
+  // initialize substep parameters
+  SubstepParams substep_params = {
+      0,  // t = 0 (time parameter)
+      1,  // substep_counter = 1 (evaluation of first substep)
+      time_step_settings_
+          .dt_,  // curr_dt = time_step_settings_.dt_ (first substep length = full time step)
+      0,         // time_step_halving_counter = 0 (full time step, therefore no substep halving yet)
+      1,         // max_num_of_substeps = 1 (1 substep to evaluate: full time step)
+      0,         // iter = 0 (0 LNL iterations for the current substep)
```
sebproell, line 2861: You can use designated initializers here: 

```suggestion
      .t = 0, 
     .substep_counter =  1, 
     ...
```

src/mat/vplast/4C_mat_vplast_reform_johnsoncook.cpp
```diff
@@ -137,13 +146,26 @@ Mat::Viscoplastic::ReformulatedJohnsonCook::evaluate_derivatives_of_plastic_stra
...
+
+    // check overflow error using these logarithms
+    if ((!log_substep && (std::log(dt) + log_deriv_sigma > 10.0) &&
+            (std::log(dt) + log_deriv_eps > 10.0)) &&
+        (log_substep && (std::log(dt) + log_deriv_sigma > std::exp(10.0)) &&
+            (std::log(dt) + log_deriv_eps > std::exp(10.0))))
+    {
+      err_status = 2;
+      return Core::LinAlg::Matrix<2, 1>{true};
+    }
```
sebproell, line 164: You are computing the same expensive functions, `log` and `exp`, and sometimes longer subexpressiones like `std::log(p) + e * (equiv_stress / yield_strength - 1.0) + std::log(e)` multiple times. This results in way more instructions than you have to do and you could extract the subexpressions.

src/mat/4C_mat_inelastic_defgrad_factors.hpp
```diff
@@ -1565,10 +1620,52 @@ namespace Mat
...
+    /*!
+     * @brief Setup new substep in the Local Newton Loop in case of an encountered evaluation error
+     *
+     * @param[in, out] substep_params parameters of the substepping procedure
+     * @param[out] sol current solution vector of the Local Newton Loop (reset to the last
+     * converged value within this method)
+     * @param[out] curr_CM current right Cauchy-Green deformation tensor, interpolated using
+     * the reference matrices of the time step (interpolated again within this method with the
+     * updated new substep length)
+     * @return error status for the new substep (0: no errors, 1: we have halved the time step too
```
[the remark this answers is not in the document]
dragos-ana: No, I only expect the two states. Thus, I now use a boolean as proposed, thanks!

src/mat/4C_mat_inelastic_defgrad_factors.hpp
```diff
@@ -1502,6 +1507,35 @@ namespace Mat
...
+    {
+      //! current time parameter ranging from 0 to the problem time step \f$ \Delta t \f$
+      double t;
+      //! counter of evaluated substeps
+      unsigned int substep_counter;
+      //! current substep size
+      double curr_dt;
+      //! number of times the problem time step \f$ \Delta t \f$ has been halved
+      unsigned int time_step_halving_counter;
+      //!  current maximum number of substeps to be evaluated within the time step \f$ \Delta t
```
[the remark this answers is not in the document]
dragos-ana: Yes, makes more sense. Thanks!

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -26,13 +26,61 @@
...
 #include <utility>
 
 FOUR_C_NAMESPACE_OPEN
 namespace
 {
+  // enum class for error types in InelasticDefgradTransvIsotropElastViscoplast, used for triggering
+  // the substepping procedures
+  enum class ErrorType
+  {
+    NegativePlasticStrain = 1,  // negative plastic strain which does not allow for evaluations
```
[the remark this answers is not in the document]
dragos-ana: This is helpful since I need to "translate" from an int to an ErrorType when communicating with the viscoplastic law. E.g. when returning the error status from `evaluate_plastic_strain_rate` in  `4C_mat_vplast_reform_johnsoncook.cpp`, I want to convert this int to an error type inside InelasticDefgradTransvIsotropElastViscoplast without providing the ErrorType class also to the viscoplasticity law.

src/mat/vplast/4C_mat_vplast_reform_johnsoncook.cpp
sebproell: Small example where I rearranged the computations to make this specific part 2.5 times faster(I did not even include the overflow check): https://quick-bench.com/q/ELEz4rotYR56n1y1IThkiFlZxWU

You don't need to do that in this PR. I just wanted to point out that this code looks really expensive and since you do this on the quadrature point I can imagine these calculations contribute significantly to your total run time.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: As explained below, I use an int to signal an error in the viscoplastic law evaluations anyway, to avoid providing the ErrorType enum class to it as well.  Since the conversion from `int` to `ErrorType` is then performed anyway,  I would see this as a matter of taste whether we pass ErrorTypes or ints to and from the material methods.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2300,164 +2358,97 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
   // viscoplasticity law terms
   double min_dt = time_step_settings_.dt_;
-  if (parameter()->bool_log_substepping())
-  {
-    min_dt = 1.0e-100;  // when using logarithmic substepping, set to extremely small value to
-                        // avoid built-in overflow error check
-  }
+
+  // declare error status (0: no errors)
+  int err_status = 0;
```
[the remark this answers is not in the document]
dragos-ana: This is the error status corresponding to the ErrorType. But as explained in both threads above, a conversion int <-> ErrorType has to be performed anyway in the current implementation of error handling.

src/mat/vplast/4C_mat_vplast_reform_johnsoncook.cpp
dragos-ana: I have included your suggestions now. Thanks a lot for the hint and for the offline discussion!

dragos-ana:
Thank you for the detailed review, @c-p-schmidt  and @sebproell!

c-p-schmidt (approved):

dragos-ana:
Based on the discussions above, I removed the `ErrorType<->int` conversion logic entirely by creating the header file `4C_mat_inelastic_defgrad_factors_service.hpp`, which I provide to both `InelasticDefgradTransvIsotropElastViscoplast` and `Viscoplastic::Law`. Additionally, I implemented some small optimizations, such as replacing division by multiplication where it was possible and reasonable to do so.

c-p-schmidt (approved):

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -0,0 +1,49 @@
...
+    FailedSolLinSystLNL,  // solution of the linear system in the Local Newton-Raphson Loop failed
+    NoConvergenceLNL,     // the Local Newton Loop did not converge for the given loop settings
+    SingularJacobian,     // singular Jacobian after converged LNL, which does not enable our
+                          // analytical evaluation of the linearization
+    FailedSolAnalytLinearization,  // solution of the linear system in the analytical linearization
+                                   // failed
+  };
+
+  /// map: error types to error messages in InelasticDefgradTransvIsotropElastViscoplast
+  extern std::map<Mat::ViscoplastErrorType, std::string> ViscoplastErrorMessages;
```
sebproell, line 42: Should this simply be a function `to_string(ViscoPlastErrorType)`?

sebproell (approved):
Only a minor comment for future improvement.

PR #104, opened 2024-12-03
Compress build directory for artifact

amgebauer:
This adds xz compression for passing the build directory between build- and test stage and removes the archive after reconstructing the build directory.


sebproell (approved):

amgebauer:
Compression is too slow. Let's try to compress with pigz, otherwise let's close it.

amgebauer:
The speed of pigz is good enough, but the compression itself is not worth the overhead.

PR #107, opened 2024-12-03
Remove scatra related implementation from the post processor

c-p-schmidt:
## Description and Context
In https://gitlab.lrz.de/baci/baci/-/merge_requests/2426, the output mechanism of the scalar transport module has been changed to what we currently still call runtime output, i.e., the vtk-based output. However, I forgot to remove the corresponding post-processor implementation.

This already let to confusion since the post-processor can still post-process data that is written for restart, as this is the same format and mechanism as the old output format. This is now removed and only the vtk-based output functionality remains. Thank you for the hint @bwirthl 

Follows https://gitlab.lrz.de/baci/baci/-/merge_requests/2426


mayrmt (dismissed):

sebproell (dismissed):

bwirthl:
@c-p-schmidt Thanks a lot for solving this so quickly!

c-p-schmidt:
The tutorial test failed because our framework tests run `pre_exodus`, `4C`, and then the `post_processor`. Since I removed the `post_processor` support for all scatra-related code, i.e., also solid-scatra-interaction, this is not possible anymore.

As we want to get rid of the post-processor completely, I would suggest that we remove the call to the post-processor in the framework tests. What do you mean?

c-p-schmidt:
The second commit implements my proposal to remove the post_processor from the framework tests.

amgebauer (approved):

PR #112, opened 2024-12-04
Move the Dockerfile for the dependencies image

ppraegla:
Move the Dockerfile for the dependencies to a new folder `docker/dependencies` in preparation of #96 , where I add a new folder for another docker build.

Thanks @amgebauer for the pre-commit checks `check-dependencies-hash` and `check-venv-up-to-date`. They already caught some of my mistakes. So, please check the changes carefully. Especially, since I changed the command to compute the dependencies hash.


amgebauer (approved):
LGTM. One thought, what do you think about if we move the Dockerfile into the dependencies folder?

sebproell (approved):

PR #85, opened 2024-11-29
Remove TSI Nitsche contact and old so3plast elements

amgebauer:
The old so3plast elements are currently not in use and block the new templated implementation of solid elements.

so3plast elements are currently the only elements that are compatible with TSI Nitsche contact. So, we also need to remove it.

Closes #67

Issue #65, opened 2024-11-29
Raise C++ standard to C++20


sebproell:
_Ported from Gitlab. Original issue [here](https://gitlab.lrz.de/baci/baci/-/issues/969)_

## Description

C++20 is out for a while and we should update. We already test with clang 18 and GCC 13 which are both compliant with this standard (in all relevant aspects).

## Possible Solution and Definition of Done
- [x] Set minimum standard in CMake
- [x] Drop GCC 9 compiler from CI
- [x] Replace forward implemented C++ 20 features with actual C++ 20 features
- [x] Check if we can use `-Wpedantic`

sebproell:
Everything is prepared for the upgrade from my side (#130). @maxfirmbach @isteinbrecher Please let me know once you have compilers ready at your institute, so we can finish this up :rocket:

ppraegla:
> @amgebauer @ppraegla When dropping GCC 9, I guess we drop the whole Ubuntu20.04 image?

Yes, sounds reasonable.

PR #114, opened 2024-12-04
Enforce warning-free clang build

amgebauer:
With #85, we can now also enforce the last remaining ignored warning of clang-18


sebproell (approved):
Nice. Will you update the LNM preset in a next PR?  We would be ready for #65 then.

c-p-schmidt (approved):

PR #116, opened 2024-12-04
Set clang as the default compiler at LNM workstations

amgebauer:
Makes LNM ready for #65, has slightly better compile times and more expressive error messages.


sebproell (dismissed):
@amgebauer AFAIK, you used this specific clang build for a few weeks already?

amgebauer:
Yes, it's my default for a few weeks and I have not yet encountered problems besides the solved so3plast warnings #85.

c-p-schmidt (approved):

sebproell:
@amgebauer It seems we would need to update libstdc++ on our system if we want to use clang with C++20. This makes me wonder if Clang is the right way to get C++20 access. We can for sure use it once we update our system and have the newer standard library available. Updating the library manually seems cumbersome (although I didn't try it).

amgebauer:
We can still think about it setting clang as standard already today as it provides better error messages and slightly better compile times compared to the current standard gcc9.4

PR #117, opened 2024-12-04
Use MPI_Comm instead of Epetra_Comm

### LINKED ISSUE #82: Replace Epetra_Comm with MPI_Comm

sebproell:
Completes #82 except for the cases where we still need to interface with other Epetra objects. Since these cases will vanish when we wrap the remaining Epetra objects, this closes #82.


src/core/comm/src/4C_comm_mpi_utils.hpp
```diff
@@ -28,113 +27,131 @@ FOUR_C_NAMESPACE_OPEN
...
+   *
+   * @note Epetra_Comm is a virtual interface and it is not trivial to create an instance on the fly
+   * and return it here due to the lack of value semantics. Trilinos functions often expect a
+   * `const Epetra_Comm&`. Therefore, this function has to maintain a global cache of Epetra_Comm
+   * objects for the respective MPI communicator. They are kept alive for the duration of the
+   * program. Don't try to create this object yourself, as you will likely run into issues with
+   * dangling references.
+   */
+  const Epetra_Comm& as_epetra_comm(MPI_Comm comm);
+
```
sebproell, line 48: This is the only non-trivial change in this PR. I added these two helpers to convert between MPI_Comm and Epetra_Comm in a few places.

src/beamcontact/4C_beamcontact_beam3contact_manager.cpp
```diff
@@ -925,16 +925,16 @@ void CONTACT::Beam3cmanager::init_beam_contact_discret()
...
   sdata.clear();
   stproc.clear();
   rdata.clear();
   allproc.clear();
 
   // build completely overlapping element map (on participating processors)
-  std::shared_ptr<Epetra_Map> newelecolmap = std::make_shared<Epetra_Map>(
-      -1, (int)erdata.size(), erdata.data(), 0, bt_sol_discret().get_comm());
+  std::shared_ptr<Epetra_Map> newelecolmap = std::make_shared<Epetra_Map>(-1, (int)erdata.size(),
+      erdata.data(), 0, Core::Communication::as_epetra_comm(bt_sol_discret().get_comm()));
```
sebproell, line 937: Note that all conversions to `Epetra_Comm` when creating an Epetra_Map will vanish once we wrap this class.

davidrudlstorfer (approved):

c-p-schmidt (approved):

PR #128, opened 2024-12-06
Add LLVM source-based coverage report generated every night

amgebauer:
This PR adds LLVM source based coverage (cf. https://clang.llvm.org/docs/SourceBasedCodeCoverage.html).

It is **much faster** than the coverage report that we used to have. In my fork, it finishes within 2h15m. I have a pro account with 40 parallel runners, so testing might take a bit longer within 4C-multiphysics, but I expect it to be below 3h.

The report itself is also a bit different:
 :heavy_check_mark:  Optimization does not affect coverage (we can use normal optimization)
 :heavy_check_mark: We see coverage for each instantiation of a template (i.e., we see which instantiations are covered, which not, optional, could be disabled if too much)
 :heavy_check_mark: For every execution branch bifurcation (e.g., an `if`-statement) we see the percentage how often a branch was taken
 :x: We don't yet have the option to sort by coverage statistics. Maybe a future version of llvm will be able to do so.
:information_source: Current bottleneck is the storage size of the Github runner (~14 GB). The build-folder already takes 4 GB + Ubuntu image etc., so there is not much storage left for testing and coverage raw output. That's why the test chunks are more fine-grained than during regular testing.
:information_source: Have a look at [the caveats according to exceptions](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#drawbacks-and-limitations).

Here is the generated report from my fork: https://amgebauer.github.io/4C/coverage_report/ (the deps-folder should now be excluded)

I will test the nightly generation and deployment in my fork this night and if it works, we can merge.

We also have to run doxygen and readthedocs generation every night to make sure it is available once we want to deploy.

:warning:  Note, I expect the deploy job to fail after merge until the first coverage pipeline ran successfully.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,162 @@
+name: coverage
+
+on:
+  schedule:
+    - cron: '0 1 * * *'
+
+jobs:
+  gcc13_coverage_build:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:28cfb9d0
+      options: --user root --env OMPI_ALLOW_RUN_AS_ROOT=1 --env OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
+    defaults:
+      run:
+        shell: bash
+    outputs:
+      test-chunks: ${{ steps.set-matrix.outputs.chunk-array }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Check docker hash
+        uses: ./.github/actions/compute-and-check-dependencies-hash
+      - uses: ./.github/actions/build_4C
+        with:
+          cmake-preset: docker_coverage
+          build-targets: full
+          build-directory: ${{ github.workspace }}/build
+          use-ccache: "false"
+      - uses: ./.github/actions/upload_4C_build
+        with:
+          build-directory: ${{ github.workspace }}/build
+          retention-days: 1
+      - uses: ./.github/actions/chunk_test_suite
+        id: set-matrix
+        # Note: We will not use previous runtimes for chunking to avoid running out of storage
+        with:
+          build-directory: ${{ github.workspace }}/build
+          source-directory: ${{ github.workspace }}
+          number-of-chunks: 40 # we need fine grained test chunks to avoid running out of storage
+
+  gcc13_coverage_test:
+    needs: gcc13_coverage_build
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:28cfb9d0
+      options: --user root --env OMPI_ALLOW_RUN_AS_ROOT=1 --env OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
+    strategy:
+      matrix:
+        test-chunk: ${{fromJson(needs.gcc13_coverage_build.outputs.test-chunks)}}
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Install dependencies
+        run: |
+          apt-get update
+          apt-get upgrade -y
+          apt-get install -y llvm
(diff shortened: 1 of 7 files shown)


.github/workflows/documentation.yml
```diff
@@ -7,6 +7,8 @@ on:
   push:
     branches:
       - 'main'
+  schedule:
+    - cron: '0 5 * * *'
```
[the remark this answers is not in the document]
amgebauer: The time itself has no real logic, except that we need to make sure that we run documentation at least once within the `retention-days`. Let me unify this with the nightly_build.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,149 @@
...
+    needs: gcc13_coverage_build
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:28cfb9d0
+      options: --user root --env OMPI_ALLOW_RUN_AS_ROOT=1 --env OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
+    strategy:
+      fail-fast: true
+      matrix:
+        # we need to make very fine grained chunks to avoid running out of storage
+        test-chunk: ['1,20', '21,40', '41,60', '61,80', '81,100', '101,120', '121,140', '141,160', '161,180',
```
sebproell, line 41: I guess this is very brittle due to the small chunk sizes? I am fine with merging but we might need to generate these chunks dynamically.

sebproell (approved):

.github/workflows/coverage.yml
c-p-schmidt: Do you intend to use the automatic chunking from your other PR here?

c-p-schmidt (approved):

.github/workflows/coverage.yml
amgebauer: Yes, once it delivers reliable results after #132 and #129 are merged.

amgebauer:
Now ready for review. I tested the workflow on my fork. The latest version is here https://amgebauer.github.io/4C/coverage_report

This will be the badge: Corverage badge

presets/docker/CMakePresets.json
```diff
@@ -93,9 +93,10 @@
         ".docker_base"
       ],
       "cacheVariables": {
-        "CMAKE_BUILD_TYPE": "DEBUG",
         "FOUR_C_ENABLE_COVERAGE": "ON",
-        "FOUR_C_TEST_TIMEOUT_SCALE": "50"
+        "FOUR_C_TEST_TIMEOUT_SCALE": "4",
```
c-p-schmidt, line 97: Really nice that it is so much faster!

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,162 @@
...
+      - name: Upload merged coverage data
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ github.job }}-4C-coverage-data-${{ matrix.test-chunk }}.profdata
+          path: |
+            ${{ github.workspace }}/4C-coverage-data-${{ matrix.test-chunk }}.profdata
+          retention-days: 1
+      - name: Upload test report
+        uses: actions/upload-artifact@v4
+        if: success() || failure()
```
c-p-schmidt, line 86: Can't that be removed? Looks as if this should always be true.

amgebauer: It is actually necessary. If we remove it completely, it will be executed only on `success()` (default behavior). We can also not use `always()` as we don't want to execute it if the workflow is canceled. This condition here is equivalent to `!cancelled()`

c-p-schmidt (approved):

sebproell (approved):
Awesome! Will you remove the old scripts and the Gitlab setup in a follow-up?

amgebauer:
yes, I will do it tomorrow after the first successful run. Then I can also add the badge to our readme.

PR #129, opened 2024-12-06
Automatically chunk ctest suite based on `test_timeout * num_procs`

amgebauer:
Currently, chunking of the test suite is a bit bittle.

This PR suggests to chunk the test suite based on `test_timeout * num_procs`.

We have the info available directly after the build (in contrast to the real runtime of the tests).

Let's see how good this actually is. I'm quite confident that it will sort out our long-running xfluid-tests.


amgebauer:
Sorry for the spam, but it took a while until I got it to work. Unfortunately, the results are not satisfying. We need to get the runtimes from the last nightly main pipeline and only fall back to the timeouts for new tests.

amgebauer:
Now ready for review. Automatic chunking of the test suite now works based on test runtime in last nightly pipeline, and test timeout for new tests. The chunk now execute between 15m and 18m.

image

.github/actions/chunk_test_suite/action.yml
```diff
@@ -0,0 +1,51 @@
...
+      id: download-junit-reports
+      continue-on-error: true
+      uses: dawidd6/action-download-artifact@v6
+      with:
+        workflow: nightly_tests.yml
+        name: ${{ inputs.junit-report-artifact-name }}
+        path: ${{ inputs.source-directory }}
+        branch: ${{ github.event.repository.default_branch }}
+        workflow_conclusion: success
+        if_no_artifact_found: fail
```
sebproell, line 33: Does this mean that we can never run PR actions if the nightly pipeline fails? This would lead to a deadlock. If this is indeed the case, we need a fallback for the chunks.

sebproell (commented):
I really like the approach! The hard-coded chunks were an eyesore.

.github/actions/chunk_test_suite/action.yml
amgebauer: no, we have `continue-on-error: true` which makes this action end with `success` even if the artifact was not found. I can use the state later with `steps.download-junit-reports.outcome`.

I use this to determine whether we found an artifact. If yes, we add `--junit-report report.xml` in the next step.

Take this as an example: https://github.com/4C-multiphysics/4C/actions/runs/12237519164/job/34174380024
The artifact was not available yesterday, the step itself failed (output: `Error: no artifacts found`), but the job succeeded.

davidrudlstorfer (approved):

PR #130, opened 2024-12-06
Fix C++ 20 incompatibilities and enforce standard

sebproell:
These are the necessary changes to compile with C++ 20. To ensure that the whole setup can actually compile C++ 20, I replaced a few features with C++20 equivalents.

Part of #65


sebproell:
clang-tidy is failing: https://github.com/4C-multiphysics/4C/actions/runs/12243905374/job/34154476947?pr=130#step:6:15

CMake seems to generate some files related to C++ modules (although we are not using them actually).

Related: 

- https://discourse.cmake.org/t/how-to-control-the-location-of-the-c-20-binary-module-interface-bmi-output-directory/7968

CMakeLists.txt
```diff
@@ -30,8 +30,13 @@ message(STATUS "Using CMake ${CMAKE_VERSION}")
...
 set(CMAKE_CXX_STANDARD 20)
-set(CMAKE_CXX_STANDARD_REQUIRED On)
-set(CMAKE_CXX_EXTENSIONS Off)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+# We do not use C++ modules (yet). Turn off scanning to avoid issues with clang-tidy.
+# If you want to add module support, this problem needs to be revisited. Our hope
+# is that CMake and/or clang-tidy will be updated to handle modules better.
+set(CMAKE_CXX_SCAN_FOR_MODULES OFF)
```
sebproell, line 39: This fixes clang-tidy, which, funnily enough, does not understand the modmap files emitted by clang. I am pretty sure this is already fixed or will be fixed in newer versions. We do not have module support anyway.

sebproell:
This works. We obviously need to drop GCC9 before merging.

maxfirmbach:
@sebproell IMCS workstations should now run with `gcc-13.2.0`. If there is anything else institute specific that needs to be updated, just write me.

sebproell:
Ready. Once we merge this, we are supporting and requiring C++20 :tada:

src/core/comm/tests/4C_comm_pack_unpack_test.np3.cpp
amgebauer, line 1: Why are these changes related to c++20?

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -470,6 +470,13 @@ namespace Core::IO
       return *this;
     }
 
+    DatFileLineIterator DatFileLineIterator::operator++(int)
+    {
+      auto copy = *this;
+      ++(*this);
+      return copy;
+    }
+
```
amgebauer, line 479: Why is this now needed?

src/core/utils/src/functions/4C_utils_symbolic_expression.cpp
```diff
@@ -23,21 +23,11 @@ FOUR_C_NAMESPACE_OPEN
...
 {
-  const std::array<std::string, 7> valid_operators = {"+", "-", "*", "/", "^", ".", ","};
-
-  const std::array<std::string, 16> valid_functions = {"acos", "asin", "atan", "cos", "sin", "tan",
-      "cosh", "sinh", "tanh", "exp", "log", "log10", "sqrt", "heaviside", "fabs", "atan2"};
-
-  const std::array<std::string, 2> valid_brackets = {"(", ")"};
-
-  const std::array<std::string, 6> reserved_words = {"pi", "e", "E", "\t", "\n", " "};
-
```
amgebauer, line 34: Also here?

amgebauer (approved):
Looks good, but I don't how some of the changes are related to c++20.

src/core/comm/tests/4C_comm_pack_unpack_test.np3.cpp
sebproell: I thought I had a bug and added more tests. The tests are correct, so I added them.

src/core/io/src/4C_io_input_file.cpp
sebproell: I don't remember exactly, but I think one of the C++ 20 range algorithms requires a proper iterator that also supports post-incrementation.

src/core/utils/src/functions/4C_utils_symbolic_expression.cpp
sebproell: I can't tell you why but the compiler gives an unused variable warning now. They were always unused.

maxfirmbach:
LGTM! I can't approve though :sweat_smile:.

mayrmt (approved):

maxfirmbach (approved):

Issue #121, opened 2024-12-05
Redundant solid material evaluations


dragos-ana:
## Description

The internal force and nonlinear stiffness computations for solid elements ( actions `Core::Elements::struct_calc_internalforce` and `Core::Elements::struct_calc_nlnstiff` in `evaluate` within `4C_solid_3D_ele_evaluate.cpp` ) use two separate calls to the solid material evaluation method `Mat::So3Material::evaluate`. This method computes both the stress and the linearization (and then disregards the one or the other depending on what we currently want to compute, see above). Therefore, we currently have a redundant material evaluation call. For materials which are computationally expensive, this results in significantly higher computation times than actually required. For example, saving the computed state and simply returning the saved linearization in the "redundant step" (while computing the stiffness contribution) led to a performance speedup of roughly factor 2 in my simulations with InelasticDefgradTransvIsotropElastViscoplast. Resolving this redundancy could therefore improve the performance significantly for expensive material evaluations.

## Possible Solution and Definition of Done

Based on a discussion with @amgebauer, we propose two possible solutions:
- [ ] Either fix this within all materials, by separating the stress and stiffness computations (e.g. replace `evaluate` by `evaluate_stress` and `evaluate_nlnstiff`, and maybe store them to not require recomputation for complex materials);
- [ ] Or store the computed contributions in the element evaluation and don't reevaluate them if they are available. This would be faster to implement, but it would lead to computing the stiffness contributions once too much per time step (per element).

I would propose a meeting for all interested parties, so that we can discuss what would be the more suitable approach. Also, feel free to add other proposals to the list. Reply under this thread if you are interested, so that we can organize this meeting.

mayrmt:
@dragos-ana @amgebauer, how would this topic relate to nonlinear solvers that do not always require both residual and linearization, e.g. during a line search in a Newton method?

sebproell:
What I have been doing to evaluate materials is based on designs in deal.II: Use evaluation flags to compute whatever you require and later query it.

Example:

```c++
    void
    reinit_point(unsigned int            q,
                 const SymmetricTensor2& linear_strain,
                 VectorizedArrayType     temperature,
                 const TsiEvalFlags&     flags)
  {
     if (contains(flags, TsiEvalFlags::stress))
       ...
     if (contains(flags, TsiEvalFlags::tanget_stress_strain))
       ...
     if (contains(flags, TsiEvalFlags::tanget_stress_strain))
       ...
```

[...]

PR #131, opened 2024-12-07
Quick fix for the material evaluate calls in the case of InelasticDefgradTransvIsotropElastViscoplast

dragos-ana:
## Description and Context

This PR implements a quick fix for the redundant material evaluations issue #121, specifically for the transversely isotropic material `InelasticDefgradTransvIsotropElastViscoplast`. Concretely, the deformation gradient, which is the input of the material evaluations, is tracked continuously. If an inelastic deformation gradient has been previously computed for a specific deformation gradient, then the solution is simply returned, instead of repeating the substepping scheme (which would yield the exact same result). A performance speed up of approx. 2 was reached with this quick fix in the exemplary simulations performed. I am aware that this quick fix is not the optimal solution. However, it is IMO the best approach until we can find a consistent solution for all materials, see issue above. Moreover, the approach of converting error types to error messages was adapted as proposed [here](https://github.com/4C-multiphysics/4C/pull/102#discussion_r1873757244), by replacing the `std::map` with a `to_string(Mat::ViscoplastErrorType)` function.

Related to Issue #121 and PR #102.

### DIFF
--- a/src/mat/4C_mat_inelastic_defgrad_factors.cpp
+++ b/src/mat/4C_mat_inelastic_defgrad_factors.cpp
@@ -17,6 +17,7 @@
 #include "4C_mat_elast_couptransverselyisotropic.hpp"
 #include "4C_mat_elasthyper_service.hpp"
 #include "4C_mat_electrode.hpp"
+#include "4C_mat_inelastic_defgrad_factors_service.hpp"
 #include "4C_mat_multiplicative_split_defgrad_elasthyper.hpp"
 #include "4C_mat_par_bundle.hpp"
 #include "4C_mat_vplast_law.hpp"
@@ -29,6 +30,7 @@
 #include <map>
 #include <memory>
 #include <stdexcept>
+#include <string>
 #include <utility>
 
 FOUR_C_NAMESPACE_OPEN
@@ -2466,8 +2468,7 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_inverse_inelast
 
     // throw error if the Local Newton Loop cannot be evaluated with the given substepping
     // settings
-    if (err_status != Mat::ViscoplastErrorType::NoErrors)
-      FOUR_C_THROW(ViscoplastErrorMessages[err_status]);
+    if (err_status != Mat::ViscoplastErrorType::NoErrors) FOUR_C_THROW(Mat::to_string(err_status));
 
     // extract the inverse inelastic defgrad from the LNL solution
     iFinM = extract_inverse_inelastic_defgrad(sol);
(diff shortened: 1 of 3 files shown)


src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2436,6 +2441,18 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_inverse_inelast
   Core::LinAlg::Matrix<3, 3> CredM(true);
   CredM.multiply_tn(1.0, FredM, FredM, 0.0);
 
+  // check whether we have already evaluated the inverse inelastic deformation gradient for the
+  // given reduced deformation gradient
+  Core::LinAlg::Matrix<3, 3> diff_defgrad{true};
+  diff_defgrad.update(1.0, FredM, -1.0, time_step_quantities_.current_defgrad_[gp_], 0.0);
+  if (diff_defgrad.norm2() < 1.0e-15)
```
sebproell, line 2448: I am wondering whether exact equality would be more appropriate in the present case. If I understand everything correctly, you really expect the exact same value for the def grad.

Side note: normally you want to scale the tolerance to the magnitude of the values in the deformation gradient.

dragos-ana: Yes, in principle, I do expect the 2-norm to be exactly 0.0. However, I was not quite sure how exact the storing of the computed defgrad is, as well as the norm computation. Modifying it to `if (diff_defgrad.norm2()) == 0.0` leads to the same results in my simulations, so the procedure is quite exact. Thanks for the hint.

sebproell: Well, you are comparing two copies of the same data, so this better be exactly equal ;)

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -1650,10 +1653,14 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::InelasticDefgradTransvIsotrop
   time_step_quantities_.current_plastic_strain_.resize(1, 0.0);  // value irrelevant at this point
   time_step_quantities_.last_substep_plastic_strain_.resize(1, 0.0);
 
-  // default values of the deformation gradient: unit tensor
+  // default values of the right CG tensor: unit tensor
   time_step_quantities_.last_rightCG_.resize(1, const_non_mat_tensors.id3x3_);
   time_step_quantities_.current_rightCG_.resize(
       1, const_non_mat_tensors.id3x3_);  // value irrelevant at this point
+
+  // default value for the current deformation gradient: 0-tensor (to make sure that the inverse
```
c-p-schmidt, line 1661:
```suggestion
  // default value for the current deformation gradient: 2-tensor (to make sure that the inverse
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -38,8 +40,47 @@ namespace Mat
...
-  /// map: error types to error messages in InelasticDefgradTransvIsotropElastViscoplast
-  extern std::map<Mat::ViscoplastErrorType, std::string> ViscoplastErrorMessages;
+  /// to_string: error types to error messages in InelasticDefgradTransvIsotropElastViscoplast
+  inline std::string to_string(Mat::ViscoplastErrorType err_type)
+  {
+    switch (err_type)
+    {
+      case Mat::ViscoplastErrorType::NegativePlasticStrain:
+        return "Error in InelasticDefgradTransvIsotropElastViscoplast: negative plastic strain!";
+        break;
```
c-p-schmidt, line 50: The `break`s are irrelevant here because you anyways return before. Furthermore, the formatting (removing obsolete linebreaks) could be enhanced a bit.

c-p-schmidt (approved):

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-corneliu-ana: Yes, thank you! I will enhance this in a new commit, along with your comment above

c-p-schmidt (approved):

dragos-corneliu-ana (approved):

dragos-ana:
Just checked whether external users can also approve...they can, but it does not "count"

PR #144, opened 2024-12-10
Fix check of NURBS elements in the evaluation of Neumann boundary conditions

eulovi:
## Description and Context

For mixed discretizations using NURBS elements and classic Lagrange finite elements, a bug was found for the evaluation of surface Neumann boundary conditions. If the Neumann boundary conditions were applied on finite elements (for example HEX8), some NURBS routines were applied, leading to segmentation errors. In this PR, this is fixed by checking the element shape instead of checking the discretization type.


src/so3/4C_so3_surface_evaluate.cpp
```diff
@@ -212,6 +208,7 @@ int Discret::Elements::StructuralSurface::evaluate_neumann(Teuchos::ParameterLis
   {
     // --------------------------------------------------
     // get knotvector
+    auto* nurbsdis = dynamic_cast<Core::FE::Nurbs::NurbsDiscretization*>(&(discretization));
```
amgebauer, line 211: why do you cast to a pointer and not to a reference?
```suggestion
    auto& nurbsdis = dynamic_cast<Core::FE::Nurbs::NurbsDiscretization>(discretization);
```

amgebauer (changes requested):

src/so3/4C_so3_surface_evaluate.cpp
sebproell: I think what @amgebauer wants to say is: you can now cast to a reference because you know that the cast will work.

amgebauer: yes. There is no further check anyway.

src/so3/4C_so3_surface_evaluate.cpp
```diff
@@ -189,11 +189,7 @@ int Discret::Elements::StructuralSurface::evaluate_neumann(Teuchos::ParameterLis
 
   // --------------------------------------------------
   // Now do the nurbs specific stuff
-  bool nurbsele = false;
-
-  auto* nurbsdis = dynamic_cast<Core::FE::Nurbs::NurbsDiscretization*>(&(discretization));
-
-  if (nurbsdis != nullptr) nurbsele = true;
+  bool nurbsele = Core::FE::is_nurbs_celltype(shape());
```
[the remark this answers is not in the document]
eulovi: Yes, that name would be better.

src/so3/4C_so3_surface_evaluate.cpp
eulovi: Yes, you're both right. I just applied the dynamic_cast that was previously in the code.

amgebauer (changes requested):
Could you please squash all commits into one? Thanks.

eulovi:
> Could you please squash all commits into one? Thanks.

@amgebauer sure! I thought I could choose the option 'Squash and merge' after this PR is approved but I just noticed this is not allowed.

eulovi:
If the current state is fine for you, could you approve @isteinbrecher, @amgebauer?

amgebauer (approved):

c-p-schmidt (approved):

PR #147, opened 2024-12-11
Fix linearization in implicit remodel fiber and fix some global renaming errors

amgebauer:
This PR fixes the linearization of the implicit remodel-fiber (2nd commit)

Additionally, it fixes the global renaming errors (1st commit).


src/mixture/4C_mixture_remodelfiber-internal.hpp
```diff
@@ -148,22 +188,32 @@ namespace Mixture
...
+      [[nodiscard]] T evaluate_d_current_lambda_r_d_lambda_f_sq() const;
+      [[nodiscard]] T evaluate_d_current_cauchy_stress_d_lambda_f_sq() const;
       /// @}
 
       /// array of G&R states (the last state in the array is the current state)
       std::array<GRState, numstates> states_;
 
+      /// current derivatives w.r.t. lambda_f
+      /// @{
+      T d_growth_scalar_d_lambda_f_sq_ = 0.0;
```
lauraengelhardt, line 213: I guess the `sq` means squared, right? Is this this the derivative of the `growth_scalar` w.r.t. to the square of `lambda_f` or the square of the derivative of the `growth_scalar` w.r.t. to `lambda_f`?

lauraengelhardt (dismissed):
I only left one minor comment - you can decide whether this is relevant or too detailed :) 
Code looks good to me - I cannot comment on the correctness of the linearization.

src/mixture/4C_mixture_remodelfiber-internal.hpp
amgebauer: yes, the name is a bit ambiguous. It's w.r.t.the square of lambda_f. I tried to make it clear in the docstring above. I used this term quite a lot in the rest of the file. I didn't added comments everywhere. The exact meaning should get clear from the context.

lauraengelhardt (approved):

davidrudlstorfer (approved):

Issue #146, opened 2024-12-11
Migrating poroelasticity to new time integration


SchaefP1:
Copied from https://gitlab.lrz.de/baci/baci/-/issues/988

## Description
Poroelasticity has to be migrated to the new time integration.

Some more standard cases of problem type Poroelasticity can already be solved with the new time integration. However, the new time integration cannot be used for many other sub-types of Poroelasticity problems. (Note that this issue focuses purely on problems of problem type Poroelasticity and not things like Multiphase_Poroelasticity, Poroelastic_scalar_transport)

Examples for settings that do not work with the new time integration based on running the existing test files for Poroelasticity with the new time integration are: DYNAMICTYP Statics, PHYSICAL_TYPE Poro_P1, COUPALGO poro_monolithicfluidsplit, poro_monolithicstructuresplit, poro_monolithicmeshtying or poro_monolithicnopenetrationsplit or using both poro and normal solid elements together (PSI). Those all seem to lead to different kind of errors when using the test files with the new time integration.

It seems only the following kind of cases are already working with the new time integraton: COUPALGO poro_monolithic with PHYSICAL_TYPE Poro and DYNAMICTYP OneStepTheta or GenAlpha. And even 2 of the test files with those setting did not run with the new time integration.

Therefore, it needs to be checked what changes are still needed to fully migrated Poroelasticity to the new time integration and those changes need to be implemented.

## Possible solution and Definition of Done
The following sub-types of Poroelasticity problems need to be migrated and other steps need to be taken:
- [x] Migrate Poroelasticity with DYNAMICTYP Statics
- [ ] Migrate Poroelasticity with Poro_p1 elements
- [ ] Migrate Poroelasticity with COUPALGO that isn't poro_monolithic
- [ ] Migrate PSI
- [ ] Migrate poro with Nitsche contact
- [ ] Migrate poro with Lagrange contact
- [ ] Check what causes remaing issues in the 2 test files that do not fit into the other steps listed above
- [ ] Test poroelasticity with the new time intergation for actual uses cases

Issue #118, opened 2024-12-05
Reduce number of elements in poromulti* and porofluidmulti* test cases


lkoeglmeier:
## Description

A lot of test cases testing the pressure-based poro formulation (poromulti* and porofluidmulti*) have a large number of elements (>1000). To reduce the test time (sometimes > 70 s), it is proposed to reduce the number of elements in the test cases without reducing their utility, i.e. the number of tested lines of code.

## Possible Solution and Definition of Done

- [ ] Modify existing poromulti* and porofluidmulti* test cases by reducing the number of elements used.

PR #148, opened 2024-12-11
Enable poro solid interaction in new time integration

SchaefP1:
## Description and Context
Poro structure interaction (using poro and normal structure elements in the same simulation) is currently only possible with the old time integration. With the new time integration it leads to a segmentation fault as the discretization is set up new but this info is not passed on to the structure adapter. This pull request fixes this by using structure_setup() after setting up the discretization new. Additionally, some lines where removed from the restart function of poro that else caused issues and seem unneeded (duplicated and do not change the tests).

The tests are identical to the ones from the old time integration only using the new time integration (1 result is updated).

Part of #146


src/particle_rigidbody/4C_particle_rigidbody_result_test.cpp
```diff
@@ -12,6 +12,8 @@
 #include "4C_particle_rigidbody_interface.hpp"
 #include "4C_utils_exceptions.hpp"
 
+#include <algorithm>
```
SchaefP1, line 15: This change is unrelated to the rest of the pull request but without this include main did not compile for me.

sebproell: Out of curiosity: which compiler are you using?

SchaefP1: gcc-14.2.0

sebproell (approved):
The C++ side is fine, although I have to admit I have a hard time understanding how the change enables a new feature.

tests/input_files/poro_3D_psi_hex8_new_struc.dat
```diff
@@ -0,0 +1,961 @@
...
+NODE 209 COORD 5.0000000000000000e+00 2.9999999999999996e+00 -2.9999999999999996e+00
+NODE 210 COORD 5.0000000000000000e+00 3.0000000000000000e+00 -5.0000000000000000e+00
+NODE 211 COORD 5.0000000000000000e+00 5.0000000000000000e+00 3.0000000000000000e+00
+NODE 212 COORD 5.0000000000000000e+00 5.0000000000000000e+00 5.0000000000000000e+00
+NODE 213 COORD 5.0000000000000000e+00 5.0000000000000000e+00 1.0000000000000000e+00
+NODE 214 COORD 5.0000000000000000e+00 5.0000000000000000e+00 -1.0000000000000000e+00
+NODE 215 COORD 5.0000000000000000e+00 5.0000000000000000e+00 -3.0000000000000000e+00
+NODE 216 COORD 5.0000000000000000e+00 5.0000000000000000e+00 -5.0000000000000000e+00
+------------------------------------------------STRUCTURE ELEMENTS
+1 SOLIDH8PORO HEX8 1 2 3 4 5 6 7 8 MAT 1 KINEM nonlinear EAS none
```
amgebauer, line 837: What kind of poro formulation is used in this testcase? Can we use the new solid poro elements?

tests/input_files/poro_3D_psi_hex8_new_struc.dat
```diff
@@ -0,0 +1,961 @@
...
+TOLINC_PRES                     1e-06
+TOLINC_PORO                     1e-06
+TOLRES_DISP                     1e-04
+TOLRES_VEL                      1e-04
+TOLRES_PRES                     1e-04
+TOLRES_PORO                     1e-04
+VECTORNORM_RESF                 L1
+VECTORNORM_INC                  Rms
+TIMESTEP                        0.01
+RESULTSEVRY                           1
```
amgebauer, line 73: This file itself is not super small and we are doing 100 computation steps, and writing output on every step.

Is this really necessary or might it be sufficient to

 * use a smaller test
 * larger timestep size / fewer timesteps
 * write output only on few points in time?

I haven't checked how long the test is running in the pipeline, but to me it looks like a little too complex and runtime could be reduced a lot with just a few modifications.

The same applies to the other 3D input file. Probably somehow related to #118.

tests/input_files/new_solidporo_pressure_velocity_based_ele_3D_psi_hex8_new_struc.dat
```diff
@@ -0,0 +1,963 @@
...
+// Surface x+
+E 6 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
+// Surface y+
+//E 5 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
+// Surface y-
+//E 3 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
+// Surface z+
+//E 1 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
+// Surface z-
+//E 2 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
```
amgebauer, line 133:
```suggestion
```

tests/input_files/poro_3D_psi_hex8_new_struc.dat
```diff
@@ -0,0 +1,961 @@
...
+// Surface x+
+E 6 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
+// Surface y+
+//E 5 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
+// Surface y-
+//E 3 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
+// Surface z+
+//E 1 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
+// Surface z-
+//E 2 - NUMDOF 3 ONOFF 1 1 1 VAL 0.0 0.0 0.0 FUNCT 0 0 0
```
amgebauer, line 131:
```suggestion
```

amgebauer (changes requested):

tests/input_files/poro_3D_psi_hex8_new_struc.dat
mayrmt: If not testing for output, there's for sure no need to write results in every time step.

tests/input_files/poro_3D_psi_hex8_new_struc.dat
SchaefP1: Pressure velocity based. The new solid poro elements also works and is tested in the other 3D test. I chose to include the old elements as well since this test exists in the old time integration.

SchaefP1:
> The C++ side is fine, although I have to admit I have a hard time understanding how the change enables a new feature.

@sebproell Thanks for the review! Basically, poroelastic PSI already works in the new time integration once the set-up is complete. The problems are only while setting it up and restarting it as some info is not correctly passed on to the new time integration and it calls evaluate during set-up while some poro specific things are not set yet. I added some comments to make it easier to understand why this is needed and also changed a bit how I handle this in the read_restart function.

tests/input_files/poro_3D_psi_hex8_new_struc.dat
lkoeglmeier: I think the new solid poro elements also work for the old time integration, or am I understanding this wrong?

SchaefP1: Yes, both versions of the solid poro element work in the old time integration. The new test cases are just copies of the test cases in the old time integration with the main change being the use of the new time integration. The changes are independent of the used version of the solid element and I only included tests for both since both are tested in the old time integration. From my side we can also remove this test case if the migration to the new solid poro elements makes adding new tests with the old solid poro elements undesirable.

lkoeglmeier: Ah sorry, now I get it. My fault, thanks for the explanation. I think we want to get rid of the old solid elements at some point, so from my point of view, it would be sufficient to have only the test cases with the new solid elements. But I am also fine if you or @amgebauer have a different opinion on this.

SchaefP1: I removed the test with the old solid elements.

amgebauer (approved):

lkoeglmeier (approved):

Issue #137, opened 2024-12-09
[Linalg] Refactor matrix assembly routines


eulovi:
> Originating from https://gitlab.lrz.de/baci/baci/-/issues/1305

## Description

We implement a bunch of routines to assemble a local dense matrix into a global sparse one ... I'm not sure if all of them are necessary or if some can be refactored.

Several things to me stand out:

- In `4C_linalg_sparsematrix.{hpp,cpp}` some of the `Core::LinAlg::SparseMatrix::assemble` routines have the input parameter `int eid`, which is never used internally. Some calls to that method set them to `0` or `-1` e.g. `mglobal.assemble(-1, Mnode, lmrow, lmrowowner, lmcol);`, thus I think this input argument can/should be deleted to avoid confusion -\> shouldn't the compiler argue about that?
- The matrix assembly routine in `4C_linalg_utils_sparse_algebra_assemble.{hpp,cpp}` seems to do the same as one implemented in sparse matrix -\> thus can be deleted?

Maybe someone can approve these observations.

## Possible Solution and Definition of Done

Matrix assembly:

- [ ] Delete unused input arguments of `Core::LinAlg::SparseMatrix::assemble` in `4C_linalg_sparsematrix.{hpp,cpp}`
- [ ] Remove matrix assembly routine from `4C_linalg_utils_sparse_algebra_assemble.{hpp,cpp}` and instead use one from `4C_linalg_sparsematrix.{hpp,cpp}`
- [ ] Remove/Refactor assembly happening in `SplitStrategy` classes in `ale`, `fluid` and `scatra`

## Related Issues and Merge Requests

* Related to #870

eulovi:
In #149 it was seen that deleting `int eid` in the `Core::LinAlg::SparseMatrix::assemble` routines was not straightforward, as it is necessary in `InterfaceSplitStrategy` from ale and fluid and its deletion might affect the performance of both libraries. @maxfirmbach pointed out, that for the moment we don't have a reliable way to access performance, especially for long term behavior and for that we would need to have a proper benchmarking / performance suite.

Therefore, I mark this issue as blocked until we can asses this.

maxfirmbach:
I would propose to actually close this issue for now. It seems like there is no straight benefit right now.

PR #149, opened 2024-12-11
Draft: Delete unused input arguments in SparseMatrix

eulovi:
> Previous MR can be found in: https://gitlab.lrz.de/baci/baci/-/merge_requests/2443

## Description and Context

As pointed out in #137, the `assemble()` methods in 4C_linalg_sparsematrix.{hpp,cpp} had the input parameter eid. Although this parameter wasn't used in `SparseMatrix`, it was found to be used to check if the given element has conditioned nodes in the strategies of `InterfaceSplitStrategy` in `4C_ale_utils.hpp` and `4C_fluid_utils.hpp`. 

As the assembly strategy in InterfaceSplitStrategy might have a performance gain, this is assessed in this draft. For now, I changed all the `InterfaceSplitStrategy` instances to `Core::LinAlg::DefaultBlockMatrixStrategy`. If this works out, I think the `InterfaceSplitStrategy` classes from `ale` and `fluid` libraries can be deleted.

Related to #137 

## Checklist

- [X] These changes maintain backwards compatibility.


eulovi:
With the current changes, all the tests pass and no test fails due to timeout. Therefore, I think it is fine to delete `InterfaceSplitStrategy` from `ale` and `fluid` and use `DefaultBlockMatrixStrategy` from `LinAlg`. Would this be a good solution for you @mayrmt, @maxfirmbach?

mayrmt:
@eulovi If you want to assess performance, it is not sufficient to just look a the test suite. You need to run large examples on a cluster to actually obtain reliable timings.

mayrmt:
This PR is open for almost one year. @eulovi, what is missing to bring this into a mergeable state?

maxfirmbach:
@mayrmt @eulovi Let's close this. We have no reliable way right now to access performance, especially for long term behavior. For that we would need to have a proper benchmarking / performance suite.

eulovi:
@mayrmt sorry for the late response, back then I didn't know how to asses performance with the changes I did. I am fine closing this PR and I will add the comment of @maxfirmbach in #137.

Issue #73, opened 2024-11-29
Merge LineDefinition and LineComponent implementation


lauraengelhardt:
## Description

For historical reasons, multiple ways exist to specify lines read from data files. These should be unified. Much work has already been done in this direction. This issue collects the last remaining steps.

The goal is to have 
 - one way to specify an expected input line
 - one reading mechanism for a line that matches the expectation
 - and one storage container `InputParameterContainer` containing the parsed values.

## Possible Solution and Definition of Done
- [x] Move all uses of `LineDefinition::extract_xxx` to `InputParameterContainer::Get<xxx>`. The `LineDefinition::read` already returns the container, one just needs to use it.
- [x] This requires a tag-like object, which can be used to mimic LineDefinition::have_named() on the container. Probably we should just store tags as bools internally.
- [x] Remove the `extract_xxx methods` from LineDefinition. Almost all places in the code that do something with input should only see the `InputParameterContainer`.
- [ ] Decide how to merge `LineDefinition` and `LineComponent` by looking at the places that define lines.

## Interested Parties

sebproell:
Due to the reasons outlined above and in #301, I want to remove the remaining `tag()`s in the input mechanism. Most tags will become groups without changes to the input files. There are two cases that require a minor input file change since they will become `bool` parameters.

- @amgebauer @mairehenke The input for Shell 7p elements uses a `tag("ANS")`. What do you think about `entry<bool>("USE_ANS")` instead?
- @davidrudlstorfer @isteinbrecher The input for beam elements uses a `tag("FAD")`. What do you think about `entry<bool>("USE_FAD")` instead?

davidrudlstorfer:
> @davidrudlstorfer @isteinbrecher The input for beam elements uses a tag("FAD"). What do you think about entry<bool>("USE_FAD") instead?

@sebproell  I am totally fine with that

amgebauer:
> @amgebauer @mairehenke The input for Shell 7p elements uses a tag("ANS"). What do you think about entry<bool>("USE_ANS") instead?

Is fine from my side. But @mairehenke's opinion counts more here.

In the long run, we might want to unify it with the solid elements interface (`TECH` that could then contain the element technologies). But we might want to do it later. Currently `TECH` only takes a string, but it might be more favorable to take a list of strings for combinations of different element technologies (`eas` and `ans`). But that's currently not possible in our legacy dat file format :roll_eyes:

mairehenke:
[...]

PR #151, opened 2024-12-11
Remove unnamed components from LineDefinition

sebproell:
Part of #73 

The LineDefinition supported special components where the name was not required in the input line. @lauraengelhardt and I think this is confusing and not necessary. Thus, this PR removes these special cases. No changes to input files are necessary which kind of proves our point ;)

FYI @mairehenke I changed internals of the shell elements. You can now use a string vector.

### DIFF
--- a/src/core/io/src/4C_io_elementreader.cpp
+++ b/src/core/io/src/4C_io_elementreader.cpp
@@ -12,6 +12,7 @@
 #include "4C_fem_general_element.hpp"
 #include "4C_fem_general_element_definition.hpp"
 #include "4C_io_input_file.hpp"
+#include "4C_io_value_parser.hpp"
 #include "4C_rebalance_print.hpp"
 
 FOUR_C_NAMESPACE_OPEN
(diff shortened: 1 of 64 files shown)


lauraengelhardt (approved):
Looks very good to me! 
I guess changes to the input files will only be necessary for the ConditionDefinition. Good to see that at least for the LineDefinition, this was not the case.

amgebauer (approved):
The changes look good, but I haven't yet understood the problematic case. Do you have an example where we used an unnamed component?

src/scatra_ele/4C_scatra_ele.cpp
```diff
@@ -93,147 +93,147 @@ void Discret::Elements::TransportType::setup_element_definition(
   std::map<std::string, Input::LineDefinition>& defs = definitions["TRANSP"];
 
   defs["HEX8"] = Input::LineDefinition::Builder()
-                     .add_int_vector("HEX8", 8)
+                     .add_named_int_vector("HEX8", 8)
```
sebproell, line 96: The problematic cases were these: previously, we added the "HEX8" tag on a higher level. Now, it is required here.
We actually need to look at the "HEX8" tag quite early; at least, this is how element input is currently implemented. Now I "peek" at the cell type tag...

src/core/io/src/4C_io_elementreader.cpp
```diff
@@ -179,15 +180,15 @@ void Core::IO::ElementReader::get_and_distribute_elements(const int nblock, cons
...
-      t >> elenumber >> eletype >> distype;
-      elenumber -= 1;
+      ValueParser parser{element_line, "While reading element line: "};
+      const int elenumber = parser.read<int>() - 1;
       gidlist.push_back(elenumber);
 
+      const auto eletype = parser.read<std::string>();
+
+      // Only peek at the distype since the elements later want to parse this value themselves.
+      const std::string distype = std::string(parser.peek());
```
sebproell, line 190: ...here.

src/scatra_ele/4C_scatra_ele.cpp
amgebauer: ah ok, I never looked into these internals of the element reader. I thought that HEX8 was already a named integer vector.

PR #154, opened 2024-12-12
PackBuffer: avoid an additional copy of data

sebproell:
We can resize the destination buffer and `memcpy` in there directly, thus, saving a copy operation.

This also fixes a warning seen in #130, although I am not exactly sure why.


mayrmt (approved):

sebproell:
Can I please get another review?

davidrudlstorfer (approved):

PR #156, opened 2024-12-13
Increase number of chunks for coverage workflow

### LINKED ISSUE #155: [BUG] Coverage report is failing

I tested it in my pro-account and runners seem to have a little more storage.

@c-p-schmidt is currently testing in his repo how we need to adapt the chunking that it works also on non-pro runners.

## Possible Solution

 * [x] Adapt chunking
 * [ ] Other way to reduce storage

- Follows #128

c-p-schmidt:
## Description and Context
For the coverage workflow, we ran into problems concerning disk storage (see https://github.com/4C-multiphysics/4C/actions/runs/12308399500). Thus, the number of chunks is increased from 40 to 60, resulting in less output data per runner, which resolved the issue on my fork.

Closes #155 
Follows https://github.com/4C-multiphysics/4C/pull/128

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           build-directory: ${{ github.workspace }}/build
           source-directory: ${{ github.workspace }}
-          number-of-chunks: 40 # we need fine grained test chunks to avoid running out of storage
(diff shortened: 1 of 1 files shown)


amgebauer (approved):
Thanks for the fix!

ppraegla (approved):

PR #157, opened 2024-12-13
Remove my_output_flag function from InputFile

sebproell:
This implementation was in the wrong place. I removed the flag. If output should be conditional, one may introduce a flag for that in the respective function in the future. I kept the same console output except for microscale simulations, which now print a few more lines.


sebproell:
Closing as this is part of #161 now.

PR #159, opened 2024-12-13
Add FILE_PATH and FILE_DIR variable substitution in input files

sebproell:
Introduce variable substitutions for `${FILE_PATH}` and `${FILE_DIR}` inside input files. This allows me to drop another special case `LineDefinition::add_named_path`, which achieved something similar but the implementation feels misplaced. The old implementation would not have worked with included files. @c-p-schmidt This should have been used for the csv function, but seems to be broken. Do we even need this feature? 

@gilrrei @ppraegla @amgebauer Does this kind of variable substitution make sense? Or is this better done by external tools (e.g. jinja). I don't want to support a lot of variables, but it seems useful to be able to refer to the file location.

Depends on #157 

Related to #73

### DIFF
--- a/apps/global_full/4C_global_full_control.cpp
+++ b/apps/global_full/4C_global_full_control.cpp
@@ -12,6 +12,7 @@
 #include "4C_global_full_inp_control.hpp"
 #include "4C_io_pstream.hpp"
 
+#include <chrono>
 
 namespace
 {
--- a/apps/pre_exodus/4C_pre_exodus_validate.cpp
+++ b/apps/pre_exodus/4C_pre_exodus_validate.cpp
@@ -29,7 +29,7 @@ void EXODUS::validate_input_file(const MPI_Comm comm, const std::string datfile)
   Global::Problem* problem = Global::Problem::instance();
 
   // create a InputFile
-  Core::IO::InputFile reader(datfile, comm, 0);
+  Core::IO::InputFile reader(datfile, comm);
 
   // read and validate dynamic and solver sections
   std::cout << "...Read parameters" << std::endl;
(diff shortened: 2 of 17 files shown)


amgebauer:
In general, I think it's a nice feature to be able to parameterize the input files with some context variables, e.g. for me, it would be valuable to encode the restart step in the input file or the output directory could also be interesting. For consistency, we can also provide variables for `FILE_PATH` and `FILE_DIR`, but I don't see the exact use-case for these. Paths should anyway be relative to the input file itself, so when would I need these variables?

gilrrei:
@sebproell I agree with @amgebauer.

I didn't understand if you're asking about introducing variables in general but some comments:
- If we implemented substitutions ourselves, we basically introduce a 'programming/script' language into the dat/yaml format. This is super difficult to maintain (implement an interpreter for: mathematical operations? strings?)
- Even if we use some existing tool like jinja, we might not be in a valid yaml/dat format anymore (`{{ blabla }}`), which means validation/autocomplete might be lost.
- One important point to note is the implicit knowledge this entails, i.e., how does this substitution work? This also includes how to document everything
- In QUEENS we had a similar problem since we wanted to initialize classes using yaml. Here, the problem emerges when creating an object, which entails creating other objects first. We 'solved' this problem by introducing a convention if the key  of the argument is of the form `<key_of_the_nested_object>_name` , the input reader looks for a section named `<key_of_the_nested_object>`, initializes it, and this is passed as argument. It's still strange but at least we maintained yaml compatibility

sebproell:
> I didn't understand if you're asking about introducing variables in general but some comments:

@gilrrei I was looking for a comment along the lines you gave above :) 

This confirms my suspicion and experience that we should not overcomplicate the file, or at least not light-heartedly.

For now, I do no longer need variables, given that #161 can work around the specific problem I was trying to solve with them here.

Side note on math and variables: For turing-completeness, I still like lua as a configuration language which has a clean syntax very close to yaml/json.

PR #160, opened 2024-12-13
Add min ele option to initial rebalancing

maxfirmbach:
## Description and Context

In interface driven problems, usually one domain dominates the simulation (e.g. in FSI the fluid domain is usually way bigger than the solid one) and thus determines the amount of processors to use in a parallel setting. At some point it can happen, that the smaller domain is oversubscribed by processors resulting in only a few elements owned per proc and thus limiting scalability or even crashing in the process, overall resulting in a very fragile solver.

This PR addresses the issue by implementing a user option to fix the minimum amount of elements a processor has to own to be used for the simulation run. Therefore, a short calculation routine is added in the initial rebalancing / domain partition algorithm similar to the interface partitioning methods already available for mortar and contact methods.

A pressure wave test case is adapted to trigger the procedure. While the overall amount of domains is three, the solid domain is rebalanced only on two, fulfilling the minimum amount of elements given.

```shell
 Procs used for redistribution: 2
 Redistributing using hypergraph .........
 ZOLTAN Load balancing method = 10 (HYPERGRAPH)

 +--------------------------------------------------------------------+
 | fill_complete() on discretization structure                         |
 +--------------------------------------------------------------------+

    discretization: structure
    +-----+---------------+--------------+-----------------+----------------+
    | PID |  n_rownodes   | n_ghostnodes |  n_rowelements  |   n_ghostele   |
    +-----+---------------+--------------+-----------------+----------------+
    |   0 |          1263 |          141 |             766 |             80 | 
    +-----+---------------+--------------+-----------------+----------------+
    |   1 |          1257 |          141 |             834 |             10 | 
    +-----+---------------+--------------+-----------------+----------------+
    |   2 |             0 |            0 |               0 |              0 | 
    +-----+---------------+--------------+-----------------+----------------+
 
 
 Procs used for redistribution: 3
 Redistributing using hypergraph .........
 ZOLTAN Load balancing method = 10 (HYPERGRAPH)
 
 +--------------------------------------------------------------------+
 | fill_complete() on discretization fluid                             |
 +--------------------------------------------------------------------+
 
    discretization: fluid
    +-----+---------------+--------------+-----------------+----------------+
    | PID |  n_rownodes   | n_ghostnodes |  n_rowelements  |   n_ghostele   |
    +-----+---------------+--------------+-----------------+----------------+
    |   0 |          1014 |          335 |             974 |             40 | 
    +-----+---------------+--------------+-----------------+----------------+
    |   1 |          1016 |          180 |             758 |            124 | 
    +-----+---------------+--------------+-----------------+----------------+
    |   2 |          1015 |          155 |             748 |            124 | 
    +-----+---------------+--------------+-----------------+----------------+
```

Interesting for @mayrmt @isteinbrecher


src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -138,9 +138,22 @@ void Core::IO::MeshReader::rebalance()
...
 
+    const int minele_per_proc = parameters_.mesh_paritioning_parameters.get<int>("MIN_ELEPROC");
+    const int max_global_procs = Core::Communication::num_mpi_ranks(comm_);
+    int min_global_procs = max_global_procs;
+
+    if (minele_per_proc > 0)
+      min_global_procs =
+          element_readers_[i].get_row_elements()->NumGlobalElements() / minele_per_proc;
+    const int procs = min_global_procs > max_global_procs ? max_global_procs : min_global_procs;
+    rebalanceParams.set<std::string>("num parts", std::to_string(procs));
```
mayrmt, line 149:
```suggestion
    const int num_procs = min_global_procs > max_global_procs ? max_global_procs : min_global_procs;
    rebalanceParams.set<std::string>("num parts", std::to_string(num_procs));
```

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -138,9 +138,22 @@ void Core::IO::MeshReader::rebalance()
...
+
     const auto rebalanceMethod = Teuchos::getIntegralValue<Core::Rebalance::RebalanceType>(
         parameters_.mesh_paritioning_parameters, "METHOD");
 
+    if (!Core::Communication::my_mpi_rank(comm_))
+      std::cout << "\nProcs used for redistribution: " << procs << "\n";
```
mayrmt, line 155: The screen output sounds a bit like it will be listing the IDs of the process. Rephrase?
```suggestion
    if (!Core::Communication::my_mpi_rank(comm_))
      std::cout << "\nNumber of procs used for redistribution: " << procs << "\n";
```
or
```suggestion
    if (!Core::Communication::my_mpi_rank(comm_))
      std::cout << "\nDistributing discretization to " << num_procs << " MPI ranks...\n";
```

src/inpar/4C_inpar_rebalance.cpp
```diff
@@ -33,6 +33,9 @@ void Inpar::Rebalance::set_valid_parameters(Teuchos::ParameterList& list)
       "Tolerance for relative imbalance of subdomain sizes for graph partitioning of unstructured "
       "meshes read from input files.",
       &meshpartitioning);
+
+  Core::Utils::int_parameter("MIN_ELEPROC", 0,
+      "Minimum no. of elements per processor for parallel redistribution.", &meshpartitioning);
```
mayrmt, line 38: Can you include the fact, that 0 disables the minimum number of elements per proc, into the documentation string?

```suggestion
  Core::Utils::int_parameter("MIN_ELEPROC", 0,
      "This parameter defines the minimum number of elements to be assigned to any MPI rank during redistribution. Use 0 to not interfere with the minimal size of a subdomain.", &meshpartitioning);
```

mayrmt (changes requested):
@maxfirmbach Looks good! I have left some comments, mostly related to screen output.

In the past, there have been issues if a proc did not own anything (though I don't remember if this was problematic for nodes or elements). For sure, `ML` could not deal with empty processors, but that does not matter anymore today.

Can you comment on such limits? Is there a minimum number of nodes or elements to be owned by each processor? (Apparently, zero elements per MPI rank work fine.)

maxfirmbach:
> In the past, there have been issues if a proc did not own anything (though I don't remember if this was problematic for nodes or elements). For sure, `ML` could not deal with empty processors, but that does not matter anymore today.

`MueLu` is able to handle empty processors and lables them `active` and `inactive`, which actually is also important for rebalancing coarse level matrices.

> Can you comment on such limits? Is there a minimum number of nodes or elements to be owned by each processor? (Apparently, zero elements per MPI rank work fine.)

Well, I think there is no clear answer to that. Zero elements / nodes / dofs per proc works, because this proc is simply ignored. A very small amount of elements / nodes / dofs per proc is difficult, because it often leads to breakdowns (especially in the aggregation process of algebraic multigrid). I like to work with around 50.000 dofs per proc, which relates to x-elements and y-nodes to procs. In the end it's highly problem dependent ...

maxfirmbach:
@mayrmt I would like to work with "MIN_DOFPROC" instead of "MIN_ELEPROC", but at the point of rebalancing, only element and node maps are available.

src/inpar/4C_inpar_rebalance.cpp
mayrmt:
```suggestion
  Core::Utils::int_parameter("MIN_ELE_PER_PROC", 0,
      "Minimum no. of elements per processor for parallel redistribution.", &meshpartitioning);
```

mayrmt:
> @mayrmt I would like to work with "MIN_DOFPROC" instead of "MIN_ELEPROC", but at the point of rebalancing, only element and node maps are available.

If you have the element map, do you already know the type of elements? If yes, can you query them for their number of nodes and type of physics, so you can estimate their number of DOFs? (I guess that this is not possible, but I wanted to raise the question.)

maxfirmbach:
> @maxfirmbach If you have the element map, do you already know the type of elements? If yes, can you query them for their number of nodes and type of physics, so you can estimate their number of DOFs? (I guess that this is not possible, but I wanted to raise the question.)

@mayrmt No, I think this is not an option. At least it seems to be overly complicated …

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -138,9 +138,22 @@ void Core::IO::MeshReader::rebalance()
...
     rebalanceParams.set<std::string>("imbalance tol", std::to_string(imbalance_tol));
 
+    const int minele_per_proc = parameters_.mesh_paritioning_parameters.get<int>("MIN_ELEPROC");
+    const int max_global_procs = Core::Communication::num_mpi_ranks(comm_);
+    int min_global_procs = max_global_procs;
+
+    if (minele_per_proc > 0)
+      min_global_procs =
+          element_readers_[i].get_row_elements()->NumGlobalElements() / minele_per_proc;
+    const int procs = min_global_procs > max_global_procs ? max_global_procs : min_global_procs;
```
isteinbrecher, line 148:
```suggestion
    const int procs = min(max_global_procs, min_global_procs);
```

isteinbrecher (commented):
@maxfirmbach looks good to me, I left one minor comment.

src/core/io/src/4C_io_meshreader.cpp
maxfirmbach: Good simplification, changed it.

src/inpar/4C_inpar_rebalance.cpp
maxfirmbach: Renamed.

src/core/io/src/4C_io_meshreader.cpp
maxfirmbach: Done.

mayrmt (approved):
@maxfirmbach thanks for the contribution!

Please rebase and squash locally. PR approvals should be retained.

isteinbrecher (approved):

maxfirmbach:
I think I'm not able to merge by myself ...

isteinbrecher:
You are not yet added to the 4C group. Seems like I don't have the rights to add you. Could somebody from the @4C-multiphysics/maintainer please add Max Firmbach to the list (and maybe extend my rights)?

mayrmt:
@maxfirmbach You should have a pending invitation to the 4C developers group. Check your GitHub account! I have sehr it a couple of weeks ago. 

This should give you the necessary rights.

PR #162, opened 2024-12-14
Add benchmarking utils in transv.isotrop. material

dragos-ana:
## Description and Context

This PR adds utilities for benchmarking the time integration algorithm of `InelasticDefgradTransvIsotropElastViscoplast` (Gauss point integration of history variables). These utilites are of great help for me in the development process of the material model. Specifically, the following work was performed:
    - A dedicated `BenchmarkUtils`  object is now utilized to track the number of substeps, the number of iterations, the computation times and the different error types that arise during computation. The functionality is currently only provided for single-processor computations (since the main usage is currently only for one-element simulations).
    - A csv writer is integrated within the `BenchmarkUtils` object, which writes runtime output for the time integration benchmarking (see quantities above).
    - The user can select whether to write dedicated runtime output for the Local Newton-Raphson Loop via the dedicated boolean variable `BENCHMARK_TIMINT`.


sebproell:
@dragos-ana I appreciate that you want to benchmark your code :+1: Depending on what you wish to achieve, I recommend looking into a benchmarking framework (I have used Google Benchmark https://github.com/google/benchmark, but others are probably available as well). The current implementation is, on the one hand, quite intrusive and, on the other hand, does not exploit all the tooling that is available for benchmarking. Feel free to talk to me directly to discuss your use case.

PR #163, opened 2024-12-16
Add coverage badge to README

amgebauer:
Adds the coverage badges to our README.

I also change the coverage color so that it is consistent with our previous colors.

 * 0-75: red
 * 75-90: orange
 * 90-100: green


davidrudlstorfer (approved):
Nice! The coverage badge has the identical design as all other 3rd party badges.

PR #161, opened 2024-12-14
InputFile can return the file a section was read from

sebproell:
Alternative to #159. In contrast to that PR, this PR keeps the old behavior and fixes it.

Issue #113, opened 2024-12-04
Export valid input options to yaml


gilrrei:
## Description
Currently, 4C has the capability to export some parameters to yaml already. This file can in the future also be used to generate the documentation etc. This feature is currently  used to create the JSON schema for .yaml input files.

However, for some sections this is not the case.

Related to [1337](https://gitlab.lrz.de/baci/baci/-/issues/1337)

## Possible Solution and Definition of Done
What should be exported?
- Description for each entry
- Expected parameters, their type and description

What should in the long term be exported?
- Meaningful defaults
- Information if the parameter is required
- Dependency on other parameters (e.g. using OST requires you to specify a value for theta)

Sections:
- [x] Everything that is converted to a Teuchos ParameterList object
- [x] Includes section
- [x] Title section
- [x] Boundary conditions
- [x] Functions
- [ ] Interface and interaction conditions
- [x] Material section
- [x] Results descriptions
- [x] Cloning material sections  
- [x] Nodeset sections
- [x] Element sections

Definition on done: Each section exports the 'What should be exported?' entries

sebproell:
I think we should tackle #73 with this issue in mind (so we only implement _one_ yaml emitter for our line definitions).

gilrrei:
Currently, bool parameters of the `Parameterlist` are exported as
```yaml
  NURBS:
    DO_LS_DBC_PROJECTION:
      type: string
      default: No
      description: "Determines if a projection is needed for least square Dirichlet boundary conditions."
      valid options:
        - Yes
        - No
        - yes
        - no
        - YES
        - NO
```
however, for meaningful validation they should be exported as:
```yaml
  NURBS:
    DO_LS_DBC_PROJECTION:
      type: boolean
      default: false
      description: "Determines if a projection is needed for least square Dirichlet boundary conditions."
```
otherwise the validation will only be happy if you use `DO_LS_DBC_PROJECTION: "yes"` (or similar). 

For the older yaml standard [1.1](https://yaml.org/type/bool.html) the booleans are:
* for true: `y, Y, yes, Yes, YES, true, True, TRUE,on, On, ON,`
* for false `n, N, no, No, NO, false, False, FALSE, off, Off, OFF`

[...]

PR #165, opened 2024-12-16
Reimplement LineDefinition as a more powerful InputSpec

sebproell:
### Description

To progress with #73 and merge our two input definitions, we need an implementation that can handle all the cases the two competing implementations have. This PR proposes this implementation. The main design goal is to have an easily readable and flexible way to construct input lines (as known from .dat format), while keeping in mind that we also want to support other formats soon (mainly yaml). In its most concise form, a line may now be written like this (see also the unit tests):

```cpp
  auto input = group({
        entry<int>("a", {.default_value = 42}),
        one_of({
            entry<double>("b"),
            group("nested",
                    {
                        entry<std::string>("c"),
                        entry<double>("d"),
                    }),
        }),
    });
```

This is a very concise style without any unnecessary clutter. Of course, this is still C++ and you can also name all the types very explicitly. The single parts that make up the line may also be stored separately and you are not forced to spell out a complete input line at once. For instance, in the `LineDefinition` we add the entries incrementally. Overall, this value-based design is a lot more flexible and, IMO, readable than the pointer-laden design of `LineComponent` which will soon also be replaced by the new implementation. We can generate emitters for yaml that work on the new implementation, so that we should be able to tackle #113 soon :tada: 

To demonstrate that this works, and to not have a third implementation, I replaced the internals of `LineDefinition` with the new class `InputSpec`. In the future, the `InputSpec` will be used directly instead of `LineDefinition`, but let's take one step at a time.

Related to #73 , ~~Depends on #161~~, Depends on #170

### DIFF
--- a/src/core/io/src/4C_io_input_parameter_container.hpp
+++ b/src/core/io/src/4C_io_input_parameter_container.hpp
@@ -46,16 +46,13 @@ namespace Core::IO
     }
   }  // namespace Internal
 
-  /*!
-  \brief A data storage container
-
-  You can store various types of data in this container and access the data by keys.
-
-  The intention of this class is to store rather 'small' units of data. Though possible,
-  it is not meant to be used at the system level to store huge data sets such as sparse matrices or
-  vectors of system length. It does therefore not support any Core::LinAlg::Vector<double>
-  or Epetra_CrsMatrix objects and is not supposed to in the future either.
-  */
+  /**
+   * A container to store dynamic input parameters. The container can store arbitrary types of
+   * parameters. Parameters can be grouped in sub-containers.
+   *
+   * This class is a core part of the input mechanism, as it contains the parsed data from the input
+   * file and grants access to it.
+   */
   class InputParameterContainer
   {
    public:
@@ -101,6 +98,21 @@ namespace Core::IO
         anydata_[name] = data;
     }
 
+    /**
+     * Access group @p name. If the group does not exist, it will be created.
+     */
+    InputParameterContainer& group(const std::string& name);
+
+    /**
+     * Access group @p name. This function throws an error if the group does not exist.
+     */
+    [[nodiscard]] const InputParameterContainer& group(const std::string& name) const;
+
+
+    /**
+     * Combine the data from another container with this one. Conflicting data will throw an error.
+     */
+    void merge(const InputParameterContainer& other);
 
     /*!
      * Get a const reference to the data stored at the key @p name from the container. An error
@@ -210,6 +222,9 @@ namespace Core::IO
 
     //! a map to store anything
     std::map<std::string, std::any> anydata_;
+
+    //! Groups present in this container. Groups are InputParameterContainers themselves.
+    std::map<std::string, InputParameterContainer> groups_;
   };  // class InputParameterContainer
 }  // namespace Core::IO
 
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -1,0 +1,896 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_IO_INPUT_SPEC_BUILDERS_HPP
+#define FOUR_C_IO_INPUT_SPEC_BUILDERS_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_io_input_parameter_container.hpp"
+#include "4C_io_input_spec.hpp"
+#include "4C_io_value_parser.hpp"
+
+#include <functional>
+#include <optional>
+#include <ostream>
+#include <variant>
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Core::IO
+{
+  namespace Internal
+  {
+    /**
+     * Helper to print values in the format of a dat file.
+     */
+    struct DatPrinter
+    {
+      template <typename T>
+      void operator()(std::ostream& out, const T& val) const
+      {
+        out << " " << val;
+      }
+
+      void operator()(std::ostream& out, bool val) const { out << " " << (val ? "true" : "false"); }
+
+      template <typename T>
+      void operator()(std::ostream& out, const std::vector<T>& val) const
+      {
+        for (const auto& v : val)
+        {
+          (*this)(out, v);
+        }
+      }
+
+      template <typename T, typename U>
+      void operator()(std::ostream& out, const std::pair<T, U>& val) const
+      {
+        (*this)(out, val.first);
+        (*this)(out, val.second);
+      }
+    };
+
+    template <typename T, typename AlwaysVoid = void>
+    constexpr bool has_print = false;
+
+    template <typename T>
+    constexpr bool has_print<T, std::void_t<decltype(std::declval<const std::decay_t<T>>().print(
+                                    std::declval<std::ostream&>(),
+                                    std::declval<const Core::IO::InputParameterContainer&>()))>> =
+        true;
+
+    template <typename T, typename AlwaysVoid = void>
+    constexpr bool has_set_default_value = false;
+
+    template <typename T>
+    constexpr bool has_set_default_value<T,
+        std::void_t<decltype(std::declval<const std::decay_t<T>>().set_default_value(
+            std::declval<Core::IO::InputParameterContainer&>()))>> = true;
+
+
+    template <typename T, typename AlwaysVoid = void>
+    struct PrettyTypeName
+    {
+      std::string operator()() { return Utils::try_demangle(typeid(T).name()); }
+    };
+
+    template <>
+    struct PrettyTypeName<std::string>
+    {
+      std::string operator()() { return "string"; }
+    };
+
+    template <>
+    struct PrettyTypeName<std::filesystem::path>
+    {
+      std::string operator()() { return "char"; }
+    };
+
+    template <typename T>
+    struct PrettyTypeName<std::vector<T>>
+    {
+      std::string operator()() { return "vector<" + PrettyTypeName<T>{}() + ">"; }
+    };
+
+    template <typename T, typename U>
+    struct PrettyTypeName<std::pair<T, U>>
+    {
+      std::string operator()()
+      {
+        return "pair<" + PrettyTypeName<T>{}() + ", " + PrettyTypeName<U>{}() + ">";
+      }
+    };
+
+    template <typename T>
+    std::string get_pretty_type_name()
+    {
+      return PrettyTypeName<T>{}();
+    }
+
+    class InputSpecTypeErasedInterface
+    {
+     public:
+      virtual ~InputSpecTypeErasedInterface() = default;
+      virtual void parse(ValueParser& parser, InputParameterContainer& container) const = 0;
+      virtual void set_default_value(InputParameterContainer& container) const = 0;
+      virtual void print(std::ostream& stream, const InputParameterContainer& container) const = 0;
+      [[nodiscard]] virtual std::unique_ptr<InputSpecTypeErasedInterface> clone() const = 0;
+    };
+
+    template <typename T>
+    struct InputSpecTypeErasedImplementation : public InputSpecTypeErasedInterface
+    {
+      template <typename T2, std::enable_if_t<!std::is_same_v<std::decay_t<T>, InputSpec>, int> = 0>
+      explicit InputSpecTypeErasedImplementation(T2&& wrapped) : wrapped(std::forward<T2>(wrapped))
+      {
+      }
+
+      void parse(ValueParser& parser, InputParameterContainer& container) const override
+      {
+        wrapped.parse(parser, container);
+      }
+
+      void set_default_value(InputParameterContainer& container) const override
+      {
+        if constexpr (Internal::has_set_default_value<T>)
+        {
+          wrapped.set_default_value(container);
+        }
+        else
+        {
+          FOUR_C_ASSERT(wrapped.data.default_value.has_value(),
+              "Implementation error: this function should only be called if the wrapped type has "
+              "an optional default value.");
+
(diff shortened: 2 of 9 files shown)


src/core/io/src/4C_io_input_line.cpp
```diff
@@ -0,0 +1,214 @@
...
+        FOUR_C_THROW("Required value '%s' not found in input line", name.c_str());
+      }
+      else if (options.store_default_values)
+      {
+        entry->set_default_value(container);
+      }
+    }
+  }
+
+  void parse_ordered(Core::IO::ValueParser& parser, const std::vector<Core::IO::Component>& line,
```
amgebauer, line 70: Where do we actually need to parse ordered?

src/core/io/src/4C_io_input_line.hpp
```diff
@@ -0,0 +1,633 @@
...
+#include <variant>
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Core::IO
+{
+  namespace Internal
+  {
+    /**
+     * Helper to print values in the peculiar format of a dat file.
```
amgebauer, line 28: yes, it is perculiar, but I don't think we need to document it here in code.
```
     * Helper to print values in the format of a dat file.
```

src/core/io/src/4C_io_input_line.hpp
```diff
@@ -0,0 +1,633 @@
...
+    template <typename StoredTypeIn, typename ScalarTypeIn>
+    struct VectorComponentData
+    {
+      using StoredType = StoredTypeIn;
+      using ScalarType = ScalarTypeIn;
+
+      /**
+       * A function that determines the size of the vector. This function is called with the
+       * already parsed content of the input line, which may be used to query the size as the
+       * value of another parameter. To create the
```
amgebauer, line 299: incomplete docs

src/core/io/src/4C_io_input_line.hpp
```diff
@@ -0,0 +1,633 @@
...
+
+      /**
+       * An optional description of the value.
+       */
+      std::string description{};
+
+      /**
+       * Whether the value is required or optional.
+       */
+      std::optional<bool> required{};
```
amgebauer, line 282: why does it need to be optional?

sebproell: Good catch! As I just discussed with @gilrrei I overengineered this part :) This will be dropped, as it's enough to pass a `default_value` which must mean the entry is optional. If an entry is required, there cannot be a default.

sebproell (commented):
Thanks for the initial review, @amgebauer! I will put this back in draft state since I have a few more improvements I can make right here.

src/core/io/tests/4C_io_input_line_test.cpp
```diff
@@ -0,0 +1,368 @@
...
+      fully_parse(parser, line, container);
+      EXPECT_EQ(container.get<int>("a"), 1);
+      EXPECT_EQ(container.group("group").get<std::string>("c"), "string");
+      EXPECT_EQ(container.group("group").get<double>("d"), 2);
+    }
+
+    {
+      InputParameterContainer container;
+      std::string stream("a 1 group c string d 2.0 b 3.0");
+      // Both options are given, so we expect an exception.
```
sebproell, line 347: Note to self: missing test here.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,884 @@
...
+     *     yes, set `default_value` to this value (`required` implicitly is `false` then). This
+     *     guarantees that you can always read the a value from the container with
+     *     InputParameterContainer::get(). A good example is a parameter that activates or
+     *     deactivates a feature, e.g., defaulting the EAS element technology to off might be
+     *     reasonable.
+     *
+     *   - If the parameter is not required, but there is no reasonable default value, set
+     *     `required = false`. This makes the parameter optional and the container might not contain
+     *     the parameter if it is not present in the input file. Use
+     *     InputParameterContainer::get_if() or InputParameterContainer::get_or() to safely retrieve
```
amgebauer, line 549: Do we really want to provide the `get_or()`? To me, it looks like it's a silent default.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,884 @@
...
+     */
+    struct DatPrinter
+    {
+      template <typename T>
+      void operator()(std::ostream& out, const T& val) const
+      {
+        out << " " << val;
+      }
+
+      void operator()(std::ostream& out, const std::string& val) const { out << " " << val; }
```
amgebauer, line 39: is this needed? I don't see the difference to the template function above.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,884 @@
...
+
+      void operator()(std::ostream& out, bool val) const { out << " " << std::boolalpha << val; }
```
amgebauer, line 41: This function has a side effect. Why not

```suggestion
      void operator()(std::ostream& out, bool val) const { out << " " << val ? "true" : "false"; }
```
?

amgebauer (approved):
Thanks for the nice work!

src/core/io/src/4C_io_input_spec_builders.hpp
sebproell: Yes, if you do this in your code, it is a silent default. It depends on whether you think about the `InputParameterContainer` in isolation, where `get_or` makes sense, or if you think about it in combination with `InputSpec`, which would provide a different way to specify defaults. `get_or` predates `InputSpec`. I am not 100% sure  what we will need, but I suspect you may be right and `get_or` should not be necessary in combination with `InputSpec`.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,882 @@
...
+      return *this;
+    }
+
+    InputSpec(InputSpec&&) noexcept = default;
+    InputSpec& operator=(InputSpec&&) noexcept = default;
+
+    /**
+     * Parse this spec from the input. This is often not the correct function to call, if you
+     * want to ensure that the input is fully parsed. Use the fully_parse() function instead.
+     */
```
lauraengelhardt, line 255:
```suggestion
    /**
     * Parse this spec from the input. This is often not the correct function to call, 
     * as it does not validate if the input has been fully parsed. If you want to ensure that the input is fully parsed,
     * use the fully_parse() function instead.
     */
```

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,882 @@
...
+    [[nodiscard]] const std::string& name() const { return data_.name; }
+
+    [[nodiscard]] const std::string& description() const { return data_.description; }
+
+    [[nodiscard]] bool required() const { return data_.required; }
+
+    [[nodiscard]] bool has_default_value() const { return data_.has_default_value; }
+
+   private:
+    CommonData data_;
```
lauraengelhardt, line 286: Docu?

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,882 @@
...
+       * size based on the value of another parameter.
+       *
+       * @note We use `int` instead of `size_t` since this is what most users write. Template
+       *      argument deduction will work a lot better with `int` than with `size_t`.
+       */
+      std::variant<int, SizeCallback> size;
+    };
+
+    template <typename StoredTypeIn>
+    struct SelectionData
```
lauraengelhardt, line 358: `ScalarData` and `VectorData` is very clear by it's name. `SelectionData` might need some additional docu.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,882 @@
...
+      std::optional<bool> required{};
+
+      /**
+       * The default value of the parameter. If this optional fields is set, the parameter is
+       * optional. If it is not set, the parameter is required.
+       */
+      std::optional<StoredType> default_value{};
+    };
+
+    struct GroupData
```
lauraengelhardt, line 379: Same for the `GroupData`'s documentation.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,882 @@
...
+        std::vector<InputSpec> specs;
+
+        void parse(ValueParser& parser, InputParameterContainer& container) const;
+        void set_default_value(InputParameterContainer& container) const;
+        void print(std::ostream& stream, const InputParameterContainer& container) const;
+      };
+
+      struct OneOfSpec
+      {
+        // A one_of spec is an unnamed group with special logic.
```
lauraengelhardt, line 471: Special logic? Does it correspond to the switch?

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,882 @@
...
+        void print(std::ostream& stream, const InputParameterContainer& container) const;
+      };
+    }  // namespace Internal
+
+    /**
+     * Create a normal entry. All entries are parameterized by a struct which contains the optional
+     * `description`, `required` and `default_value` fields. The following examples demonstrate how
+     * entries can be created:
+     *
+     * @code
```
lauraengelhardt, line 488: Very nice!

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,882 @@
...
+     * @endcode
+     *
+     * After parsing an InputSpec with fully_parse(), the value of the entry can be retrieved from
+     * an InputParameterContainer. The details depend on the `required` and `default_value` fields:
+     *
+     *   - `required` is used to determine whether the parameter is required in the input file. Note
+     *     that by default, if `required` is not set, the parameter is implicitly required. Failing
+     *     to read a required parameter will result in an exception. On the other hand, if a
+     *     parameter is not required, it is optional and can be left out in the input file. The
+     *     container that is filled by the fully_parse() function will not contain the parameter.
```
lauraengelhardt, line 517:
```suggestion
     *   - `required` is used to determine whether the parameter is required in the input file. Note
     *     that by default, if `required` is not set, the parameter is implicitly required. Failing
     *     to read a required parameter will result in an exception. If sucessfully parsed, the container that is filled by the fully_parse() function will contain the parameter and its value. On the other hand, if a
     *     parameter is not required, it is optional and can be left out in the input file. The
     *     container that is filled will not contain the non-required parameter.
```

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -0,0 +1,882 @@
...
+     *    entry<double>("alpha_m"),
+     *    entry<double>("gamma"),
+     *  }),
+     *  });
+     * @endcode
+     *
+     * This InputSpecs requires either the "OneStepTheta" group or the "GenAlpha" group to be
+     * present in the input. If both or none of them are present, an exception is thrown.
+     */
+    [[nodiscard]] InputSpec one_of(std::vector<InputSpec> specs);
```
lauraengelhardt, line 726: This is similar to the `SwitchComponent`, right?

src/core/io/src/4C_io_input_parameter_container.hpp
```diff
@@ -210,6 +222,9 @@ namespace Core::IO
 
     //! a map to store anything
     std::map<std::string, std::any> anydata_;
+
+    //! Groups present in this container
```
lauraengelhardt, line 226:
```suggestion
    //! Groups present in this container.
    //! Each group is an InputParameterContainer, identified by its name as the map key.
```

lauraengelhardt (commented):
I left some comments, mostly concerning the documentation and some more descriptive error messages (even though they are already better than before - so consider this rather a suggestion):

src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -0,0 +1,430 @@
...
+      EXPECT_EQ(b[2], 3.0);
+      EXPECT_EQ(container.get<std::string>("c"), "string");
+    }
+
+    {
+      InputParameterContainer container;
+      std::string stream("a 3 b 1.0 2.0 c string");
+      ValueParser parser(stream);
+      FOUR_C_EXPECT_THROW_WITH_MESSAGE(fully_parse(parser, line, container), Core::Exception,
+          "Could not parse 'c' as a double value");
```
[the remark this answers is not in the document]
sebproell: Yes, the error messages can be improved further. The reason I have not done this yet is that the `.dat` format is bound to be deprecated, and much of the parsing logic is simpler for YAML.

src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    {
+      InputParameterContainer container;
+      std::string stream("a 1 group c string");
+      // Note: we start to parse the group, but the entries are not complete, so we backtrack.
+      // The result is that the parts of the group remain unparsed.
+      ValueParser parser(stream);
+      FOUR_C_EXPECT_THROW_WITH_MESSAGE(
+          fully_parse(parser, line, container), Core::Exception, "still contains 'group c string'");
```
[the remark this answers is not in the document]
sebproell: Fully agree. Ideally, you want a kind of stack trace of the input file. I already have something in mind for that, but as I said above, I will tackle this later.

src/core/io/src/4C_io_input_spec_builders.hpp
sebproell: Actually, I don't even need this struct any longer due to my latest changes. Will be removed.

src/core/io/src/4C_io_input_spec_builders.hpp
sebproell: Exactly. `one_of` can be used to achieve what `SwitchComponent` does.

lauraengelhardt (approved):
Thanks for tackling this! I have no further comments ☺️

amgebauer (approved):

PR #169, opened 2024-12-17
CMake: automatically pick up unit tests

sebproell:
This PR implements the same logic for unit tests that we already use for sources. We no longer need to spell out every test file in the `CMakeLists.txt`. The number of MPI ranks and OMP threads is detected based on the file extension.


cmake/functions/four_c_auto_define_tests.cmake
```diff
@@ -0,0 +1,230 @@
...
+  cmake_parse_arguments(
+    _parsed
+    "${options}"
+    "${oneValueArgs}"
+    "${multiValueArgs}"
+    ${ARGN}
+    )
+
+  if(DEFINED _parsed_UNPARSED_ARGUMENTS)
+    message(SEND_ERROR "There are unparsed arguments: ${_parsed_UNPARSED_ARGUMENTS}")
```
davidrudlstorfer, line 21:
```suggestion
    message(FATAL_ERROR "There are unparsed arguments: ${_parsed_UNPARSED_ARGUMENTS}")
```

Should we use a fatal error here? https://cmake.org/cmake/help/v3.16/command/message.html#:~:text=FATAL_ERROR,continue%20processing%2C%20but%20skip%20generation.

We used a fatal error in other places and this would stop the processing and generation

cmake/functions/four_c_auto_define_tests.cmake
```diff
@@ -0,0 +1,230 @@
...
+    MPIEXEC_EXTRA_OPTS_FOR_TESTING_LIST UNIX_COMMAND ${MPIEXEC_EXTRA_OPTS_FOR_TESTING}
+    )
+
+  set(mpi_arguments
+      ${MPIEXEC_EXTRA_OPTS_FOR_TESTING_LIST}
+      -np
+      1
+      $<TARGET_FILE:${_target}>
+      --gtest_output=xml:unittest_reports/${_target}_report.xml
+      )
```
davidrudlstorfer, line 71: We could really think about our cmake formatting settings and increase the line length 😅

davidrudlstorfer (commented):
Really nice changes! That looks flawless now.

I only have two minor comments.

@sebproell and what do you think if we add some upper limit of procs per test? For example throw an error if someone wants to run a unit test with more than 5 procs

sebproell:
> @sebproell and what do you think if we add some upper limit of procs per test? For example throw an error if someone wants to run a unit test with more than 5 procs

I agree that using 5 ranks is usually questionable. I don't like to maintain another check for this, though.

davidrudlstorfer (approved):

sebproell:
I'll close to test how Slack notifications work. Will reopen.

amgebauer (approved):

PR #173, opened 2024-12-23
Introduce typos hook

sebproell:
Typos are annoying to fix during review (if they are even detected there). Luckily, there is [typos](https://github.com/crate-ci/typos), a configurable and fast spell checker specifically written for code bases. This PR enables `typos` in the pre-commit hooks. I suggest going over the different file types we have and start enforcing the check.

Special words that we do not want to be corrected are listed in `.typos.toml`. For now, I only applied the check to parts of the C++ sources to find the special cases we need to ignore. Of course, we could also fix some of these cases in the future.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,6 +72,15 @@ repos:
           # Allow the first line to be a shebang separated by a newline.
           - --detect-license-in-X-top-lines=3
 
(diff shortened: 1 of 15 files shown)


davidrudlstorfer (approved):
Nice! Also that the `thr` went away 🚀

mayrmt (approved):

PR #177, opened 2024-12-28
Use InputSpec in internals of MaterialDefinition

sebproell:
Part of #73, Follows #165

### DIFF
--- a/src/mat/4C_mat_electrode.cpp
+++ b/src/mat/4C_mat_electrode.cpp
@@ -24,7 +24,8 @@ Mat::PAR::Electrode::Electrode(const Core::Mat::PAR::Parameter::Data& matdata)
       chimax_(matdata.parameters.get<double>("CHI_MAX")),
       ocpmodel_(string_to_ocp_model(matdata.parameters.get<std::string>("OCP_MODEL"))),
       ocpparanum_(matdata.parameters.get<int>("OCP_PARA_NUM")),
-      ocppara_(matdata.parameters.get<std::vector<double>>("OCP_PARA")),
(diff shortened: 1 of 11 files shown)


amgebauer (approved):

src/mat/4C_mat_electrode.cpp
```diff
@@ -24,7 +24,8 @@ Mat::PAR::Electrode::Electrode(const Core::Mat::PAR::Parameter::Data& matdata)
       chimax_(matdata.parameters.get<double>("CHI_MAX")),
       ocpmodel_(string_to_ocp_model(matdata.parameters.get<std::string>("OCP_MODEL"))),
       ocpparanum_(matdata.parameters.get<int>("OCP_PARA_NUM")),
-      ocppara_(matdata.parameters.get<std::vector<double>>("OCP_PARA")),
+      ocppara_(matdata.parameters.get_or<std::vector<double>>(
+          "OCP_PARA", std::vector<double>(ocpparanum_, 0.0))),
```
c-p-schmidt, line 28:
```suggestion
          "OCP_PARA", std::vector<double>(ocpparanum_, std::numeric_limits<double>::infinity()))),
```

I think some default value that is certainly wrong/unphysical would be better here. But this I can also do in a follow-up. I anyways already have an almost finished branch with changes to this file.

c-p-schmidt (approved):

src/mat/4C_mat_electrode.cpp
sebproell: Agreed. Many of the `.default_value` parameters should rather be `.required = false`. I mostly did a regex replace in this PR.

PR #181, opened 2025-01-03
[Mat] Adding anisotropic thermo material

maxfirmbach:
## Description and Context

This PR adds the anisotropic version of Fourier's law to 4C following:
```math
q = -\kappa{\nabla}T
```
with
```math
\kappa=
\begin{bmatrix}
\kappa_x & 0 & 0 \\
0 & \kappa_y & 0 \\
0 & 0 & \kappa_z 
\end{bmatrix}
```

Added unit-testing and a very rudimentary framework test, still thinking about a proper one. The material implementation is a simple copy of the isotropic Fourier's law, but adding the thermal conductivity as vector quantity and therefore changing some methods like the evaluate() call.

I put this into draft mode as one might come up with a more elegant implementation of the material itself, e.g. combining both Fourier's law into one single material, this might not be so trivial though ...


tests/input_files/thermo3D_FBC_anisotropic_statics.dat
```diff
@@ -0,0 +1,113 @@
...
+TIMESTEP                        1.0
+TOLRES                          1e-08
+TOLTEMP                         1e-10
+------------------------------------------------THERMAL DYNAMIC/ONESTEPTHETA
+THETA                           0.5
+--------------------------------------------------------------------SOLVER 1
+NAME                            Thermal_Solver
+SOLVER                          UMFPACK
+-------------------------------------------------------------------MATERIALS
+MAT 1 THERM_FourierAniso CAPA 420 NUMSPACEDIM 3 CONDUCT 52 52 52
```
amgebauer, line 50: It would make sense to actually use an anisotropic material here

src/inpar/4C_inpar_validmaterials.cpp
```diff
@@ -2170,6 +2170,22 @@ std::shared_ptr<std::vector<std::shared_ptr<Mat::MaterialDefinition>>> Input::va
...
+  {
+    auto m = std::make_shared<Mat::MaterialDefinition>("THERM_FourierAniso",
+        "anisotropic (linear) Fourier's law of heat conduction",
+        Core::Materials::m_th_fourier_aniso);
+
+    add_named_real(m, "CAPA", "volumetric heat capacity");
+    add_named_int(m, "NUMSPACEDIM", "Number of space dimension");
+    add_named_real_vector(m, "CONDUCT",
+        "Diagonal tensor values defining the thermal conductivity in each space direction",
+        "NUMSPACEDIM");
```
amgebauer, line 2184: Can we allow any positive definite tensor here instead of only a diagonal one?

src/mat/4C_mat_fourieraniso.cpp
```diff
@@ -0,0 +1,131 @@
...
+ *----------------------------------------------------------------------*/
+void Mat::FourierAniso::pack(Core::Communication::PackBuffer& data) const
+{
+  // pack type of this instance of ParObject
+  int type = unique_par_object_id();
+  add_to_pack(data, type);
+
+  // matid
+  int matid = -1;
+  if (params_ != nullptr) matid = params_->id();  // in case we are in post-process mode
```
amgebauer, line 65: I don't think that pack is called in post-process mode.

```suggestion
  int matid = params_->id();
```

src/mat/4C_mat_fourieraniso.cpp
```diff
@@ -0,0 +1,131 @@
...
+    if (Global::Problem::instance()->materials()->num() != 0)
+    {
+      const int probinst = Global::Problem::instance()->materials()->get_read_from_problem();
+      Core::Mat::PAR::Parameter* mat =
+          Global::Problem::instance(probinst)->materials()->parameter_by_id(matid);
+      if (mat->type() == material_type())
+        params_ = static_cast<Mat::PAR::FourierAniso*>(mat);
+      else
+        FOUR_C_THROW("Type of parameter material %d does not fit to calling type %d", mat->type(),
+            material_type());
```
amgebauer, line 89:
```suggestion
        FOUR_C_ASSERT_ALWAYS(mat->type() == material_type(), "Type of parameter material %d does not fit to calling type %d", mat->type(), material_type());
            
      params_ = static_cast<Mat::PAR::FourierAniso*>(mat);
```

vryy (commented):
The logic of this PR is fine to me.

src/mat/4C_mat_fourieraniso.cpp
```diff
@@ -0,0 +1,131 @@
...
+void Mat::FourierAniso::evaluate(const Core::LinAlg::Matrix<3, 1>& gradtemp,
+    Core::LinAlg::Matrix<3, 3>& cmat, Core::LinAlg::Matrix<3, 1>& heatflux) const
+{
+  // conductivity tensor
+  cmat.clear();
+  for (int i = 0; i < 3; ++i) cmat(i, i) = params_->conduct_[i];
+
+  // heatflux
+  heatflux.multiply_nn(cmat, gradtemp);
+}
```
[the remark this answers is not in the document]
sebproell: This is not easily possible since `evaluate` is a virtual function and cannot be a template.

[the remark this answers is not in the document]
sebproell: Maybe, but the more significant issue for me is not to duplicate 300 lines of the boilerplate of `FourierIso` just to make the conductivity anisotropic. This can be integrated into the old material which should probably just be called `Fourier` then.

maxfirmbach:
I think one main issue is that in the isotropic case, the conductivity is given as a scalar value and several places in the code retrieve the "isotropic" conductivity as scalar value. In the general case it would be a vector or even a matrix, how would one define the isotropic case there? Do some special procedure to return a scalar again?

sebproell:
@maxfirmbach I checked, and there are only a few places that call the `conductivity()` function. The ones in scatra actually cast to `Mat::Soret`. The ones in so3 should be irrelevant since this code will be removed. So it should be simple to remove the scalar conductivity function.

maxfirmbach:
> @maxfirmbach I checked, and there are only a few places that call the `conductivity()` function. The ones in scatra actually cast to `Mat::Soret`. The ones in so3 should be irrelevant since this code will be removed. So it should be simple to remove the scalar conductivity function.

@sebproell `Mat::Soret` inherits from `Mat::FourierIso`, so they always go hand in hand. Is there a timeline for the deletion of the old `so3` elements/functionality?

sebproell:
@maxfirmbach You can remove the `conductivity()` function from `FourierIso` and only keep it in `Soret`. The inheritance relation could be removed as well.

> Is there a timeline for the deletion of the old so3 elements/functionality?

@amgebauer?

tests/input_files/thermo3D_FBC_anisotropic_statics.dat
maxfirmbach: This test will be replaced and currently guarantees that the material, at minimal, can represent the isotropic case by using a full conductivity tensor.

src/mat/4C_mat_fourieraniso.cpp
maxfirmbach: Removed the if statement.

src/mat/4C_mat_fourieraniso.cpp
maxfirmbach: Changed the respective code parts.

src/inpar/4C_inpar_validmaterials.cpp
```diff
@@ -2585,13 +2585,28 @@ std::shared_ptr<std::vector<std::shared_ptr<Mat::MaterialDefinition>>> Input::va
...
+  {
+    auto m = std::make_shared<Mat::MaterialDefinition>("MAT_Fourier",
+        "anisotropic (linear) Fourier's law of heat conduction", Core::Materials::m_thermo_fourier);
+
+    m->add_component(entry<double>("CAPA", {.description = "volumetric heat capacity"}));
+    m->add_component(entry<int>(
+        "CONDUCT_PARA_NUM", {.description = "Number of thermal conductivity parameters"}));
+    m->add_component(entry<std::vector<double>>(
+        "CONDUCT", {.description = "Values defining the row-wise thermal conductivity tensor",
+                       .size = from_parameter<int>("CONDUCT_PARA_NUM")}));
```
mayrmt, line 2609:
```suggestion
    m->add_component(entry<int>(
        "CONDUCT_PARA_NUM", {.description = "Number of entries in the thermal conductivity tensor"}));
    m->add_component(entry<std::vector<double>>(
        "CONDUCT", {.description = "Values of the thermal conductivity tensor in row-wise ordering",
                       .size = from_parameter<int>("CONDUCT_PARA_NUM")}));
```

maxfirmbach:
@amgebauer @sebproell @vryy This PR is ready for a second round of reviews regarding the `Fourier` material implementation (testing is still work in progress).

amgebauer:
> @maxfirmbach You can remove the `conductivity()` function from `FourierIso` and only keep it in `Soret`. The inheritance relation could be removed as well.
> 
> > Is there a timeline for the deletion of the old so3 elements/functionality?
> 
> @amgebauer?

Sorry, I missed this. The hit is within the so3_surface which will remain (under a different name).

amgebauer:
> I would prefer it if you adapted the old Fourier material to cover both cases. This is a tiny change where you make the conductivity tensor anisotropic. The new features to enter input will help as well, once we merge #177.

Is this really desirable? It's a different constitutive law having a scalar and a tensor version. Yes, the tensor version can also represent the isotropic case, but with the downside of a matrix-vector product instead of scaling a vector (and that you cannot easily recover the scalar conductivity, which might be necessary). Of course, that might be irrelevant for many 4C simulations. However, given the fact that these equations are written differently on paper, I would also handle them differently within the code.

What is problematic, though, is that we need to write hundreds of lines of boilerplate for such a simple constitutive model. But that's not what we can change within this PR.

src/mat/4C_mat_fourier.hpp
```diff
@@ -19,74 +16,58 @@
...
 
       /// volumetric heat capacity
       const double capa_;
+
       /// heat conductivity
-      const double conduct_;
+      const std::vector<double> conduct_;
+
+      /// number of conductivity components
+      const int conduct_para_num_;
```
vryy, line 43: I would assume `conduct_para_num_` is the same as `conduct_.size()`, no?

src/mat/4C_mat_fourier.hpp
```diff
@@ -19,74 +16,58 @@
...
-  ///
-  /// This is a FourierIso's law of isotropic, instationary heat conduction
-  ///
-  /// <h3>References</h3>
-  /// <ul>
-  /// <li> [1] GA Holzapfel, "Nonlinear solid mechanics", Wiley, 2000.
-  /// </ul>
-  ///
-  /// \author dano
-  /// \date 09/09
```
vryy, line 81: I would suggest to keep this reference, and add `The anisotropic version is the extension of isotropic case with thermal conductivity treated as tensor.`

tests/input_files/consinter3D_tsi_hex27_hex27.dat
```diff
@@ -105,7 +105,7 @@ TEKO_XML_FILE                   xml/block_preconditioner/thermo_solid.xml
 AZREUSE                         10
 -------------------------------------------------------------------MATERIALS
 MAT 1 MAT_Struct_ThrStVenantK YOUNGNUM 1 YOUNG 2.1e08 NUE 0.0 DENS 7.85e-06 THEXPANS 1.1e-05 CAPA 0.0 CONDUCT 1030.0 INITTEMP 273.15
-MAT 2 THERM_FourierIso CAPA 0.0 CONDUCT 1030.0
+MAT 2 MAT_Fourier CONDUCT_PARA_NUM 1 CAPA 0.0 CONDUCT 1030.0
```
vryy, line 108: Is it possible to induce `CONDUCT_PARA_NUM` from the number of parameters since it's an inherent property? If it's not possible due to the logic of the code, then leave it as it is.

unittests/mat/CMakeLists.txt
```diff
@@ -5,4 +5,4 @@
 #
 # SPDX-License-Identifier: LGPL-3.0-or-later
 
-four_c_auto_define_tests(mat)
+four_c_auto_define_tests(mat)
```
vryy, line 8: probably some mismatching file ending

src/mat/4C_mat_fourier.cpp
```diff
@@ -0,0 +1,203 @@
...
+      conduct_para_num_(matdata.parameters.get<int>("CONDUCT_PARA_NUM"))
+{
+  /*
+  TODO: there are 5 options ...
+ - one value is given -> scalar case (this is our default case)
+ - two values are given -> 2d diagonal
+ - three values are given -> 3d diagonalCONDUCT_PARA_NUM
+ - four values are given -> 2d full
+ - nine values are given -> 3d full
+ */
```
vryy, line 32: something still needs to be added?

src/mat/4C_mat_fourier.cpp
```diff
@@ -0,0 +1,203 @@
...
+      break;
+    }
+    case 4:  // full tensor is given
+    {
+      cmat(0, 0) = params_->conduct_[0];
+      cmat(0, 1) = params_->conduct_[1];
+      cmat(1, 0) = params_->conduct_[2];
+      cmat(1, 1) = params_->conduct_[3];
+      break;
+    }
```
vryy, line 141: obviously, the `CONDUCT_PARA_NUM` parameter is to cover the non-diagonal thermal conductivity tensor, rather than the dimension. In that case, I would name it `CONDUCT_TYPE` or similar.

sebproell:
> What is problematic, though, is that we need to write hundreds of lines of boilerplate for such a simple constitutive model. But that's not what we can change within this PR.

@amgebauer I agree that this is the underlying problem, and my desire to unify the two models was mostly for this reason.

maxfirmbach:
@sebproell @amgebauer So the current version is implementing a generic version of `Fourier`'s law using the underlying tensor, with different options to define it.

src/mat/4C_mat_fourieriso.cpp
```diff
@@ -1,137 +0,0 @@
...
-/*----------------------------------------------------------------------*
- |  calculate for 3D                                         dano 09/09 |
- *----------------------------------------------------------------------*/
-void Mat::FourierIso::evaluate(const Core::LinAlg::Matrix<3, 1>& gradtemp,
-    Core::LinAlg::Matrix<3, 3>& cmat, Core::LinAlg::Matrix<3, 1>& heatflux) const
-{
-  // conductivity tensor
-  cmat.clear();
-  for (int i = 0; i < 3; ++i) cmat(i, i) = params_->conduct_;
-
```
maxfirmbach, line 132: E.g. the conductivity value was given as scalar in the input file, but internally the tensor is still created by setting the values onto the diagonal, so the matrix-vector is done anyway.

amgebauer:
> @sebproell @amgebauer So the current version is implementing a generic version of `Fourier`'s law using the underlying tensor, with different options to define it. The old implementation is not scalar either ... it just put's the isotropic scalar value on the diagonal.

Interesting. For the isotropic case, this seems to be an over-complicated implementation. I'm totally fine to leave it like so. We might implement a faster isotropic implementation once we can define materials with less boilerplate.

src/mat/4C_mat_fourieraniso.cpp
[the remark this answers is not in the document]
maxfirmbach: Done 👍

src/mat/4C_mat_fourier.cpp
maxfirmbach: I took the name `THERM_PARA_NUM` as example from another material.

src/mat/4C_mat_fourier.hpp
maxfirmbach: Added the documentation again.

src/inpar/4C_inpar_validmaterials.cpp
maxfirmbach: Updated the description strings.

tests/input_files/thermo3D_FBC_anisotropic_statics.dat
maxfirmbach: Reverted this test and added a new, true anisotropic one.

maxfirmbach:
I updated some things and added a new truely anisotropic test case. Geometry is a annulus shape with thickness with an anisotropic thermal conduction on it.

tests/input_files/.goutputstream-E6T302
vryy, line 1: I think this file shall be removed

amgebauer (changes requested):

vryy (approved):
There is a small remnant. Other than that, I think that's fine.

tests/input_files/.goutputstream-E6T302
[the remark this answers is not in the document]
maxfirmbach: Right, not sure where this comes from ...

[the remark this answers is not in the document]
maxfirmbach: File deleted.

maxfirmbach: File deleted.

doc/readthedocs/elementsthermo.rst
```diff
@@ -3,6 +3,7 @@ Thermo Elements
 
 The only material, which is used for thermo elements in the input examples is
 
-- THERM_FourierIso
+- MAT_Fourier
+(- MAT_Soret)
```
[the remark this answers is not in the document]
maxfirmbach: Parentheses removed.

tests/input_files/thermo3D_annulus_anisotropic_statics.dat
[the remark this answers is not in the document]
maxfirmbach: I refactored the test case, which now has way less elements ... I also had to adapt the result slightly due to that.

doc/readthedocs/elementsthermo.rst
```diff
@@ -3,6 +3,7 @@ Thermo Elements
 
 The only material, which is used for thermo elements in the input examples is
 
-- THERM_FourierIso
+- MAT_Fourier
+- MAT_Soret
 
-it seems.
\ No newline at end of file
+it seems.
```
sebproell, line 9: What is the point of "it seems"? Please remove it.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2572,32 +2572,37 @@ std::shared_ptr<std::vector<std::shared_ptr<Mat::MaterialDefinition>>> Global::v
...
-    matsoret->add_component(entry<double>("CAPA", {.description = "volumetric heat capacity"}));
-    matsoret->add_component(entry<double>("CONDUCT", {.description = "thermal conductivity"}));
-    matsoret->add_component(entry<double>("SORET", {.description = "Soret coefficient"}));
+    m->add_component(entry<double>("CAPA", {.description = "volumetric heat capacity"}));
+    m->add_component(entry<int>(
+        "CONDUCT_PARA_NUM", {.description = "Number of thermal conductivity parameters"}));
+    m->add_component(entry<std::vector<double>>(
+        "CONDUCT", {.description = "Values defining the row-wise thermal conductivity tensor",
+                       .size = from_parameter<int>("CONDUCT_PARA_NUM")}));
+    m->add_component(entry<double>("SORET", {.description = "Soret coefficient"}));
```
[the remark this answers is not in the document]
maxfirmbach: I put a more detailed description in place.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2572,32 +2572,45 @@ std::shared_ptr<std::vector<std::shared_ptr<Mat::MaterialDefinition>>> Global::v
...
-    auto m = std::make_shared<Mat::MaterialDefinition>("THERM_FourierIso",
-        "isotropic (linear) Fourier's law of heat conduction", Core::Materials::m_th_fourier_iso);
+    auto m = std::make_shared<Mat::MaterialDefinition>("MAT_Fourier",
+        "anisotropic (linear) Fourier's law of heat conduction", Core::Materials::m_thermo_fourier);
 
     m->add_component(entry<double>("CAPA", {.description = "volumetric heat capacity"}));
-    m->add_component(entry<double>("CONDUCT", {.description = "thermal conductivity"}));
+    m->add_component(entry<int>("CONDUCT_PARA_NUM",
+        {.description = "Parameter representing the relevant number of entries in the thermal "
+                        "conductivity tensor. Setting it's value to 1 resembles a scalar "
```
sebproell, line 2583:
```suggestion
                        "conductivity tensor. Setting its value to 1 resembles a scalar "
```

src/mat/4C_mat_fourier.hpp
```diff
@@ -19,74 +16,70 @@
...
   namespace PAR
   {
-    /*----------------------------------------------------------------------*/
-    /// material parameters for FourierIso material
+    /// material parameters for Fourier material
     ///
     /// <h3>Input line</h3>
-    /// MAT 1 THERM_FourierIsoIso CAPA 1.0 COND 1.0
-    class FourierIso : public Core::Mat::PAR::Parameter
+    /// MAT 1 MAT_Fourier CAPA 1.0 NUMSPACEDIM 3 CONDUCT 1.0 0.1 10.0
```
sebproell, line 38: I would remove this. This gets outdated and is only correct for a dat input file.

maxfirmbach:
Before merging I will clean up the commits a little bit and fix the checkcode test.

src/scatra_ele/4C_scatra_ele_calc_sti_diffcond.cpp
```diff
@@ -716,11 +722,16 @@ void Discret::Elements::ScaTraEleCalcSTIDiffCond<distype>::mat_fourier(
...
-  const std::shared_ptr<const Mat::FourierIso> matfourier =
-      std::static_pointer_cast<const Mat::FourierIso>(material);
+  const std::shared_ptr<const Mat::Fourier> matfourier =
+      std::static_pointer_cast<const Mat::Fourier>(material);
+
   densn = densnp = densam = matfourier->capacity();
-  diff_manager()->set_isotropic_diff(matfourier->conductivity(), 0);
-}  // Discret::Elements::ScaTraEleCalcSTIDiffCond<distype>::mat_soret
+
+  std::vector<double> k = matfourier->conductivity();
```
sebproell, line 730:
```suggestion
 const std::vector<double>& k = matfourier->conductivity();
```

Not essential but there is no reason to make a copy. On the other hand, the hacky code you modified here (not yours) has more serious problems.

doc/readthedocs/elementsthermo.rst
maxfirmbach: Removed and description updated.

tests/input_files/thermo3D_annulus_anisotropic_statics.dat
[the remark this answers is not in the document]
amgebauer: It is still a very large test with now 2232 elements. The material law is so simple that a test with <10 elements suffices. I think I have seen a test with just 3 elements in a prior version. Why not just taking that one?

mayrmt (changes requested):
Some minor requests for change, mostly related to language and documentation.

tests/input_files/thermo3D_annulus_anisotropic_statics.dat
```diff
@@ -0,0 +1,6185 @@
...
+-------------------------------------------------------------FUNCT1
+COMPONENT 0 SYMBOLIC_FUNCTION_OF_SPACE_TIME 100
+-------------------------------------------------------------FUNCT2
+COMPONENT 0 SYMBOLIC_FUNCTION_OF_SPACE_TIME 1
+----------------------------------------DESIGN SURF DIRICH CONDITIONS
+DSURF 2
+// dirichlet_surf_inner
+E 1 - NUMDOF 1 ONOFF 1 VAL 1.0 FUNCT 1
+// dirichlet_surf_outer
+E 2 - NUMDOF 1 ONOFF 1 VAL 1.0 FUNCT 2
```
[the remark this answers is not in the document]
maxfirmbach: @mayrmt I would still define it via functions to be honest … even if it‘s just constant values.

tests/input_files/thermo3D_annulus_anisotropic_statics.dat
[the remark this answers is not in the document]
maxfirmbach: I would like to use multigrid on it, everything element based should be covered by unit-tests. I can still add a framework test with just three elements, if necessary …. There will be a follow up PR using that framework test too.

vryy (approved):

amgebauer (approved):

PR #184, opened 2025-01-06
Move poro with Nitsche contact to new time integration

SchaefP1:
## Description and Context
This PR allows Nitsche contact for poroelasticity with new time integration. This requires some small changes in the set-up, implementing the function `CONTACT::NitscheStrategyPoro::integrate` and accessing the contact integrator in `PoroElast::Monolithic`. The new test files are created as copy of existing test files modified to use new time integration. One result check was slightly adapted, all other remain the same as in the old time integration.

Part of #146


src/contact/4C_contact_strategy_factory.cpp
```diff
@@ -538,6 +539,20 @@ void CONTACT::STRATEGY::Factory::read_and_check_input(Teuchos::ParameterList& pa
...
     }
+    case Core::ProblemType::poroelast:
+    case Core::ProblemType::poroscatra:
+    {
+      const Teuchos::ParameterList& stru = Global::Problem::instance()->structural_dynamic_params();
+      const Teuchos::ParameterList& porodyn =
+          Global::Problem::instance()->poroelast_dynamic_params();
+      // porotimefac = 1/(theta*dt) --- required for derivation of structural displacements!
+      double porotimefac =
+          1 / (stru.sublist("ONESTEPTHETA").get<double>("THETA") * stru.get<double>("TIMESTEP"));
```
c-p-schmidt, line 550: So, it only works for OST. Do you make sure that you can only call it with OST?

src/contact/4C_contact_strategy_factory.cpp
```diff
@@ -538,6 +539,20 @@ void CONTACT::STRATEGY::Factory::read_and_check_input(Teuchos::ParameterList& pa
...
+      params.set<double>("porotimefac", porotimefac);
+      params.set<bool>(
+          "CONTACTNOPEN", porodyn.get<bool>("CONTACTNOPEN"));  // used in the integrator
```
c-p-schmidt, line 553:
```suggestion
          "CONTACT_NO_PENETRATION", porodyn.get<bool>("CONTACT_NO_PENETRATION"));  // used in the integrator
```

c-p-schmidt (commented):
So you need to keep the old time integration implementation as well? Or, why don't you just delete the code and input files concerning Nitsche contact using the old time integration?

SchaefP1:
> So you need to keep the old time integration implementation as well? Or, why don't you just delete the code and input files concerning Nitsche contact using the old time integration?

Thanks for the review! Not everything in poroelastic works in the new time integration at the moment so I would suggest to keep Nitsche contact using the old time integration until we moved everything.

src/contact/4C_contact_strategy_factory.cpp
SchaefP1: Done.

c-p-schmidt (approved):

knarfnitram (approved):

PR #188, opened 2025-01-07
Drop GCC9 from Github actions

sebproell:
GCC9 only supports a small subset of the C++20 standard and will no longer be supported.

Part of #65, Blocks #130


.github/workflows/docker.yml
```diff
@@ -53,33 +37,6 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
           image-name: ${{ env.REGISTRY }}/${{ env.PROJECT_NAMESPACE }}/4c-dependencies-ubuntu24.04
 
-  build-and-push-ubuntu20_04:
```
sebproell, line 56: I assume there is no point in keeping this image?

sebproell:
> We ran the gcc9 test without assertions and gcc13 with assertions. The build without assertions is still covered by the clang build, right?

Yes.

Issue #189, opened 2025-01-07
[Trilinos] Migration to new package stack meta-issue


maxfirmbach:
# Description

This issue is for organizing and documenting progress of the migration from the old `Epetra` stack to the modern `Tpetra` stack. Specific tasks to be done are organized in issues with the respective label marking them. An overview of the progress can be found in [Migration from Epetra to Tpetra](https://github.com/4C-multiphysics/4C/milestone/1).

Goals:

- Migrate the entire linear algebra stack to `Tpetra`
- Switch 4C's Trilinos dependencies to the `Tpetra` stack

# Background

4C relies on Trilinos for the implementation of sparse linear algebra capabilities.
We currently use the `Epetra` stack, which will be phased-out from Trilinos by the end of 09/2025 (at the latest) -- and even right now, the `Epetra` stack is not actively developed anymore, so we cannot easily benefit from new features. Finally, `Epetra` is limited to MPI-only mode, so we cannot use on-node parallelism or accelerator devices (as they appear in more and more of todays computing clusters).

The modern `Tpetra` stack is the successor of `Epetra`. It is actively developed, it forms the basis for a modern solver stack in Trilinos, and comes with a hand-in-hand `Kokkos` integration for MPI+X parallelism.

The switch to `Tpetra` is mandatory, such that we can benefit from developments in Trilinos in the future.

The current strategy is to divide the transition into tree phases to ensure an orderly transition. The first phase includes preliminary steps in which the dependency on `Epetra` is reduced. If possible, the functionalities which still remains from `Epetra` are encapsulated by using wrappers.
In the second phase, the conversion of the sparse matrices will take place.
In the third and final phase, all remaining dependencies on `Epetra` that were not removed in the previous phases will be removed.

A presentation about packages, which will be archived can be found [HERE](https://trilinos.github.io/pages/community/trilinos_user_meetings/TUG_2023/2023-10-31/03_New-Leadership-Structure_Ober_2023-10-31_TUG_Trilinos_New_Leadership_Structure.pdf) (as presented at TUG 2023). Those include the following relevant packages for BACI:
- `Epetra` / `EpetraExt`
- `Isorropia`
- `Intrepid`
- `Amesos`
- ~~`AztecOO`~~
- `Ifpack`
- ~~`ML`~~

In addition, `Epetra` code from all other packages will be removed.

[...]

Issue #176, opened 2024-12-28
[Trilinos] Update to version XX


maxfirmbach:
## Description
At some point we need to update Trilinos again. This issue documents what other issues or PR's are related to it and are relevant for 4C:
- Getting https://github.com/trilinos/Trilinos/pull/13505, which increases the order of quadrature rules in `Intrepid2`.
- Getting https://github.com/trilinos/Trilinos/pull/13578, which resolves a bug in the `Kokkos` based aggregation in `MueLu`.
- Getting https://github.com/trilinos/Trilinos/pull/13608, which resolves a bug in the distance Laplacian dropping for the non-`Kokkos` based aggregation in `MueLu`.
- Getting https://github.com/trilinos/Trilinos/pull/13696, which enables on-node reindexing of non-contigous GIDs of `Epetra` style operators for solvers like `Umfpack` and `KLU2` in `Amesos2`.

mayrmt:
FYI, there is a new Trilinos release with official deprecations of the Epetra stack: https://github.com/trilinos/Trilinos/releases/tag/trilinos-release-16-1-0

Removal of the Epetra stack in the fall of 2025 is confirmed once again.

PR #191, opened 2025-01-07
Draft: Refactor Gauss points and move to Intrepid2

maxfirmbach:
## Description and Context

This MR introduces a small refactoring of the Gauss points and deletion of the Gauss point cache done by @sebproell. In addition the `Intrepid` based functions and classes are moved to `Kokkos` and `Intrepid2`.

Currently put in draft mode, as a `Trilinos` update beforehand is mandatory for merging.

* Blocked by #176 
* Part of #189

### DIFF
--- a/src/core/fem/src/general/utils/4C_fem_general_utils_polynomial.cpp
+++ b/src/core/fem/src/general/utils/4C_fem_general_utils_polynomial.cpp
@@ -10,8 +10,8 @@
 #include "4C_fem_general_utils_integration.hpp"
 #include "4C_utils_singleton_owner.hpp"
 
-#include <Intrepid_FieldContainer.hpp>
-#include <Intrepid_PointTools.hpp>
+#include <Intrepid2_PointTools.hpp>
+#include <Kokkos_DynRankView.hpp>
 #include <Shards_CellTopology.hpp>
 
 FOUR_C_NAMESPACE_OPEN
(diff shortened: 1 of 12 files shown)


sebproell:
@maxfirmbach I think you need to use `#if`, not `#ifdef` ;)

src/core/fem/src/general/utils/4C_fem_general_utils_polynomial.cpp
```diff
@@ -10,8 +10,13 @@
 #include "4C_fem_general_utils_integration.hpp"
 #include "4C_utils_singleton_owner.hpp"
 
+#if FOUR_C_TRILINOS_INTERNAL_VERSION_GE(2025, 2)
+#include <Intrepid2_PointTools.hpp>
```
eulovi, line 14: With the Trilinos update, do you still need this guarded code? This applies for all the `#if FOUR_C_TRILINOS_INTERNAL_VERSION_GE`

mayrmt: @eulovi Yes, this is still needed, because we also still support older Trilinos versions, that do not have Intrepid2 enabled. The simultaneous support of different versions allows us to update Trilinos without having to force everyone update right away.

eulovi (approved):

PR #193, opened 2025-01-08
Update clang-format

sebproell:
We are using a rather old version of clang-format. To make sure it can properly format newer language constructs, I suggest updating to the latest version. Since this implies some reformatting I also activated the setting to always align pointers/references at the type.


.clang-format
```diff
@@ -13,7 +13,12 @@ ColumnLimit: 100 # Break lines after 100 columns/characters
 MaxEmptyLinesToKeep: 3 # Max. number of emtpy lines
 NamespaceIndentation: All # Add one level of indentation per namespace nesting
 SpacesInContainerLiterals: false # Do not insert spaces after and before parenthesis, here []
-Standard: c++17 # Use the Cpp17 code style
+
+# Always put '&' and '*' with the type.
+DerivePointerAlignment: false
+PointerAlignment: Left
+
+Standard: c++20 # Use the Cpp17 code style
```
amgebauer, line 21:
```suggestion
Standard: c++20 # Use the Cpp20 code style
```
or (what I prefer):
```suggestion
Standard: c++20
```

amgebauer (changes requested):

.clang-format
mayrmt: I'd also vote for the second option. This is self-explaining, so no need for a comment.

sebproell: Agreed. I'll also update/remove the other comments that are not useful.

mayrmt (approved):

amgebauer (approved):

PR #195, opened 2025-01-08
Replace SFINAE in core with C++20 concepts

sebproell:
Use the more expressive and powerful C++ 20 `requires` syntax instead of `std::enable_if`. I applied it to all code in the core module and some places that use the same concepts. I also added the case style for concepts to our `clang-tidy` config.


bgoderbauer (approved):
I really like this new syntax! It makes it much more readable, IMO.

amgebauer (approved):

Issue #50, opened 2024-11-29
Implement solid shell elements (hex8, wedge6) in new solid elements


amgebauer:
## Description

Move solid shell element to new solid elements

## Possible Solution and Definition of Done

* [x] Implement shell ANS element technology
    * [x] hex8
    * [x] wedge6
* [x] Implement shell EAS element technology
    * [x] hex8
* [x] Implement shell EAS+ANS element technology #227 
    * [x] hex8 #227 
    * [x] wedge6 #227 
* [ ] Delete old solid shell elements (and last remaining test with stc)

## Related Issues and Merge Requests

* Related to https://gitlab.lrz.de/baci/baci/-/issues/872
* Part of https://gitlab.lrz.de/baci/baci/-/issues/868

amgebauer:
@isteinbrecher there is one test remaining with stc (scaled thickness conditioning) and solid shell elements. Is it still needed? The test (stc-mltube-sosh8.dat) has been added 14 years ago.

isteinbrecher:
I am not familiar with that one. @mayrmt do you know more about that?

mayrmt:
STC has been proposed in this paper:

```
@article{Kloeppel2011b,
	author = {Kl\"oppel, Thomas and Gee, Michael W. and Wall, Wolfgang A.},
	journal = {Computer Methods in Applied Mechanics and Engineering},
	number = {9--12},
	pages = {1301--1310},
	title = {A scaled thickness conditioning for solid- and solid-shell discretizations of thin-walled structures},
	volume = {200},
	year = {2011}}
```
It is a technique to improve the conditioning of linear systems arising from shell models. If possible, I'd keep it.

amgebauer:
> If possible, I'd keep it

ok. Keeping itself is not possible, we need to reimplement it for the new solid elements. We need someone who will do it. I can guide where we can put the implementation.

mayrmt:
@amgebauer do you mean that the STC stuff needs to be re-implemented in the new solid elements?

amgebauer:
> @amgebauer do you mean that the STC stuff needs to be re-implemented in the new solid elements?

yes

mayrmt:
@amgebauer STC can be removed, since it is tied to a very specific element formulation and has not been actively used for several years.

PR #227, opened 2025-01-09
Implement solid shell element with EAS+ANS

amgebauer:
This implements the remaining missing functionality of EAS+ANS.

Part of #50


src/solid_3D_ele/4C_solid_3D_ele_calc_lib_shell_ans.hpp
```diff
@@ -0,0 +1,568 @@
...
+FOUR_C_NAMESPACE_OPEN
+namespace Discret::Elements
+{
+  template <Core::FE::CellType celltype>
+  struct SamplingPoints
+  {
+  };
+
+  template <>
+  struct SamplingPoints<Core::FE::CellType::hex8>
```
isteinbrecher, line 26: What exactly are the `SamplingPoints`?

src/solid_3D_ele/4C_solid_3D_ele_calc_lib_shell_ans.hpp
```diff
@@ -0,0 +1,568 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_SOLID_3D_ELE_CALC_LIB_SHELL_ANS_HPP
```
isteinbrecher, line 8: Would it be possible to unify the implementation in this file to not have the two implementations that are not exactly duplicate code but look very much the same?

Also, maybe the file name `src/solid_3D_ele/4C_solid_3D_ele_calc_shell_ans_lib.hpp` would be better then, both files are are next to each other in the repository directory tree.

tests/input_files/solid_ele_wedge6_Standard_shell_eas_ans.4C.yaml
```diff
@@ -0,0 +1,71 @@
+# Test of the templated implementation of solid elements
```
isteinbrecher, line 1: Are the two newly added test cases testing additional functionality compared to the old ones?

Can you remove the `.4C` in the filename?

src/solid_3D_ele/4C_solid_3D_ele_calc_lib_shell_ans.hpp
amgebauer: That are the ANS sampling points.

src/solid_3D_ele/4C_solid_3D_ele_calc_lib_shell_ans.hpp
amgebauer: Everything that is not the same is moved into the library function. It does not make much sense to merge them into one file as it's then much more complex.

I can of course move the `lib` to the end of the filename if you find that better.

tests/input_files/solid_ele_wedge6_Standard_shell_eas_ans.4C.yaml
amgebauer: >Are the two newly added test cases testing additional functionality compared to the old ones?

they are much simpler that the existing tests and consistent to the other tests that we have for the solid elements.

isteinbrecher (approved):

davidrudlstorfer (approved):

PR #231, opened 2025-01-09
Update LNM cluster presets

davidrudlstorfer:
## Description and Context

Update the LNM Cluster config to actually use the new GCC 13.2 compiler

Also add FOUR_C_OPENMPI_ROOT to remove the necessity that each user loads the module within their bash.rc file

### DIFF
--- a/presets/lnm/cluster/CMakePresets.json
+++ b/presets/lnm/cluster/CMakePresets.json
@@ -23,8 +23,6 @@
         ".lnm_cluster"
       ],
       "cacheVariables": {
-        "CMAKE_CXX_COMPILER": "/cluster/mpi/gcc/openmpi/4.1.5/bin/mpic++",
-        "CMAKE_C_COMPILER": "/cluster/mpi/gcc/openmpi/4.1.5/bin/mpicc",
         "FOUR_C_BOOST_ROOT": "/cluster/lib/gcc/9.1.0/boost_1_86_0",
         "Python3_ROOT_DIR": "/lnm/miniconda3/envs/python_3_12_env/bin"
       }
@@ -35,11 +33,7 @@
       "description": "Build release version of 4C on thought at LNM",
       "inherits": [
         ".lnm_cluster"
-      ],
-      "cacheVariables": {
-        "CMAKE_CXX_COMPILER": "/opt/openmpi/4.1.5/gcc/bin/mpic++",
-        "CMAKE_C_COMPILER": "/opt/openmpi/4.1.5/gcc/bin/mpicc"
-      }
+      ]
     }
   ]
 }


davidrudlstorfer:
@sebproell regarding our short discussion - the user still has to load cmake 3.30.5 in their bash.rc file

presets/lnm/cluster/CMakePresets.json
```diff
@@ -23,8 +23,9 @@
         ".lnm_cluster"
       ],
       "cacheVariables": {
-        "CMAKE_CXX_COMPILER": "/cluster/mpi/gcc/openmpi/4.1.5/bin/mpic++",
-        "CMAKE_C_COMPILER": "/cluster/mpi/gcc/openmpi/4.1.5/bin/mpicc",
+        "FOUR_C_OPENMPI_ROOT": "/opt/openmpi/4.1.5/",
```
sebproell, line 26:
```suggestion
        "FOUR_C_MPI_ROOT": "/opt/openmpi/4.1.5/",
```

davidrudlstorfer: Thanks - changed it

presets/lnm/cluster/CMakePresets.json
```diff
@@ -23,8 +23,9 @@
         ".lnm_cluster"
       ],
       "cacheVariables": {
-        "CMAKE_CXX_COMPILER": "/cluster/mpi/gcc/openmpi/4.1.5/bin/mpic++",
-        "CMAKE_C_COMPILER": "/cluster/mpi/gcc/openmpi/4.1.5/bin/mpicc",
+        "FOUR_C_MPI_ROOT": "/opt/openmpi/4.1.5/",
```
sebproell, line 26: Did you accidentally flip the settings for thought and bruteforce?

davidrudlstorfer: Damnit :joy: thanks for further looking into it - changed it - hopefully now its correct

amgebauer (approved):

davidrudlstorfer:
Unfortunately, both cluster configs failed and I am rather clueless why it fails. Maybe someone (@sebproell, @amgebauer) has a clue?

https://gitlab.lrz.de/baci/baci-config-lnm/-/pipelines/2298444

sebproell:
@davidrudlstorfer the reason is that the system-installed `libstc++` is used when executing 4C. @amgebauer and I kind of suspected that but thought we should give it a try. This is a frequent problem when building another compiler and trying to compile executables with it that remember which `libstc++` should be used. We can also go with the `module load` solution instead. In this case, we would drop all the paths and require a proper `module load` in CI and by our coworkers.

davidrudlstorfer:
> @davidrudlstorfer the reason is that the system-installed libstc++ is used when executing 4C. @amgebauer and I kind of suspected that but thought we should give it a try. This is a frequent problem when building another compiler and trying to compile executables with it that remember which libstc++ should be used. We can also go with the module load solution instead. In this case, we would drop all the paths and require a proper module load in CI and by our coworkers.

Thanks for the explanation - one question remains - you write that the problem happens during execution but it actually fails during linking - shouldn't this work?

For me the module load approach is fine - so let's just stick with it if it works and remove all the unnecessary variables from the presets.

sebproell:
>  Thanks for the explanation - one question remains - you write that the problem happens during execution but it actually fails during linking - shouldn't this work?

I did not check carefully and thought the problem appears once we execute 4C to generate the metadata file. You are right it happens during linking which is ...  interesting.

davidrudlstorfer:
As discussed I've now reverted the approach to the old one, i.e., load all modules within the bashrc file and I've removed the hardcoded paths to the compilers from the cluster preset

I started a pipeline here https://gitlab.lrz.de/baci/baci-config-lnm/-/pipelines/2299473

Let's see if it works now...

davidrudlstorfer:
@amgebauer @sebproell  the pipeline succeeded on `bruteforce` and `thought`- therefore, this can be merged to finally fix the LNM pipeline

Once this is merged I will inform all colleagues regarding the module update within their .bashrc file

sebproell (approved):

amgebauer (approved):
Good that we will have these changes in our first release.

PR #233, opened 2025-01-09
Replace custom cubic spline interpolation and polynomial implementation from mat_electrode

c-p-schmidt:
## Description and Context
So far, we had a custom implementation of the evaluation of a polynomial and a cubic spline interpolation from CSV data inside the mat_electrode. However, this is also available from the function framework which is now used to replace those two custom implementations.

Some minor changes along the way.


c-p-schmidt:
@gilrrei I hope you won't miss the cryptic error message when something is wrong with the CSV file 😉

src/inpar/4C_inpar_validmaterials.cpp
```diff
@@ -1004,7 +1004,7 @@ std::shared_ptr<std::vector<std::shared_ptr<Mat::MaterialDefinition>>> Input::va
     // model for half cell open circuit potential of electrode
     matelectrode->add_component(entry<std::string>(
         "OCP_MODEL", {.description = "model for half cell open circuit potential of electrode: "
-                                     "Redlich-Kister, Taralov, Polynomial, csv"}));
+                                     "Function, Redlich-Kister, Taralov"}));
```
sebproell, line 1007: This could be a `selection` so you can map to the enum constant right away. Alternatively, you could also use something like

```
group("OCP_MODEL", {
  one_of({
    entry<int>("FUNCT_NUM"),
    group("Redlich-Kistler", {...}),
    group("Taralov", {...}),
  },
  store_index_as<OCPModels>("MODEL_TYPE")
});
```

This would make much of the validation you do later superfluous.

c-p-schmidt: Thank you for the hint and help!

dragos-ana:
Looks good to me, thanks for adding this feature! 
Is there any way to get rid of the input parameter `OCP_PARA_NUM`  for the Redlich-Kister model? This could be deduced from `OCP_PARA`, right? It was removed here for the Taralov model and I wonder which problems may arise when removing it also for Redlich-Kister.

dragos-ana (approved):
Looks good to me, thanks for adding this feature!
Is there any way to get rid of the input parameter `OCP_PARA_NUM` for the Redlich-Kister model? This could be deduced from `OCP_PARA`, right? It was removed here for the Taralov model and I wonder which problems may arise when removing it also for Redlich-Kister.

sebproell:
@dragos-ana The `OCP_PARA_NUM` is currently required because the dat file does not have any internal structure which would allow us to know how many entries to read for `OCP_PARA`. Once we switch to yaml, (soon TM) this parameter is ofc obsolete. We could only remove it for Redlich-Kister, because this model always requires 13 values.

dragos-ana:
@sebproell Thanks for the quick clarification!

sebproell (approved):

Issue #74, opened 2024-11-29
Versioning of our input file


amgebauer:
## Description

Introduce versioning of our input files and some script that can advance an old input file to a newer one.

Much easier if we would have a standard format (see https://gitlab.lrz.de/baci/baci/-/issues/981), but probably needed already for the transition)

ischeider:
In the maintainer meeting on Dec. 12, 2024 the following was decided:

### Release schedule

Releases can provide motivation to bring in developments prior to the next release. They are useful for applications to refer to a specific version of the code.

Release schedule: 1-2 releases per year

Versioning scheme: Year + number, e.g. 2025.1 (first release in 2024), 2025.2 (second release in 2025), reset minor release to 1 at the beginning of each year

ischeider:
Personally, I wonder whether a patch level indicating the (nightly?) build version can be added. Or might we have an internal versioning in addition to the official tags?  I am still dreaming of a versioning database, with which some tool may automatically update input files based on its version.

In principle, the commit string also does the trick, but it's not so intuitive, the age of an input file is not easily recognizable from there.

ppraegla:
> In the maintainer meeting on Dec. 12, 2024 the following was decided:
> ### Release schedule
> 
> Releases can provide motivation to bring in developments prior to the next release. They are useful for applications to refer to a specific version of the code.
> 
> Release schedule: 1-2 releases per year
> 
> Versioning scheme: Year + number, e.g. 2025.1 (first release in 2024), 2025.2 (second release in 2025), reset minor release to 1 at the beginning of each year

The versioning scheme is not 100 % fixed yet. From our point of view, it seems reasonable. However, I will check if the proposed versioning is standard and if there might be other pitfalls to consider.

gilrrei:
For #514 to work properly, we need a input file versioning scheme.

Some points that @sebproell and I identified:
- Versioning for input files has nothing to do with the code version. Therefore to avoid confusion, this should be reflected in the versioning scheme (e.g., don't do `v25.2`)
- The input file version has to be increased if the input changed (parameters **and** infrastructure). The easiest way to detect this, is if the metadata file changes in a PR
- A continuous counter should be fine. If major version are wanted, the continuous counting should be done in the minor versions `1.23` (vi for version input, a better name is welcome)
- We should add a metadata section to the input #48 with the possibility to provide the version of the input file:
  - ```YAML
    metadata:
      - input file version: 1.23
    ALE:
      - the rest of the input
    ```
[...]

PR #235, opened 2025-01-10
Introduce CalVer versioning and print it

sebproell:
As discussed in #74


sebproell:
The printout now looks like this:
```
******************************************************
*                                                    *
*                         4C                         *
*                                                    *
*                 version 2025.1-pre                 *
*                                                    *
*                      git SHA1                      *
*      d02460f02321e1b71457be2062d55e52fdb6eb40      *
*                                                    *
******************************************************
```

cmake/setup_install.cmake
```diff
@@ -97,7 +97,7 @@ configure_package_config_file(
 write_basic_package_version_file(
   ${CMAKE_BINARY_DIR}/cmake/templates/4CConfigVersion.cmake
   VERSION ${FOUR_C_VERSION_STRING}
-  COMPATIBILITY AnyNewerVersion
+  COMPATIBILITY ExactVersion
```
sebproell, line 100: We are breaking compatibility all the time, so I changed that. FYI @vryy

mayrmt (dismissed):

VERSION
```diff
@@ -0,0 +1 @@
+2025.1
```
sebproell, line 1: We might want to use this info in Doxygen and ReadTheDocs as well. @ischeider

maxfirmbach (dismissed):

CMakeLists.txt
```diff
@@ -46,9 +46,28 @@ set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
...
-# The version number.
-set(FOUR_C_VERSION_MAJOR 1)
-set(FOUR_C_VERSION_MINOR 0)
+# Read the version from the VERSION file. Instead of encoding the version in the CMakeLists.txt,
+# the file allows easy access to the version from other scripts.
+# Following https://calver.org/, the version string has to look like this:  YYYY.MINOR[.PATCH][-MODIFIER]
+# where PATCH is optional and implicitly 0 if not present.
+file(STRINGS "${PROJECT_SOURCE_DIR}/VERSION" FOUR_C_VERSION_FULL LIMIT_COUNT 1)
+string(
+  REGEX MATCH "([0-9]+)\\.([0-9]+)(\\.([0-9]+))?(-([a-zA-Z0-9]+))?" _dummy ${FOUR_C_VERSION_FULL}
```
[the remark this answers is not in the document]
ppraegla: Why do you use ` \\.` to match a period and not `\.`?

amgebauer:
@sebproell @c-p-schmidt and me had a discussion over lunch and we came to the conclusion that it makes sense to already now include the patch-version, i.e. starting with `2025.1.0`. Reasoning:

 * That allows us to use default tooling (e.g. we can use the docker action https://github.com/docker/metadata-action to automatically detect the version string from the tag name, including the `latest` tag if it's the latest release tag). Writing it ourselves is cumbersome, especially detecting the `latest` tag. We are then compatible with semver and all of its available tools.
 * We are quite sure that we will at some point need a patch for a release. Examples: We realize after release that accidentally some important configurations do not build. In that case, we would need so release a patch. Or, we want to enable an older released version of 4C on a different compiler to reproduce an example from a publication. If someone volunteers to patch the old version, why should we refrain from releasing the patched version if it is helpful for others? It would be weird if we sometimes have 2 and sometimes 3 components in our version. That makes tools that consume us more complicated.

CMakeLists.txt
```diff
@@ -46,9 +46,28 @@ set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
...
-set(FOUR_C_VERSION_MAJOR 1)
-set(FOUR_C_VERSION_MINOR 0)
+# Read the version from the VERSION file. Instead of encoding the version in the CMakeLists.txt,
+# the file allows easy access to the version from other scripts.
+# Following https://calver.org/, the version string has to look like this:  YYYY.MINOR.PATCH[-MODIFIER]
+# where PATCH is optional and implicitly 0 if not present.
+file(STRINGS "${PROJECT_SOURCE_DIR}/VERSION" FOUR_C_VERSION_FULL LIMIT_COUNT 1)
+string(
+  REGEX MATCH
+        "([0-9][0-9][0-9][0-9])\.([0-9]+)\.([0-9]+)(-([a-zA-Z0-9]+))?"
```
amgebauer, line 56:
```suggestion
        "([0-9][0-9][0-9][0-9])\.([0-9]+)\.([0-9]+)(-([a-z0-9]+))?"
```
Let's not allow capital letters. Docker does also not support capital letters. That avoids that we have to use different flags for docker and our version.

amgebauer (approved):

mayrmt (approved):

PR #236, opened 2025-01-10
Automatically tag docker images on release

amgebauer:
This PR adds automatically tags to the docker images `4c` and `4c-dependencies-ubuntu24-04`.

The following tags are added:
 * `latest`
 * `2025.1.0` for the git tag name `v2025.1.0`.
 * branch-name
 * sha-hash (only for `4c`)

Note additionally,

* `latest` will now refer to the latest release
* `main` is now what used to be `latest` before


amgebauer:
We can make us the life a lot easier if we use `2025.1.0` instead `2025.1`. See #235

PR #239, opened 2025-01-10
Emit metadata for InputSpec

sebproell:
This PR allows to call `emit_metadata` for an `InputSpec` which will write all the knowledge we have about that spec into a yaml file. We do not yet use it for the `4C_metadata.yaml` file but this can follow soon if we agree this format looks sensible in general.

Part of #113, Depends on #237 

FYI @gilrrei


sebproell:
@gilrrei Based on our discussion last week: I am now using `anonymous_group` and `group` in code and in the metadata. I hope this makes it clear that the former doesn't have a name.

amgebauer (approved):

lauraengelhardt (approved):

PR #241, opened 2025-01-12
InputFile determines huge sections automatically

### LINKED ISSUE #60: Allow to read the mesh from standard mesh formats
## Description

The mesh should not be part of the input file. Instead, it should be directly read from a (standard) mesh format.

Materials and other element properties could then be specified for blocks instead for each element, e.g. (suppose we have a yaml input file):

```yaml
element_blocks:
    name_of_block:
        MAT: 1
        KINEM: nonlinear
        EAS: none
        ELEMENT: SOLID
```

The mesh-format that we then read must be able to have some requirements:

 * should be a standard format (exodus, gmsh, xdmf, ...)
 * nodesets / sidesets for boundary conditions
 * element and nodal data for fiber directions
 * reasonably fast to read in a mesh and distribute it in parallel. This may either include parallel reading or a better implementation of parallel communication.

- [ ] Decide on a format
- [ ] Check which library to use to read the mesh

* Blocks
* Is blocked by
* Follows
* Precedes
* Related to
* Part of
* Composed of

sebproell:
This PR fixes a problem where huge yaml files containing a huge discretization cannot be read effectively when using many MPI ranks. Input is now read like this, independent of the underlying format:
- Read the complete input file (+ any includes) on rank 0
- Determine huge sections (more than 10,000 entries)
- Communicate everything except the huge sections to all ranks.
- If a huge section is requested on a rank > 0, communicate it then. This means that `lines_in_section` must be a collective call. To support reading on rank 0 only, we now have `lines_in_section_rank_0_only`. Note that this distinction existed before as well but is now made explicit in the function names.

This implementation is much simpler than the on-the-fly reading we had previously. It seems reasonable that the input file must not be so large, that it cannot even fit in memory at once. Note that all of this is temporary anyway until we resolve #60.

As a small test, I checked the total wall time for input (also include creating the discretization etc.) for an example with 1e6 elements (200 MB file size):

- old implementation, dat: around 60s
- old implementation, yaml: around 100s (not relevant because this would have consumed a large amount of memory)
- new implementation, dat: around 30s
- new implementation, yaml: around 50s

Note that yaml is slower because we currently convert the yaml file into a dat-like representation internally. This will change. The message here is not that yaml is inherently slower. I just wanted to ensure that the new implementation is faster than the old one.

### DIFF
--- a/src/core/io/src/4C_io_input_file.cpp
+++ b/src/core/io/src/4C_io_input_file.cpp
@@ -24,6 +24,7 @@
 
 #ifdef FOUR_C_ENABLE_FE_TRAPPING
 #include <cfenv>
+#include <format>
 #endif
 
 
@@ -34,6 +35,12 @@ namespace Core::IO
   namespace
   {
     constexpr double tolerance_n = 1.0e-14;
+
+    /**
+     * Sections that contain at least this number of entries are considered huge and are only
+     * available on rank 0.
+     */
+    constexpr std::size_t huge_section_threshold = 10'000;
 
     /*----------------------------------------------------------------------*/
     /*----------------------------------------------------------------------*/
@@ -55,7 +62,7 @@ namespace Core::IO
     {
       // safety check: Is there a duplicate of the same parameter?
       if (list.isParameter(key))
-        FOUR_C_THROW("Duplicate parameter %s in sublist %s", key.c_str(), list.name().c_str());
+        FOUR_C_THROW("Duplicate parameter '%s' in sublist '%s'", key.c_str(), list.name().c_str());
 
       if (key.empty()) FOUR_C_THROW("Internal error: missing key.", key.c_str());
       // safety check: Is the parameter without any specified value?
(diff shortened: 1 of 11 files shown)


src/core/io/src/4C_io_input_file.cpp
```diff
@@ -343,21 +299,27 @@ namespace Core::IO
...
         }
 
-        content.emplace_back("--" + section_name);
-        section_to_file_mapping[section_name] = file_path.string();
+        FOUR_C_ASSERT_ALWAYS(content_by_section.find(section_name) == content_by_section.end(),
+            "Section '%s' is defined again in file '%s'.", section_name.c_str(), file_path.c_str());
+
+        auto& current_content = content_by_section[section_name];
+        current_content.file = file_path;
+        std::list<std::string> list_of_lines;
```
amgebauer, line 322:
```suggestion
        std::forward_list<std::string> list_of_lines;
```

amgebauer (approved):

src/core/io/src/4C_io_input_file.cpp
sebproell: True. I don't even know if `list` makes sense here; it was what we used even before I started incrementally refactoring this class. Since this is all related to the dat format, I would just leave it.

PR #243, opened 2025-01-13
Fix a false positive gcc warning in particles

sebproell:
Fixes a warning I get when compiling `-O3` w/o assertions with GCC 13.2 installed via Spack (aka LNM release config). At least the code also becomes simpler this way.

Side note: this once more shows that we will need patches for releases ;) It was just a coincidence that I noticed this warning and it would have almost made it into the release.

### DIFF
--- a/src/particle_rigidbody/4C_particle_rigidbody.cpp
+++ b/src/particle_rigidbody/4C_particle_rigidbody.cpp
@@ -1339,8 +1339,8 @@ void ParticleRigidBody::RigidBodyHandler::broadcast_rigid_body_positions()
 
     data.add_to_pack(rigidbody_k);
 
-    for (int i = 0; i < 3; ++i) data.add_to_pack(pos_k[i]);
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (approved):

src/particle_rigidbody/4C_particle_rigidbody.cpp
```diff
@@ -1339,8 +1339,8 @@ void ParticleRigidBody::RigidBodyHandler::broadcast_rigid_body_positions()
 
     data.add_to_pack(rigidbody_k);
 
-    for (int i = 0; i < 3; ++i) data.add_to_pack(pos_k[i]);
```
amgebauer, line 1342: Looks like the `std::vector` should have been an `std::array`

amgebauer (approved):

c-p-schmidt (approved):

PR #249, opened 2025-01-14
Reuse ccache from clang18_build for clang18_install_test

amgebauer:
We only have limited amount of storage available for our cache. Since we build clang18 anyway in the stage before, we can reuse the cache from that stage and don't need our own.

fyi @vryy

### DIFF
--- a/.github/actions/build_4C/action.yml
+++ b/.github/actions/build_4C/action.yml
@@ -15,21 +15,35 @@ inputs:
     description: Use ccache to speed up the build
     required: false
     default: "true"
+  cache-key:
+    description: Key to restore the cache for ccache (default <job-name>-<preset-name>)
+    required: false
+    default: ""
   additional-cmake-flags:
     description: Additional flags to pass to CMake
     required: false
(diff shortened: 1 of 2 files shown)


vryy:
This is interesting, because `ninja install` is invoked during installation testing. I noticed that it sometimes recompiles the whole thing even after copying the build from other workflow. So removing `ccache` will have no influence, but I want to know if we can skip recompiling.

amgebauer:
ok, now I see. It looks like a full recompilation is always necessary (maybe because we checkout the source files again). With ccache activated, it might have happened that recompilation was much faster.

Maybe, instead of deactivating ccache, we need to adapt the ccache key so that the cache from the previous clang18 build stage will be used instead of creating a new one.

amgebauer:
Now ready for review. We now reuse the cache from the previous build stage. As a consequence, copying the build folder from the previous build stage is not necessary.

vryy (approved):

sebproell (approved):

PR #238, opened 2025-01-10
Consistently use separators in input conditions (Part II)

lauraengelhardt:
## Description and Context
This PR adds separators to the conditions defined in src/inpar and adapts the input files accordingly.
See #224 for a detailed explanation.

## Related Issues and Merge Requests
Follows #224 
Part of https://github.com/4C-multiphysics/4C/issues/73

PR #255, opened 2025-01-15
WIP: Remove none_allowed option from LineComponent

lauraengelhardt:
## Description and Context

Remove the `none_allowed` option from `LineComponent`.
Instead of writing `none` in the input file, we now directly use `-1` (which is how an integer `none` was interpreted until now). 

Note for future use: 
- Assuming you want to define an `IntComponent` that can be `none`: Your best option would be to make the `IntComponent` `optional`. 
- Assuming you want to define an `IntVectorComponent`, where any entry can be `none`, e.g. `FUNCT 0 none 0`. You may use `IntVectorComponent` and define a value that represents `none`. Up to now, this was usually `-1`, but this is of course up to you.

Depends/based on #250, consider only the last 2 commits
Related to https://github.com/4C-multiphysics/4C/issues/73


mayrmt:
From a user's perspective, this change not intuitive: "none" has a clear meaning (e.g. I do not refer to a function on purpose), but "-1" requires the implicit knowledge of code internals, that this mean "none". 

To be honest, I am not convinced that this makes the code more user friendly.

mayrmt:
I totally see the value of consistency. 

Yet, usability is also important. And we have tried hard in the past to hide implementation details from the user.

Let's see if we can come up with another solution!

sebproell:
@mayrmt FYI, We can read `none` values in a concise manner once we merge #258. @lauraengelhardt will use this to update this PR.

mayrmt:
Great!

sebproell:
@lauraengelhardt This is outdated and can be closed?

PR #242, opened 2025-01-13
Consistently use separators in MORTAR input conditions (Part III)

lauraengelhardt:
## Description and Context
This PR adds separators to all conditions in `4C_inpar_mortar.cpp` and another coupling condition in `4C_inpar_ehl.cpp`.
See https://github.com/4C-multiphysics/4C/pull/224 for a detailed explanation.

## Related Issues and Merge Requests
Follows https://github.com/4C-multiphysics/4C/pull/224
Follows https://github.com/4C-multiphysics/4C/pull/238
Part of https://github.com/4C-multiphysics/4C/issues/73

PR #256, opened 2025-01-15
Consistently use separators in input conditions (Part IV/IV)

lauraengelhardt:
## Description and Context
This PR unifies the separator name and the actual integer/real/etc. component name for all remaining conditions. 
Now, all components should have the same name for the separator and the actual value.  
This is the last PR of this kind ;) 
See https://github.com/4C-multiphysics/4C/pull/224 for a detailed explanation.

Follows https://github.com/4C-multiphysics/4C/pull/224
Follows https://github.com/4C-multiphysics/4C/pull/242
Part of https://github.com/4C-multiphysics/4C/issues/73

Note: The first two commits are from #255, I can rebase once this is merged.


tests/input_files/elch_1D_DLcap_linearKinetics_galvanostat.dat
```diff
@@ -76,8 +76,8 @@ SCATRA DIS scatra SPECIAL QUANTITY meaneta VALUE -7.50901487373067666e-02 TOLERA
 SCATRA DIS scatra SPECIAL QUANTITY meancur VALUE -1.49999999999984851e+02 TOLERANCE 1e-07
 ----------------------ELECTRODE BOUNDARY KINETICS POINT CONDITIONS
 DPOINT 2
-E 1 - ID 0 POT -0.1 FUNCT none NUMSCAL 1 STOICH -1  E- 1 EPSILON -1 ZERO_CUR 0 linear ALPHA 1 I0 40 GAMMA 0.0 REFCON 0.1 DL_SPEC_CAP 10
-E 2 - ID 1 POT  0.0 FUNCT none NUMSCAL 1 STOICH -1  E- 1 EPSILON -1 ZERO_CUR 0 linear ALPHA 1 I0 40 GAMMA 0.0 REFCON 0.1 DL_SPEC_CAP 0
+E 1 - ConditionID 0 POT -0.1 FUNCT none NUMSCAL 1 STOICH -1  E- 1 EPSILON -1 ZERO_CUR 0 linear ALPHA 1 I0 40 GAMMA 0.0 REFCON 0.1 DL_SPEC_CAP 10
```
c-p-schmidt, line 79:
```suggestion
E 1 - CONDITION_ID 0 POT -0.1 FUNCT none NUMSCAL 1 STOICH -1  E- 1 EPSILON -1 ZERO_CUR 0 linear ALPHA 1 I0 40 GAMMA 0.0 REFCON 0.1 DL_SPEC_CAP 10
```
Is there a reason why the ConditionID is not in SCREAMING_SNAKE_CASE as most other parameters?

lauraengelhardt: The separator was named `ID`, and the component itself `ConditionID`:
```
cond->add_component(std::make_shared<Input::SeparatorComponent>("ID"));
cond->add_component(std::make_shared<Input::IntComponent>("ConditionID"));
```

We need/want to use the same name for both now, so either change all inputs from `ID` to `ConditionID` or change `ConditionID` in the code to `ID`. 
The last option is currently not possible (or not as trivial), because this condition, and several other conditions rely on the rather obscure methods `evaluate_condition` and `evaluate_scalars`. Those methods query the value stored under `ConditionID` from the parameter container. 
https://github.com/4C-multiphysics/4C/blob/cf18f915add5b60ef7753e512896e96287012029/src/core/fem/src/discretization/4C_fem_discretization_evaluate.cpp#L420

Hence, if you want to name the parameter different from `ConditionID` we need to clean up `evaluate_condition` and `evaluate_scalars` (which we should do, but this cannot be done in this MR).

sebproell (approved):

tests/input_files/elch_1D_DLcap_linearKinetics_galvanostat.dat
c-p-schmidt: Thank you for the explanation. We can work towards consistent input afterwards.

c-p-schmidt (approved):

PR #265, opened 2025-01-17
Automatically filter clang-tidy options in CI

amgebauer:
Currently, we have two clang-tidy config files that we need to keep in sync. One for the editors containing all sets of helpful hints, and one for the pipeline containing only those that are enforced. They went out of sync in the last 2 months. To avoid this in the future, We now automatically filter the enforced checks so that we only have to maintain the single clang-tidy config file.


sebproell (commented):
Nice! Please also update this comment: https://github.com/4C-multiphysics/4C/blob/main/.clang-tidy#L63-L64

sebproell (approved):

davidrudlstorfer (approved):

PR #267, opened 2025-01-17
Remove -U from setup virtual environment

### LINKED ISSUE #266: Setup of python environment hangs without internet connection
## Description
I noticed that the setup of the python environment during configuring demands for internet connection. This can be annoying, e.g., when sitting in a train. Simply adding a test case triggers the configure process and one gets blocked, then. 
You may have noticed that even with internet connection it takes a second (or two) to load the python environment. 

Possible remedy:
- [ ] Cache the python environment
- [ ] Reload from cache if no internet connection is available

## Interested Parties
@sebproell @mairehenke @SchaefP1

amgebauer:
@georghammerl could you check whether this already fixes the problem?

Closes #266

### DIFF
--- a/cmake/python/set_up_python.cmake
+++ b/cmake/python/set_up_python.cmake
@@ -71,7 +71,6 @@ _execute_process(
   PROCESS_COMMAND
   ${FOUR_C_PYTHON_VENV_BUILD}/bin/pip
   install
-  -U
   pip==24.3.1
   wheel==0.45.0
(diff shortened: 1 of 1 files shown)


georghammerl (approved):

georghammerl:
Solves #266 
@amgebauer Thank you very much for the immediate help!

sebproell (approved):

PR #269, opened 2025-01-17
Move general so3 element service routines to core

amgebauer:
Routines for interpolation and gauss point extrapolation are moved into core from so3 elements.


amgebauer:
Now ready for review. I additionally had to move the so3 dependency from the tsi modules since that was problematic in the unity build for clang-tidy.

sebproell (approved):

src/so3/4C_so3_element_service.cpp
```diff
@@ -1,94 +0,0 @@
...
-    const Core::LinAlg::Matrix<3, 1>&, const std::vector<double>&);
-template std::vector<double>
-Discret::Elements::project_nodal_quantity_to_xi<Core::FE::CellType::tet10>(
-    const Core::LinAlg::Matrix<3, 1>&, const std::vector<double>&);
-template std::vector<double>
-Discret::Elements::project_nodal_quantity_to_xi<Core::FE::CellType::wedge6>(
-    const Core::LinAlg::Matrix<3, 1>&, const std::vector<double>&);
-template std::vector<double>
-Discret::Elements::project_nodal_quantity_to_xi<Core::FE::CellType::nurbs27>(
-    const Core::LinAlg::Matrix<3, 1>&, const std::vector<double>&);
```
bgoderbauer, line 92: Would explicit instantiations be interesting enough to keep? It may not have much influence on compile time, but for inexperienced readers, it would maybe be nice to know where it is used.

bgoderbauer (approved):

src/so3/4C_so3_element_service.cpp
amgebauer: I decided to move the function to the header (as it is short and does not have many dependencies). The explicit instantiations are, therefore, not needed anymore. This list does not guarantee that these functions are used somewhere. And in all cases (except the deprecated so3-elements), the function is anyway used with a template parameter and not explicitly with the celltype.

PR #271, opened 2025-01-20
Store none_allowed components as Noneables in the InputParameterContainer

lauraengelhardt:
## Description and Context
Currently, components with the option `none_allowed` set to true, are stored as `-1` in the `InputParameterContainer`.
Using the `Noneable` type introduced in #258, we now store such component values as optionals in the container.

This PR adapts the read functionalities of the `LineComponent` accordingly and adapts the getting and checking of the component values in the code base.

Please see the preceding discussion in #255.

Follows https://github.com/4C-multiphysics/4C/pull/258
Related to https://github.com/4C-multiphysics/4C/issues/73

### DIFF
--- a/src/beam3/4C_beam3_euler_bernoulli_evaluate.cpp
+++ b/src/beam3/4C_beam3_euler_bernoulli_evaluate.cpp
@@ -271,17 +271,17 @@ int Discret::Elements::Beam3eb::evaluate_neumann(Teuchos::ParameterList& params,
     time = params.get("total time", -1.0);
 
   // find out whether we will use a time curve and get the factor
-  const auto* tmp_funct = &condition.parameters().get<std::vector<int>>("FUNCT");
+  const auto tmp_funct = condition.parameters().get<std::vector<Core::IO::Noneable<int>>>("FUNCT");
 
   // get values and switches from the condition
   // "ONOFF" is related to the first 6 flags of a line Neumann condition in the input file;
   // value 1 for flag i says that condition is active for i-th degree of freedom
-  const auto* onoff = &condition.parameters().get<std::vector<int>>("ONOFF");
+  const auto onoff = condition.parameters().get<std::vector<int>>("ONOFF");
 
   // val is related to the 6 "VAL" fields after the "ONOFF" flags of the Neumann condition
   // in the input file; "VAL" gives the values of the force as a multiple of the prescribed load
   // curve
-  const auto* val = &condition.parameters().get<std::vector<double>>("VAL");
+  const auto val = condition.parameters().get<std::vector<double>>("VAL");
 
 #ifndef BEAM3EBDISCRETELINENEUMANN
 // funct is related to the 6 "FUNCT" fields after the val field of the Neumann condition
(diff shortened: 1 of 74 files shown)


lauraengelhardt:
The two commits are solely for review purposes - we'll merge once we are done here.

src/beam3/4C_beam3_euler_bernoulli_evaluate.cpp
```diff
@@ -271,7 +271,8 @@ int Discret::Elements::Beam3eb::evaluate_neumann(Teuchos::ParameterList& params,
     time = params.get("total time", -1.0);
 
   // find out whether we will use a time curve and get the factor
-  const auto* tmp_funct = &condition.parameters().get<std::vector<int>>("FUNCT");
+  const auto* tmp_funct =
+      &condition.parameters().get<std::vector<Core::IO::Noneable<int>>>("FUNCT");
```
sebproell, line 275: Since you reworked this anyway, would you mind removing the pointless pointers here? I leave it to you.

src/so3/4C_so3_tet4_evaluate.cpp
```diff
@@ -789,12 +777,15 @@ int Discret::Elements::SoTet4::evaluate_neumann(Teuchos::ParameterList& params,
...
       {
         // function evaluation
-        const int functnum = (funct) ? (*funct)[dim] : -1;
-        const double functfac =
-            (functnum > 0) ? Global::Problem::instance()
-                                 ->function_by_id<Core::Utils::FunctionOfSpaceTime>(functnum - 1)
-                                 .evaluate(xrefegp.data(), time, dim)
-                           : 1.0;
+        double functfac = 1.0;
+        if ((*funct)[dim].has_value() && (*funct)[dim].value() > 0)
```
sebproell, line 781: A possible follow-up could be to remove `-1` in input files and use `none` instead.

lauraengelhardt: We would have to change `0` and `-1` to `none`. I was thinking about this as well but was unsure if we want to allow `0` as well.

sebproell: Since this ID is supposed to refer to one of the functions identified by positive numbers, I would say, we only allow `none` or one of the positive numbers. In any case, I would not change that here, but we could put it on our mental list of possible cleanups.

src/fluid_ele/4C_fluid_ele_boundary_parent_calc.cpp
```diff
@@ -1877,7 +1877,7 @@ void Discret::Elements::FluidBoundaryParent<distype>::evaluate_weak_dbc(
 
   // get values and switches from condition
   // (assumed to be constant on element boundary)
-  const auto functions = wdbc_cond->parameters().get<std::vector<Core::IO::Noneable<int>>>("FUNCT");
+  const auto functions = wdbc_cond->parameters().get<std::vector<int>>("FUNCT");
```
sebproell, line 1880: Should this be changed to `Noneable`?

sebproell (approved):

src/fluid_ele/4C_fluid_ele_boundary_parent_calc.cpp
lauraengelhardt: I had to change this back (see [7b2f175](https://github.com/4C-multiphysics/4C/pull/271/commits/7b2f175e1fa6ffb57c3a3b98fa3d753748527bf7)), because otherwise the test f2_channel20x20_drt_weak.dat fails. It's probably one of those cases where conditions are not retrieved from the input file but set up somewhere along the way...

sebproell:
> The two commits are solely for review purposes - we'll merge once we are done here.

@lauraengelhardt Do you still want to do that?

amgebauer (approved):

lauraengelhardt:
I squashed the commits, but it seems your approvals are kept. Thanks for the review!

PR #277, opened 2025-01-23
InputFile::Fragment to access input file content independent of underlying format

sebproell:
We used to convert every input fragment into a dat-style string, even if it came from yaml. This PR prepares for a change in the internals of the input mechanism, such that we always match our expectation (InputSpec) against the content (InputFile::Fragment) and receive the parameters (InputParameterContainer) without doing unnecessary conversion to a format that is less powerful. We can still read both formats. This PR does not change the timings for a 1e6 element input file for .dat or .yaml.

My overall vision is to have the following two functions which all (parameter) input/output will be going through:

```
// reading input (added in this PR)
InputFile::Fragment::match(InputSpec) -> InputParameterContainer

// wrtiting into an input file
InputFile::append(InputSpec, InputParameterContainer) -> InputFile::Fragment
```

### DIFF
--- a/src/core/io/src/4C_io_input_file.cpp
+++ b/src/core/io/src/4C_io_input_file.cpp
@@ -8,6 +8,8 @@
 #include "4C_io_input_file.hpp"
 
 #include "4C_comm_mpi_utils.hpp"
+#include "4C_io_input_spec.hpp"
+#include "4C_io_value_parser.hpp"
 #include "4C_utils_string.hpp"
 
 #include <ryml.hpp>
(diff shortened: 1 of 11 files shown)


src/core/io/src/4C_io_input_file.cpp
```diff
@@ -290,20 +423,57 @@ namespace Core::IO
...
+        {.base_path = std::filesystem::path(pimpl_->section->file).parent_path()}};
+    InputParameterContainer container;
+    try
+    {
+      spec.fully_parse(parser, container);
+    }
+    catch (const Core::Exception& e)
+    {
+      return std::nullopt;
+    }
```
amgebauer, line 453: The match function is currently only used in a context where they anyways throw an error afterwards. Couldn't you just call this method `parse` and leave the error if it fails? Some of the errors that might occur are checks for implementation errors that are now hidden.

sebproell: Yes, that's correct. I thought about changing the `fully_parse` function to `match` as well. Ideally, it could return a `std::expected<InputParameterContainer, ErrorLog>` (but that's C++23 :cry:) where `ErrorLog` would give you detailed information on which parts exactly did not parse. I"ll do this in one of the next PRs.

For now, I'll remove the `try` block as you suggested.

amgebauer (approved):

src/core/io/src/4C_io_input_file.cpp
amgebauer: Looking at the compiler support of c++23 and at the calendar (it's 2025!) we should just use c++23 instead of inventing an own version of `std::expected`.

davidrudlstorfer (approved):

PR #282, opened 2025-01-24
Enable -Wpedantic warnings

### LINKED ISSUE #65: Raise C++ standard to C++20

sebproell:
This enforces strict compliance with standard C/C++ and warns of non-standard compiler extensions. If the need for these extensions arises, we can still disable the warning, but for now the code compiles in this pedantic mode.

The only things I needed to fix are named `__VA_ARGS__` and a few superfluous semicolons.

Closes #65

### DIFF
--- a/cmake/setup_global_options.cmake
+++ b/cmake/setup_global_options.cmake
@@ -36,6 +36,7 @@ endfunction()
 # Enable all warnings that are supported by the compiler
 enable_compiler_flag_if_supported("-Wall")
 enable_compiler_flag_if_supported("-Wextra")
+enable_compiler_flag_if_supported("-Wpedantic")
 enable_compiler_flag_if_supported("-Wvla")
 
 # Disable unused parameter detection since there would be too many hits to fix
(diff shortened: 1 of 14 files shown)


davidrudlstorfer (dismissed):

sebproell:
I'll close this. Clang and GCC are in disagreement on how to define a specific destructor and there is a change in C++23 which clarifies this, see: https://stackoverflow.com/questions/68751682/is-a-class-templates-name-in-scope-for-a-qualified-out-of-line-destructors-def

`-Wpedantic` is not useful enough to me to bother with this.

PR #283, opened 2025-01-24
Remove hacky HDF5 flags and update to newer API

sebproell:
I noticed these flags which we are for sure not allowed to fiddle with.

### DIFF
--- a/src/core/io/src/4C_io.cpp
+++ b/src/core/io/src/4C_io.cpp
@@ -747,7 +747,8 @@ void Core::IO::DiscretizationWriter::new_step(const int step, const double time)
       write_file = true;
     }
 
-    resultgroup_ = H5Gcreate(resultfile_, groupname.str().c_str(), 0);
+    resultgroup_ =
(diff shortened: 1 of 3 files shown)


davidrudlstorfer (approved):

sebproell:
Oh, the situation seems worse than I thought: https://github.com/4C-multiphysics/4C/actions/runs/12946014412/job/36109666454?pr=282#step:5:849

Apparently we change the HDF5 version string to get a different API :see_no_evil: I haven't checked the docs yet but I doubt that we are allowed to do that

sebproell:
>  but I doubt that we are allowed to do that

I was wrong and we may use the defines to get access to an older API. See https://support.hdfgroup.org/documentation/hdf5/latest/group___h5_g.html#ga187cee27a9fc4f1a311eb19b0522c7b8

The question is if we want to do that. The API we are using is 18 years old, so I doubt there is any HDF5 installation that doesn't support the new API.

sebproell:
I now updated the relevant calls to the newer API.

davidrudlstorfer (approved):

c-p-schmidt (approved):

maxfirmbach:
Looks good!

maxfirmbach (approved):

PR #284, opened 2025-01-24
Use InputSpec in ConditionDefinition

sebproell:
Thanks to all the work by @lauraengelhardt, we can now use `InputSpec` for the conditions :tada:  This change will enable us to fully remove the `LineDefinition` class but I'll do this in a next PR since this one is already quite large.

Part of #73


tests/input_files/elch_1D_electrodes_galvanostat.dat
```diff
@@ -67,8 +67,8 @@ SCATRA DIS scatra NODE 200 QUANTITY phi1 VALUE 1.17933135469771067e+00 TOLERANCE
 SCATRA DIS scatra NODE 200 QUANTITY phi2 VALUE -1.71411948005193533e-03 TOLERANCE 1.8e-11
 -------------------------ELECTRODE DOMAIN KINETICS SURF CONDITIONS
 DSURF  2
-E 1 - ConditionID 0 POT -0.01 FUNCT none NUMSCAL 1 STOICH -1  E- 1 ZERO_CUR 0 Butler-Volmer A_S 1717.5 ALPHA_A 0.5 ALPHA_C 0.5 I0 36.0 GAMMA 0.0 REFCON 1.5 DL_SPEC_CAP 0.0 END
```
sebproell, line 70: @c-p-schmidt I took the liberty to remove the useless `END` instead of adding an unnecessary `tag()` for that.

src/inpar/4C_inpar_s2i.cpp
```diff
@@ -163,352 +164,218 @@ void Inpar::S2I::set_valid_conditions(
...
+                            Inpar::S2I::kinetics_butlervolmerreducedlinearized},
+                    }),
+                entry<int>("NUMSCAL"),
+                entry<std::vector<int>>(
+                    "STOICHIOMETRIES", {.size = from_parameter<int>("NUMSCAL")}),
+                entry<int>("E-"),
+                entry<double>("K_R"),
+                entry<double>("ALPHA_A"),
+                entry<double>("ALPHA_C"),
+                entry<bool>("IS_PSEUDO_CONTACT"),
```
sebproell, line 210: @c-p-schmidt I chose this style to write the spec since it is equivalent to what we had before. The input would become a lot nicer and simpler if you treat the different "KINETIC_MODEL" choices as groups with their own parameters. As discussed I'll leave this a potential future improvement.

c-p-schmidt: Sounds good to me.

c-p-schmidt (approved):

PR #286, opened 2025-01-24
InputSpec: rename anonymous_group -> all_of

sebproell:
In discussion with @c-p-schmidt and @lauraengelhardt we realized that `anonymous_group` is actually `all_of`, the counterpart of `one_of`.


c-p-schmidt (approved):
Thank you for improving the naming and adding the additional check.

amgebauer (approved):

PR #287, opened 2025-01-26
Clean up Condition input

sebproell:
Use `InputSpec` directly in conditions to remove `LineComponent` #73. This PR includes a few steps:
- Sanitize a few input strings in `selection` that are remapped to slightly different strings. This is hard to understand and I think we should deliberately _not_ support that. If you want to remap a `selection` to something different, at least use and enum. `selection` function is enhanced to facilitate this. This is the only small functional change.
- Remove unnecessary shared_ptr on ConditionDefinition and replace old `add_named_XXX` functions with equivalent `InputSpec`.
- Remove the now unused `LineComponent`.

This concludes the major part of #73


sebproell:
Ready now. The only change in functionality is in the first commit.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -295,6 +295,10 @@ namespace Core::IO
    */
   namespace InputSpecBuilders
   {
+    // Import the Noneable type into the InputSpecBuilders namespace to make it easier to use.
+    using Core::IO::none;
+    using Core::IO::Noneable;
```
mayrmt, line 300: We We have been reluctant with unscoped `using` statements in the past. This one is not unscoped, though also not tightly encapsulated and part of a header file, so it might be included somewhere else.

I do not object in this case. Just wanted to build awareness...

sebproell: The whole intention is that you include the header, write `using namespace Core::IO::InputSpecBuilders;` and thus also gain access to the `Noneable` and  `none` symbols. See https://godbolt.org/z/4qYe4T951

lauraengelhardt (approved):
Very nice! Great to see the LineComponent removed :+1:

amgebauer (approved):

Issue #197, opened 2025-01-08
Unify output writers


eulovi:
> This issue was originally created by @mayrmt in https://gitlab.lrz.de/baci/baci/-/issues/140 on 2018-11-19.

## Expectations
We need a common strategy to output simulation results. In particular, there should only be a single user interface for outputting simulation results.

## Current Behavior
We have multiple possibilities to write output. Among them:
- `IO::DiscretizationWriter` and subsequent application of the `post_drt_ensight` filter
- `IO::DiscretizationWriter` and subsequent application of a VTK-based post filter
- VTK-based runtime output (no post-filter needed)

There are two problems common to all of these possibilities:
- They are not compatible to each other.
- None of output approaches can be used in every scenario in Baci.

Example: Solid-to-beam meshtying needs the RuntimeVTK-output for beam visualization, but the VTK+post-filter output for stresses in the solid volume.

## Motivation and Context
Baci allows us to solve complex interactions between multiple fields and discretizations. We also should be able to actually look at the results. This issue is of particular importance in coupled problems where all fields have to use the same output mechanism. In particular, interaction of beams with solids (@isteinbrecher) or beams with fluids (@NHagmeyer) will suffer from this.

More generally speaking, we need a single interface to the output mechanism to guarantee that any future development will be compatible with other fields/problems in Baci. Otherwise, a feature developer has to implement output in both (or even more) output tools (and we all know that this is probably not gonna happen).

## Definition of Done
Ideally
- [ ] Have a common output format and switch all output in Baci to that format

For now:
- [ ] Discuss current inconsistencies and limitations
- [ ] Develop a strategy on how to resolve these.

## Possible Solution
From a developer's perspective, one just wants to pass a quantity to an abstract output tool without being concerned with the actual details of output writing. Hence, we need a common look-and-feel interface for all our output.

If we want to allow for both ensight- or VTK-based output, this distinction should only be done in the input file and in the interior of this common output tool, but not a algorithm level. If done properly, a developer doesn't have to think about the particular output tool.

We could introduce a common base class for writing output. Two classes derive from that, one for writing runtime output, one for writing post-filter-based output. Which one is used, is decided when creating the output writer (maybe using a Factory concept and some control via the input file). Any post-filter (ensight or VTK) is independent of that and has to be executed when necessary.

## Interested Parties
@baci/baci\_maintainer

PR #293, opened 2025-01-27
[thermo] Add temperature runtime output

maxfirmbach:
## Description and Context

This PR adds runtime output for the nodal temperature vector and some basic element and node information to `thermo`. It doesn't replace the old output writing process yet, but marks a initial step for the switch to vtk-runtime output. The current implementation is very similar to the one done in `fluid`.

Vtk runtime output has the nice feature that different quantities can be added easily for output (e.g. material properties as the conductivity), which is not the case for the old one.

I would keep this in `draft` mode for now and might remove the label after the first set of comments and opinions.

#197


src/inpar/4C_inpar_IO_runtime_output_thermo.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include <Teuchos_ParameterList.hpp>
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Inpar
+{
+  namespace IORuntimeOutput
+  {
+    namespace THERMO
```
sebproell, line 20:
```suggestion
    namespace Thermo
```

src/inpar/4C_inpar_IO_runtime_output_thermo.cpp
```diff
@@ -0,0 +1,58 @@
...
+        Teuchos::ParameterList& sublist_IO_output_thermo =
+            sublist_IO_output.sublist("THERMO", false, "");
+
+        // whether to write output for thermo
+        Core::Utils::bool_parameter(
+            "OUTPUT_THERMO", "No", "write thermo output", &sublist_IO_output_thermo);
+
+        // whether to write temperature state
+        Core::Utils::bool_parameter(
+            "TEMPERATURE", "No", "write pressure output", &sublist_IO_output_thermo);
```
sebproell, line 40:
```suggestion
            "TEMPERATURE", "No", "write temperature output", &sublist_IO_output_thermo);
```

src/inpar/4C_inpar_IO_runtime_output_thermo.hpp
```diff
@@ -0,0 +1,34 @@
...
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_INPAR_IO_RUNTIME_OUTPUT_THERMO_HPP
+#define FOUR_C_INPAR_IO_RUNTIME_OUTPUT_THERMO_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_utils_parameter_list.fwd.hpp"
+
+#include <memory>
```
sebproell, line 15:
```suggestion
```

src/thermo/src/4C_thermo_timint.cpp
```diff
@@ -49,6 +53,7 @@ Thermo::TimInt::TimInt(const Teuchos::ParameterList& ioparams,
       solveradaptolbetter_(tdynparams.get<double>("ADAPTCONV_BETTER")),
       dbcmaps_(std::make_shared<Core::LinAlg::MapExtractor>()),
       output_(output),
+      runtime_output_writer_(nullptr),
```
sebproell, line 56:
```suggestion
```

src/thermo/src/utils/4C_thermo_discretization_runtime_output_params.hpp
```diff
@@ -0,0 +1,107 @@
...
+#include "4C_utils_parameter_list.fwd.hpp"
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Discret
+{
+  namespace Elements
+  {
+    /** \brief Input data container for output at runtime for thermo*/
+    class ThermoRuntimeOutputParams
```
sebproell, line 22: You can reduce all this boilerplate down to a simple struct 

```
struct ThermoRuntimeOutputParams
{
        /// whether to write pressure output
      bool output_temperature_state = false;

      /// whether to write the owner of elements
      bool output_element_owner = false;

      /// whether to write the element GIDs
      bool output_element_gid = false;

      /// whether to write the node GIDs
      bool output_node_gid = false;
};
```
which you fill with the correct values instead of calling `init` and `setup`.

You could also move the struct to the thermo_timint header (maybe even into the TimInt class), if you only need it there.

sebproell (changes requested):
Thanks for enabling the modern output for thermo :+1: I only have C++-related requests; the implemented functionality makes sense, although you should test it.

isteinbrecher:
What would be needed to be able to fully switch to runtime output in the thermo module @maxfirmbach ?

maxfirmbach:
> What would be needed to be able to fully switch to runtime output in the thermo module @maxfirmbach ?

There‘s also the option to write the heatflux and temperature gradient as well as the energy values … I haven‘t looked into those (because I‘m not really interested in those currently 😅).

Besides that I think there shouldn‘t be any issues … I haven‘t looked into the whole restart thing too …

mayrmt (changes requested):
@maxfirmbach, thanks for adding this!

I found some more copy-paste oversights where "pressure" needs to be replaced by "temperature".

maxfirmbach:
> @maxfirmbach, thanks for adding this!
> 
> I found some more copy-paste oversights where "pressure" needs to be replaced by "temperature".

That is possible … I will look into it 👍.

maxfirmbach:
> Thanks for enabling the modern output for thermo 👍 I only have C++-related requests; the implemented functionality makes sense, although you should test it.

I will add some kind of vtk output test 👍.

maxfirmbach:
@isteinbrecher I prepared the runtime output to write the missing quantities. This means adding the related infrastructure, but not yet writing the actual states for heatflux, temperature gradient and energy.

maxfirmbach:
@isteinbrecher and me discussed the implementation of the missing output quantities offline and decided to do this in a separate PR at some point.

src/thermo/src/4C_thermo_timint.cpp
```diff
@@ -9,12 +9,16 @@
...
+#include "4C_io_discretization_visualization_writer_mesh.hpp"
+#include "4C_io_visualization_parameters.hpp"
 #include "4C_linalg_serialdensematrix.hpp"
 #include "4C_linalg_serialdensevector.hpp"
 #include "4C_thermo_ele_action.hpp"
 #include "4C_thermo_resulttest.hpp"
 #include "4C_timestepping_mstep.hpp"
 #include "4C_utils_function.hpp"
+#include "4C_utils_parameter_list.hpp"
+#include "4C_utils_shared_ptr_from_ref.hpp"
```
sebproell, line 21: I didn't see where you need this include.

src/thermo/src/4C_thermo_timint.hpp
```diff
@@ -195,6 +197,9 @@ namespace Thermo
     //! print summary after step
     void print_step() override = 0;
 
+    //! assemble data for runtime writer output
```
sebproell, line 200: What do you mean by "assemble" here? I would document what you write and not just repeat the function name.

src/thermo/src/4C_thermo_timint.hpp
```diff
@@ -531,8 +536,36 @@ namespace Thermo
...
+      /// whether to write the element GIDs
+      bool output_element_gid = false;
+
+      /// whether to write the node GIDs
+      bool output_node_gid = false;
+    };
+
     std::shared_ptr<Core::IO::DiscretizationWriter> output_;  //!< binary output
-    bool printlogo_;                                          //!< true: enjoy your cuppa
+    std::shared_ptr<Core::IO::DiscretizationVisualizationWriterMesh>
```
sebproell, line 564: A `std::optional` would probably convey more clearly that this member is not always needed.

sebproell (approved):
A few optional ideas for improvement.

maxfirmbach:
@sebproell I added the `std::optional`, not sure if I use it in a correct way though ...

src/thermo/src/4C_thermo_timint.cpp
```diff
@@ -379,11 +377,11 @@ void Thermo::TimInt::read_restart_state()
  *----------------------------------------------------------------------*/
 void Thermo::TimInt::write_runtime_output()
 {
-  runtime_output_writer_->reset();
+  runtime_output_writer_.value()->reset();
```
sebproell, line 380:
```suggestion
  runtime_output_writer_->reset();
```
If you unconditionally access the value, just write this.

Also, I should add something like `FOUR_C_ASSERT(runtime_output_writer_.has_value(), "Runtime output not initialized.")`. If the function is public, it should probably even be `FOUR_C_ASSERT_ALWAYS`.

src/thermo/src/4C_thermo_timint.hpp
```diff
@@ -561,7 +561,7 @@ namespace Thermo
     };
 
     std::shared_ptr<Core::IO::DiscretizationWriter> output_;  //!< binary output
-    std::shared_ptr<Core::IO::DiscretizationVisualizationWriterMesh>
+    std::optional<std::shared_ptr<Core::IO::DiscretizationVisualizationWriterMesh>>
```
sebproell, line 564:
```suggestion
    std::optional<Core::IO::DiscretizationVisualizationWriterMesh>
```

Sorry if that was not clear before. This is what I meant.

maxfirmbach: Done.

src/thermo/src/4C_thermo_timint.hpp
maxfirmbach: I agree, makes it more straightforward to understand.

src/thermo/src/4C_thermo_timint.hpp
maxfirmbach: Changed the documentation.

src/thermo/src/4C_thermo_timint.cpp
```diff
@@ -377,11 +377,13 @@ void Thermo::TimInt::read_restart_state()
...
 {
-  runtime_output_writer_.value()->reset();
+  FOUR_C_ASSERT_ALWAYS(runtime_output_writer_.has_value(), "Runtime output not initialized.");
+
+  runtime_output_writer_.value().reset();
 
   if (runtime_output_params_.output_temperature_state)
   {
-    runtime_output_writer_.value()->append_result_data_vector_with_context(
+    runtime_output_writer_.value().append_result_data_vector_with_context(
```
sebproell, line 386:
```suggestion
    runtime_output_writer_->append_result_data_vector_with_context(
```

You can just access the optional like that as you asserted that it is non-empty.

sebproell (approved):

src/thermo/src/4C_thermo_timint.cpp
maxfirmbach: Right, that works now :+1:.

maxfirmbach:
I wait for another approval and squash things together before merging.

mayrmt (approved):

isteinbrecher (approved):

PR #296, opened 2025-01-28
InputFile can read arbitrarily nested yaml objects

sebproell:
An important step towards enabling useful yaml input: we can now process arbitrarily structured yaml and provide access to it via the `InputFile` class. All interpretation of the data still goes through a dat-style string, but this could now more easily be changed in the future.


src/core/io/src/4C_io_input_file.cpp
```diff
@@ -433,13 +478,36 @@ namespace Core::IO
...
+        std::ostringstream ss;
+        ss << node;
+        // replace all the yaml markup characters with spaces to receive a dat-style string
+        auto& str = pimpl_->section->as_yaml().dat_style_string;
+        str = ss.str();
+        std::replace(str.begin(), str.end(), '\n', ' ');
+        std::replace(str.begin(), str.end(), ':', ' ');
+        std::replace(str.begin(), str.end(), ',', ' ');
+        std::replace(str.begin(), str.end(), '[', ' ');
+        std::replace(str.begin(), str.end(), ']', ' ');
```
amgebauer, line 498:
```suggestion
        std::replace(str.begin(), str.end(), '[', ' ');
        std::replace(str.begin(), str.end(), ']', ' ');
        std::replace(str.begin(), str.end(), '{', ' ');
        std::replace(str.begin(), str.end(), '}', ' ');
```
Don't we also need these?

tests/input_files/solid_ele_hex8_Standard_eas_none_stressout.4C.yaml
```diff
@@ -35,12 +35,21 @@ STRUCTURAL DYNAMIC:
...
+      MATIDS: [10]
+      DENS: 0.1
+  - MAT: 10
+    ELAST_CoupNeoHooke:
+      YOUNG: 10
+      NUE: 0.25
 FUNCT1:
-  - COMPONENT 0 SYMBOLIC_FUNCTION_OF_SPACE_TIME t
+  - COMPONENT: 0
+    SYMBOLIC_FUNCTION_OF_SPACE_TIME: t
```
amgebauer, line 49: A struct would be nicer here instead of a list, but this is probably not so easy to realize.

sebproell: This would require more refactoring in the input mechanism of functions. We currently read more or less isolated "lines" (in dat file lingo) and build a function from these. Every line holds a component. I think the natural way to map lines to yaml is an array.

src/core/io/src/4C_io_input_file.cpp
amgebauer: Once we directly read yaml without converting it to dat internally, testing json explicitly should not be necessary. So not so important for this intermediate step.

c-p-schmidt (approved):

PR #299, opened 2025-01-28
Provide more input metadata for schema generation

sebproell:
Part of #113 

Includes a few of the more complicated input file sections. The remaining ones need a little bit of polishing and simplification to be as easy to add.

Draft as I will iterate over this state with @gilrrei to see which (more or less arbitrary) structure makes the most sense for our 4C_metadata.yaml.

### DIFF
--- a/apps/global_full/4C_global_full_main.cpp
+++ b/apps/global_full/4C_global_full_main.cpp
@@ -269,8 +269,7 @@ int main(int argc, char* argv[])
   {
     if (Core::Communication::my_mpi_rank(lcomm) == 0)
     {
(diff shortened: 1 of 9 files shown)


sebproell:
The current state is an improvement in the metadata we write out and already includes quite a lot of feedback by @gilrrei (thanks :pray:). Other work on the yaml side I have lying around will conflict with this change, so I would like to merge soon. We can continue tweaking this in follow-up PRs when we learn more about what is required for the schema generation.

amgebauer (approved):

davidrudlstorfer (approved):

PR #301, opened 2025-01-29
Remove input tag() in functions

sebproell:
As described in #73, the `tag()` function is rather questionable as we nowadays have better ways to describe input, e.g. `group()`, `one_of` and `selection`. This PR removes the tags from the functions. Note that I favored backward-compatibility and only performed restructuring that does not break existing files.

### DIFF
--- a/src/core/utils/src/functions/4C_utils_function_manager.cpp
+++ b/src/core/utils/src/functions/4C_utils_function_manager.cpp
@@ -69,6 +69,18 @@ void Core::Utils::add_valid_builtin_functions(Core::Utils::FunctionManager& func
 {
   using namespace IO::InputSpecBuilders;
 
+  auto time_info = all_of({
+      entry<int>("NUMPOINTS"),
+      one_of({
+          group("BYNUM",
+              {
+                  entry<std::vector<double>>("TIMERANGE", {.size = 2}),
+              },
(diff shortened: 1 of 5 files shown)


davidrudlstorfer (dismissed):

src/core/utils/src/functions/4C_utils_function_manager.cpp
```diff
@@ -80,33 +92,35 @@ void Core::Utils::add_valid_builtin_functions(Core::Utils::FunctionManager& func
...
-                                   {
-                                     // When NUMPOINTS is not set, then we still allow for a
-                                     // single DESCRIPTION entry
-                                     return 1;
-                                   }
-                                 }}),
-          tag("PERIODIC", {.default_value = false}),
-          entry<double>("T1", {.required = false}),
-          entry<double>("T2", {.required = false}),
+          one_of({
```
sebproell, line 95: This example nicely demonstrates how much more expressive the new InputSpec can be. All the implicit dependencies between the parameters are now made explicit. Instead of having a bunch of optional parameters, we now nicely group the parameters that should be entered together.

Best thing: with the work of @gilrrei this will soon be available through your editor if you use yaml files.

amgebauer (approved):

c-p-schmidt (approved):

PR #311, opened 2025-01-31
[doc] Describe types of maps from Petra Object Model

### LINKED ISSUE #309: Document row/col/range/domain map
## Description

Parallelization in 4C is based on Trilinos data structures and, thus, has adopted Trilinos nomenclature such as row map, column map, range map, and domain map. To ease their understanding for new developers, we should document this in our developer guide -- either by pointing to a good Trilinos resource or writing our own documentation.

- [x] Check for available explanations of these maps
- [x] Identify suitable location in our developer guide
- [x] Put links or documentation in our developer guide

## Interested Parties

@bgoderbauer (I opened this issue due to your comment in #304, thanks for that!)

mayrmt:
## Description and Context

In the developer guide, explain the different types of maps (row, column, domain, range) from the Petra Object Model and exemplify them for linear algebra operations as well as FEM meshes.

Closes #309

### DIFF
--- a/doc/readthedocs/petra_object_model.rst
+++ b/doc/readthedocs/petra_object_model.rst
@@ -1,0 +1,93 @@
+.. _petra-object-model:
+
+Distributed computations with the Petra Object Model
+====================================================
+
+|FOURC| is heavily based on the Trilinos project.
+For this reason, management of data distributed across a parallel machine
+follows Trilinos' **Petra Object Model**.
+
+Petra Object Model
+------------------
+
+The Petra Object Model is a foundational framework within the Trilinos project,
+designed to facilitate the construction and manipulation of distributed memory objects in parallel computing environments.
+It provides a structured approach to managing data distribution and communication across multiple MPI processes,
+ensuring efficient parallel computations.
+
+It uses the concept of a **map** for the
+
+- **Distribution of global objects**: A map defines how a global object, such as a vector or matrix, is partitioned across the local memories of different MPI processes.
+- **Global-to-Local Index Translation**: A map manages the translation between global indices (representing the entire dataset) and local indices (specific to a processes' subset), facilitating efficient data access and manipulation.
+
+Types of Maps -- a linear algebra point of view
+-----------------------------------------------
+
+The Petra Object Model uses four maps: row map, column map, domain map, range map
+
+Let's assume the matrix-vector product
+
+.. math::
+  y = Ax
+
+with vectors :math:`x` and :math:`y` as well as the matrix :math:`A` being stored in a distributed memory environment.
+
+- **Row Map**: The row map defines how the rows of a distributed matrix or the entries of a distributed vector are partitioned across multiple processes. It specifies which rows are owned by each process, facilitating efficient parallel operations. For instance, in a sparse matrix, the row map determines the distribution of matrix rows among processes.
+- **Column Map**: The column map pertains to the columns of a distributed matrix. It indicates which columns are needed by each process to perform local computations. This is particularly important in sparse matrix-vector multiplications, where a process may require access to certain columns that are owned by other processes. The column map ensures that each process knows which columns to access and possibly import from other processes.
+- **Domain Map**: The domain map defines the distribution of the input vector space for an operator or matrix. In the context of the exemplary matrix-vector product, the domain map describes how the elements of the vector :math:`x` are distributed across processes. This ensures that each process knows which portions of the input vector it is responsible for during the computation.
+- **Range Map**: Conversely, the range map defines the distribution of the output vector space for an operator or matrix. In the context of the exemplary matrix-vector product, the range map describes how the elements of the resulting vector :math:`y`  are distributed across processes. This ensures that each process knows which portions of the output vector it is responsible for after the computation.
+
+.. admonition:: Example
+
+  Assume the matrix-vector multiplication :math:`y=Ax` to read
+
+  .. math::
+    \begin{bmatrix} y_1\\y_2\\y_3 \end{bmatrix}
+    = \begin{bmatrix} 2 & -1 & 0\\-1 & 2 & -1\\0 & -1 & 2 \end{bmatrix}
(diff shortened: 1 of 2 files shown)


mayrmt:
Starting on slide 20 of [this presentation](https://trilinos.github.io/pdfs/4-UsingTrilinosLinearSolvers.pdf), there are some examples of maps in the context of matrices/linear operators. If desired, we could replicate parts of them in our documentation.

doc/readthedocs/petra_object_model.rst
```diff
@@ -0,0 +1,46 @@
+.. _petra-object-model:
+
+Distributed computation with the Petra Object Model
+===================================================
+
+|FOURC| is hevily based on the Trlinos project.
```
davidrudlstorfer, line 6:
```suggestion
|FOURC| is heavily based on the Trilinos project.
```

davidrudlstorfer (commented):
@mayrmt thanks for adding this - although I'm now approximately one year into 4C I did not understand the concepts of the different maps until reading the proposed changes - this helps a lot.

doc/readthedocs/petra_object_model.rst
sebproell: I noticed that there are quite a few of these typos. Why does our typos hook not catch them? :thinking:

sebproell: I had a look and the typos hook works. Apparently `heavily` is not a common misspelling, so they haven't included that inside typos.

mayrmt: Fixed it.

mayrmt:
@davidrudlstorfer I added a concrete example of the maps for a simple 3x3 matrix. Let me know if this is helpful. (This example is taken from the Trilinos slides referenced above. There are other variants in there, but I'd keep it simple for now.)

mayrmt:
I have added some explanations and a concrete example. For now, I will leave it at this. 

This PR is ready for review.

/cc @bgoderbauer

doc/readthedocs/petra_object_model.rst
```diff
@@ -0,0 +1,91 @@
...
+
+  - Rows are fully owned by a single processor.
+  - Row map = Domain map = Range map (all maps are 1-to-1)
+  - Column map is not 1-to-1
+
+Types of Maps -- a finite element point of view
+-----------------------------------------------
+
+While domain map and range map are strongly tied to matrices,
+the concept of row and column maps transfers directly to finite element discretizations using an overlapping domain decomposition to distribute the mesh across a parllel computer.
```
sebproell, line 88: Is the fact that the domain decomposition is overlapping relevant here? I am not 100% sure but with "overlapping domain decomposition" you seem to be referring to the fact that we like to evaluate elements on multiple ranks and do on-processor assembly into the dofs owned by the relevant ranks (so the "overlapping" part are the elements that are evaluated multiple times, right?). If we do off-processor assembly (only evaluating every element once), we still need the concept of row and column maps, don't we? An element at the interface between subdomains might need "column dofs" from other ranks for evaluation and the associated matrix rows will need the see columns as well for multiplication. For me, "overlapping domain decomposition" is associated more with linear solvers than with the MPI layout (see https://en.wikipedia.org/wiki/Domain_decomposition_methods), although they are related in the sense that you probably need a decomposed MPI layout for the linear solve technique to make any sense.

Long story short, I think we should either explain what we mean by overlapping domain decomposition or leave it out since everything you write seems to be true without this condition.

sebproell, line 88:
```suggestion
the concept of row and column maps transfers directly to finite element discretizations using an overlapping domain decomposition to distribute the mesh across a parallel computer.
```

doc/readthedocs/petra_object_model.rst
```diff
@@ -0,0 +1,91 @@
...
+------------------
+
+The Petra Object Model is a foundational framework within the Trilinos project,
+designed to facilitate the construction and manipulation of distributed memory objects in parallel computing environments.
+It provides a structured approach to managing data distribution and communication across multiple processors,
+ensuring efficient parallel computations.
+
+It uses the concept of a **map** for the
+
+- **Distribution of global objects**: A map defines how a global object, such as a vector or matrix, is partitioned across the local memories of different processors (MPI ranks).
```
sebproell, line 20: I know that we (myself included) frequently use these terms interchangeably:

- "(MPI) rank": an integer [0,N) where N is the associated MPI communicator size
- "(MPI) process": the actual OS process id
- "processor": Not sure, but sounds like the whole CPU (core) to me? So more hardware than software.

WDYT about prefacing this section with a short note on MPI and what the terms mean? Also, I recently isolated all or most of the explicit MPI calls. I think it would be helpful for developers to explain how Petra handles MPI communication for them and explain when they need explicit MPI communication themselves. I am volunteering to do it as a follow-up if there is interest in having that. But let's watch out not to replicate trivially searchable things ;) 

Would you mind replacing "processor" with one of the other two, depending on the context?

doc/readthedocs/petra_object_model.rst
mayrmt: @sebproell, that is a good point! Indeed, I was referring to the case of evaluating (column) elements twice and only assembling into "row dofs". Your example with off-processor assembly is correct to. It naturally arises from overlapping DD, but does not fully fit into the current description.

Yes, DD methods can be used as solvers, yet their sole usage for distributing data across processes is valid on its own in my point of view. Nevertheless, I agree that the fact of "overlapping domain decomposition" does not really matter here. I will remove its mentioning.

doc/readthedocs/petra_object_model.rst
mayrmt: I usually prefer "process / processes" over "processor / processors" to avoid confusion with hardware. Somehow, this made it into the draft anyways. 

I will go for "MPI process".

sebproell (approved):
Nice! I would squash the commits.

davidrudlstorfer (approved):

PR #316, opened 2025-02-03
Allow to read real yaml as input

sebproell:
This PR finally enables us to read "real" yaml without first converting it into the dat-style string. Crucially, this allows us to read vectors of arbitrary lengths without an additional `NUM<xxx>` parameter, which is also tested in this PR. Furthermore, I made sure that the material input uses the `InputFile::Fragment::match` method which selects between dat style parsing and yaml matching depending on what we got from the input file. The `solid_ele_hex8_Standard_eas_none_stressout.4C.yaml` is a test for the yaml case.


src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -1069,6 +1195,10 @@ Core::IO::InputSpec Core::IO::InputSpecBuilders::user_defined(std::string name,
       Internal::UserDefinedSpec<DataType>{.name = name,
           .data = std::forward<DataType>(data),
           .parse = parse,
+          .match =
+              [](ConstYamlNodeRef, InputParameterContainer&, IO::Internal::Matches&) {  // TODO
```
amgebauer, line 1199: ?

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -508,11 +506,32 @@ namespace Core::IO
...
+      // We need to check whether the node contains structured yaml or just a dat-style string.
+      if (node.is_val())
+      {
+        std::string_view dat_style_string{node.val().data(), node.val().size()};
+        Core::IO::ValueParser parser{dat_style_string,
+            {.base_path = std::filesystem::path(pimpl_->section->file).parent_path()}};
+        InputParameterContainer container;
+        spec.fully_parse(parser, container);
+        return container;
+      }
```
amgebauer, line 529: It would be nice to allow this only for certain sections (those that will disappear anyway once we can read the mesh from standard formats)

src/core/io/src/4C_io_input_spec_builders.hpp
sebproell: Thanks! I think we should probably get rid of the `user_defined` entry. Writing one is probably more difficult than enhancing the `InputSpecBuilders` library. And apart from the special case of @lauraengelhardt, we have never needed this (and there just for reasons of the material design).

src/core/io/src/4C_io_input_file.cpp
sebproell: This could make sense. To do this, I first need to let the InputFile know which sections we know. Right now, this knowledge is highly implicit.

src/core/io/src/4C_io_input_spec_builders.hpp
sebproell: I'll just throw an error in this case for now. I will discuss with @lauraengelhardt if we can remove `user_defined`.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -126,6 +126,40 @@ namespace Core::IO
       return PrettyTypeName<T>{}();
     }
 
+    /**
+     * A small wrapper around a vector that provides a few convenience functions related to storing
+     * matched node IDs.
+     */
+    class Matches
```
sebproell, line 133: I implemented much better tracing of the match process with nice error messages. Since this PR contains enough non-trivial code already, I would push that in a follow-up. The whole `Matches` class will change to something much more powerful but for the initial version, it gets the job done.

davidrudlstorfer (approved):

amgebauer (approved):

PR #328, opened 2025-02-06
Fix print_unknown_section

davidrudlstorfer:
## Description and Context

Fix a bug in the input file parsing mechanism where an unknown section was not recorded properly. If the unknown section is not known it was not correctly set.

For example, `PROBLEM TYP` was not recognized and the default values were used.

### DIFF
--- a/src/core/io/src/4C_io_input_file.cpp
+++ b/src/core/io/src/4C_io_input_file.cpp
@@ -687,7 +687,10 @@ namespace Core::IO
     }
 
     // the following section names are always regarded as valid
-    record_section_used("TITLE");
+    // NOTE: this is a temporary solution to avoid unused section warnings for restarts.
+    //       Once the input file knows the expected sections, this can be removed.
+    record_section_used("ALE ELEMENTS");
+    record_section_used("FLUID ELEMENTS");
     record_section_used("FUNCT1");
     record_section_used("FUNCT2");
     record_section_used("FUNCT3");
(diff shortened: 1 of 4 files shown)


amgebauer:
@gilrrei that was the problem that we also observed

davidrudlstorfer:
Kudos to @sebproell for determining and fixing the problem :pray:

amgebauer:
It's also time that we get rid of the ParameterList. Specifying the problem type should be mandatory

amgebauer (approved):

gilrrei:
I'm not sure how/if it is related, but there are test files with function numbers FUNCTX [(here for example)](https://github.com/4C-multiphysics/4C/blob/0aab69bd480c777ff970dad4d51447ea8a6d5674/tests/input_files/poromultielastscatra_2D_quad4_tumor_angiogenesis.dat#L228) larger than 20, whereas [here](https://github.com/4C-multiphysics/4C/blob/0aab69bd480c777ff970dad4d51447ea8a6d5674/src/core/io/src/4C_io_input_file.cpp#L710), only 20 are set to used

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -708,6 +711,10 @@ namespace Core::IO
     record_section_used("FUNCT18");
     record_section_used("FUNCT19");
     record_section_used("FUNCT20");
+    record_section_used("NODE COORDS");
```
amgebauer, line 714: Instead of reverting this change, can we remove these `FUNCT*` here so that we can only define a function if it is actually used in the codebase? If I have observed the error correctly before, we just need to remove the unused and empty sections from a few header files in the tutorial tests.

It is kind of weird, that we ignore unused `FUNCT1` ... `FUNCT20`, but would not ignore `FUNCT21`.

amgebauer (commented):
Instread

src/core/io/src/4C_io_input_file.cpp
davidrudlstorfer: If I understood @sebproell correctly this will be entirely removed in the upcoming days - therefore, we wanted to keep the changes small.

amgebauer (approved):

sebproell (approved):

PR #330, opened 2025-02-07
InputSpec: clearly state supported types for input

sebproell:
Also fix a small bug where selection prints the wrong value for enum classes.

### DIFF
--- a/src/core/io/tests/4C_io_input_spec_test.cpp
+++ b/src/core/io/tests/4C_io_input_spec_test.cpp
@@ -72,6 +72,33 @@ namespace
     }
   }
 
+  TEST(InputSpecTest, EnumClassSelection)
+  {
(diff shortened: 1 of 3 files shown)


src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -637,10 +664,10 @@ required: true
   - name: e
     type: selection
     required: false
-    default: 1
+    default: e1
     choices:
-      e1: 1
-      e2: 2
+      - name: e1
+      - name: e2
```
sebproell, line 670: FYI @gilrrei: `default` is now correctly printed as the value you need to enter and I removed the internal values from the choices.

amgebauer (approved):

PR #331, opened 2025-02-07
Create some basic utility functions to integrate over a celltype in any dimension

amgebauer:
This PR adds some utility functions for integrating over any celltype in core.

These functions are similar to that what we also have in the new solid elements. I just generalized them a bit, so that they are not connected to solids, and can also handle manifolds (e.g., a surface in the 3d space).

I will use this in another PR to extract the spring-dashpots from so3_surface.

I will also reuse this in the new solid elements, but I want to keep the changes in here small.


sebproell (dismissed):

sebproell (approved):
The new test is easier to read and understand :+1:

davidrudlstorfer (approved):

Issue #303, opened 2025-01-30
Implement Solid-Poro Element with new templated solid elements


amgebauer:
## Description

This issue is to document progress of replacing `so3_poro*` and `wall1_poro*` with a respective solid-poro element that uses the new templated solid element implementation.

The pressure-based formulation has been implemented with !1570. Still missing is the pressure-velocity based formulation.

The pressure velocity based formulation also needs to implement `GetCauchyNDirAndDerivativesAtXi(xi, displacements, pressure, concentration, n, dir)` see https://gitlab.lrz.de/baci/baci/-/issues/1162

## Possible Solution and Definition of Done

- [x] Implement pressure based formulation https://gitlab.lrz.de/baci/baci/-/merge_requests/1570
- [x] Implement pressure veloctiy based formulation for `SOLID*PORO` elements
- [ ] Implement pressure veloctiy based formulation for `SOLID*POROP1*` elements
- [x] Implement pressure veloctiy based formulation for `SOLID*POROSACTRA` elements #392 
- [x] Implement `GetCauchyNDirAndDerivativesAtXi(xi, displacements, pressure, concentration, n, dir)` for Nitsche contact with pressure-velocity based formulation #335

## Related Issues and Merge Requests

* Part of

amgebauer:
Who is involved in poro? Can we setup a meeting to discuss the formulations that are not yet migrated and how we can do so?

Proposed dates:
- :one: 5.2. 11:00
- :two: 6.2. 10:00
- :three: 7.2. 13:00
- :four: 12.2. 13:00
- :five: 13.2. 13:00

Mentioning @SchaefP1 @lkoeglmeier @knarfnitram @vryy

amgebauer:
Thank you for taking part at the poll. Then let's fix the 7.2 at 13:00. See you then. I will send the zoom link via Slack.

PR #335, opened 2025-02-07
Enable Nitsche contact in new pressure-velocity-based solid-poro element

amgebauer:
It was easier than initially expected. Now, we can use Nitsche contact in new pressure-velocity-based solid-poro elements.

Part of #303

@lkoeglmeier @SchaefP1 @knarfnitram


tests/input_files/new_solidporo_pressure_velocity_based_nitsche_contact3D_twoelem.dat
```diff
@@ -0,0 +1,274 @@
...
+ITEMIN                          1
+RESULTSEVERY                           1
+TOLRES_GLOBAL                   1e-8
+TOLINC_GLOBAL                   1e-8
+NORM_INC                        AbsGlobal
+NORM_RESF                       AbsGlobal
+NORMCOMBI_RESFINC               And
+SECONDORDER                     No
+CONTIPARTINT                    Yes
+CONTACT_NO_PENETRATION          Yes
```
SchaefP1, line 116: Thanks for implementing this! The code looks good to me. One remark to the test: This parameter `CONTACT_NO_PENETRATION` is critical for how poro contact behaves and is relevant here by - if I remember correctly - deciding if `if (linearizations.d_cauchyndir_dp || linearizations.solid.d_cauchyndir_dn || linearizations.solid.d_cauchyndir_ddir || linearizations.solid.d_cauchyndir_dxi)` is true or false (in the case of this test it should be true). So long term there should also be a test with `CONTACT_NO_PENETRATION no` like for the so3 element. But since the two implementations do basically the same I'd be ok with waiting with this until the so3 element gets removed to avoid almost duplicate tests - up to you depending on what you prefer.

SchaefP1 (dismissed):

tests/input_files/new_solidporo_pressure_velocity_based_nitsche_contact3D_twoelem.dat
amgebauer: good point, I have actually not tested any other test, so it might be good to also add one with `CONTACT_NO_PENETRATION no`. Let me do so.

amgebauer: I had to make some additional changes in the code to make it work. Good, that you raised the point.

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_pressure_velocity_based_evaluate.cpp
```diff
@@ -231,6 +234,71 @@ int Discret::Elements::SolidPoroPressureVelocityBased::evaluate(Teuchos::Paramet
...
+      {
+        return solid->get_normal_cauchy_stress_at_xi(
+            *this, this->struct_poro_material(), disp, xi, n, dir, linearizations.solid);
+      },
+      solid_calc_variant_);
+
+  if (!pressures) return cauchy_stress_n_dir;
+
+  const double n_dot_dir = n.dot(dir);
+  if (std::abs(n_dot_dir) > 1e-30)
```
sebproell, line 255: What's the significance of 1e-30 here?

sebproell (approved):

c-p-schmidt (approved):

SchaefP1 (approved):

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_pressure_velocity_based_evaluate.cpp
amgebauer: I just copied it from the previous implementation. I discussed this with @c-p-schmidt @SchaefP1, and we are both unsure. We suspect that this has been added to avoid some unnecessary computations for the common case that `n` and `dir` are perpendicular. However, skipping the whole part results in an error in the linearization (w.r.t. `n` or `dir`). It might be the case that this error in the linearization does not matter down the road in solid-poro Nitsche contact, but still, the solid-poro elemnts should return a consistent result. The additional computations (basically a few additions and multiplications) should be negligible to that what needs to be done for the solid element, so I think we can remove this check here.

c-p-schmidt (approved):

SchaefP1 (approved):

PR #338, opened 2025-02-07
Minor cleanup new structure vtk output

c-p-schmidt:
## Description and Context
Manual export to column map is unnecessary and thus removed. Furthermore, some minor beauty changes in related files.

### DIFF
--- a/src/structure_new/src/output/4C_structure_new_discretization_runtime_output_params.cpp
+++ b/src/structure_new/src/output/4C_structure_new_discretization_runtime_output_params.cpp
@@ -23,6 +23,7 @@ Discret::Elements::StructureRuntimeOutputParams::StructureRuntimeOutputParams()
       output_velocity_state_(false),
       output_element_owner_(false),
       output_element_gid_(false),
+      output_element_material_id_(false),
       output_element_ghosting_(false),
       output_node_gid_(false),
       output_stress_strain_(false),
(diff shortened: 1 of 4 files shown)


src/structure_new/src/output/4C_structure_new_discretization_runtime_output_params.cpp
```diff
@@ -61,9 +62,11 @@ void Discret::Elements::StructureRuntimeOutputParams::init(
     auto io_strain =
         Teuchos::getIntegralValue<Inpar::Solid::StrainType>(io_parameter_list, "STRUCT_STRAIN");
     if (io_stress == Inpar::Solid::stress_none and io_strain == Inpar::Solid::strain_none)
+    {
       FOUR_C_THROW(
           "If stress / strain runtime output is required, one or two of the flags STRUCT_STRAIN / "
           "STRUCT_STRESS in the --IO section has to be set.");
+    }
```
mayrmt, line 69: Technically, these flags are always set (but maybe not activated), since we work with default parameters for values, that the user did not specify in the input file:
```suggestion
    {
      FOUR_C_THROW(
          "If stress / strain runtime output is required, one or two of the flags STRUCT_STRAIN / "
          "STRUCT_STRESS in the --IO section has to be activated.");
    }
```

mayrmt (dismissed):
@c-p-schmidt, looks good!

amgebauer (dismissed):

sebproell (dismissed):

sebproell (approved):

mayrmt (approved):

PR #343, opened 2025-02-09
Read key-value input parameters via InputSpec

sebproell:
This PR migrates all input done via Teuchos::ParamterList to the new `InputSpec.` This gives us a lot saner input for the key-value sections, notably:
- actually require parameters to be set without providing useless defaults. best example: PROBLEMTYPE defaults to FSI
- allow vector input, e.g. some beam parameters used string input and manually parse out double values from that.
- grouping of parameters that should be required together, e.g. selecting GenAlpha requires you to choose the respective coefficients.

I have not performed any of these steps yet, as the goal of this PR is to search-replace the implementation with as little effort as possible to ensure that we can handle all existing input.

This is a major step in improving the input. When I started this work, we had _five_ different ways to specify parts of the input file and read it. After this PR, there is finally a single one, although more cleanup will follow.


src/core/linear_solver/src/amgnxn/4C_linear_solver_amgnxn_preconditioner.cpp
```diff
@@ -179,22 +181,15 @@ Core::LinearSolver::AmGnxnInterface::AmGnxnInterface(Teuchos::ParameterList& par
   if (amgnxn_type == "XML")
   {
     // Parse the whole file
-    std::string amgnxn_xml = amglist.get<std::string>("AMGNXN_XML_FILE", "none");
-    if (amgnxn_xml == "none") FOUR_C_THROW("The input parameter AMGNXN_XML_FILE is empty.");
-    if (not(amgnxn_xml == "none"))
+    auto amgnxn_xml = amglist.get<Core::IO::Noneable<std::filesystem::path>>("AMGNXN_XML_FILE");
+    if (!amgnxn_xml) FOUR_C_THROW("The input parameter AMGNXN_XML_FILE is 'none'.");
```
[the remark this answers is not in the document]
maxfirmbach: Looks good! This was always error prone ...

src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_muelu.cpp
```diff
@@ -71,8 +74,7 @@ void Core::LinearSolver::MueLuPreconditioner::setup(bool create, Epetra_Operator
 
       Teuchos::ParameterList& inverseList = muelulist_.sublist("MueLu Parameters");
 
-      std::string xmlFileName = inverseList.get<std::string>("MUELU_XML_FILE");
-      if (xmlFileName == "none") FOUR_C_THROW("MUELU_XML_FILE parameter not set!");
+      auto xmlFileName = inverseList.get<std::string>("MUELU_XML_FILE");
```
maxfirmbach, line 77: What is the difference here in comparison to `tekolist_.sublist("Teko Parameters").get<std::filesystem::path>("TEKO_XML_FILE");`? Shouldn't it be the same code?

sebproell: Initially, I tried to adapt the code in such a way that as little code as possible is changed. Looking at the diff now, I agree that this should be the same code everywhere. Ill see if I an unify that.

bgoderbauer (approved):

Issue #211, opened 2025-01-08
Adapt output for all fields to refactored output writers


eulovi:
> This issue was originally created by @isteinbrecher in https://gitlab.lrz.de/baci/baci/-/issues/899 on 2022-11-16.

## Description

Ensure that every field in BACI that writes visualization output uses the api from the output writer ([898](https://gitlab.lrz.de/baci/baci/-/issues/898))

To get an overview of the current state in BACI, we ask every @baci/baci_developers to append their case and current type of output (runtime / post) to the following list:
- structure
  - classic solid elements: **runtime** (users: @isteinbrecher @agebauer @NHagmeyer @maxfirmbach @eulovi @lengelhardt @scheider )
  - beam elements: **runtime** (users: @isteinbrecher @NHagmeyer @maxfirmbach)
  - nurbs (only displacement): **runtime** (users: @isteinbrecher @eulovi)
  - nurbs (displacement and stress): **post** (users: @isteinbrecher @eulovi)
  - solid scatra elements (for ssi, ssti) **post** (users: @cschmidt @ssinzig)
- beaminteraction: **runtime** (users: @isteinbrecher @maxfirmbach)
- fluid-beam interaction: **runtime** (users: @NHagmeyer)
- classical fluid (3d, contains a bug for 2D): **runtime** (users: @NHagmeyer)
- mortar interfaces: **post** (users: @mayrmt @eulovi @lengelhardt)
- monolithic FSI: **post** (users: @mayrmt)
- poro-multiphase-scatra: **post** (users: @b.wirthl @lkoeglmeier)
- contact (old time integration): **post** (users: @ChrSte, @mayrmt @lengelhardt)
- reduced airways: **post** (users: @geitner)
- scalar transport / elch: **post** (users: @cschmidt @ssinzig)

## Possible Solution and Definition of Done
- [ ] Query @baci/baci_developers how they currently write visualization output
- [ ] Migrate the current runtime output to the refactored output writer (possibly already done within [898](https://gitlab.lrz.de/baci/baci/-/issues/898))
- [ ] Implement visualization output for fields that currently only use `drt_post`

## Related Issues and Merge Requests

* Blocks
* Is blocked by
* Follows [897](https://gitlab.lrz.de/baci/baci/-/issues/897), [898](https://gitlab.lrz.de/baci/baci/-/issues/898)
* Precedes
* Related to
* Part of [896](https://gitlab.lrz.de/baci/baci/-/issues/896)
* Composed of

[...]

PR #360, opened 2025-02-12
Remove STRUCT_CURRENT_VOLUME output option

c-p-schmidt:
## Description and Context
In #211, it was decided that this output option is not desired/necessary anymore. It was anyways untested. Thus, it is removed including all related methods.

Part of #211


src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
```diff
@@ -1644,71 +1644,6 @@ void Solid::ModelEvaluator::Structure::determine_optional_quantity()
...
-  Core::LinAlg::SerialDenseMatrix empty_dummy_mat;
-  Core::LinAlg::SerialDenseVector empty_dummy_vec;
-
-  Solid::Elements::EvalErrorFlag ele_eval_error = Solid::Elements::ele_error_none;
-  for (unsigned elid = 0; elid < my_num_reles; ++elid)
-  {
-    Core::Elements::Element* rele = discret().l_row_element(elid);
-    rele->location_vector(discret(), la, false);
-
-    eval_data().set_action_type(Core::Elements::analyse_jacobian_determinant);
```
amgebauer, line 1681: This one is also just used here.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
```diff
@@ -1644,71 +1644,6 @@ void Solid::ModelEvaluator::Structure::determine_optional_quantity()
   evaluate_internal(eval_mat.data(), eval_vec.data());
 }
 
-/*----------------------------------------------------------------------------*
- *----------------------------------------------------------------------------*/
-bool Solid::ModelEvaluator::Structure::determine_element_volumes(
-    const Core::LinAlg::Vector<double>& x, std::shared_ptr<Core::LinAlg::Vector<double>>& ele_vols)
-{
-  // set action in params-interface
-  eval_data().set_action_type(Core::Elements::struct_calc_mass_volume);
```
amgebauer, line 1653: I think you can also remove this element action.

amgebauer (commented):
Looks good! I think you can additionally remove some implementations also on the element level.

c-p-schmidt:
> Looks good! I think you can additionally remove some implementations also on the element level.

You're right. Thank you for the hint!

amgebauer (approved):

davidrudlstorfer (approved):

ischeider:
Sorry, I missed the history of this action. Is there a reason to remove the current volume output? (except that it is untested). It could be a nice feature and I used it (in ABAQUS) many times, e.g. to evaluate the volume fractions of a composite structure or integrating internal variables over the volume in a post-processing step. And testing for this feature is probably easy, isn't it?

amgebauer:
@ischeider Evaluating the volume of the discretization (or a subset) are a few lines in pyvista (Probably 2 lines, maybe 3 if you count `import pyvista`). And then you are very flexible since you decide over which part you want to integrate. Especially, if you anyways want integrate over some other variables.

c-p-schmidt:
@ischeider: Even if you do not use `pyvista` but `ParaView`, you also directly obtain the element volumes by application of the Filter `CellSize`.

PR #353, opened 2025-02-11
Add solid mechanics tutorial

ischeider:
## Description and Context

This PR provides a solid mechanics tutorial for simulating a structure, which deforms plastically. 
The tutorial is aims at introducing the reader into solid mechanics simulations, also explaining some of the main sections and parameters to be set.

Also, the test is added to the framework-tests. Since the whole simulation would take more time than the timeout allows, the number of steps is reduced in the .head file.

The general structure of the tutorials chapter is slightly changed, some more information about the preprocessor cubit has moved to an introductory section, this is why also the fluid tutorial has changed.

For some more versatility, the material model gets a new parameter, TOL for the local Newton-Raphson iteration, which is given in many other models as well.

## Related Issues and Merge Requests
Related to issue #202

PR #361, opened 2025-02-12
Improve GTN

vryy:
## Description and Context

Improve some aspects of the local coupled damage-plasticity model GTN:

+ Fix the calculation of derivatives in case of using hardening function.
+ Add stabilization parameter for effective damage.
+ Add example with damage limit (fC).

!2205, !2465 (old gitlab) #353

### DIFF
--- a/src/mat/4C_mat_plasticgtn.cpp
+++ b/src/mat/4C_mat_plasticgtn.cpp
@@ -42,6 +42,7 @@ Mat::PAR::PlasticGTN::PlasticGTN(const Core::Mat::PAR::Parameter::Data& matdata)
       functionID_hardening_(matdata.parameters.get<int>("HARDENING_FUNC")),
       fc_(matdata.parameters.get<double>("FC")),
       kappa_(matdata.parameters.get<double>("KAPPA")),
+      ef_(matdata.parameters.get<double>("EF")),
       f0_(matdata.parameters.get<double>("F0")),
       fn_(matdata.parameters.get<double>("FN")),
       sn_(matdata.parameters.get<double>("SN")),
@@ -96,7 +97,6 @@ void Mat::PlasticGTN::pack(Core::Communication::PackBuffer& data) const
 void Mat::PlasticGTN::unpack(Core::Communication::UnpackBuffer& buffer)
 {
   isinit_ = true;
-
 
   Core::Communication::extract_and_assert_id(buffer, unique_par_object_id());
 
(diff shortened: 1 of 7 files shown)


ischeider (commented):
IMO, the parameter EF is quite unusual. If it improves convergence, might it make sense to provide a default != 0 ? Just wondering.

vryy:
> IMO, the parameter EF is quite unusual. If it improves convergence, might it make sense to provide a default != 0 ? Just wondering.

It's a newly devised parameter to help convergence in difficult cases. If it's 0 then the default GTN behavior is expected. Only when convergence is difficult I would suggest to adjust it (with care).

vryy:
@ischeider could you review shortly before any breaking changes in input files occur, to avoid further rebase?

sebproell:
@vryy Currently, there are no more breaking changes in the input files planned from my side.

vryy:
@sebproell thanks for the information. In any case, this PR is ready for review.

mayrmt:
@vryy If `EF` is a purely algorithmic parameter that is meant to improve the convergence properties of an iterative method but ideally should not change the final result, wouldn’t it be possible to adjust it over the iterations so that it vanishes at the moment of convergence?

One would then need an update formula that adjusts the parameter in each iteration. This is similar to the idea behind pseudo-transient continuation, a method used to achieve more robust convergence in Newton methods.

It could also be that this idea is completely off track — I’m not deeply familiar with the material model.

ischeider (approved):
It's an important bugfix, I hope the additional parameter can improve convergence, and maybe the additional example brings the coverage above 70% ;-)

src/mat/4C_mat_plasticgtn.cpp
```diff
@@ -215,6 +215,20 @@ void Mat::PlasticGTN::evaluate(const Core::LinAlg::Matrix<3, 3>* defgrd,
...
+  {
+    // reinitialize the hardening function pointer.
+    // this is to make sure the analysis works consistently with hardening by function. It happens
+    // that during redistribution or restart analysis, the global functions are reassigned after
+    // this constitutive law is set up. This will effectively null out the hardening function
+    // pointer.
+    hardening_function_ =
+        &Global::Problem::instance()->function_by_id<Core::Utils::FunctionOfAnything>(
+            functionID_hardening);
+  }
```
amgebauer, line 230: I think you can also do this during setup and unpack, then you don't need this check here.

vryy:
> It's an important bugfix, I hope the additional parameter can improve convergence, and maybe the additional example brings the coverage above 70% ;-)

This improvement does not exist in the literature, but still not sufficient to constitute a publication. I really hope we can find a way to resolve the mesh dependency problem in the softening domain :)))

src/mat/4C_mat_plasticgtn.cpp
vryy: @amgebauer thanks for the hint. The same problem happens for `Mat::PlasticNlnLogNeoHooke`. I will push the PR to fix both.

PR #363, opened 2025-02-12
Fix asan in reduced lung

bgoderbauer:
## Description and Context

I used the wrong vector for a postprocessing step, which led to a wrong access ID with more ranks.


amgebauer (approved):

c-p-schmidt (approved):

c-p-schmidt:
Thank you for the quick fix!

PR #366, opened 2025-02-12
Input: ensure printable names and descriptions

sebproell:
Catch unprintable characters, notably `\f`, early.

FYI @gilrrei @ischeider

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.cpp
+++ b/src/core/io/src/4C_io_input_spec_builders.cpp
@@ -479,6 +479,43 @@ Core::IO::Internal::MatchEntry& Core::IO::Internal::MatchEntry::append_child(
 }
 
 
+InputSpecTypeErasedBase::InputSpecTypeErasedBase(InputSpecTypeErasedBase::CommonData data)
+    : data(std::move(data))
+{
+  const auto check_for_unprintable_chars =
+      [](const std::string& check_string, const auto& field_type)
+  {
+    auto first_unprintable_char =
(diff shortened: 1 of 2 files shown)


src/core/io/src/4C_io_input_spec_builders.cpp
```diff
@@ -479,6 +479,43 @@ Core::IO::Internal::MatchEntry& Core::IO::Internal::MatchEntry::append_child(
...
 
 
+InputSpecTypeErasedBase::InputSpecTypeErasedBase(InputSpecTypeErasedBase::CommonData data)
+    : data(std::move(data))
+{
+  const auto check_for_unprintable_chars =
+      [](const std::string& check_string, const auto& field_type)
+  {
+    auto first_unprintable_char =
+        std::ranges::find_if_not(check_string, [](char c) { return std::isprint(c) || c == '\n'; });
```
gilrrei, line 489: Does it make sense to allow line breaks :thinking: ? What would be such a meaningful case?

sebproell: Line breaks don't make a ton of sense. But, as you might have guessed, we, of course, have them in the input right now;)

gilrrei: Ok, makes sense

amgebauer (approved):
Changes look good. Are other special characters handled correctly? What if I'm having a `"` in the description?

sebproell:
@amgebauer Having `\"` in the description works.

maxfirmbach (approved):

PR #369, opened 2025-02-13
Enable warnings as errors in doxygen

amgebauer:
Will never work, but let's see how many errors we get.


amgebauer:
Most warnings are:

* argument '...' of command [@]param is not found in the argument list: 1051
* Found unknown command '...': 740
* warning: explicit link request to '...' could not be resolved: 345
* no matching [...] member for: 121
* unbalanced grouping commands: 110
* Unsupported xml/html tag: 101
* found documented return type for ... that does not return anything: 92
* unable to resolve reference to ... for \ref command: 86
* warning: member ... belongs to two different groups. The second one found here will be ignored.: 86
* documented symbol ... was not declared or defined: 61
* no uniquely matching class member found for: 59

and a few more.

PR #372, opened 2025-02-13
Add missing definitions of class `Core::LinAlg::Solver`

vovannikov:
Without these 2 additional headers the code does not compile with `Trilinos` built in debug mode. The error was here:
```
include/Teuchos_TypeNameTraits.hpp:58:27: error: invalid use of incomplete type ‘class FourC::Core::LinAlg::Solver’
   58 |       return demangleName(typeid(T).name());
```


amgebauer (approved):

mayrmt (approved):

maxfirmbach:
@vovannikov Thanks for the fix!

PR #373, opened 2025-02-13
Let InputFile know about the InputSpec

### LINKED ISSUE #113: Export valid input options to yaml

sebproell:
This is a major PR and closes #113 :tada: 

- All sections are now exported to the `4C_metadata.yaml` file. `InputFile` now takes over this responsibilty.
- When reading into an `InputFile`, you will immediately get an error when a section is unknown. There is no need to remember to call a complicated special function for that (see #328).

Importantly, the implementation of the input mechanism is still generic, as all the knowledge about sections is handed in from the outside. This is also demonstrated by the unit tests, which read completely different sections than we use in the actual input files. Currently, this is done from a central place in the global module. Future modularization activities can move this into different modules.


c-p-schmidt (approved):

amgebauer (approved):

src/global_data/4C_global_data_read.cpp
```diff
@@ -44,6 +44,119 @@
...
+      global_legacy_module_callbacks().AttachFunctionDefinitions(functionmanager);
+
+      auto valid_functions = functionmanager.valid_function_lines();
+
+      // TODO Move the function number inside the general function section to stop relying
+      // on explicit function numbers.
+      for (unsigned i : std::views::iota(1, 30))
+      {
+        section_specs["FUNCT" + std::to_string(i)] = valid_functions;
+      }
```
amgebauer, line 90: Can you open an issue about this?

PR #306, opened 2025-01-31
Move inpar to physics

davidrudlstorfer:
## Description and Context

Moving inpar files to the corresponding physics modules

## Related Issues and Merge Requests

Part of #59

PR #381, opened 2025-02-15
[thermo] Add energy CSV runtime output

### LINKED ISSUE #367: [thermo] Remove energy file writing
## Description

The `thermo` module offers the possibility to write the energy state described by a scalar value per timestep into an external file. Due to #197, I would opt to change this.

As the energy state is only given by a single scalar quantity, traditional vtk-runtime output is not an option. In addition it‘s a quantity of secondary interest and thus not majorly important as e.g. the temperature state.

I would propose to add the energy to the console output in a certain verbosity level so the user is still able to see it without writing an extra file.

Any other suggestions or opinions?

- [x] Decide on how to treat the energy state output
- [x] Move the energy file writing to `core::IO::RuntimeCsvWriter`
- [x] Delete the old implementation

## Interested Parties

@isteinbrecher @c-p-schmidt

maxfirmbach:
## Description and Context

This PR removes the old thermal energy output mechanism and replaces it by a runtime csv based writer. In addition relevant input file sections are added as well as testing (the old output mechanism seemed to be untested).

It would be nice to have an identical way for writing csv based data over the whole code similar to vtk ... right now this doesn't seem to be the case. I kept the csv runtime writing in thermo similar to how vtk is done, from input file, parameters and so on.

Closes #367.


src/inpar/4C_inpar_IO_runtime_output_thermo.cpp
```diff
@@ -23,42 +23,58 @@ namespace Inpar
...
+              "NODE_GID", "No", "write 4C internal node GIDs", sublist_IO_vtk_output_thermo);
 
-        sublist_IO_output_thermo.move_into_collection(list);
+          sublist_IO_vtk_output_thermo.move_into_collection(list);
+        }
+
+        // csv runtime output
+        {
+          Core::Utils::SectionSpecs sublist_IO_csv_output{sublist_IO, "RUNTIME CSV OUTPUT"};
+          Core::Utils::SectionSpecs sublist_IO_csv_output_thermo{sublist_IO_csv_output, "THERMO"};
```
sebproell, line 66: First, I apologize for the current state of these parameters, this is not fully migrated to the `InputSpecs` yet, although they are used internally. The way we gather these parameters is still rather convoluted. 

Second, I am not a big fan of adding another of these inverted dependencies: Why should we have a global `RUNTIME CSV OUTPUT` section which is then filled with `THERMO` specific information? From a modularization perspective, we actively try to avoid and refactor such global collections (see e.g. #306). Can you move the thermo-specific choice to one of the thermo sections?

maxfirmbach: @sebproell Ah of course, I didn't really think about that while changing the input. I can move it to some of the thermo sections 👍.

maxfirmbach:
@sebproell What does the arrow related to the failing pipeline exactly mean? I don't get it.

sebproell:
> @sebproell What does the arrow related to the failing pipeline exactly mean? I don't get it.

Some things you entered were not parsed. I assume you removed some parameters in code but forgot to also remove them in the input files?

maxfirmbach:
@sebproell Right + some wrong parameter names from my side :sweat:, this is fiddly.

sebproell:
Feel free to drop me all input-related confusion in a Slack message and I can see if we can improve things. In this case, the error is already much better for yaml files. I will still improve the dat error and print the offending section. I'll add this to upcoming PRs

maxfirmbach:
This should be ready now.

sebproell (approved):

mayrmt (approved):

Issue #375, opened 2025-02-14
[Input] Make input sections for functions more flexible


sebproell:
_Originally posted by @amgebauer in https://github.com/4C-multiphysics/4C/pull/373#discussion_r1955919966_

PR #382, opened 2025-02-16
Allow 4C to convert dat to yaml

sebproell:
This PR allows to convert a dat input file yo yaml via `./4C --to-yaml <in> <out>`. The conversion is quite powerful in the sense that we correctly identify the implicit conventions in dat files and construct structured yaml input from the dat-style strings. For example, this 

```
----------------------------------------------------------MATERIALS

MAT 0   MAT_Electromagnetic   CONDUCTIVITY 0 PERMITTIVITY 1.732050808 PERMEABILITY 1.732050808
```

is correctly converted to

```
MATERIALS:
  - MAT: 0
    MAT_Electromagnetic:
      CONDUCTIVITY: 0
      PERMITTIVITY: 1.732050808
      PERMEABILITY: 1.732050808
```


ischeider:
Thanks for the definitely useful PR! Out of curiosity: what happens when a parameter cannot be converted? Does it stop immediately or just print it out and continue? I would definitely prefer the latter.

sebproell:
>  a dat file that has an invalid parameter.

 @ischeider Ah, good question! A little bit of background: dat does not have the typical concept of keys, values, objects and lists. The only thing this "format" does is structure lines of text into sections (denoted by `--`). This is one of the main reasons why the dat format is such a pain to work with. Take for example, XML and yaml which you can convert into equivalent representations irrespective of file contents. We cannot do the same for dat to yaml.
The only way to read a dat file in a meaningful way, is when you know what you expect. This is now encoded in the `InputSpec` class. Long story short: we have to read in the whole file and interpret the dat content, at the same time we **validate all parameters**, and then we write everything out as yaml.

src/global_data/4C_global_data_read.cpp
```diff
@@ -82,20 +82,23 @@ namespace
...
-      // in the test files.
-      for (unsigned i : std::views::iota(1, 30))
-      {
-        section_specs["FUNCT" + std::to_string(i)] = valid_functions;
-      }
+      // The FUNCT sections are special and do not fit into the usual pattern of sections and the
+      // capabilities of InputSpec. The special FUNCT<n> section is not supposed to be entered by
+      // users but we use this information inside the input file. Not pretty, but it works.
+      // TODO remove this hack by restructuring the input of functions.
+      section_specs["FUNCT<n>"] = valid_functions;
```
sebproell, line 89: @isteinbrecher @davidrudlstorfer Since I had an intermediate state that would have broken meshpy again, I opted for a new approach without the magic number.

@gilrrei I think we need a special case in schema for this. I am open to a different spelling of `FUNCT<n>`, if this makes this easier for you. 

Related to #375

tests/input_files/mixture_solid_material_viscoelasthyper.4C.yaml
```diff
@@ -1,70 +1,163 @@
...
-  STRUCT_STRAIN: gl
-  STRUCT_STRESS: 2pk
+  STRUCT_STRESS: Yes
+  STRUCT_STRAIN: Yes
 IO/RUNTIME VTK OUTPUT:
-  OUTPUT_DATA_FORMAT: ascii
   INTERVAL_STEPS: 1
-  EVERY_ITERATION: no
+  OUTPUT_DATA_FORMAT: ascii
 IO/RUNTIME VTK OUTPUT/STRUCTURE:
```
sebproell, line 7: Regarding the conversion mechanism (here demonstrated for old yaml files):

I made sure that sections are ordered the same way after conversion. Within a section, I did not bother to keep this order. The order instead corresponds the order in the `InputSpec`. Also note how we automatically removed default values (here: `EVERY_ITERATION: no`).

sebproell (commented):
I am running the test suite again with every file force-fully converted to yaml. Some tests will not work this way, but this is related to the test infrastructure (multiscale stuff). I will drop the HACK commit afterwards. 

What is a good strategy to test the conversion mechanism? Right now, I would not convert all input files to yaml. @davidrudlstorfer Should I add an option to our `four_c_test` function to perform file conversion and apply it to a few files?

sebproell:
This is now ready for review. The last commit adds 10 tests where I convert dat input to yaml before running the test. I chose these tests across the most used physics and depending on special sections in their files.

davidrudlstorfer (approved):

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -124,6 +124,7 @@ endfunction()
 #                         If two files are provided the second input file is restarted based on the results of the first input file.
 
 # optional:
+# CONVERT_TO_YAML:                If set, convert the input file to YAML format before running the test
```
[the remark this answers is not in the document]
davidrudlstorfer: As we've discussed offline I did not look into your changes accurate enough.

You are writing the temporary yaml file into the build directory - that's why recycling the standard test function won't work (it is checked whether the file exists in the source test input file directory)

With #56 this could be achieved in a simpler fashion - but that's not yet done.

So let's get this merged

apps/global_full/4C_global_full_main.cpp
```diff
@@ -273,6 +344,16 @@ int main(int argc, char* argv[])
...
+  else if ((argc == 4) && (strcmp(argv[1], "--to-yaml") == 0))
+  {
+    if (Core::Communication::my_mpi_rank(lcomm) == 0)
+    {
+      Core::IO::InputFile input_file = Global::set_up_input_file(lcomm);
+      input_file.read(argv[2]);
+      std::ofstream output_file(argv[3]);
+      input_file.write_as_yaml(output_file);
+    }
+  }
```
amgebauer, line 356: Maybe we should at some point cleanup our input and use a library like https://github.com/p-ranav/argparse

sebproell: Absolutely :+1: Not sure if we need a lib for that but note that Teuchos also has some code for that; we use it in pre_exodus.

PR #386, opened 2025-02-17
InputFile: allow arbitrary FUNCT<n> sections

### LINKED ISSUE #385: [BUG] 4C can't handle more than 30 functions any more

It seems like with #373 the ability of 4C to handle more than 30 functions was removed.

## Expectations
4C should be able to handle arbitrarily many functions (maybe limited by some integer value range, but not by an arbitrarily chosen value of 30).

## Steps to Reproduce
The attached input file with 500 functions works as expected with 6a75fd555dad8a603d76ee921e38f2e012fd55f9 but fails with the current `main` branch with an error like
```
  Section 'FUNCT31' is not a valid section name.
```

Attached input file (has to be renamed to `.dat`)
[a_lot_of_functions.txt](https://github.com/user-attachments/files/18823298/a_lot_of_functions.txt)

- Blocks
- Follows
- Related to #373
- Part of
- Composed of

## Interested Parties

@knarfnitram @sebproell

sebproell:
Closes #385

### DIFF
--- a/src/global_data/4C_global_data_read.cpp
+++ b/src/global_data/4C_global_data_read.cpp
@@ -83,7 +83,8 @@ namespace
       auto valid_functions = functionmanager.valid_function_lines();
 
       // TODO Move the function number inside the general function section to stop relying
-      // on explicit function numbers.
+      // on explicit function numbers. The magic number 30 is equal to the highest number appearing
+      // in the test files.
       for (unsigned i : std::views::iota(1, 30))
       {
         section_specs["FUNCT" + std::to_string(i)] = valid_functions;
--- a/src/core/io/src/4C_io_input_file.cpp
+++ b/src/core/io/src/4C_io_input_file.cpp
@@ -259,10 +259,19 @@ namespace Core::IO
 
       bool is_section_known(const std::string& section_name) const
       {
-        return valid_sections_.find(section_name) != valid_sections_.end() ||
+        // The input for functions might introduce an arbitrary number of sections called
+        // FUNCT<n>, where n is a number. As long as this input is not restructured, we need this
+        // manual hack.
+        const bool is_hacky_function_section =
+            section_name.starts_with("FUNCT") &&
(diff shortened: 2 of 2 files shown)


src/global_data/4C_global_data_read.cpp
```diff
@@ -83,7 +83,8 @@ namespace
       auto valid_functions = functionmanager.valid_function_lines();
 
       // TODO Move the function number inside the general function section to stop relying
-      // on explicit function numbers.
+      // on explicit function numbers. The magic number 30 is equal to the highest number appearing
+      // in the test files.
       for (unsigned i : std::views::iota(1, 30))
```
isteinbrecher, line 88: Is there a reason we still need this? Should the fix in `is_section_known` not allow for all function IDs?

sebproell, line 88: I validated that this should fix the problem by reducing the magic number to 2 and running minimal tests. I would like to keep the magic number at 30 for now to get proper schema validation.

isteinbrecher (commented):
Thanks for looking into this so quickly @sebproell

src/global_data/4C_global_data_read.cpp
isteinbrecher: That answers my question!

isteinbrecher: See previous comment by @sebproell

isteinbrecher (approved):

src/global_data/4C_global_data_read.cpp
sebproell: We should definitely work to remove it. I will check with @gilrrei when he is back, if the schema regex features can help us here. On the other hand, the input should really change to something like:

```
FUNCTIONS:
  - ID: 1
    EXPRESSION: x*y
  - ID: 100
    ...
```

But let's not get ahead of ourselves. This would require an automatic input file upgrade mechanism.

mayrmt (approved):

PR #388, opened 2025-02-17
Rework `Core::FE::extract_my_values()`

amgebauer:
Until now, extracting the local dofs always looks like this:

```c++
    std::vector<double> disp;
    Core::FE::extract_my_values(global_displacements, disp, global_ids);
```

With this PR, we can simply write

```c++
    const std::vector<double> disp = Core::FE::extract_values(global_displacements, global_ids);
```
(allow const-variables, easier readable).

* Commit 1: Implements the function in Core
* Commit 2: Applies it in solid_3D_ele
* Commit 3: Replaces it everywhere

### DIFF
--- a/src/ale/4C_ale_ale3_evaluate.cpp
+++ b/src/ale/4C_ale_ale3_evaluate.cpp
@@ -129,11 +129,9 @@ int Discret::Elements::Ale3::evaluate(Teuchos::ParameterList& params,
   {
     case calc_ale_laplace_material:
     {
-      std::vector<double> my_dispnp;
       std::shared_ptr<const Core::LinAlg::Vector<double>> dispnp =
           discretization.get_state("dispnp");
-      my_dispnp.resize(lm.size());
-      Core::FE::extract_my_values(*dispnp, my_dispnp, lm);
+      std::vector<double> my_dispnp = Core::FE::extract_values(*dispnp, lm);
 
       Ale3ImplInterface::impl(this)->static_ke_laplace(
           this, discretization, elemat1, elevec1, my_dispnp, mat, false);
@@ -142,11 +140,9 @@ int Discret::Elements::Ale3::evaluate(Teuchos::ParameterList& params,
     }
     case calc_ale_laplace_spatial:
     {
-      std::vector<double> my_dispnp;
       std::shared_ptr<const Core::LinAlg::Vector<double>> dispnp =
           discretization.get_state("dispnp");
-      my_dispnp.resize(lm.size());
-      Core::FE::extract_my_values(*dispnp, my_dispnp, lm);
+      std::vector<double> my_dispnp = Core::FE::extract_values(*dispnp, lm);
 
       Ale3ImplInterface::impl(this)->static_ke_laplace(
           this, discretization, elemat1, elevec1, my_dispnp, mat, true);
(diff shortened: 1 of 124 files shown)


mayrmt:
@amgebauer I like the new interface! 

How many non-auto-replacable occurrences are there still in the code? I am debating, whether we should switch everything right now instead of having two implementations for the same feature.

amgebauer:
> @amgebauer I like the new interface!
> 
> How many non-auto-replacable occurrences are there still in the code? I am debating, whether we should switch everything right now instead of having two implementations for the same feature.

I could fix 300 automatically, but there are still around 300 remaining. That's a lot of manual work. The old feature of course reuses the new one. So, we only have one implementation, but two interfaces.

src/ale/4C_ale_ale3_evaluate.cpp
```diff
@@ -129,11 +129,10 @@ int Discret::Elements::Ale3::evaluate(Teuchos::ParameterList& params,
   {
     case calc_ale_laplace_material:
     {
-      std::vector<double> my_dispnp;
+      ;
```
mayrmt, line 132: This looks like an artifact from auto-replace.

mayrmt (commented):
I am not happy with leaving the old interface in there, but I can see, that it is hard to do, since automation might be impossible.

src/core/fem/src/general/utils/4C_fem_general_extract_values.hpp
```diff
@@ -14,34 +14,112 @@
...
+   * @note Use @p Core::FE::extract_values() instead
+   *
+   * @param global
+   * @param local
+   * @param lm
+   */
+  inline void extract_my_values(const Core::LinAlg::Vector<double>& global,
+      std::vector<double>& local, const std::vector<int>& lm)
+  {
+    local = extract_values(global, lm);
```
sebproell, line 93: I would implement these the other way round. Now I have to make a copy even when I use this signature that could reuse memory.

sebproell (commented):
Most times, I like to return by value as it gives a much more natural flow to the code. However, I am not sure if it is the best decision in such core FEM routines. The new implementation is going to allocate every time it is called for a very short-lived small piece of memory. I think you should at least flip the implementation as commented below and also comment on when to call which function. 4C is not optimized to a high degree, so this might not be the most important point but I still want to raise awareness.

amgebauer:
> Most times, I like to return by value as it gives a much more natural flow to the code. However, I am not sure if it is the best decision in such core FEM routines. The new implementation is going to allocate every time it is called for a very short-lived small piece of memory. I think you should at least flip the implementation as commented below and also comment on when to call which function. 4C is not optimized to a high degree, so this might not be the most important point but I still want to raise awareness.

The function is used in 90% of the cases in a way where we anyway create the std::vector before. Additionally, the function is a pure header with a template, so the compiler can inline it (and will probably do). I'm confident that this will end up in the same compiled code. For optimized pieces of code, this PR also provides a version that returns a fixed-sized `std::array`.

With some further regex-search-and-replace, I could remove the old interface.

sebproell (approved):
@amgebauer As I said, it isn't terribly important to have the old function. I just wanted to note that there is a world where it is useful to fill a chunk of pre-allocated memory and not reallocate all the time. As you correctly pointed out, we mostly do not reuse the memory in the loops, so I am OK with merging.

mayrmt (approved):

PR #389, opened 2025-02-17
Move output of accelerations to vtk output in new structure time integration

c-p-schmidt:
## Description and Context
As part of #211, the output of accelerations is moved from binary-based output to vtk-based output. Old output is removed, including the post-processor implementations; test cases are cleaned w.r.t. the option that is not available anymore.

One test is added to test the introduced vtk-based acceleration output.

Part of #211


tests/input_files/ref/membrane_patch_new_struct/vtk-files/structure-00006-0.vtu
amgebauer, line 1: Is there a slightly smaller suitable test to test the new output?

amgebauer (approved):

tests/input_files/ref/membrane_patch_new_struct/vtk-files/structure-00006-0.vtu
c-p-schmidt: You mean because of the approximately 10k lines of the reference files?

I intentionally chose this test because the next quantity to migrate (I locally already have it) is the membrane thickness. For that, this seems to be the only reasonable test and testing that additional quantity is then only additional lines in the vtu files.

However, I could deactivate other quantities like displacements, velocity and strain output that are already part of other tests to save some lines.

c-p-schmidt: Now it's approximately 7k lines less. Thanks for the hint!

tests/list_of_tests.cmake
```diff
@@ -2113,6 +2112,7 @@ four_c_test_vtk(embeddedmesh_two_cubes_nurbs_1_straight_up_vtk.dat embeddedmesh_
 four_c_test_vtk(f3_channel_EOS_hex27_runtime_output.dat f3_channel_EOS_hex27_10x10x10_quad_inflow.dat 3 xxx-fluid.pvd ref/fluid_runtime_output-fluid.pvd 1e-08)
 four_c_test_vtk(fsi_runtime_output.dat fsi_dc_part_ait_ga_ga.dat 2 xxx-fluid.pvd ref/fsi_runtime_output-fluid.pvd 1e-08 1)
 four_c_test_vtk(mat_druckerprager_RetToCone-vtk-plastic.dat mat_druckerprager_RetToCone.dat 2 xxx-structure.pvd ref/mat_druckerprager_RetToCone/RetToCone-structure.pvd 1e-08 9.5)
+four_c_test_vtk(membrane_patch_new_struct_vtk.dat membrane_patch_new_struct.dat 2 xxx-structure.pvd ref/membrane_patch_new_struct/membrane.pvd 1e-08 0.6)
```
isteinbrecher, line 2115: Can you also add one iteration step, to test that the iteration output works?

amgebauer (approved):

sebproell (approved):

tests/list_of_tests.cmake
c-p-schmidt: Do you mean activating the `EVERY_ITERATION` flag and also comparing one of those files?

isteinbrecher: Yes, exactly.

c-p-schmidt: OK, but do we really benefit from it? I just adapted a method that now requires additional input (the accelerations) which is called from the methods that write after every step or after every iteration.

Since the basic functionality (writing after every step and every iteration) is already tested for the structure module, there is no real additional benefit; maybe besides that, both function calls actually get the accelerations and not a different quantity. Or am I missing something?

Please don't get me wrong. Adding this would not be an effort (only rerunning the simulation and copying the output). I'm just hesitant to add about 3-4k more lines to the repo if the benefit is unclear.

isteinbrecher: I personally like to have the iteration output tested because this is currently the only way we have toncheck if the linearizations are correct (or at least don’t change). But I get your points, so feel free to continue as you wish.

PR #393, opened 2025-02-18
Emit type metadata in a easier parseable format

sebproell:
As requested by @gilrrei


src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -735,9 +736,12 @@ required: true
...
     default: 42
   - name: b
-    type: vector<double>
+    type: vector
+    value_type:
+      noneable: true
+      type: double
     required: false
-    default: [1,2,3]
+    default: [1,none,3]
```
gilrrei, line 744: I just realised we were talking about the same thing but this solution won't unfortunately work. 

The problem is there is no type `none` in yaml/json. Instead it's called `null` (awful name, super unintuitive). So if we want to have  schema validation we have to use `null`. Otherwise we can't create a schema to differentiate between arbitrary strings and `none` losing validation in these cases.

amgebauer: I think that's just because json comes from javascript where `none` is `null`. In my eyes, it's consistent to adopt `null` here.

sebproell: OK, that's no problem. 

IMO, `null`, `none` and `nil` are commonly used terms for the [billion dollar mistake](https://en.wikipedia.org/wiki/Null_pointer). Obviously, yaml uses `null` for that. If we generate python data classes we use `None` and we if I one day implement my personal favorite, lua input, we would use `nil` in there. The input in 4C just interprets these appropriately.

The only question is: what should we call the ability to not have a value in C++ code? The type I use internally is `std::optional`, but I don't like to use that directly because people will confuse it with the `required` property of `InputSpec`. Thus, @lauraengelhardt and I opted for `Noneable`.

gilrrei: In python you sometimes find similar constructs commonly called `NotSet`

sebproell: I personally like Haskell's name `Maybe` a lot. We can bikeshed a bit more for this name, but the PR makes sense regardless of the outcome. I will push the `null` change here.

src/core/io/src/4C_io_yaml.hpp
```diff
@@ -201,7 +201,7 @@ namespace Core::IO
   void read_value_from_yaml(ConstYamlNodeRef node, std::optional<T>& value)
   {
     FOUR_C_ASSERT_ALWAYS(node.node.has_val(), "Expected a value node.");
-    if (node.node.has_val() && node.node.val() == "none")
+    if (node.node.val_is_null())
```
sebproell, line 204: This convenience function was added to ryml in a recent release, so I updated the dependecy.

davidrudlstorfer (approved):

amgebauer (approved):

PR #400, opened 2025-02-19
1361 ifpack move parameters and options to xml file

dharinib98:


mayrmt:
@dharinib98, this branch needs to be rebased to resolve merge conflicts.

src/core/linear_solver/src/method/4C_linear_solver_method_linalg.cpp
```diff
@@ -237,11 +237,15 @@ Teuchos::ParameterList translate_four_c_to_ifpack(const Teuchos::ParameterList&
 {
   Teuchos::ParameterList ifpacklist;
 
-  ifpacklist.set("fact: level-of-fill", inparams.get<int>("IFPACKGFILL"));
+  /*ifpacklist.set("fact: level-of-fill", inparams.get<int>("IFPACKGFILL"));
   ifpacklist.set("partitioner: overlap", inparams.get<int>("IFPACKOVERLAP"));
   ifpacklist.set("schwarz: combine mode",
       inparams.get<std::string>("IFPACKCOMBINE"));    // can be "Zero", "Add", "Insert"
   ifpacklist.set("schwarz: reordering type", "rcm");  // "rcm" or "metis" or "amd"
+  */
```
mayrmt, line 245:
```suggestion
```

src/core/linear_solver/src/method/4C_linear_solver_method_linalg.cpp
```diff
@@ -237,11 +237,15 @@ Teuchos::ParameterList translate_four_c_to_ifpack(const Teuchos::ParameterList&
...
+
+  std::string xmlfile = inparams.get<std::string>("IFPACK_XML_FILE");
```
mayrmt, line 247:
```suggestion
  const std::string xmlfile = inparams.get<std::string>("IFPACK_XML_FILE");
```
... and also at other places throughout this PR.

tests/input_files/beam3r_line2_backweuler_browndyn_largesys.dat
```diff
@@ -61,8 +61,7 @@ AZTOL                  1.0E-3
 AZCONV                 AZ_r0
 AZITER                 1000
 AZSUB                  50
-IFPACKGFILL            1
-IFPACKOVERLAP          1
+IFPACK_XML_FILE                 xml/ifpack.xml
 IFPACKCOMBINE          Zero
```
mayrmt, line 65:
```suggestion
IFPACK_XML_FILE        xml/ifpack.xml
IFPACKCOMBINE          Zero
```

mayrmt (commented):
The changes to the input files change the preconditioner for some tests, since the new xml file always use zero overlap and a fill level of `0`, while some tests have used different settings. I don't say, that you have to change this. I just want to create awareness for this change.

What are the plans to actually dis-allow the old way of configuring Ifpack preconditioners>

tests/input_files/beam3r_line2_backweuler_browndyn_largesys.dat
sebproell: Yet another reason to use yaml for input files :)

dharinib98: yes, this is also the case with overlap values, not all files use a value of 0. Based on the number of tests that fail, we can think about a solution by maybe adding another xml file for the  higher gfill values

mayrmt: We can have one or two default xml files in the central xml file collection. If individual tests require a specialized preconditioned, then we can put the xml file just next to the input file.

tests/input_files/beam3r_line2_backweuler_browndyn_largesys.dat
```diff
@@ -61,8 +61,7 @@ AZTOL                  1.0E-3
 AZCONV                 AZ_r0
 AZITER                 1000
 AZSUB                  50
-IFPACKGFILL            1
-IFPACKOVERLAP          1
+IFPACK_XML_FILE                 xml/ifpack.xml
```
mayrmt, line 64: This places the xml file into the top level directory of all xml files for solvers/preconditioners. There is an existing folder structure. @maxfirmbach, should we sort it into the folder structure? If yes, where? Into a new directory?

maxfirmbach: @dharinib98 @mayrmt We need to figure out a place, not sure where it fits best, in theory we could just create a folder called `preconditioner` and put all preconditioner related .xml files there with an updated naming scheme. So `xml/preconditioner/ifpack.xml` would work for me right now.

maxfirmbach:
> The changes to the input files change the preconditioner for some tests, since the new xml file always use zero overlap and a fill level of `0`, while some tests have used different settings. I don't say, that you have to change this. I just want to create awareness for this change.
> 
> What are the plans to actually dis-allow the old way of configuring Ifpack preconditioners>

@mayrmt We will use the provided .xml file for now and see how many tests fail. From that point on we slowly work towards passing all tests with a minimal set of configs.

src/inpar/4C_inpar_solver.cpp
```diff
@@ -62,15 +62,8 @@ namespace Inpar::SOLVER
...
-          "The amount of overlap used for the ifpack \"ilu\" preconditioner.", list);
-
-      Core::Utils::int_parameter("IFPACKGFILL", 0,
-          "The amount of fill allowed for an internal \"ilu\" preconditioner.", list);
-
-      std::vector<std::string> ifpack_combine_valid_input = {"Add", "Insert", "Zero"};
-      Core::Utils::string_parameter("IFPACKCOMBINE", "Add",
-          "Combine mode for Ifpack Additive Schwarz", list, ifpack_combine_valid_input);
+      Core::Utils::string_parameter(
+          "IFPACK_XML_FILE", "none", "xml file defining any IFPACK options", &list);
```
sebproell, line 66: We can read files in a better way (so that relative paths are correctly resolved). See e.g. https://github.com/4C-multiphysics/4C/blob/main/src/inpar/4C_inpar_solver.cpp#L104-L108

src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_ifpack.cpp
```diff
@@ -44,24 +46,30 @@ void Core::LinearSolver::IFPACKPreconditioner::setup(bool create, Epetra_Operato
 
     pmatrix_ = std::make_shared<Epetra_CrsMatrix>(*A_crs);
 
-    // get the type of ifpack preconditioner from solver parameter list
-    std::string prectype = solverlist_.get("Preconditioner Type", "ILU");
-    const int overlap = ifpacklist_.get("IFPACKOVERLAP", 0);
+    std::string xmlFileName = ifpacklist_.get<std::string>("IFPACK_XML_FILE");
```
mayrmt, line 49:
```suggestion
    const std::string xmlFileName = ifpacklist_.get<std::string>("IFPACK_XML_FILE");
```

src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_ifpack.cpp
```diff
@@ -44,24 +46,30 @@ void Core::LinearSolver::IFPACKPreconditioner::setup(bool create, Epetra_Operato
...
+    Teuchos::ParameterList ifpack_params;
+
+    auto comm = Core::Communication::to_teuchos_comm<int>(
+        Core::Communication::unpack_epetra_comm(pmatrix_->Comm()));
+
+    Teuchos::updateParametersFromXmlFileAndBroadcast(
+        xmlFileName, Teuchos::Ptr(&ifpack_params), *comm);
+
+    auto prectype = ifpack_params.get<std::string>("Preconditioner type");
+    auto overlap = ifpack_params.get<int>("Overlap");
```
mayrmt, line 61: Const qualifiers?
```suggestion
    const std::string prectype = ifpack_params.get<std::string>("Preconditioner type");
    const int overlap = ifpack_params.get<int>("Overlap");
```

src/inpar/4C_inpar_solver.cpp
dharinib98:
```cpp
    {
      list.specs.emplace_back(
          Core::IO::InputSpecBuilders::entry<Core::IO::Noneable<std::filesystem::path>>(
              "IFPACK_XML_FILE", {.description = "xml file defining any IFPACK preconditioner",
                                    .default_value = Core::IO::Noneable<std::filesystem::path>("/home/a11bdhba/Ddrive/Codes/4C/tests/input_files/xml/preconditioner/ifpack.xml")}));
    }
```

I gave the absolute path since I did not know how to get the absolute one form the relative one. @sebproell Would you mind helping me?

sebproell: I think the underlying problem is that you should not use a default value here.

sebproell: Sorry, I didn't see that you are trying to use some older constructs. Could you rebase this onto main? The `Core::IO::Noneable` is now simply `std::optional`.

dharinib98: The tests that fail currently are due to the following error:

```
 terminate called after throwing an instance of 'Teuchos::Exceptions::InvalidParameterName'
  what():  Error!  The parameter "IFPACK_XML_FILE" does not exist
in the parameter (sub)list "ANONYMOUS".
```

and these tests do not have the IFPACK_XML_FILE in their respective dat files. Can someone check if there is something fundamentally wrong with this branch?

src/core/linear_solver/src/method/4C_linear_solver_method_linalg.cpp
```diff
@@ -237,11 +237,8 @@ Teuchos::ParameterList translate_four_c_to_ifpack(const Teuchos::ParameterList&
 {
   Teuchos::ParameterList ifpacklist;
 
-  ifpacklist.set("fact: level-of-fill", inparams.get<int>("IFPACKGFILL"));
-  ifpacklist.set("partitioner: overlap", inparams.get<int>("IFPACKOVERLAP"));
-  ifpacklist.set("schwarz: combine mode",
-      inparams.get<std::string>("IFPACKCOMBINE"));    // can be "Zero", "Add", "Insert"
-  ifpacklist.set("schwarz: reordering type", "rcm");  // "rcm" or "metis" or "amd"
+  auto xmlfile = inparams.get<std::optional<std::filesystem::path>>("IFPACK_XML_FILE");
+  if (xmlfile) ifpacklist.set("IFPACK_XML_FILE", xmlfile->string());
```
sebproell, line 241: Here you set `IFPACK_XML_FILE` but only if the optional path is given in the input file...

src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_ifpack.cpp
```diff
@@ -44,23 +48,31 @@ void Core::LinearSolver::IFPACKPreconditioner::setup(bool create, Epetra_Operato
 
     pmatrix_ = std::make_shared<Epetra_CrsMatrix>(*A_crs);
 
-    // get the type of ifpack preconditioner from solver parameter list
-    std::string prectype = solverlist_.get("Preconditioner Type", "ILU");
-    const int overlap = ifpacklist_.get("IFPACKOVERLAP", 0);
+    const std::string xmlFileName = ifpacklist_.get<std::string>("IFPACK_XML_FILE");
```
sebproell, line 51: ... Here you unconditionally access `IFPACK_XML_FILE`.

sebproell (commented):
@dharinib98 I think the following might be your problem:

src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_ifpack.cpp
dharinib98: Once I have an if condition, the tests that have no IFPACK_XML_FILE still fail, like in the pipeline.

mayrmt: @dharinib98, I had a look at some of the failing tests. At first glance, I only saw tests that fail with a segmentation fault after the calculation of the first residual in Belos and before the first Belos iterations. That is exactly the point in the simulation workflow, where one would setup a preconditioner. So, maybe access to the IFPACK file is still not working properly.

You can run those tests locally to speed-up the development cycle, so you don't have to wait for the pipeline to complete.

src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_ifpack.cpp
```diff
@@ -44,23 +48,31 @@ void Core::LinearSolver::IFPACKPreconditioner::setup(bool create, Epetra_Operato
 
     pmatrix_ = std::make_shared<Epetra_CrsMatrix>(*A_crs);
 
-    // get the type of ifpack preconditioner from solver parameter list
-    std::string prectype = solverlist_.get("Preconditioner Type", "ILU");
-    const int overlap = ifpacklist_.get("IFPACKOVERLAP", 0);
+    if (ifpacklist_.isParameter("IFPACK_XML_FILE"))
```
dharinib98, line 51: So from what I understood, because of this condition now, the tests that use ILU but did not have IFPACK parameters specified (which means now there is no IFPACK_XML_FILE parameter defined) fail.

dharinib98: @mayrmt This is also why the access to the xml file (just for these tests) is not working like you suggested. So I tried to specify the  ` IFPACK_XML_FILE` path explicitly in some of these tests and it seems to work atleast locally. Either we specify this for all the others if this sounds like a solution or we have to change the structure of the if condition so all tests, irrespective of the ` IFPACK_XML_FILE` being defined, go through the preconditioning as it was before.

mayrmt: @dharinib98, do you mean, that ILU has been the implicit default (and was used, even if it has not been defined in the input file) and now needs to be explicitly defined by specifying a path to an xml file?

dharinib98: I was just checking that, it looks like Belos is the reason and not ILU. So the issue is with the tests that use Belos and have no explicit definition of the ifpack xml file.

mayrmt:
@dharinib98, yes, I can have a look -- probably not today, but soon.

mayrmt:
@dharinib98, before your change, 4C passed some settings to Ifpack that overwrote Ifpack's default behavior. When reading a plain Ifpack xml file (without these parameters), Ifpack behaves differently since it relies on its internal defaults.

I have added the old 4C defaults to the xml file. This should restore the old behavior and make tests pass again. 

If the change in default values is the only reason for failing tests, then I am happy. We can still decide whether to keep the old defaults as recommendations to the user or if we want to rely on Ifpack's default settings and adapt 4C test results accordingly. If we change test results, at least we then know that the change is safe and ok, since it only came from a change in default settings of an external package.

dharinib98:
@mayrmt Thanks for your inputs

dharinib98:
@mayrmt I see that with this recent change, there are two new tests that fail. Would it help to make a separate xml for just these two tests?

mayrmt:
@dharinib98 Which ones? `tutorial_fsi`? They are failing for a different reason.

dharinib98:
@mayrmt These two are the ones I meant and they used to pass: 
f3_bfs_xwall_8x8x8_bl1_turbinflow.dat
xfluid_kimmoin_stat_24x24x1_wedge15_levelset.dat

tests/input_files/xfluid_kimmoin_stat_24x24x1_wedge15_levelset.dat
```diff
@@ -99,8 +99,7 @@ AZPREC                          ILU
 AZREUSE                         1
 AZTOL                           1.0E-8
 AZCONV                          AZ_r0
-IFPACKOVERLAP                   3
-IFPACKGFILL                     3
+IFPACK_XML_FILE                 xml/preconditioner/ifpack.xml
```
mayrmt, line 102: @dharinib98, this test fails since the old version used an overlap of 3 and a fill level of 3, while the new version in the xml file uses an overlap of 0 and a fill level of 0. I have replicated this locally: by setting the "old" values in the new xml file, this tests passes.

Please go over all failing test files and make sure to check, whether the configuration in the xml file resembles the old configuration or not. If not, please verify that restoring the old configuration fixes the problem.

If the tests fail due to a lack of convergence, then we need another xml file with the old settings. If the tests fail just for slight differences in the results, we might change the results.

Regarding lack of convergence: you have to look at the solver output manually and maybe even compare iteration counts in `main` and your branch. Krylov solvers can hit the maximum iteration count, but then do not stop the simulation, but just continue. In this case, the rest of the simulation can become useless (as it is the case for this test for example) and the result tests fail due to the lack of a properly converged linear solver.

mayrmt:
> most tests that were failing before were the ones using umfpack and had the ifpack xml path(again something that I added).

Why were they failing? Can we catch that and print a meaningful error message for the user?

> After removing this unnecessary addition, all these tests (except xfluid_taylor_couette_NavSlip_eps1.0_25x25_krylov_levelset_altgeogeneration.dat and the tutorials) passed on my machine locally, but somehow they still seem to fail here.

I believe that the tutorial tests need the xml file to be copied from the source directory to the build directory. This is usually done in the test macro of the tutorial tests, though I am afraid that it can't handle the case of multiple xml files, which would be required for the 3D FSI tutorial.

dharinib98:
I'm not entirely sure what was going wrong with the UMFPACK tests. The error was a segmentation fault, and Belos was mentioned in the error string. I still find this strange, since the input explicitly specifies UMFPACK. However, something in the backend was still passing a Belos solver path, which led to an improperly initialized preconditioner being accessed.

I modified the code (with some hard coding) to allow preconditioner initialization even when no IFPACK_XML_FILE is specified. I believe this should resolve the remaining test failures.

mayrmt:
@dharinib98 If we use UMFPACK, we should not need a preconditioner.

dharinib98:
@mayrmt Yes, that is true. But that is what I do not understand yet. This can be a temporary solution if all the tests pass. I feel we need to investigate why UMFPACK uses preconditioning for us to be able to come up with a sustainable solution for these tests.

src/inpar/4C_inpar_solver.cpp
```diff
@@ -57,17 +57,8 @@ namespace Inpar::SOLVER
...
-
-        parameter<int>("IFPACKGFILL",
-            {.description = "The amount of fill allowed for an internal \"ilu\" preconditioner.",
-                .default_value = 0}),
-
-
-        deprecated_selection<std::string>("IFPACKCOMBINE", {"Add", "Insert", "Zero"},
-            {.description = "Combine mode for Ifpack Additive Schwarz", .default_value = "Add"}),
+        Core::IO::InputSpecBuilders::parameter<std::optional<std::filesystem::path>>(
+            "IFPACK_XML_FILE", {.description = "xml file defining any Ifpack preconditioner"}),
```
mayrmt, line 61: Since these documentation strings will be used elsewhere as input to the 4C documentation, let's start to be more expressive here:
```suggestion
        Core::IO::InputSpecBuilders::parameter<std::optional<std::filesystem::path>>(
            "IFPACK_XML_FILE", {.description = "This parameter describes the absolute or relative path to an xml file containing the configuration of a Trilinos/Ifpack preconditioner. The content of this xml file needs to follow Ifpack guidelines. Consult the Trilinos/Ifpack documentation and user guide for more information on valid Ifpack parameters.."}),
```

tests/input_files/beam_ifpack.xml
```diff
@@ -0,0 +1,7 @@
+<ParameterList name="Ifpack">
+  <!-- General settings for Ifpack preconditioners -->
+  <Parameter name="Preconditioner type" type="string" value="ILU"/>
+  <Parameter name="Overlap" type="int" value="1"/>
+  <!-- Level-of-fill setting for Ifpack ILU preconditioner -->
```
mayrmt, line 5: Such lines could be removed. 
```suggestion
```
I don't have a strong opinion on that, but they seem not to provide additional value.

tests/input_files/embeddedmesh_cantilever_beam.dat
```diff
@@ -36,6 +36,7 @@ DIVERCONT                             stop
 NAME                                  Struct_Solver
 SOLVER                                Belos
 SOLVER_XML_FILE                       xml/linear_solver/iterative_gmres_template.xml
+IFPACK_XML_FILE                       xml/preconditioner/ifpack.xml
```
mayrmt, line 39: This is now a weird scenario: We specify to use Belos as `SOLVER` and rely on `AZSOLVE`'s default value GMRES. We also rely on the `AZPREC`'s default value Ifpack for the preconditioner, but have to specify the `IFPACK_XML_FILE` explicitly.

I don't have a better solution for now, but I just want to point this out. This happens in other input files as well.

tests/input_files/f3_cha_xwall_ifpack.xml
```diff
@@ -0,0 +1,7 @@
+<ParameterList name="Ifpack">
+  <!-- General settings for Ifpack preconditioners -->
+  <Parameter name="Preconditioner type" type="string" value="ILU"/>
+  <Parameter name="Overlap" type="int" value="5"/>
```
mayrmt, line 4: If this is required to get the test passing, then let's do it. Just FYI: this is a huge overlap, so I'd not recommend that in day-to-day practice.

tests/input_files/homogenized-constrained-mixture-membrane-cyl-aniso-expl-nofibdir.dat
```diff
@@ -28,6 +28,7 @@ LINEAR_SOLVER                   1
 THETA                           0.5
 --------------------------------------------------------------------SOLVER 1
 SOLVER                          UMFPACK
+IFPACK_XML_FILE                 xml/preconditioner/ifpack.xml
```
mayrmt, line 31: This must not be necessary! We need to be able to construct direct solvers without having to specify a preconditioner.

tests/input_files/xml/preconditioner/ifpack.xml
```diff
@@ -0,0 +1,9 @@
+<ParameterList name="Ifpack">
+  <!-- General settings for Ifpack preconditioners -->
+  <Parameter name="Preconditioner type" type="string" value="ILU"/>
+  <Parameter name="Overlap" type="int" value="0"/>
+  <!-- Level-of-fill setting for Ifpack ILU preconditioner -->
```
mayrmt, line 5: The parameter name seems to be self-explaining, so no need for this comment. One needs to look at the Ifpack documentation anyways for further information on possible parameters.
```suggestion
```

mayrmt (changes requested):
@dharinib98 thanks for looking into all special cases! I have left some comments.

Direct solvers must not need a preconditioner. This needs to be fixed before merging.

mayrmt:
@dharinib98 I had a look at the failing framework test. It can't find the new xml file for the Ifpack preconditioner. I believe, that we also need to copy it to the build directory. The framework test macro has one input argument to define a file to be copied, however we would need to copy two files now.

@sebproell, would #641 be useful here? Do we need to adapt the framework test macro to copy a list of files? Do we need another approach?

mayrmt:
@dharinib98 You can use the new feature from #642 for the framework tests.

sebproell:
@dharinib98 Could you clean the commits a little? For instance, you have a lot of commits named "Rebase changes". I am not sure where these come from -- rebasing should not produce new commits, just rewrite old ones.

See also our CONTRIBUTING guidelines (
https://github.com/4C-multiphysics/4C/blob/main/CONTRIBUTING.md?plain=1#L43-L44)

tests/input_files/poromultielastscatra_2D_quad4_part_reaction.dat
```diff
@@ -62,18 +62,18 @@ CONVTOL                         1e-8
 --------------------------------------------------------------------SOLVER 1
 NAME                            solver
 SOLVER                          UMFPACK
+IFPACK_XML_FILE                 xml/preconditioner/ifpack.xml
```
mayrmt, line 65: Can we fix this before merging?

mayrmt (commented):
There are still many input files using UMFPACK where we now add an ifpack preconditioned.

mayrmt (dismissed):
Thanks for tackling this!

dharinib98:
I just want to point this out here. Although the explicit ifpack xml file definition has been removed from the input files, I do not think this completely resolves the dependency of UMFPACK on the ifpack preconditioner as the code is structured in such a way that the preconditioner is initialized even when the input file does not specify the ifpack xml file (more like a default behaviour). This could be the starting point for a new issue to further work on this dependency. Also another point to be noted: The following tests inspite of using UMFPACK have had Ifpack parameters defined in their respective input files. However this will not be the case after this merge.
```
fpsi_ofsiinterface.dat
fsi_ow3D_mtr_drt.dat
homogenized-constrained-mixture-cyl-aniso-impl-nofibdir.dat
scatra_transport_neumann.dat
scatra_transport_neumann_lin.dat
scatra_transport_neumann_local.dat
xff_beltrami_Div.dat
xff_beltrami_MomMom.dat
xff_beltrami_TesTes.dat
xff_beltrami_TesTes_NIT_embedded_evp.dat
xff_beltrami_TesTes_NIT_embedded_hk_ele_vol_div_by_ele_surf.dat
xff_beltrami_TesTes_NIT_embedded_hk_vol_equivalent.dat
xff_channel_simple_stat.dat
xff_shear_flow_2D_instat_initialfield_OST.dat
xff_shear_flow_2D_instat_neumann.dat
xff_shear_flow_2D_stat_bf_Nitsche_alesided_inflowaveraged_dens_7.dat
xff_shear_flow_2D_stat_neum_moredofs_circ_Nitsche_alesided.dat
xff_shear_flow_2D_stat_neum_moredofs_circ_Nitsche_alesided_inflowaveraged.dat
xff_shear_flow_2D_stat_neum_moredofs_circ_Nitsche_xsided.dat
xff_shear_flow_2D_stat_neum_moredofs_circ_Nitsche_xsidedDiv.dat
xff_shear_flow_2D_stat_neum_moredofs_circ_Nitsche_xsided_inflowstab_onesided.dat
xff_shear_flow_2D_stat_neumann.dat
xff_shear_flow_2D_stat_neumann_hex20.dat
xff_shear_flow_2D_stat_neumann_hex20_Div.dat
xff_shear_flow_2D_stat_neumann_inflowstab_onesided.dat
xff_shear_flow_2D_stat_neumann_moredofset_circle.dat
xff_shear_flow_2D_stat_neumann_moredofset_quad.dat
xff_shear_flow_stat.dat
```

sebproell:
> I think there's a bit of confusion going on related to the input parameters ... if Umfpack is used as linear solver no additional parameters need to be given, 

That's a great application for the new `InputSpecBuilders::selection` : ) (after #598 is done)

dharinib98:
If this now looks good, I can squash the commits and we can go ahead with this PR

mayrmt (approved):

PR #403, opened 2025-02-19
[epetra] Wrapper for Epetra_CrsGraph

### LINKED ISSUE #397: Create Wrapper for Epetra_CrsGraph
## Description

For the sake of centralizing all calls to `Epetra_CrsGraph` one needs to provide a containership class wrapping the respective method calls.

The class could roughly look like:
```cpp
class Core::LinAlg::Graph
{
  public:
    Graph(Epetra_DataAccess cv, const Epetra_Map& map, int nnz_per_row)
     : graph_(Epetra_CrsGraph(cv, map, nnz_per_row));

    int InsertGlobalIndices(int global_row, int num_indices, int *indices)
    {
      vector_.InsertGlobalIndices(global_row, num_indices, indices);
    }

    ...

    Epetra_CrsGraph getEpetraGraph()
    {
      return graph_;
    }

  private:
    Epetra_CrsGraph graph_;
}
```

- [ ] Provide a containership class `Core::LinAlg::Graph` for `Epetra_CrsGraph` (e.g. simple explanation given in https://www.geeksforgeeks.org/containership-in-c/)
- [ ] Verify which methods are actively used in 4C and need to be added to the 🚢 class + unit testing of these methods (see e.g. for reference https://docs.trilinos.org/dev/packages/epetra/doc/html/classEpetra__CrsGraph.html)
  - `Epetra_CrsGraph`
  - ...
- [ ] Replace occurrences of `Epetra_CrsGraph` in 4C with the 🚢 class `Core::LinAlg::Graph`, run all tests, decide where it makes sense to use `getEpetraGraph()` to return the underlying `Epetra` object to directly interface with `Trilinos` functions
- [ ] Refactor the class to use a `Epetra_FECrsGraph` as container object, adapt the class methods if necessary and replace occurrences of `Epetra_FECrsGraph` in 4C by the overhauled 🚢 class

* Related to #189 

## Interested Parties

@mayrmt @maxfirmbach

knarfnitram:
## Description and Context

Add a wrapper named `Core::LinAlg::Graph` to centralize all calls of `Epetra_CrsGraph`. 

The commits of the MR are structured in the following way:

1. Create new Interface named `Graph` 
2. Perform global search and replace of the `Epetra_CrsGraph`
3. Fix necessary places by adapting references, pointers etc.

Closes #397 .

### DIFF
--- a/src/adapter/4C_adapter_fld_fluid.hpp
+++ b/src/adapter/4C_adapter_fld_fluid.hpp
@@ -13,12 +13,12 @@
 #include "4C_inpar_fluid.hpp"
 #include "4C_inpar_poroelast.hpp"
 #include "4C_linalg_fixedsizematrix.hpp"
+#include "4C_linalg_graph.hpp"
 #include "4C_linalg_vector.hpp"
 #include "4C_utils_exceptions.hpp"
 #include "4C_utils_parameter_list.fwd.hpp"
 #include "4C_utils_result_test.hpp"
 
-#include <Epetra_CrsGraph.h>
 #include <Epetra_Map.h>
 #include <Epetra_Operator.h>
 
@@ -262,8 +262,10 @@ namespace Adapter
 
     /// provide access to the turbulence statistic manager
     virtual std::shared_ptr<FLD::TurbulenceStatisticManager> turbulence_statistic_manager() = 0;
+
     /// provide access to the box filter class for dynamic Smaorinsky model
     virtual std::shared_ptr<FLD::DynSmagFilter> dyn_smag_filter() = 0;
+
     virtual std::shared_ptr<FLD::Vreman> vreman() = 0;
 
     /// reset state vectors (needed for biofilm simulations)
@@ -419,7 +421,7 @@ namespace Adapter
     virtual double eval_time() const = 0;
 
     /// redistribute the fluid discretization and vectors according to nodegraph in std. mode
-    virtual void redistribute(const std::shared_ptr<Epetra_CrsGraph> nodegraph) = 0;
+    virtual void redistribute(const std::shared_ptr<Core::LinAlg::Graph> nodegraph) = 0;
 
 
     //@}
@@ -571,6 +573,7 @@ namespace Adapter
 
     /// return maximum for iteration steps
     virtual int itemax() const = 0;
+
     /// set maximum for iteration steps
     virtual void set_itemax(int itemax) = 0;
 
--- a/src/core/linalg/src/sparse/4C_linalg_graph.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_graph.hpp
@@ -1,0 +1,145 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_LINALG_GRAPH_HPP
+#define FOUR_C_LINALG_GRAPH_HPP
+
+
+#include "4C_config.hpp"
+
+#include <Epetra_CrsGraph.h>
+#include <Epetra_Export.h>
+#include <Epetra_FECrsGraph.h>
+
+#include <memory>
+
+
+// Do not lint the file for identifier names, since the naming of the Wrapper functions follow the
+// naming of the Epetra_CrsGraph
+
+// NOLINTBEGIN(readability-identifier-naming)
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Core::LinAlg
+{
+
+  class Graph
+  {
+   public:
+    //! Creates a Epetra_CrsGraph object and allocates storage.
+    Graph(Epetra_DataAccess CV, const Epetra_BlockMap& RowMap, const int* NumIndicesPerRow,
+        bool StaticProfile = false);
+
+    Graph(Epetra_DataAccess CV, const Epetra_BlockMap& RowMap, int NumIndicesPerRow,
+        bool StaticProfile = false);
+
+    Graph(const Graph& other);
+
+    Graph& operator=(const Graph& other);
+
+    ~Graph() = default;
+
+    /// Copy constructor from Epetra_CrsGraph to Epetra_CrsGraph
+    explicit Graph(const Epetra_CrsGraph& Source);
+
+    /// Copy constructor from Epetra_FECrsGraph to Epetra_CrsGraph
+    explicit Graph(const Epetra_FECrsGraph& Source);
+
+    //! return the reference to the Epetra_CrsGraph
+    const Epetra_CrsGraph& get_Epetra_CrsGraph() const { return *graph_; }
+
+    Epetra_CrsGraph& get_Epetra_CrsGraph() { return *graph_; }
+
+    //! Returns the Column Map associated with this graph.
+    const Epetra_BlockMap& ColMap() const { return (graph_->ColMap()); }
+
+    //! Returns a pointer to the Epetra_Comm communicator associated with this graph.
+    const Epetra_Comm& Comm() const { return (graph_->Comm()); }
+
+    //! Extract a list of elements in a specified global row of the graph. Put into storage
+    //! allocated by calling
+    int ExtractGlobalRowCopy(int GlobalRow, int LenOfIndices, int& NumIndices, int* Indices) const
+    {
+      return graph_->ExtractGlobalRowCopy(GlobalRow, LenOfIndices, NumIndices, Indices);
+    }
+
+    int ExtractGlobalRowCopy(
+        long long GlobalRow, int LenOfIndices, int& NumIndices, long long* Indices) const
+    {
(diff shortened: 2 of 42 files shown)


knarfnitram:
Seems like my pre-commit hook is not working as intended, I'll fix that and update it.

maxfirmbach:
@sebproell Have there been changes to the clang formatting style recently?

sebproell:
@maxfirmbach You mean #193?

maxfirmbach:
The blank spaces between class names etc. seem to be very weird and I don't really understand where this is coming from ... and that clang formatting isn't complaining.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
```diff
@@ -0,0 +1,163 @@
...
+    int ExtractGlobalRowCopy(int GlobalRow, int LenOfIndices, int& NumIndices, int* Indices) const
+    {
+      return graph_->ExtractGlobalRowCopy(GlobalRow, LenOfIndices, NumIndices, Indices);
+    }
+
+    int ExtractGlobalRowCopy(
+        long long GlobalRow, int LenOfIndices, int& NumIndices, long long* Indices) const
+    {
+      return graph_->ExtractGlobalRowCopy(GlobalRow, LenOfIndices, NumIndices, Indices);
+    }
```
mayrmt, line 80: Where do we operate with `long long`?

maxfirmbach: I guess we could just use `int` there to be honest.

knarfnitram: I have removed the functions containing the type`long long` if possible. For `NumGlobalIndices` seems to be only an interface of type `long`.

src/adapter/4C_adapter_fld_fluid.hpp
```diff
@@ -30,8 +30,11 @@ FOUR_C_NAMESPACE_OPEN
 namespace Core::LinAlg
 {
   class SparseMatrix;
+
```
mayrmt, line 33: Are these additional empty lines coming from our code formatting or is this an intentional change?

mayrmt (commented):
I am fine with the technical changes. The changes in white spaces seem weird, though.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+    //! Returns the number of indices in the global graph.
+    int NumGlobalNonzeros() const { return graph_->NumGlobalNonzeros(); }
+
+    //! Remove a list of elements from a specified global row of the graph.
+    int RemoveGlobalIndices(int GlobalRow, int NumIndices, int* Indices);
+
+    const Epetra_BlockMap& RowMap() const { return graph_->RowMap(); }
+
+    explicit Graph(const std::shared_ptr<Core::LinAlg::Graph>& graph);
```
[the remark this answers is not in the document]
knarfnitram: yes, that should be removed.

src/adapter/4C_adapter_fld_fluid.hpp
knarfnitram: I assume it is a combination of clion during rebasing. I have removed most of the unnecessary lines, only in a few cases where I don't consider them irritating, I have not reverted them.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
```diff
@@ -0,0 +1,153 @@
...
+#include <Epetra_CrsGraph.h>
+#include <Epetra_Export.h>
+#include <Epetra_FECrsGraph.h>
+
+#include <memory>
+
+// Do not lint the file for identifier names, since the naming of the Wrapper functions follow the
+// naming of the Epetra_Vector
+
+// NOLINTBEGIN(readability-identifier-naming)
```
amgebauer, line 23: It would be nicer if this wrapper follows our naming conventions. Can this be just removed?

src/core/linalg/src/sparse/4C_linalg_graph.hpp
```diff
@@ -0,0 +1,153 @@
...
+    int RemoveGlobalIndices(int GlobalRow, int NumIndices, int* Indices);
+
+    const Epetra_BlockMap& RowMap() const { return graph_->RowMap(); }
+
+    explicit Graph(const std::shared_ptr<Core::LinAlg::Graph>& graph);
+
+
+   private:
+    //! The actual Epetra_CrsGraph object.
+    std::shared_ptr<Epetra_CrsGraph> graph_;
```
amgebauer, line 143: Can you make  it a `uniqe_ptr`? I think we don't want to share ownership of the graph?

src/core/linalg/src/sparse/4C_linalg_graph.hpp
```diff
@@ -0,0 +1,150 @@
...
+    /// Copy constructor from Epetra_FECrsGraph to vector
+    explicit Graph(const Epetra_FECrsGraph& Source);
+
+
+    //! (Implicit) conversions: they all return references or RCPs, never copies
+    const Epetra_CrsGraph& get_ref_of_Epetra_CrsGraph() const { return *graph_; }
+
+    Epetra_CrsGraph& get_ref_of_Epetra_CrsGraph() { return *graph_; }
+
+    std::shared_ptr<Epetra_CrsGraph> get_ptr_of_Epetra_CrsGraph() { return graph_; }
```
amgebauer, line 58: This is only used in places where we directly dereference the shared pointer. You can also remove this one here and just use `get_ref_of_Epetra_CrsGraph()`

mayrmt: One could call it `get_Epetra_CrsGraph()`, since the return type does not need to be encoded into the function name.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
knarfnitram: I thought, that this would make some troubles due to Epetra naming, but it seems that it's no problem, I, therefore removed the lines.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
knarfnitram: Yes, since we are passing the reference to trilinos it seems that we can use a `unique_ptr`.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
knarfnitram: I replaced the corresponding functions, and since we only have one getter function (which returns the reference), I renamed the signature accordingly to `get_Epetra_CrsGraph()`, to simplify the name.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
sebproell: I think we added this line for the vector wrappers since the search and replace effort in that case was so high that we didn't want to bother with renaming at the same time. I am fine with taking one step at a time, keeping the `NOLINT` also for this class for now. What is important is to get the interfaces of all the distributed LinAlg building blocks under our control. Then we can gradually change it to our needs.

amgebauer: We should tackle it in a follow up though.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+    Graph& operator=(const Graph& other);
+
+    ~Graph() = default;
+
+    /// Copy constructor from Epetra_CrsGraph to vector
+    explicit Graph(const Epetra_CrsGraph& Source);
+
+    /// Copy constructor from Epetra_FECrsGraph to vector
+    explicit Graph(const Epetra_FECrsGraph& Source);
```
mayrmt, line 51: What does this comment mean? How does a vector play in here?

src/core/linalg/src/sparse/4C_linalg_graph.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+    //! return the reference to the Epetra_CrsGraph
+    const Epetra_CrsGraph& get_Epetra_CrsGraph() const { return *graph_; }
+
+    Epetra_CrsGraph& get_Epetra_CrsGraph() { return *graph_; }
+
+    // Functions
+
```
mayrmt, line 59: Either remove or make it a doxygen group.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
knarfnitram: I have opened a follow up issue https://github.com/4C-multiphysics/4C/issues/414

mayrmt (approved):

sebproell (approved):
I just now realized that you put all definitions into the header. I think for the graph class, we could move all of them into the implementation file since the functions do rather heavy work and inlining might not happen anyway. Can be changed later. I might do a small cleanup pass over the wrappers we already have and throw out the functions we don't need and can do this as well.

I am fine with merging -- this is progress! Please fix up the commits though.

PR #406, opened 2025-02-20
[NOX] Remove "Modified" parameter list

### LINKED ISSUE #405: [Trilinos] Current develop branch breaks 4C testing
The current develop branch of Trilinos breaks all tests related to the new structure time integration, specifically all test cases that use `NOX` with the following error:
```shell
terminate called after throwing an instance of 'Teuchos::Exceptions::InvalidParameterName'
what():  Error, the parameter {name="Modified",type="ParameterList",value="..."}
in the parameter (sub)list "ROOT->STRUCT NOX->Direction->Newton"
was not found in the list of valid parameters!
```

Something seems off with the parameter list entry "Modified". Looks like a special thing we are doing / implementing.

maxfirmbach:
## Description and Context

The `Modified` parameter list we internally add to NOX breaks the new structure time integration tests with the current Trilinos development branch. To my knowledge and in-depth search in the input files, these options are not used at all, neither in NOX nor in our own code?!
In addition the `Modified` list is only retrieved once in `get_number_of_modified_newton_corrections()`, but the relevant parameter is never found and thus this function always returns 0 (also based on the coverage report). With this PR I propose to just delete the parameter list and the related function, there's no benefit in keeping it.

Closes #405

### DIFF
--- a/src/structure_new/src/implicit/4C_structure_new_impl_generic.cpp
+++ b/src/structure_new/src/implicit/4C_structure_new_impl_generic.cpp
@@ -223,7 +223,7 @@ void NOX::Nln::PrePostOp::IMPLICIT::Generic::runPostIterate(const ::NOX::Solver:
 {
   double step = 0.0;
   const bool isdefaultstep = get_step(step, solver);
-  const int num_corrs = get_number_of_modified_newton_corrections(solver);
+  const int num_corrs = 0;
 
   impl_.model_eval().run_post_iterate(solver, step, isdefaultstep, num_corrs);
 }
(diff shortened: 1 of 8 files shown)


sebproell (dismissed):

src/structure_new/src/implicit/4C_structure_new_impl_generic.cpp
```diff
@@ -223,7 +223,7 @@ void NOX::Nln::PrePostOp::IMPLICIT::Generic::runPostIterate(const ::NOX::Solver:
 {
   double step = 0.0;
   const bool isdefaultstep = get_step(step, solver);
-  const int num_corrs = get_number_of_modified_newton_corrections(solver);
+  const int num_corrs = 0;
```
mayrmt, line 226: If `num_corrs` is always zero now, can it maybe removed entirely (also from `run_post_iterate()`)?

maxfirmbach: I didn't had a detailed look into all of these functions, I just replicated the behavior of the status quo without having to access the deleted parameter list.

maxfirmbach: `num_corrs` was always zero even before my changes.

mayrmt: It only has 5 occurrences throughout the code base. I'd love to see it removed, since we now know for sure, that it does not have any effect.

maxfirmbach: Done.

mayrmt (approved):

c-p-schmidt (approved):

sebproell (approved):

PR #417, opened 2025-02-21
CI: make 4C upload and download more flexible

sebproell:
With the help of @amgebauer we realized that the upload of  a 4C build performed in the actions is suboptimal as it previously used absolute paths. This will not work when the build is reused in a different environment, but it isn't very clear even in the present use case. This PR packages the build relative to the build dir and extracts it to a parameterized destination.

### DIFF
--- a/.github/actions/upload_4C_build/action.yml
+++ b/.github/actions/upload_4C_build/action.yml
@@ -13,7 +13,11 @@ runs:
   steps:
     - name: Package build directory
       shell: bash
-      run: tar -cf $GITHUB_WORKSPACE/4C_build.tar $GITHUB_WORKSPACE/build
+      run: |
+        cd $BUILD_DIRECTORY
+        tar -cf $GITHUB_WORKSPACE/4C_build.tar .
+      env:
+        BUILD_DIRECTORY: ${{ inputs.build-directory }}
     - name: Upload build folder
       uses: actions/upload-artifact@v4
       with:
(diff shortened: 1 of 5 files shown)


.github/actions/upload_4C_build/action.yml
```diff
@@ -13,7 +13,9 @@ runs:
   steps:
     - name: Package build directory
       shell: bash
-      run: tar -cf $GITHUB_WORKSPACE/4C_build.tar $GITHUB_WORKSPACE/build
+      run: |
+        cd ${{ inputs.build-directory }}
+        tar -cf $GITHUB_WORKSPACE/4C_build.tar .
```
amgebauer, line 18: Not sure, but I think you need this.
```suggestion
      run: |
        cd $BUILD_DIRECTORY
        tar -cf $GITHUB_WORKSPACE/4C_build.tar .
      env:
        BUILD_DIRECTORY: ${{ inputs.build-directory }}
```

amgebauer (commented):
-

.github/actions/upload_4C_build/action.yml
sebproell: I saw that we have this elsewhere. Tbh, I don't understand why this would change things but I believe we had some trouble with this in the past?

amgebauer (approved):

amgebauer (approved):

ppraegla (approved):

amgebauer (approved):

PR #419, opened 2025-02-22
Quote string values and descriptions in yaml

sebproell:
This will help to distinguish `yes` from `"yes"`, where the former is a bool and the latter a string in yaml.

FYI @gilrrei

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -1363,7 +1363,7 @@ void Core::IO::InputSpecBuilders::Internal::EntrySpec<T>::emit_metadata(ryml::No
 
   if (!data.description.empty())
   {
-    node["description"] << data.description;
+    emit_value_as_yaml(node["description"], data.description);
   }
   emit_value_as_yaml(node["required"], data.required.value());
   if (data.default_value.has_value())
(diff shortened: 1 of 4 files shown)


mayrmt:
Out of curiosity: Why not switch all to bool? Yes implies, that there is a no. Other values do not make sense. I don't see, why one wants to encode this as string.

gilrrei:
> Out of curiosity: Why not switch all to bool? Yes implies, that there is a no. Other values do not make sense. I don't see, why one wants to encode this as string.

The problem lies in the ENUM cases, for example `STRUCT_COUPLING_STRESS`. The choices are `No, no, NO, Yes, yes, YES, Cauchy, cauchy, 2PK, 2pk` which means the choices are a mix of bools and strings

sebproell:
@mayrmt Your reasoning is correct but as @gilrrei already pointed out, we have cases where we use strings "yes", "no", "some_other_option", ... as possible choices. See this nice list of all the problematic cases: https://github.com/4C-multiphysics/4C/issues/113#issuecomment-2612370911

If we clean this up, we can forbid all bool-like values as inputs to the `selection` InputSpec.

mayrmt:
@sebproell Do we know how many of the "yes/no/some_other_value" exist? If it's a small number, we could clean this up right away.

sebproell:
>  Do we know how many of the "yes/no/some_other_value" exist?

As I said above, @gilrrei compiled this list https://github.com/4C-multiphysics/4C/issues/113#issuecomment-2612370911. I hope that's all of them. I will transfer that comment into a new issue. I am all for cleaning that up but, just to be clear,  this PR is useful in its own right, regardless of whether we clean this up.

gilrrei (approved):

amgebauer (approved):

PR #421, opened 2025-02-22
Allow to read yaml files with --- document markers

sebproell:
Yaml files can start with `---` to mark the beginning of a document. This didn't work before. The absurd thing is that we cannot test it since our formatted `yamlfmt` either wants to insert the `---` marker everywhere or nowhere. There is no drawback to inserting it in all our yaml files but it feels wrong to do that for the purpose of having a few test files. Opinions?

### DIFF
--- a/src/core/io/src/4C_io_input_file.cpp
+++ b/src/core/io/src/4C_io_input_file.cpp
@@ -460,14 +460,32 @@ namespace Core::IO
       std::ifstream file(file_path);
       std::ostringstream ss;
       ss << file.rdbuf();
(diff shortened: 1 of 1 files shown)


gilrrei (approved):

gilrrei:
The fact that it doesn't work with yamlfmt is crazy to me, it's such a base feature... We could create a python script that reads in the files and dumps them again. This should not be difficult to do, some libraries even preserve comments

davidrudlstorfer (approved):
In my opinion we should leave the marker out of our test files

Issue #423, opened 2025-02-22
Input parameter `MASSLIN` in `STRCUTURAL DYNAMIC` with unclear possible values


mayrmt:
## Description

The input parameter `MASSLIN` in `STRCUTURAL DYNAMIC` has three options: no, standard, rotations

The option "standard" is never tested. It also does not occur alone in code, but always in combination with "rotations". One could think about removing the "standard" option.

## Possible Solution and Definition of Done

- [ ] Understand the difference between "standard" and "rotations"
- [ ] Either test or remove "standard"

davidrudlstorfer:
I do not know the specifics of the `MASSLIN` option - but it only makes sense when using the `GenAlphaLieGroup` time integration (that's what I have been told)

Maybe @isteinbrecher knows more about it regarding its influence within the beam framework and specifically with the time integration scheme

amgebauer:
I think `standard` is related to https://gitlab.lrz.de/baci/baci/-/merge_requests/2384 and can be removed.

mayrmt:
@amgebauer, thanks! That clarifies the picture.

There are two cases where we needed it in the past:

- rotational degrees for freedom (cf. `GenAlphaLieGroup`) --> we still need this for beams
- AC framework (removed) where the mass has changed due to growth phenomena and, thus, the mass matrix needed to be re-computed. This one did not use rotational DOFs.

Since the AC framework is gone, the `standard` option can be removed.

vryy:
I believe `MASSLIN` as ` standard` refers to nonlinear mass as the lines below

https://github.com/4C-multiphysics/4C/blob/905ce667039f4fa83850365d056bea1797e16574/src/structure/4C_structure_timint.cpp#L366-L377

Although I never tried nonlinear mass in structural problem myself, but there is case of nonlinear mass for porous media related problem, i.e., `poroelast`. I would suggest to check how `MASSLIN` is handled there, although I believe it is `standard` by default.

There's even a MR in the old repo to enable `MASSLIN standard` for `structure_new`, although now's in stale mode

https://gitlab.lrz.de/baci/baci/-/merge_requests/1431

Issue #422, opened 2025-02-22
Input parameter `CONTACT_REGULARIZATION` seems to be untested


mayrmt:
## Description

The input parameter `CONTACT_REGULARIZATION` in `CONTAT DYNAMICS` has two possible values, `no` and `tanh`. It looks like the `tanh` option is untested.

## Possible Solution and Definition of Done

- [x] Understand the meaning of this parameter
- [x] Either test the `tanh` option or remove the parameter and related code entirely

Issue #420, opened 2025-02-22
[Input] Do not allow selection choices that are a mix of bool and string


sebproell:
> Currently, bool parameters of the `Parameterlist` are exported as
> ```yaml
>   NURBS:
>     DO_LS_DBC_PROJECTION:
>       type: string
>       default: No
>       description: "Determines if a projection is needed for least square Dirichlet boundary conditions."
>       valid options:
>         - Yes
>         - No
>         - yes
>         - no
>         - YES
>         - NO
> ```
> however, for meaningful validation they should be exported as:
> ```yaml
>   NURBS:
>     DO_LS_DBC_PROJECTION:
>       type: boolean
>       default: false
>       description: "Determines if a projection is needed for least square Dirichlet boundary conditions."
> ```
> otherwise the validation will only be happy if you use `DO_LS_DBC_PROJECTION: "yes"` (or similar). 
> 
> For the older yaml standard [1.1](https://yaml.org/type/bool.html) the booleans are:
> * for true: `y, Y, yes, Yes, YES, true, True, TRUE,on, On, ON,`
> * for false `n, N, no, No, NO, false, False, FALSE, off, Off, OFF`
> 
> For the new yaml standard [1.2](https://yaml.org/spec/1.2.2/#10212-boolean), only `true` and `false` are used. I propose to stick to the new one as it reduce complexity. *Note*: this is currently no problem for the reading and writing of 4C input files, but to create the schema, additional information/code is required.
> 
> *Note*: There are some cases where the valid options are a mix of bools and strings in the yaml sense. Here are all the parameters I could find.
> 
> | section                                          | parameter                | valid options                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
> |--------------------------------------------------|--------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
> | ALE DYNAMIC                                      | MESHTYING                | **no**, meshtying, meshsliding                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
> | CONTACT DYNAMIC                                  | CONTACT_REGULARIZATION   | **no**, tanh                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
[...]

PR #424, opened 2025-02-22
Convert some `yes`/`no` input values to non-boolean strings

mayrmt:
⚠️ BREAKING CHANGE in input parameters

## Description and Context

In the light of #420, this PR ramens some input options to strings that avoid confusion with boolean sounding expressions `yes` and `no`.

Details:

- `BEAMS_DAMPING`: has been converted to an actual boolean input parameter
- `STC_SCALING`: to deactivate it, use `Inactive` (instead of `no`) 
- `MASSLIN`: to deactivate it, use `none` (instead of `no`)
- `DAMPING` in solid dynamics can now either be `None`, `Rayleigh` or `Material`. The old "yes" has used Rayleigh damping under the hood, so it has been removed and must be called `Rayleigh` to be explicit about it.
- `CONTACT_REGULARIZATION`: to deactivate it, use `none` (instead of `no`)

- Arose from #420 
- Triggered the opening of #422 #423


sebproell:
@mayrmt Regarding the use of `none`: We have `Noneable<T>` as a wrapper around type T that allows to take a special "none" value. In the code, this works like `std::optional<T>` (it's that type under the hood). I've discussed this with @lauraengelhardt and @amgebauer and we are not 100% sure how we want to handle things. I just wanted to let you know that `none` (in dat) and `null`(in yaml) are special keywords that explicitly mark the absence of a parameter. Thus, it might make sense to *only* allow the "none" value if the parameter is indeed `Noneable`. We can proceed with this MR, as the code would need a cleanup pass in that respect anyway, but I wanted to get the thought out.

mayrmt:
@sebproell This is the only remaining error:

```
Start 2813: tutorial_fsi_3d-p2-fw
201/288 Test #2813: tutorial_fsi_3d-p2-fw .....................................................................................................................................***Failed  Error regular expression found in output. Regex=[ ERROR ]  0.75 sec
File /__w/4C/4C/tests/framework-test/tutorial_fsi_3d.e was created with EXODUS II library version 4.98
Mesh consists of 2772 Nodes, 1280 Elements, organized in 
2 ElementBlocks, 39 NodeSets, 0 SideSets 

Your 63 definitions do not match the 41 entities in your mesh!
(This is OK, if more than one BC is applied to an entity, e.g in FSI simulations)
creating and checking 4C input file       --> framework_test_output/tutorial_fsi_3d/xxx.dat
...Ensure positive element jacobians        in....0.0186117 secs
...Writing dat-file                         in....0.0541864 secs
...Read parameters


=========================================================================
PROC 0 ERROR in /__w/4C/4C/src/core/io/src/4C_io_input_file.cpp, line 922:
Error message:
  Error while parsing section 'STRUCTURAL DYNAMIC': PROC 0 ERROR in /__w/4C/4C/src/core/io/src/4C_io_input_spec_builders.hpp, line 1473:
Error message:
  Invalid value 'None'
------------------

------------------
 0# FourC::Core::IO::InputFile::match_section(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, FourC::Core::IO::InputParameterContainer&) const [clone .cold] in /__w/4C/4C/build/lib4C.so
 1# FourC::Core::IO::read_parameters_in_section(FourC::Core::IO::InputFile&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Teuchos::ParameterList&) in /__w/4C/4C/build/lib4C.so
 2# FourC::Global::read_parameter(FourC::Global::Problem&, FourC::Core::IO::InputFile&) in /__w/4C/4C/build/lib4C.so
 3# FourC::EXODUS::validate_input_file(ompi_communicator_t*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) in ./pre_exodus
 4# main in ./pre_exodus
 5# 0x00007F0D642A01CA in /lib/x86_64-linux-gnu/libc.so.6
 6# __libc_start_main in /lib/x86_64-linux-gnu/libc.so.6
 7# _start in ./pre_exodus

=========================================================================


--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------
```

The failing test is special, since it executes `pre_exodus` to build a dat file.

I am confident, that the line in the `head`-file is correct. Do you have an idea what is going wrong here?

src/inpar/4C_inpar_structure.cpp
```diff
@@ -189,12 +187,10 @@ namespace Inpar
...
           "LOADLIN", false, "Use linearization of external follower load in Newton", sdyn);
 
-      Core::Utils::string_to_integral_parameter<Solid::MassLin>("MASSLIN", "No",
+      Core::Utils::string_to_integral_parameter<Solid::MassLin>("MASSLIN", "none",
           "Application of nonlinear inertia terms",
-          tuple<std::string>("No", "no", "Standard", "standard", "Rotations", "rotations"),
-          tuple<Solid::MassLin>(
-              ml_none, ml_none, ml_standard, ml_standard, ml_rotations, ml_rotations),
-          sdyn);
+          tuple<std::string>("none", "standard", "rotations"),
```
sebproell, line 192: @mayrmt Compare this....

tests/framework-test/tutorial_fsi_3d.head
```diff
@@ -222,7 +222,7 @@ MATERIALTANGENT                 analytical
 //
 // Application of nonlinear inertia terms
 //     No,no,Standard,standard,Rotations,rotations
-MASSLIN                         No
+MASSLIN                         None
```
sebproell, line 225: ... to this. This should be the error. Anyways, the error can be better here. Fix coming up.

mayrmt: Thanks for the helping hand! I will fix the capitalization.

c-p-schmidt:
@mayrmt: Thank you for the initiative to clean the input. However, one question arose when I looked through this pull request. Shouldn't we aim for a consistent style of the input parameters?
At some locations, you, e.g., used `none`; at others, `None.` Also in #426 the parameters were capitalized, i.e. `None`.

mayrmt:
@c-p-schmidt This might be interesting in the long run. This PR partially addresses #420, nothing else. 

At the moment, we do not even have a policy on how to format input parameters. Currently, inconsistencies are much worse than just the question of capitalization. For example, many fields use different input parameter keys for the same quantity. Even if the capitalization in the values matches everywhere, there is no common touch and feel across the entire input file.

sebproell (approved):

PR #427, opened 2025-02-23
Make bool params take a bool default value

sebproell:
@mayrmt I was too slow to push this; now you pushed #426, #425, #424. But I think we have almost no overlap :crossed_fingers: 

- First commit fixes "NONLINEARBC", "STRONG_REDD_3D_COUPLING_TYPE", and "PRESSAVGBC". 
- Second commit moves the signature of `bool_parameter` to take a bool value. 
- ~Third commit is an alternative to a commit in #424 as it changes `BEAMS_DAMPING` to a bool.~ Dropped

Part of #420


sebproell:
@mayrmt I'll drop the third commit.

c-p-schmidt (approved):

mayrmt (approved):

Issue #182, opened 2025-01-03
[Mat] Create nodal material vector


maxfirmbach:
## Description

For strongly anisotropic and heterogeneous materials it can make sense to assemble a nodal material vector and use it as additional information in the linear solver. For a very simple material with just one parameter the calculation routine could look like the following:
1. Create empty material vector based on the noderowmap of the discretization
2. Loop over nodes
3. Get the associated element
4. Retrieve material information
5. Average material property into the nodal vector

I mainly have to questions:
1. Where to place this feature? It seems like I can't put it somewhere into `core` as this would create a dependency on the material files.
2. How to implement step 4? To me it seems like I need a big switch and cast to a specific material, with the cases being the material that support the material vector feature.

## Possible Solution and Definition of Done

- [ ] Discuss how to realize the implementation and resolve open questions.
- [ ] Implement the calculation routine for a nodal material vector.
- [ ] Create appropriate unit- and framework-tests.

## Interested Parties

@mayrmt @sebproell

sebproell:
> Where to place this feature? It seems like I can't put it somewhere into core as this would create a dependency on the material files.

I agree that `core` is not the right place as this feature is too specific to be there. 

> How to implement step 4? To me it seems like I need a big switch and cast to a specific material, with the cases being the material that support the material vector feature.

You can use `Discretization::evaluate` taking a function (this one: https://github.com/4C-multiphysics/4C/blob/main/src/core/fem/src/discretization/4C_fem_discretization.hpp#L1705-L1709). You pass it a function that does whatever you need to do to recover the material values. This function works on elements instead of nodes, but this is no problem, as you need to call the material per element and distribute your desired results to all element nodes (performing averaging if you like).

PR #428, opened 2025-02-23
[discretization] Add method to average element values into adjacent nodes

maxfirmbach:
## Description and Context

This PR adds a little method enabling to average element based values into adjacent nodes, by using a respective overlapping (column) element map vector as input (e.g. evaluated on the discretization beforehand) and outputs a node (row) map vector containing the averaged values.

There are two open points from my side:
1. I'm not sure if a similar method already exists ... at least I haven't found one.
2. I wasn't sure where to put this method, the current place doesn't feel right, but all others I thought about aren't better either.

Part of #182


maxfirmbach:
Not sure who else comes into question for review, it's always the same people anyway.

src/core/fem/src/discretization/4C_fem_discretization_utils.cpp
```diff
@@ -121,4 +121,35 @@ void Core::FE::Utils::do_initial_field(const Core::Utils::FunctionManager& funct
   }
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+Core::LinAlg::MultiVector<double> Core::FE::Utils::average_element_to_nodal_values(
+    const Discretization& discretization,
+    const Core::LinAlg::MultiVector<double>& element_multi_vector)
```
mayrmt, line 128: Data is const. And the type can be seen from the declaration and must not be part of the argument's name.
```suggestion
    const Core::LinAlg::MultiVector<const double>& element_data)
```

src/core/fem/src/discretization/4C_fem_discretization_utils.cpp
```diff
@@ -121,4 +121,35 @@ void Core::FE::Utils::do_initial_field(const Core::Utils::FunctionManager& funct
...
+{
+  Core::LinAlg::MultiVector<double> nodal_material_vector = Core::LinAlg::MultiVector<double>(
+      *discretization.node_row_map(), element_multi_vector.NumVectors(), true);
```
mayrmt, line 131: I know, that you will use this to compute a "nodal material vector", but this generic function might be used for other things. Maybe just `nodal_vector`?
```suggestion
  Core::LinAlg::MultiVector<const double> nodal_vector = Core::LinAlg::MultiVector<double>(
      *discretization.node_row_map(), element_multi_vector.NumVectors(), true);
```
or `nodally_averged_vector`?

src/core/fem/src/discretization/4C_fem_discretization_utils.cpp
```diff
@@ -121,4 +121,35 @@ void Core::FE::Utils::do_initial_field(const Core::Utils::FunctionManager& funct
...
+      Core::LinAlg::Vector<double> element_vector = element_multi_vector(col);
+      double nodal_value = 0.0;
+      for (int k = 0; k < num_elements; k++)
+      {
+        const int global_element_id = adjacent_elements[k]->id();
+        const int local_element_id = element_vector.Map().LID(global_element_id);
+        nodal_value = nodal_value + element_vector[local_element_id];
+      }
+      // replace by average value
+      nodal_material_vector.ReplaceGlobalValue(node->id(), col, nodal_value / num_elements);
```
mayrmt, line 148: Can't you do this locally, if you know the local id of the node?

src/core/fem/src/discretization/4C_fem_discretization_utils.cpp
```diff
@@ -121,4 +121,35 @@ void Core::FE::Utils::do_initial_field(const Core::Utils::FunctionManager& funct
   }
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+Core::LinAlg::MultiVector<double> Core::FE::Utils::average_element_to_nodal_values(
+    const Discretization& discretization,
+    const Core::LinAlg::MultiVector<double>& element_multi_vector)
+{
```
sebproell, line 129: At the start you probably want to check some preconditions: are the vector sizes consistent with data in the discretization? Is the discretization even filled? etc.

sebproell (commented):
I am not sure about this function. It feels a little too special to put into core. It might make more sense to just put this into the implementation file where you need it. If you belive it is really that useful in a general case, we can put it in core, but then I would like some better docs and asserts.

src/core/fem/src/discretization/4C_fem_discretization_utils.cpp
maxfirmbach: Yes, should be possible to also do locally.

maxfirmbach:
> I am not sure about this function. It feels a little too special to put into core. It might make more sense to just put this into the implementation file where you need it. If you belive it is really that useful in a general case, we can put it in core, but then I would like some better docs and asserts.

I thought so too, that‘s why I raised the question 😅 Might really be just better placed in the application/physics itself.

isteinbrecher (commented):
Thanks for the contribution @maxfirmbach , I left some comments.

Maybe a general question, do we need this implementation? I assume you implement this to port the rest of the thermo output. First of all, I don't think the output of cell based data on nodes is a good practice, second, if you really want this, you can simply output the cell based data and apply a vtk filter `CellToPointData` in post processing.

src/core/fem/src/discretization/4C_fem_discretization_utils.cpp
isteinbrecher: I think this is still a processor local operation, only with global IDs instead of local ones.

maxfirmbach:
> Thanks for the contribution @maxfirmbach , I left some comments.
> 
> Maybe a general question, do we need this implementation? I assume you implement this to port the rest of the thermo output. First of all, I don't think the output of cell based data on nodes is a good practice, second, if you really want this, you can simply output the cell based data and apply a vtk filter `CellToPointData` in post processing.

@isteinbrecher Yes I need this function for internal calculations, which are node/dof based. It's not related to output in any way, so doing `paraview` filters is not an option.

maxfirmbach:
I will rebase this after #432 is done, with the respective setup of thermo unit testing.

src/core/fem/src/discretization/4C_fem_discretization_utils.cpp
maxfirmbach: That's right. I can't put things off-proc anyway.

mayrmt: Local indexing avoids the index lookup, that global indexing performs under the hood. That's the entire reason why I made this suggestion.

isteinbrecher: Since we start of with the global `node->id()` we have to to the lookup one way or the other, but maybe I am missing something.

maxfirmbach: Yes, the one from the column map, which I don‘t want in this case.

Issue #310, opened 2025-01-31
[thermo] Element name doesn't appear in element definitions


maxfirmbach:
## Description
I was trying to setup a unit-test with a small thermo element based discretization.
I'm trying to call `element_lines(inputData.elementtype_, inputData.distype_);` with `inputData.elementtype_ = "THERMO"`, which results in an instant error that this type of element definition is not in the definitions list.
In `/src/core/fem/src/general/element/4C_fem_general_element_definition.cpp`, line 147:
```shell
 The following test failed:
   definitions_.contains(name)
 Error message:
   Element type not found: THERMO

```
I printed the respective data structure and it seems like all element types are in there besides "THERMO" ...

I'm setting things up like this:
```cpp
  void create_material_in_global_problem()
  {
    Core::IO::InputParameterContainer mat_fourier;
    mat_fourier.add("CAPA", 420.0);
    mat_fourier.add("CONDUCT_PARA_NUM", 3);
    mat_fourier.add("CONDUCT", std::vector<double>{1.0, 1.0, 1.0});

    Global::Problem::instance()->materials()->insert(
        1, Mat::make_parameter(1, Core::Materials::MaterialType::m_thermo_fourier, mat_fourier));
  }

  // Serial discretization tests
  class ElementToNodalValuesTest : public testing::Test
  {
   public:
    ElementToNodalValuesTest()
    {
      create_material_in_global_problem();

      comm_ = MPI_COMM_WORLD;
      test_discretization_ = std::make_shared<Core::FE::Discretization>("thermo", comm_, 3);

      Core::IO::cout.setup(false, false, false, Core::IO::standard, comm_, 0, 0, "dummyFilePrefix");

      // results in 27 nodes
      inputData_.bottom_corner_point_ = std::array<double, 3>{0.0, 0.0, 0.0};
      inputData_.top_corner_point_ = std::array<double, 3>{1.0, 1.0, 1.0};
      inputData_.interval_ = std::array<int, 3>{2, 2, 2};
      inputData_.node_gid_of_first_new_node_ = 0;

      inputData_.elementtype_ = "THERMO";
      inputData_.distype_ = "HEX8";
      inputData_.elearguments_ = "MAT 1";

      Core::IO::GridGenerator::create_rectangular_cuboid_discretization(
          *test_discretization_, inputData_, true);

      test_discretization_->fill_complete(false, false, false);
    }

    void TearDown() override { Core::IO::cout.close(); }

   protected:
[...]

PR #432, opened 2025-02-24
[thermo] Add discretization unit test

maxfirmbach:
## Description and Context

This PR adds two small unit tests are added to guarantee a correct setup of a thermo discretization in serial and parallel to be able to include thermo based unit tests in the next PR's.

Part of #310

### DIFF
--- a/src/thermo/tests/4C_thermo_discretization_test.cpp
+++ b/src/thermo/tests/4C_thermo_discretization_test.cpp
@@ -1,0 +1,97 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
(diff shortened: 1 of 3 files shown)


src/thermo/src/element/4C_thermo_element.hpp
```diff
@@ -27,571 +27,573 @@ namespace Core::FE
   class Discretization;
 }  // namespace Core::FE
 
-namespace Thermo
+namespace Discret
```
sebproell, line 30: Isn't `Thermo` the namespace we want in the future? `Discret::Elements` is conceptually not a modular namespace

sebproell (commented):
Do we gain something from this test? It looks like code path is covered by the input file tests. I am not sure which "units" are tested here.

maxfirmbach:
@sebproell This would mainly serve as entry point for further `thermo` unit tests, but I can also add this later on with the addition of the actual unit testing -> closing.

PR #433, opened 2025-02-24
Fix MPI_COMM nullptr

ppraegla:
## Description and Context
The 4C build fails when compiling with intel-mpi. This PR fixes the build errors.
- In some places the MPI_COMM was initialized with or assigned a `nullptr`. This works for openmpi as MPI_COMM is a pointer. But for intel-mpi MPI_COMM is an int. Therefore, the nullptr is replaced by the implementation-independent `MPI_COMM_NULL`.
- With MPI_COMM as int, there were two constructors with the same argument types. I removed one of them. The removed constructor was only used in tests.
- I also fixed a bug where only one process would hit an MPI barrier.


sebproell (dismissed):

mayrmt (approved):

ppraegla:
I will look into the failing tests. On one proc `levelset_sv_elliptic.dat` runs through successfully. But on 3 procs it hangs. So, I might have done something wrong with the MPI barrier.

mayrmt (approved):

sebproell (approved):

PR #435, opened 2025-02-24
Enable all sections in JSON schema

gilrrei:
## Description and Context
This PR enables JSON schemas for all sections
- The python YAML library is changed to ruamel.yaml to allow for YAML 1.2
- The metadata functionality is split into two parts `metadata_utils.py` (creates ptython objects, will also be used for documentation etc) and the script that actually creates the schema
- I used inheritance to keep the dataclasses well interfaced (@sebproell don't worry, in python they are chill)
- Compared to the previous iteration, the script is much more complex. This is due to the fact that all sections are supported now, allowing for nesting information
- This change requires python 3.10, shouldn't be a problem since ubuntu 24.04 will ship with 3.12

To test the schema file, I converted all the .dat input test files to YAML and checked if they were valid. The validation was successful for all the files, except the ones using NURBS with `something KNOTVECTORS` since
```yaml
STRUCTURE KNOTVECTORS:
  - NURBS_DIMENSION 3
  - BEGIN NURBSPATCH
  - ID 1
  - NUMKNOTS 8
  - DEGREE 2
  - TYPE Interpolated
  - +0.000000000000
  - +0.000000000000
  - +0.000000000000
 ```
the last three entries are not strings as expected (can also be seen by the syntax highlighting on gh). Here a valid approach looks like this:
 ```yaml
STRUCTURE KNOTVECTORS:
  - NURBS_DIMENSION 3
  - BEGIN NURBSPATCH
  - ID 1
  - NUMKNOTS 8
  - DEGREE 2
  - TYPE Interpolated
  - "+0.000000000000"
  - "+0.000000000000"
  - "+0.000000000000"
 ```

Closes #319

### DIFF
--- a/utilities/create_json_schema.py
+++ b/utilities/create_json_schema.py
@@ -5,169 +5,592 @@
 #
 # SPDX-License-Identifier: LGPL-3.0-or-later
 
-"""Create json schema from 4C yaml file."""
-import yaml
-import warnings
+"""Create JSON schema for YAML validation."""
+
 import json
 import argparse
+from pathlib import Path
 
-FOURC_TO_JSON_SCHEMA_DICT = {
-    "double": "number",
-    "int": "integer",
-    "bool": "boolean",
-    "string": "string",
-}
+from metadata_utils import (
+    NotSet,
+    NAMED_TYPES,
+    MISSING_DESCRIPTION,
+    FOURC_BASE_TYPES_TO_JSON_SCHEMA_DICT,
+    Primitive,
+    Vector,
+    Map,
+    Selection,
+    Group,
+    List,
+    All_Of,
+    One_Of,
+    metadata_object_from_file,
+)
+from dataclasses import asdict
 
 
-def map_property_to_json_schema(
-    properties_original: dict,
-    key_original: str,
-    properties_schema: dict,
-    key_schema: str,
-    filter=lambda x: x,
-    optional: bool = False,
+def json_schema(
+    *,
+    schema_type="object",
+    title=None,
+    description=MISSING_DESCRIPTION,
+    noneable=False,
+    default=NotSet(),
+    enum=None,
+    const=NotSet(),
+    properties=None,
+    required=None,
+    oneOf=None,
+    minProperties=None,
+    maxProperties=None,
+    patternProperties=None,
+    additionalProperties=False,
+    items=None,
+    minItems=None,
+    maxItems=None,
 ):
-    value = properties_original.pop(key_original, None)
-    if value is None:
-        if not optional:
-            raise ValueError(f"Value {key_original} was not provided!")
-    else:
-        properties_schema[key_schema] = filter(value)
+    """Create JSON schema dict."""
+
+    def set_if_not_none(schema, name, value):
+        if value is not None:
+            schema[name] = value
+
+    def set_if_true(schema, name, value):
+        if value:
+            schema[name] = value
+
+    def set_if_set(schema, name, value):
+        if not isinstance(value, NotSet):
+            schema[name] = value
+
+    schema = {}
+
+    # Properties for any kind of schema
+    set_if_not_none(schema, "title", title)
+    schema["description"] = description or MISSING_DESCRIPTION
+    set_if_not_none(schema, "type", schema_type)
+    if schema_type:
+        if noneable:
+            schema["type"] = [schema_type, "null"]
+        else:
+            schema["type"] = schema_type
+    set_if_true(schema, "enum", enum)
+
+    # Allow None since it could be a desired value
+    set_if_set(schema, "const", const)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: utilities/create_json_schema.py [hunk 1]
```diff
@@ -5,169 +5,592 @@
...
+            parameter.specs[i] = add_named_specs(spec, named_specs)
+        return parameter
 
+    raise TypeError(f"Unknown metadata type {type(parameter)}: {parameter}")
 
-def load_fourc_yaml_and_escape_bools(fourc_yaml_path: str) -> dict:
-    """Load 4C yaml file.
 
-    Since in 4C currently bools are not supported in the inputs, they need some special treatment to avoid converting them to bool while loading. Once this is supported in the code, this functionally can be dropped.
+def condense_all_of(all_of):
```
gilrrei, line 496: This function is required as JSON schemas require a different approach to all ofs and one ofs compared to the metadata file. Also the approach I chose requires to copy the named parameters to the oneOf options which requires this recursive approach.

utilities/create_json_schema.py
```diff
@@ -5,169 +5,592 @@
...
+
+        if new_all_ofs:
+            raise TypeError("No new all_ofs possible!")
+
+    # No all_of left
+    all_ofs = None
+
+    if not one_ofs:
+        raise TypeError(
+            "New one_ofs have to be introduced other this function would have exited earlier"
```
sebproell, line 545: I don't get what you mean here.

sebproell (commented):
This is amazing :star_struck:

utilities/create_json_schema.py
gilrrei: Cryptic, I'll change it. I meant that there have to be one_ofs if this stage of the function is reached. Otherwise, other conditions should have been met, and the function would have returned something different. I tried to add as many errors as possible since debugging this recursive and nested code is not a good time.

sebproell: OK, so this is more of an `assert` statement. And I fully agree to throw in one of these whenever you make an assumption :+1:

sebproell (dismissed):
:rocket:

utilities/metadata_utils.py
```diff
@@ -0,0 +1,334 @@
...
+    def __post_init__(self):
+        self.type = self._type
+        for i, spec in enumerate(self.specs):
+            if isinstance(spec, dict):
+                spec = _metadata_object_from_dict(spec)
+                self.specs[i] = spec
+        try:
+            self.specs = sorted(self.specs, key=lambda x: x.name)
+        except:
+            breakpoint()
```
amgebauer, line 110: This is a leftover. Remove the try-except?

utilities/create_json_schema.py
```diff
@@ -5,169 +5,592 @@
...
+        metadata_path (str, pathlib.Path): 4C emitted metadata file path
+        json_schema_path (str, pathlib.Path): JSON schema file path
+    """
+    metadata_4C, title_section, code_metadata = metadata_object_from_file(metadata_path)
+    schema = get_schema(metadata_4C)
+    schema["properties"].update(
+        {title_section: schema_for_description_section(title_section)}
+    )
+    schema["patternProperties"] = {
+        "^FUNCT[1-9][0-9]*$": schema["properties"].pop("FUNCT<n>")
```
amgebauer, line 591: Nice, not allowing FUNCT0 👍🏼

amgebauer (commented):
Thanks for this nice contribution!

As discussed offline, we need to make sure in cmake that the python script is only called for python 3.10 or above, I guess. Is this possible? @sebproell

sebproell:
> As discussed offline, we need to make sure in cmake that the python script is only called for python 3.10 or above, I guess. Is this possible?

Yes. We just need to check the `Python_VERSION` in cmake here:

https://github.com/4C-multiphysics/4C/blob/main/apps/global_full/CMakeLists.txt#L44

sebproell:
On second thought, is it absolutely necessary to have python 3.10 here? My python is only 3.8. We should make it as easy as possible to get access to the schema for all our users.

utilities/metadata_utils.py
gilrrei: Thanks, fixed

gilrrei:
> On second thought, is it absolutely necessary to have python 3.10 here? My python is only 3.8. We should make it as easy as possible to get access to the schema for all our users.

Python 3.10 is necessary because of pattern matching. In theory, this could be changed to if-else, but Python 3.8 is already in eol.

amgebauer (approved):

sebproell:
FYI, if you want 4C to use a newer python version you can set the following in your CMake preset:

```
"Python3_ROOT" : "path/to/dir/containing/bin/python"
```

Important: the path should _not_ contain the `bin/python` itself.

Issue #437, opened 2025-02-25
[InputSpec] Remove .required field and use std::optional


sebproell:
## Description

`InputSpecBuilders::parameter` and `selection` take an option `.required`. This options exists for historical reasons to enable the initial transformation of input parameters that existed in 4C to the new mechanism. It should be removed and replaced with the `Noneable` type wrapper. When the `.required` filed is gone, there is no more confusion with `std::optional` and `Noneable` can simply become that type.

## Possible Solution and Definition of Done
- [x] Remove the 286 occurrences of `.required = false`
- [x] Remove the field from InputSpec and simplify implementation
- [x] Replace `Noneable` with `std::optional`

## Interested Parties

@lauraengelhardt @gilrrei

PR #441, opened 2025-02-25
Replace most parameters with .required = false with better constructs

sebproell:
This PR replaces most occurrences of `.required = false` by:
- removing outdated paramters
- making parameters required, or
- making parameters `Noneable`

I need to discuss a few remaining occurrences with a few people. So this is only the first part but can be merged already.

Part of #437

### DIFF
--- a/src/mat/4C_mat_anisotropy.cpp
+++ b/src/mat/4C_mat_anisotropy.cpp
@@ -86,10 +86,19 @@ void Mat::Anisotropy::set_number_of_gauss_points(int numgp) { numgp_ = numgp; }
 void Mat::Anisotropy::read_anisotropy_from_element(
     const Core::IO::InputParameterContainer& container)
 {
+  // This method might be called even though there is no anisotropy possible for the element.
+  // In this case, we just return without doing anything.
+  using OptionalFiber = Core::IO::Noneable<std::vector<double>>;
+  const bool any_fibers_found =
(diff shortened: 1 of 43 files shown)


src/mat/4C_mat_anisotropy.cpp
```diff
@@ -86,10 +86,19 @@ void Mat::Anisotropy::set_number_of_gauss_points(int numgp) { numgp_ = numgp; }
 void Mat::Anisotropy::read_anisotropy_from_element(
     const Core::IO::InputParameterContainer& container)
 {
+  // This method might be called even though there is no anisotropy possible for the element.
+  // In this case, we just return without doing anything.
+  using OptionalFiber = Core::IO::Noneable<std::vector<double>>;
+  const bool any_fibers_found =
+      container.get_if<OptionalFiber>("FIBER1") or container.get_if<OptionalFiber>("RAD") or
+      container.get_if<OptionalFiber>("AXI") or container.get_if<OptionalFiber>("CIR");
+  if (!any_fibers_found) return;
```
sebproell, line 95: @lauraengelhardt @amgebauer The fiber stuff is kind of crazy since functions that read fibers are called for elements that can never have them. I cannot fix all of that at once, so I opted for this temporary solution. At least it makes some convoluted call chains more explicit.

lauraengelhardt (approved):

PR #442, opened 2025-02-25
Remove CONTACT_REGULARIZATION

### LINKED ISSUE #422: Input parameter `CONTACT_REGULARIZATION` seems to be untested

mayrmt:
## Description and Context

This feature has been added 10 years ago. Today, it is

- unclear what it is supposed to do,
- untested,
- broken with an overflow error.

We have been doing robust contact simulations without for several years, so apparently it is not super important.

All in all, there seems not to be much value in keeping it, so it is removed.

- I looked into this parameter due to #420.
- Closes #422

### DIFF
--- a/src/contact/4C_contact_abstract_data_container.cpp
+++ b/src/contact/4C_contact_abstract_data_container.cpp
@@ -66,7 +66,6 @@ CONTACT::AbstractStrategyDataContainer::AbstractStrategyDataContainer()
       isselfcontact_(false),
       friction_(false),
       non_smooth_contact_(false),
-      regularized_(false),
       dualquadslavetrafo_(false),
       trafo_(nullptr),
       invtrafo_(nullptr),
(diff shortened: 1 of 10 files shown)


mayrmt:
This hasn't been tested and it is broken, so I do not expect it to break any input files. So, I refrain from adding the `breaking change` label.

sebproell (approved):
So, this closes #422?

mayrmt:
@sebproell, yes, it does.

c-p-schmidt (approved):

PR #447, opened 2025-02-26
Improve input of four conditions (spring dashpot, impendance, volumetric flow, traction correction)

lauraengelhardt:
## Description and Context

This PR improves the input for 4 conditions:
- Make functions noneable for spring dashpot conditions and impendance condition.
- First, rename the parameters for the volumetric flow and total traction correction condition. Second, if `NORMAL` and `CENTEROFMASS` is set to `self_evaluate`, the user does not have to supply the corresponding vector anymore. This change is the reason for the breaking change label.

--


src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -557,48 +573,50 @@ void CONSTRAINTS::SpringDashpot::evaluate_robin(std::shared_ptr<Core::LinAlg::Sp
...
-          }
-          else
+          if (function_id_nonlinear_stiffness[dof].has_value() &&
+              function_id_nonlinear_stiffness[dof].value() > 0)
           {
             std::array<double, 3> displ = {(*disp)[0], (*disp)[1], (*disp)[2]};
             force_disp = Global::Problem::instance()
                              ->function_by_id<Core::Utils::FunctionOfSpaceTime>(
-                                 (numfuncnonlinstiff)[dof] - 1)
+                                 function_id_nonlinear_stiffness[dof].value() - 1)
```
sebproell, line 607: I am surprised to still see a `-1` after #280. Does this mean the `function_id_nonlinear_stiffness` in the input is larger by one compared to the actual ID?

lauraengelhardt: All tests pass after removing the -1.
We have no test that uses `FUNCTNONLINSTIFF != none` and `NUMDOF 1` which is where this branch arrives.
I'll remove the -1.

src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -43,13 +43,20 @@ namespace
...
-  double scale_by_optional_function(double baseline_value, int num_func, double total_time)
+  double scale_by_optional_function(
+      double baseline_value, Core::IO::Noneable<int> funct_id, double total_time)
   {
-    return num_func ? Global::Problem::instance()
-                              ->function_by_id<Core::Utils::FunctionOfTime>(num_func)
-                              .evaluate(total_time) *
-                          baseline_value
-                    : baseline_value;
+    if (funct_id.has_value() && funct_id.value() > 0)
```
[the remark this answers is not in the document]
lauraengelhardt: I'll change for this condition.

For all other functions, I have a PR almost ready, that replaces former -1's in the input files by nones. If we then remove those checks, we basically enforce using none instead of -1 (or any non-valid function IDs).

lauraengelhardt:
Just for your information: @sebproell, @gilrrei, and I discussed how to tackle the current error during the metadata creation. We have come up with a solution, so this PR is on hold until this issue is resolved.

sebproell:
After #510, this PR can be reactivated.

lauraengelhardt:
Closing this PR, as it has become outdated.

PR #450, opened 2025-02-26
Add tests for linear solid poro elements

amgebauer:
I found that linear kinematics in solid-poro was not tested.

I reused some nonlinear tests and switched to `linear`. I tested it with the old solid elements and now use the new ones. The results are the same. BUT: Also the old solid-poro elements were untested with linear kinematics. So, I would actually not trust the results too much.

@lkoeglmeier @knarfnitram @SchaefP1 I leave the decision up to you whether we want to merge this. We may also remove linear kinematics from solid-poro since it was untested for several years.


lkoeglmeier:
Thank you for bringing this up and adding the test case @amgebauer. Personally, I would vote to delete the linear kinematics part for now, as we should take a closer look at it before using it. Adding the tests might make us forget about it.

knarfnitram:
We could keep the linear kinematics and add a warning that says to check this code before using it again. 
Since I mainly used the non-linear kinematics anyway, we can remove it from my site.

amgebauer:
ok, then let's remove it. Recreating the current state is easy. The hard work is verifying that everything is correct.

PR #455, opened 2025-02-27
Make all parameter types noneable

gilrrei:
## Description and Context
All parameter types can be set to `null` if provided by the metadata file.

### DIFF
--- a/utilities/metadata_utils.py
+++ b/utilities/metadata_utils.py
@@ -244,6 +244,7 @@ class List(Parameter):
 
     size: int = None
     spec: Parameter = None
+    noneable: bool = False
     _type = "list"
 
     def __post_init__(self):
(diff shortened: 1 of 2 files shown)


utilities/metadata_utils.py
```diff
@@ -244,6 +244,7 @@ class List(Parameter):
 
     size: int = None
     spec: Parameter = None
+    noneable: bool = False
```
sebproell, line 247: I guess this makes sense although I haven't thought about it. I would enter an empty list as

```
list: []
```

instead of

```
list: null
```
correct?

sebproell (approved):

utilities/metadata_utils.py
gilrrei: both are fine IMO, the same holds for vectors of course

amgebauer (approved):

PR #458, opened 2025-02-27
Use std::optional instead of Noneable

### LINKED ISSUE #437: [InputSpec] Remove .required field and use std::optional

sebproell:
As outlined in #437, we can do the following things:

- `Noneable` is just `std::optional` now since we no longer have the `.required` field, so there should be no confusion about what this type means.
- When reading `parameter<std::optional<int>>("name")`, you expect that this parameter "name" does not have to be given in the input file and, if indeed not given, its value should be empty. Thus, the code is changed to no longer accept any `.default_value` if your parameter is a `std::optional<T>`.

Closes #437

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -649,7 +649,7 @@ namespace Core::IO
     namespace Internal
     {
       template <SupportedType T>
-      struct EntrySpec
+      struct ParameterSpec
       {
         std::string name;
         using StoredType = T;
@@ -1158,7 +1158,7 @@ namespace Core::IO
 // --- template definitions --- //
 
 template <Core::IO::SupportedType T>
-void Core::IO::InputSpecBuilders::Internal::EntrySpec<T>::parse(
+void Core::IO::InputSpecBuilders::Internal::ParameterSpec<T>::parse(
     ValueParser& parser, InputParameterContainer& container) const
 {
   if (parser.peek() == name)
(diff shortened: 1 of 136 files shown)


amgebauer (dismissed):

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -192,8 +192,8 @@ namespace Core::IO
       void operator()(ryml::NodeRef node, size_t* size)
       {
         FOUR_C_ASSERT(node.is_map(), "Expected a map node.");
-        // Pull up the "noneable" aspect. The fact that this type wraps another type is specific
-        // to C++ and not relevant to other tools. Simply knowing that a type can be "none" is
+        // Pull up the std::optional aspect. The fact that this type wraps another type is specific
+        // to C++ and not relevant to other tools. Simply knowing that a type can be empty is
         // enough for them.
         emit_value_as_yaml(node["noneable"], true);
```
sebproell, line 198: I talked to @gilrrei and we thought we just stick with the name `noneable` in the `metadata` for now. Writing `optional` here would be confusing since we also have a `required` field (only in the metadata!). Maybe we will come up with something better over time. In the end, it does not matter too much and the language barriers between C++, yaml, json schema, etc. will always require implicit knowledge.

amgebauer (approved):

sebproell:
This is ready.

lauraengelhardt (approved):

PR #466, opened 2025-02-28
WIP: Consistently use none to indicate non-set function ids

lauraengelhardt:
## Description and Context
- Replaces `0`s by `none`s for all input function ids (`FUNCT`s).
- Remove all checks for non-positive function ids before calling `function_by_id`.

If you now use 0, -1, or any other invalid value in the input file to indicate you do not want to supply a `FUNCT`, an error is thrown by `function_by_id`. 

I'll supply a python script I used for the replacements if needed. 

--


lauraengelhardt:
Let's see what tests fail. I'll let you know once this is ready.

lauraengelhardt:
This PR is ready.

sebproell:
@lauraengelhardt Thanks! This makes things much more consistent. I just wonder how big the fallout from this change is going to be. As far as I can tell, this will break every input file. Since we had a lot of breaking changes anyway, we might as well throw this one in as well. On the other hand, we should really start versioning of the input file (#74) and allow to perform these adaptations automatically.

lauraengelhardt:
Closing this PR. As @sebproell noted, such changes only reasonable if we have an automatic upgrade mechanism for input files (see #74)

PR #467, opened 2025-02-28
Enable vtk output in pressure-based arteries

lkoeglmeier:
This pr enables the vtk-based output in the pressure-based artery elements

- a vtk-test was added to test the functionality
- the old output mechanism was deleted 

Thank you @c-p-schmidt for your help!


src/art_net/4C_art_net_impl_stationary.cpp
```diff
@@ -460,6 +473,42 @@ void Arteries::ArtNetImplStationary::prepare_time_loop()
   return;
 }
 
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+void Arteries::ArtNetImplStationary::collect_runtime_output_data()
+{
+  // write domain decomposition for visualization (only once at step 0!)
+  if (step_ == upres_ or step_ == 0) visualization_writer_->append_element_owner("Owner");
```
c-p-schmidt, line 481: I have seen that it already used to be like this. However, if the comment should be true, you need to change as follows:
```suggestion
  if (step_ == 0) visualization_writer_->append_element_owner("Owner");
```

src/art_net/4C_art_net_impl_stationary.hpp
```diff
@@ -87,14 +88,20 @@ namespace Arteries
       FOUR_C_THROW("LoadState() not available for stationary formulation");
     }
 
+    //! collect runtime output data
+    void collect_runtime_output_data();
+
+    //! write data required for restart
+    virtual void output_restart();
```
c-p-schmidt, line 95: It does not need to be virtual as you do not override it.
```suggestion
    void output_restart();
```

src/art_net/4C_art_net_impl_stationary.cpp
```diff
@@ -476,30 +525,18 @@ void Arteries::ArtNetImplStationary::output(
   TEUCHOS_FUNC_TIME_MONITOR("             + output of solution");
 
   // solution output and potentially restart data
-  if (do_output())
+  if (step_ % upres_ == 0)
```
c-p-schmidt, line 528: Is `do_output()` used somewhere else? Otherwise, you could remove the method.

c-p-schmidt (commented):
The changes look good to me. It is nice to see that your problem is now only using vtk-based output. 

Please don't forget to mark the methods in https://github.com/4C-multiphysics/4C/issues/211 to be resolved after this is merged. Thanks!

bwirthl (approved):

c-p-schmidt:
I forgot one thing in the initial review, sorry about that. Can you please check if you can delete parts in the `post_processor`.
Besides that, I'm fine with the changes.

bwirthl (dismissed):

lkoeglmeier:
> I forgot one thing in the initial review, sorry about that. Can you please check if you can delete parts in the `post_processor`. Besides that, I'm fine with the changes.

I have also removed the option of processing solid/structure data in the postprocessor, as we can already use the Vtk output there too.

c-p-schmidt (dismissed):

lkoeglmeier:
Sorry for the mess. I deleted the remaining post processing tests and updated the commit message.

c-p-schmidt (approved):

bwirthl (approved):

PR #468, opened 2025-02-28
Remove \author and \date comments

### LINKED ISSUE #91: Proposal to remove \author and \date from doxygen
> Originating from https://gitlab.lrz.de/baci/baci/-/issues/1322

## Description

In the past, some people used the doxygen key words `\author`, `@author`, `\date` or `@date` to provide information about the origin of an implementation. Today, I don't see great value in that for several reasons:

1. This can be tracked in git.
2. These statements often use abbreviations, so that it might not even be possible to easily identify the actual person.
3. It clutters the source code.

Hence, this issue proposes to remove these statements from the entire code base.

- [x] Decide to remove such statements
- [x] Remove them

sebproell:
As described in #91. All occurrences of \author, \date, @author and @date are removed. This was surprisingly easy to regex replace :) 

Closes #91


src/contact/4C_contact_meshtying_abstract_strategy.hpp
```diff
@@ -541,8 +541,8 @@ namespace CONTACT
      *
      *  \param bt (in): Desired matrix block type, e.g. displ_displ, displ_lm, ...
      *
-     *  \date 05/2016
-     *  \author hiermeier */
+
+     *  */
```
lauraengelhardt, line 545: Does this destroy the doxygen formating?

lauraengelhardt (approved):

apps/post_monitor/4C_post_monitor.cpp
```diff
@@ -1692,8 +1692,7 @@ FOUR_C_NAMESPACE_CLOSE
  *
  * Note: Works in seriell version only! Requires to read one instance of the discretisation!!
  *
- * \author chfoe
- * \date 11/07
+
```
mayrmt, line 1695: This breaks the style: empty line instead of a leading asterisk. Purely cosmetic, of course.

src/adapter/4C_adapter_scatra_base_algorithm.hpp
```diff
@@ -37,7 +37,7 @@ namespace Adapter
 {
   /// general scalar transport field interface for multiphysics problems
   /*!
-  \date 07/08
+
```
mayrmt, line 40: The comment block is now empty.

src/adapter/4C_adapter_str_structure_new.hpp
```diff
@@ -213,7 +213,7 @@ namespace Adapter
     virtual void set_step_end(int step_end) = 0;
 
     /// Take the time and integrate (time loop)
-    /// \date 11/08
+
```
mayrmt, line 216: Just deleting this line would be the better option.

src/beam3/4C_beam3_triad_interpolation.hpp
```diff
@@ -30,20 +30,17 @@ namespace LargeRotations
 
     /** \brief Standard Constructor
      *
-     *  \author grill
-     *  \date 01/17 */
+     */
```
mayrmt, line 33: This could be condensed into a one-line comment.

mayrmt (approved):
@sebproell Thanks!

I am aware, that these changes have been done by automation, so some changes do not result in the optimal style. I highlighted some examples, just in case you see an easy way to fix them.

sebproell:
@mayrmt We can do another pass over all comments at any time. One thing I would like to do is to sanitize all doxygen blocks, such that they 

- start with `/**` or `/**` followed by a line break
- have an asterisk `*` in every comment line
- end with `*/` 

This makes the comment readable, distinct, and formattable.

src/contact/4C_contact_meshtying_abstract_strategy.hpp
sebproell: No, this should be OK.

Issue #465, opened 2025-02-28
Introduce magic_enum?


sebproell:
## Description
Lack of reflection is a constant annoyance in C++ and responsible for boilerplate code. A prominent example is the mapping of enum constants to string values. Luckily, there is [`magic_enum`](https://github.com/Neargye/magic_enum) which provides static reflection for enum constants and makes string converters obsolete. The addition of this lib would help a lot in the new InputSpec, as we could start to aggressively require enums while not burdening users with the `to_string` boilerplate code. Inclusion can be done without user intervention through `fetch_content`.

I have wanted this lib on multiple occasions already. I think it is time to just include it. Let me know if you see reasons against that. FWIW, deal.II included this lib a while back ;)

## Benefits

- Conversion between enum constants and their string representation is always consistent and up-to-date --> less error-prone
- Allows to remove all `to_string` functions for enums (if we make the values consistent) --> less boilerplate
- The simple conversion allows to use enums more proactively such as here https://github.com/4C-multiphysics/4C/pull/474/files#r1977180742 --> use the type system in a convenient manner

## Possible Solution and Definition of Done

- [x] Consider using this lib
- [ ] If we want it, add it --> #474 
- [ ] Finally, start removing all the obsolete to_string functions!

mayrmt:
@sebproell thanks for bringing this up! I had a quick look at `magic_enum` and, at first glance, this looks like a reasonable proposal.

I am wondering: how will we use this library? What are intended use cases and scenarios? How will we deal with our existing handling of `enum`s? Will this library simplify working with 4C also for non-expert coders?

PR #469, opened 2025-02-28
InputSpec: new selection for common use case

sebproell:
Together with @gilrrei and @lauraengelhardt, we identified a common use case when creating InputSpecs. One often wants to selectively activate more parameters based on the value of a certain parameter. This PR enables that based on strings. I am a bit reluctant to add another overload taking a enum-to-string mapping, and would rather want to use magic_enum for that, see #465.

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -833,6 +833,19 @@ namespace Core::IO
                      [&](const auto& val) { return this->operator()(val.second, size_info + 1); });
         }
       };
+
+
+      //! Helper for selection().
+      struct BasedOn
+      {
+        std::string name;
+      };
+
+      //! Helper for selection().
+      struct FromChoices
+      {
+        std::map<std::string, InputSpec> choices;
+      };
     }  // namespace Internal
 
     /**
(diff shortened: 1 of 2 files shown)


src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -1655,6 +1721,24 @@ Core::IO::InputSpec Core::IO::InputSpecBuilders::parameter(
...
+  specs.reserve(choices.choices.size());
+  for (auto&& [choice_name, choice_spec] : choices.choices)
+  {
+    specs.emplace_back(all_of({
+        selection<std::string>(selector.name, {choice_name},
+            {.description = "Selects which other parameters are allowed in this group."}),
+        std::move(choice_spec),
+    }));
+  }
+  return group(name, {one_of(specs)}, data);
```
sebproell, line 1739: @gilrrei For now, this implemented in terms of existing features. We can think about using a separate implementation internally to emit special info for this case. However, I would like to wait with that until we converged on the exact semantics of the new function.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -1033,6 +1046,48 @@ namespace Core::IO
     [[nodiscard]] InputSpec group(
         std::string name, std::vector<InputSpec> specs, GroupData data = {});
 
+    /**
+     * This function is used to select one of multiple InputSpecs based on the value of a parameter
+     * with name given in @p selector. The selector parameter is always put inside a group with the
+     * given @p name. For every possible value of the selector, a different InputSpec is expected
+     * inside the group @p name. For example:
```
lauraengelhardt, line 1053:
```suggestion
     * This function is used to select one of multiple InputSpecs based on the value of the parameter @p selector. 
     * For every possible value of the selector, a different InputSpec is expected.
     * The selector parameter and the selected InputSpec lives inside a group with the @p name. For example:
```

Maybe like this?

lauraengelhardt (commented):
Very nice! 
Maybe just `Internal::From`, but I'll leave that up to you.

lauraengelhardt (approved):

gilrrei (approved):

PR #471, opened 2025-03-01
Implement pressure-velocity based solid poro element with primary porosity (solid-poro-p1)

amgebauer:
This is the last large building block that is missing to finish #303 and to remove the `so3` folder 🥳 

Yet, this is a draft. I'm no expert in solid-poro and might not have chosen the best names for a few options. I will shortly add threads in the code.

The first two commits are already part of #464 and the last one in #472. I just wanted to open the PR so that we can resolve the last remaining points quickly and merge once I'm back.

Closes #303 

@lkoeglmeier @bwirthl @maxiludwig @SchaefP1 @knarfnitram

PR #472, opened 2025-03-01
Implement optional pack/unpack helper

amgebauer:
Cherry-picked from #471 with some additional fixes of doxygen.

### DIFF
--- a/src/core/comm/src/4C_comm_pack_helpers.hpp
+++ b/src/core/comm/src/4C_comm_pack_helpers.hpp
@@ -11,7 +11,6 @@
 #include "4C_config.hpp"
 
 #include "4C_comm_pack_buffer.hpp"
-#include "4C_comm_parobject.hpp"
 #include "4C_linalg_fixedsizematrix.hpp"
 #include "4C_linalg_serialdensematrix.hpp"
 #include "4C_linalg_serialdensevector.hpp"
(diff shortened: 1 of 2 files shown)


src/core/comm/src/4C_comm_pack_helpers.hpp
```diff
@@ -86,6 +86,24 @@ namespace Core::Communication
    */
   void add_to_pack(PackBuffer& data, const Packable auto& obj) { obj.pack(data); }
 
+  /*!
+   * \brief Add stuff to the end of a char vector data
+   *
+   * This method is templated for std::optional<T>
+   * \param[in,out] data char string stuff shall be added to
+   * \param[in] stuff std::optional<T> that get's added to stuff
+   */
```
sebproell, line 95:
```suggestion
  /*!
   * \brief Add stuff to the end of a char vector data
   *
   * This method is templated for std::optional<T>
   * \param[in,out] data char string stuff shall be added to
   * \param[in] stuff std::optional<T> that gets added to stuff
   */
```

Given that you fixed doxygen: it would probably make the most sense to document these functions once and refer to a general doc for all the overloads. Just a thought; I might do that.

sebproell (approved):
Thanks!

lauraengelhardt (approved):

PR #473, opened 2025-03-01
CMake: add FOUR_C_ENABLE_DEVELOPER_MODE

sebproell:
This option changes the details of the build to give a more pleasant developer experience. For now, it enables the linking of unit tests for modules without first building lib4C, i.e., it trades faster recompilation against increased binary sizes.


amgebauer (dismissed):

sebproell:
This does not yet work. The problem is that -- due to the dependency cycles -- all modules except `core` require all other modules and cannot be used without them.

sebproell:
This should work now. In developer mode,  we are now building an extra module lib (at no additional cost, just disk space) if a module is marked as `NO_CYCLE`. This is only true for `core` for now. We use this lib for linking unit tests. I added a few more arg parsers and sanity checks along the way, since I frequently mess up the setup otherwise.

The module lib will become more useful with more modules being truly modular. I've been planning to add this anyway.

amgebauer (approved):

c-p-schmidt (approved):

PR #474, opened 2025-03-02
Add magic_enum dependency and allow parameter to take enums

### LINKED ISSUE #465: Introduce magic_enum?

sebproell:
This PR adds `magic_enum` (see #465) which allows us to take enums in `InputSpecBuilders::parameter` without any manual (and obvious) string conversions.

Going forward, this has the potential to replace the old `selection`. Note: this means the value of an enum parameter in the input should always be spelled exactly as the enum constant in code and we do not remap the values to slightly different spellings. That is an upside in my eyes, but I still wanted to let you know.

Closes #465


src/global_legacy_module/4C_global_legacy_module.cpp
```diff
@@ -345,12 +345,8 @@ namespace
...
                     parameter<bool>("SPECIAL"),
                 }),
-                selection<Solid::ResultTest::TestOp>("OP",
-                    {{"unknown", Solid::ResultTest::TestOp::unknown},
-                        {"sum", Solid::ResultTest::TestOp::sum},
-                        {"max", Solid::ResultTest::TestOp::max},
-                        {"min", Solid::ResultTest::TestOp::min}},
-                    {.default_value = Solid::ResultTest::TestOp::unknown}),
+                parameter<Solid::ResultTest::TestOp>(
+                    "OP", {.default_value = Solid::ResultTest::TestOp::unknown}),
```
sebproell, line 349: This diff nicely shows one of the benefits of magic enum: We can skip the manual mapping of strings to enums and are guaranteed to have consistent values.

sebproell:
Thoughts?

src/core/io/src/4C_io_value_parser.hpp
```diff
@@ -190,6 +193,24 @@ namespace Core::IO
...
+      read_internal(string);
+      auto val = magic_enum::enum_cast<Enum>(string);
+      if (val)
+      {
+        value = *val;
+      }
+      else
+      {
+        FOUR_C_THROW(std::format("Could not parse value '{}' as an enum constant of type '{}'.",
+            string, magic_enum::enum_type_name<Enum>()));
```
amgebauer, line 210: It's time that we change the internals of `FOUR_C_THROW`.

src/core/io/src/4C_io_yaml.hpp
```diff
@@ -159,6 +167,24 @@ namespace Core::IO
...
+    auto substr = node.node.val();
+    auto val = magic_enum::enum_cast<T>(std::string_view(substr.data(), substr.size()));
+    if (val)
+    {
+      value = *val;
+    }
+    else
+    {
+      FOUR_C_THROW(std::format("Could not parse value '{}' as an enum constant of type '{}'.",
+          std::string_view(substr.data(), substr.size()), magic_enum::enum_type_name<T>()));
```
amgebauer, line 184: Printing all possible values of the enums would help here a lot.

src/core/io/src/4C_io_value_parser.hpp
sebproell: I agree and will open an issue for that.

src/core/io/src/4C_io_yaml.hpp
sebproell: In theory, you are right. However, you are not going to see this error when reading input. The place to print the possible values would be within the `MatchTree`.

c-p-schmidt (approved):
It's nice that we no longer have to manually write the mapping of strings to enums.

sebproell:
@mayrmt Since you recently improved some input parameters: what do you think about this PR?

mayrmt (approved):

mayrmt:
Sure 🚀

PR #476, opened 2025-03-03
Add CMake option FOUR_C_ENABLE_METADATA_GENERATION

sebproell:
Allows to turn off metadata and schema generation in environments where it does not work. It stays on by default.

FYI @ppraegla @tnschmid

### DIFF
--- a/apps/global_full/CMakeLists.txt
+++ b/apps/global_full/CMakeLists.txt
@@ -32,29 +32,30 @@ add_custom_command(
     ${CMAKE_BINARY_DIR}/${FOUR_C_EXECUTABLE_NAME_LEGACY}
   COMMENT "Creating symbolic link"
   )
-
-if(Python3_VERSION VERSION_GREATER_EQUAL "3.10")
-  # Create a json schema file for the input
-  add_custom_command(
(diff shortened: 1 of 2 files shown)


amgebauer (approved):
This change is useful on its own, but what is the problem on these target systems?

I would prefer the solution where `FOUR_C_ENABLE_METADATA_GENERATION` will fail if python is smaller than 3.10. We could even make it non-default: Metadata generation is only valuable if you also configure the metadata in your ide. So, you have to take additional manual actions.

sebproell:
> but what is the problem on these target systems

The problem that motivated this change is that we have to run 4C for metadata generation and for that 4C always calls MPI_Init. If the environment you use for building cannot correctly run with MPI, this will not work. There is probably another solution for that but we thought -- as you say yourself -- that this options is correct regardless.

sebproell:
> Metadata generation is only valuable if you also configure the metadata in your ide

It depends. You are talking about the json schema file and here I agree. However, we will be able to do a lot more with the `4C_metadata.yaml` file, e.g. generate docs, use it input generators, etc. I am also fine with turning it off by default if you think this is more reasonable.

Issue #463, opened 2025-02-28
[Input] Remove remnants of old input mechanisms and use InputSpec directly


sebproell:
## Description

The various ways to read input have all been superseded by `InputSpec`. However, some old interfaces still exist. Internally, they already call the new functionality. To get a consistent user experience and show that people can actually use the new feature everywhere, we should remove these. 

## Possible Solution and Definition of Done

- [ ] Remove the helpers in `Core::Utils`
  - [ ] Remove `bool_parameter`
  - [ ] Remove `int_parameter`
  - [ ] Remove `double_parameter`
  - [ ] Remove `string_parameter`
  - [ ] Remove `string_to_integral_parameter`
  - [ ] Remove temporary helper `SectionSpecs`

lauraengelhardt:
I just started removing `int_parameter`; seems to be doable with a regex search and replace quite well.

sebproell:
@lauraengelhardt Thanks! Please coordinate with @maxiludwig who also removes some helpers.

sebproell:
I'll replace the `string_to_integral`.

PR #479, opened 2025-03-04
Remove old input interfaces for bool, double and string

maxiludwig:
Removes the calls of input helper functions `Core::Utils::bool_parameter`, `Core::Utils::double_parameter`, `Core::Utils::string_parameter`. `int_parameter` and `string_to_integral_parameter` should follow soon. See #463

### DIFF
--- a/src/scatra/4C_scatra_timint_hdg.cpp
+++ b/src/scatra/4C_scatra_timint_hdg.cpp
@@ -47,7 +47,6 @@ ScaTra::TimIntHDG::TimIntHDG(const std::shared_ptr<Core::FE::Discretization>& ac
       padapterrorbase_(params->get<double>("PADAPTERRORBASE")),
       padaptdegreemax_(params->get<int>("PADAPTDEGREEMAX")),
       elementdegree_(nullptr)
-
 {
 }
 
--- a/src/ale/4C_ale_input.cpp
+++ b/src/ale/4C_ale_input.cpp
@@ -22,9 +22,11 @@ void ALE::set_valid_parameters(std::map<std::string, Core::IO::InputSpec>& list)
 
   Core::Utils::SectionSpecs adyn{"ALE DYNAMIC"};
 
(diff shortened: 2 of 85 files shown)


sebproell:
@maxiludwig The failing check is partly my fault, since another change disabled implicit conversion between types. Here you pass an `int` 1 as a default value for the `double` parameter.

```
/__w/4C/4C/src/inpar/4C_inpar_elemag.cpp:28:24: error: invalid initialization of reference of type ‘FourC::Core::IO::InputSpecBuilders::ParameterDataIn<double>&&’ from expression of type ‘<brace-enclosed initializer list>’
   28 |       parameter<double>("TAU", {.description = "Stabilization parameter", .default_value = 1}));
      |       ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

If you rebase your changes onto main and build locally, you should see all the errors.

General comment: I am not sure how to handle implicit conversions. In the case of passing an `int` when a `double` is expected , an implicit conversion seems OK, but passing e.g. a `bool` for an `int` seems wrong.

src/scatra/4C_scatra_timint_hdg.cpp
```diff
@@ -47,7 +47,6 @@ ScaTra::TimIntHDG::TimIntHDG(const std::shared_ptr<Core::FE::Discretization>& ac
       padapterrorbase_(params->get<double>("PADAPTERRORBASE")),
       padaptdegreemax_(params->get<int>("PADAPTDEGREEMAX")),
       elementdegree_(nullptr)
-
```
sebproell, line 50: There are some spurious formatting changes in this PR. They are an improvement, so I am OK with keeping them.

sebproell (approved):

c-p-schmidt (approved):

PR #484, opened 2025-03-06
Fix post_setup calls for new structural time integration

dragos-ana:
- Separated setup and post_setup for the new time integration to avoid post_setup evaluation of faulty states during restart.
- Adapted the various problem types containing structure fields to the modified implementation.
- Currently, there are 9 / 2943 tests still failing...

## Description and Context

The method `Solid::TimeInt::Base::setup` calls both the `setup` and the `post_setup` method of the time integrator. This becomes problematic during `Solid::TimeInt::Base::read_restart`, since the time integrator calls `post_setup` two times within this method for the current implementation. Specifically, the first call within the `setup` method of the time integrator is faulty because the restart information is not read yet - in simulations, this means that we evaluate 
$\boldsymbol{F} = \boldsymbol{I}$ Gauss-point-wise before evaluating the correct $\boldsymbol{F}$ obtained from the restart information. For materials with internal variables, e.g., for `InelasticDefgradTransvIsotropElastViscoplast`, this can lead to evaluation problems, since the internal variables may have evolved significantly before the restart, not allowing for the evaluation of $\boldsymbol{F} = \boldsymbol{I}$ at each Gauss point anymore.
In discussions with @amgebauer and @c-p-schmidt, we have decided that the faulty `post_setup` call has to removed, and that the `Solid::Integrator::equilibrate_initial_state` method shall be renamed to something more expressive (basically, the method evaluates the mass matrix and computes consistent initial accelerations). This PR does both tasks. Specifically, it adds a separate ` Solid::TimeInt::Base::post_setup` method which calls the `post_setup` of the time integrator only in the case that we don't currently restart - conversely, the ` Solid::TimeInt::Base::read_restart` method still correctly calls the time integrator `post_setup` after reading the restart information.
These changes required adaptations for the structure field post setup routine in various problem types, such as SSI, TSI, PaSI, etc. Currently, there are still 9 tests which fail, predominantly due to small changes in the test results:

-[]  poromultielastscatra_2D_quad4_starting_dbc.dat-p2
-[] sohex8_multiscale_macro.dat-p2-restart_from_same_input
-[] sohex8_multiscale_macro_2micro.dat-p2-restart_from_same_input
-[] sosh8_freeflying_ruler_sta_centrdiff_new_restart.dat-p2_from_sosh8_freeflying_ruler_sta_centrdiff_new.dat-p2-restart_step_70
-[] ssi_3D_hex8_quad4_meshtying.dat-p2
-[] ssi_twoway_3D_line2_elch_s2i_butlervolmer_multiscale.dat-p1
-[] ssi_twoway_3D_line2_elch_s2i_butlervolmer_multiscale.dat-p1-restart_from_same_input
-[] sohex8_multiscale_macro.dat-nestedPar-restart
-[] sohex8_multiscale_macro_2micro.dat-nestedPar-restart

As I am not an expert in either of the failing problem types, I would need some help / guidance on how the failing tests can be adapted.


c-p-schmidt:
Thank you for this draft. I'll look into the `SSI` tests probably at the beginning of next week.

georghammerl:
Thank your for your draft. I will look into the multiscale topic.

bwirthl:
I fixed the poromultielastscatra_2D_quad4_starting_dbc test.

dragos-ana:
@bwirthl Thanks a lot for your contribution!

c-p-schmidt:
After quite some debugging I found the origin of the deviation between the old and new implementation concerning the failing `ssi_twoway`  tests. It originates from this line: https://github.com/4C-multiphysics/4C/blob/main/src/ssi/4C_ssi_str_model_evaluator_partitioned.cpp#L156.

The difference is that in the current main version, the mesh tying was neglected during the calculation of the initial accelerations as the ssi-algorithm was not yet set up, and thus, the if statement was evaluated as false.

Now that the setup task and the calculation of the initial accelerations are separated, the ssi-algorithm is completely set up, and the mesh tying condition is taken into account during the calculation of the initial accelerations.

Thus, this refactoring removes a bug. I'll update the results of the affected tests accordingly and push as soon as I have also resolved the second ssi-related failing test case.

amgebauer:
Once this is merged, can we then remove the optionality of the scalars from the `SolidScatra` elements?

dragos-ana:
@c-p-schmidt thank you for taking the time to analyze the tests and determine this bug!

c-p-schmidt:
> Once this is merged, can we then remove the optionality of the scalars from the `SolidScatra` elements?

I'll also look into this.

c-p-schmidt:
Brief info on the commits I added:
- Fix ssi_twoway tests: Adapts the result checks according to the reported bug (see https://github.com/4C-multiphysics/4C/pull/484#issuecomment-2775440591)
- Only communicate the scatra state to: This fixes the `ssi_3D_hex8_quad4_meshtying.dat-p2` by preventing to call set_scalar when it is a ssi one way coupled simulation solid to scatra, so in such cases trying to communicate the scatra state does not make sense
- First step of removing optional from solidscatra: I started to remove the first optionals from solidscatra. We can also decide to not do this within this PR. @amgebauer: Feel free to check if more can be removed

c-p-schmidt:
Unassigning myself as reviewer as I have pushed to this PR.

src/contact/4C_contact_nitsche_integrator_ssi.cpp
```diff
@@ -206,46 +206,20 @@ void CONTACT::IntegratorNitscheSsi::so_ele_cauchy_struct(Mortar::Element& mortar
...
+        auto* solid_scatra_ele =
+            dynamic_cast<Discret::Elements::SolidScatra*>(mortar_ele.parent_element());
+
+        FOUR_C_ASSERT_ALWAYS(solid_scatra_ele,
+            "Nitsche contact is not implemented for this element (expecting SOLIDSCATRA "
+            "element)!");
+
+        return solid_scatra_ele->get_normal_cauchy_stress_at_xi(mortar_ele.mo_data().parent_disp(),
+            mortar_ele.mo_data().parent_scalar(), parent_xi, gp_normal, test_dir, linearizations);
+      });
```
amgebauer, line 222: Nice! I think we can also remove this in the poro Nitsche framework now. And in the element implementation of poro, we can even delete a lot more of similar constructs. But we can do this in a separate PR. This one is already quite large.

amgebauer (approved):

c-p-schmidt:
Can someone please have a look? This is ready to review.

davidrudlstorfer (approved):

PR #494, opened 2025-03-11
Remove elemag

### LINKED ISSUE #493: Remove `elemag`
### Discussed in https://github.com/4C-multiphysics/4C/discussions/355

<sup>Originally posted by **davidrudlstorfer** February 11, 2025</sup>
Based on the recent PR #333 it turned out that there are some problematic things happening in `elemag` (most importantly that its built upon `scatra`). Additionally, it is currently used by no one. Can we remove it?

If `elemag` can be removed we could also think about removing the `DiscretizationHDG` - it's mostly used my `elemag` and also somehow in `fluid`. Its removal would allow us to further improve the modularization (move the `ProblemType` out of core).</div>

### Results of the discussion

`elemag` can be removed and `HDG` stays.

davidrudlstorfer:
## Description and Context

This PR removes electromagnetism (elemag) and electromagnetic diffusion (EMD) as well as connected boundary conditions.

Closes #493 

## Disclaimer

As this is my first large removal of code in 4C - please check all removed files during the review - thanks!

Locally everything builds and the minimal tests succeed


davidrudlstorfer:
This PR also removes the old output for elemag (see #211)

c-p-schmidt (dismissed):

src/core/fem/src/discretization/4C_fem_discretization_hdg.cpp
```diff
@@ -473,7 +473,7 @@ void Core::FE::Utils::DbcHDG::do_dirichlet_condition(const Teuchos::ParameterLis
     Teuchos::ParameterList initParams;
 
     const auto problem_type = *params.get<const Core::ProblemType*>("problem_type");
-    if (problem_type == Core::ProblemType::elemag or problem_type == Core::ProblemType::scatra)
+    if (problem_type == Core::ProblemType::scatra)
```
amgebauer, line 476: Could you check whether this can be removed completely? To me it looks like it originates from the fact that elemag problems evolve to scatra at runtime.

src/fluid/4C_fluid_DbcHDG.cpp
```diff
@@ -203,8 +203,7 @@ void FLD::Utils::DbcHdgFluid::do_dirichlet_condition(const Teuchos::ParameterLis
     Core::LinAlg::SerialDenseMatrix elemat1, elemat2;
     Core::Elements::LocationArray dummy(1);
     Teuchos::ParameterList initParams;
-    if (Global::Problem::instance(0)->get_problem_type() == Core::ProblemType::elemag or
-        Global::Problem::instance(0)->get_problem_type() == Core::ProblemType::scatra)
+    if (Global::Problem::instance(0)->get_problem_type() == Core::ProblemType::scatra)
```
amgebauer, line 206: same here

amgebauer (approved):

davidrudlstorfer:
I've now rebased onto latest main due to conflicts after #491

src/core/fem/src/discretization/4C_fem_discretization_hdg.cpp
davidrudlstorfer: I think this cannot be removed - I just checked when/by whom this was added - and this was added a long time before all of the elemag stuff. It was originally done by Martin Kronbichler / Michael Hiermeier.

Is there another way to check this or is this good enough?

The same applies for the second case down below

amgebauer: ok, then it might be necessary.

amgebauer (approved):

PR #495, opened 2025-03-11
Assume yaml filename for --to-yaml if not given

ischeider:
## Description and Context

As of now, for the tool `4C --to-yaml` it is necessary to define the the dat file input AND the yaml output. Since I presume that one often might want to simply replace the .dat suffix by .yaml for the output filename, it shouldn't be necessary to define the output explicitly. 
This is what the PR mainly does. Overwriting an existing yaml file is prevented.

In addition, when the given input file (also for 4C simulations in general) does not exist, an error message will be thrown. As of now, if a yaml file name is given as input - and the respective file does not exist - the message was `Error message:   The input file '<whatever_your_inputfile_is>.yaml' must contain a map as root node`, which I don't find intuitive.

See discussion #490 .

### DIFF
--- a/apps/global_full/4C_global_full_main.cpp
+++ b/apps/global_full/4C_global_full_main.cpp
@@ -27,6 +27,7 @@
 #include <unistd.h>
 
 #include <csignal>
+#include <filesystem>
 #include <format>
 #include <iostream>
 
@@ -47,7 +48,7 @@ namespace
         << "\n"
         << "SYNOPSIS\n"
         << "\t"
-        << "4C [-h | --help] [-p | --parameters] [--dat_to_yaml] [-d | --datfile] [-ngroup=<x>] \\ "
+        << "4C [-h | --help] [-p | --parameters] [--to-yaml] [-d | --datfile] [-ngroup=<x>] \\ "
            "\n"
            "\t\t[-glayout=a,b,c,...] [-nptype=<parallelism_type>] \\ \n"
         << "\t\t<dat_name> <output_name> [restart=<y>] [restartfrom=restart_file_name] \\ \n"
@@ -65,8 +66,9 @@ namespace
         << "\t--parameters or -p\n"
         << "\t\tDumps information about the parameters for consumption by additional tools.\n"
         << "\n"
-        << "\t--to-yaml <in_file_name> <out_file_name>\n"
-        << "\t\tRewrites a dat file to a yaml file.\n"
+        << "\t--to-yaml <in_file_name> [<out_file_name>]\n"
+        << "\t\tRewrites a dat file to a yaml file. (default output: "
+           "<in_file_name_without_suffix>.4C.yaml\n"
         << "\n"
         << "\t--datfile or -d\n"
         << "\t\tPrint example dat_file with all available parameters.\n"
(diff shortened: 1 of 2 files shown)


apps/global_full/4C_global_full_main.cpp
```diff
@@ -343,14 +344,35 @@ int main(int argc, char* argv[])
...
+      if (argc == 3)
+      {
+        outputfile_name = inputfile_name.replace_extension("yaml");
+        if (std::filesystem::exists(outputfile_name))
+        {
+          printf("You did not provide an output file name.\n");
+          printf("The default is to replace the original suffix by .yaml.\n");
+          printf("However, the file '%s' already exists. I will not continue\n",
+              outputfile_name.c_str());
+          MPI_Finalize();
```
ppraegla, line 364:
```suggestion
```
I think, this is not necessary as `MPI_Finalize` will be called by the `struct CleanUpMPI` in the destructor.

apps/global_full/4C_global_full_main.cpp
```diff
@@ -343,14 +344,35 @@ int main(int argc, char* argv[])
...
     {
+      std::filesystem::path inputfile_name = argv[2];
+      std::filesystem::path outputfile_name;
       Core::IO::InputFile input_file = Global::set_up_input_file(lcomm);
-      input_file.read(argv[2]);
-      std::ofstream output_file(argv[3]);
+      input_file.read(inputfile_name);
+      if (argc == 3)
+      {
+        outputfile_name = inputfile_name.replace_extension("yaml");
```
ppraegla, line 357:
```suggestion
        outputfile_name = inputfile_name.replace_extension("4C.yaml");
```
Should the extension be `4C.yaml`?

ppraegla (changes requested):
Thanks, for fixing the bug.

Could you also adapt the help output [here](https://github.com/4C-multiphysics/4C/blob/main/apps/global_full/4C_global_full_main.cpp#L68) to signal that the output file name is optional
```
"\t--to-yaml <in_file_name> [<out_file_name>]\n"
```

apps/global_full/4C_global_full_main.cpp
ischeider: why? If I have <problem_name>.yaml, what else could that be?

sebproell: @ischeider We should use `.4C.yaml` to indicate that this file follows the 4C schema. See also the TGM where Laura shows this in the live demo.

sebproell: @ischeider I think you still need to change that to `.4C.yaml` here?

apps/global_full/4C_global_full_main.cpp
```diff
@@ -65,8 +66,9 @@ namespace
         << "\t--parameters or -p\n"
         << "\t\tDumps information about the parameters for consumption by additional tools.\n"
         << "\n"
-        << "\t--to-yaml <in_file_name> <out_file_name>\n"
-        << "\t\tRewrites a dat file to a yaml file.\n"
+        << "\t--to-yaml <in_file_name> [<out_file_name>]\n"
+        << "\t\tRewrites a dat file to a yaml file. (default output: "
+           "<in_file_name_without_suffix>_4C.yaml\n"
```
sebproell, line 71:
```suggestion
           "<in_file_name_without_suffix>.4C.yaml\n"
```

sebproell (approved):

Issue #59, opened 2024-11-29
Move inpar files to related module


sebproell:
_Ported from Gitlab. Original issue [here](https://gitlab.lrz.de/baci/baci/-/issues/962)_

## Description

`inpar_`-files should not exist on the top-level; the files should move into their respective physics modules. This allows to define the input parameters locally for each module itself and removes a central dependency `inpar` that pollutes every physics module.

## Possible Solution and Definition of Done

- [ ] `inpar` folder removed, all files moved to physics.

davidrudlstorfer:
@bwirthl @lkoeglmeier I am currently moving inpar files again and I stumbled over the poro files again. One question:

We currently have the following poro files in inpar:

- 4C_inpar_poro.hpp (this just holds a random enum - the entire file can be deleted as far as I can tell, see #1007)
- 4C_inpar_poroelast.cpp/.hpp
- 4C_inpar_poroscatra.cpp/hpp

Do you have any preferences where to move the remaining poro inpar files? Or is this more complicated and needs to be split up?

lkoeglmeier:
yes, the first file can be deleted - thanks! `4C_inpar_poroelast.cpp/.hpp` and `4C_inpar_poroscatra.cpp/hpp `belong to the pressure_velocity_based poro formulation. I would suggest moving them to the `poroelast `and `poroelast_scatra` folder, respectively?

davidrudlstorfer:
@ppraegla Do you have any preferences/ideas on where I could/should put the particle input?

ppraegla:
> [@ppraegla](https://github.com/ppraegla) Do you have any preferences/ideas on where I could/should put the particle input?

Ideally, this would be combined with #92. So, the particle input can be directly placed in the new `particle` folder. I have this on my agenda. But currently no time to work on it. I can look at it in two weeks.

davidrudlstorfer:
So slowly this issue is coming to an end 👀 

The only files that remain are `io_runtime` files (8 files consisting of corresponding headers and implementations).

I currently do not have a real plan on where to put those files.

For example `4C_inpar_io.cpp` contains the general IO features and has a dependency on structure. But it also contains some top-level fluid parameters, binning and so on. Should this therefore be split up even more?

Maybe someone with more 4C knowledge (@4C-multiphysics/maintainer) could check this out. I would be open for any form of input (👀) and discussion.

lauraengelhardt:
I just took a quick look but was not really involved in modularization ;) 

[...]

Issue #81, opened 2024-11-29
Unify the folder structure


sebproell:
_Original issue: https://gitlab.lrz.de/baci/baci/-/issues/1021_

## Description

This is a first suggestion by the taskforce modularization for the future directory structure of the code basis:

```
4C/
├── core/
│   ├── discretization/
│   │   ├── tests
│   │   ├── src
│   │   └── ...
│   ├── utils
│   ├── comm
│   └── ...
├── physics/
│   ├── solid/
│   │   ├── tests
│   │   ├── src/
│   │   │   ├── alg
│   │   │   ├── ele
│   │   │   ├── utils
│   │   │   └── ...
│   │   └── ...
│   ├── fluid
│   ├── scatra
│   ├── ale
│   └── ...
├── doc
├── app
├── tests/
│   └── input_files
├── cmake
└── ...
```

The idea is that the second directory layer (discretization, utils, comm, etc. // solid, fluid, etc.) represent individual targets that can be built and linked against, independent of the others. The inspiration for this structure is taken from the Trilinos project. 

The naming in "core" is chosen to represent related groups of functionality currently available in 4C, e.g. the content of lib, fem_general and others operating on the discretization will be grouped together in the directory "discretization". This will hopefully make it easier to navigate and search for functionality. 

Clarifying comment for the idea behind some directories:

- core: This contains modules that form the core of 4C's shared functionality. These modules are widely used within a lot or all physics implementations. We will move (and split up) folders like `lib`, `geometry`, `io`, ... here.
- physics: This contains modules that provide the implementation of everything physics-specific (elements, special time integration rules, special post-processing or output, ...). 
- app: This contains the files that are necessary to build the executable (4C-release/debug). In here, we would like to eventually selectively enable only a subset of all modules (user-selected). For now this should mainly contain the files from `global_full` and other files that gather all the options that are provided by the various physics (e.g. module_registry).

The modularization team would continue working with this structure in the current efforts to separate the core from the physics modules. 
Any feedback is welcome.

## Tasks

- [ ] Move unit tests closer to their source code
- [ ] Merge related directories, e.g. `fluid` and `fluid_ele`
- [ ] Use the internal `src` and `tests` structure

PR #496, opened 2025-03-12
Restructure beaminteraction potential - Part I

davidrudlstorfer:
## Description and Context

This PR is the first part of a small restructuring/cleaning of the beaminteraction framework.

As a first step I want to move the beaminteraction-potential stuff into a dedicated subfolder (this PR) - the second PR should move the associated inpar files to this new structure - and the third one should rework the entire input based on the new input mechanism (a lot of doubles are read in as strings in the beaminteraction framework)

Preparation for next step => part of #59

The same procedure described above should be applied in the near future to the remaining beaminteraction submodels


src/beaminteraction/potential/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# This file is part of 4C multiphysics licensed under the
+# GNU Lesser General Public License v3.0 or later.
+#
+# See the LICENSE.md file in the top-level for license information.
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+four_c_auto_define_module()
+
+# TODO add dependencies
```
davidrudlstorfer, line 10: I have one question (I probably once did it but I forgot how to do it)

How do I add the correct dependencies here? There has to be some automatic way for that...

sebproell: What exactly do you want to add here? The problem is that beaminteraction is not modular and cyclically depends on everything else in 4C. So there is no single correct set of dependencies. If everything compiles, you are good to go for now.

isteinbrecher (dismissed):

src/beaminteraction/potential/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+
+four_c_add_dependency(${AUTO_DEFINED_MODULE_NAME} ${_dependencies})
```
sebproell, line 12:
```suggestion
```

davidrudlstorfer: Thanks

src/beaminteraction/potential/CMakeLists.txt
davidrudlstorfer: Thanks for the in-person discussion and explanation - I've removed the dependencies

davidrudlstorfer:
I've now directly moved the potential folder into a `src` folder to reflect our intended module structure.

This would be ready for review again

sebproell (dismissed):
I suggest moving the remaining files in `beaminteraction` into the new `src`. I'll leave it to you if you want to do it here.

isteinbrecher (dismissed):

davidrudlstorfer:
@sebproell thanks for reminding me of #81 

I've moved everything into `src` - now everything builds and works locally

isteinbrecher (approved):

sebproell (approved):

Issue #483, opened 2025-03-06
Allow for more flexible versions of dependencies


sebproell:
## Description

Right now, we only support a single version of a dependency. As discussed e.g. in #288, for developers working at the intersection of a dependency and 4C, this is cumbersome. Instead we should allow for a range of versions and provide ways to test with newer versions without requiring an immediate update for all other users. @mayrmt and I suggest the following strategy:

- Track a list of supported versions: this is a mapping from commit hash of the dependency to an _internal_ version name of that dependency. It is important to be independent of the release schedule of our dependencies and to have a way to specify versions between releases since these can be too slow (e.g. only 1 per year in deal.II, ~1 per year in Triilinos). So we need something like this:
```
# hash internal_version_name external_version_name (for documentation)
abc123 2025.1 unknown
xyz789 2025.2 16.1.0
```
- Provide the usual testing macros to check the version, e.g. `#if FOUR_C_TRILINOS_VERSION_GREATER(2025,2)`
- Bonus: during configuration we can check whether the version of a dependency is known and warn if it not.
- For CI testing: We thought it makes more sense to test the newest versions in PRs since this is the one that active developments should be done with. The older/oldest version that we still support can be tested nightly.
- Over time, e.g. when doing a release, we can drop older supported versions.

This strategy mostly applies to Trilinos now, and will apply to deal.II soon. Of course, it is extensible to any other dependency if we decide to work more closely on it.

## Possible Solution and Definition of Done

- [x] Introduce the necessary helpers 
- [ ] Adapt testing

sebproell:
https://github.com/4C-multiphysics/4C/pull/673 reveals that we

- need better testing for versions
- need a way to have requirements between dependencies

The newer ArborX version requires a newer Kokkos version, which requires a newer Trilinos version.

PR #499, opened 2025-03-12
Extend LNM presets for Ubuntu 24.04

davidrudlstorfer:
## Description and Context

Based on the new Ubuntu 24.04 install we need to use the built-in gcc and g++ compiler. This PR extends the presets so we have the new setup as standard and a backup for the old Ubuntu 20.04 installs available.

Once all workstations are switched we can remove the presets for 20.04

### DIFF
--- a/presets/lnm/workstation/CMakePresets.json
+++ b/presets/lnm/workstation/CMakePresets.json
@@ -15,9 +15,9 @@
         "FOUR_C_TRILINOS_ROOT": "/lnm/lib/packages/trilinos/16-1-0_v2/release-shared",
         "BUILD_SHARED_LIBS": "ON",
         "CMAKE_BUILD_TYPE": "RELEASE",
-        "CMAKE_CXX_COMPILER": "/lnm/spack_packages/linux-ubuntu20.04-ivybridge/gcc-13.2.0/bin/g++",
+        "CMAKE_CXX_COMPILER": "g++",
         "CMAKE_CXX_COMPILER_LAUNCHER": "ccache",
-        "CMAKE_C_COMPILER": "/lnm/spack_packages/linux-ubuntu20.04-ivybridge/gcc-13.2.0/bin/gcc",
+        "CMAKE_C_COMPILER": "gcc",
         "CMAKE_C_COMPILER_LAUNCHER": "ccache"
       }
     },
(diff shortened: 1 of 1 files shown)


sebproell:
This breaks all user configs that inherit `lnm_workstation` but I guess this is what we get for using inheritance ;) 

Maybe it is time to stop providing such presets. With planned changes like #483, these predefined configs become less and less useful.

davidrudlstorfer:
> This breaks all user configs that inherit lnm_workstation but I guess this is what we get for using inheritance ;)

Yes I am aware of that - but what would be the alternative? It breaks either way - if we decide to switch it now to a new name the user has to select the new name. Once the switch is complete we will probably rename everything back so it breaks again. I think the user always has to update it twice 😅

> Maybe it is time to stop providing such presets. With planned changes like https://github.com/4C-multiphysics/4C/issues/483, these predefined configs become less and less useful.

So each user has to do their own presets? Do we somewhere document what is a good "base setup"? Because this is more or less a preset itself - in my opinion it can stay

sebproell:
The problem is that we are trying to sync up changes in local infrastructure with a repo state but these two are not naturally in sync. If we remove the presets, this problem does not exist. Right now we are pretending that a certain config works although this can never be fully true.

isteinbrecher:
@maxfirmbach do we also need to update our presets?

davidrudlstorfer:
@amgebauer @c-p-schmidt opinions? Either update the presets or remove them entirely from the repo?

@ppraegla how do you handle it for SAM?

amgebauer:
> @amgebauer @c-p-schmidt opinions? Either update the presets or remove them entirely from the repo?
> 
> @ppraegla how do you handle it for SAM?

I think it is time to remove them completely. We can leave the current ones in the repo until we migrated all workstations to Ubuntu 24.04.

davidrudlstorfer:
Sounds like we have decided that we want to remove the LNM (and probably also the IMCS) presets.

Therefore, I'll close this PR and open an issue to remove the LNM presets once the Ubuntu 24.04 upgrade is finished. Thanks for the discussion!

isteinbrecher:
At the moment I don't think we will remove the IMCS presets in the near future, as all our coworkers and students rely on them.

PR #505, opened 2025-03-13
Remove BACI symbolic link

davidrudlstorfer:
## Description and Context

Remove the last remnant of BACI - the symbolic link to the 4C executable


sebproell (approved):
Let's see how many complaints that "4C does not work" we will get after this change.

amgebauer (approved):

ppraegla (approved):

PR #510, opened 2025-03-14
Rework new selection to work with magic_enum

sebproell:
Just as we don't like to use untyped strings, we also do not want to use them for the `selection`. This PR changes the new selection from #469 to work with enums.

FYI @bwirthl

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.cpp
+++ b/src/core/io/src/4C_io_input_spec_builders.cpp
@@ -481,6 +481,13 @@ namespace
           }
           break;
         }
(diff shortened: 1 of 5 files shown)


davidrudlstorfer (approved):

davidrudlstorfer (approved):

sebproell:
We need another look here.

sebproell:
>We need another look here.

Apparently my comment was misleading, sorry for that. This PR is ready. I meant that we need another _review_ here please :)

amgebauer (approved):

PR #511, opened 2025-03-14
Membranes can now be written to output using vtk-based output

c-p-schmidt:
## Description and Context
Part of getting the new solid rid of the old output mechanism.

Part of #211


src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_data.hpp
```diff
@@ -615,11 +615,32 @@ namespace Solid
...
       }
 
+      //! get stress data vector
+      inline const std::shared_ptr<std::vector<char>>& get_opt_quantity_data() const
+      {
+        return opt_quantity_data_ptr_;
+      }
+
+      //! get nodal post processed optional quantity data vector
+      inline const std::shared_ptr<Core::LinAlg::MultiVector<double>>&
```
sebproell, line 630: Does this really need to be shared_ptr?

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
```diff
@@ -874,6 +887,90 @@ void Solid::ModelEvaluator::Structure::evaluate_analytical_error()
...
+
+  // set dummy evaluation vectors and matrices
+  std::array<std::shared_ptr<Core::LinAlg::Vector<double>>, 3> eval_vec = {
+      nullptr, nullptr, nullptr};
+  std::array<std::shared_ptr<Core::LinAlg::SparseOperator>, 2> eval_mat = {nullptr, nullptr};
+
+  evaluate_internal_specified_elements(
+      eval_mat.data(), eval_vec.data(), discret().element_row_map());
+
+  auto EvaluateGaussPointData = [&](const std::vector<char>& raw_data)
```
sebproell, line 930: Why CamelCase?

sebproell (dismissed):

src/inpar/4C_inpar_IO_runtime_vtk_output_structure.cpp
```diff
@@ -86,6 +86,13 @@ namespace Inpar
                 Inpar::Solid::GaussPointDataOutputType::gauss_points),
             sublist_IO_VTK_structure);
 
+        Core::Utils::string_to_integral_parameter<Inpar::Solid::OptQuantityType>(
+            "OPTIONAL_QUANTITY", "No", "Output of an optional quantity",
```
amgebauer, line 90: What does this `OPTIONAL_QUANTITY` mean? I might would expect a list of quantities to output?

amgebauer (dismissed):

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_data.hpp
c-p-schmidt: Inside the class it is still a shared pointer and I think it also has to be like this at the moment.

But I changed to a set method and a get method returning a const reference, such that the interaction should at least be better. I would postpone more cleanup to later pull requests.

amgebauer (approved):

c-p-schmidt:
Please have a look.

PR #512, opened 2025-03-14
Improve POINT COUPLING CONDITIONS

ischeider:
## Description and Context

The condition `DESIGN POINT COUPLING CONDITIONS` is a multipoint constraint, where all nodes of a point set are coupled in one or more specified directions to a master node (which is the node with the lowest node number in the set).
Up to now, if a node is coupled to a node in a specific direction by this condition, it cannot be coupled to another node, even if the direction is different.
The change provided here improves this behavior; occurance of a node in more than one point coupling condition (in different directions) is now possible. An example validates this behavior.

Remark: Still this feature only works if all nodes are on the same process. To change this, is a TODO for the future.


tests/input_files/sohex8_two-connected-pointcouplings.yaml
```diff
@@ -0,0 +1,259 @@
+TITLE:
+  - "elastic strip made of two materials, pulled at ymax"
+  - "dirichlet surface boundary conditions: symmetry at xmin, ymin, zmin "
+  - "point couplings (plane surface conditions) at xmax, zmax"
+  - "only runs on a single process (as of now)!"
```
isteinbrecher, line 5: Are the points with the MPCs at the same spatial position? If so, you can try
```
---------------------------------------------------------------MESH PARTITIONING
METHOD                                monolithic
```
this might produce a parallel distribution that still lets you run the test on 2 cores. This would at least show that the code works in parallel as long as all nodes are on the same rank.

isteinbrecher (commented):
Thanks for the contribution @ischeider !

isteinbrecher:
@davidrudlstorfer this is not possible at the moment. What comes closes to your description is a `LocSys` condition, which transforms the local coordinates to a rotated coordinate system. You could transform the coordinate system such that the first DOF of the node matches your local $e_1$ direction. The transformation is defined by a (pseudo-)rotation vector, which can even be governed by a function of (space and?) time - but as far as I know there is no way to parametrize the rotated coordinate system with other system DOFs (I think this would also lead to some quite complex linearizations - go AD!).

What exactly are you trying to solve? Maybe this can be achieved with beam-to-beam contact?

davidrudlstorfer:
@isteinbrecher thanks for the quick and detailed answer - I do not want to hijack the present PR to further discuss this topic - I'll open a discussion some time soon

tests/input_files/sohex8_two-connected-pointcouplings.yaml
ischeider: The point cloud at least for the cases I have in mind is really large, commonly it covers two surfaces of a cube (in the example, each surface has 15 of the 45 overall nodes). Thus, keeping these nodes on one process and partition only the rest is probably not very useful. Only for very large structures we

src/core/fem/src/dofset/4C_fem_dofset.cpp
```diff
@@ -210,87 +210,100 @@ int Core::DOFSets::DofSet::assign_degrees_of_freedom(
...
-    maxnodenumdf = numdfrownodes.MaxValue();
+    maxnodenumdf = num_dof_rownodes.MaxValue();
     get_reserved_max_num_dofper_node(maxnodenumdf);  // XFEM::XFEMDofSet set to const number!
 
+    std::vector<std::vector<int>> onoffcond;  // vector of onoff status for each condition
+    std::vector<int> numdofcond;              // vector of NUMDOF for each condition
+    std::vector<std::vector<int>> nodeids;    // vector of node IDs for each condition
+    std::vector<int> mgids;                   // vector of master node IDs for each condition
+
+    for (int k = 0; k < (int)couplingconditions.size(); ++k)
```
georghammerl, line 232: @ischeider You might use range based loops, e.g.,

    for (auto* const condition : couplingconditions)
    {
      onoffcond.push_back(condition->parameters().get<std::vector<int>>("ONOFF"));
      numdofcond.push_back(condition->parameters().get<int>("NUMDOF"));
      const auto* const conditioned_node_ids = condition->get_nodes();
      nodeids.push_back(*conditioned_node_ids);
      mgids.push_back(*conditioned_node_ids->begin());
      ...
    }

m-frey:
Thanks! Looks good to me.

georghammerl (commented):
@ischeider Thanks a lot. I added some comments.

isteinbrecher (approved):

PR #521, opened 2025-03-14
Convert and cleanup contact enums

mayrmt:
## Description and Context

Following PR #516, this PR converts all `enum` into `enum class` in contact input as well as removes many unused enums along with their utility methods.

Follows #516.


src/contact/4C_contact_input.hpp
```diff
@@ -32,10 +32,10 @@ namespace CONTACT
 
   /// Type of contact adhesion law
   /// (this enum represents the input file parameter ADHESION)
-  enum AdhesionType
+  enum class AdhesionType
   {
-    adhesion_none,  ///< no adhesion
-    adhesion_bound  ///< fix bounded adhesion
+    none,  ///< no adhesion
+    bound  ///< fix bounded adhesion
```
sebproell, line 38:
```suggestion
    bounded  ///< fix bounded adhesion
```

To increase the symmetry between enum and input name? I will run a script in the context of #463 to automatically replace such `string_to_integral_parameter` calls with the new equivalents. If the string options exactly match the enum names, I can directly replace with `paramter<Enum>` which is the best case. Otherwise, I will use `deprecated_selection`. So, since you clean this anyway, do you think you can drop the `None` spelling and the `b` options as well or would this break many files?

mayrmt: Done.

sebproell (dismissed):

davidrudlstorfer (dismissed):
Nice cleanup!

c-p-schmidt (approved):

mayrmt:
I have converted all `enum`s in contact input to `enum class`. This PR is now ready for review and merge.

davidrudlstorfer (approved):
Nice!

(Maybe you can squash the commits)

src/contact/4C_contact_input.hpp
```diff
@@ -22,83 +22,76 @@ namespace CONTACT
...
+    vague,      ///< no solving strategy defined
+    lagmult,    ///< method of Lagrange multipliers
+    penalty,    ///< penalty approach
+    uzawa,      ///< Uzawa augmented Lagrange approach
+    nitsche,    ///< Nitsche contact solution method
+    ehl,        ///< method for elasto-hydrodynamic lubrication
+    multiscale  ///< method for contact of rough surfaces with a multi scale approach
   };
 
   inline std::string solving_strategy_to_string(enum SolvingStrategy stype)
```
c-p-schmidt, line 54: This method is not required anymore, right?

src/contact/4C_contact_input.hpp
```diff
@@ -22,83 +22,76 @@ namespace CONTACT
...
   /// Type of contact friction law
   /// (this enum represents the input file parameter FRICTION)
-  enum FrictionType
+  enum class FrictionType
   {
-    friction_none = 0,  ///< no friction
-    friction_stick,     ///< perfect stick
-    friction_tresca,    ///< Tresca friction law
-    friction_coulomb    ///< Coulomb friction law
+    None,    ///< no friction
```
c-p-schmidt, line 27: Is there a reason why you decided to start with an uppercase letter here, whereas all other definitions are purely lowercase?

c-p-schmidt (dismissed):
Thanks for cleaning the enums within contact!

src/contact/4C_contact_input.hpp
mayrmt: It is still called multiple times, so it cannot be removed right now. I have left the migration to `magic_enum` for future work.

src/contact/4C_contact_input.hpp
mayrmt: I found it weird to spell actual names with small letters, however I will change that to comply with our naming convention. Fix will come in a minute.

c-p-schmidt (dismissed):

src/contact/4C_contact_input.hpp
sebproell: I think we have never finished the discussion on how to name enum constants. The enum type is CamelCase,but for the values inside we currently use CamelCase and snake_case. With the introduction of `magic_enum` we should consider that the style we choose is also the style we should use in input files.

mayrmt:
All tests pass now. Previously, I overlooked some `int` access to an `enum` due to a wild chain of c-style casts, that were not caught by the compiler.

sebproell (approved):

Issue #228, opened 2025-01-09
Move Readthedocs to Readthedocs


amgebauer:
## Description

We currently host readthedocs within github pages. We should move readthedocs to readthedocs so that we can publish documentation for multiple versions.

## Interested Parties

@ischeider

mayrmt:
If we are on a paid plan, we then can cancel it.

ischeider:
For open source projects we may use the community plan (free of charge). Version history is included.

mayrmt:
@ischeider, do you know if we are paying for a read-the-docs subscription?

gilrrei:
You could also use a different sphinx theme.  For example, [pydata](https://pydata-sphinx-theme.readthedocs.io/en/stable/) provides a dropdown menu for multiple releases natively, see [here](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html).

mayrmt:
Closing as completed. The documentation is now hosted at https://4c-multiphysics.github.io/4C/documentation/

PR #523, opened 2025-03-15
Reorganize our various docs

### LINKED ISSUE #352: Rename readthedocs to documentation
### Discussed in https://github.com/4C-multiphysics/4C/discussions/321

<sup>Originally posted by **ppraegla** February  5, 2025</sup>
The 4C documentation is called 'readthedocs'. However, readthedocs is a hosting platform for sphinx documentation while our actual documentation is written with Markdown/reStructuredText and generated with sphinx. Even if we decide to move the documentation to the readthedocs page (see #228 ) it is still sphinx.</div>

sebproell:
- Put everything under top-level doc since `doc_removed_code` always bothered me at the top-level.
- Rename `readthedocs` to `documentation` in targets, files, docs (closes #352)

I thought about doing this in smaller steps, but I believe it might be simpler to just look at the results in one go. You can check my fork to see how the files are now organized (https://github.com/sebproell/4C/tree/organize-docs).

In the next step, I think we should organize the files inside the `documentation` folder into something that is more indicative of the outline we use for the documentation. I always have a hard time finding the correct file I need to edit.


utilities/assets/badges/documentation_readthedocs.svg
sebproell, line 1: One idea I had for this badge: we could put the current version, i.e., 2025.2.0-dev to the right, since this is the version of the documentation we will link to. Thoughts?

mayrmt (dismissed):

sebproell:
I removed the old `readthedocs` job as required. After we merge this, we should add `documentation` instead.

doc/documentation/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+    COMMAND ${FOUR_C_EXECUTABLE_NAME} -h > ${_sphinx_OUT_DIR}/reference_docs/4C-help.txt
+    COMMAND
+      cmake ${PROJECT_SOURCE_DIR} --list-presets >
+      ${_sphinx_OUT_DIR}/reference_docs/4C-cmake-presets.txt
+    COMMAND
+      ${SPHINX_EXECUTABLE} -W -n -q -b html # warnings are errors and be nit-picky
+      -i "${CMAKE_CURRENT_SOURCE_DIR}" -i "${_sphinx_OUT_DIR}/reference_docs" -i
+      "${PROJECT_SOURCE_DIR}/tests/framework-test" -i "${_sphinx_TPLDEPEND_DIR}" -o
+      "${_sphinx_HTML_DIR}" -c "${_sphinx_OUT_DIR}" -s "${PROJECT_BINARY_DIR}/doc/tmp"
+    DEPENDS create_rtd ${FOUR_C_EXECUTABLE_NAME}
```
ppraegla, line 49:
```suggestion
    DEPENDS create_documentation ${FOUR_C_EXECUTABLE_NAME}
```

doc/documentation/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+  find_package(Sphinx REQUIRED)
+
+  set(_sphinx_THEME sphinx_rtd_theme)
+
+  configure_file(${_sphinx_CONFIG_IN} ${_sphinx_CONFIG_OUT} @ONLY)
+  message("Sphinx configure file conf.py is written based on ${_sphinx_CONFIG_IN}")
+
+  add_custom_target(
+    documentation ALL
+    COMMAND create_rtd
```
ppraegla, line 39:
```suggestion
    COMMAND create_documentation
```
There might also be other references to rtd

sebproell: This refers to the helper executable that exports data from 4C. This should be replaced by scripts that work directly on the 4C metadata. I did not rename anything for this folder. Do you want me to do it?

doc/documentation/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+# Set up the main documentation based on Sphinx
+
+four_c_process_global_option(FOUR_C_BUILD_DOCUMENTATION "Build the documentation" OFF)
+if(FOUR_C_BUILD_DOCUMENTATION)
+
+  # add doc-specific cmake modules
+  list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+
+  # set input and output files and directories
+  set(_sphinx_SOURCE_DIR ${PROJECT_SOURCE_DIR}/doc/documentation)
```
ischeider, line 18: To me it's not clear why the variables are now prefixed with _sphinx ( I would rather use DOCUMENTATION*), but maybe it's a matter of taste.

ischeider (dismissed):

doc/documentation/CMakeLists.txt
sebproell: The underscore is just a convention to indicate a local variable, and `sphinx` is what we use these for. They are not visible beyond this scope. I try to avoid all caps variables in CMake unless they are pre-defined or have some kind of global scope.

utilities/assets/badges/documentation_readthedocs.svg
ischeider: But then the tags moves with each new merged commit. Shouldn't be a tag pointing to a "constant" commit? If not, I support the idea.

sebproell: Well, the `dev` suffix indicates that this is the development version, so yes the content changes all the time. I was thinking that we might host the docs in a way (#228) that once can select the version, either a release one, or the current development version on main.

mayrmt (approved):

ischeider (approved):

doc/documentation/CMakeLists.txt
ppraegla: I would try to remove every occurrence of `rtd` because rtd implies that we use readthedocs which we don't. But if you want to rework the helper scripts anyway, I'm also fine with leaving it.

PR #524, opened 2025-03-15
Remove helper to convert Yes/No to bool

sebproell:
~~We should not need this. The coverage report claims it has a few uses, but I have no idea for which parameters. Let's run the actions and see what fails.~~

Remove a helper which was used for a single parameter "Disable Implicit Weighting" which is now entered as a `bool`.


davidrudlstorfer (approved):
Thanks for removing this 🙏🏻 This was on my ToDo List for this week 😉

c-p-schmidt (approved):

PR #526, opened 2025-03-15
Move solid nullspace to solid_3D_ele

amgebauer:


src/solid_3D_ele/4C_solid_3D_ele_nullspace.hpp
```diff
@@ -0,0 +1,101 @@
...
+  /*!
+   * @brief Helper function for the nodal nullspace of solid elements in any dimension
+   *
+   * @param node_coordinates (in):    coordinates of the node to compute the nullspace on
+   * @param x0 (in):      center of discretization
+   */
+  template <unsigned dim>
+    requires(dim == 2 || dim == 3)
+  Core::LinAlg::SerialDenseMatrix compute_solid_null_space(
+      const std::vector<double>& node_coordinates, const double* x0)
```
sebproell, line 28: Maybe:

```suggestion
      std::span<const double, dim> node_coordinates, std::span<double, dim> x0)
```

amgebauer: I agree, but this is used by all elements and needs to be cleaned up in the whole codebase, so out of the scope of this PR.

c-p-schmidt (approved):

PR #527, opened 2025-03-16
Add further 1D Gauss rules

dragos-ana:
## Description and Context

Added further 1D line Gauss rules (11-point up to 15-point). These are a necessary prerequisite to the inverse scaling and squaring evaluation of the matrix logarithm, for which I will create a PR in the upcoming days. The values for the Gauss rules are taken from [here](https://pomax.github.io/bezierinfo/legendre-gauss.html).


mayrmt:
Could you use Intrepid/Intrepid2 instead?

amgebauer:
> Could you use Intrepid/Intrepid2 instead?

Before @dragos-ana can use intrepid2, the tpetra-taskforce needs to do progress and provide the appropriate infrastructure.

amgebauer (approved):
I think you have verified the added integration rules?

dragos-ana:
@mayrmt, @amgebauer:
After receiving the hint yesterday, I managed to replace my current workflow in evaluating the matrix logarithm and use `Intrepid::Cubature` instead of the added 1D Gauss rules (similar to what is done in `4C_fem_general_utils_gausspoints.cpp`). Basically, the added Gauss rules of this PR are not required anymore, but they may still be added for a more complete set of Gauss rules, which can then be replaced by the `Intrepid2` framework. What is your opinion as maintainers? Should the Gauss rules still be added or should I close the PR?
@amgebauer yes, they are verified!

amgebauer:
@dragos-ana Do you now use `Intrepid::Cubature` directly in your material model? It would be better if you use some core utility function for that so that we only have to replace it in one central place and not in many.

dragos-ana:
Understood. After this PR is merged, I will push the changes to the matrix logarithm, and then we can decide which approach is best out of the two you suggested, @amgebauer !

PR #528, opened 2025-03-16
Switch error and assert macros to std::format

### LINKED ISSUE #482: Use std::format for throw and assert macros
## Description

`FOUR_C_THROW` and `FOUR_C_ASSERT` currently do `printf`-style formatting. With C++20 we can use `std::format` for much easier formatting. The annoying `c_str()` that needs to be appended to all strings right now is then no longer necessary.

sebproell:
Switch from C-style formatting, i.e., `%d`, `%f` etc., to `std:;format`, i.e., mostly just `{}`, unless you want to do something special.

This wasn't too difficult and actually fixed a lot of bugs (wrong format specifiers, missing specifies, too many specifies, etc.). We can now statically assert a lot of these cases. Since people might be used to the old way, I think we should leave in the check that errors _at compile-time_ if you use C-style format specifiers.

Note that I added a forward implementation of `std::unreachable` from C++23 since one beam function needs this.

Closes #482


c-p-schmidt (approved):

sebproell:
Compile time increases by quite a bit (1h46 min vs 1h22 min for clang). Let me see if I can do something about that.

sebproell:
I already identified some enum-to-string functions that are implemented in headers as `static` and are recompiled over and over again. Most can just be deleted, or at least moved to the implementation.

 I will remove these but would like to proceed here. The switch to `std::format` is what we want.

davidrudlstorfer (approved):

amgebauer (dismissed):

davidrudlstorfer (approved):

amgebauer (approved):

PR #530, opened 2025-03-17
Rework matrix functions framework

dragos-ana:
## Description and Context

- Generalized the exp/log framework for matrices to the more general matrix functions framework. The matrix square root is now also included as a further matrix function in this commit.
- Improved error management in this framework using a dedicated enum class with possible error types.
- For each of the matrix functions, we now have enum classes for the computation method to be utilized.
- The matrix logarithm routines were enhanced with the widely used inverse scaling and squaring method (Higham: Functions of Matrices, Algorithm 11.10) and the Pade approximation (partial fraction expansion as presented in Higham: Functions of Matrices, Eq. 11.18).
- Square root unit tests were added and the unit tests of the matrix functions now always test all the available computation methods.

#527 was created as a prerequisite to this PR.


src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1494 @@
...
+    if (X.norm2() == 0.0)
+    {
+      err_status = MatrixFunctErrorType::no_errors;
+      return Core::LinAlg::Matrix<dim, dim>{true};
+    }
+
+    // initialize m-point Gauss integration points and weights
+    FE::IntPointsAndWeights<1> int_pts_wts{
+        FE::num_gauss_points_to_gauss_rule<Core::FE::CellType::line2>(*pade_order)};
+
```
dragos-ana, line 330: @amgebauer This is the current status of the Gaussian quadrature used in the Pade approximation. I used the available core utilities, so that the migration to Intrepid2 is more straightforward, as proposed in my latest [PR](https://github.com/4C-multiphysics/4C/pull/527#issuecomment-2728869067).

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.hpp
```diff
@@ -0,0 +1,307 @@
...
+                         // iterations is exceeded)
+  };
+
+  /// make sure MatrixFunctErrorType is stream-insertable
+  inline std::ostream& operator<<(
+      std::ostream& stream, const MatrixFunctErrorType& matrix_funct_err_type)
+  {
+    stream << magic_enum::enum_name(matrix_funct_err_type);
+    return stream;
+  }
```
sebproell, line 52: Let's do this differently: I will soon add a header `4C_utils_enum.hpp` which provides such convenience magically without requiring extra definitions. You would only need to include the header in the file where you want to print. The `magic_enum` header is rather heavy so we don't want to pull it in if we simply need the enums here.

sebproell (dismissed):
No detailed review but I noticed this.

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1494 @@
...
+  if (calc_method == Core::LinAlg::SymMatrixExpFirstDerivCalcMethod::default_method)
+  {
+    if (norm < 0.3)
+      calc_method = Core::LinAlg::SymMatrixExpFirstDerivCalcMethod::taylor_series;
+    else
+      calc_method = Core::LinAlg::SymMatrixExpFirstDerivCalcMethod::eigenproj_based;
+  }
+
+  // further norm checking to determine appropriate computation procedure
+  if (calc_method == Core::LinAlg::SymMatrixExpFirstDerivCalcMethod::taylor_series)
```
c-p-schmidt, line 795: You could use a switch case to switch between the different methods.

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1494 @@
...
+  }
+
+  if (calc_method == Core::LinAlg::MatrixSqrtCalcMethod::db_iter_scaled_product)
+  {
+    // compute dim-dimensional unit tensor
+    Core::LinAlg::Matrix<dim, dim> id{true};
+    for (unsigned int i = 0; i < dim; ++i) id(i, i) = 1.0;
+
+    // compute exponent for scaling factor \f$ 1/(2 n) \f$
+    const double exponent = 1.0 / 2.0 / dim;
```
c-p-schmidt, line 69:
```suggestion
    const double exponent = 1.0 / (2.0 * dim);
```

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1494 @@
...
+    Core::LinAlg::MatrixFunctErrorType& err_status, Core::LinAlg::MatrixExpCalcMethod calc_method)
+{
+  // declare output matrix
+  Core::LinAlg::Matrix<dim, dim> output(true);
+
+  // get norm of input matrix
+  double mat_norm = input.norm2();
+
+  // direct calculation for zero-matrix
+  if (mat_norm == 0.)
```
c-p-schmidt, line 181: Really only a 0, not numerically 0 like smaller than e.g. 1e-16?

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1494 @@
...
+  // declare output variable
+  Core::LinAlg::Matrix<6, 6> output(true);
+
+  // compute 4-th order identity tensor
+  Core::LinAlg::Matrix<6, 6> id4sharp(true);
+  for (int i = 0; i < 3; i++) id4sharp(i, i) = 1.0;
+  for (int i = 3; i < 6; i++) id4sharp(i, i) = 0.5;
+
+  // direct calculation for zero-matrix
+  if (norm == 0.)
```
c-p-schmidt, line 779: Really only an exact 0, similar to above?

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1494 @@
...
+      if (n % 2 == 1)
+        Core::LinAlg::Tensor::add_symmetric_holzapfel_product(
+            output, Xn.at((n - 1) / 2), Xn.at((n - 1) / 2), .25 / fac[n]);
+    }
+  }
+  else if (calc_method == Core::LinAlg::SymMatrixExpFirstDerivCalcMethod::eigenproj_based)
+  {
+    double EWtolerance = 1.e-12;
+
+    Core::LinAlg::Matrix<3, 3> EV(input);
```
c-p-schmidt, line 857: Are these the eigenvalues and eigenvectors? Please use more descriptive names.

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1494 @@
...
+    //  d X^2 / d X  =  1/2 * (  delta_jk X_lj + delta_il X_kj
+    //                         + delta_jl X_ik + delta_kj X_il )
+    //
+    // y_i = log(x_i)
+    // dy_i / dx_j = delta_ij 1/x_i
+
+    Core::LinAlg::Matrix<3, 3> id2(true);
+    for (int i = 0; i < 3; i++) id2(i, i) = 1.0;
+    //  // --------------------------------- switch by number of equal eigenvalues
+    if (abs(EW(0, 0) - EW(1, 1)) < EWtolerance &&
```
c-p-schmidt, line 878: Also below?
```suggestion
    if (std::abs(EW(0, 0) - EW(1, 1)) < EWtolerance &&
```

c-p-schmidt (commented):
Thank you for this contribution. I left some commends.

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
dragos-ana: As discussed offline, this is not really my code, but these are leftovers - the renaming of the previous files
`4C_linalg_utils_densematrix_exp_log.{hpp, cpp}` to `4C_linalg_utils_densematrix_funct.{hpp, cpp}` did not work, strangely.
Nonetheless, I agree that this should not be 0.0, but a "numerical" 0. Thus, I adapted it.

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
dragos-ana: Same as above

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
dragos-ana: Again, not my code. But fixed now, thank you!

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
dragos-ana: Fixed

dragos-ana:
After discussions with @c-p-schmidt and @sebproell, I addressed the issue of optional parameters (such as the Padé order, which is only required for some specific computation methods) via function overloading. Moreover, the `.cpp` file has been reorganized - smaller evaluation functions were created for each computation method separately and are called within their respective function.

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1769 @@
...
+      n++;
+      facn *= n;
+      tmp.multiply(1.0, tmp2, input, 0.0);
+      tmp2 = tmp;
+      output.update(1. / facn, tmp, 1.);
+    }
+
+    // throw error if no convergence is reached after the maximum number
+    // of terms
+    if (n > n_max)
```
c-p-schmidt, line 187: Is this even possible to reach? In the while statement, you check for n<n_max and increment afterward, so the maximum value of n should be n_max, right?

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1769 @@
...
+    {
+      nIter++;
+      nfac *= nIter;
+      fac.push_back(nfac);
+      tmp1.multiply(tmp2, input);
+      Xn.push_back(tmp1);
+      tmp2 = tmp1;
+    }
+
+    if (nIter > 50)
```
c-p-schmidt, line 278: Same here and further below.

src/core/linalg/src/dense/4C_linalg_utils_densematrix_funct.cpp
```diff
@@ -0,0 +1,1769 @@
...
+
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+void Core::LinAlg::sym_matrix_3x3_exp_2nd_deriv_voigt(const Core::LinAlg::Matrix<3, 3>& input,
+    Core::LinAlg::Matrix<3, 3>& exp, Core::LinAlg::Matrix<6, 6>& dexp_mat,
+    Core::LinAlg::Matrix<6, 6>* ddexp_mat, Core::LinAlg::MatrixFunctErrorType& err_status)
+{
+  // declare arrays of derivatives
+  Core::LinAlg::Matrix<3, 3> matrix_exp_1st_deriv[6];
```
c-p-schmidt, line 1627: Why don't you use a std::array?
```suggestion
  std::array<Core::LinAlg::Matrix<3, 3>,6> matrix_exp_1st_deriv;
```

c-p-schmidt (dismissed):

dragos-ana:
@c-p-schmidt Thanks again for the review! While I certainly agree with your remaining comments, I must again point out that all of the exponential evaluation routines are not my code - they have only been simply relocated within the framework I created. Their improvement was not the goal of this PR, but this is quite hard to see due to the fact that the renaming of the file has not functioned properly in git. I would prefer the improvement of these routines to be part of a separate PR.

dragos-ana:
The version I just pushed contains another commit with necessary changes for rebasing on the current main.

c-p-schmidt (approved):

c-p-schmidt:
Can someone have a look?

lauraengelhardt (approved):
Looks good, thanks for tackling this!

PR #531, opened 2025-03-17
BugFix: Add missing ConditionType enum strings

ischeider:
## Description and Context

In the `Core::Conditions` class, a method `to_string` exists for `enum ConditionType`.
However, not all condition types have a corresponding string in the respective switch clause.
For now I am confident I added all missing ones. 

However, there is no guarantee that people don't add their new condition in the to_string method.

### DIFF
--- a/src/core/legacy_enum_definitions/4C_legacy_enum_definitions_conditions.cpp
+++ b/src/core/legacy_enum_definitions/4C_legacy_enum_definitions_conditions.cpp
@@ -31,6 +31,8 @@ std::string_view Core::Conditions::to_string(const Core::Conditions::ConditionTy
       return "Surface Neumann boundary condition";
     case Core::Conditions::VolumeNeumann:
       return "Volume Neumann boundary condition";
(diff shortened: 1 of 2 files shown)


sebproell (dismissed):
This should be done with magic_enum and, as you realized, we should not bother with this kind of manual housekeeping. In case this is meant for docs, we can do it but then the docs are rather poor in the present case.

ischeider:
I discussed it with Georg. The problem here is that the enums are sometimes different from their strings, so that we loose (just a bit) of information. The remark is anyway valid. 

In the directory `legacy_enum_definitions`, none of the enums use the magic_enum feature, right? And some are enum classes, some are not. Therefore, I would like to keep it as it is for now.

sebproell:
Sure, we keep it here -- that's why I approved ;) Just wanted to make you aware that we have magic_enum.

mayrmt (dismissed):

sebproell (approved):

mayrmt (approved):

PR #541, opened 2025-03-19
445 reduce size of docker image

### LINKED ISSUE #445: Reduce the size of the docker images
## Description
Our docker images are quite large.
- 7.53 GB for `4c:main`
- 4.24 GB for `4c-dependencies-ubuntu24.04:main`

This is a problem, e.g., when the 4C image is used in other projects.
- Queens wants to use the prebuilt 4C image in their GitHub Actions. But they are limited by the available runner storage of 14 GB where the 4C image already consumes more than half of the storage.
- meshpy wants to use the 4C image with mybinder. But the image is too large such that mybinder runs into a timeout.

One immediate solution is to build a docker image with only the 4C executable. Removing the 4C source folder will already save 2.5 GB.

The docker image with just the 4C executable should be our default image. It should be as small as possible. We can still have our current 4C image with the source code and the tests. We could name it `4c-full:main`. This image is great for quickly checking out 4C. 

- [ ] Create a minimal image to run 4C (without source code and tests). This will be `4c:main`
- [ ] Create an image with the full 4C build and source code. Formerly, `4c:main` -> `4c-full:main`
- [ ] Try to reduce dependencies image as well

## Interested Parties
@isteinbrecher @gilrrei @danielwolff1

ppraegla:
## Description and Context
Provide a new docker image with only the 4C executable. Some unnecessary packages which are not required to run 4C, like paraview and clang, have been removed. The image with the full 4C build `ghcr.io/4c-multiphysics/4c:main` is 7.53GB. The new `4c-minimal:main` image is now at 2.73GB. Right now, I just remove things from the `4c-dependencies` image. This can be refactored in the future once we have a better way to manage our different docker images.

Additional info
- The PR also fixes an "error" for the prebuilt 4C image. As this image is tagged with main, it should also depend on the dependencies image `main`. This has not caused any harm yet because the dependencies `latest` and `main` tags point to the same image.
- The 4C `install` target only installs the 4C executable. So, `pre_exodus` or the `post-processor` are not available in the `4c-minimal` image.
- The changes of this PR will only be visible once it is merge. But I tested it in my fork. Trust me ;) 
- You can check out the image in [my fork](https://github.com/users/ppraegla/packages/container/4c-minimal/377400868?tag=main)
- This is only a first shot at reducing the image size. It should be possible to reduce the size even further when we remove everything only required for building 4C (gcc, all the *-dev packages, ...)
- This image should be used in GitHub actions when you need a prebuilt 4C

Closes #445

### DIFF
--- a/docker/prebuilt_4C_minimal/Dockerfile
+++ b/docker/prebuilt_4C_minimal/Dockerfile
@@ -1,0 +1,44 @@
+# This file is part of 4C multiphysics licensed under the
+# GNU Lesser General Public License v3.0 or later.
+#
+# See the LICENSE.md file in the top-level for license information.
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+FROM ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:main
+
+USER root
+
+# Remove unnecessary packages
+RUN apt-get update && apt-get -y remove \
+      doxygen \
+      clang* \
+      llvm* \
+  && apt-get -y autoremove \
+  && rm -rf /var/lib/apt/lists/* \
+  && rm -rf /usr/share/doc
+
+# Remove testing dependencies
+RUN rm -r /opt/4C-dependencies-testing/
+
+RUN git clone --depth 1 https://github.com/4C-multiphysics/4C.git 4C-source
+
+RUN mkdir 4C-source/build && cd 4C-source/build && \
+  cmake ../ --fresh -DCMAKE_INSTALL_PREFIX=/home/user/4C --preset=docker_minimal && \
+  cmake --build . --target install -- -j `nproc`
+
+RUN ln -s /home/user/4C/bin/4C /usr/local/bin/4C
+
+# Remove build artifacts
+RUN rm -r /home/user/4C-source
+
+# Copy everything to a new image to get rid of the intermediate layers
+FROM scratch
+COPY --from=0 / /
+LABEL org.opencontainers.image.description="Image containing the 4C executable"
+LABEL org.4c-multiphysics.project=4C
+
+USER user
+WORKDIR /home/user
+
+CMD [ "/bin/bash" ]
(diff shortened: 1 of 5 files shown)


docker/prebuilt_4C_minimal/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# Copy everything to a new image to get rid of the intermediate layers
+FROM scratch
+COPY --from=0 / /
+LABEL org.opencontainers.image.description="Image containing the 4C executable"
+LABEL org.4c-multiphysics.project=4C
+
+USER user
+WORKDIR /home/user
+
+CMD [ "/bin/bash" ]
```
amgebauer, line 42: Does it make sense to define 4C as the entry point of the image?

amgebauer (dismissed):

docker/prebuilt_4C_minimal/Dockerfile
ppraegla: I'm not sure. I would say that `ENTRYPOINT` is not what we want right now. This means that the docker image always executes 4C. I don't know how well this plays with the GitHub Actions.
So, I would leave it as is.

FYI, all the other images also define `CMD [ "/bin/bash" ]`. I just had to define it here because I used the `scratch` image

davidrudlstorfer: Could we define an alias within the image, i.e., `4C`?

ppraegla: It would probably make sense to create a symlink and put it in a directory that is in the search path so you can call 4C with just `4C`.

amgebauer (approved):

gilrrei:
Thanks from QUEENS-side for the fix. Even though we had a workaround related to the big image, I think that for the future this is super valuable. I love it when the communities interact :family:

sebproell (approved):
Very nice and I like the README :+1:

Issue #506, opened 2025-03-13
Add (optional) deal.II dependency


sebproell:
## Description

At LNM, we plan to use deal.II in combination with 4C for a few projects. I will present some more details on these projects soon-ish. Regardless, we will require deal.II as a dependency. At first, this should be optional and only if necessary and generally useful, it will become mandatory. This is a good opportunity to test next steps in modularization. 

I will add a few conversion mechanisms between 4C Discretization and deal.II Trangulation alongside the changes, to ensure that everything indeed works.

## Possible Solution and Definition of Done

- [ ] Add cmake options to enable deal.II (related to #483)
- [ ] Add an image that allows to test with deal.II
- [ ] Add first batch of conversion functions to really test deal.II in 4C

sebproell:
Let me take this comment to give some information about my ideas for the interaction of deal.II and 4C:

- I already have a few tested converters from 4C to deal.II discretizations and associated vectors. They will serve as a basis to try out various ideas listed below. I will merge these soon, once I have the testing infrastructure in place.

- I will investigate how we can modernize our discretization, element, node, material, ... infrastructure. The implementation is coming from a time when hardware demands were just different. This kind of design cannot exploit hardware to its full extent and introduces overhead. Also, if you have ever written something based on the discretization you may have realized it is quite cumbersome to work with. C++ has evolved significantly in the last 20 years and there is more knowledge for good designs. Using a deal.II Triangulation internally, may be a route to unlock new features at a rather small implementation cost. For this project, I am mainly concerned with the more classical finite elements where `dim==spacedim`, not so much with particles or reduced-dimensional models.

- A first potential payoff from all of this, is the ability to have adaptive mesh refinement in 4C.

- A second potential payoff is the ability to concentrate on our core competencies: coupled multiphysics. deal.II is specifically built to provide FE infrastructure. I find that very interesting stuff, but I guess most of our users want to build their coupled models and don't necessarily care about writing nice helpers to evaluate shape functions. This is something deal.II does very well. Now that the taskforce solid elements brought the elements in shape, a logical step could be to make them simpler, more generic and more performant (maybe not everything to the same degree ;) Note: this would of course be an intrusive change and is unrealistic as long as I keep the dependency optional.

[...]

PR #545, opened 2025-03-20
Build depencies + deal.II image

sebproell:
Part of #506 

Build a separate dependency image with deal.II added on top of our other dependencies.

### DIFF
--- a/docker/dependencies_dealii/Dockerfile
+++ b/docker/dependencies_dealii/Dockerfile
@@ -1,0 +1,32 @@
+# This file is part of 4C multiphysics licensed under the
+# GNU Lesser General Public License v3.0 or later.
+#
+# See the LICENSE.md file in the top-level for license information.
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+FROM ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:latest
+
+USER root
+
+# Install deal.II
+
+# Create directory for dependencies
+ARG NPROCS=12
+ENV NPROCS=$NPROCS \
+    INSTALL_DIR="/opt/4C-dependencies"
+
+COPY dependencies /dependencies
+# Install cmake
+RUN /dependencies/dealii/install.sh ${INSTALL_DIR}
+
+# Copy everything to a new image to get rid of the previous layers
+FROM scratch
+COPY --from=0 / /
+LABEL org.opencontainers.image.description="Image containing all the dependencies required for building and testing 4C with deal.II"
+LABEL org.4c-multiphysics.project=4C
+
+USER user
+WORKDIR /home/user
+
+CMD [ "/bin/bash" ]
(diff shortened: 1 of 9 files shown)


docker/dependencies_dealii/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+COPY dependencies /dependencies
+# Install cmake
+RUN /dependencies/current/dealii/install.sh ${INSTALL_DIR}
+
+# Copy everything to a new image to get rid of the previous layers
+FROM scratch
+COPY --from=0 / /
+LABEL org.opencontainers.image.description="Image containing the 4C executable"
+LABEL org.4c-multiphysics.project=4C
```
sebproell, line 27: @ppraegla What should I put here?

dependencies/current/dealii/install.sh
```diff
@@ -0,0 +1,58 @@
...
+
+git clone https://github.com/dealii/dealii.git
+cd dealii
+git checkout $VERSION
+mkdir build
+cd build
+
+# Install p4est as part of deal.II
+wget --no-verbose https://p4est.github.io/release/p4est-2.8.tar.gz
+../doc/external-libs/p4est-setup.sh p4est-2.8.tar.gz $INSTALL_DIR
```
sebproell, line 33: This script internally untars the archive, which gives some ownership warnings. The resulting image worked for me locally, so I did not try to fix that.

sebproell (commented):
I tested this locally and I was able to compile and run tests for a branch withe 4C-deal.II converter tests. Before I build the images: @ppraegla Can I get a review if this is how I should set up the images?

dependencies/current/dealii/install.sh
```diff
@@ -0,0 +1,58 @@
...
+  -D DEAL_II_WITH_TRILINOS=ON \
+  -D DEAL_II_WITH_BOOST=ON \
+  -D DEAL_II_FORCE_BUNDLED_BOOST=ON \
+  -D TRILINOS_DIR=$INSTALL_DIR \
+  -D P4EST_DIR=$INSTALL_DIR/FAST \
+  ..
+
+make -j${NPROCS} install
+cd ..
+rm -rf dealii
```
ppraegla, line 58: Does this work? Is another `../` missing? With `../` you go from the `build` directory to the `dealii` directory.

docker/dependencies_dealii/Dockerfile
ppraegla: Something like this?
```suggestion
LABEL org.opencontainers.image.description="Image containing all the dependencies required for building and testing 4C with dealii"
LABEL org.4c-multiphysics.project=4C
```

ppraegla (changes requested):
The overall Dockerfile looks good. 

I think it would make sense to put the dealii dependency in `dependencies/dealii` and not under `current`. For me, the dealii dependency has an experimental status like trilinos_develop. This would also allow you to work with this dependency more freely as you don't have to adapt the dependencies hash every time you change something (This requires that the utility function to compute the hash is adapted once). We can still move `dealii` to `current` in the future if it becomes necessary.

PR #551, opened 2025-03-21
Emit knowledge about elements to metadata

sebproell:
We know many things about the elements we can input, so write it out in the metadata file.

### DIFF
--- a/src/core/fem/src/general/element/4C_fem_general_element_definition.hpp
+++ b/src/core/fem/src/general/element/4C_fem_general_element_definition.hpp
@@ -45,6 +45,8 @@ namespace Core::Elements
     /// return line definitions for given element type
     const Core::IO::InputSpec& element_lines(std::string name, std::string cell_type);
 
+    const auto& definitions() const { return definitions_; }
+
    private:
     /// input line definitions per element type
     std::map<std::string, std::map<std::string, Core::IO::InputSpec>> definitions_;
--- a/src/core/io/src/4C_io_input_file.cpp
+++ b/src/core/io/src/4C_io_input_file.cpp
@@ -262,6 +262,11 @@ namespace Core::IO
       std::map<std::string, InputSpec> valid_sections_;
       std::vector<std::string> legacy_section_names_;
 
+      /**
+       * Additional specs for legacy sections that are not fully known.
+       */
+      std::map<std::string, InputSpec> legacy_partial_specs_;
+
       bool is_section_known(const std::string& section_name) const
       {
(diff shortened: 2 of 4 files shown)


gilrrei (approved):
Looks good, thanks for the work

gilrrei:
> Small example how this looks, @gilrrei
> 
> ```
> legacy_element_specs:
>   type: one_of
>   specs:
>     - name: ALE2
>       type: group
>       required: true
>       defaultable: false
>       specs:
>         - type: one_of
>           description: 'one_of {QUAD4, QUAD8, QUAD9, TRI3, TRI6}'
> ```

Wouldn't it make sense to use selection instead of one_of for `QUAD4, QUAD8, ...`? If  the element parameters are the same independent of the element shape functions, maybe something like:
```yaml
legacy_element_specss:
  type: one_of
  specs:
    - name: ALE2
      type: group
      required: true
      defaultable: false
      specs:
        - name: element # with some better name
          description: "selection for QUAD4, QUAD8, QUAD9, TRI3, TRI6"
          type: selection
          required: true
          selector: type
          choices:
            - name: QUAD4
              spec:
                - name: nodes
                  type: vector
                  size: 4
                  value_type:
                    type: int
                  required: true
            - name: QUAD8
              spec:
                - name: nodes
                  type: vector
                  size: 8
                  value_type:
                    type: int
                  required: true
            - name: QUAD9
              spec:
                - name: nodes
                  type: vector
                  size: 9
                  value_type:
                    type: int
                  required: true
        - name: MAT # this is an element parameter that is not related to the shape functions
          type: int
          required: true
```

sebproell:
> Wouldn't it make sense to use selection

Yes, I think it would make sense. However, the input currently looks like this:

`<id> <element_type> QUAD4 1 2 3 4 ...`

So, there is nothing existing that can be interpreted as the `selector` here. The distinguishing `QUAD 4` is already the key for the node IDs. I don't see a backward compatible way to introduce this.

gilrrei:
> > Wouldn't it make sense to use selection
> 
> Yes, I think it would make sense. However, the input currently looks like this:
> 
> `<id> <element_type> QUAD4 1 2 3 4 ...`
> 
> So, there is nothing existing that can be interpreted as the `selector` here. The distinguishing `QUAD4` is already the key for the node IDs. I don't see a backward compatible way to introduce this.

I thought so. No worries, this part will then move to FourCIPP. Since the long-term goal is to remove the geometry altogether, I think this is fair.

davidrudlstorfer (approved):

Issue #552, opened 2025-03-21
[LinAlg] Reduce Epetra_Map and Epetra_BlockMap occurances


knarfnitram:
## Description

Create a wrapper named `4C_linalg_map.*` to centralize all `Epetra_Map` calls.  

## Possible Solution and Definition of Done

Related to `Epetra_Map`:
- [x] Provide the Wrapper and replace occurrences (see https://github.com/4C-multiphysics/4C/pull/554/) 
- [x] Adjust to style convention ( ` remove // NOLINTBEGIN(readability-identifier-naming)` for `4C_linalg_map.*` )  (see #817)
- [x] Try to remove the function `domain_map_not_epetra`  by adjusting the inheritance of `SparseMatrixBase`
- [x] Before the `Epetra_BlockMap` can be replaced, usa a view as introduced in #557 for classes  `Graph`, `Multivector`, `BlockSparseMatrixBase` and ` SparseMatrixBase` (Search for `mutable std::optional<Core::LinAlg::Map> `)

Related to `Epetra_BlockMap`:
- [x] utilize regex search and  replace `Epetra_BlockMap` by `Core::LinAlg::Map`, but also the respective `get_block_map()` function call by `get_map()` (see #799)
- [x] remove the member function `get_block_map()` from `Core::LinAlg::Map` (see #799)

PR #555, opened 2025-03-23
Rename get_map to get_block_map of Vector

knarfnitram:
## Description and Context

Renames the existing `get_map()` function to `get_block_map()` as it currently returns the `Epetra_BlockMap` object.

Part of #552

### DIFF
--- a/apps/post_monitor/4C_post_monitor.cpp
+++ b/apps/post_monitor/4C_post_monitor.cpp
@@ -524,7 +524,7 @@ void FluidMonWriter::write_result(
 {
   // get actual result vector
   auto resvec = result.read_result("velnp");
-  const Epetra_BlockMap& velmap = resvec->get_map();
+  const Epetra_BlockMap& velmap = resvec->get_block_map();
   // do output of general time step data
   outfile << std::right << std::setw(20) << result.step();
   outfile << std::right << std::setw(20) << std::scientific << result.time();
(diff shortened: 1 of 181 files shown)


sebproell (dismissed):

mayrmt:
@knarfnitram this new name appears to expose an implementation detail to the user. I am wondering: Is it important to let the user know about the difference? 

I don't remember to have ever cared about the difference between map and block map, but that may have changed in the meantime.

sebproell:
We are returning `Epetra_BlockMap`, which is the more general type as `Epetra_Map` derives from it. So the new name reflects that. I assumed this change is more for our own sanity than for the user. @knarfnitram could you briefly explain if there is a deeper reason for renaming?

knarfnitram:
Yes, so I think (or hope ) that it should not matter, but I am not 100% sure.
 Therefore I am trying to change as few objects as possible and not mix things up, renaming this function helps with preserving the current status. 
 In a next step one can either create a wrapper for the `BlockMap` or simply try to add it to this interface here and see what happens.

mayrmt:
I don't mind the change. I am just wondering whether we need the distinction. Do we ever explicitly use a block map object? Do we really need a wrapper for it?

mayrmt (approved):

sebproell:
AFAIK, Tpetra only has `Map`. Thus, I'd hope we can get rid of the distinction as well.

sebproell (approved):

PR #554, opened 2025-03-22
Provide Wrapper for the Epetra map 

knarfnitram:
## Description and Context

Add the wrapper of the Epetra Map Function to 4c.
- Rename of Vector function(see #555 )
- Replace occurrences of `EpetraMap` with `Map` 
- Fix Map access across code base

## Related Issues and Pull Requests

First step of #552 .

PR #557, opened 2025-03-23
Create LinAlg::View for our wrappers

sebproell:
We frequently want to view Epetra datastructure as if they were our own wrapper. This class enables such a behavior with minimal modifications in the wrapper classes. We already had a basic version for vectors and I realized we can generalize it a little bit, to work for any combination of source type and wrapper type. 

Should help with some issues in #554

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_vector.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_vector.hpp
@@ -12,6 +12,7 @@
 #include "4C_config.hpp"
 
 #include "4C_linalg_multi_vector.hpp"
+#include "4C_linalg_view.hpp"
 
 #include <Epetra_FEVector.h>
 #include <Epetra_IntVector.h>
@@ -20,13 +21,10 @@
 #include <memory>
 
 
-
 FOUR_C_NAMESPACE_OPEN
 
 namespace Core::LinAlg
 {
-  template <typename T>
-  class VectorView;
 
   // Sparse Vector which will replace the Epetra_Vector
   template <typename T>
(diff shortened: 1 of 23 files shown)


src/core/linalg/src/sparse/4C_linalg_view.hpp
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * A helper struct to easily specify that one of our linear algebra classes is a wrapper for
+   * another class. This struct should be specialized for each of our linear algebra classes and
+   * contain a single typedef `type` for the class that is wrapped.
+   */
+  template <typename T>
+  struct WrapperFor
+  {
+    static_assert("You need to specialize this struct for the wrapper class.");
+  };
```
sebproell, line 31: We specialize this struct for our wrappers as you can see for Vector and MultiVector in this PR.

src/core/linalg/src/sparse/4C_linalg_vector.hpp
```diff
@@ -304,6 +302,24 @@ namespace Core::LinAlg
...
+      ret->vector_ = std::make_shared<Epetra_Vector>(Epetra_DataAccess::View, view, 0);
+      return ret;
+    }
+
+    [[nodiscard]] static std::shared_ptr<const Vector<T>> create_view(const Epetra_Vector& view)
+    {
+      std::shared_ptr<Vector<T>> ret(new Vector<T>);
+      ret->vector_ = std::make_shared<Epetra_Vector>(Epetra_DataAccess::View, view, 0);
+      return ret;
+    }
```
sebproell, line 320: We want to support const and non-const views, so both are created here.

src/core/linalg/src/sparse/4C_linalg_vector.hpp
```diff
@@ -407,50 +411,12 @@ namespace Core::LinAlg
...
-    std::shared_ptr<VectorType>& get_non_owning_rcp_ref() { return view_vector_; }
-
-   private:
-    //! Store inside an RCP because our interfaces frequently take references to RCPs.
-    std::shared_ptr<VectorType> view_vector_;
+  template <>
+  struct WrapperFor<Epetra_Vector>
+  {
+    using type = Vector<double>;
   };
```
sebproell, line 419: Here is the specialization. Together with the static method, this is enough for Views to work :)

davidrudlstorfer (approved):
fancy

knarfnitram (approved):
Thanks for generalizing the view @sebproell , I 'll take a look, how I can utilize this for #554 .

PR #558, opened 2025-03-24
Build an image with deal.II

sebproell:
I forgot I need to use a special branch name to update the image. This replaces #545

Part of #506


docker/dependencies_dealii/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# This file is part of 4C multiphysics licensed under the
+# GNU Lesser General Public License v3.0 or later.
+#
+# See the LICENSE.md file in the top-level for license information.
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+FROM ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:latest
```
sebproell, line 8: @ppraegla I just use latest here. I have not included a dependency on the other dependency image. Is this OK? This means that I manually need to rebuild the deal.II image after updating the other dependencies. I thought we should not over-engineer the workflow at this point but I would be curious to get your feedback on this.

ppraegla: You should use the tag `main`. We use `latest` to tag the latest release version. But apart from that your approach sounds good.

amgebauer (commented):
As discussed offline, the deal.ii-docker image should directly build after the regular docker dependencies image is build. Having a second manual intervention is error prone.

docker/dependencies_dealii/Dockerfile
ppraegla: I don't see a use case where we would depend an image on the latest release image. So, I would leave it out and introduce it once we need it.

amgebauer: I don't really see how this should work if we update a dependency. We need to rebuild our dependencies image in order to run tests. Since we also want to run tests with deal.ii, we also need to update the dependencies image containing deal.ii. This needs to happen before we merge the changes in the dependencies. In that case, `main` points to the wrong base image.

ppraegla: Now I understand what you mean. The correct tag should be the dependencies hash. To avoid updating the Dockerfile with the dependencies hash, you need to pass it as a build argument to docker.

.github/workflows/docker_dealii.yml
```diff
@@ -0,0 +1,52 @@
+name: Docker with deal.II
+
+on:
+  workflow_dispatch:
```
ppraegla, line 4:
```suggestion
on:
  workflow_dispatch:
  workflow_run:
    workflows: [docker]
    types: [completed]
```
You could add this to automatically build the dealii image after the docker workflow built the dependencies image. (Please check if this makes sense)

docker/dependencies/compute_dependencies_hash.sh
```diff
@@ -9,4 +9,4 @@
 # Exit the script at the first failure
 set -e
 
-find dependencies docker/dependencies -not -wholename '*/trilinos_develop/*' -not -name 'README.md' -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8
+find dependencies docker/dependencies -not -wholename '*/trilinos_develop/*' -not -wholename '*/dealii/*' -not -name 'README.md' -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8
```
ppraegla, line 12: It may make sense to explicitly name what to folders to check instead of excluding all other folders.
```suggestion
find dependencies/current dependencies/testing docker/dependencies -not -name 'README.md' -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8
```
This should give you the same files. Please double-check

sebproell:
@ppraegla @amgebauer Thanks for the guidance! I now updated everything and rebuild the image. Let me know what you think about the generalizations I made. I am not sure at which point one should stop to abstract the actions and just make a copy for the new case.

.github/workflows/docker.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       id-token: write
       #
     needs: check-if-ubuntu24_04-build-is-required
-    if: github.event_name == 'workflow_dispatch' && ${{ needs.check-if-ubuntu24_04-build-is-required.outputs.build_docker_image
+    if: ${{github.event_name == 'workflow_dispatch' && needs.check-if-ubuntu24_04-build-is-required.outputs.build_docker_image
```
sebproell, line 49: This was a bug that I fixed along the way.

.github/workflows/docker_dealii.yml
```diff
@@ -2,19 +2,64 @@ name: Docker with deal.II
 
 on:
   workflow_dispatch:
+  workflow_run:
+    workflows: [Docker]
+    types: [completed]
```
sebproell, line 7: I can't get this to work. I've built the image manually but the docker_dealii workflow is not triggered after the Docker workflow. I already tried both spellings, `docker` and `Docker`.

amgebauer: See my comment above. I think it only runs if the workflow is on the default branch (and then it also runs on the state of the default branch). I think this is not what we want.

amgebauer (commented):
It looks like I forgot to submit my previous comments.

sebproell:
Looks like this now: https://github.com/4C-multiphysics/4C/actions/runs/14133399924

amgebauer (dismissed):

docker/dependencies/compute_dependencies_hash.sh
```diff
@@ -9,4 +9,4 @@
 # Exit the script at the first failure
 set -e
 
-find dependencies docker/dependencies -not -wholename '*/trilinos_develop/*' -not -name 'README.md' -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8
+find dependencies/current dependencies/testing docker/dependencies -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8
```
ppraegla, line 12: Should we keep excluding a possible README.md? Otherwise, we would need to adapt the script again if we decide to add more documentation. Though, I don't feel super strong about it.
```suggestion
find dependencies/current dependencies/testing docker/dependencies -not -name 'README.md' -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8
```

docker/dependencies_dealii/compute_dependencies_hash.sh
```diff
@@ -0,0 +1,12 @@
...
+# GNU Lesser General Public License v3.0 or later.
+#
+# See the LICENSE.md file in the top-level for license information.
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+# Exit the script at the first failure
+set -e
+
+find dependencies/current dependencies/dealii dependencies/testing docker/dependencies docker/dependencies_dealii -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8
```
ppraegla, line 12: Same as above
```suggestion
find dependencies/current dependencies/dealii dependencies/testing docker/dependencies docker/dependencies_dealii -not -name 'README.md' -type f -exec sha1sum {} \; | sort | sha1sum | cut -c -8
```

.github/workflows/docker.yml
```diff
@@ -59,9 +59,61 @@ jobs:
...
           image-name: ${{ env.REGISTRY }}/${{ env.PROJECT_NAMESPACE }}/4c-dependencies-ubuntu24.04
-          dependencies-hash: ${{ needs.check-if-ubuntu24_04-build-is-required.outputs.dependencies_hash
+          dependencies-hash: ${{ needs.check-if-build-dependencies-is-required.outputs.dependencies_hash
             }}
-          ubuntu-version: "24.04"
+          base-image: "ubuntu:24.04"
+
+  check-if-build-dependencies-dealii-is-required:
+    runs-on: ubuntu-latest
+    if: github.ref_name != 'main'
```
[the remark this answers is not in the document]
sebproell: That's from a prior version. Changed.

docker/dependencies/compute_dependencies_hash.sh
sebproell: Well, I would need to adapt the hash now for this ;) I would keep it.

ppraegla: Yes. I'm fine if you keep it. But we should keep it in mind.

amgebauer (approved):

PR #561, opened 2025-03-25
Add dummy job to set up docker image with deal.II

sebproell:
I need to get this workflow onto main so I can manually trigger the build within https://github.com/4C-multiphysics/4C/pull/558.

Part of #506

### DIFF
--- a/.github/workflows/docker_dealii.yml
+++ b/.github/workflows/docker_dealii.yml
@@ -1,0 +1,20 @@
+name: Docker with deal.II
+
+on:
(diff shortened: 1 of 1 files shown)


ppraegla (approved):

amgebauer:
What is the exact problem that you solve with this dummy workflow?

sebproell:
I cannot start a workflow to build an image if the workflow is not on main.

amgebauer (approved):

Issue #539, opened 2025-03-19
[BUG] roughcontact2d_mirco_varying_roughness fails with clang compiler


amgebauer:
## Bug Description: What's going wrong?

The test `` fails with the clang-compiler resulting in failure of our nightly pipeline and the coverage pipeline

The error is

```
====== Restart for field 'Structure' written in step 1
Checking results of 4 tests:
STRUCTURE: structure   dispx    at node   9	 is WRONG --> actresult= 4.70579746827954670e-01, givenresult= 4.70518611340847825e-01, abs(diff)= 6.11354871068447991e-05 > 1.00000000000000004e-10
STRUCTURE: structure   dispy    at node   9	 is WRONG --> actresult=-1.56951023417044033e+00, givenresult=-1.56950026119692709e+00, abs(diff)= 9.97297351323922499e-06 > 1.00000000000000006e-09
STRUCTURE: structure   dispx    at node 197	 is WRONG --> actresult= 4.70836795356221416e-01, givenresult= 4.70776232133924499e-01, abs(diff)= 6.05632222969165745e-05 > 1.00000000000000004e-10
STRUCTURE: structure   dispy    at node 197	 is WRONG --> actresult=-8.43002552254935367e+00, givenresult=-8.43003556702593393e+00, abs(diff)= 1.00444765802620850e-05 > 1.00000000000000006e-09
```

## Expectations

- [ ] Tests should pass

## Your Environment

- **Relevant repo SHA1s:**
- **Relevant configure flags or configure script:**
- **Operating system and version:**
- **Compiler and TPL versions:**

## Related Issues and Pull Requests

- Follows #535 

## Interested Parties

@RShaw026

sebproell:
- This also fails with MIRCO's main branch.
- It fails in debug and release.
- It fails with `-ftrapping-math` turned on.

amgebauer:
The deploy job now also fails and our documentation is now not up-to-date anymore.

ppraegla:
Should we remove the job temporarily? There does not seem to be an apparent fix for the problem.
Due to the failing nightly pipeline, the new minimal docker image is also not built.

RShaw026:
Should I design another test which uses the same functionalities see if we get the same issue? @isteinbrecher

sebproell:
>  Should I design another test which uses the same functionalities and see if we get the same issue?

[...]

PR #562, opened 2025-03-25
Disable failing test with unclear fix

sebproell:
See #539


davidrudlstorfer (approved):

amgebauer (approved):

RShaw026:
I am okay with disabling the test for now. I will write another test with different parameters and check if we run into the same issue with the tolerance.

RShaw026 (approved):

PR #572, opened 2025-03-26
Only communicate small sections when reading yaml

sebproell:
So far we always copied the whole internal representation of the yaml tree to all ranks. That is certainly not ideal. This PR only copies sections if they are small. The large sections (which will vanish eventually with #60) are not copied. They are only processed on rank 0 in the code.

@ppraegla Can you give this a try for the large particle example you mentioned?

FYI @isteinbrecher

### DIFF
--- a/src/core/io/src/4C_io_input_file.cpp
+++ b/src/core/io/src/4C_io_input_file.cpp
@@ -33,6 +33,30 @@ namespace Core::IO
   namespace
   {
     std::string to_string(const ryml::csubstr str) { return std::string(str.data(), str.size()); };
+
+    // Copy the content of the source node to the destination node. Useful to copy a node from one
+    // tree to another.
+    void deep_copy(ryml::NodeRef src, ryml::NodeRef dst)
+    {
+      dst.set_type(src.type());
+
+      if (src.has_key())
+      {
+        dst.set_key(src.key());
+      }
+      if (src.has_val())
+      {
+        dst.set_val(src.val());
+      }
+      // Recursively copy children
+      for (size_t i = 0; i < src.num_children(); ++i)
+      {
+        c4::yml::NodeRef src_child = src.child(i);
+        c4::yml::NodeRef dst_child = dst.append_child();
+        deep_copy(src_child, dst_child);
+      }
+    };
+
   }  // namespace
 
   namespace Internal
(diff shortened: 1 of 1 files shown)


isteinbrecher:
How is this related to #241?

I always thought that we do the full input reading on rank 0. What data is build up locally at each rank from the yaml file and cannot be communicated by communicating our 4C internal objects?

sebproell:
> I always thought that we do the full input reading on rank 0. 

Yes, that's correct.

> What data is build up locally at each rank from the yaml file and cannot be communicated by communicating our 4C internal objects?

No rank other than rank 0 ever reads the file. In consuming code, interaction with the input file always goes over our class `InputFile`. This class reads the file on rank 0, sends the data to all ranks, and the ranks can query it. Only the geometry sections, or legacy string sections, as we started to call them are _not_ communicated. As you might know, we process these sections only on rank 0 and send out the finished objects (elements, nodes) to other ranks. This was not changed in any fundamental way when yaml was added. 

However, I think we can simplify the logic a bit: we already know that the legacy sections are the problematic ones. So we can probably drop the heuristic of 10,000 lines to determine a huge section and just use the legacy section names.

> How is this related to https://github.com/4C-multiphysics/4C/pull/241?

That PR changed the logic to the 10,000 line heuristic. Back then, we internally converted yaml to dat. So the yaml branch had that logic as well. When I introduced yaml as an actual internal representation, I did not apply the heuristic right away. This is done in this PR.

isteinbrecher:
But just as a general question, why do we need to communicate yaml data and can not communicate the corresponding parameter list?

Do node set IDs fall under this (current) limit of 10000 entries? That would be bad, because then the node sets would not be available on all ranks. I think we don't have a test case that covers this in our test suite. Is this an issue?

sebproell:
> why do we need to communicate yaml data and can not communicate the corresponding parameter list?

@isteinbrecher We technically could do that but then we would need to implement some kind of serialization for these `ParameterList` or `InputParameterContainer` types. We currently don't have that. The (unparsed) input content is a kind of serialization which works perfectly well for broadcasting the information.

Regarding the nodes: let me check that. As I said, I would like to remove the huge section heuristic and clearly specify if a section is available everywhere or only on rank 0. I believe you are right that node set ids need to be processed on every rank right now.

sebproell:
@isteinbrecher I checked: we already read the nodesets on rank 0 only and communicate the result :) I will open another PR which ensures that we don't break this and makes this a little more explicit.

sebproell:
@ppraegla Timed this on a small cluster (thanks a lot for your help!):

2 nodes, with total of 128 cores:
- input 1: 17 mio particles, file 1.5 GB
    - dat (main): 2.185e+02
    - yaml (main): killed with signal 9 out of memory after 1 minute
    -  yaml(PR 572): 2.371e+02
- input 2: 10 mio particles, file 900 MB
    - dat (main) : 1.281e+02
    - yaml (main): 1.576e+02
    - yaml(PR 572): 1.392e+02

So the changes in this PR are correct :+1: After this PR we have less than 10% runtime overhead for yaml. We could probably reduce this, but I think this is not a relevant difference. What's more important is that the memory is now in check, as we no longer duplicate the internal info on all ranks.


Ready for review.

ppraegla (approved):

amgebauer (approved):

PR #581, opened 2025-03-28
Use ValueParser for parsing of nodes

sebproell:
This helps to understand what we can actually parse, so I can write this out as metadata. Also fixes a peculiar behavior: in a line like 

```
NODE 1 COORD 1.0 1.0 1.0
```

you could actually enter anything for the `COORD` token. AFAICT, only red airways entered something different here.

I tested this on my ~1e6 element example (with plain nodes) and could not see a difference in input time.


amgebauer (dismissed):

sebproell:
[This error](https://github.com/4C-multiphysics/4C/actions/runs/14133049281/job/39598329597?pr=581#step:7:126) is annoying. Seems that `stod` cannot parse subnormal doubles. We should really use `from_chars` but clang hasn't implemented that. On top of that, `from_chars` does not support a leading `+` :face_exhaling:

amgebauer (approved):

davidrudlstorfer (approved):

PR #582, opened 2025-03-28
Update custom pre-commit hook with improved file name and include checks

davidrudlstorfer:
## Description and Context

Following an idea from @sebproell the following changes emerged. The old logic of the file name and include check was not very elegant and flawed.

This PR now introduces a more concise and easier to understand pre-commit hook. The hook is now separated into a check for correct file names and a check for correct includes.

If errors are thrown it looks like this:

wrong file names:

```
C++ file names...........................................................Failed
- hook id: cpp-filename
- exit code: 1

The following files do not adhere to our file naming convention:

    src/core/fem/tests/geometric_search/4C_geometric_search_create_bounding_volumes_test.hpp
    src/core/fem/tests/geometric_search/4C_geometric_search_distributed_test.np3.cpp
    src/core/fem/tests/geometric_search/4C_geometric_search_test.cpp
    src/core/fem/tests/geometry/4C_geometry_element_volume_test.cpp
    src/core/utils/tests/functions/4C_function_library_test.cpp
    src/core/utils/tests/functions/4C_functionvariables_test.cpp
    src/core/utils/tests/functions/4C_symbolic_expression_test.cpp
    src/core/utils/tests/numerics/4C_cubic_spline_interpolation_test.cpp

Please rename the files to match the naming convention.
A valid filename must:
   - Be prefixed with '4C_' followed by the module name
   - If it is a test file, it should end with '_test'
```

wrong imports (exemplary):

```
C++ includes.............................................................Failed
- hook id: cpp-include
- exit code: 1
- files were modified by this hook

The following includes do not adhere to our include convention:

    * src/beaminteraction/src/potential/4C_beaminteraction_potential_pair_beam_to_beam.cpp:26: #include "memory.h"

    * src/beaminteraction/src/potential/4C_beaminteraction_potential_pair_beam_to_beam.cpp:27: #include "Sacado.hpp"

    Please fix the includes in the files to match the naming convention.
    A valid include must:
        - not contain relative navigation (. or ../) 
        - include external files with angle brackets

Includes from the following files have been automatically fixed:

    * src/beaminteraction/src/potential/4C_beaminteraction_potential_pair_beam_to_beam.cpp

    Please review the changes and commit them.

```


utilities/four_c_utils/src/four_c_utils/check_filenames.py
```diff
@@ -0,0 +1,90 @@
...
+            "   - If it is a test file, it should end with '_test'"
+        )
+
+        sys.exit(1)
+
+    sys.exit(0)
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
amgebauer, line 90: The function does not return

```suggestion
    main()
```

utilities/four_c_utils/src/four_c_utils/check_includes.py
```diff
@@ -0,0 +1,120 @@
...
+        print("\n    Please review the changes and commit them.\n")
+
+    if invalid_file_includes_data or fixed_file_includes_data:
+        sys.exit(1)
+
+    sys.exit(0)
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
amgebauer, line 120:
```suggestion
    main()
```

utilities/four_c_utils/src/four_c_utils/check_filenames.py
```diff
@@ -0,0 +1,90 @@
...
+        print(
+            "\nPlease rename the files to match the naming convention.\n"
+            "A valid filename must:\n"
+            "   - Be prefixed with '4C_' followed by the module name\n"
+            "   - If it is a test file, it should end with '_test'"
+        )
+
+        sys.exit(1)
+
+    sys.exit(0)
```
amgebauer, line 86:
```suggestion
```

utilities/four_c_utils/src/four_c_utils/check_includes.py
```diff
@@ -0,0 +1,120 @@
...
+        print("Includes from the following files have been automatically fixed:\n")
+        for fixed_file in fixed_file_includes_data.keys():
+            print(f"    * {fixed_file}")
+
+        print("\n    Please review the changes and commit them.\n")
+
+    if invalid_file_includes_data or fixed_file_includes_data:
+        sys.exit(1)
+
+    sys.exit(0)
```
amgebauer, line 116:
```suggestion
```

utilities/four_c_utils/src/four_c_utils/check_includes.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if ".." in line or "./" in line:
+            invalid_include_lines.append([line_no + 1, line])
+            continue
+
+        if "4C" in line:
+            match = re.search(r"#include <(4C_\w+\.hpp)>", line)
+            if match:
+                new_line = f'#include "{match[1]}"\n'
+                replace_line_in_file(file, line_no, new_line)
```
amgebauer, line 69: I think this removes and comments followed by the include which is not necessarily wanted.

```suggestion
            match = re.search(r"#include <(4C_\w+\.hpp)>(.*)", line)
            if match:
                new_line = f'#include "{match[1]}"{match[2]}'
                replace_line_in_file(file, line_no, new_line)
```

.pre-commit-config.yaml
```diff
@@ -100,18 +100,27 @@ repos:
...
-        name: C++ file names and include style
-        description: 'Check that all C++ files have the correct file name and include style'
+      - id: cpp-filename
+        name: C++ file names
+        description: 'Check that all C++ files have the correct file name'
+        language: system
+        types_or: [c++, c]
+        exclude: 'tests/cut_test/|mm$|dummy.cpp'
+        entry: utilities/python-venv/bin/check-filenames
+        require_serial: true # To give a non duplicated error output
```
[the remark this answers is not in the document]
davidrudlstorfer: Thanks for the feedback!

So pre-commit chunks all files to get some sort of "multi" processing. This in turn executes our custom pre-commit hook multiple times with a different batch of files

For the old pre-commit hooks this resulted in multiple error write outs if errors are detected. Personally, I really dislike this approach because the same error message is printed multiple times with different files and errors. If we execute it "serially" - all files are checked in one go.

If the custom pre-commit hooks introduced in this PR are executed with all files in the repo it takes a fraction of a second. Therefore, I would prefer to run all files in serial.

Let me know what you think about it:)

utilities/four_c_utils/src/four_c_utils/check_filenames.py
davidrudlstorfer: Thanks for the good catch! These were forgotten leftovers from a first draft. I've updated everything accordingly

utilities/four_c_utils/src/four_c_utils/check_includes.py
davidrudlstorfer: That's a good idea! Thanks for the specific suggestion - works flawlessly 🙏🏻

.pre-commit-config.yaml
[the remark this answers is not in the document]
sebproell: Okay, I thought so. I would still run it in parallel. I don't really care about slightly suboptimal write-outs. Otherwise, I need to run the hook multiple times if my auto-fixable files are in different batches, correct?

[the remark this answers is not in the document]
davidrudlstorfer: What do you mean with running multiple times?

Pre-commit just hands over all files in one go to the pre-commit hook rather than split in multiple "processes"

All files are checked with one pre-commit run - either with serial or "parallel" execution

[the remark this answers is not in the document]
sebproell: Sorry, I did not read your comment properly: 

> If we execute it "serially" - all files are checked in one go.

This answers my question :)

utilities/four_c_utils/src/four_c_utils/check_includes.py
```diff
@@ -0,0 +1,118 @@
...
+
+        if not line.startswith("#include"):
+            continue
+
+        if ".." in line or "./" in line:
+            invalid_include_lines.append([line_no + 1, line])
+            continue
+
+        if "4C" in line:
+            match = re.search(r"#include <(4C_\w+\.hpp)>(.*)", line)
```
sebproell, line 66: Wouldn't you precompile this regex? Not important here, I guess; I just want to learn some idiomatic python along the way.

sebproell (approved):

utilities/four_c_utils/src/four_c_utils/check_includes.py
davidrudlstorfer: I actually did not know the difference - but here https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile it is suggested that the regex is cached either way and it has no influence on the speed or anything else

amgebauer (approved):

Issue #62, opened 2024-11-29
Discuss introduction of a more powerful math/tensor library


amgebauer:
## Description

I propose to include a new math library for our computations on the Gauss point. The library should implement tensors of arbitrary rank (also symmetric tensors) and the usual tensor products. Why? The implementation gets much closer to the equations on the paper (readability) and will be faster in implementing them. We also then don't need Voigt notations.

Features that I currently miss when using `LINALG::Matrix` for tensor quantities:

* Tensors instead of a Matrix
* Symmetric tensors, so that we don't have to mess with voigt notation (and convert forth and back all the time)
* tensor products (not just standard matrix multiplications and scalar products)
* fourth order tensors for the linearization

## Possible Solution and Definition of Done

- [ ] Look for libraries out there. Candidates: [The deal.ii tensor](https://github.com/dealii/dealii/blob/master/include/deal.II/base/tensor.h), [Trilinos MiniTensor](https://trilinos.github.io/minitensor.html), [fastor](https://github.com/romeric/Fastor), [KokkosKernels](https://kokkos.org/about/kernels/) [eigen](https://gitlab.com/libeigen/eigen)
- [ ] Include it.
- [ ] Use it.

## Related Issues and Merge Requests

* Related to https://gitlab.lrz.de/baci/baci/-/issues/363 https://gitlab.lrz.de/baci/baci/-/issues/746 

## Interested Parties

@4C-multiphysics/developers

mrp089:
We're looking into `Eigen` to do that: https://github.com/SimVascular/svFSIplus/issues/297
There is also a brief discussion of `FEBio` and `deal.ii` here: https://github.com/SimVascular/svFSIplus/issues/177

amgebauer:
Thank you for your input @mrp089. I added Eigen to the list of candidates.

amgebauer:
We now have an implementation for tensors in 4C with #583 #770 #807 #812 #853 #854 #855 #862 #869 #879 #898 #902 #908 #926. TGM with slides (and recording) will be here #909

PR #583, opened 2025-03-28
Implement a Tensor class on-top of our fixed-size matrix routines

amgebauer:
This is a first draft of a tensor class in 4C. Only few tensor operations are defined.

~~Currently, I'm using [Fastor](https://github.com/romeric/Fastor) as the backend for tensor operations.~~

I'm using the same internals as the `Core::LinAlg::Matrix`. This ensures that
(a) we use the same internal ordering of the values for tensors of rank 2
(b) we don't use different math kernels for basically the same thing.
This has of course the drawback that our own implementation is less flexible and also not really optimized for current hardware. But we might want to switch to a more powerful backend and then tensors and matrices profit the same.

Note, the Tensor as suggested here is just a different object but the mathematicals are identical. Tensor operations are more flexible, but also not faster than our existing LinAlg::Matrix, except maybe that the values are not dynamically allocated.

```cpp
Core::LinAlg::Tensor<double, 3, 3> A = {{
  {1.1, 1.2, 1.3},
  {2.1, 2.2, 2.3},
  {3.1, 3.2, 3.3},
}};
Core::LinAlg::Tensor<double, 3> b = {{1.0, 2.0, 3.0}};

double det = Core::LinAlg::det(A);
double trace = Core::LinAlg::trace(A);
Core::LinAlg::Tensor<double, 3, 3> A_inv = Core::LinAlg::inv(A);

Core::LinAlg::Tensor<double, 3> c = Core::LinAlg::dot(A, b);
double d = Core::LinAlg::double_contraction(A, A);

Core::LinAlg::Tensor<double, 3, 3, 3, 3> E = Core::LinAlg::dyadic(A, A);

// supports addition of two tensors
Core::LinAlg::Tensor<double, 3, 3> B = A + A;

// also supports views to other tensors
Core::LinAlg::TensorView<double, 3, 3> A_view(A);
```

#62

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_tensor.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_tensor.hpp
@@ -1,0 +1,940 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_LINALG_TENSOR_HPP
+#define FOUR_C_LINALG_TENSOR_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_linalg_fixedsizematrix.hpp"
+#include "4C_linalg_tensor_meta_utils.hpp"
+
+#include <algorithm>
+#include <array>
+#include <concepts>
+#include <cstddef>
+#include <cstring>
+#include <span>
+#include <tuple>
+#include <type_traits>
+#include <utility>
+
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Core::LinAlg
+{
+  template <typename Tensor>
+  concept TensorConcept = requires(const Tensor& t) {
+    { t.data() } -> std::convertible_to<const typename Tensor::value_type*>;
+  };
+
+  template <typename Tensor>
+  concept Rank1TensorConcept = TensorConcept<Tensor> && Tensor::rank() == 1;
+
+  template <typename Tensor>
+  concept Rank2TensorConcept = TensorConcept<Tensor> && Tensor::rank() == 2;
+
+  template <typename Tensor>
+  concept SquareTensor =
+      Rank2TensorConcept<Tensor> && Tensor::template extent<0>() == Tensor::template extent<1>();
+
+
+  namespace Internal
+  {
+
+    enum class StorageType
+    {
+      owning,
+      view
+    };
+
+    /*!
+     * @brief General tensor class for dense tensors of arbitrary rank
+     *
+     * This class is designed for small tensors typically used in physics simulations, such as
+     * strains, stresses, and linearizations at the Gauss point level. The tensor dimensions are
+     * usually small (e.g., 2x2 or 3x3), with a rank of 1 or 2, and rarely rank 4. This
+     * results in typically 9 or fewer entries (rarely 81 for 4-tensors in 3D). The data is
+     * stored on the stack avoiding dynamic memory allocation and ensuring efficient memory access
+     * via typically high locality.
+     *
+     * Key characteristics:
+     * - **Stack Allocation**: The tensor's data (if it is owning) is stored on the stack, avoiding
+     * the overhead of dynamic memory allocation.
+     * - **Temporary Objects**: Operations on tensors may create temporary objects, but these do not
+     *   involve dynamic memory allocation, ensuring efficient computation.
+     * - **Use Case**: This class is ideal for small-scale tensor operations at the Gauss point
+     * level in finite element simulations. It allows to write physical equations close to their
+     * pendant on paper.
+     *
+     * Comparison with @p LinAlg::Matrix:
+     * - **Tensor**: Designed for small, fixed-size data stored on the stack. Suitable for
+     * operations at the Gauss point level (e.g., strains, stresses).
+     * - **Matrix**: Designed for larger, dynamically allocated data. Suitable for element-level
+     *   calculations (e.g., stiffness and mass matrices). For example, a stiffness matrix for a
+     *   hex27 element may have dimensions 81x81 with 6561 entries, requiring dynamic memory
+     * allocation.
+     *
+     * Shared Backend:
+     * Both @p LinAlg::Tensor and @p LinAlg::Matrix share the same backend for matrix and vector
+     * operations, ensuring consistency and avoiding duplication of functionality.
+     *
+     * @tparam Number Type of the tensor elements (e.g., `double`).
+     * @tparam storage_type Specifies whether the tensor owns its data (`StorageType::owning`) or is
+     * a view (`StorageType::view`).
+     * @tparam n Dimensions of the tensor, specified as a variadic template parameter.
+     */
+    template <typename Number, StorageType storage_type, std::size_t... n>
+    class Tensor
+    {
+     public:
+      using value_type = Number;
+      using shape_type = std::integer_sequence<std::size_t, n...>;
+
+      static constexpr std::size_t rank_ = sizeof...(n);
+      static constexpr std::size_t size_ = (n * ...);
+
+      static_assert(storage_type == StorageType::view or
+                        std::is_same_v<std::remove_cv_t<value_type>, value_type>,
+          "Owning Tensor must have a non-const, non-volatile value_type");
+      static_assert(storage_type == StorageType::owning or
+                        std::is_same_v<std::remove_volatile_t<value_type>, value_type>,
+          "TensorView must have a non-volatile value_type");
+      static_assert(sizeof...(n) > 0, "Tensor must have at least rank one");
+      static_assert(
+          []() constexpr
+          {
+            std::array shape = {n...};
+            return std::ranges::all_of(shape, [](auto dim) { return dim > 0; });
+          }(),
+          "The extents of all axes must larger than zero");
+
+      // Container is either a std::array for an owning tensor or a std::span for a view
+      using ContainerType = std::conditional_t<storage_type == StorageType::owning,
+          std::array<Number, size_>, std::span<Number, size_>>;
+
+     private:
+      ContainerType data_;
+
+     public:
+      /*!
+       * @brief Default constructor
+       *
+       * @note Only an owning tensor has the default constructor
+       */
+      Tensor()
+        requires(std::is_default_constructible_v<ContainerType>)
+      = default;
+
+      Tensor(const Tensor&) = default;
+
+      /*!
+       * @brief Move constructor
+       *
+       * @note Copy operation of a view is cheap, so we don't need to have a move constructor
+       */
+      Tensor(Tensor&&) noexcept
+        requires(storage_type == StorageType::owning)
+      = default;
+      Tensor& operator=(const Tensor&) noexcept = default;
+
+      /*!
+       * @brief Move assignment operator
+       *
+       * @note Copy operation of a view is cheap, so we don't need to have a move assignment
+       * operator
+       */
+      Tensor& operator=(Tensor&&) noexcept
+        requires(storage_type == StorageType::owning)
+      = default;
+      ~Tensor() = default;
+
+      /*!
+       * @brief Construct an owning tensor from given data
+       *
+       * You can assign the tensor via
+       *
+       * @code
+       * Tensor<double, 3, 4> A = {{
+       *     {1.0, 2.0, 3.0, 4.0},
+       *     {5.0, 6.0, 7.0, 8.0},
+       *     {9.0, 10.0, 11.0, 12.0}
+       * }};
+       * @endcode
+       *
+       */
+      Tensor(const TensorInitializerList<Number, n...>::type& lst)
+        requires(std::is_default_constructible_v<ContainerType>);
+
+      /*!
+       * @brief Create a view on an owning tensor
+       */
+      Tensor(Tensor<Number, StorageType::owning, n...>& view_on)
+        requires(storage_type == StorageType::view)
+          : data_(view_on.container())
+      {
+      }
+
+      /*!
+       * @brief Create a constant view on a const owning tensor
+       */
+      Tensor(const Tensor<std::remove_cv_t<Number>, StorageType::owning, n...>& view_on)
+        requires(storage_type == StorageType::view && std::is_const_v<Number>)
+          : data_(view_on.container())
+      {
+      }
+
+      /*!
+       * @brief Create a copy of a view
+       */
+      Tensor(const Tensor<std::remove_cv_t<Number>, StorageType::view, n...>& other)
+        requires(storage_type == StorageType::owning)
+      {
+        std::ranges::copy(other.container(), data_.begin());
+      }
+
+      /*!
+       * @brief Access to the underlying raw data of the tensor
+       *
+       * @return Number*
+       */
+      [[nodiscard]] Number* data() { return data_.data(); }
+
+      /*!
+       * @brief Access to the underlying raw data of the tensor for readonly access
+       *
+       * @return Number*
+       */
+      [[nodiscard]] const Number* data() const { return data_.data(); }
+
+      /*!
+       * @brief Access to the underlying container (std::array or std::span, depending on whether it
+       * is owning or a view)
+       *
+       * @return Number*
+       */
+      [[nodiscard]] ContainerType& container() { return data_; }
+
+      /*!
+       * @brief Access to the underlying readonly container (std::array or std::span, depending on
+       * whether it is owning or a view)
+       *
+       * @return Number*
+       */
+      [[nodiscard]] const ContainerType& container() const { return data_; }
+
+      /*!
+       * @brief Indexing operator to access individual values of the tensor (without bound checks)
+       *
+       * @param i
+       * @return Number&
+       */
+      [[nodiscard]] Number& operator()(decltype(n)... i);
+
+      /*!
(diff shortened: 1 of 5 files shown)


amgebauer:
I tried in a small benchmark whether it is problematic to split `A = 3 * A + 4 * B * C`(A, B, C are 3x3 matrices) into its subcomponents
(1) scaling `3 * A`
(2) matrix-matrix multiplication `B * C`
(3) scaling `4 * (B * C)`
(4) Addition `(3 * A) + (4 * B * C)`
instead of using the overload of our `Core::LinAlg::Matrix<>` that does this "in one go".

Clang is able to optimize both versions so that the results of the benchmark are the same: https://quick-bench.com/q/DrzKaU_pjZVwWWqGBunYw_H4WXE. GCC is not as good in optimizing this equation. The separated version is ~27% slower.

The benefit is that we can write the c++ code as we would write it on paper, i.e.,
```cpp
Tensor<double, 3, 3> A = {...};
Tensor<double, 3, 3> B = {...};
Tensor<double, 3, 3> C = {...};

A = 3.0 * A + 4.0 * B * C;
```
instead of
```cpp
Matrix<3, 3> A;
Matrix<3, 3> B;
Matrix<3, 3> C;

A.multiply(4.0, A, B, 3.0);
```
I think the benchmark problem is also somehow pessimistic as this kind of multiplication-scaling-addition example is only rarely used in material evaluations and many other operations compile to the same code.

Of course, we can also define these specialized methods for the Tensor, but then we again move away from how we write these equations on paper. Before doing that, we should exchange our own-cooked implementation with a better-optimized linear algebra backend for these operations.

Theoretically, there are packages out there (like [Fastor](https://github.com/romeric/Fastor)) that apply some compile-time optimizations via lazy evaluations. With that technique, it should be possible to end up in the same performance also with GCC (but that also has cons).

amgebauer:
This PR is now ready for some feedback.

sebproell:
Very cool, @amgebauer! I will check this out locally and have a look :)

mayrmt:
@amgebauer I am wondering: To what extent is this a reimplementation of the fixed-size matrix? What are the key advantages of the new objects?

amgebauer:
> @amgebauer I am wondering: To what extent is this a reimplementation of the fixed-size matrix? What are the key advantages of the new objects?

Thanks for the question. Please read the description of the PR.

mayrmt:
> Thanks for the question. Please read the description of the PR.

I wouldn't have asked, if the answer was there.

amgebauer:
> > Thanks for the question. Please read the description of the PR.
> 
> 
> 
> I wouldn't have asked, if the answer was there.

Ok, then reformulate the questions more precisely, because I don't know what to answer without repeating the description of the PR and the linked issues. Or ask what is unclear with the description.

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -0,0 +1,366 @@
...
+      [[nodiscard]] static constexpr std::size_t size() { return size_; }
+
+      /*!
+       * @brief Returns the dimension of the i-th axis of the tensor
+       *
+       * @tparam i
+       */
+      template <std::size_t i>
+        requires(i < rank())
+      [[nodiscard]] static constexpr auto size_of()
```
sebproell, line 254:
```suggestion
      [[nodiscard]] static constexpr auto extent()
```
?

`size_of` is confusing since there is the `sizeof` operator which does something very different.

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -0,0 +1,366 @@
...
+    /*!
+     * @brief This is a general tensor class for dense tensors of arbitrary rank.
+     *
+     * @Note: The tensor might own the data or be just a view depending on the StorageType.
+     *
+     * @tparam Number Type of the tensor elements
+     * @tparam storage_type Whether the tensor owns the data or is just a view
+     * @tparam n Dimensions of the tensor of arbitrary rank
+     */
+    template <typename Number, StorageType storage_type, std::size_t... n>
```
sebproell, line 65: Do we need the flexibility of passing arbitrary dimensions? Many classical operations (inv, det, trace) work on square tensors. In deal.II, there is only a `rank` and a `dim` argument. This has served me well, although I am admittedly not a tensor power user :) Maybe it's obvious why we need this flexibility; in this case, I am eager to learn.

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -0,0 +1,366 @@
...
+      constexpr void fill(const Number& value) { std::ranges::fill(data_, value); }
+
+      /*!
+       * @brief Add another tensor onto this tensor
+       *
+       * @tparam OtherTensor
+       */
+      template <typename OtherTensor>
+        requires(std::is_same_v<typename OtherTensor::shape_type, typename OtherTensor::shape_type>)
+      constexpr Tensor<Number, storage_type, n...>& operator+=(const OtherTensor& B)
```
sebproell, line 280: I tried to use this in a `consteval` function. It does not work yet since `DenseFunctions::update` is not `constexpr`. Could be addressed at a later point.

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -0,0 +1,366 @@
...
+      }
+
+      /*!
+       * @brief Add a scalar value on every element of the tensor
+       *
+       * @tparam OtherTensor
+       */
+      template <typename Scalar>
+        requires(std::is_arithmetic_v<Scalar>)
+      constexpr Tensor<Number, storage_type, n...>& operator+=(const Scalar b)
```
sebproell, line 308: This operation seems a bit fishy from a mathematical perspective. Does this operation typically appear?

src/core/linalg/src/dense/4C_linalg_tensor.hpp
amgebauer: From the implementation perspective, this add-on was quite easy and, mathematically, a tensor is not limited to vector spaces of the same size. But I agree, in most physical applications you are right. Different dimensions might only be needed to map between vector spaces of different dimension. For rank-1 tensors, we could use LinAlg::Matrix for that mapping, for higher order rank tensors, we would need this generalization.

I don't need this in my own applications, but (not sure whether it is actually a good example) think of 2D mechanics: In plain strain conditions, you map a 2D rank 2 tensor (strains) to a 3D rank 2 tensor (stress).

src/core/linalg/src/dense/4C_linalg_tensor.hpp
amgebauer: True, it's too much python style and violates too many mathematical rules :grimacing:.

I had the following in mind during implementation: (e.g., adding a unit rank 2 tensor)

```c++
Tensor<double, 3, 3> A{};
[...]
// Add unit tensor
diagonal_view(A) += 1;
```

But I think we don't need this convenience. And we would need a strided view.

This is more expressive:
```c++
Tensor<double, 3, 3> A{};
[...]
// Add unit tensor
A += diag<double, 3>(1);
```

src/core/linalg/src/dense/4C_linalg_tensor.hpp
amgebauer: I've talked to various developers from different modules (thanks @davidrudlstorfer for beams, @mairehenke for shells and membranes, @c-p-schmidt for space-charge-layers etc.). I hope I understand everything correctly. For beams it is not necessary since the components are separated and the manifold is 1D. Similar to the space-charge layers in ssi applications. However, there are plans to add more physical effects of the reduced layers and that might also require mappings between different dimensions. For the shell7p formulation, there is no usecase since the strains are 3d in that case. However, in our membrane formulation, we implemented mappings between 2D and 3D vector spaces (see also eqn 43 in https://doi.org/10.1002/nme.1620350511).

Since it's not much overhead the implementation (and also some interesting (not complex!) template challenges), I tend to keep the general version.

sebproell: Fine for me :+1:

isteinbrecher:
I like the way one can write these calculations with this implementation. Can we make the `Linalg::Matrix` use these improved operators as well? Maybe make the `Linalg::Matrix` a special case of the tensor implementation.

amgebauer:
> I like the way one can write these calculations with this implementation. Can we make the `Linalg::Matrix` use these improved operators as well? Maybe make the Linalg::Matrix a special case of the tensor implementation.

Yes, we can add these operators to `LinAlg::Matrix` as well (separately), but we have to be more careful there.

A Tensor is typically very small (typical dimension is `2` or `3` and rank is, typically `1` or `2`, maximum `4`, i.e., < `81` entries, typically `9` or less). The internal data of the Tensor is on the stack, so we don't need to dynamically allocate memory (we internally use `std::array<double, size>`, no `double*`). For the operations, we create many temporary objects, but have no dynamic memory allocation (no `new double[size]`).

For the Matrix that is different. It also has to represent large matrices (stiffness matrix of `hex27` is a `81x81` matrix with `6561` entries), so we typically dynamically allocate the memory. This can of course significantly impact performance. The compiler might be able to optimize also these cases, but I have not tested that.

I think we need to keep the `LinAlg::Matrix` for calculations on the element level (stiffness matrix, mass matrix etc) and use the `LinAlg::Tensor` for the implementation of the physics on the Gauss point level (for strains, stresses, linearization, etc.).

Importantly, both use the same backend, i.e., we don't duplicate any matrix/vector operations.

I will add this to the docstring of Tensor.

isteinbrecher:
Thanks for the explanation, makes sense.

Could that potentially become an issue if, for example I have a 2nd order tensor in R^3 and I use FAD types with 81 dependent variables? Or would we in that case prefer the "old" implementation that can also do dynamic memory allocation? (I remember there is some magic value that decides on static/dynamic memory allocation in the Linalg::Matrix)

amgebauer:
> Thanks for the explanation, makes sense.
> 
> Could that potentially become an issue if, for example I have a 2nd order tensor in R^3 and I use FAD types with 81 dependent variables? Or would we in that case prefer the "old" implementation that can also do dynamic memory allocation? (I remember there is some magic value that decides on static/dynamic memory allocation in the Linalg::Matrix)

Then you might end up searching for stack overflow on stack overflow :thinking:

At least DFad should not be a problem because it dynamically allocates the additional variables. The compiler might then be less able to optimize it.

amgebauer:
I now added a generalized transpose:

```c++
Tensor<double, 2, 3> A;
A_t = Core::LinAlg::reorder_axis<1, 0>(A); // results in Tensor<double, 3, 2>

Tensor<double, 2, 3, 4> B;
B_tt = Core::LinAlg::reorder_axis<2, 1, 0>(B); // results in Tensor<double, 4, 3, 2>

// A_weird_t = Core::LinAlg::reorder_axis<1, 1, 0>(A); // does not compile
// A_weird_t = Core::LinAlg::reorder_axis<3, 1, 0>(A); // does not compile
```
The reordering of the tensor entries is computed at compile time, so this should be quite efficient.

amgebauer:
Feel free to review. I would suggest to merge this in this state. I suggest to add missing operations later once needed (I think the important ones will come up quickly). I will then update some of my materials, do some performance tests in practice. Based on that, we can tweak the implementation a bit further and start using it in more applications.

src/core/linalg/src/dense/4C_linalg_tensor_operations.hpp
```diff
@@ -0,0 +1,416 @@
...
+      {
+        std::array<std::size_t, (shape * ...)> index_mapping = {0};
+
+        for (std::size_t i = 0; i < (shape * ...); ++i)
+        {
+          auto old_index = get_md_index<OrderType::column_major, TensorBoundCheck::no_check,
+              get_axis(new_order)...>(i);
+
+          // auto new_index = std::make_tuple(
+          //     std::get<new_order>(old_index)...);  // this mapping must be vice versa!!
```
sebproell, line 94: ?

src/core/linalg/src/dense/4C_linalg_tensor_operations.hpp
```diff
@@ -0,0 +1,416 @@
...
+#ifndef FOUR_C_LINALG_TENSOR_OPERATIONS_HPP
+#define FOUR_C_LINALG_TENSOR_OPERATIONS_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_linalg_fixedsizematrix.hpp"
+#include "4C_linalg_tensor.hpp"
+#include "4C_linalg_tensor_meta_utils.hpp"
+
+#include <bits/utility.h>
```
sebproell, line 17:
```suggestion
#include <utility>
```

src/core/linalg/src/dense/4C_linalg_tensor_meta_utils.hpp
```diff
@@ -0,0 +1,216 @@
...
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_LINALG_TENSOR_META_UTILS_HPP
+#define FOUR_C_LINALG_TENSOR_META_UTILS_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_utils_exceptions.hpp"
+
+#include <bits/utility.h>
```
sebproell, line 15:
```suggestion
#include <utility>
```

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -0,0 +1,382 @@
...
+
+#ifndef FOUR_C_LINALG_TENSOR_HPP
+#define FOUR_C_LINALG_TENSOR_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_linalg_fixedsizematrix.hpp"
+#include "4C_linalg_tensor_meta_utils.hpp"
+
+#include <bits/utility.h>
```
sebproell, line 16:
```suggestion
#include <utility>
```

src/core/linalg/src/dense/4C_linalg_tensor_operations.hpp
sebproell, line 1: Wouldn't I include this file basically always when I am using a tensor? Does it make sense to merge it into the main tensor header?

src/core/linalg/src/dense/4C_linalg_tensor_operations.hpp
```diff
@@ -0,0 +1,416 @@
...
+    Core::LinAlg::Tensor<value_type, n, m> dest;
+    DenseFunctions::multiply<value_type, m, k, n>(dest.data(), A.data(), B.data());
+    return dest;
+  }
+
+  template <typename TensorLeft, typename TensorRight>
+    requires(Rank2TensorConcept<TensorLeft> && Rank2TensorConcept<TensorRight> &&
+             TensorLeft::template extent<0>() == TensorRight::template extent<0>() &&
+             TensorLeft::template extent<1>() == TensorRight::template extent<1>())
+  constexpr auto ddot(const TensorLeft& A, const TensorRight& B)
```
sebproell, line 224: These operations are all very nice! Could you add docs? And, again, I suggest to move the definition at the end of the header and just have the declaratiion with docs here.

sebproell (commented):
Looks very nice and useful :sunglasses:

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -0,0 +1,342 @@
...
+  using Tensor = Internal::Tensor<T, Internal::StorageType::owning, n...>;  // owns the memory
+
+  template <typename T, std::size_t... n>
+  using TensorView =
+      Internal::Tensor<T, Internal::StorageType::view, n...>;  // view onto's other memory
+}  // namespace Core::LinAlg
+
+FOUR_C_NAMESPACE_CLOSE
+
+#include "4C_linalg_tensor_operations.hpp"  // IWYU pragma: export
```
amgebauer, line 340: Not sure whether this is actually a good idea. I don't wan that this file explodes in size :grimacing:

src/core/linalg/src/dense/4C_linalg_tensor_meta_utils.hpp
```diff
@@ -0,0 +1,216 @@
...
+    // initializers
+    using type = typename TensorInitializerList<Number, n...>::type[n1];
+  };
+
+  template <typename Number, std::size_t... n>
+  const Number* get_view_to_first_element(
+      const typename TensorInitializerList<Number, n...>::type& lst)
+  {
+    // This is to get a view to the first element in the nested array
+    return static_cast<const Number*>(static_cast<const void*>(&lst));
```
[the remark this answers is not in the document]
amgebauer: I need a pointer to the raw data, and this should return a pointer to a C-style array.

The initializer type is a Number[n1][n2][n3]... and I need to get the raw values (i.e., the pointer to the first value)

src/core/linalg/src/dense/4C_linalg_tensor_operations.hpp
amgebauer: yes. I'm not super convinced of my new solution: https://github.com/4C-multiphysics/4C/pull/583#discussion_r2027654938 :grimacing: What do you suggest?

src/core/linalg/src/dense/4C_linalg_tensor.hpp
sebproell: Hm, including a file at the bottom is quite unusual. If you want to split the files for organization purposes,  provide a header that does nothing but include all the parts. This file can be included by users. But for the first implementation, I would inline the operations here.

davidrudlstorfer (approved):

PR #585, opened 2025-03-29
Enhance build time analysis with ninjatracing

sebproell:
I learned about this nice tool [ninjatracing](https://github.com/nico/ninjatracing) which allows to embed the detailed output from `-ftime-trace` within the project-wide info in .ninja_log. The results can be viewed as a flame graph like this (e.g. here: https://ui.perfetto.dev/):

image

This PR adds this information to the existing nightly job and provides the trace as an artifact.

Credit for the idea: https://www.youtube.com/watch?v=Oih3K-3eZ4Y

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -136,6 +136,11 @@ jobs:
           repository: aras-p/ClangBuildAnalyzer
           ref: bae0cb488cce944bfc3da9850a69ad621701ebef # version 1.6.0
           path: ClangBuildAnalyzer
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (approved):

davidrudlstorfer (approved):

amgebauer (approved):
Should we upload the report for easier access next to the coverage report?

sebproell:
> Should we upload the report for easier access next to the coverage report?

Could be interesting. Chrome can open this file type natively, AFAIK. However, I guess the user base will be rather small. We can do this later if we see a need.

PR #592, opened 2025-04-01
Upload metadata after build

gilrrei:
## Description and Context
This PR adds the `4C_metadata.yaml` file to the artifacts. This is useful for FourCIPP since we can download the file for testing directly from the artifacts.

The JSON schema is already uploaded.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -3,10 +3,10 @@ name: buildtest
 on:
   pull_request:
     branches:
-      - 'main'
+      - main
   push:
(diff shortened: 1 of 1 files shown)


ppraegla (dismissed):

davidrudlstorfer (approved):

davidrudlstorfer (dismissed):

c-p-schmidt (approved):

amgebauer (approved):
I think the nightly jobs are not uploading schema and metadata, but we can tackle this in another PR.

PR #597, opened 2025-04-02
CI: new job to compile with no optional dependencies

### LINKED ISSUE #509: [CI] Test with all optional dependencies disabled
## Description

A related previous discussion may be found in https://gitlab.lrz.de/baci/baci/-/issues/1187

We should add a check, either to PR testing or to nightly, that tests with all optional dependencies tuned off.

sebproell:
This new job turns off the dependencies that are optional. I only added a minimal test in PRs. I propose to test this in PRs, since it should not be slower than the rest of the pipeline. The question is how we should add this to nightly. Yes/no? With full tests?

Closes #509, Useful first step to test #506

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -221,6 +221,55 @@ jobs:
           yaml_extension_short: .4C.yml
           json_extension: .4C.json
 
+  gcc13_no_optional_dependencies_build:
+    runs-on: ubuntu-latest
+    container:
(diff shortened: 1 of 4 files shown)


amgebauer:
Yes, I would add it to nightly with full tests to catch that we only use optional dependencies in tests that we correctly exclude in the full test suite.

amgebauer (approved):

ppraegla (approved):

Issue #598, opened 2025-04-02
Deprecation of dat format in favor of yaml


sebproell:
This is a meta issue to track progress on https://github.com/4C-multiphysics/4C/milestone/4. Much of the content has been discussed in #263.

## Roadmap for migration

During the migration dat stays functional although we will advise everybody to move away from it.

- [x] Ensure that reading the large geometry sections from yaml files is as performant as reading dat-files for large examples (approx 5e7 DOF, 2e7 nodes, 2e7 elements).
- [x] FourCIPP seems very helpful in a functional ~~and sufficiently performant~~ state (max performance overhead of 0.1s --> completed within https://github.com/4C-multiphysics/fourcipp/pull/11). If useful, common (python) infrastructure may emerge during these steps. 
- [x] Convert some of the existing test cases to yaml while keeping some in dat format in order to test both input paths of 4C and thus avoid untested code.
- [x] Warn if a dat file is used, point out how easily it can be converted with https://github.com/4C-multiphysics/4C/pull/382 (=deprecation) (--> @sebproell)
- [x] Migrate public and open input generators (meshpy, lnmmeshio, cubitpy) to directly produce yaml input files. Goal: everybody needs to be able to create native yaml files from scratch.
  - [x] lnmmeshio
  - [x] meshpy
  - [x] cubitpy
- [x] If pre-exodus is still desired, it should be ported to read a yaml header as indicated here: https://github.com/4C-multiphysics/4C/discussions/490#discussioncomment-12463272
  - [x] Remove `pre_exodus` and use direct exodus input #751 
- [x] Purge mentions of dat from the documentation. Either use neutral wording (e.g. "input file") or, where useful, refer to yaml directly.
  - [x] What about ABAQUS to dat converter https://4c-multiphysics.github.io/4C/documentation/tools/tools.html#abaqus-to-fourc-input-converter -> Remove the tools page
- [x] Specify the required input format in the documentation, so that users know what 4C expects (and do not have to reconstruct it from the source code).
  - [x] Refer to exodus files as `EXODUS` (see https://github.com/4C-multiphysics/4C/pull/769#discussion_r2086188106)
- [x] Document necessary steps to create a yaml input file from scratch.
- [x] Input generators (besides pre_exodus) should be mentioned in the documentation! Different tutorials may use different input generators. Ideally, a comparison of the input generators with their strenghts is provided to offer new users to optimally choose the input generator. (Done but more tutorials and links would be helpful if anything exists.)
[...]

PR #601, opened 2025-04-02
Warn when using a dat file in a 4C run

sebproell:
Part of #598

### DIFF
--- a/apps/global_full/4C_global_full_control.cpp
+++ b/apps/global_full/4C_global_full_control.cpp
@@ -24,6 +24,17 @@ namespace
                .count() *
            1.0e-3;
   }
+
+  void print_dat_deprecation_warning()
+  {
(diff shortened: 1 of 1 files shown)


isteinbrecher (approved):

apps/global_full/4C_global_full_control.cpp
```diff
@@ -70,4 +89,13 @@ void ntam(int argc, char* argv[])
     Core::IO::cout << "\nTotal wall time for CALCULATION: " << std::setw(10) << std::setprecision(3)
                    << std::scientific << tc << " sec \n\n";
   }
+
+  // Let's print the deprecation warning again to increase the chance users will see it.
```
isteinbrecher, line 93: Good idea, I am wondering if anybody will even see the note at the beginning - I for sure would not.

PR #602, opened 2025-04-02
Fix: Change fiber compression condition to I4 < 1

tuchpaul:
Changed the condition that excludes fiber contribution under compression from I4 < 0 to I4 < 1, since fibers are already compressed before I4 drops below 1 - with I4 being the 4th pseudo-invariant of the right Cauchy-Green deformation tensor -.


src/mixture/4C_mixture_constituent_remodelfiber_lib.hpp
```diff
@@ -45,7 +45,7 @@ namespace Mixture
   [[nodiscard]] inline T get_exponential_fiber_strain_energy(
       const PAR::ExponentialFiberParameters& params, const T I4)
   {
-    if (I4 < 0 && !params.supports_compression_)
+    if (I4 < 1 && !params.supports_compression_)
       FOUR_C_THROW("The fiber is under compression, but does not support that.");
```
amgebauer, line 49:
```suggestion
    FOUR_C_ASSERT_ALWAYS(params.supports_compression_ || I4 >= 1, "The fiber is under compression, but does not support that.");
```
It's now also possible to write this in one line (but the current version is also fine).

amgebauer (dismissed):
Thanks for your first contribution @tuchpaul and for fixing this!

c-p-schmidt (dismissed):

amgebauer:
It looks like that this condition is too strict for prestressing (also, fibers must only not be under compression in the converged state). 

From Holzapfel and Ogden (2009), https://doi.org/10.1098/rsta.2009.0091
> However, when they are under compression, their contribution is minimal becausethe fibres do not support compression. For this reason, we include these terms in the energy function only if I4 f > 1 or I4 s > 1, as appropriate.

So, let's not throw an error, but just exclude them, i.e.,

```c++
if (I4 < 1 && !params.supports_compression_)
{
  return 0.0;
}
```

amgebauer:
I think you forgot to commit the removal of the error. Can you also cleanup the commit history a bit and squash all commits into one single one (e.g., with `git rebase -i [...]`)?

amgebauer (approved):

c-p-schmidt (approved):

sebproell:
@tuchpaul Thanks for the first contribution! Could you add your commit email to your Github account so the commit  will be correctly attributed to you?

PR #588, opened 2025-03-31
InputSpec: refactor internals and add better errors

sebproell:
The first commit shuffles some code around (in the implementation). The second commit enhances various user-facing error messages when input does not match the specification.

PR #603, opened 2025-04-02
Validate size read from_parameter for yaml files

sebproell:
I was trying to be too clever here and thought we don't need the check. 

Error looks like this now:

```
Could not match this input

DESIGN SURF DIRICH CONDITIONS:
  - E: 1
    NUMDOF: 3
    ONOFF: [1,1,1]
    VAL: [0,0,0]
    FUNCT: [0,0,0,4]


against the given input specification. This was the best attempt to match the input:

[!] Candidate list 'DESIGN SURF DIRICH CONDITIONS'
  [!] The following list entry did not match:
    {
      [ ] Matched parameter 'E'
      [ ] Matched parameter 'NUMDOF'
      [ ] Matched parameter 'ONOFF'
      [ ] Matched parameter 'VAL'
      [!] Candidate parameter 'FUNCT' (value has incorrect size)
      [ ] Defaulted deprecated_selection 'TAG'
    }
```

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.cpp
+++ b/src/core/io/src/4C_io_input_spec_builders.cpp
@@ -1229,6 +1229,51 @@ namespace
     return flattened_specs;
   }
 
+  // Convert an InputSpec of the form all_of(x,y, one_of(a,b,c))
+  // into one_of(all_of(x,y,a), all_of(x,y,b), all_of(x,y,c)).
+  std::vector<Core::IO::InputSpec> push_all_of_into_one_of(std::vector<Core::IO::InputSpec> specs)
+  {
+    // Find the one_of spec and pull it out of the list. Assert that it is only one.
+    auto one_of_it = std::ranges::find_if(
+        specs, [](const auto& spec) { return spec.impl().data.type == InputSpecType::one_of; });
+
+    if (one_of_it == specs.end())
+    {
+      // No one_of spec found, nothing to do.
+      return specs;
+    }
+
+    // Check that there is no other one_of spec in the list.
+    if (std::find_if(one_of_it + 1, specs.end(), [](const auto& spec)
+            { return spec.impl().data.type == InputSpecType::one_of; }) != specs.end())
+    {
+      FOUR_C_THROW("Cannot have multiple one_of specs in the same all_of spec.");
+    }
+    using namespace InputSpecBuilders;
+    // Take the choices from the one_of spec and enhance them with the other specs that we got
+    // passed in this function.
+    auto& one_of_spec = *one_of_it;
+    const auto& old_one_of_choices =
+        dynamic_cast<InputSpecTypeErasedImplementation<OneOfSpec>&>(one_of_spec.impl())
+            .wrapped.specs;
+    // specs now only contains the other specs that are not one_ofs. These are pulled into the
+    // one_of specs.
+    std::vector<Core::IO::InputSpec> new_one_of_choices;
+    for (const auto& choice : old_one_of_choices)
+    {
+      // Make sure to keep the ordering.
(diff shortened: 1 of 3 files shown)


sebproell:
Sorry for that @shervas :innocent: This should fix your problem, and, in combination with #588, you should see a nice error. Thanks for trying out the new features!

amgebauer (dismissed):

davidrudlstorfer (dismissed):

src/core/io/src/4C_io_input_spec_builders.cpp
```diff
@@ -1229,6 +1229,51 @@ namespace
     return flattened_specs;
   }
 
+  // Convert an InputSpec of the form all_of(x,y, one_of(a,b,c))
+  // into one_of(all_of(x,y,a), all_of(x,y,b), all_of(x,y,c)).
+  std::vector<Core::IO::InputSpec> push_all_of_into_one_of(std::vector<Core::IO::InputSpec> specs)
```
sebproell, line 1234: @gilrrei and I have been talking about this function for a longer time. It proofed to be very useful to tackle another subtle bug that may happen in certain, poorly written InputSpecs lacking substructure. In particular, this occurred for the `FUNCTION`s.

src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -705,9 +705,9 @@ description: 'one_of {a, b, one_of {c, one_of {d, e, f}}}'
             {.default_value = std::vector<std::optional<double>>{1., std::nullopt, 3.}, .size = 3}),
         one_of({
             all_of({
-                parameter<std::map<std::string, std::string>>(
-                    "b", {.default_value = std::map<std::string, std::string>{{"key", "abc"}},
-                             .size = 1}),
+                parameter<std::map<std::string, std::string>>("string to string",
+                    {.default_value = std::map<std::string, std::string>{{"key", "abc"}},
```
sebproell, line 709: And I also found this bug in the tests: Of course we cannot have another "b" here.

src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -750,40 +750,112 @@ description: 'one_of {a, b, one_of {c, one_of {d, e, f}}}'
       spec.emit_metadata(yaml);
       out << tree;
 
-      std::string expected = R"(type: all_of
-description: 'all_of {a, b, one_of {all_of {b, c}, triple_vector, group}, e, eo, group2, list, selection_group}'
-required: true
+      std::string expected = R"(type: one_of
+description: 'one_of {all_of {a, b, string to string, c, e, eo, group2, list, selection_group}, all_of {a, b, triple_vector, e, eo, group2, list, selection_group}, all_of {a, b, group, e, eo, group2, list, selection_group}}'
```
sebproell, line 754: The exemplary metadata explodes quite drastically but we exercise the input mechanism in a rather unrealistic manner. The real metadata file increases from 40k lines to 43k lines.

gilrrei: As I mentioned to you @sebproell, this isn't an issue and is expected. The 4C input space is huge, we are talking about 467  sections and 192 materials.

gilrrei (approved):
With `push_all_of_into_one_of`, the input is now even closer to the [JSON schema](https://json-schema.org/draft/2020-12) standard. I think it's a good sign that both of us came to this conclusion independently, especially since it aligns with an existing and established [standard](https://json-schema.org/draft/2020-12).

At first, some of the JSON schema logic seems restrictive, however, in 99% of the cases when we want a more 'flexible' input approach that deviates from the standard, it's due to a lack of structure, i.e. grouping and naming of the nested parameters.

amgebauer (approved):

PR #605, opened 2025-04-02
Fix enum printing via magic_enum

dragos-ana:
## Description and Context

- Enums were previously not recognized as StreamInsertable concepts due to namespace / lookup issues.
- This was fixed by moving the PrintHelper functions to the corresponding parameter container template file and including the corresponding enum utilities file.
- A dedicated unit test was added checking the stream-insertability.

Fixed by @sebproell and me.

Fixes the bug discussed in [588](https://github.com/4C-multiphysics/4C/pull/588#discussion_r2021107936)

### DIFF
--- a/src/core/io/tests/4C_io_input_parameter_container_test.cpp
+++ b/src/core/io/tests/4C_io_input_parameter_container_test.cpp
@@ -11,6 +11,8 @@
 
 #include <Teuchos_ParameterList.hpp>
 
+#include <ostream>
+
 namespace
 {
   using namespace FourC;
@@ -60,7 +62,7 @@ namespace
         pl.get<std::vector<Teuchos::ParameterList>>("list")[1].sublist("group").get<int>("b"), 2);
   }
 
-  TEST(InputParameterContainerTest, StreamInsertableEnum)
+  TEST(InputParameterContainerTest, PrintEnum)
   {
     enum class TestEnum
     {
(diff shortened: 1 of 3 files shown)


src/core/io/tests/4C_io_input_parameter_container_test.cpp
```diff
@@ -59,4 +59,22 @@ namespace
...
+      C
+    };
+
+    InputParameterContainer container;
+    container.add("enum", TestEnum::A);
+
+    // check that the enum is stream-insertable
+    static_assert(
+        Core::IO::Internal::InputParameterContainerImplementation::StreamInsertable<TestEnum>,
+        "TestEnum must be stream-insertable");
```
sebproell, line 78: You are currently testing an implementation detail. An alternative would be to print the container and check the output.

sebproell: Actually, I meant to replace this assert with the alternative. Sorry if that wasn't clear enough. The point is that testing implementation details is annoying because pure internal refactorings suddenly require test changes.

dragos-ana: Done, thanks!

sebproell (approved):

c-p-schmidt (approved):

PR #607, opened 2025-04-03
Check filename before reading file in --to-yaml

ppraegla:
The --to-yaml converter aborts when the output file already exists. To avoid the unnessary reading of the input file in this case, the reading of the file is now done after checking the output file. This will save some seconds.

This is useful for converting large files. The converter now aborts after 10 seconds instead of 30 seconds for a input file of 3 GB.

### DIFF
--- a/apps/global_full/4C_global_full_main.cpp
+++ b/apps/global_full/4C_global_full_main.cpp
@@ -350,8 +350,6 @@ int main(int argc, char* argv[])
     {
       std::filesystem::path inputfile_name = argv[2];
       std::filesystem::path outputfile_name;
-      Core::IO::InputFile input_file = Global::set_up_input_file(lcomm);
-      input_file.read(inputfile_name);
       if (argc == 3)
       {
         outputfile_name = inputfile_name.replace_extension("4C.yaml");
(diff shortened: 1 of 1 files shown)


sebproell (commented):
Thanks, definitely the right order of doing things.

> The converter now aborts after 10 seconds instead of 30 seconds for a input file of 3 GB.

Should't it just be a fraction of a second now? What is taking 10 seconds? :thinking:

ppraegla:
> Thanks, definitely the right order of doing things.
> 
> > The converter now aborts after 10 seconds instead of 30 seconds for a input file of 3 GB.
> 
> Should't it just be a fraction of a second now? What is taking 10 seconds? 🤔

I also wondered about that. I was content with one third of the original time. But I will look into it.

amgebauer (approved):

sebproell (approved):
I think we can merge.

ppraegla:
I haven't had time to look at why the input reading still takes 10 seconds. But I will provide any insights or open a new PR once I have more information.

sebproell:
@ppraegla Isn't the point that there shouldn't be any reading at all?

ppraegla:
> @ppraegla Isn't the point that there shouldn't be any reading at all?

Sorry, I did not phrase it properly. I meant the `to-yaml` execution branch should not take 10 seconds until it reaches the point where the actual reading takes place.

Issue #571, opened 2025-03-26
Switch DOMAIN sections to InputSpec


gilrrei:
## Description
The domain sections (currently, the only tested ones are STRUCTURE and FLUID DOMAIN) are part of the `legacy_string_sections.` However, they should be treated like the parameter sections and switched to `InputSpec`.

## Possible Solution and Definition of Done
Move to InputSpec

amgebauer:
This is also much nicer since we have it then everything in one section. I think that would then also that this feature is more often used.

gilrrei:
> This is also much nicer since we have it then everything in one section. I think that would then also that this feature is more often used.

I didn't know how cool this feature is. For testing and benchmarking this is actually handy

gilrrei:
@sebproell did the nice job of reviving the`DOMAIN` sections which is super nice! One shortcomming is that it still uses the `x+` notation etc that you have to do in the `DNODE TOPOLOGY` which is unintuitive.

In a talk with @sebproell we came to the conclusion that we could approach it the same way we do for `X GEOMETRY`, i.e. do not explicitly create DNODE sets, instead provide it in the condition sections directly via some parameter. This makes DOMAIN usage more consistent with GEOMETRY which is the more general case.

Alternatively which would also be nice is to provide nodesets via some conditions like `x<10`

PR #612, opened 2025-04-04
Fix d node topology

davidrudlstorfer:
## Description and Context

After doing some roundtrip tests (convert all .dat files from 4C to .yaml -> read in all 4C.yaml files with fourcipp -> dump all files again -> run all simulations with 4C) the following problem emerged

Similar to #581. One was able to enter anything into a design section, i.e.,

```
---------------------------------------------------------DLINE-NODE TOPOLOGY
NODE  1 4C_is_cool 1
---------------------------------------------------------DSURF-NODE TOPOLOGY
NODE  1 THIS_IS_THE_BEST_SIMULATION_ON_EARTH 1
```

This PR checks if the actual keyword in each line is identical to the current section (see first commit). The second commit fixes `DSURF` -> `DSURFACE` and the third commit fixes `DVOL` -> `DVOLUME`

@sebproell this could be improved in the future by not asserting this check but using the parser identical to #581 

ping @gilrrei this should solve our observed problem

#581

### DIFF
--- a/src/core/io/src/4C_io_input_file_utils.cpp
+++ b/src/core/io/src/4C_io_input_file_utils.cpp
@@ -19,6 +19,7 @@
 #include "4C_io_value_parser.hpp"
 #include "4C_io_yaml.hpp"
 #include "4C_linalg_utils_densematrix_communication.hpp"
+#include "4C_utils_exceptions.hpp"
 #include "4C_utils_string.hpp"
 
 #include <Teuchos_ParameterList.hpp>
@@ -345,6 +346,11 @@ void Core::IO::read_design(InputFile& input, const std::string& name,
     if (nname == "NODE")  // plain old reading of the design nodes from the input file
     {
       stream >> nodeid >> dname >> dobj;
+
+      FOUR_C_ASSERT_ALWAYS(name == dname || (name == "DSURF" && dname == "DSURFACE") ||
+                               (name == "DVOL" && dname == "DVOLUME"),
+          "Wrong design node name: {}. Expected {}.", dname, name);
+
       topology[dobj - 1].insert(nodeid - 1);
     }
     else  // fancy specification of the design nodes by specifying min or max of the domain


davidrudlstorfer:
Let's see how many tests fail when one changes 1e6 lines :rofl:

mayrmt:
Although this has never been a problem before in practice, I can see the value in checking for valid entries in lines in the input file. I am wondering though, if that actually requires to change the section heading, effectively breaking every existing input file.

sebproell:
> if that actually requires to change the section heading, effectively breaking every existing input file.

I agree; let's not do that if we can avoid it. We can change this once we have the upgrade mechanism. For now, why not keep the section heading and explicitly accept both spellings?

davidrudlstorfer:
Sounds good - @sebproell where would I need to change the code to allow for both section names?

The section name appears more than once within `global_data/4C_global_data_read.cpp`

sebproell:
I would only leave the old section name. But in the (hand-written) parser logic you could accept either spelling. Basically you enhance the `FOUR_C_ASSERT_ALWAYS` to check for the special `VOLUME` and `SURFACE` cases.

gilrrei:
@sebproell @davidrudlstorfer To keep in mind if the parser is changed as in #581. I don't know the reasons historically, but the section parsing for the node sets is, in part, complicated due to the `DOMAIN` functionality. If you use `* DOMAIN`, the `EDGE`, `CORNER` etc are also defined in the `D*-NODE TOPOLOGY` sections, even though they are fundamentally different, see #571.

sebproell (approved):
@davidrudlstorfer The "diff in the diff", i.e., going from 1e6 lines changed down to 6 lines, is probably the most extreme I've seen so far :)

davidrudlstorfer:
@gilrrei thanks for looking into it on the fourcipp side.

I would still like to merge this PR as otherwise one could always enter completely random stuff and 4C would accept it. To ensure that it works within fourcipp we should still merge this

gilrrei:
> @gilrrei thanks for looking into it on the fourcipp side.
> 
> I would still like to merge this PR as otherwise one could always enter completely random stuff and 4C would accept it. To ensure that it works within fourcipp we should still merge this

Ups, I closed this by mistake. I fully agree this should be solved here.

gilrrei (approved):

PR #616, opened 2025-04-04
Usage of converged mechanical state for pseudo contact in ssi

c-p-schmidt:
## Description and Context
Working with the pseudo contact implementation within the ssi framework showed that it is favorable to use the converged mechanical state instead of the current (probably especially since the linearization was not implemented).
Thus, the converged mechanical state is used instead now.

The results of the two test cases are adapted accordingly.


src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_generic.hpp
```diff
@@ -417,10 +417,11 @@ namespace Solid
        *  example the EAS state or the condensed Lagrange multiplier state. The
        *  global state in terms of the x-vector is stored more globally.
        *
-       *  */
-      virtual void create_backup_state(const Core::LinAlg::Vector<double>& dir) {
+       *  \author hiermeier \date 12/17 */
```
sebproell, line 420:
```suggestion
```

sebproell (approved):

src/ssi/4C_ssi_str_model_evaluator_base.hpp
```diff
@@ -102,16 +107,20 @@ namespace Solid::ModelEvaluator
...
     void update_step_element() override {}
 
-    void update_step_state(const double& timefac_n) override {}
-
-    void write_restart(
-        Core::IO::DiscretizationWriter& iowriter, const bool& forced_writerestart) const override
+    void update_step_state(const double& timefac_n) override
     {
+      if (mechanical_stress_state_n_ != nullptr)
+        mechanical_stress_state_n_->update(1.0, *mechanical_stress_state_np_, 0.0);
```
dragos-ana, line 113: Maybe nitpicking, but I would define this also in the cpp file for consistency with the `write_restart` and `read_restart methods` which are also 'small'.

dragos-ana (approved):
Looks good to me - left a small comment.

PR #618, opened 2025-04-05
[CMake] Simplifiy installation logic

sebproell:
Inspired by a problem in #591, this PR simplifies the installation logic.


cmake/configure/configure_MIRCO.cmake
```diff
@@ -34,7 +34,7 @@ else() # Fetch MIRCO from GIT repository
     GIT_TAG ${MIRCO_GIT_TAG}
     )
   fetchcontent_makeavailable(mirco)
-  set(FOUR_C_MIRCO_ROOT "${CMAKE_INSTALL_PREFIX}/lib/cmake/mirco")
+  set(FOUR_C_MIRCO_ROOT "${CMAKE_INSTALL_PREFIX}")
```
sebproell, line 37: I'll have to check why MIRCO is not working. I tested the installation locally and AFAICT, it worked. A possible reason is that MIRCO is a bit inconsistent in its naming. The package is often called `mirco_lib` but the installation directory is called `mirco`. I'll check again locally.

sebproell: Well, it works locally :man_shrugging: I reverted this change for now.

PR #620, opened 2025-04-06
Clean up boost dependency

sebproell:
I accidentally included a boost file in the LinAlg::View header, so I cleaned this dependency a little bit.

- Remove string functions that are trivial to write with std
- Remove unused includes


mayrmt (approved):

sebproell:
This error doesn't make any sense: https://github.com/4C-multiphysics/4C/actions/runs/14293804045/job/40058417874?pr=620

The build works locally. 

@amgebauer WDYT? Could this somehow be an issue with `ccache`? But which part of the change could give `ccache` trouble?

amgebauer (approved):

PR #595, opened 2025-04-02
Clean up and improve docs on 4C installation

sebproell:
Throw out things that are outdated or too specific. Mention a few (recent) improvements to the installation process.

Overall goal: try to write the docs in a way that they do not get outdated. Refer to external documentation if possible.

PR #621, opened 2025-04-06
Add Google Benchmark for micro-benchmark tests

### LINKED ISSUE #611: Add Google Benchmark for quick performance tests of 4C submodules
### Discussed in https://github.com/4C-multiphysics/4C/discussions/610

<sup>Originally posted by **amgebauer** April  3, 2025</sup>
Every once in a while, I want to micro-benchmark some submodules in 4C (Currently #583). I either use quick-bench.com (often annyoing to copy all needed code snippets and sometimes even impossible), or I have a local patch that adds Google Benchmark to 4C (often annoying to get the old patch to work).

I would suggest add Google Benchmark infrastructure in 4C (similar to unit tests) and run the benchmark in ctest (with some option to reduce the number of iterations). Running the benchmarks is not for monitoring performance (rather difficult to achieve in the CI), but only for testing that everything works.

https://github.com/google/benchmark</div>

## Definition of Done

* [ ] Add Google Benchmark as external dependency (similar to Google Test), default `OFF`
* [ ] Add benchmarks to testsuite with limited iterations (for testing, not benchmarking)

amgebauer:
This provides infrastructure for micro-benchmarking tests. Useful for profiling or comparing different implementations. Setup is very similar to Google Test.

I added a few benchmark tests for the tensors added in #583.

Closes #611

### DIFF
--- a/cmake/functions/four_c_auto_define_benchmark_tests.cmake
+++ b/cmake/functions/four_c_auto_define_benchmark_tests.cmake
@@ -1,0 +1,116 @@
+# This file is part of 4C multiphysics licensed under the
+# GNU Lesser General Public License v3.0 or later.
+#
+# See the LICENSE.md file in the top-level for license information.
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+function(_set_up_benchmark_test_target _module_under_test _target)
+  message(VERBOSE "Setting up benchmark test target ${_target}")
+
+  add_executable(
+    ${_target}
+    ${PROJECT_SOURCE_DIR}/tests/benchmark_tests/4C_benchmark_tests_main.cpp ${_parsed_SOURCE}
+    )
+
+  # Store benchmark test executables directly inside the tests/ directory
+  set_target_properties(${_target} PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/tests)
+
+  # Do not try to build benchmark tests as unity files.
+  set_target_properties(${_target} PROPERTIES UNITY_BUILD OFF)
+
+  # All libraries are linked as PRIVATE since a benchmark test executable cannot be used as a dependency itself.
+
+  # Common dependencies for benchmark tests
+  target_link_libraries(${_target} PRIVATE four_c_private_compile_interface)
+
+  target_link_libraries(${_target} PRIVATE ${_module_under_test}_unit_test_deps)
+
+  target_link_libraries(${_target} PRIVATE benchmark::benchmark)
+
+  add_test(NAME ${_target} COMMAND $<TARGET_FILE:${_target}> --benchmark_dry_run)
+  set_tests_properties(${_target} PROPERTIES TIMEOUT ${UNITTEST_TIMEOUT})
+  set_tests_properties(${_target} PROPERTIES PROCESSORS 1)
+  set_tests_properties(${_target} PROPERTIES ENVIRONMENT "OMP_NUM_THREADS=1")
+
+  add_dependencies(benchmarktests ${_target})
+endfunction()
+
+##
+# Pickup all the source files in the current directory and add them to a benchmark test target.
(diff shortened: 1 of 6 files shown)


cmake/setup_tests.cmake
```diff
@@ -50,6 +50,37 @@ else()
...
+  # Define a convenience target for all benchmark tests and add it to 'full'
+  # All benchmark test executables should add themselves as a dependency to 'benchmarktests'
+  # disable tests from google benchmark
+  set(BENCHMARK_ENABLE_TESTING OFF)
+  add_custom_target(benchmarktests)
+  add_dependencies(full benchmarktests)
+
+  if(TARGET benchmark_main)
+    message(
+      FATAL_ERROR "A target <benchmark_main> has already been included by a TPL."
```
sebproell, line 67:
```suggestion
      FATAL_ERROR "A target <benchmark_main> has already been included by another library."
```

I tried to avoid the word "TPL" in #595.

src/core/linalg/benchmark_tests/4C_linalg_tensor_benchmark.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include <benchmark/benchmark.h>
+
+FOUR_C_NAMESPACE_OPEN
+template <std::size_t... size>
+Core::LinAlg::Tensor<double, size...> get_random_tensor()
+{
+  Core::LinAlg::Tensor<double, size...> tensor;
+  std::generate(tensor.container().begin(), tensor.container().end(),
+      []() { return static_cast<double>(std::rand()) / std::rand(); });
```
sebproell, line 18: Let's avoid the UB and not potentially divide by zero.

tests/benchmark_tests/4C_benchmark_tests_main.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "4C_utils_singleton_owner.hpp"
+
+#include <benchmark/benchmark.h>
+#include <Kokkos_Core.hpp>
+#include <mpi.h>
+
+int main(int argc, char** argv)
+{
+  using namespace FourC;
+  benchmark::Initialize(&argc, argv);
```
sebproell, line 19: Should this maybe happen after MPI and Kokkos?

cmake/functions/four_c_auto_define_benchmark_tests.cmake
```diff
@@ -0,0 +1,151 @@
...
+# Pickup all the source files in the current directory and add them to a benchmark test target.
+# Recursively add all subdirectories that contain CMakeLists.txt files. A module under test
+# may be supplied via MODULE. If no module name is supplied, the name of the parent module is used.
+#
+# A source file is analyzed for certain suffixes, to determine how the test should be run. The supported
+# suffixes are:
+#
+# - .npX.: Run the test with X MPI processes.
+# - .threadsX.: Run the test with X OpenMP threads.
+##
```
sebproell, line 76: Copy-paste relic?

sebproell (commented):
Regarding CMake: I guess we don't need all the MPI and OMP logic for the benchmark. For testing, we just want to make sure the benchmark compiles and runs. For a real measurement, selecting appropriate MPI/OMP settings is a user task. So I think you can remove much of the logic you copied over from gtest.

tests/benchmark_tests/4C_benchmark_tests_main.cpp
amgebauer: I don't think this matters since the initialize does not use neither mpi nor kokkos. But it looks cleaner.

cmake/setup_tests.cmake
```diff
@@ -50,6 +51,37 @@ else()
...
```
sebproell, line 68: You adapted the "TPL" related to gtest. Can you also change this one?

sebproell (approved):

sebproell (approved):

tuchpaul (approved):

PR #624, opened 2025-04-07
Add zlib to cmake build process and update imcs cluster preset

knarfnitram:
## Description and Context

On the IMCS cluster the `zlib` version is not located under `/usr/lib/*` . In such cases, the `-lz` flag is required to ensure proper linkage. To avoid manually maintaining this flag within the IMCS cluster preset, I integrated `zlib`  by using  `find_package` within the cmake build strategy, allowing cmake to automatically provide the appropriate include and link flags during the build process.

Additionally the IMCS preset is adapted to the new cluster configuration.


mayrmt (approved):

presets/imcs/cluster/CMakePresets.json
```diff
@@ -21,14 +21,10 @@
...
-        "CMAKE_CXX_COMPILER": "$env{OPENMPI_ROOT}/bin/mpic++",
-        "CMAKE_C_COMPILER": "$env{OPENMPI_ROOT}/bin/mpicc",
         "FOUR_C_TRILINOS_ROOT": "$env{TRILINOS_ROOT}",
-        "FOUR_C_QHULL_ROOT": "$env{QHULL_ROOT}",
-        "FOUR_C_BOOST_ROOT": "$env{BOOST_ROOT}",
-        "FOUR_C_HDF5_ROOT": "$env{HDF5_ROOT}",
-        "FOUR_C_CLN_ROOT": "$env{CLN_ROOT}",
-        "FOUR_C_WITH_FFTW": "OFF"
+        "FOUR_C_WITH_FFTW": "ON",
+        "FOUR_C_CXX_FLAGS": "-msse3",
```
sebproell, line 26: Why this one? Wouldn't you want native optimization, i.e., `FOUR_C_ENABLE_NATIVE_OPTIMIZATIONS`?

sebproell (commented):
The zlib change is fine, but you need to add the installation boiler plate. I will try to make this a bit simpler.  For now, just copy and adapt the logic from Boost. Make sure to include the `template` file.

presets/imcs/cluster/CMakePresets.json
knarfnitram: Thanks for the hint, didn't know about this flag.

sebproell (changes requested):
@knarfnitram You also need to add something in 4CConfig.cmake.in and 4CSettings.cmake.in. Just take the surrounding code there as an example. As I said, we should generate this content but to make progress here, just add it.

sebproell (dismissed):
Thanks! Our installation setup is very noisy; I just had to deal with it in #626. Once these PRs are merged, I think I can simplify this a little.

davidrudlstorfer (approved):

PR #626, opened 2025-04-07
[deal.II] Converters for 4C discretization and vectors into deal.II equivalents 

sebproell:
This PR adds converters that allow to construct the equivalent deal.II Triangulation from a 4C Discretization. Additionally, one can convert vectors associated with equivalent FE spaces on the two. These are useful building blocks to further investigate how we can profit from deal.II in this code base.

Part of #506


src/deal_ii/src/4C_deal_ii_context_implementation.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace DealiiWrappers::Internal
+{
+  template <int dim, int spacedim = dim>
+  struct ContextImplementation
+  {
+    //! Store the local mapping between deal.II cells and 4C elements
+    std::map<int, int> cell_index_to_element_lid;
```
amgebauer, line 32: Isn't an unordered_map the better option here? Typically, you only create this (rather large) mapping once and then you only it as a look-up table, right?

src/deal_ii/src/4C_deal_ii_context_implementation.hpp
```diff
@@ -0,0 +1,82 @@
...
+    const auto& fe = context.pimpl_->finite_elements[0];
+
+    if (fe.degree == 1)
+    {
+      // simple linear mapping
+      context.pimpl_->mapping.push_back(dealii::MappingQ<dim, spacedim>(1));
+    }
+    else if (fe.degree == 2)
+    {
+      // create a MappingQEulerian that takes the shift into account
```
amgebauer, line 73: Isn't something missing here?

src/deal_ii/src/4C_deal_ii_dofs.cpp
```diff
@@ -0,0 +1,86 @@
...
+      return std::make_unique<dealii::hp::MappingCollection<dim, spacedim>>(
+          dealii::MappingQ<dim, spacedim>(1));
+    }
+    else if (fe_collection[0].degree == 2)
+    {
+      // complicated case: we need to find the shift of the 4C nodes and create a mapping with
+      // these
+
+      return std::make_unique<dealii::hp::MappingCollection<dim, spacedim>>(
+          dealii::MappingQ<dim, spacedim>(2));
```
amgebauer, line 70: I don't really understand this. Where is the complicated stuff happening? You can also just write

```
FOUR_C_ASSERT_ALWAYS(fe_collection[0].degree == 1 || fe_collection[0].degree == 2, "Polynomial FEs with degree higher than 2 are not supported.");
return std::make_unique<dealii::hp::MappingCollection<dim, spacedim>>(
          dealii::MappingQ<dim, spacedim>(fe_collection[0].degree));
```

amgebauer (dismissed):

src/deal_ii/src/4C_deal_ii_triangulation.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Core::FE
+{
+  class Discretization;
+}
+
+namespace DealiiWrappers
+{
+  /**
+   * Create an equivalent dealii::Triangulation @p tria from a given Core::FE::Discretization
```
marksidoro, line 27: I think its unclear what equivalent means here since it has different results for different dealii::Triangulations. Specifically only in the case of a d:f:T the implementation will have the same partitioning as 4C. In the other cases deal.II might currently repartition the cells in any way it sees fit. Meaning that for some 4C elements that are locally owned triangulation cells might not be. I think we should either clarify this or repartition in the same way as 4C does.

src/deal_ii/src/4C_deal_ii_triangulation.hpp
```diff
@@ -0,0 +1,56 @@
...
+   * @p discretization. The given @p tria can be either
+   *
+   *  - (serial) dealii::Triangulation
+   *  - dealii::parallel::distributed::Triangulation (abbreviated p:d:T)
```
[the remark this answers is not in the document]
sebproell: Yes, p:s:T would work. I've never used it in a productive setting; I would not enable it for now.

marksidoro (approved):

src/deal_ii/src/4C_deal_ii_vector_conversion.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * @brief Create a Map object that maps from 4C dofs to deal.II gids.
+   *
+   * This function is the inverse of create_dealii_to_four_c_map().
+   */
+  template <int dim, int spacedim = dim>
+  Core::LinAlg::Map create_four_c_to_dealii_map(
+      const dealii::DoFHandler<dim, spacedim>& dof_handler,
+      const Core::FE::Discretization& discretization, const Epetra_Map& dealii_to_four_c_map);
```
amgebauer, line 64: Why is it taking an `Epetra_Map` and not a `Core::LinAlg::Map`? And I think it is actually never used somewhere?

amgebauer (dismissed):

src/deal_ii/src/4C_deal_ii_vector_conversion.hpp
sebproell: Right -- removed it.

amgebauer (approved):

marksidoro (approved):

PR #627, opened 2025-04-07
Switch ssti input files from dat to yaml

c-p-schmidt:
## Description and Context
Based on the fix in https://github.com/4C-multiphysics/4C/pull/625 and the script https://github.com/4C-multiphysics/4C/issues/598#issuecomment-2774776846 provided by @davidrudlstorfer I converted the `ssti` input files from dat to yaml format.


c-p-schmidt:
The problem is that --to-yaml creates the absolute paths to files that are read for the created 4C.yaml file; however, this obviously only works on the local machine but not for the pipeline.

davidrudlstorfer (approved):

sebproell (approved):

PR #631, opened 2025-04-08
Docker: don't use a separate image for deal.II

sebproell:
I introduced deal.II into the docker build  here: https://github.com/4C-multiphysics/4C/pull/558 However, the way I did it, was a bit overengineered. I talked with @amgebauer and @ppraegla, and we now believe the better way is to add deal.II to the normal image. Otherwise, we get in trouble with out dependency hashing. Also, we usually want to test with all optional dependencies enabled except in the new job #509.

We should revisit our optional installed dependencies (backtrace, deal.II, paraview, mathja) and create a minimal image but this is a separate task.

Blocks #626


amgebauer (approved):

sebproell:
I rebased #626 onto this branch to check whether deal.II is functional.

sebproell:
deal.II can be found and I can run the tests in #626. I need to fix a few more things over there, but this PR is ready.

ppraegla (approved):

PR #632, opened 2025-04-08
Migrate pre_exodus to yaml

sebproell:
Since a few people are using `pre_exodus` (see #630), this PR brings it to yaml. Specifically, you can use it as before with the following changes to the options and behavior:

- `--exo`: stays the same
- `--bc`: stays the same (custom format)
- `--head`: has to be yaml now (You can use `4C --to-yaml` on an existing dat-style header to convert!)
- `--dat` is renamed to `--out`. Default is to name the file like the exo file with `.4C.yaml` extension.
- You no longer get a `default.head` file. This file does not make any sense for the new input mechanism. Use the schema file to write a header from scratch (or FourCIPP).

I don't think we need any kind of backward compatibility. This tool is used to create new input files, so it is fine to only allow yaml from now on.

Technical note: this PR makes the front-end correct by only working with yaml there. I am currently using the conversion mechanisms from dat to yaml internally. This can easily be changed to write in yaml (in another PR). After all it is just a dat-style string.

FYI @bwirthl @lkoeglmeier @c-p-schmidt @RShaw026


tests/list_of_tests.cmake
```diff
@@ -1956,9 +1956,6 @@ endif (FOUR_C_WITH_MIRCO)
 # four_c_test_cut_test
 four_c_test_cut_test(2)
 
-# four_c_test_pre_processing: generate default head file and test pre_exo using generated head file
-four_c_test_pre_processing(pre_exo_cube 1)
```
davidrudlstorfer, line 1960: I think then we can also trash one further custom cmake testing function in `cmake/functions/four_c_testing_functions.cmake`?

davidrudlstorfer: Oh sorry - haven't seen it

davidrudlstorfer (approved):

lkoeglmeier (dismissed):

c-p-schmidt (approved):
Looks good, thank you @sebproell

davidrudlstorfer (approved):

c-p-schmidt (approved):

PR #642, opened 2025-04-11
Cmake: copy arbitray number of support files for framework tests

sebproell:
Required for #400 

@dharinib98 @mayrmt This should help you.

### DIFF
--- a/tests/list_of_tests.cmake
+++ b/tests/list_of_tests.cmake
@@ -1962,12 +1962,12 @@ four_c_test_nested_parallelism(sohex8_multiscale_macro_2micro.dat sohex8_multisc
 four_c_test_nested_parallelism(tsi_heatconvection_monolithic.dat tsi_heatconvection_monolithic.dat "")
 
 # testing the whole framework: from pre_exodus and 4C to the post-filter:
-four_c_test_framework(tutorial_battery 2 "")
-four_c_test_framework(tutorial_contact_3d 2 "")
-four_c_test_framework(tutorial_fluid 2 "")
-four_c_test_framework(tutorial_fsi 2 "")
-four_c_test_framework(tutorial_fsi_3d 2 fsi_part_struct_solver.xml)
-four_c_test_framework(tutorial_solid 3 xml/multigrid/elasticity_template.xml)
+four_c_test_framework(PREFIX tutorial_battery NP 2)
+four_c_test_framework(PREFIX tutorial_contact_3d NP 2)
+four_c_test_framework(PREFIX tutorial_fluid NP 2)
+four_c_test_framework(PREFIX tutorial_fsi NP 2)
+four_c_test_framework(PREFIX tutorial_fsi_3d NP 2 COPY_FILES ${PROJECT_SOURCE_DIR}/tests/input_files/fsi_part_struct_solver.xml)
+four_c_test_framework(PREFIX tutorial_solid NP 3 COPY_FILES ${PROJECT_SOURCE_DIR}/tests/input_files/xml/multigrid/elasticity_template.xml)
 
 # four_c_test_post_processing
 four_c_test_post_processing(constr3D_MPC_direct.dat 2 2 ndxyz ndxyz 25)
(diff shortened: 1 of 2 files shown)


tests/list_of_tests.cmake
```diff
@@ -1962,12 +1962,12 @@ four_c_test_nested_parallelism(sohex8_multiscale_macro_2micro.dat sohex8_multisc
...
-four_c_test_framework(tutorial_fluid 2 "")
-four_c_test_framework(tutorial_fsi 2 "")
-four_c_test_framework(tutorial_fsi_3d 2 fsi_part_struct_solver.xml)
-four_c_test_framework(tutorial_solid 3 xml/multigrid/elasticity_template.xml)
+four_c_test_framework(PREFIX tutorial_battery NP 2)
+four_c_test_framework(PREFIX tutorial_contact_3d NP 2)
+four_c_test_framework(PREFIX tutorial_fluid NP 2)
+four_c_test_framework(PREFIX tutorial_fsi NP 2)
+four_c_test_framework(PREFIX tutorial_fsi_3d NP 2 COPY_FILES ${PROJECT_SOURCE_DIR}/tests/input_files/fsi_part_struct_solver.xml)
+four_c_test_framework(PREFIX tutorial_solid NP 3 COPY_FILES ${PROJECT_SOURCE_DIR}/tests/input_files/xml/multigrid/elasticity_template.xml)
```
sebproell, line 1970: @dharinib98 @mayrmt You would append your new file from #400 in the back. COPY_FILES takes an arbitrary number of arguments.

tests/list_of_tests.cmake
```diff
@@ -1962,12 +1962,12 @@ four_c_test_nested_parallelism(sohex8_multiscale_macro_2micro.dat sohex8_multisc
...
-four_c_test_framework(tutorial_contact_3d 2 "")
-four_c_test_framework(tutorial_fluid 2 "")
-four_c_test_framework(tutorial_fsi 2 "")
-four_c_test_framework(tutorial_fsi_3d 2 fsi_part_struct_solver.xml)
-four_c_test_framework(tutorial_solid 3 xml/multigrid/elasticity_template.xml)
+four_c_test_framework(PREFIX tutorial_battery NP 2)
+four_c_test_framework(PREFIX tutorial_contact_3d NP 2)
+four_c_test_framework(PREFIX tutorial_fluid NP 2)
+four_c_test_framework(PREFIX tutorial_fsi NP 2)
+four_c_test_framework(PREFIX tutorial_fsi_3d NP 2 COPY_FILES ${PROJECT_SOURCE_DIR}/tests/input_files/fsi_part_struct_solver.xml)
```
davidrudlstorfer, line 1969: why do we want to fully specify the path to the copy file here?

In my opinion we should stay consistent with all file paths

the regular test files do also assume the path to the input file folder?


```suggestion
four_c_test_framework(PREFIX tutorial_fsi_3d NP 2 COPY_FILES fsi_part_struct_solver.xml)
```

sebproell: I thought about the case where you put the support file in the `framework_tests` directory.

tests/list_of_tests.cmake
mayrmt: @sebproell Great! Thanks a lot!

mayrmt (approved):

tests/list_of_tests.cmake
sebproell: That being said, yes, let's try to be more consistent with the paths. Maybe this is something for your planned cleaning of these helper functions?

mayrmt: To be fully path independent of the source directory, we would need to copy all test files to the build directory.

davidrudlstorfer: > I thought about the case where you put the support file in the framework_tests directory.

Is this planned?

But I am fine with the current solution - if one has ideas on a better approach regarding our test paths please add it to #56

davidrudlstorfer (approved):

PR #644, opened 2025-04-11
Correct license identifier

georghammerl:
The correct SPDX identifier of our chosen license includes the "-or-later".

## Interested parties
@4C-multiphysics/maintainer


sebproell (approved):
True. It is correct in the license header in our field :+1:

mayrmt (approved):

c-p-schmidt (approved):

PR #646, opened 2025-04-11
Throw if iterative linear solver does not converge

mayrmt:
## Description and Context

Failed convergence of iterative linear solvers has lead to trouble in the past --- either in solver development or on applications, where "bad" Newton increments are likely to disturb/destroy the convergence of the outer Newton solver. This PR finally throws, if the iterative solver did not converge in the maximum number of iterations.

## Additional Information

If necessary for the pipeline to pass or if desired by the community, I can add an input option to issue a warning instead. In order to be on the safe side, the new default behavior will be to throw.

UPDATE: I have added the option `THROW_IF_UNCONVERGED`, so that the user can overwrite default behavior and only gets a warning instead of an error.

### DIFF
--- a/src/core/linear_solver/src/method/4C_linear_solver_method_iterative.cpp
+++ b/src/core/linear_solver/src/method/4C_linear_solver_method_iterative.cpp
@@ -157,19 +157,33 @@ int Core::LinearSolver::IterativeSolver<MatrixType, VectorType>::solve()
 
   Belos::ReturnType ret = newSolver->solve();
 
-  int my_error = 0;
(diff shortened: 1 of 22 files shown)


sebproell (dismissed):

davidrudlstorfer (dismissed):

mayrmt:
I had to fix a typo.

sebproell (dismissed):

davidrudlstorfer (dismissed):

sebproell (approved):

davidrudlstorfer (approved):

PR #650, opened 2025-04-11
Split rebalancing into more useful parts

sebproell:
Split up a long function into reusable functions. This is useful so I can use all the nice rebalancing techniques for a discretization read from an exodus file. Refactoring powered by Copilot.

### DIFF
--- a/src/core/io/src/4C_io_meshreader.cpp
+++ b/src/core/io/src/4C_io_meshreader.cpp
@@ -109,141 +109,124 @@ void Core::IO::MeshReader::read_mesh_from_dat_file(int& max_node_id)
   read_nodes(input_, node_section_name_, element_readers_, max_node_id);
 }
 
-/*----------------------------------------------------------------------*/
-/*----------------------------------------------------------------------*/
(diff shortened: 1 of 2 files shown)


davidrudlstorfer (dismissed):

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -109,141 +109,124 @@ void Core::IO::MeshReader::read_mesh_from_dat_file(int& max_node_id)
...
+      default:
+        FOUR_C_THROW("Appropriate partitioning has to be set!");
+    }
+
+    return {rowmap, colmap};
+  }
+
+  void rebalance_discretization(Core::FE::Discretization& discret,
+      const Core::LinAlg::Map& row_elements,
+      const Core::IO::MeshReader::MeshReaderParameters& parameters, MPI_Comm comm)
```
sebproell, line 189: We do not require the `node ids` for this function, so I dropped it.

c-p-schmidt (approved):

PR #651, opened 2025-04-11
Rework tensor interpolation framework (+Bugfix)

dragos-ana:
## Description and Context

This PR provides a rework of the tensor interpolation framework, originally [developed](https://onlinelibrary.wiley.com/doi/10.1002/nme.7373) by @abhi-satheesh and implemented by me in our code framework. Specifically, the following work was performed: 
- Added error types for tensor interpolation to control whether the interpolation was successful or not, and (possibly) why not.
- Added struct of parameters used for interpolation.
- Added methods for the numerical computation of the interpolation gradient, i.e., the derivative of the interpolated matrix with respect to the interpolation location. Also added a dedicated unit test for it.
- Added dedicated methods for the computation of the material and spatial stretch tensors.
- Fixed bug in the rotation interpolation: the base matrix is now correctly determined and decomposed (polar decomposition) before entering the loop. This fixes the determination of the relative rotation matrices, where the base matrix decomposition was previously not always available.
- Added various rotation and eigenvalue interpolation types, as shown in the publication of Satheesh et al. 10.1002/nme.7373. Only one combination (R-LOG) is currently implemented.
- Added a more general 1D tensor interpolation unit test, where the intermediary results of the interpolation algorithm are verified in detail. Specifically, we look here  at the interpolated eigenvalues of the logaritmic weighted average routine and the interpolated rotation vectors.
- Corrected several things in the documentation (e.g., the journal names).


dragos-ana:
@abhi-satheesh @c-p-schmidt:  Since you are the developers of the method, I think that this might be of interest to you. 
@c-p-schmidt: As discussed offline, we currently only implement the R-LOG variant of the interpolation. However, all the other types were added in enum classes, serving as a starting point for further implementation.

c-p-schmidt (dismissed):

dragos-ana:
Thanks you for reviewing the PR, @c-p-schmidt and @sebproell!

sebproell (approved):

c-p-schmidt (approved):

Issue #317, opened 2025-02-04
Inconsistencies in 4C input file settings


gilrrei:
# Overview
This issue acts as a collection of inconsistencies in input specifications that were raised by 4C users. This is not related to any type of input files, but with the provided information itself. 

*Note*: The goal of this issue is not to criticize past design decisions, instead to highlight what could be improved, also when adding new parameters, materials etc.  

### Numbering
- For some sections the numbering is implicit in the section name (`SOLVER 1`, `FUNCT2`) whereas other do this within the sections themselves (`MAT 1 ...`).

### Naming
- Abbreviations vs full words, sometimes even for similar functionality, e.g. `OneStepTheta` vs `OST` vs `One_Step_Theta`
- Convention for multi-words names: `two_words` vs `twowords` vs `two-words` vs `two words` vs `TwoWords`
- Lowercase vs uppercase vs a mix
- Case sensitivity: parameter names are case sensitive whereas for bools  `yes`, `YES` etc is allowed
- Use of formula symbol as parameter name `alpha` vs `K`
- `MAT_` prefix in materials
- Deactivation key `No, byfunct` vs `None, none, No, no, Screen, screen, File, file, Both, both`
- Use of non letters in values or naming `PREDICTOR  d(n)` or `STD=COPY_and_GHOST=COPY/GP`

### Arrays
- Variable to store the number of entries `NUMPOINTS 6 TIMES 1 2 3 4 5 6` vs expected `DOMAINBOUNDINGBOX  1e12 1e12 1e12 1e12 1e12 1e12`

### Implicit knowledge
- The use of `-1` to trigger special behavior, e.g. `ABSTOLRES`  
- Implicit defaults `DAMPING  no, No, NO, yes, Yes, YES, Rayleigh, Material`, what does yes do?
- Sizes of arrays

PR #652, opened 2025-04-13
Support mesh input from exodus files

sebproell:
This PR implements the basis to get rid of `pre_exodus` (#630, #60).

The basic idea is the following: Instead of specifying boundary conditions in our wonky `bc` format, use the nice yaml syntax and add this information to your normal input file. The exodus mesh can already be read by 4C (see also #640), so you now just specify the mesh file in your main input file. AFAIK, this would make `pre_exodus` obsolete.

I ported the `tutorial_contact` to this new style and provide some comments on that below.

This PR is a draft to get some feedback on the user interface. I will split out more PRs from this one to prepare the code (I've already started with #647, #650, #649).


tests/input_files/contact3D_exodus_in.4C.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+STRUCTURE GEOMETRY:
+  FILE: "contact3D_exodus_in.exo"
+  ELEMENT_BLOCKS:
+    - ID: 1
+      ELEMENT_NAME: SOLID
+      ELEMENT_DATA: "MAT 1 KINEM nonlinear"
+    - ID: 2
+      ELEMENT_NAME: SOLID
+      ELEMENT_DATA: "MAT 2 KINEM nonlinear"
```
sebproell, line 17: 2. This new section tells us which file to read and which elements to assign. This should be very familiar to pre_exodus users, since it is exactly what you entered in the `bc` file before. Note that the `ELEMENT_DATA` should of course become a clean yaml structure like

```
ELEMENT_DATA:
  MAT: 2
  KINEM: nonlinear
```

but this can also be tackled later. The inline dat-style string is closer to the old code, so I kept this for now.

tests/input_files/contact3D_exodus_in.4C.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+DESIGN SURF MORTAR CONTACT CONDITIONS 3D:
+  - E: 1
+    ENTITY_TYPE: node_set_id
+    InterfaceID: 1
+    Side: Slave
+  - E: 2
+    ENTITY_TYPE: node_set_id
+    InterfaceID: 1
+    Side: Master
```
sebproell, line 68: 3. Boundary conditions gained the new parameter `ENTITY_TYPE` which tells us, what the number `E` actually means. By default, if you specify nothing, the `ENTITIY_TYPE` is `legacy_id` which refers to the `DVOL-NODE` etc. topology which was hand-written. If you specify `node_set_id`, as was done here, we refer to a node set in the exodus file. This is much simpler and does not require the indirect renumbering we did before.

tests/input_files/contact3D_exodus_in.4C.yaml
```diff
@@ -0,0 +1,82 @@
...
+  TIMESTEP: 0.02
+  NUMSTEP: 6
+  MAXTIME: 1
+  DAMPING: "Rayleigh"
+  M_DAMP: 1e-05
+  K_DAMP: 1e-05
+  TOLDISP: 1e-08
+  LINEAR_SOLVER: 1
+  # Important setting: ensure that we use the file input
+  GEOMETRY: file
```
gilrrei, line 30: Why not directly add the elements and their definition in the `STRUCTURAL DYNAMIC` section?
```yaml
geometry:
  file: "contact3D_exodus_in.exo"
  element_blocks:
    - ID: 1
      type: SOLID
      data: "MAT 1 KINEM nonlinear"
    - ID: 2
      type: SOLID
      data: "MAT 2 KINEM nonlinear"
```
Maybe without the `ELEMENT_` prefix in the parameters. Also, related to #317, it would be nice to agree upon a naming convention, e.g., lower vs upper case parameters...

gilrrei: Also, I am pretty sure that the node sets/element blocks/etc in exodus files can have strings as names, not only integers. This would be an amazing feature to enable here as well!

mayrmt: How I think about the in put sections is that `STRUCTURAL DYNAMIC` refers to the time integration aspects (and has been cluttered with lots of other stuff over the years)... So, I like the new geometry section. 

Example: Let's assume, we have a clear separation in time integration and nonlinear solver stuff (as e.g. in FSI or (almost) in fluid problems. Then, it looks more natural to add a geometry section rather than stuffing geometry information into the time integration section. Yes, we are not there yet, but this could be a direction for future restructurings.

src/core/fem/src/condition/4C_fem_condition.hpp
```diff
@@ -32,6 +32,27 @@ namespace Core::Elements
...
+
+    /**
+     * Refers to a node set ID.
+     */
+    node_set_id,
+
+    /**
+     * Refers to an element block ID.
+     */
+    element_block_id,
```
mayrmt, line 54: For the future: Exodus also offers "side sets". Although `pre_exodus` had some side-set-code in it, we never used them in 4C. This might be worthwhile to investigate in the future to see, if and how side sets are superior and simpler in specifying boundary conditions on surfaces.

mayrmt (commented):
@sebproell thanks! At first glance, I like the proposed solution and also the three steps to include mesh and boundary condition information into the input file.

I can see many options for improvement, especially when we know more about certain features and information, that we can incorporate into the exodus file and then directly read it from there.

A topic for the future: at some point, we might want to look into partitioning of exodus meshes. From `Core::IO::MeshReader::get_exodus_mesh_on_rank_zero()`, I deduce that the mesh is still read on proc 0 only. Totally fine! In the future, one could think about partitioning the mesh and then read it in parallel. I just want to mention the option without favoring it -- there are use cases for both approaches. I am happy to talk about that, when the time comes.

sebproell (commented):
I opened #653 to gather various points for future work.

sebproell:
FYI: all input now takes 20s instead of 30s for 1e6 elements on my machine (with 10 MPI ranks). This is just a rough check, and the details of reading need more attention, but we are moving in the right direction.

tests/input_files/contact3D_exodus_in.4C.yaml
ischeider: If the `GEOMETRY: file` option is not used for this purpose, I guess the whole keyword can be erased. The box option is clearly identified by the section keyword `--FLUID|STRUCTURE DOMAIN`, so we don't have to declare it in the DYNAMICS, or is there any catch in it?

mayrmt (dismissed):
Overall, this looks good to me!

mayrmt (approved):

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -81,9 +90,16 @@ void Core::IO::MeshReader::read_and_partition()
...
-  create_inline_mesh(max_node_id);
+  if (exodus_readers_.size() > 0)
+  {
+    read_mesh_from_exodus();
+  }
+  else
+  {
+    read_mesh_from_dat_file(max_node_id);
+    rebalance();
+    create_inline_mesh(max_node_id);
```
isteinbrecher, line 101: Why do we call `rebalance` for the dat file and not for the exodus case?

tests/input_files/contact3D_exodus_in.4C.yaml
isteinbrecher, line 1: Should we add a result test to this file, to check that everything worked as expected?

isteinbrecher (commented):
Thanks for looking in to this @sebproell ! I think this is a big step towards #60. I have two general questions:
- In the current state, is it possible to combine exodus with "explicit" node/element definitions in the input file?
- If we want to support additional file formats, would the main change be just adding another reader similar to `read_mesh_from_exodus`, specific to that new file type?

isteinbrecher (approved):

mayrmt (approved):

PR #655, opened 2025-04-14
Add wedge6 to embeddedmesh

eulovi:
## Description and Context

In this PR, the wedge element is added to the embedded mesh framework. With this, an interface can cut through wedge elements using the cut library. Then the resulting integration cells are processes by the embedded mesh framework. To achieve this, the wedge6 element was added to the geometry pair library. To do this completely, it is necessary to add a framework test in the geometry pair library I think. @isteinbrecher, could we take a look at this?


tests/input_files/embeddedmesh_block_nurbs27_hex8_wedge6.dat
```diff
@@ -0,0 +1,642 @@
...
+22 SOLID WEDGE6 96 65 64 97 69 68  MAT 1 KINEM nonlinear
+// -----------------------------------------------------------------------------
+// Full script used to create this input file.
+// -----------------------------------------------------------------------------
+//"""
+//Script to analyse the applicability of Lagrange multipliers:
+//
+//In this test, we have an embedded mesh (FEM) and a
+//overlapping mesh (IGA) that have similar mesh sizes.
+//The stiffness of both discretizations
```
isteinbrecher, line 556: This should be removed.

tests/list_of_tests.cmake
```diff
@@ -494,6 +494,7 @@ four_c_test(TEST_FILE elch_pulseplating_hex27_moving_boundary.dat NP 3 RESTART_S
 four_c_test(TEST_FILE elch_RCE_rotsymm_pbc.dat NP 2)
 four_c_test(TEST_FILE elch_straight_convection.dat NP 2)
 four_c_test(TEST_FILE elch_test.dat NP 2)
+four_c_test(TEST_FILE embeddedmesh_block_nurbs27_hex8_wedge6.dat NP 1)
```
isteinbrecher, line 497: Can we run this test in parallel?

isteinbrecher (commented):
Thanks for the contribution @eulovi , I left some minor comments.

tests/list_of_tests.cmake
eulovi: Sure!

isteinbrecher (dismissed):

src/geometry_pair/4C_geometry_pair_element.hpp
```diff
@@ -41,7 +41,8 @@ namespace GEOMETRYPAIR
     //! hexahedron
     hexahedron,
     //! tetraeder
-    tetraeder
+    tetraeder,
+    wedge
```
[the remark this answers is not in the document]
eulovi: Oh yes, you're right. I just added the comment to be consistent with the previous code :) could you reapprove?

eulovi:
@isteinbrecher could you re-approve?

Issue #288, opened 2025-01-27
[Trilinos] Update to newer version


maxfirmbach:
## Description

I would like to update the Trilinos version, which is used by 4C. In the past few months there has been quite a lot of bug fixes (see #176).

## Possible Solution and Definition of Done

- [ ] Update on IMCS
- [ ] Update on LNM

## Interested Parties

@mayrmt @c-p-schmidt @davidrudlstorfer

mayrmt:
Good idea!

davidrudlstorfer:
Sounds like a good plan @maxfirmbach

One thing I would like to mention is that it probably would make sense to somehow "limit" or "plan" the Trilinos Updates. For me the Trilinos Updates involve some effort to update on the clusters. This update would mark the third Trilinos Update in the last 6 months.

Would it make sense to somehow tie the Trilinos Updates to 4C release versions? For example create a new 4C release every 6 months and update Trilinos with these releases? Either before or after?

I understand that prominent bug fixes are important but I do not want that we end up updating Trilinos 6 times a year :D

maxfirmbach:
> Sounds like a good plan [@maxfirmbach](https://github.com/maxfirmbach)
> 
> One thing I would like to mention is that it probably would make sense to somehow "limit" or "plan" the Trilinos Updates. For me the Trilinos Updates involve some effort to update on the clusters. This update would mark the third Trilinos Update in the last 6 months.
> 
> Would it make sense to somehow tie the Trilinos Updates to 4C release versions? For example create a new 4C release every 6 months and update Trilinos with these releases? Either before or after?
> 
> I understand that prominent bug fixes are important but I do not want that we end up updating Trilinos 6 times a year :D

As discussed offline, I'm okay with that proposal 👍 .

mayrmt:
@davidrudlstorfer, yes, an update of Trilinos requires time and effort -- so do all other infrastructure steps in 4C such as upgrading compilers or any other part of our development environment or changes in in put parameters and so on...

I do not want to tie Trilinos updates to any schedule. For people who do their research at the intersection of 4C and Trilinos, such a strategy would impose restrictions on their working environment and in fact force them to keep their finalized work in their own branches (at least in their own Trilinos installation). Since research happens in Trilinos as well, we need to be able to update Trilinos when the need arises. 

We could (and I think we should) discuss ideas of making Trilinos updates more independent of the installations at different institutes. I have some thoughts on that. Feel free to reach out, if you want to know more.

[...]

PR #659, opened 2025-04-14
[Trilinos] Update version in docker

### LINKED ISSUE #176: [Trilinos] Update to version XX

maxfirmbach:
## Description and Context

Updating the Trilinos hash (version) in docker.

Closes #176 #288


dependencies/supported_version/Trilinos.txt
```diff
@@ -1 +1 @@
-1eab15637f2998d1e86fe127b78200f2c9687cb5 2025.1 none
\ No newline at end of file
+11a16481fa1af48e3cfdb0aa84c29617ce94479f 2025.1 16.1.0
```
sebproell, line 1: @maxfirmbach Please change this as follows:

```suggestion
11a16481fa1af48e3cfdb0aa84c29617ce94479f 2025.2 16.1.0
1eab15637f2998d1e86fe127b78200f2c9687cb5 2025.1 none
```

This way, the new version works, as well as the old version. Otherwise everybody will get warnings when running with the old version.
Sorry if that was not clear. Maybe we should document it in the CMake/dependency internal docs?

ppraegla: @sebproell What is the plan for testing the supported versions? If the new dependencies image contains the new Trilinos version, we don't know if the old version still works.

sebproell: @maxfirmbach we should not drop support for an old version out of nowhere. With your proposed change, everybody using the old version will get a warning that the version is unknown. 4C then assumes that their version is _more recent_ than the latest known version, which is wrong. Yes, users may fix this by specifying the version manually, but this is not really intended here. I specifically designed the mechanism for what you are trying to do here: if you add a new version, you add it. Removing an old version is breaking. I think we should only do this in releases. 
Maybe we even need to deprecate it first?

sebproell (approved):

Issue #661, opened 2025-04-15
[Trilinos] Move direct solver from Amesos to Amesos2


maxfirmbach:
## Description

Currently we use the `Amesos` package from Trilinos as interface to our direct solvers namely `UMFPACK` and `SuperLU_DIST`. Being able to switch to `Tpetra` means we need to switch to the successor of `Amesos` the newer package called `Amesos2`, which can handle `Epetra` **AND** `Tpetra` data structures.
The calls to the respective solvers in the input files and rest of `4C` should stays the same.

## Possible Solution and Definition of Done

**Preperation**
- [x] Turn `Amesos2` to ON in the Trilinos-configs
- [x] Find respective `UMFPACK` and `SuperLU_DIST` versions working with `Amesos` and `Amesos2` 

**Switching**
- [x] Write an interface to Amesos2
- [x]  Solve possible issues on the way:
  * Only matrices with maps starting from 0…n are supported (we currently use the `Epetra_Reindexer` class for this, might wanna include something in Amesos2 based on reindexing, currently an issue in Trilinos, see https://github.com/trilinos/Trilinos/issues/13205):
    * solved for `UMFPACK`(see https://github.com/trilinos/Trilinos/pull/13696)
    * solved for `SuperLU_DIST` (see https://github.com/trilinos/Trilinos/issues/13987, https://github.com/trilinos/Trilinos/pull/13990)
    * make sure things are tested for `Epetra` and `Tpetra` (see https://github.com/trilinos/Trilinos/pull/14013)
- [x] Update the `Trilinos` version to include the changes made (see https://github.com/4C-multiphysics/4C/pull/684)
- [x] Check if `Amesos2` works correctly by using our current ctest routines
  * Move problematic test cases to `Umfpack` (see https://github.com/4C-multiphysics/4C/pull/681)
  * Adjust tolerance of remaining failing tests
  * Fix issue in explicit strucutre time integration (see #683, #711)

**Further clean-up and deprecation**
- [ ] Remove the #if statements related to `Amesos` code, due to backwards compatibility

maxfirmbach:
This issue is done, thus closing.

PR #662, opened 2025-04-15
[Trilinos] Move direct solver from Amesos to Amesos2

maxfirmbach:
## Description and Context

This PR moves the direct solver interface from `Amesos` to `Amesos2`. In addition the reindexing and linear problem data structure (which are a hard dependency on `Epetra`) are no longer necessary and therefore removed.

In total we have two direct solvers commonly in use throughout our testing suit:
- `UMFPACK`: This should work without any problems, some tests may fail due to tolerances, but there shouldn't be a conceptual problem with using it anymore.
- `Superlu_DIST`: The story here is different some tests pass? many don't. ~~Currently I don't understand the reason why this is failing.~~ Mostly related to gaps in the global index map. Some tests also fail, due to the parallel nature of the preordering done by `ParMETIS`.

The work on this PR is not finished yet. I wan't to get an impression of how many tests are failing and proof my suspicion that those will all be related to `Superlu_DIST` in some way or another.

Related to #661


.typos.toml
```diff
@@ -11,6 +11,7 @@ ue = "ue" # elastic stretch tensor
 ser = "ser" # "switched evolution relaxation" (used for pseudo transient continuation)
 ded = "ded" # derivative w.r.t. electrode variables
 uncomplete = "uncomplete" # used in the context of the uncomplete() call fo a sparse matrix
+Equil = "Equil" # "Equilibration" (used to pre-order a linear system)
```
mayrmt, line 14: Can't we just use the full word instead of our own abbreviation? Or is this particular spelling required by Superlu_DIST?

src/core/linear_solver/src/method/4C_linear_solver_method_direct.hpp
```diff
@@ -64,11 +65,19 @@ namespace Core::LinearSolver
     //! system of equations
     std::shared_ptr<MatrixType> a_;
 
+#if FOUR_C_TRILINOS_INTERNAL_VERSION_GE(2025, 3)
+    //! an abstract amesos solver that can be any of the amesos concrete implementations
```
mayrmt, line 69:
```suggestion
    //! an abstract Amesos2 solver that can be any of the Amesos2 concrete implementations
```

mayrmt (commented):
@maxfirmbach No matter if we use Amesos or Amesos2, we always run a direct solver. Do you know, why some tests need to relax the result tolerance by several orders of magnitude?

mayrmt (dismissed):

mayrmt:
@4C-multiphysics/developers can we have another review for this PR please?

dharinib98 (approved):

src/fluid_ele/4C_fluid_ele_xwall_boundary.cpp
```diff
@@ -139,91 +138,18 @@ void Discret::Elements::FluidXWallBoundary::location_vector(const Core::FE::Disc
...
-        //      for (unsigned j=0; j<dof.size(); ++j)
-        //      {
-        //        lmowner.push_back(owner);
-        //        lm.push_back(dof[j]);
-        //      }
+
 
         // fill the vector with face dofs
         if (this->num_dof_per_face(0) > 0)
           FOUR_C_THROW("set face_ from private to protected and uncomment");
```
mayrmt, line 152: Probably not at the core of this PR, but the error message is wrong now, since the code to be uncommented is deleted.

@sebproell is this part of one of your recent changes?

mayrmt (approved):

mayrmt (approved):
@maxfirmbach I think we should merge. 

I had a look at the EHL tests. Tolerances and material properties are set in a very particular way, which in my opinion justifies the change in tolerance.

isteinbrecher (approved):

dharinib98 (approved):

PR #663, opened 2025-04-16
Some cleanup of the ssi module

c-p-schmidt:
## Description and Context
I'll do some cleanup of the ssi module, where this is the first part to keep the PR reviewable. Some of the main things I did:
  - remove unnecessary header inclusions
    - improve readability of variable names
    - remove unnecessary references to authors
    - clang tidy suggestions:
      - add const's
      - add nodiscard's
      - designated initalization
      - remove redundant qualifiers to improve formatting and thereby readability

### DIFF
--- a/src/ssi/4C_ssi_base.cpp
+++ b/src/ssi/4C_ssi_base.cpp
@@ -22,7 +22,6 @@
 #include "4C_io_input_file.hpp"
 #include "4C_linalg_utils_sparse_algebra_create.hpp"
 #include "4C_mat_par_bundle.hpp"
-#include "4C_membrane.hpp"
 #include "4C_rebalance_binning_based.hpp"
 #include "4C_scatra_timint_implicit.hpp"
 #include "4C_scatra_timint_meshtying_strategy_s2i.hpp"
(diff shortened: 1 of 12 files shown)


dragos-ana (approved):
Left some comments regarding other methods which can be made `const`. Thanks for the cleanup!

dragos-ana (approved):

src/ssi/4C_ssi_base.cpp
```diff
@@ -401,21 +398,19 @@ void SSI::SSIBase::init_discretizations(MPI_Comm comm, const std::string& struct
   }
   // read in the micro field, has to be done after cloning of the scatra discretization
   auto input_file_name = problem->output_control_file()->input_file_name();
-  Global::read_micro_fields(*problem, std::filesystem::path(input_file_name).parent_path());
+  read_micro_fields(*problem, std::filesystem::path(input_file_name).parent_path());
```
sebproell, line 401: You are now using argument-dependent lookup instead of qualified lookup. Is there a specific reason for that? I have no strong opinion, but I would say that ADL is often confusing and implicit.

sebproell (commented):
I like the cleanup except for one thing.

src/ssi/4C_ssi_base.cpp
c-p-schmidt: As just discussed. Clang-tidy highlighted that they are not needed, and I saw that formatting and thus readability of the code is improved when removing them, in my opinion.

sebproell (approved):

PR #666, opened 2025-04-16
Merge porofluidmulti namespaces

bwirthl:
This PR merges the three porofluid-multi-phase namespaces into one, currently called `PoroPressureBased`.

Further clean-up will be necessary to
- Rename the classes
- Rename the tests
- Clean up the enums

@lkoeglmeier @shervas This is my initial suggestion. Now that the resulting naming conflicts are resolved, we can discuss which name we prefer.

### DIFF
--- a/src/porofluid_pressure_based/4C_porofluid_pressure_based_dyn.cpp
+++ b/src/porofluid_pressure_based/4C_porofluid_pressure_based_dyn.cpp
@@ -38,7 +38,7 @@ void porofluidmultiphase_dyn(int restart)
   // access the problem
   Global::Problem* problem = Global::Problem::instance();
 
-  // print problem type and logo
+  // print problem type
   if (Core::Communication::my_mpi_rank(comm) == 0)
   {
     std::cout << "###################################################" << std::endl;
--- a/src/porofluid_pressure_based/4C_porofluid_pressure_based_dyn.cpp
+++ b/src/porofluid_pressure_based/4C_porofluid_pressure_based_dyn.cpp
@@ -41,7 +41,6 @@ void porofluidmultiphase_dyn(int restart)
   // print problem type and logo
   if (Core::Communication::my_mpi_rank(comm) == 0)
   {
-    PoroPressureBased::print_logo();
     std::cout << "###################################################" << std::endl;
     std::cout << "# YOUR PROBLEM TYPE: " << Global::Problem::instance()->problem_name()
               << std::endl;
--- a/src/porofluid_pressure_based/4C_porofluid_pressure_based_dyn.cpp
+++ b/src/porofluid_pressure_based/4C_porofluid_pressure_based_dyn.cpp
@@ -86,7 +86,7 @@ void porofluidmultiphase_dyn(int restart)
       case Inpar::ArteryNetwork::ArteryPoroMultiphaseScatraCouplingMethod::ntp:
       {
         actdis->fill_complete();
-        nearbyelepairs = POROFLUIDMULTIPHASE::Utils::extended_ghosting_artery_discretization(
+        nearbyelepairs = POROFLUIDMULTIPHASE::extended_ghosting_artery_discretization(
             *actdis, arterydis, evaluate_on_lateral_surface, arterycoupl);
         break;
       }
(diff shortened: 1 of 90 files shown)


src/porofluid_pressure_based/4C_porofluid_pressure_based_dyn.cpp
```diff
@@ -41,7 +41,6 @@ void porofluidmultiphase_dyn(int restart)
   // print problem type and logo
```
lkoeglmeier, line 41:
```suggestion
  // print problem type
```

lkoeglmeier (approved):
Thank you very much for your work @bwirthl! It looks very good to me. I just added a small comment that I noticed while reviewing. You don't have to change this comment of course, I just thought you might want to adjust it if you need to change something else anyway.

src/porofluid_pressure_based/4C_porofluid_pressure_based_utils.cpp
```diff
@@ -539,32 +539,7 @@ double POROFLUIDMULTIPHASE::Utils::calculate_vector_norm(
   {
     FOUR_C_THROW("Cannot handle vector norm");
     return 0;
-  }
-}  // calculate_vector_norm()
-
-/*----------------------------------------------------------------------*
- |                                                    kremheller 03/17  |
- *----------------------------------------------------------------------*/
-void POROFLUIDMULTIPHASE::print_logo()
```
c-p-schmidt, line 548: Sad to see this go away ;-)

src/porofluid_pressure_based_elast/4C_porofluid_pressure_based_elast_input.hpp
```diff
@@ -23,32 +23,14 @@ namespace Core::Conditions
...
-  enum VectorNorm
-  {
-    norm_undefined,
-    norm_l1,         //!< L1/linear norm
-    norm_l1_scaled,  //!< L1/linear norm scaled by length of vector
-    norm_l2,         //!< L2/Euclidean norm
-    norm_rms,        //!< root mean square (RMS) norm
-    norm_inf         //!< Maximum/infinity norm
+    undefined,
+    twoway_partitioned,
```
c-p-schmidt, line 32: As you have the name twoway_partitioned, I also thought oneway_partitioned might be available. But since this is not the case, why not simply distinguish between partitioned and monolithic?

src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_input.hpp
```diff
@@ -26,48 +26,25 @@ namespace Core::Conditions
...
-    norm_rms,        //!< root mean square (RMS) norm
-    norm_inf         //!< Maximum/infinity norm
-  };
-
-  //! Handling of non-converged nonlinear solver
-  enum DivContAct
-  {
-    divcont_stop,     ///< abort simulation
-    divcont_continue  ///< continue nevertheless
+    twoway_partitioned_nested,
```
c-p-schmidt, line 34: The same here. You could think about renaming.
But it can, of course, be in a follow-up, as this would add further changes and probably also changes to input files.

src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_utils.cpp
```diff
@@ -258,81 +259,4 @@ void PoroMultiPhaseScaTra::Utils::assign_material_pointers(const std::string& st
...
-  else
-  {
-    FOUR_C_THROW("Cannot handle vector norm");
-    return 0;
-  }
-}  // calculate_vector_norm()
-
-/*----------------------------------------------------------------------*
- *----------------------------------------------------------------------*/
-void PoroMultiPhaseScaTra::print_logo()
```
c-p-schmidt, line 314: How often did Johannes duplicate it? 🤣

c-p-schmidt (approved):
Changes look good to me. Thank you for the contribution.

shervas (approved):

src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_input.hpp
bwirthl: I cleaned up the enums but will postpone all further clean ups with require changes to the input files to a follow-up.

shervas (approved):

lkoeglmeier (approved):

PR #667, opened 2025-04-17
Collect all mesh input helpers inside MeshReader

sebproell:
Provide a minimal interface and hide all the implementation details.


c-p-schmidt (dismissed):

c-p-schmidt (approved):
The new interface is way nicer! Thanks!

Issue #653, opened 2025-04-14
Support direct input of exodus meshes


sebproell:
#652 implements a basic version of direct exodus mesh input for structure problems.

### List of points to consider for exodus mesh support

- [ ] Add support for all physics
- [ ] Add support for conditions on element blocks
- [ ] Basic performance checks: time and memory, reading on multiple ranks?
- [ ] Investigate side sets?

This is a subtask of #60.

sebproell:
More detailed timings for 1e6 elements:


```
                                        TimeMonitor results over 10 processors

Timer Name                                     MinOverProcs     MeanOverProcs    MaxOverProcs    MeanOverCallCounts    
-----------------------------------------------------------------------------------------------------------------------
Core::IO::MeshReader::read_mesh_from_exodus    20.66 (1)        20.67 (1)        20.68 (1)       20.67 (1)             
Rebalance::RebalanceGraph                      12.97 (1)        12.97 (1)        12.97 (1)       12.97 (1)             
Rebalance::rebalance_node_maps                 12.99 (1)        12.99 (1)        12.99 (1)       12.99 (1)             
exodus in                                      1.136e-06 (1)    0.05333 (1)      0.5333 (1)      0.05333 (1)           
make elements rank 0                           3.34e-07 (1)     0.1718 (1)       1.718 (1)       0.1718 (1)            
make nodes rank 0                              3.33e-07 (1)     0.03243 (1)      0.3243 (1)      0.03243 (1)           
rebalance                                      17.28 (1)        17.57 (1)        17.6 (1)        17.57 (1)             
=======================================================================================================================
```

This means that almost all the time is spent for rebalancing! The construction of the discretization takes around 2.5 s (down from 14s if reading/constructing from dat).

sebproell:
@maxfirmbach Yes.

```
                                         TimeMonitor results over 10 processors

Timer Name                                       MinOverProcs    MeanOverProcs    MaxOverProcs    MeanOverCallCounts    
------------------------------------------------------------------------------------------------------------------------
Core::IO::MeshReader::Rebalance                  17.31 (1)       18.62 (1)        18.77 (1)       18.62 (1)             
Core::IO::MeshReader::read_mesh_from_dat_file    13.09 (1)       13.22 (1)        14.4 (1)        13.22 (1)             
Rebalance::RebalanceGraph                        12.91 (1)       12.91 (1)        12.91 (1)       12.91 (1)             
[...]

PR #668, opened 2025-04-17
Enabling reading element blocks direct exodus mesh

shervas:
This PR enables reading the nodes in the element blocks directly from an exodus mesh.

Checked in porofluid multiphase analysis, where in this case the initial conditions are applied to a block.


sebproell:
Can you rebase onto main? You included a few commits that are already merged. The failed check looks like a (rare) unrelated failure. Let's see if this goes away when you push again.

sebproell:
@c-p-schmidt With this, we could now port the `tutorial_battery` to the direct exodus input.

c-p-schmidt:
> @c-p-schmidt With this, we could now port the `tutorial_battery` to the direct exodus input.

Nice, I'll look into that, but I'm not sure whether I will manage today.

src/global_data/4C_global_data_read.cpp
```diff
@@ -1521,6 +1552,9 @@ void Global::read_conditions(
   std::map<int, std::vector<int>> node_sets;
   get_node_sets_from_mesh(node_sets, mesh_reader);
 
+  std::map<int, std::vector<int>> element_block_nodes;
+  get_element_block_nodes_from_mesh(element_block_nodes, mesh_reader);
```
sebproell, line 1556: Reminder to myself: this might be problematic for very large meshes. We already have an open point to evaluate the performance for larger examples in #653.

sebproell (approved):

c-p-schmidt (approved):
Nice!

PR #678, opened 2025-04-22
Update documentation on doxygen

sebproell:
This page was quite outdated. I am wondering whether we should remove most of the content here? The features of doxygen are listed elsewhere. Instead, we should probably explain a little more what constitutes good documentation.

Of course, the easiest way to get good documentation is to already have really good documentation so everybody mimics it in their new code (see e.g. deal.II).

### DIFF
--- a/doc/documentation/doxygen.rst
+++ b/doc/documentation/doxygen.rst
@@ -12,33 +12,25 @@ What is Doxygen and why does |FOURC| rely on it?
 
 Doxygen is a documentation generator, i.e. a tool for writing software reference documentation.
 The documentation is written within the code, so documentation lives close by to the code that it is documenting.
-This makes it relatively easy to keep up to date.
+This makes it relatively easy to keep up-to-date.
 Doxygen can cross reference documentation and code, so that the reader of a document can easily refer to the actual code.
+
+In contrast to the high-level documentation you are currently reading, Doxygen generates the more technical
+documentation of the code's API and relevant internals.
 
 |FOURC|'s Policy
 ~~~~~~~~~~~~~~~~
 
-|FOURC| requires a set of minimum Doxygen items for each code entity.
-
-At a bare minimum, |FOURC| requires that
-
-- any source code file has a file header consisting of the tags
-    - ``@file``
-    - ``@brief``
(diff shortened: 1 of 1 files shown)


doc/documentation/doxygen.rst
```diff
@@ -253,47 +216,36 @@ When documenting an ``enum``, use something along the lines of the following:
...
+       * Description of another value of the enum.
+       */
+      something_else,
     }; // end of enum EnumName
 
 Typedefs and Usings
 """"""""""""""""""""""
 
-When documenting a ``typedef``, the syntax is essentially the same as that used for member data:
+When documenting a ``using`` declaratin, the syntax is essentially the same as that used for member data:
```
mayrmt, line 234:
```suggestion
When documenting a ``using`` declaration, the syntax is essentially the same as that used for member data:
```

doc/documentation/doxygen.rst
```diff
@@ -574,18 +497,16 @@ There are a handful of different ways to include code blocks in Doxygen, but the
...
+     *     int   argc,
+     *     char* argv[])
+     *   {
+     *     std::cout << "Hello World!" << std::endl;
+     *   } // end of main()
+     *  \endcode
      *  that prints "Hello World!".
      */
 
 The argument to the ``\code{}`` command is a file extension that'll tell Doxygen what kind of syntax highlighting it should use.
```
mayrmt, line 511:
```suggestion
The argument to the ``\code{}`` command is a file extension that will tell Doxygen what kind of syntax highlighting it should use.
```

mayrmt (approved):
@sebproell thanks for the fixes! I am actually wondering, whether we could/should remove even more of the doxygen example code. I think that we need to specify, what type of documentation we want, but we don't have to show how to achieve this with doxygen. Everybody should be able to figure out doxygen syntax on their own.

The current state is fine for me to be merged. If others also fell that this could be reduced even further, we can do it in this or another PR.

ppraegla (approved):

sebproell:
Okay, let's merge, and I'll remove the detailed explanation in favor of some comments on how to write useful documentation.

Issue #192, opened 2025-01-08
[Trilinos] Switch from Intrepid to Intrepid2


maxfirmbach:
## Description

Currently we use `Intrepid` to obtain quadrature rules (points and weights) to calculate the local element matrices. The package will be deprecated soon, thus we need to switch to it's successor `Intrepid2`, which is fully build on `Kokkos`.

## Possible Solution and Definition of Done

**Preparation**
- [x] Discuss which functionality to port and overall integration of the package: For now we agreed on a one-to-one replacement of `Intrepid` with `Intrepid2` with additional features might be included later on  
- [x] Find `Intrepid` relevant code  
- `src/core/fem/src/general/utils/4C_fem_general_utils_gausspoints.cpp`
- `src/core/fem/src/general/utils/4C_fem_general_utils_polynomial.cpp`
- [x] Turn `Intrepid2` to ON in the Trilinos-configs

**Switching**
- [x] Replace `Intrepid` code parts with `Intrepid2` (see 1261-Move-from-intrepid-to-intrepid2)
- [x] Adapt unit- and framework testing if necessary

**Further clean-up and deprecation**
- [x] Remove `Intrepid` from the Trilinos-configs related to docker and github workflows (see https://github.com/4C-multiphysics/4C/pull/684)
- [x] Remove the `#if` statements related to `Intrepid` code, due to backwards compatibility

maxfirmbach:
This issue is done, thus closing.

PR #684, opened 2025-04-24
[docker] Update Trilinos config and hash

maxfirmbach:
## Description and Context

This PR again updates the `Trilinos` hash (for the sake of getting #662 done), adds a new version number to it and removes `Intrepid` from the configs.

Related to #661 and #192


dependencies/supported_version/Trilinos.txt
```diff
@@ -1,2 +1,3 @@
+8cdf0375d278598c12edc3fd7059cb133c6f4426 2025.3 16.2.0
 11a16481fa1af48e3cfdb0aa84c29617ce94479f 2025.2 16.1.0
 1eab15637f2998d1e86fe127b78200f2c9687cb5 2025.1 none
```
ppraegla, line 3: What is the plan with the different Trilinos version when you work on #662? Are there going to be `ifdef`s for the Amesos2 code that just works with the new Trilinos version or do we have to drop the other Trilinos versions?

@sebproell We need to come up with something to test the different versions.

ppraegla: Thanks for the explanation. 

I also thought about testing the oldest and newest version as a compromise. I will think about how we can implement the testing. Maybe something like a nightly test for the oldest Trilinos version, and an additional flag we can set in a PR that triggers testing with the oldest Trilinos version. This may be helpful if a PR touches a lot of Trilinos stuff, like #662. Though this requires some more thinking about how we manage the different Docker images in that case.

ppraegla (approved):

sebproell (approved):
Let's see how things develop with the supported versions.

dependencies/supported_version/Trilinos.txt
vovannikov: What is this list about? Is it planned to make 4C with all of these versions? I have made a modification to Trilinos recently, that, according to my plan, is required for the upcoming changes in the one step solver of 4C (it won't work with the old versions).

sebproell: @vovannikov You would add the newest version at the top of the list and give it a new internal version (2025.4). The third entry is a purely informative name for the version of Trilinos itself. If you pick a commit from develop, this would be "None". You need to guard your code with `FOUR_C_TRILINOS_INTERNAL_VERSION_GE`. Regarding testing: we plan to eventually test the newest version and the oldest version in the list. Currently, we only test the newest version.

Issue #608, opened 2025-04-03
Creating tutorials from test files


gilrrei:
## Description
We already have some nice tutorials in the repo :) However, I just realised we could use Python with [Jinja2](https://github.com/pallets/jinja/) to create templated tutorials based on (existing) test input files (here written in markdown):

#### Example template
<pre>

# Some good tutorial name

This is a tutorial for blabla and is build upon {{ test_file("solid_ele_hex8_Standard_shell_eas_ans.4C.yaml") }}

## Setting up the problem

In this example we investigate a {{ reference_documentation(["PROBLEM TYPE","PROBLEMTYPE","Structure"]) }}:

```yaml
{{ section("solid_ele_hex8_Standard_shell_eas_ans.4C.yaml", "PROBLEM TYPE")}}
```

with {{ len(load("solid_ele_hex8_Standard_shell_eas_ans.4C.yaml", "STRUCTURE ELEMENTS")) }} elements:
```yaml
{{ section("solid_ele_hex8_Standard_shell_eas_ans.4C.yaml", "STRUCTURE ELEMENTS")}}
```
</pre>

#### Rendered template (links not working here, but of course afterwards they would)

<details>

<summary>Click me for the rendered tutorial </summary>

# Some good tutorial name

This is a tutorial for blabla and is build upon [solid_ele_hex8_Standard_shell_eas_ans.4C.yaml](path_to_file)

## Setting up the problem

In this example we investigate a [Structure](path_to_documentation):

```yaml
PROBLEM TYPE:
  PROBLEMTYPE: Structure
```

with 1 elements:
```yaml
STRUCTURE ELEMENTS:
  - 1 SOLID HEX8 1 5 6 2 3 7 8 4 MAT 1 KINEM nonlinear TECH shell_eas_ans
```
</details>

## The idea:
Where we provide utility functions, such as:
- `test_file`: path to the test files
- `reference_documentation`: reference documentation for this particular parameter which is also generated by the metadata file
- `load`: load the content of the file
- `section`: add a YAML section here. The function could even generate the <pre> ```yaml </pre>

The point is we provide infrastructure to write tutorial **templates** that get their data from other sources (test input file, metadata) that are always synced with the code. This way, we only have to change the keys in the tutorials, and with nice error messages in the functions listed above, it becomes easy to fix whatever is needed. QUEENS already uses Jinja2 in the tutorials, so expertise can be borrowed by friendly projects.

## Interested Parties
@sebproell @ischeider @shervas

[...]

PR #376, opened 2025-02-14
Make PROBLEMTYPE a required parameter

sebproell:
4C just gained the ability to do other things than FSI ;)

PR #694, opened 2025-04-25
[Documentation] Introduce jinja templating for tutorials

ischeider:
## Description and Context

With this PR, yaml information can be directly included in the tutorials (and in principle also in all other documentation files) based on the new script `include_yaml.py`, which does the conversion. Similar to the code snippet shown in issue #608, the control strings are inserted into the tutorial template like this:

```
{{ section("solid_ele_hex8_Standard_shell_eas_ans.4C.yaml", "PROBLEM TYPE")}}
```

If a file is used more than once in a tutorial, a variable containing the file definition can be placed on the top of the file:

```
{% set solid_tutorial_base_file = "tutorial_solid.4C.yaml" %}
{% set base_content = load_input_file(solid_tutorial_base_file)%}

3D Solid tutorial 
===================
{{ section_dump(base_content, "PROBLEM TYPE") }}

Here comes normal text...

{{ section_dump(base_content, ["STRUCTURE GEOMETRY", "INCLUDES"]) }}

```
It is possible to enter a number of section commands as a list, as one can see above.

The commands (like `section_dump`) are defined in the python script, the list of recognized commands can be extended as needed. The conversion is triggered by the sphinx configuration script (`doc/conf.py.in`), where the function `include_yaml.convert(...)` is called.

For now, the solid mechanics tutorial contains these jinja commands, and the complete workflow has changed - it does not use pre_exodus anymore, but the 4C input is simply created by the exodus II geometry and the additional header and boundary condition information. However, all other tutorials still don't have any jinja controls and the input is still the old one. Feel free to change these input files as well.

Note, that even most of the tutorials are written in restructuredText, the jinja controls also work for markdown files. Then the jinja control must have an additional filetype parameter, for example:

```
{{ section("solid_ele_hex8_Standard_shell_eas_ans.4C.yaml", "PROBLEM TYPE", "md")}}
```

Is related to: #608 (jinja templating), #653 (direct exodus read)

Interested parties: @gilrrei @shervas


doc/documentation/python/include_yaml.py
```diff
@@ -0,0 +1,88 @@
...
+        rststring = ""
+        yaml_data = yaml.safe_dump(data, sort_keys=False).split("\n")
+        for line in yaml_data:
+            rststring += "    " + line + "\n"
+        rststring += "\n"
+        return ".. code-block:: yaml\n\n" + rststring + "\n"
+    elif filetype == "md":
+        return "```yaml\n" + yaml.safe_dump(data, sort_keys=False) + "```\n"
+    else:
+        return yaml.safe_dump(data, sort_keys=False)
```
gilrrei, line 37:
```suggestion
    if filetype == "rst":
        rststring = ""
        yaml_data = yaml.safe_dump(data, sort_keys=False).split("\n")
        for line in yaml_data:
            rststring += "    " + line + "\n"
        rststring += "\n"
        return ".. code-block:: yaml\n\n" + rststring + "\n"
    elif filetype == "md":
        return "```yaml\n" + yaml.safe_dump(data, sort_keys=False) + "```\n"
    else:
        raise TypeError(f"Filetype {filetype} for yaml_dump cannot be recognized yet.")
```

doc/documentation/python/include_yaml.py
```diff
@@ -0,0 +1,88 @@
...
+
+def convert(template_path, rendering_path, input_file_path):
+    global PATH_TO_TESTS
+    PATH_TO_TESTS += input_file_path
+    target_dir = pathlib.Path(rendering_path)
+
+    for template_file in pathlib.Path(template_path).glob("*.j2"):
+
+        template = jinja2.Template(template_file.read_text())
+        tutorial_name = pathlib.PurePath(template_file).stem
```
gilrrei, line 58:
```suggestion
        tutorial_name = template_file.stem
```

doc/documentation/python/include_yaml.py
```diff
@@ -0,0 +1,88 @@
...
+            template.render(
+                section_dump=section_dump,
+                load_input_file=load_input_file,
+                yaml_dump=yaml_dump,
+                len=len,
+            )
+        )
+    for filetype in RECOGNIZED_TYPES:
+        for file in pathlib.Path(template_path).glob("*." + filetype):
+            shutil.copy(file, target_dir)
```
gilrrei, line 71: I don't fully get it, what does this step do :sweat_smile: ?

doc/documentation/src/tutorial_templates/tut_solid_preexo.rst.j2
```diff
@@ -1,3 +1,9 @@
+{% set solid_tutorial_base_file = "tutorial_solid.4C.yaml" %}
+{% set solid_tutorial_bc_file = "tutorial_solid_bc.4C.yaml" %}
+{% set solid_tutorial_header_file = "tutorial_solid_head.4C.yaml" %}
+{% set base_content = load_input_file(solid_tutorial_base_file)%}
+{% set bc_content = load_input_file(solid_tutorial_bc_file)%}
+{% set header_content = load_input_file(solid_tutorial_header_file)%}
```
gilrrei, line 6:
```suggestion
{% set base_content = load_input_file(solid_tutorial_base_file) %}
{% set bc_content = load_input_file(solid_tutorial_bc_file) %}
{% set header_content = load_input_file(solid_tutorial_header_file) %}
```

gilrrei (commented):
Thanks for introducing this! I like it a lot!

With #376, validating partial 4C input files will lead to validation errors (based on the schema) if the required section is not present, i.e. if you use `INCLUDES`. For this #674, added a second schema for these cases and I had the idea to do a different ending `4C.part.yaml` to make it easier to differentiate between a complete and hierarchical input file.

For this  small example, it might be easier to add all the information into a single `4C.yaml`, in particular since this is the default way to use 4C input.

doc/documentation/python/include_yaml.py
ischeider: There may be files in the tutorial_templates directory, which are not templates but already restructuredText or markdown files. These are simply copied to the target directory. Currently this holds for the `tutorials.rst` file.

Actually it would be possible to have another directory for the files which are not templates, but if it would be named "tutorials", this directory would be overwritten by the symbolic link to the directory containing the converted files.

Should we rather put those files in a different directory, and which name should this have then - or how should we name the directory for the converted tutorial files?

gilrrei: I see, thanks for the info. My first instinct would be to put the templates in a separate folder and exclude those in the sphinx build (using exclude patterns). Afterward, during rendering, the rendered tutorials would be written in the 'normal' tutorial directory where the other files already exist. However, that's just a personal preference

doc/documentation/python/include_yaml.py
```diff
@@ -0,0 +1,85 @@
...
+        section_names = [section_names]
+    yaml_dict = {}
+    for section in section_names:
+        yaml_dict[section] = input_file_data[section]
+    return yaml_dump(yaml_dict, filetype)
+
+
+def convert(template_path, rendering_path, input_file_path):
+    global PATH_TO_TESTS
+    PATH_TO_TESTS += input_file_path
```
gilrrei, line 49: Sorry, I missed this in the first review round. Why do we need this to be a global variable? Also, why modify the variable?

ischeider: @gilrrei The directory of the jinija files (`input_file_path`) is not known at build time here, but only at runtime. Since we call the function `convert` to create the documentation, but need the variable in the independent function `load_input_file`,  a global variable seemed the easiest way to get the information from one place to the other. In the example you provided, the variable was hard-coded, which I wanted to avoid.

gilrrei: I see. But why
```python
PATH_TO_TESTS += input_file_path
```
instead of 
```python
PATH_TO_TESTS = input_file_path
``` 
?

gilrrei (approved):

shervas (approved):

PR #696, opened 2025-04-25
Add vtu output to reduced lung

bgoderbauer:
## Description and Context

Adds a basic implementation for runtime Vtu output to `reduced_lung`. This will be extended with more features in the future when more models and better data structures are available.

I also renamed some variables along the way.

### DIFF
--- a/src/reduced_lung/4C_reduced_lung_helpers.cpp
+++ b/src/reduced_lung/4C_reduced_lung_helpers.cpp
@@ -176,6 +176,61 @@ namespace ReducedLung
 
     return column_map;
   }
+
+  void collect_runtime_output_data(
+      Core::IO::DiscretizationVisualizationWriterMesh& visualization_writer,
+      const std::vector<Airway>& airways, const std::vector<TerminalUnit>& terminal_units,
+      const Core::LinAlg::Vector<double>& dofs, const Core::LinAlg::Map* element_row_map)
+  {
+    Core::LinAlg::Vector<double> pressure_in(*element_row_map, true);
+    Core::LinAlg::Vector<double> pressure_out(*element_row_map, true);
+    Core::LinAlg::Vector<double> flow_in(*element_row_map, true);
+    for (const auto& airway : airways)
+    {
+      [[maybe_unused]] int err = pressure_in.replace_local_value(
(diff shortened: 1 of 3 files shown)


src/reduced_lung/4C_reduced_lung_helpers.cpp
```diff
@@ -176,6 +176,61 @@ namespace ReducedLung
...
+          "Internal error: replace_local_value for runtime output (q_in from terminal units) did "
+          "not "
+          "work.");
+    }
+    visualization_writer.append_result_data_vector_with_context(
+        pressure_in, Core::IO::OutputEntity::element, {"p_1"});
+    visualization_writer.append_result_data_vector_with_context(
+        pressure_out, Core::IO::OutputEntity::element, {"p_2"});
+    visualization_writer.append_result_data_vector_with_context(
+        flow_in, Core::IO::OutputEntity::element, {"q_in"});
```
sebproell, line 232: Your current approach works. However, I thought that this kind of reindexing/interpretation of the vector entries can be done by `append_result_data_vector_with_context`? 

Does this work?
```suggestion
    visualization_writer.append_result_data_vector_with_context(
        dofs, Core::IO::OutputEntity::element, {"p_1", "p_2", "q_in"});
```

ping @c-p-schmidt

isteinbrecher: @sebproell If the ordering in `dofs` is `{p_1_el0, p_2_el0, q_in_el0, p_1_el1, p_2_el1, q_in_el1, ...}` then your suggestion should work. This whole function can likely be replaced by your single line suggestion.

bgoderbauer: Thanks! This is good to know for further refining the output. For now, I would like to keep it in this general form that relies on the indices given in the elements, since the ordering in `dofs` will change with reduced-order elements with more than three dofs. For example, airways with elastic walls have an extra dof `flow_out` and need special treatment in one form or another.

I will return to this topic when we have a better data layout for all the models of different complexities.

sebproell (approved):

c-p-schmidt (approved):
Looks reasonable to me. Further refining the output once your data layout is converged is a good idea, as it should drastically reduce the number of required code lines.

PR #697, opened 2025-04-25
Fix testing tolerances on Ubuntu24

maxfirmbach:
## Description and Context

Some tests fail on my local Ubuntu24 workstation. I think I posted the full list at some point, this PR adjusts the testing tolerance of a bunch of them, such that they pass locally.

@c-p-schmidt For some of the `ssi` related tests the tolerance adjustment is rather high, maybe you can have a detailed look at them if that is still okay.

@isteinbrecher For the `beam` related test, the nonlinear solver did not converge, due to some weird behavior of the predictor. I now use a constant displacement predictor, which works well.

For the scatra turbulence test the linear solver did not converge anymore at some point, thus switching to a more robust incomplete factorization preconditioner instead of algebraic multigrid.


isteinbrecher:
I am fine with the changes to `tests/input_files/beam3r_herm2line3_static_beam_to_solid_surface_contact_ironing_penalty_gap_variation_gauss_point_projection.dat`

c-p-schmidt:
@maxfirmbach: Sorry for the late reply. I had a first look at the `ssi_mono` test, and it seems like the setup is rather crazy with extreme charge rates leading to, for example, spatial oscillations.
So this test might need a rework.

But can you give some more information on your setup? Do you also use other libraries, or compiler settings like -march=native, or is it just different hardware, but the plain settings from the IMCS preset?

mayrmt (approved):

sebproell (dismissed):

c-p-schmidt:
> I had a talk with @c-p-schmidt about the `ssi`, `sti` tests and he will do a little rework, till then this is on hold, so no merging yet.

I'm already at it, but I intend to do this at a larger scale, i.e., for all `ssi`, `ssti`, and `sti` tests. However, I'll focus on the relevant tests in this PR first to avoid unnecessarily postponing the progress here and adapt the other tests in a second round once I've finished the rest.

sebproell (approved):

mayrmt (approved):

c-p-schmidt (approved):

PR #698, opened 2025-04-25
Add Paul Tuch to .mailmap

tuchpaul:
Added my name to the .mailmap


amgebauer:
As long as you only committed with a single mail, we don't need to add you to the mailmap. This file is only for git to group author contributions even if they used different mail addresses.

sebproell:
@tuchpaul As discussed on Friday, you don't need this PR.

PR #700, opened 2025-04-25
Move red_airways input to physics

davidrudlstorfer:
## Description and Context

Separate the input for red_airways from the bio input.

Part of #59


davidrudlstorfer:
@c-p-schmidt maybe the 4C help hotline knows the problem I encounter again :D

```bash
terminate called after throwing an instance of 'Teuchos::Exceptions::InvalidParameterType'
  what():  Error!  The parameter "solver type" exists
in the parameter (sub)list "ANONYMOUS"
but it does not contain any validator needed to extract
an integral value of type "FourC::Airway::RedAirwaysDyntype"!
```

I am a bit clueless on why this fails.

Also @bgoderbauer maybe you could check if these changes make sense regarding the reduced_lung as well. The main point of this PR is to move some input out of inpar again

bgoderbauer:
Thanks for tackling this! :raised_hands: 
Moving the input to red_airways makes a lot of sense. I fully support this change.

bgoderbauer (approved):

sebproell (approved):

PR #701, opened 2025-04-25
Remove PoroFluidMultiphase meshtying std

bwirthl:
Remove unnecessary meshtying class `std` which meant no meshtying at all.


src/porofluid_pressure_based/4C_porofluid_pressure_based_timint_implicit.cpp
```diff
@@ -536,7 +532,11 @@ void PoroPressureBased::TimIntImpl::solve()
 /*----------------------------------------------------------------------*
  | contains the call of linear/nonlinear solver             vuong 08/16 |
  *----------------------------------------------------------------------*/
-void PoroPressureBased::TimIntImpl::update() { strategy_->update(); }
+void PoroPressureBased::TimIntImpl::update()
+{
+  assert(artery_coupling_active_ == true);
+  meshtying_->update();
```
bwirthl, line 538: These methods should only be called if artery-coupling is active. Not sure how to handle these cases better. Would probably require more refactoring?

src/porofluid_pressure_based/4C_porofluid_pressure_based_timint_implicit.cpp
```diff
@@ -793,7 +796,8 @@ std::shared_ptr<const Core::LinAlg::Map> PoroPressureBased::TimIntImpl::dof_row_
  *----------------------------------------------------------------------*/
 std::shared_ptr<const Core::LinAlg::Map> PoroPressureBased::TimIntImpl::artery_dof_row_map() const
 {
-  return strategy_->artery_dof_row_map();
+  assert(artery_coupling_active_ == true);
+  return meshtying_->artery_dof_row_map();
```
bwirthl, line 800: Same here. And below.

src/porofluid_pressure_based/4C_porofluid_pressure_based_timint_implicit.cpp
bwirthl: I changed this to `FOUR_C_ASSERT`.

src/porofluid_pressure_based/4C_porofluid_pressure_based_timint_implicit.cpp
```diff
@@ -2396,7 +2450,9 @@ void PoroPressureBased::TimIntImpl::fd_check()
  *----------------------------------------------------------------*/
 std::shared_ptr<Adapter::ArtNet> PoroPressureBased::TimIntImpl::art_net_tim_int()
 {
-  return strategy_->art_net_tim_int();
+  FOUR_C_ASSERT(artery_coupling_active_,
```
sebproell, line 2453: Maybe even `FOUR_C_ASSERT_ALWAYS`? Depends on whether you think it is plausible that users can end up here by choice of input parameters or calling public APIs. Then, I would go for the always-version. If this would be more of an implementation error, the current version is fine.

bwirthl: This would be an implementation error.

src/porofluid_pressure_based/4C_porofluid_pressure_based_timint_ost.cpp
```diff
@@ -139,7 +139,10 @@ void PoroPressureBased::TimIntOneStepTheta::compute_time_derivative()
 void PoroPressureBased::TimIntOneStepTheta::update()
 {
   // call base class
-  PoroPressureBased::TimIntImpl::update();
+  if (artery_coupling_active_)
+  {
+    PoroPressureBased::TimIntImpl::update();
+  }
```
sebproell, line 145: That's a bit surprising. The name of the base class function is generic, yet you only call it when you have the coupling. This kind of OOP code is hard to review; there might be a reason I am not seeing due to the lack of arguments to this function.

sebproell (approved):

src/porofluid_pressure_based/4C_porofluid_pressure_based_timint_ost.cpp
bwirthl: I totally agree. I had a closer look at this and will try to improve it as part of the next PR where I plan to merge the two time integration classes (implicit and OST).

c-p-schmidt (approved):

src/porofluid_pressure_based/4C_porofluid_pressure_based_meshtying_strategy_artery.cpp
```diff
@@ -18,20 +18,27 @@
...
 #include "4C_porofluid_pressure_based_utils.hpp"
 
-#include <Teuchos_TimeMonitor.hpp>
+#include <Teuchos_StandardParameterEntryValidators.hpp>
 
 FOUR_C_NAMESPACE_OPEN
 
 
 /*----------------------------------------------------------------------*
  | constructor                                (public) kremheller 04/18 |
```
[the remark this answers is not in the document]
bwirthl: No wait, this was intentional, without I cannot compile. What should be there in your opinion? I need `Teuchos::getIntegralValue` and before this was included via the base class.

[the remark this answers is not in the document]
sebproell: The comment. Just leave it here ;)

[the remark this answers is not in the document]
c-p-schmidt: I was just referring to these unnecessary comments like
`| constructor                            (public) kremheller 04/18 |`

and not to the changed include. Sorry, if that was unclear.

Issue #683, opened 2025-04-23
[BUG] New explicit structure time integration results in Amesos error


maxfirmbach:
## Bug Description: What's going wrong?

The following tests
```terminal
sosh8_freeflying_ruler_sta_ab2_new.dat-p2
sosh8_freeflying_ruler_sta_centrdiff_new.dat-p2_for_sosh8_freeflying_ruler_sta_centrdiff_new_restart.dat-p2-restart_step_70
sosh8_freeflying_ruler_sta_centrdiff_new_restart.dat-p2_from_sosh8_freeflying_ruler_sta_centrdiff_new.dat-p2-restart_step_70
sosh8_freeflying_ruler_sta_expleuler_new.dat-p1
```
pass the `ctest` routine despite **a lot** of `AMESOS ERROR -22` errors usually resulting in a segfault/throw, due to a breakdown in the symbolic or numeric factorization procedure. This definitely seems wrong and such kind of errors should not appear, in particular in our testing ...

## Expectations

- [ ] Testing runs with the listed test routines without errors.

## Possible Solution

It seems like something is wrong with the system matrix.

## Steps to Reproduce

Execute the tests from above with the current `main` branch and look at the output.

mayrmt:
@maxfirmbach If one uses a consistent mass matrix, it needs to be inverted, which requires a single linear solve. This could be avoided by using a lumped (=diagonal) mass matrix. I don't know for sure, but I guess that this is the main motivation in the present scenario.

The use of the `SingleStep` class from `NOX` comes from the integration of `NOX` into the new solid time integration.

vovannikov:
I think @mayrmt is right regarding the reason for the use of `SingleStep` solver: it is indeed a way to inverse the mass matrix for this kind of equation:
![Image](https://github.com/user-attachments/assets/d91d461f-eb68-41f5-84da-0445d9766056)
It is actually a common approach for a generalized explicit time integrator that can get an arbitrary mass matrix and has no apriori knowledge on whether it is consistent, lumped or of some other type.

@maxfirmbach, I also think that you are correct and even for lumped matrices the exact same procedure based on `SingleStep` from `NOX` is used, since the explicit integrator is not aware of whether the matrix is lumped or not. What @mayrmt probably meant, is that diagonal mass matrix is always invertible (unless it has a zero on the diagonal), so the linear solver won't have any issues working with it.

[...]

PR #704, opened 2025-04-25
[structure] Debug time adaptivity with joint explicit

maxfirmbach:
## Description and Context

Using the new structure time integration with time adaptivity based on the joint explicit scheme results in a singular system matrix consisting of only zeros for the explicit time integration scheme. This PR serves for debugging and discussion purpose.

There are several open questions to me:
- Why is the system matrix for the explicit time integrator zero and thus singular? (see #711)
- Why do we even need to solve a linear system for the explicit time integrator in the first place (and especially in the lumped mass matrix case)? (explanation given in #683)
- Why do we funnel things through the NOX single step class for the explicit time integration case? (explanation given in #683)
- It seems like the linear solve, correct mass matrix usage etc. doesn't has an influence on the values generated by the auxiliary time integrator and thus the error estimation if a time step is accepted or not.

The current implementation has a bad taste to it ... not only is it actively throwing errors but not stopping it also seems to be a nice feature with a once again hacky vibe to it.

Related to #683

### DIFF
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_linearsystem.cpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_linearsystem.cpp
@@ -292,6 +292,13 @@ bool NOX::Nln::LinearSystem::applyJacobianInverse(Teuchos::ParameterList& linear
   // Zero out the delta X of the linear problem if requested by user.
   if (zeroInitialGuess_) result.init(0.0);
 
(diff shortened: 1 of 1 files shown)


mayrmt:
Since this is probably not supposed to be merged soon, I took the liberty to convert it into draft mode.

vovannikov:
@maxfirmbach I think this should be closed.

PR #705, opened 2025-04-26
[Epetra] Remove unused includes

maxfirmbach:
## Description and Context

This PR removes a bunch of `Epetra` related includes. Now the remaining includes reflect the work done so far way better. `Epetra` objects left over in the physics are mostly related to:
- `Epetra_FEVector`
- `Epetra_FEGraph`
- `Epetra_CrsMatrix`
- `Epetra_Import`
- `Epetra_Export`

which matches the list related to linear algebra objects listed in #189.

Related to #189.

### DIFF
--- a/src/cardiovascular0d/4C_cardiovascular0d.hpp
+++ b/src/cardiovascular0d/4C_cardiovascular0d.hpp
@@ -16,10 +16,6 @@
 #include "4C_linalg_vector.hpp"
 #include "4C_utils_parameter_list.fwd.hpp"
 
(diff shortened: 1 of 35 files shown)


mayrmt (approved):
Any chance that we can automate to protect ourselves against adding unwanted Epetra includes?

sebproell (approved):
@mayrmt I think we do not add so much new code that I would implement a check for Epetra includes. Let's watch out in review. We should check that everybody uses the LinAlg data structures in new code.

mayrmt:
@sebproell I am fine with that.

PR #711, opened 2025-04-28
Call `post_setup()` for the auxiliary time integrator in `StructureTimeAdaJoint`

### LINKED ISSUE #683: [BUG] New explicit structure time integration results in Amesos error

vovannikov:
## Description and Context
The original implementation of `StructureTimeAdaJoint` did not call `post_setup()` for its auxiliary time integrator. This resulted in the empty mass matrix, which is not shared between the two integrators, since the auxiliary one has its own global state. I fixed that. As the result, Amesos should not suffer from `AMESOS ERROR -22` errors.

I also removed the wrapper member, as it looks obsolete.

Fixes #683
Related to #704

@maxfirmbach @mayrmt @vryy


mayrmt (approved):

mayrmt:
@vovannikov could you please open an issue about the double-evaluation of the mass matrix in certain scenarios, so that we don't forget about it? Thanks.

vovannikov:
@maxfirmbach In fact while debugging, I also observed that this modification did not change the behavior of the auxiliary integrator at all, the numbers are exactly as before. But for this the behavior of the explicit integrators as a part of the joint integration strategy has to be checked, I did not go that far. But I think it is worth doing so.

Moreover, even if the results provided by the auxiliary integrator had changed, this still could have had no influence on the overall behavior of the joint time integration strategy, since the auxiliary integrator is used there only for error estimation. And if the error remains below the threshold, then the results keep unaltered.

vovannikov:
> That's true, but shouldn't the time adaptivity feature be tested with something that actually alters the stepping procedure, due to the error estimation being over some threshold? Also something for a separate PR though.

It definitely should, you are absolutely right. But it is indeed for a separate PR.

I have posted another issue #715 listing these TODOs.

amgebauer:
> @vovannikov Added a commit for post_setup() in the restart case for the joint integrator, please have a look if this makes sense to you (no errors are thrown there anymore).

@c-p-schmidt and @dragos-ana might be able to help here.

c-p-schmidt:
> @vovannikov Added a commit for post_setup() in the restart case for the joint integrator, please have a look if this makes sense to you (no errors are thrown there anymore).

As part of the post_setup, the mass matrix is calculated, which is also required for restart simulations. So this might have been the problem you encountered.

mayrmt (dismissed):

vovannikov:
> @vovannikov Added a commit for post_setup() in the restart case for the joint integrator, please have a look if this makes sense to you (no errors are thrown there anymore).

I will have a look now, I think there should be a bit more consistent way to call `post_setup()` in this case. As it is now it looks like for the non-restart case it gets called eventually twice for the auxiliary integrator.

vovannikov:
@maxfirmbach It is OK, I just removed the "else" branch as it is unnecessary.

mayrmt (dismissed):

amgebauer (approved):
Can you cleanup the commits a bit?

c-p-schmidt:
I'm fine with the solution. Just one thought.

Maybe it would be easier to understand if you would not override the `post_setup()` method and instead use the else-case in the `setup_auxiliary` method.

This way, the behavior for reading restart simulations and the "original" simulation would be the same. In contrast, now for the "original" simulation, the post setup of the auxiliary integrator is performed with the overridden post_setup call, while for the restart simulation,n this is performed in the setup_auxiliary method.

c-p-schmidt (approved):

vovannikov:
@c-p-schmidt I got your idea. I have to say that neither of these options is perfect due to the limitations posed by the current design. The main reason to override `Adapter::StructureTimeAdaJoint::post_setup()` is that it is consistent across the code. Mainly, all `post_setup()` calls are located inside someone's else `post_setup()` with some exceptions. The general pattern is that by default `post_setup()` is called, but for a restart situation - `read_restart()` is called.

Another reason to keep that call inside `Adapter::StructureTimeAdaJoint::post_setup()` is that there could be something happening between `Adapter::StructureTimeAdaJoint::setup_auxiliary()` and `Adapter::StructureTimeAdaJoint::post_setup()`. I do not know whether this is intentional, but the current design provides this possibility. I also see from the code that this is violated in quite many parts of the code, but anyways, I stick myself to that what I retrieved as a default behavior.

For this reason I actually do not like that `sta_->post_setup()` is currently invoked inside `Adapter::StructureTimeAdaJoint::setup_auxiliary()` for the restart case, but let it stay there. Actually, probably a better solution would be to override `Adapter::StructureTimeAdaJoint::read_restart()`. In this case  `Adapter::StructureTimeAdaJoint::post_setup()` and  `Adapter::StructureTimeAdaJoint::read_restart()` would kind of mirror each other for different scenarios, both contain explicit calls `sta_->post_setup()`, thus making the intention clear and transparent. What do you think of this option?

c-p-schmidt:
I agree that the main issue is the current design, i.e., that you need to call a sequence of methods (in the right order) to obtain a working object to work with afterwards.

I  like your suggestion to override the `read_restart()` method. Given the current constraints, this seems to be the best option.

vovannikov:
> @vovannikov How should we proceed? Overwrite `read_restart()`?

@maxfirmbach yep, I have force-pushed the corresponding commit.

mayrmt (approved):

c-p-schmidt (dismissed):

vovannikov:
Test `sosh8_freeflying_ruler_sta_centrdiff_new_restart.dat-p2_from_sosh8_freeflying_ruler_sta_centrdiff_new.dat-p2-restart_step_70` fails, but we need to discuss whether this the failure is correct. It looks like the final results differ a bit from the reference:
```
STRUCTURE: structure   dispx    at node  96	 is WRONG --> actresult= 6.35319789272100061e-02, givenresult= 6.45527052251677474e-02, abs(diff)= 1.02072629795774139e-03 > 9.99999999999999955e-08
STRUCTURE: structure   dispy    at node  96	 is WRONG --> actresult= 2.83621339654209270e-01, givenresult= 2.84707177730445016e-01, abs(diff)= 1.08583807623574558e-03 > 9.99999999999999955e-08
STRUCTURE: structure   dispz    at node  96	 is WRONG --> actresult= 2.35909515950326054e-01, givenresult= 2.36765386647539267e-01, abs(diff)= 8.55870697213212406e-04 > 9.99999999999999955e-08
STRUCTURE: structure   velx     at node  96	 is WRONG --> actresult= 4.74431012143168118e+01, givenresult= 5.08800443477653062e+01, abs(diff)= 3.43694313344849434e+00 > 1.00000000000000005e-04
STRUCTURE: structure   vely     at node  96	 is WRONG --> actresult= 4.74231668368538521e+01, givenresult= 4.99727206533389108e+01, abs(diff)= 2.54955381648505863e+00 > 1.00000000000000005e-04
STRUCTURE: structure   velz     at node  96	 is WRONG --> actresult= 2.36757291008216271e+01, givenresult= 2.39628837868776827e+01, abs(diff)= 2.87154686056055652e-01 > 1.00000000000000005e-04
```
The old results are correct since they are in agreement with those obtained with no restart used at all. However, I still think that doing `post_setup()` after initialization of displacements, velocities and accelerations makes much more sense. At least, this is the pattern for the main integration schemes:
https://github.com/4C-multiphysics/4C/blob/185ceb38473d11086ae8a40455fafbdff1ad920e/src/structure_new/src/4C_structure_new_timint_base.cpp#L868-L879
This requires further investigation. I would defer it for #715. In the meantime, I would proceed with the implementation proposed by @maxfirmbach.

mayrmt (approved):

c-p-schmidt:
@vovannikov: Your suggestion is also fine for me.

src/adapter/4C_adapter_str_timeada_joint.cpp
```diff
@@ -185,7 +184,13 @@ void Adapter::StructureTimeAdaJoint::integrate_step_auxiliary()
   const Solid::TimeInt::Base& sta = *sta_;
   const auto& gstate = sta.data_global_state();
 
+  double norm_before = 0, norm_after = 0;
```
c-p-schmidt, line 187: I guess this was meant for debugging. Can you remove it?

vovannikov: Oh, yep, fixed.

PR #712, opened 2025-04-28
Exodus: reorder indices for hex27

### LINKED ISSUE #707: [BUG] Wrong node numbering for HEX27 elements when reading exodus file
## Bug Description

Exodus meshes can be read by 4C directly. This is done in the input file section `STRUCTURE GEOMETRY`. However, the second order cubes (4C designation HEX27) are stored in a different order in exodus II file format, namely the face center nodes and the volume center node. 

Exodus order is shown here:
[Exodus Manual](https://sandialabs.github.io/seacas-docs/exodusII-new.pdf), page 31

The order in 4C is shown here:
[4C Documentation](https://4c-multiphysics.github.io/4C/documentation/celltypereference.html#hex27)

In pre_exodus, a validate_element_jacobian step is performed after mesh reading, which does the reordering
(one may notice the reordering by the log output`..Ensure positive element jacobians...Successfully rewinded 1 elements. For details see *.err file`.

This step is missing for the direct 4C input. Since the HEX27 element is the only second-order element containing face center nodes, it might be sufficient to simply reorder this element directly (hard coded).

## Possible Solution

I see two options:
Either one includes the validate step after the reading for each discretization (function `read_mesh_from_exodus` in `4C_io_meshreader.cpp`), or one renumbers the HEX27 elements directly while reading (and keeps skipping the validate step, which is probably okay).

## Steps to Reproduce

A simple input file setting is attached below (the yaml input file and the exodus II geometry file). The error message, which appears - at least in two configurations on my computer - is

```
Error message:
  determinant of 3x3 matrix is zero
```

## Your Environment

- **Relevant repo SHA1s:** 0d15c52fd64b83fa57f1692acee8585bbaf6380e
- **Relevant configure flags or configure script:** release version
- **Operating system and version:** Ubuntu 24.04.2 LTS
- **Compiler and TPL versions:** 13.3.0

- Blocks
- Related to
- Part of
- Composed of

sebproell:
Closes #707

I further simplified the duplication we have in our Exodus code and removed the Shape enum.

### DIFF
--- a/apps/pre_exodus/4C_pre_exodus_validate.cpp
+++ b/apps/pre_exodus/4C_pre_exodus_validate.cpp
@@ -15,6 +15,7 @@
 #include "4C_global_legacy_module.hpp"
 #include "4C_io_input_file.hpp"
 #include "4C_linalg_utils_densematrix_multiply.hpp"
+#include "4C_utils_enum.hpp"
 
 FOUR_C_NAMESPACE_OPEN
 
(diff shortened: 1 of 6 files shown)


mayrmt (dismissed):

mayrmt (approved):

ischeider (approved):
Thanks for modifying the code alongside with the twist of Hex27 elements!

PR #718, opened 2025-04-29
Put magic_enum reflection into EnumTools namespace

sebproell:
This PR properly encapsulates the magic_enum dependency in a namespace `EnumTools`. Previously, I tried to make the enum IO operators available globally but this cannot work due to the lookup rules of C++.

FYI @dragos-ana


dragos-ana (approved):
Looks good to me. Thanks for taking the time to improve this!

amgebauer (approved):

PR #724, opened 2025-04-30
List all our CMake variables in the documentation

sebproell:
I've been meaning to do that for quite some time: add a list of all variables that CMake understands to the documentation. Looks like this:

image

The better configuration necessary for #664 will only make sense if people know which variables exist. The table is not perfect yet, as seen for the entries with `UNINITIALIZED` type. We can add a nicer entry for these in the next steps.


davidrudlstorfer (approved):

georghammerl:
@sebproell Thank you for the list. I wonder why the `TEST_TIMEOUT_SCALE` is a string? Additionally, the default value (I assume for boolean -> false) would be nice to mention.

georghammerl (approved):

PR #726, opened 2025-04-30
Simplify porofluid time integration classes

bwirthl:
This PR merges the two time integration classes into one because only one time integration method is available (one-step-theta). However, the implementation was split into two classes.


src/adapter/4C_adapter_porofluid_pressure_based_wrapper.hpp
```diff
@@ -24,7 +24,7 @@ namespace Core::FE
 
 namespace PoroPressureBased
 {
-  class TimIntImpl;
+  class Porofluid;
 }
```
sebproell, line 28: This seems to be the central class for the solution procedure so why not `TimeIntegrator`, `Operator`, `Algorithm` etc. I don't really understand the name `PoroPressureBased::Porofluid`. The class is too large to name it so your name may be as good or bad as my suggestions but I stumbled over this.

sebproell (dismissed):

src/porofluid_pressure_based/4C_porofluid_pressure_based.hpp
```diff
@@ -93,7 +93,7 @@ namespace PoroPressureBased
     /*--- set, prepare, and predict ------------------------------------------*/
 
     //! add global state vectors specific for time-integration scheme
-    virtual void add_time_integration_specific_vectors() = 0;
+    void compute_intermediate_values() { return; };
```
c-p-schmidt, line 96: You probably don't need this method if it does nothing.
```suggestion
```

c-p-schmidt (approved):

src/porofluid_pressure_based/4C_porofluid_pressure_based.hpp
bwirthl: Thanks for catching this! Very true.

c-p-schmidt (approved):

PR #729, opened 2025-05-01
Remove INVALID_TEMPLATE_QUALIFIER

sebproell:
We got the macro `INVALID_TEMPLATE_QUALIFIER` from Teuchos, but it expands to nothing. It probably expanded to `template` at some point. I scraped Trilinos' history because I was curious. This diff https://github.com/trilinos/Trilinos/commit/e84d37591efa83266c521a9e7cac4c38fb4a4985#diff-4512baff44d86eb6c683befac0f69efc98d906b6f47c762ec0e5a0a8af4e59c0R3-R12 suggests that there were once buggy compilers.

### DIFF
--- a/src/adapter/4C_adapter_str_structure_new.cpp
+++ b/src/adapter/4C_adapter_str_structure_new.cpp
@@ -433,16 +433,15 @@ void Adapter::StructureBaseAlgorithmNew::set_model_types(
     case Core::ProblemType::ssi:
     case Core::ProblemType::ssti:
     {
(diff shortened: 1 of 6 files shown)


mayrmt (approved):
@sebproell Should this also be addressed upstream in Trilinos?

sebproell:
> Should this also be addressed upstream in Trilinos?

Probably. My quick Github search did not reveal any way how this macro can become active.

PR #732, opened 2025-05-02
Make the developer docs nicer

sebproell:
This is a mixed bag of doc improvements:

- Allow table content to wrap: This looks much better than the scrolling table, IMO.
- Remove the "Additional information" section in developer docs and pull the content up one level.
- I experienced problems with inconsistent docs when rebuilding locally. Thus, we now wipe the output folder before building. I didn't notice a difference in build time.


davidrudlstorfer (approved):
Nice!

c-p-schmidt (approved):

PR #733, opened 2025-05-02
Make internal node numbering consistent with input

### LINKED ISSUE #731: [BUG] Does the exodus read feature change the results?
## Bug Description: Results for at least one example depend on the geometry input (directly in .yaml or via exodus read) 

During my update of the solid tutorial (see PR #694) I first noticed that reading the HEX27 elements needs special treatment, which in the meantime has been implemented by @sebproell in PR #712 (thanks for that!). 

Now it turns out that the results are different depending on the kind of geometry input.

## Expectations

Of course the results should be the same no matter whether the exodus input is read by `pre_exodus` or 4C itself. At least on my computer and on our cluster it's not.

## Steps to Reproduce

You may verify this behavior by the following procedure:

1. generate the 4C input file of the solid tutorial, e.g.,  by running `ctest -FC cleanup -R tutorial_solid`
   (File `framework_test_output/tutorial_solid/xxx.4C.yaml` is being created), the test should run successfully. 
2. Replace all lines from `DNODE-NODE TOPOLOGY: []` till the end by:
```
STRUCTURE GEOMETRY:
  FILE: "tutorial_solid.e"
  ELEMENT_BLOCKS:
    - ID: 1
      ELEMENT_NAME: SOLID
```
3. Copy the exodus file tutorial_solid.e from the `tests/framework_test/tutorial_solid.e` to the directory, where the test ran, or replace the FILE option inside `STRUCTURE GEOMETRY` with the full path name.
4. Add the option `ENTITY_TYPE: node_set_id` to each boundary condition entry (4 times in `DESIGN SURF DIRICH CONDITIONS`). Otherwise an error will be thrown immediately (actually I don't know why this is necessary depending on the geometry read option).
5. Save the yaml input file.
6. Run the problem again (not the ctest!).

An error is thrown, because the results are different:
```
STRUCTURE: structure   dispy   (elongation) at node 2477         is WRONG --> actresult= 4.56117520898904338e-01, givenresult= 4.56123591074363621e-01, abs(diff)= 6.07017545928290048e-06 > 9.99999999999999980e-13
STRUCTURE: structure   dispx   (necking) at node   6     is WRONG --> actresult=-4.11160719183618420e-02, givenresult=-4.25664885567924312e-02, abs(diff)= 1.45041663843058916e-03 > 9.99999999999999980e-13
```

## Your Environment

- **Relevant repo SHA1s:*d476353a392d947cf210d20472bfbf5524ff8146*
- **Relevant configure flags or configure script:*RELEASE*
- **Operating system and version:*Ubuntu 24.04 LTS*
- **Compiler and TPL versions:*gcc 13.3*

Same behavior comes up on the cluster (same repo SHA1, CentOS 7.9, gcc 13.2)

Blocks #694

sebproell:
This PR makes the internal node numbers equal to the node numbers specified in the dat-style or Exodus input.

Closes #731

### DIFF
--- a/src/ale/4C_ale_resulttest.cpp
+++ b/src/ale/4C_ale_resulttest.cpp
@@ -32,7 +32,6 @@ void ALE::AleResultTest::test_node(
   if (dis != aledis_->name()) return;
 
   int node = container.get<int>("NODE");
-  node -= 1;
 
   int havenode(aledis_->have_global_node(node));
   int isnodeofanybody(0);
--- a/src/art_net/4C_art_net_artery_resulttest.cpp
+++ b/src/art_net/4C_art_net_artery_resulttest.cpp
@@ -46,7 +46,6 @@ void Arteries::ArteryResultTest::test_node(
   if (dis != dis_->name()) return;
 
   int node = container.get<int>("NODE");
-  node -= 1;
 
   int havenode(dis_->have_global_node(node));
   int isnodeofanybody(0);
--- a/src/core/fem/src/general/element/4C_fem_general_element.cpp
+++ b/src/core/fem/src/general/element/4C_fem_general_element.cpp
@@ -198,11 +198,10 @@ void Core::Elements::Element::set_node_ids(const int nnode, const int* nodes)
 
 /*----------------------------------------------------------------------*/
 /*----------------------------------------------------------------------*/
-void Core::Elements::Element::set_node_ids_one_based_index(
+void Core::Elements::Element::set_node_ids(
     const std::string& distype, const Core::IO::InputParameterContainer& container)
 {
   nodeid_ = container.get<std::vector<int>>(distype);
-  for (int& i : nodeid_) i -= 1;
   node_.resize(0);
 }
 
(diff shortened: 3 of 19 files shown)


ischeider (approved):
Thanks for the quick fix! Also the whole -"1 procedure" seems more consistent across the problem types.

bgoderbauer (approved):

sebproell:
This PR reveals loads of interesting things. I suspect we depend on some implicit numbering a lot more than we would like. Some observations:

- Some contact tests have the correct results but iteration numbers go up: https://github.com/4C-multiphysics/4C/actions/runs/14796604344/job/41548992303#step:7:483
- Scatra multi-scale somehow still depends on zero-based nodes
- Cardiovascular has very small deviations.
- Specifying  a node to test results on for the `DOMAIN` input is questionable since the node numbering is a pure implementation detail. Same holds for `ParticleWall`. We should test results at geometric coordinates instead.

sebproell:
We should not depend on any numbering. However, there seem to be some very nasty dependencies on the zero-based numbering, and I am not sure I will be able to cleanly resolve these here. @maxfirmbach can you point me to the code where you think we do something questionable based on node numbering.

Overall, getting this right seems even more important for the future robustness of the code base.

mayrmt:
@sebproell @maxfirmbach there is currently work in progress to replace the degree of freedom dependent aggregation scheme by a node based aggregation scheme. This should not depend on any conversions between dofs and nodes. 

I expect this to be merged into Trilinos within the next couple of weeks

mayrmt:
@sebproell @maxfirmbach The necessary changes to Trilinos have been merged into Trilinos in the mean time. Necessary changes to 4C are work in progress, cf. #878, but require a Trilinos update in 4C in order to be mergable.

PR #735, opened 2025-05-04
Use the same internal numbering for direct Exodus input

### LINKED ISSUE #731: [BUG] Does the exodus read feature change the results?

sebproell:
This is a kind of alternative to #733. Closes #731

### DIFF
--- a/src/core/io/src/4C_io_exodus.cpp
+++ b/src/core/io/src/4C_io_exodus.cpp
@@ -107,7 +107,8 @@ namespace
 /*----------------------------------------------------------------------*
  |  ctor (public)                                              maf 12/07|
  *----------------------------------------------------------------------*/
-Core::IO::Exodus::Mesh::Mesh(const std::string exofilename)
+Core::IO::Exodus::Mesh::Mesh(std::filesystem::path exodus_file, MeshParameters mesh_data)
+    : mesh_data_(mesh_data)
 {
   int error;
   int CPU_word_size, IO_word_size;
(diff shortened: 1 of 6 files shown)


ischeider (approved):
You didn't change the FOUR_C_THROW calls after an if-clause into a FOUR_C_ASSERT. Is there any policy which one to use generally?

sebproell:
@ischeider I don't think there is a policy. Do you want to add some guidance on that? This could go into the coding guidelines. I use both styles, depending on the context. Sometimes, I want to construct a complex error message. In this case, the manual if is useful.

PR #736, opened 2025-05-04
Use std::source_location in exception macros

sebproell:


sebproell:
It is absurd that this change makes the xfsci test time out. Results are still correct. IIRC, a similar side effect also happened due to another change to the exceptions for the xfluid tests. Are we throwing a lot of exceptions in this framework for control flow?

sebproell:
I moved the evaluation of the formatting string such that it is only evaluated when the assertion is violated (as was the case before). Now the tests pass.

amgebauer (approved):

c-p-schmidt (approved):

PR #739, opened 2025-05-05
Clang-tidy: check for uninitialized members

sebproell:
We had UB from uninitialized members in the past. Let's see what this check finds.


sebproell:
This has too many false positives, IMO. Closing.

PR #745, opened 2025-05-07
Remove unused code in Condition utils

sebproell:
Removed unused code (see coverage report) originating from `4C_fem_condition_utils.hpp/.cpp`. Sometimes the removal spread into XFEM and some other classes.


davidrudlstorfer (approved):

src/scatra/4C_scatra_utils.cpp
```diff
@@ -18,6 +18,52 @@
 
 FOUR_C_NAMESPACE_OPEN
 
+namespace
+{
+  void assert_same_nodes(const Core::Conditions::Condition* const condition1,
```
sebproell, line 23: @c-p-schmidt I think this function is too special to warrant a place in core, so I moved it to this place. Actually, it isn't related to conditions that much; it mostly checks whether to vectors contain the same elements.

c-p-schmidt (approved):

PR #753, opened 2025-05-08
Refactor the preconditioner reuse logic

maxfirmbach:
## Description and Context

As the title states. Removing the `create` boolean from the preconditioner setup call and letting the iterative solver directly decide.


sebproell (approved):
Nice, I like to move `if`s out of functions, and this PR nicely illustrates why :)

PR #758, opened 2025-05-09
Minor improvement to LinAlg::View

sebproell:
- Use `unique_ptr`
- Add an example to docs
- Remove the `shared_ptr` accessor. No need to propagate these further than necessary. We can use `shared_ptr_from_ref` instead.


sebproell:
Forgot one `shared_ptr`; now ready.

davidrudlstorfer (approved):

amgebauer (approved):

PR #759, opened 2025-05-09
[LinAlg] Add FE_Graph capability to `Core::LinAlg::Graph`

### LINKED ISSUE #404: [linalg] Make LinAlg::Graph use Epetra_FECrsGraph
## Description

As outlined in #403 we are wrapping the `Epetra_CrsGraph` by our own `Core::LinAlg::Graph`, but currently omit an implementation of `Epetra_FECrsGraph` in the wrapper class.
I would propose to make `Epetra_FECrsGraph` the standard case, as there are some parts in the code that make use of its off-processor assembly capabilities (which are really nice to have). Performance wise there is not a lot of difference.

From an implementation point of view I only see one issue. If there's a `Trilinos` function returning an `Epetra_CrsGraph` it seems difficult to turn this into an `Epetra_FECrsGraph` to make it compatible to `Core::LinAlg::Vector` (@sebproell any idea how to handle this case?).

- [ ] Replace the `Epetra_CrsGraph` private member of `Core::LinAlg::Graph` with `Epetra_FECrsGraph` and modify the overwritten method calls
- [ ] Replace all occurrences of `Epetra_FECrsGraph` in the `4C` code base with the new `Core::LinAlg::Graph`

## Interested Parties

 @knarfnitram

maxfirmbach:
## Description and Context

This PR proposes a way to include the `Epetra_FECrsGraph` into our `Core::LinAlg::Graph` wrapper. For clarification, I tried to replace the internal `Epetra_CrsGraph` member with `Epetra_FECrsGraph`, which did not work out. Thus, I replicated what is done in `SparseMatrix`, which is not the prettiest thing, but it works and is consistent to what is already done.

I would be happy about some other opinions though. 

* Closes #404

### DIFF
--- a/src/core/fem/src/discretization/4C_fem_discretization_partition.cpp
+++ b/src/core/fem/src/discretization/4C_fem_discretization_partition.cpp
@@ -11,8 +11,6 @@
 #include "4C_linalg_utils_densematrix_communication.hpp"
 #include "4C_utils_exceptions.hpp"
 
-#include <Epetra_FECrsGraph.h>
-
 FOUR_C_NAMESPACE_OPEN
 
 /*----------------------------------------------------------------------*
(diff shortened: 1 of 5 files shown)


sebproell (approved):
I think this is OK as a first implementation :+1:  Since you mentioned the SparseMatrix: there it is different since our "wrapper" inherits, and is not giving us control over the interface.


The more of Epetra we get under our control, the more we have the ability to abstract away details. Epetra has a few bugs/surprises in terms of const-correctness, slicing of data on copy, etc. The less we expose, the better. For instance, we could think about only ever using the `FE` (i.e., off-processor assembly) types. But, as you probably realized, we need to get all the types under our control first. Probably we should switch to Tpetra first anyway, since this package probably has other bugs/surprises we need to deal with :laughing:

sebproell:
> I was thinking of giving the graph type as template parameter, not sure if this is a good idea though. Giving it as function input is a bit weird.

My rule of thumb is: don't use a template if you don't know why. If you were to use one here, the two variants can no longer bind to a `Graph&` reference (unless you template that as well). Let's just go with what you have.

knarfnitram (approved):

PR #762, opened 2025-05-12
Fix ASAN pipeline

### LINKED ISSUE #760: [BUG] Failng ASAN test with SuperLU_dist
NIghtly ASAN tests `beam3r_herm2line3_elastoplastic_axialNeum_hard.dat-p2` fails somewhere in SuperLU_dist See https://github.com/4C-multiphysics/4C/actions/runs/14949609770/job/41999635796#step:7:475

Possible caused by #663

maxfirmbach:
## Description and Context

Problematic test now uses `Umfpack` instead of `SuperLU_DIST`. This should avoid the memory problem.

Closes #760


sebproell (approved):

amgebauer:
@maxfirmbach does it make sense to open an issue? To me it seems weird that we (sometimes) run into a segmentation fault if we use superlu for this test.

PR #765, opened 2025-05-12
Remove unnecessary particle output

ppraegla:
Follow up of cce0df77b512952df408dead231c82ecc95bc8af. Before this commit, the timing of reading a block of particles was printed. After the refactoring the printing was done after each read particle.


sebproell (approved):
Sorry for that, and thanks for cleaning it up!

davidrudlstorfer (approved):

PR #770, opened 2025-05-12
Implement dot product between rank 1 tensors

amgebauer:
This combination is missing in the current implementation.

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_tensor.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_tensor.hpp
@@ -460,6 +460,22 @@ namespace Core::LinAlg
   constexpr auto transpose(const Rank2TensorConcept auto& A);
 
   /*!
+   * @brief Computes the dot product of two rank-1 tensors
+   *
+   * This function performs the inner product between two vectors a * b. Both sizes need to match.
+   *
+   * @tparam TensorLeft The type of the rank-1 tensor (vector).
+   * @tparam TensorRight The type of the rank-1 tensor (vector).
+   * @param a The rank-1 tensor (vector).
+   * @param b The rank-1 tensor (vector).
+   * @return scalar resulting from the dot product.
+   */
+  template <typename TensorLeft, typename TensorRight>
+    requires(Rank1TensorConcept<TensorLeft> && Rank1TensorConcept<TensorRight> &&
+             TensorLeft::template extent<0>() == TensorRight::template extent<0>())
+  constexpr auto dot(const TensorLeft& a, const TensorRight& b);
+
+  /*!
    * @brief Computes the dot product of a rank-2 tensor and a rank-1 tensor.
    *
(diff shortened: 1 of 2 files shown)


src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -459,6 +459,22 @@ namespace Core::LinAlg
...
   constexpr auto transpose(const Rank2TensorConcept auto& A);
 
+  /*!
+   * @brief Computes the dot product of two rank-1 tensors
+   *
+   * This function performs the scalar product between two vector a * b. Both sizes need to match.
+   *
+   * @tparam TensorLeft The type of the rank-1 tensor (vector).
+   * @tparam TensorRight The type of the rank-1 tensor (vector).
+   * @param a The rank-1 tensor (matrix).
```
mayrmt, line 469: The comment says "matrix", but this is also a "vector", right?

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -459,6 +459,22 @@ namespace Core::LinAlg
    */
   constexpr auto transpose(const Rank2TensorConcept auto& A);
 
+  /*!
+   * @brief Computes the dot product of two rank-1 tensors
+   *
+   * This function performs the scalar product between two vector a * b. Both sizes need to match.
```
mayrmt, line 465: Would "inner product" be a more accurate description?

mayrmt (commented):
What happens if the data type of both vectors is different, e.g. double and int? Will everything fall in place automatically or dues this need its own test?

amgebauer:
> What happens if the data type of both vectors is different, e.g. double and int? Will everything fall in place automatically or dues this need its own test?

All data types that are compatible should work, but we don't have tests for this yet. I will add these in another PR since they are also needed for the existing operations.

mayrmt (dismissed):

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -459,6 +459,22 @@ namespace Core::LinAlg
...
+  /*!
+   * @brief Computes the dot product of two rank-1 tensors
+   *
+   * This function performs the inner product between two vectors a * b. Both sizes need to match.
+   *
+   * @tparam TensorLeft The type of the rank-1 tensor (vector).
+   * @tparam TensorRight The type of the rank-1 tensor (vector).
+   * @param a The rank-1 tensor (vector).
+   * @param b The rank-1 tensor (vector).
+   * @return double scalar resulting from the dot product.
```
sebproell, line 471:
```suggestion
   * @return Scalar value resulting from the dot product.
```

It's not going to be a double necessarily, right?

sebproell (approved):

sebproell (approved):

mayrmt (approved):

PR #775, opened 2025-05-14
Redistribute discret for point coupling conditions

georghammerl:
## Description and Context
Point coupling conditions rely on master and slave nodes to be owned by the same processor. This requirement comes from the current procedure how assign degrees of freedom works. To enable parallel runs with point coupling conditions, a redistribution is performed to ensure proper parallel distribution of conditioned nodes.

## Interested parties
@ischeider


src/core/fem/src/condition/4C_fem_condition_multipointcoupling.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  nodesonthisproc.clear();
+  std::vector<int> newnodesonthisproc(row_node_set.begin(), row_node_set.end());
+
+  //--------------------------------------------------
+  // build noderowmap for new distribution of nodes
+  // the constructed rowmap
+  std::shared_ptr<Core::LinAlg::Map> newrownodemap =
```
sebproell, line 203: This shared_ptr seems unnecessary. Same for `newcolnodemap`. Also, what do you think about naming this `new_row_node_map` etc?

src/core/fem/src/condition/4C_fem_condition_multipointcoupling.hpp
```diff
@@ -0,0 +1,78 @@
...
+{
+  class Discretization;
+}  // namespace Core::FE
+
+namespace Core::Conditions
+{
+  /*!
+  \brief  update node row map for point coupling conditions
+  master and connected slave nodes must be owned by the same processor
+  */
```
sebproell, line 34: This does not help me understand what the class does.

src/core/fem/src/condition/4C_fem_condition_multipointcoupling.hpp
```diff
@@ -0,0 +1,78 @@
...
+  class MultipointcouplingConditions
+  {
+   public:
+    //! constructor which checks the discretization for point coupling conditions
+    MultipointcouplingConditions(
+        std::shared_ptr<Core::FE::Discretization> dis, bool verbose = true);
+
+    //! Destructor
+    virtual ~MultipointcouplingConditions() = default;
```
sebproell, line 43: Why?

```suggestion
```

src/core/fem/src/condition/4C_fem_condition_multipointcoupling.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+    //! are there pointcoupling conditions in use
+    bool has_pcc() const { return has_pointcoupling_conditions_; }
+
+    //! make sure that proper node distribution is achieved for point coupling conditions
+    void redistribute_for_pointcoupling_conditions();
```
sebproell, line 49: This is called only once, directly after the constructor. So it can at least be private, but I would prefer to remove this completely form the header. If you want an implementation heleper, that is fine, but I would put it in the cpp file.

src/core/fem/src/condition/4C_fem_condition_multipointcoupling.hpp
```diff
@@ -0,0 +1,78 @@
...
+   public:
+    //! constructor which checks the discretization for point coupling conditions
+    MultipointcouplingConditions(
+        std::shared_ptr<Core::FE::Discretization> dis, bool verbose = true);
+
+    //! Destructor
+    virtual ~MultipointcouplingConditions() = default;
+
+    //! are there pointcoupling conditions in use
+    bool has_pcc() const { return has_pointcoupling_conditions_; }
```
sebproell, line 46: Unused.

src/core/fem/src/condition/4C_fem_condition_multipointcoupling.cpp
```diff
@@ -0,0 +1,243 @@
...
+    int myn = static_cast<int>(row_node_set.size());
+    int gn = 0;
+
+    Core::Communication::sum_all(&myn, &gn, 1, discret_->get_comm());
+    if (gn != discret_->num_global_nodes())
+    {
+      FOUR_C_THROW(
+          "Unmatching numbers of nodes before and after call Redistribution. Nodemap "
+          "constructor will crash.\n");
+    }
```
sebproell, line 194: Is this more of an `FOUR_C_ASSERT`, i.e., an implementation check?

sebproell (changes requested):
After reading through everything, I think what's hiding behind the class `MultipointcouplingConditions` is really a function like `redistribute_for_point_coupling_conditions`. I would prefer if you just write that function by fusing the constructor arguments with the function you already wrote with the same name.

ischeider (dismissed):
I tested the code with an example of 45e3 tet10 elements and 69e3 nodes on 1, 2, 4, and 8 processes, and the speed up using Belos with a GMRES solver was ( compared to simulation with 1 cpu on the main branch):

- 1 cpu: 0.99
- 2 cpus: 1.90
- 4 cpus: 2.48
- 8 cpus: 2.66

src/core/fem/src/condition/4C_fem_condition_multipointcoupling.hpp
georghammerl: Removed

src/core/fem/src/condition/4C_fem_condition_multipointcoupling.cpp
georghammerl: I'd prefer to have it in a all runs actually.

src/core/fem/src/condition/4C_fem_condition_multipointcoupling.hpp
georghammerl: Removed

src/adapter/4C_adapter_str_structure_new.cpp
```diff
@@ -144,52 +142,88 @@ void Adapter::StructureBaseAlgorithmNew::setup_tim_int()
...
-    };
+      const std::vector<int>* conditioned_nodes = pcc->get_nodes();
+      for (const int node_id : *conditioned_nodes)
+      {
+        if (actdis_->have_global_node(node_id))
+        {
+          const Core::Nodes::Node* node = actdis_->g_node(node_id);
+          const Core::Elements::Element* first_element = node->elements()[0];
+
+          if (!(dynamic_cast<const Discret::Elements::Beam3Base*>(first_element)) &&
```
isteinbrecher, line 158: Why does the proposed approach not work for beams?

src/adapter/4C_adapter_str_structure_new.cpp
```diff
@@ -144,52 +142,88 @@ void Adapter::StructureBaseAlgorithmNew::setup_tim_int()
...
+              {currpos[0] + radius, currpos[1] + radius, currpos[2] + radius}};
+        }
+        else
+          return Core::Binstrategy::DefaultRelevantPoints{}(discret, ele, disnp);
+      };
 
-    Core::Rebalance::rebalance_discretizations_by_binning(binning_params,
-        Global::Problem::instance()->output_control_file(), actdis_vec, correct_node,
-        determine_relevant_points, true);
+      Core::Rebalance::rebalance_discretizations_by_binning(binning_params,
```
isteinbrecher, line 223: I am pretty sure we still want to do this for point couplings. While not optimal, the binning re-distribution helps to put spatially close elements on the same processor. This does not ensure that nodes from point couplings are on the same processor (this is the reason for this PR) but it usually works out quite OK and results in a better partitioning than the graph based partitioning (where point couplings are not taken into account). Your `redistribute_for_point_coupling_conditions` function should be applied after the discretisation is distributed with binning.

src/core/fem/src/condition/4C_fem_condition_point_coupling_redistribution.cpp
```diff
@@ -0,0 +1,255 @@
...
+    if (gn != discret.num_global_nodes())
+    {
+      FOUR_C_THROW(
+          "Unmatching numbers of nodes before and after call Redistribution. Nodemap "
+          "constructor will crash.\n");
+    }
+  }
+
+  const std::vector<int> new_row_nodes(row_node_set.begin(), row_node_set.end());
+  row_node_set.clear();
```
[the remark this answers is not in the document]
georghammerl: True, this does not have effect. If I would like to free memory (as these sets may be large): What about

src/adapter/4C_adapter_str_structure_new.cpp
georghammerl: Each beam has distint nodes (no common nodes of different beam elements). Hence, the graph does not know of beam elements which are close to each other (at least not with the graph how I use it). So, beams demand for geometric close (graph) analysis which is done with the binning approach (at the moment).

src/core/fem/src/condition/4C_fem_condition_point_coupling_redistribution.cpp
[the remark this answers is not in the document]
sebproell: This set goes out of scope 40 lines further down, and its memory is freed. Am I missing something?

src/core/fem/src/condition/4C_fem_condition_point_coupling_redistribution.hpp
```diff
@@ -0,0 +1,37 @@
...
+}  // namespace Core::FE
+
+namespace Core::Conditions
+{
+  /*!
+  \brief  The procedure to assign degrees of freedom relies for point coupling conditions on a
+  layout in which the master and connected slave nodes must be owned by the same processor. This
+  function redistributes the row nodes accordingly.
+  */
+  //! make sure that proper node distribution is achieved for point coupling conditions
```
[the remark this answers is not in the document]
georghammerl: Removed

src/core/fem/src/condition/4C_fem_condition_point_coupling_redistribution.hpp
```diff
@@ -0,0 +1,37 @@
...
+  void redistribute_for_point_coupling_conditions(
+      std::shared_ptr<Core::FE::Discretization> discret, bool verbose);
```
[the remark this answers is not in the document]
georghammerl: Changed -> Thanks

src/core/fem/src/condition/4C_fem_condition_point_coupling_redistribution.cpp
```diff
@@ -0,0 +1,255 @@
...
+      {.assign_degrees_of_freedom = true,
+          .init_elements = true,
+          .do_boundary_conditions = true,
+          .do_extended_ghosting = true});
+
+  if (myrank == 0 && verbose)
+    std::cout << "\nparallel redistributed discretization due to point coupling condition(s)"
+              << std::endl;
+
+  Core::Rebalance::Utils::print_parallel_distribution(discret);
```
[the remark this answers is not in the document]
georghammerl: Removed (unnecessary "verbose" flag)

src/adapter/4C_adapter_str_structure_new.cpp
georghammerl: The changes in this PR only come into action if conditioned nodes on solid elements are in the discretization. This has not yet worked so far in parallel. Hence, the changes in this PR cannot be seen as an optional appendix.

Assuming binning is applied first for beam/solid problems (with conditioned solid nodes), the changes in this PR would afterwards destroy the "bin-achieved" distribution. Therefore, this should not be done. The other way round, applying first the redistribution of this PR on beam/solid problems would be destroyed by a following binning redistribution. Hence, I am not aware of a practicable solution how these two mechanism can be combined nicely (at all).

isteinbrecher: > This has not yet worked so far in parallel

I am pretty sure that this already works in parallel. At the moment, the redistribution is done by a binning approach, which usually puts close nodes on the same processor (probably the monolithic approach introduced by @maxfirmbach would be even better). So at the moment this fails, if after this "geometric close" partitioning, the coupling nodes are not on the same rank. In practice this usually worked out quite well. If I remember our discussion a couple of months ago, you have conditions where the coupled nodes are not at the same spatial position. This will likely fail with the binning approach. But in general, for coupled nodes at the same position (which in my experience is a very common case) the binning partitioning is usually better than the graph partitioning. So I am not sure if we should simply switch to your proposed approach for all problems. Can we make this optional? Or even better, one could select the partitioning algorithm (binning, graph based, monolithic in the future, ...) and then your redistribution to put coupled nodes that are not on the same processor is run after that (I don't see a problem why this should not work).

> the changes in this PR would afterwards destroy the "bin-achieved" distribution.

Yes and no, one would simply move the coupled nodes to the "main" coupling nodes if they are not on the same rank. If the binning partitioning already provides this, then your implementation would not change the parallel distribution, correct?

georghammerl: The current implementation (with binning) exclusively works if nodes are at exactly the same position that need to be coupled. If they are just close, it may work out or may not work out if I get the implementation correctly. The application this PR targets to is on spatially distributed nodes which are conditioned (and thus share common dofs). This does not work with binning (or any other purely geometric partitioning). The proper connectivity is achieved with the graph as it is done in this PR.

src/adapter/4C_adapter_str_structure_new.cpp
georghammerl: > I am not sure I fully understand your point. Nodes that are connecting multiple beams along are connected to these elements the same way solid elements are.

The test cases I found, only couple nodes from different beams. The coupled nodes are exactly at the same position (=touching beams). Due to the fact that the coupled nodes are exactly at the same position, the binning approach pulls them onto the same rank (and assign degrees of freedom works). If they would be just close, this is not guaranteed.

> Besides that, with the current implementation, you only check if there is a single solid element, so if there is a mixed mesh, you will still end up with beam elements in your proposed approach (if I have read the code correctly).

This point is valid, actually a check is necessary that catches the case with constraint beam nodes and spatially distributed constraint solid nodes. This does not work with either of the implementations correctly. Maybe we can discuss offline how this could go together. I do not see a solution for this at the moment.

sebproell (dismissed):
I am approving for the code design aspects.

ischeider (dismissed):

src/adapter/4C_adapter_str_structure_new.cpp
georghammerl: @isteinbrecher As discussed, I added the additional redistribution after the binning. This leads to trouble because at the end of the binning redistribution, `fillcomplete` is called which tries to assign dofs. This is not (yet) possible for distributed point couplings in this moment as the proper distribution is achieved afterwards. It is quite tricky.

georghammerl: The additional redistribution can be seen as a repair step if binning was not successful. @isteinbrecher Please have a look.

src/adapter/4C_adapter_str_structure_new.cpp
```diff
@@ -187,9 +185,27 @@ void Adapter::StructureBaseAlgorithmNew::setup_tim_int()
...
+    {
+      Core::Rebalance::rebalance_discretizations_by_binning(binning_params,
+          Global::Problem::instance()->output_control_file(), actdis_vec, correct_node,
+          determine_relevant_points, true);
+    }
+    catch (const Core::Exception& e)
+    {
+      // this can happen due to improper node distribution in assign degrees of freedom
+      node_distribution_repair_necessary = 1;
+    }
```
sebproell: Hm, 4C isn't really exception-safe, so doing try-catch on a large algorithmic component seems to beg for trouble. Can you guarantee that the objects are in a useable/defined state if an exception is thrown somewhere in the middle? If that was always in this PR I have missed it, sorry for the late notice in this case.

src/adapter/4C_adapter_str_structure_new.cpp
isteinbrecher: I like this approach, because it extends the existing implementation rather than changing it. But I also agree with @sebproell regarding the exception. What exception is thrown?

georghammerl: I switched to a newly introduced custom exception: `NodalDistributionException`
This exception is thrown at exactly one place and catched in another place. All other exceptions are passing and thus inconsistent states should (hopefully) be avoided in the catch case.

src/adapter/4C_adapter_str_structure_new.cpp
georghammerl: Resolve here due to duplicate from above.

src/core/fem/src/dofset/4C_fem_dofset.cpp
```diff
@@ -292,8 +292,8 @@ int Core::DOFSets::DofSet::assign_degrees_of_freedom(
         }
       }
       if (!allononeproc)
-        FOUR_C_THROW(
-            "ERROR: Nodes in point coupling condition must all be on same processor (for now).");
+        throw NodalDistributionException(
+            "ERROR: Nodes in point coupling condition must all be on same processor.");
```
sebproell: If I am not missing something, you will only throw this exception on ranks where you have offending nodes. Are you missing a `broadcast` before doing so?

georghammerl: `FOUR_C_THROW` does also not account for this aspect. The catching side is responsible for taking care of this communication. Please check in [4C_adapter_str_structure_new.cpp](https://github.com/4C-multiphysics/4C/pull/775/files#diff-6f347d05bce3ea89aa766ba061b096cf4e0c3ca39f3286003da70ab96877426a) the implementation after the catch block.

sebproell: Sorry, I still don't get it. In line 360 of this file you do an `import` with the `dof_row_map`, so you do MPI communication. Won't this hang if one of the ranks never reaches this line?

georghammerl: Now, I got it. Will adapt. Thanks!

georghammerl: please see latest changes

georghammerl:
Let's see whether this is now the final version @sebproell @isteinbrecher ?!

sebproell (approved):

georghammerl:
I briefly squashed into one commit.

sebproell (approved):

PR #776, opened 2025-05-14
Cleanup pointers and const-ness for Condition

sebproell:
This PR escalated quite drastically in size. However, the changes are simple:

- Removes a lot of  `shared_ptr<Condition>`
- Fixes `const`-ness issues for many functions taking `Condition`

I'll point out some things in the diff.


src/core/fem/src/discretization/4C_fem_discretization.hpp
```diff
@@ -1147,46 +1147,28 @@ namespace Core::FE
...
-    returned and behaviour is therefore non-deterministic. This method should
-    therefore only be used in cases where the user is sure that name is unique.
-
-    \note Conditions attached to the discretization have to be
-          completely redundant meaning that nodal cloud in the
-          condition is the same on each processor and spans all
-          nodes that hold this condition
-
-    \param name (in): Name of condition
+        const std::string& name, std::vector<const Core::Conditions::Condition*>& out) const;
```
sebproell, line 1166: FYI, The cascade of changes was mostly triggered by this signature change. As you can see, `const`-ness was broken here before: although the function is `const`, we got a non-`const` pointer and could modify the condition, although there are operations on the `Condition` that would require an internal rebuild of the discretization.

src/cardiovascular0d/4C_cardiovascular0d.cpp
```diff
@@ -165,13 +165,16 @@ Utils::Cardiovascular0D::Cardiovascular0D(std::shared_ptr<Core::FE::Discretizati
     {
       coupcondID[i] = cardiovascular0dstructcoupcond_[i]->parameters().get<int>("coupling_id");
 
+      // TODO: This hacks the condition parameters.
+      auto& parameters = const_cast<Core::IO::InputParameterContainer&>(
+          cardiovascular0dstructcoupcond_[i]->parameters());
```
sebproell, line 170: I had to insert code like this in a dozen places or so. The `const_cast` makes it explicit that we hack things in the condition in a way that is not visible to the Discretization. This was the case before as well.

sebproell: No, it's not UB, since the thing we are const-casting is not const.

amgebauer: We cast away const from the parameters which is a direct member of the const condition. Are you sure that this is correct?

amgebauer: ah, now I understand. How often do we have these constructs?

At least I understand why we do it in the cardiovascular BC. I found them in cv0d, locsys and fluid. I already thought about reimplementing cv0d, but there are no concrete plans.

amgebauer (approved):

PR #778, opened 2025-05-15
Disable metadata generation in lnm cluster presets

maxdinkel:
## Description and Context

This pull request disables the metadata generation in the LNM cluster presets. If enabled, the automated build of 4C on a cluster node in the [Queens](https://github.com/queens-py/queens) cluster test fails due to an MPI error (see https://github.com/queens-py/queens/pull/146#issuecomment-2848493756).  Of course it is also possible to add this build option as an additional variable in the queens pipeline setup (see https://github.com/queens-py/queens/pull/146#discussion_r2088429776), but we think it would make sense to disable the metadata generation on the clusters by default (this is also the default at the IMCS cluster).

Tagging @davidrudlstorfer @sbrandstaeter @knarfnitram


davidrudlstorfer (approved):
Thanks for the change @maxdinkel !

knarfnitram (approved):

c-p-schmidt (approved):

PR #779, opened 2025-05-15
[NOX] Simplify one step solver

vovannikov:
## Description and Context
The changes made to Trilinos in https://github.com/trilinos/Trilinos/pull/14017 do not longer require some additional methods in the NOX Group to overcome the problems of the NOX SingleStep solver. These things can now be removed.

I am now also preparing a new docker image with the newer Trilinos version.

## Related Issues and Pull Requests
#359, #379

Issue #379, opened 2025-02-14
[NOX] Clean up of `NOX::Solver::SingleStep`


vovannikov:
## Description
The more I look into the code, the more I figure out that certain "non-optimal" design desicions in 4C were made due to mistakes in the design of `Trilinos` itself. And I think this is the right time to figure our how they should tackled properly.

There is `SingleStep` solver in 4C which is based on `NOX::Solver::SingleStep` from `Trilinos` and its `solve()` function looks like this:
https://github.com/4C-multiphysics/4C/blob/67d3f5310bfa26b46febd3901bc7939756e7588a/src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_singlestep.cpp#L103-L124

You may see that the `step()` function is customized with some strange calls:
```c++
  nln_group.set_is_valid_newton(true);  // to circumvent the check in ::NOX::Solver::SingleStep
  nln_group.set_is_valid_rhs(false);    // force to compute the RHS
  nln_group.reset_x();                  // to initialize the solution vector to zero
```
Note that `set_is_valid_newton()` and `set_is_valid_rhs()` are used only in the implementation of the single step solver, not anywhere else in the code, which makes me think that it was a clear hack to resolve some problem.

Indeed, this implementation is dictated by this piece of code
https://github.com/trilinos/Trilinos/blob/62ff542b83d63ad9a56b91ef7af177c485459dc2/packages/nox/src/NOX_Solver_SingleStep.C#L131-L138

Here the author optimized memory usage. Well, `const_cast` is not nice, but what makes it worse is that `getNewton()` can thrown an exception:
https://github.com/trilinos/Trilinos/blob/62ff542b83d63ad9a56b91ef7af177c485459dc2/packages/nox/src-epetra/NOX_Epetra_Group.C#L547-L555

And the only way to set flag `isValidNewton` to `true` is to call `Group::computeNewton()`, and that call is not performed in `NOX::Solver::SingleStep`. Aparently, this makes `NOX::Epetra::Group` absolutely incompatible with `NOX::Solver::SingleStep`. And what it makes the situation even worse is that this behavior is not consistent between different group implementations in `Trilinos` itself. For instance, `getNewton()` in `Thyra::Group` can not throw a similar exception because it has not any checks inside: https://github.com/trilinos/Trilinos/blob/62ff542b83d63ad9a56b91ef7af177c485459dc2/packages/nox/src-thyra/NOX_Thyra_Group.C#L750-L753.

[...]

Issue #359, opened 2025-02-11
Reducing number of NOX-based solvers


vovannikov:
## Description
This is the first part of work in relation to #300.

As discussed on the meeting, the number of solver classes is excessive and should be reduced. I would start with this part of the `NOX` related refactoring, since this seems to be the simplest task, should be good to get familiar with the other classes too. I just would like to discuss the preliminary plan and a few consequences of this decision, such that everyone is aware of them.

If no objections, I will start with these modifications. Most probably, something else appears while working on these items.

## Possible Solution and Definition of Done

Tell us what needs to happen and outline a possible solution. If possible, provide a step-by-step task list along the lines of:
- [ ] Delete the base class `Solid::Nln::SOLVER::Generic` since we agreed to rely solely on `NOX`  ([src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_generic.hpp](https://github.com/4C-multiphysics/4C/blob/be9fcea5f90f1add5f1cee799477363de8bb64bd/src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_generic.hpp))
- [ ] Delete or refactor class `Solid::Nln::SOLVER::NOX` ([src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_nox.hpp](https://github.com/4C-multiphysics/4C/blob/be9fcea5f90f1add5f1cee799477363de8bb64bd/src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_nox.hpp))
- [ ] Delete classes, which implement a specific solver via setting up certain options in the parameters list, this should be done via the corresponding XML files. These are:
`Solid::Nln::SOLVER::FullNewton` ([src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_fullnewton.hpp](https://github.com/4C-multiphysics/4C/blob/be9fcea5f90f1add5f1cee799477363de8bb64bd/src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_fullnewton.hpp))
`Solid::Nln::SOLVER::PseudoTransient` ([src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_ptc.hpp](https://github.com/4C-multiphysics/4C/blob/be9fcea5f90f1add5f1cee799477363de8bb64bd/src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_ptc.hpp))
[...]

PR #780, opened 2025-05-15
[docker] Update Trilinos config and hash for #779 (NOX SingleStep solver modifications)

vovannikov:
## Description and Context
To proceed with #779, Trilinos needs to be updated.

I have also updated `BASE_IMAGE` to Ubuntu 24.04, since this is the base image which effectively was used for the previous docker images despite the setting specified. So I think it makes sense to update it in order others would not later spend time understanding why individual packages do not compile with gcc-9.4.

#359, #379


dependencies/supported_version/Trilinos.txt
```diff
@@ -1,3 +1,4 @@
+06db4c850654feacabdaed61ee8308219266b6a5 2025.4 16.3.0
 8cdf0375d278598c12edc3fd7059cb133c6f4426 2025.3 16.2.0
 11a16481fa1af48e3cfdb0aa84c29617ce94479f 2025.2 16.1.0
 1eab15637f2998d1e86fe127b78200f2c9687cb5 2025.1 none
```
amgebauer, line 4: If I understand it correctly, the update of Trilinos is required for work in our NOX framework. If so, we need to remove all incompatible versions from this list here (once we rely on the changes in nox).

dependencies/supported_version/Trilinos.txt
```diff
@@ -1,3 +1,4 @@
+06db4c850654feacabdaed61ee8308219266b6a5 2025.4 16.3.0
```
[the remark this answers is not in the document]
amgebauer: 16.2.0 should be tied to `8cdf0375d278598c12edc3fd7059cb133c6f4426`, so I think it is actually none.

dependencies/supported_version/Trilinos.txt
vovannikov: _So far_, there is no need to do that, but along the refactoring this need may become inevitable. I would keep them for a while at least.

amgebauer (approved):

ppraegla (approved):

PR #787, opened 2025-05-16
Enable LinAlg::View for Map

sebproell:
This PR enables the `LinAlg::View` class for _any type_, regardless of whether it has native view support. I'll point out how this works in the diff.

I apply it to `LinAlg::Map` here.

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_sparsematrixbase.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_sparsematrixbase.hpp
@@ -134,46 +134,18 @@ namespace Core::LinAlg
     int global_row_index(int local_row_index) const { return sysmat_->GRID(local_row_index); }
 
     /// Returns the Epetra_Map object associated with the rows of this matrix.
-    const Core::LinAlg::Map& row_map() const
-    {
-      if (!row_map_)
-      {  // check if view is uninitialized
-        row_map_ = Core::LinAlg::Map(sysmat_->RowMap());
-      }
-      return *row_map_;
-    }
+    const Core::LinAlg::Map& row_map() const { return row_map_.sync(sysmat_->RowMap()); }
 
     /// Returns the  Epetra_Mapobject that describes the set of column-indices that appear in
     /// each processor's locally owned matrix rows.
-    const Core::LinAlg::Map& col_map() const
-    {
-      if (!column_map_)
-      {  // check if view is uninitialized
-        column_map_ = Core::LinAlg::Map(sysmat_->ColMap());
-      }
-      return *column_map_;
-    }
+    const Core::LinAlg::Map& col_map() const { return column_map_.sync(sysmat_->ColMap()); }
 
     /// Returns the Epetra_Map object associated with the domain of this matrix operator.
-    const Map& domain_map() const override
-    {
-      if (!domain_map_)
-      {  // check if view is uninitialized
-        domain_map_ = Core::LinAlg::Map(sysmat_->DomainMap());
-      }
-      return *domain_map_;
-    }
+    const Map& domain_map() const override { return domain_map_.sync(sysmat_->DomainMap()); }
 
 
     /// Returns the Epetra_Map object associated with the range of this matrix operator.
-    const Core::LinAlg::Map& range_map() const
-    {
-      if (!range_map_)
-      {  // check if view is uninitialized
-        range_map_ = Core::LinAlg::Map(sysmat_->RangeMap());
-      }
-      return *range_map_;
-    }
+    const Core::LinAlg::Map& range_map() const { return range_map_.sync(sysmat_->RangeMap()); }
 
 
     /// Returns the current UseTranspose setting.
@@ -302,10 +274,10 @@ namespace Core::LinAlg
     /// internal epetra matrix (Epetra_CrsMatrix or Epetra_FECrsMatrix)
     std::shared_ptr<Epetra_CrsMatrix> sysmat_;
 
-    mutable std::optional<Core::LinAlg::Map> range_map_;
-    mutable std::optional<Core::LinAlg::Map> row_map_;
-    mutable std::optional<Core::LinAlg::Map> domain_map_;
-    mutable std::optional<Core::LinAlg::Map> column_map_;
+    mutable View<const Map> range_map_;
+    mutable View<const Map> row_map_;
+    mutable View<const Map> domain_map_;
+    mutable View<const Map> column_map_;
   };
 
 }  // namespace Core::LinAlg
--- a/src/core/linalg/src/sparse/4C_linalg_view.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_view.hpp
@@ -23,7 +23,7 @@ namespace Core::LinAlg
    * contain a single typedef `type` for the class that is wrapped.
    */
   template <typename T>
-  struct WrapperFor
+  struct EnableViewFor
   {
     static_assert("You need to specialize this struct for the wrapper class.");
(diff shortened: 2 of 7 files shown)


src/core/linalg/src/sparse/4C_linalg_sparsematrixbase.hpp
```diff
@@ -138,7 +138,7 @@ namespace Core::LinAlg
     {
       if (!row_map_)
       {  // check if view is uninitialized
-        row_map_ = Core::LinAlg::Map(sysmat_->RowMap());
+        row_map_.emplace(sysmat_->RowMap());
```
sebproell, line 141: @maxfirmbach Note that this PR will _not_ fix the problems we get when somebody replaces maps (at least I think so). This needs to be further isolated.

src/core/linalg/src/sparse/4C_linalg_view.hpp
```diff
@@ -28,6 +29,37 @@ namespace Core::LinAlg
...
   };
 
+
+  /**
+   * A wrapper around unique_ptr, which propagates constness to the pointee.
+   * This type is useful to hold wrapped objects in various LinAlg wrappers in combination with
+   * the View mechanism.
+   */
+  template <typename T>
+  struct ConstPropagatingUniquePtr
```
vovannikov, line 39: Probably, should `Deleter` be a part of the `ConstPropagatingUniquePtr` as the case for the original `unique_ptr` itself? This would then mimic the declaration of the latter:
```c++
template<
    class T,
    class Deleter = std::default_delete<T>
>
```

src/core/linalg/src/sparse/4C_linalg_view.hpp
```diff
@@ -28,6 +29,37 @@ namespace Core::LinAlg
...
+
+    const T& operator*() const { return *obj; }
+    T& operator*() { return *obj; }
+
+    const T* operator->() const { return obj.get(); }
+    T* operator->() { return obj.get(); }
+
+    //! The actual unique_ptr. You may access it directly, but be aware that constness will not
+    //! be propagated this way.
+    std::unique_ptr<T, Deleter> obj;
```
vovannikov, line 60: Is this really required to be public? I would prevent the direct access to `obj`.

src/core/linalg/src/sparse/4C_linalg_view.hpp
```diff
@@ -105,7 +141,9 @@ namespace Core::LinAlg
 
    private:
     //! Source content wrapped in our own type.
-    std::unique_ptr<WrapperType> view_;
+    //! We need to share ownership so we can make copies of the view (which refer to the same
+    //! underlying object).
+    std::shared_ptr<WrapperType> view_;
```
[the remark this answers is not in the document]
amgebauer: But why shared pointer? Do we want the view to have the ownership? Why not `WrapperType*`?

src/core/linalg/src/sparse/4C_linalg_view.hpp
sebproell: I'm torn on this. For a general-purpose class, yes. For the case of the views, I always want a flexible deleted, so I can make the pointer non-owning for the view case.

src/core/linalg/src/sparse/4C_linalg_view.hpp
sebproell: Changed.

src/core/linalg/src/sparse/4C_linalg_view.hpp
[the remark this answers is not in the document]
sebproell: The `View` owns the WrapperType.

src/core/linalg/src/sparse/4C_linalg_view.hpp
vovannikov: You still can use an empty deleter by default, something like this:
```c++
template <typename T, typename Deleter = std::function<void(T*)>>
struct ConstPropagatingUniquePtr
```

With this syntax you can be less verbose relying extensively on CTAD, like this:
```c++
const Foo cf1;
ConstPropagatingUniquePtr cp1(&cf1, [](auto* ptr) {});
```

With `Deleter = std::function` you are less flexible with lambdas. Have a look at this code I made for myself when playing with your wrapper: https://godbolt.org/z/fc8q71rz1, try to fallback to your version and you see the problem.

By the way, can you tell me, what kind of behavior exactly you tried to achieve? To propagate const from the underlying being wrapped object or to force additional constness from the wrapper itself? To me it looks like the latter. It is still a propagation, but depending on who propagates to what. For me it was not obvious from the name, as I thought initially about the other way around propagation :)

I made that godbolt problem to understand that. So, basically, the only difference from `unique_ptr` is the case 4, am I right? And this is what you wanted to change?

sebproell: Your comments lead me to the conclusion that this type should not be named anything related to `unique_ptr` :)  I am also tripping over the slight deviations from a `std::unique_ptr`.

My requirements are the following: a kind of `Box` type (#784), that can be empty and where I can replace the contained value. The contained value needs to be a pointer or reference type, so I can also insert a viewing, non-owning pointer. The `Box` should be a thin layer around the contained type and propagate const-ness from the outside to the value. After writing all this down: it seems what I am describing here is basically a `OwnerOrViewPtr`.

src/core/linalg/src/sparse/4C_linalg_sparsematrixbase.hpp
```diff
@@ -138,9 +138,9 @@ namespace Core::LinAlg
...
       }
-      return *row_map_;
+      return row_map_->sync(sysmat_->RowMap());
```
sebproell, line 143: @knarfnitram @maxfirmbach This call should ensure that we update the view if necessary. Minimal tests passed, let's see what the checks on Github say. I need to add more docs and cleanup, but I think this might work :pray:

src/core/linalg/src/sparse/4C_linalg_view.hpp
vovannikov: > propagate const-ness from the outside to the value

OK, then it is really different from what `unique_ptr` does. It is clear now for me. Sounds like a `ConstEnforcingWrapper` to me or something like that :) 

How do you plan to replace the contained value? By accessing directly `obj`? Again, to me this looks like the replication of the `unique_ptr`'s behavior provided by methods `reset()` and `release()`.

With this in mind, out of general curiosity? Why do you need to impose additional const-ness on the underlying value object? Why is the default `unique_ptr` behavior not sufficient?

Regardless of this, I still think it is worth moving deleter to the template parameter due to enhanced flexibility. Even `unique_ptr` can not be used with CTAD if a user provides a custom deleter, whereas your class can! :)

sebproell:
Thanks for all your input! Could you carefully review from scratch again? We now have:

- `OwnerOrView` as a general helper (this evolved out of the `ConstPropagatingUniquePtr` class)
- Generalization of `LinAlg::View` such that it can be synced. Please check the docs for the idea.
- I applied it to the `Map` and inside the `SparseMatrix` as a test.

src/core/linalg/src/sparse/4C_linalg_view.hpp
sebproell: @vovannikov I basically need a container that propagaes const so that this works: https://github.com/4C-multiphysics/4C/pull/787#discussion_r2093246398

vovannikov (approved):

knarfnitram (approved):

PR #792, opened 2025-05-20
Remove doc/reference_docs dir before creating new documentation

ischeider:
# Description and Context

Similar to what @sebproell already did in PR #732 (deleting the documentation directories before building a new one), here I propose to delete another directory, which can lead to confusion if old/outdated files are left.

Remark: The directory `${_sphinx_OUT_DIR}/reference_docs` is created during the call of `create_rtd`. This program can also handle a directory name as argument - which is not given so far, such that the default `reference_docs` is used. One might also enforce the directory name as argument, so that the directories are always consistent (not done so far).

PR #732

## Interested parties
@sebproell


doc/documentation/CMakeLists.txt
```diff
@@ -23,9 +23,6 @@ if(FOUR_C_BUILD_DOCUMENTATION)
   set(_sphinx_HTML_DIR ${_sphinx_OUT_DIR}/html)
 
   file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/doc/tmp)
```
sebproell, line 25: Should this one also move into the `add_custom_target`?

sebproell (dismissed):

doc/documentation/CMakeLists.txt
ischeider: For line 25 I agree.

sebproell (approved):

PR #793, opened 2025-05-20
Rework of ssi monolithic meshtying strategy

c-p-schmidt:
## Description and Context
The SSI MeshtyingStrategy stored pointers to other objects within SSI, namely the SSIMaps and SSIMeshTying. This is now changed such that those objects are handed in to the respective method as a const reference whenever required
Further changes:
- added [[nodiscard]] where reasonable
- changed to std::unique_ptr as return type of the factory

### DIFF
--- a/src/ssi/4C_ssi_dyn.cpp
+++ b/src/ssi/4C_ssi_dyn.cpp
@@ -11,6 +11,7 @@
 #include "4C_io_control.hpp"
 #include "4C_rebalance_print.hpp"
 #include "4C_ssi_monolithic.hpp"
+#include "4C_ssi_monolithic_meshtying_strategy.hpp"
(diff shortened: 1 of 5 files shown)


sebproell (approved):
Maybe some methods can now become free functions?

c-p-schmidt:
> Maybe some methods can now become free functions?

You are right, I'll work on this in a future pull request.

c-p-schmidt:
> LGTM. I'm not the biggest fan of `do_uncomplete` of matrices though ... but of course not related to this PR 😉.

I can remember that Stephan and I designed this several years ago and implemented a logic that incomplete is only called when additional contributions are expected, i.e., the graph changes. If there is a better solution, I'm happy to get your ideas. ;-)

PR #795, opened 2025-05-20
Remove more `get_condition` from Node class

sebproell:
I am trying to remove the `Condition*` from the `Node` object.  This PR improves a few utility functions to find nodes carrying conditions. I then use these functions to find these node IDs instead of asking the node object directly. There are still some places left where the old functions are used, but I would like to get this merged as a first step.


sebproell:
I would like to merge this batch of changes so I don't have to keep rebasing.

davidrudlstorfer (approved):

sebproell:
This would still be ready.

PR #788, opened 2025-05-19
Remove LinAlg::Map constructors taking Epetra_Comm

ppraegla:
This reduces the use of `as_epetra_comm()` to convert the `MPI_Comm` to `Epetra_Comm` when using the LinAlg::Map wrapper.

LinAlg::Map::Comm() was renamed to EpetraComm(). A new function LinAlg::Map::Comm() now returns the MPI_Comm.

Closes #763

PR #799, opened 2025-05-21
Replace Epetra_BlockMap with Map

knarfnitram:
## Description and Context

As the title indicates, this MR adapts `Epetra_BlockMap` occurrences to our Wrapper `Core::LinAlg::Map`.

- Provides a unit test `4C_linalg_map_test.np2.cpp ` to ensure that the `Epetra_Map` of a `Vector` or `Multivector` may be exchanged.
- Utilizes the view introduced in #790 for the `Core::LinAlg::Vector`.

Related to #552 .

### DIFF
--- a/src/core/comm/src/4C_comm_utils.cpp
+++ b/src/core/comm/src/4C_comm_utils.cpp
@@ -361,7 +361,8 @@ namespace Core::Communication
           vecmap, Core::Communication::num_mpi_ranks(lcomm) - 1);
 
     // export full vectors to the two desired processors
-    Core::LinAlg::MultiVector<double> fullvec(proc0map->get_epetra_map(), vec.NumVectors(), true);
+    Core::LinAlg::MultiVector<double> fullvec(
+        proc0map->get_epetra_block_map(), vec.NumVectors(), true);
     Core::LinAlg::export_to(vec, fullvec);
 
     const int myglobalrank = Core::Communication::my_mpi_rank(gcomm);
@@ -509,7 +510,8 @@ namespace Core::Communication
           domainmap, Core::Communication::num_mpi_ranks(lcomm) - 1);
 
     // export full matrices to the two desired processors
-    Epetra_Import serialimporter(serialrowmap->get_epetra_map(), rowmap.get_epetra_map());
+    Epetra_Import serialimporter(
+        serialrowmap->get_epetra_block_map(), rowmap.get_epetra_block_map());
     Core::LinAlg::SparseMatrix serialCrsMatrix(*serialrowmap, 0);
     serialCrsMatrix.import(matrix, serialimporter, Insert);
     serialCrsMatrix.complete(*serialdomainmap, *serialrowmap);
--- a/src/core/comm/src/4C_comm_utils.cpp
+++ b/src/core/comm/src/4C_comm_utils.cpp
@@ -14,7 +14,6 @@
 #include "4C_linalg_utils_sparse_algebra_manipulation.hpp"
 
 #include <Epetra_Import.h>
-#include <Epetra_Map.h>
 
 #include <iomanip>
 #include <sstream>
@@ -348,8 +347,8 @@ namespace Core::Communication
       return false;
     }
 
-    // do stupid conversion from Epetra_BlockMap to Core::LinAlg::Map
-    const Epetra_BlockMap& vecblockmap = vec.Map();
+    // do stupid conversion from Core::LinAlg::Map to Core::LinAlg::Map
+    const Core::LinAlg::Map& vecblockmap = vec.get_map();
     Core::LinAlg::Map vecmap(vecblockmap.NumGlobalElements(), vecblockmap.NumMyElements(),
         vecblockmap.MyGlobalElements(), 0, vec.Comm());
 
@@ -423,7 +422,7 @@ namespace Core::Communication
           std::stringstream diff;
           diff << std::scientific << std::setprecision(16) << maxdiff;
           std::cout << "vectors " << name << " do not match, difference in row "
-                    << fullvec.Map().GID(i) << " between entries is: " << diff.str().c_str()
+                    << fullvec.get_map().GID(i) << " between entries is: " << diff.str().c_str()
                     << std::endl;
         }
         maxdiff = std::max(maxdiff, difference);
--- a/src/core/linalg/src/sparse/4C_linalg_graph.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_graph.hpp
@@ -114,7 +114,8 @@ namespace Core::LinAlg
       int err = 0;
 
       if (graphtype_ == CRS_GRAPH)
-        err = graph_->FillComplete(domain_map.get_epetra_map(), range_map.get_epetra_map());
+        err = graph_->FillComplete(
+            domain_map.get_epetra_block_map(), range_map.get_epetra_block_map());
       else if (graphtype_ == FE_GRAPH)
         err = static_cast<Epetra_FECrsGraph*>(graph_.get())
                   ->GlobalAssemble(domain_map.get_epetra_map(), range_map.get_epetra_map());
--- a/src/core/linalg/src/sparse/4C_linalg_graph.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_graph.hpp
@@ -18,7 +18,7 @@
 #include <Epetra_FECrsGraph.h>
 
 #include <memory>
-
+#include <optional>
 
 
 FOUR_C_NAMESPACE_OPEN
@@ -65,9 +65,6 @@ namespace Core::LinAlg
     const Epetra_CrsGraph& get_epetra_crs_graph() const { return *graph_; }
 
     Epetra_CrsGraph& get_epetra_crs_graph() { return *graph_; }
-
-    //! Returns the Column Map associated with this graph.
-    const Epetra_BlockMap& col_map() const { return (graph_->ColMap()); }
 
     //! Returns a pointer to the Epetra_Comm communicator associated with this graph.
     const Epetra_Comm& get_comm() const { return (graph_->Comm()); }
@@ -168,13 +165,34 @@ namespace Core::LinAlg
     //! Remove a list of elements from a specified global row of the graph.
     int remove_global_indices(int GlobalRow, int NumIndices, int* Indices);
 
-    const Epetra_BlockMap& row_map() const { return graph_->RowMap(); }
+    //! Returns the Row Map associated with this graph.
+    const Map& row_map() const
+    {
+      if (!row_map_)
+      {  // check if view is uninitialized
+        row_map_ = Core::LinAlg::Map(graph_->RowMap());
+      }
+      return *row_map_;
+    }
+
+    //! Returns the Column Map associated with this graph.
+    const Map& col_map() const
+    {
+      if (!col_map_)
+      {  // check if view is uninitialized
+        col_map_ = Core::LinAlg::Map(graph_->ColMap());
+      }
+      return *col_map_;
+    }
 
    private:
     GraphType graphtype_;
 
     //! The actual Epetra_CrsGraph object.
     std::unique_ptr<Epetra_CrsGraph> graph_;
+
+    mutable std::optional<Map> row_map_;
+    mutable std::optional<Map> col_map_;
   };
 }  // namespace Core::LinAlg
 FOUR_C_NAMESPACE_CLOSE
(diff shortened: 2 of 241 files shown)


src/core/linalg/src/sparse/4C_linalg_map.cpp
```diff
@@ -66,6 +66,31 @@ std::unique_ptr<const Core::LinAlg::Map> Core::LinAlg::Map::create_view(const Ep
   ret->map_ = Utils::make_view<Epetra_Map>(const_cast<Epetra_Map*>(&view));
   return ret;
 }
+std::unique_ptr<Core::LinAlg::Map> Core::LinAlg::Map::create_view(Epetra_BlockMap& view)
+{
+  // cast the Epetra_BlockMap to Epetra_Map, this cast can go wrong.
+  const Epetra_Map* epetra_map_ptr = static_cast<const Epetra_Map*>(&view);
```
sebproell, line 72: As you said yourself in the comment, this does not necessarily work. I think you should use `dynamic_cast` and check if the casted pointer is non-null. If it is indeed null, we throw an exception with a nice explanation like:

"You are trying to view an Epetra_BlockMap as a LinAlg::Map. This only works if the map you are viewing is of the derived type Epetra_Map. This is not true for the map you passed to this function."

src/core/linalg/src/sparse/4C_linalg_map.cpp
```diff
@@ -66,6 +66,31 @@ std::unique_ptr<const Core::LinAlg::Map> Core::LinAlg::Map::create_view(const Ep
...
+
+  // We may safely cast away const here, since the returned object is const and we use a
+  // OwnerOrView to ensure that we only ever access the viewed map through
+  // const methods.
+  std::unique_ptr<Map> ret(new Map);
+  ret->map_ = Utils::make_view<Epetra_Map>(const_cast<Epetra_Map*>(epetra_map_ptr));
```
sebproell, line 78: Also, for the non-const version, just don't cast to const Epetra_Map above, then you don't need const_cast here.

src/core/linalg/tests/4C_linalg_map_test.np2.cpp
```diff
@@ -0,0 +1,111 @@
...
+      // set up communicator
+      comm_ = MPI_COMM_WORLD;
+
+      // ensure that this test is run in parallel
+      int size;
+      MPI_Comm_size(MPI_COMM_WORLD, &size);
+      if (size <= 1)
+      {
+        GTEST_SKIP() << "This test requires more than 1 MPI rank.";
+      }
```
sebproell, line 42: That's unnecessary? We always add a dummy test to all our test executables that checks for the correct number of ranks, so that a test run with the wrong number is reported as failed.

src/core/linalg/tests/4C_linalg_map_test.np2.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include "4C_linalg_sparsematrix.hpp"
+#include "4C_linalg_vector.hpp"
+
+#include <memory>
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace
+{
+  class ExchangeMapTest : public testing::Test
```
sebproell, line 23: I believe the current tests you wrote (which I like!) would be easier to read if you remove this fixture class. There is little to no setup required and I would put this in the test, so I don't have to jump around the code to see what `starting_map` is.

src/core/linalg/tests/4C_linalg_map_test.np2.cpp
```diff
@@ -0,0 +1,111 @@
...
+  TEST_F(ExchangeMapTest, Vector)
+  {
+    // create a vector
+    auto vector = Core::LinAlg::Vector<double>(starting_map_->get_epetra_map(), true);
+
+    const Epetra_Map& expected_map = starting_map_->get_epetra_map();
+    const Epetra_Map& actual_map = vector.get_map().get_epetra_map();
+
+    // check if underlying maps are identical.
+    ASSERT_TRUE(actual_map.SameAs(expected_map));
```
sebproell, line 58: Not super critical, but you usually use the `EXPECT_` variant of gtest macros unless you really want to abort further testing should the check fail. Is what you think should happen here?

sebproell (commented):
Looks neat, @knarfnitram! I like to see that the `View` can help us here.

src/core/linalg/src/sparse/4C_linalg_map.cpp
knarfnitram: I tried the `dynamic_cast`, but for the creation of the `const Epetra_BlockMap` view, but unfortunately it does not work. If I do so all tests fail on runtime. For the `non-const view ` I added it.

src/core/linalg/tests/4C_linalg_map_test.np2.cpp
knarfnitram: ok, I didn't know that and removed the check.

src/core/linalg/tests/4C_linalg_map_test.np2.cpp
knarfnitram: Thanks for the comment, I removed the fixture object completely.

src/core/linalg/src/sparse/4C_linalg_map.cpp
knarfnitram: In the cases I viewed `dynamic_cast` did return a nullptr.

src/core/linalg/tests/4C_linalg_map_test.np2.cpp
knarfnitram: I think the tests may continue, thanks for the hint, i adapted it to `EXPECT`.

src/core/linalg/src/sparse/4C_linalg_map.cpp
sebproell: Well, we still need to fix this. It is UB to just view the memory of a C++ object as an object of another type unless we do some specific manipulations to tell the language and compiler that we can do this deliberately. I believe we cannot do this in the case of inheritance safely, so we need to think about some other way.

sebproell: One way could be to wrap either the block or non-block Map (via a `variant`) and use the according implementation.

src/core/comm/src/4C_comm_utils.cpp
```diff
@@ -348,8 +347,8 @@ namespace Core::Communication
       return false;
     }
 
-    // do stupid conversion from Epetra_BlockMap to Core::LinAlg::Map
-    const Epetra_BlockMap& vecblockmap = vec.Map();
+    // do stupid conversion from Core::LinAlg::Map to Core::LinAlg::Map
```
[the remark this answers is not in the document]
knarfnitram: This is a good indication for clean up: Since we now use the map wrapper we can remove the comment and use the map directly. i therefore would tackle this separately.

src/core/linalg/src/sparse/4C_linalg_graph.hpp
```diff
@@ -168,13 +165,34 @@ namespace Core::LinAlg
...
+      return *col_map_;
+    }
 
    private:
     GraphType graphtype_;
 
     //! The actual Epetra_CrsGraph object.
     std::unique_ptr<Epetra_CrsGraph> graph_;
+
+    mutable std::optional<Map> row_map_;
```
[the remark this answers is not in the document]
knarfnitram: Yes, you are right, however i would like to this seperatly from this MR.

src/core/linalg/src/sparse/4C_linalg_multi_vector.hpp
```diff
@@ -19,6 +19,7 @@
 #include <mpi.h>
 
 #include <memory>
+#include <optional>
```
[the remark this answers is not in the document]
knarfnitram: I think not, thanks for catching that.

knarfnitram:
Thanks for the help and feedback @sebproell and @maxfirmbach .
I added the `Epetra_BlockMap` to the class  `Map`  by using  `variant`. It's now possible to distinguish between these two objects from the interface. The corresponding casts should now be fine or lead to an error. Let's see if the tests are passing.

src/core/linalg/tests/4C_linalg_map_test.np2.cpp
```diff
@@ -0,0 +1,106 @@
...
+    const Epetra_Map& expected_map = starting_map->get_epetra_map();
+    const Epetra_BlockMap& actual_map = vector.get_map().get_epetra_block_map();
+
+    // check if underlying maps are identical.
+    EXPECT_TRUE(actual_map.SameAs(expected_map));
+
+    // create a new map with block map and different index base
+    auto new_epetra_block_map = std::make_shared<Epetra_BlockMap>(
+        NumGlobalElements, 1, 1, Core::Communication::as_epetra_comm(comm));
+    auto new_map = std::make_shared<Core::LinAlg::Map>(*new_epetra_block_map);
```
knarfnitram, line 46: All constructors are currently based on the `Epetra_Map`, is that a problem/should I add an additional explicit constructor for the `BlockMap` ? 
If you want to create a `Map` based on an `Epetra_BlockMap` you currently need this workaround, if the `BlockMap` is not available from another Epetra_object.

src/core/linalg/src/sparse/4C_linalg_map.hpp
```diff
@@ -27,11 +28,29 @@ FOUR_C_NAMESPACE_OPEN
...
 
+
+
 namespace Core::LinAlg
 {
-
   class Map
   {
+    // This wrapper may have two different variants
+    // Either it holds a "view" to the Epetra_Map or Epetra_BlockMap
```
sebproell, line 38:
```suggestion
    // Either it holds an  Epetra_Map or Epetra_BlockMap
```

Minor: I wouldn't call this "view" here.

src/core/linalg/src/sparse/4C_linalg_map.hpp
```diff
@@ -45,122 +64,289 @@ namespace Core::LinAlg
...
+      // If the Epetra_Map is stored try to cast it.
+      auto& map = *std::get<Utils::OwnerOrView<Epetra_Map>>(map_);
+      auto* block_map = dynamic_cast<const Epetra_BlockMap*>(&map);
+
+      if (block_map == nullptr)
+      {
+        FOUR_C_THROW("The dynamic cast from Epetra_Map to Epetra_BlockMap failed.");
+      }
+
+      return *block_map;
```
sebproell, line 125:
```suggestion
      return map;
```

This up-cast will always work. An `Epetra_Map` is-a `Epetra_BlockMap`. Same below.

src/core/linalg/src/sparse/4C_linalg_map.hpp
```diff
@@ -45,122 +64,289 @@ namespace Core::LinAlg
...
+    {
+      return visit_variant([](const auto& map) { return map.MinMyGID(); });
+    }
 
-    MPI_Comm Comm() const { return Core::Communication::unpack_epetra_comm(map_->Comm()); }
+    //! Access function for Epetra_Comm communicator.
+    MPI_Comm Comm() const
+    {
+      return Core::Communication::unpack_epetra_comm(
+          visit_variant([](const auto& map) -> const Epetra_Comm& { return map.Comm(); }));
```
sebproell, line 294: This call was introduced in #788 and requires the inclusion of the expensive `4C_comm_mpi_utils.hpp` header. Which increased the compile time a lot. Since you are working on this file anyway, could you move this function to the implementation and remove the header-in-header inclusion?

src/core/linalg/src/sparse/4C_linalg_map.hpp
```diff
@@ -45,122 +64,289 @@ namespace Core::LinAlg
...
     [[nodiscard]] static std::unique_ptr<const Map> create_view(const Epetra_Map& view);
+    [[nodiscard]] static std::unique_ptr<Map> create_view(Epetra_BlockMap& view);
+    [[nodiscard]] static std::unique_ptr<const Map> create_view(const Epetra_BlockMap& view);
 
    private:
     Map() = default;
 
-    //! The actual Epetra_Map object.
-    Utils::OwnerOrView<Epetra_Map> map_;
+    // store view of Epetra_BlockMap or Epetra_Map
```
sebproell, line 343:
```suggestion
    //! stores an Epetra_BlockMap or Epetra_Map
```

sebproell (commented):
Nice! Let's go with the `variant` for now.

sebproell (dismissed):
@knarfnitram While looking at the code locally, I simplified the variant handling a little bit. I would push it as a follow-up.

sebproell (approved):

PR #810, opened 2025-05-23
NOX: remove obsolete setup() method

vovannikov:
## Description and Context
Some additional clean up after #798. Mainly, the removal of unnecessary `setup()` method and the internal flags. Now ctor does everything required for initialization.

#359, #379, #798

### DIFF
--- a/src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_generic.cpp
+++ b/src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_generic.cpp
@@ -21,14 +21,12 @@ Solid::Nln::SOLVER::Generic::Generic(
     const std::shared_ptr<Solid::TimeInt::BaseDataGlobalState>& gstate,
     const std::shared_ptr<Solid::TimeInt::BaseDataSDyn>& sdyn,
     const std::shared_ptr<Solid::TimeInt::NoxInterface>& noxinterface,
-    const std::shared_ptr<Solid::Integrator>& integrator,
+    const std::shared_ptr<Solid::Integrator>& integr,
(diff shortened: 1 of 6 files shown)


src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_generic.cpp
```diff
@@ -21,14 +21,12 @@ Solid::Nln::SOLVER::Generic::Generic(
     const std::shared_ptr<Solid::TimeInt::BaseDataGlobalState>& gstate,
     const std::shared_ptr<Solid::TimeInt::BaseDataSDyn>& sdyn,
     const std::shared_ptr<Solid::TimeInt::NoxInterface>& noxinterface,
-    const std::shared_ptr<Solid::Integrator>& integrator,
+    const std::shared_ptr<Solid::Integrator>& intgr,
```
mayrmt, line 24: What's the benefit of this change? The new abbreviation is less descriptive than the original.

mayrmt (dismissed):

src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_generic.cpp
vovannikov: Just to avoid name conflict with with the method `integrator()`, which previously was not an issue, since initialization was in a separate finction, but now I have moved it to ctor. I did not want to use `this->integrator()` since `this` is not frequently used across the code in general (and I agree with this).

mayrmt (approved):

mayrmt (approved):

sebproell (approved):

PR #811, opened 2025-05-23
Add image for oldest supported dependencies

### LINKED ISSUE #483: Allow for more flexible versions of dependencies

sebproell:
This PR adds a docker image with the oldest supported dependencies. This image is updated in the same workflow as the existing image. We propose to run a nightly build on this image to catch wrong `#ifdefs` or APIs (see e.g. #806)  (unchached, so we don't overload the cache as much). Nightly should be enough since very few PRs change anything in the dependencies. We can always move it to PRs if the check fails too frequently.

Co-authored by @ppraegla 

Closes #483

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -8,7 +8,7 @@ jobs:
   gcc13_assertions_build:
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:502560ce
+      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:8115e93e
       options: --user root --env OMPI_ALLOW_RUN_AS_ROOT=1 --env OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
     defaults:
       run:
@@ -41,7 +41,7 @@ jobs:
     needs: gcc13_assertions_build
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:502560ce
+      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:8115e93e
       options: --user root --env OMPI_ALLOW_RUN_AS_ROOT=1 --env OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
     strategy:
       fail-fast: false
@@ -93,7 +93,7 @@ jobs:
   clang18_build:
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:502560ce
+      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:8115e93e
       options: --user root --env OMPI_ALLOW_RUN_AS_ROOT=1 --env OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
     defaults:
       run:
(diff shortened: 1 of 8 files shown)


ppraegla:
> We can always move it to PRs if the check fails too frequently.

If required, we could also add a manual trigger. This way, the job can be triggered as needed when a PR does something with Trilinos.

Changes LGTM (but I also co-authored them :wink:  ). Maybe someone else can approve.

.github/workflows/nightly_tests.yml
```diff
@@ -240,10 +240,29 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - name: Check docker hash
+        uses: ./.github/actions/compute-and-check-dependencies-hash
+      - uses: ./.github/actions/build_4C
+        with:
+          cmake-preset: docker_clang
+          build-targets: full
+          build-directory: ${{ github.workspace }}/build
+          use-ccache: "false"
+
```
amgebauer, line 261: Why no testing?

sebproell: This was suggested by @ppraegla; I would be fine with tests as well.

ppraegla: I don't think that tests are necessary. If the build is successful, the code will likely work. We already use a lot of resources for testing. But I'm also fine with adding the tests as well.

sebproell: I added tests.

ppraegla (approved):

amgebauer (approved):

.github/workflows/nightly_tests.yml
```diff
@@ -240,10 +240,80 @@ jobs:
...
 
+  clang18_build_oldest_supported_dependencies:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04-oldest-supported:8115e93e
+      options: --user root --env OMPI_ALLOW_RUN_AS_ROOT=1 --env OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
+    defaults:
+      run:
+        shell: bash
+    steps:
```
ppraegla, line 251: You forgot 
```
outputs:
test-chunks: ${{ steps.set-matrix.outputs.chunk-array }}
```
That's why the nightly pipeline fails https://github.com/4C-multiphysics/4C/actions/runs/15220171403
I will push a fix.

PR #816, opened 2025-05-23
Cleanup of scatra timint implicit

c-p-schmidt:
## Description and Context
While starting to work on removing the call of the 4C-coded block preconditioning, I got annoyed by so many things in `scatra_timint_implicit`. Thus, I decided to do some cleanup first, which includes, e.g., removing init methods and moving the relevant work to the constructor, removing unused methods, fixing naming, etc.

### DIFF
--- a/src/sti/4C_sti_partitioned.cpp
+++ b/src/sti/4C_sti_partitioned.cpp
@@ -50,19 +50,23 @@ STI::Partitioned::Partitioned(MPI_Comm comm,  //! communicator
     case Inpar::STI::CouplingType::twoway_thermotoscatra_aitken:
     {
       // initialize increment vectors
-      scatra_field()->phinp_inc() =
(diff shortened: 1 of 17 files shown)


src/sti/4C_sti_partitioned.cpp
```diff
@@ -84,11 +88,9 @@ STI::Partitioned::Partitioned(MPI_Comm comm,  //! communicator
     default:
     {
       FOUR_C_THROW("What the hell?!");
```
sebproell, line 90: Should this be `std23::unreachable()`?

sebproell (approved):

src/sti/4C_sti_partitioned.cpp
c-p-schmidt: Thank you for the hint and the offline discussion. I'll adapt this in complete scatra, sti etc. in a follow-up.

PR #819, opened 2025-05-23
Simplify access into internal variant items of Map

sebproell:
Instead of visting the variant, we can access the two possible options with a uniform `(const) Epetra_Map&`.

Follows #799


knarfnitram (approved):
@sebproell  thanks for refactoring it, this simplifies the internal access a lot.

Issue #207, opened 2025-01-08
Remove non-configurable pre-processor define flags (and their underlying code)


eulovi:
> This issue was originally created by @mayrmt in https://gitlab.lrz.de/baci/baci/-/issues/1310 on 2024-07-25.

## Description

Over years, 4C has accrued a large number of pre-processor define flags (combined with `#ifdef` / `#ifndef` directives) to (de-)activate parts of the code depending on the state of the define flag. Most of them are hard-coded, thus do not change their state. Hence, the code guarded by these hard-coded define flags has not been compiled or tested for years.

To clean the code, this issue proposes to remove such hard-coded define flags. If combined with an `#ifdef` or `#ifndef` construct, the deactivated code shall be removed as well, since it has not been tested for several years anyways.

## Possible Solution and Definition of Done

For every pre-processor define that can not be controlled via a CMake option:

- [ ] Remove the define (either active or commented)
- [ ] Identify and remove the unused code (this depends on the state of the pre-processor define (active or inactive) and on the choice of `#ifdef` vs. `#ifndef`

### Example

This code
```c++
#define SOME_OLD_DEFINE_FLAG

  std::cout << "Hello, world." << std::endl;

#ifdef SOME_OLD_DEFINE_FLAG
  std::cout << "This code is compiled, executed, and tested." << std::endl;
#else
  std::cout << "This code is neither compiled nor tested, thus shall be removed." << std::endl;
#endif

  std::cout << "Hello, world." << std::endl;
```
shall be transformed into
```c++
  std::cout << "Hello, world." << std::endl;
  std::cout << "This code is compiled, executed, and tested." << std::endl;
  std::cout << "Hello, world." << std::endl;
```

## Interested Parties

@baci/baci_developers

PR #821, opened 2025-05-23
Avoid include cycles

sebproell:
This PR adds a nice modularity check based on an implementation in deal.II: https://github.com/dealii/dealii/pull/18013

The cycle in `cut` is only a thing if the `#ifdef` were turned on, but we want to remove these anyway (see #207).


davidrudlstorfer:
Really nice check 👌

Issue #755, opened 2025-05-09
[LinAlg] Create wrapper classes for `Epetra_Import` and `Epetra_Export`


maxfirmbach:
## Description

Create a wrapper named e.g. 4C_linalg_transfer.* to centralize all calls to `Epetra_Import` and `Epetra_Export`.

I would consider this an relatively easy issue and a good one to start, if you did not contribute as much so far.

## Possible Solution and Definition of Done

Related to `Epetra_Import` (currently I see 77 overall occurrences either as objects, in comments or as string):
- [x] Provide a wrapper class for `Epetra_Import` (`Core::LinAlg::Map` is a good reference)
- [x] Replace all mentions of `Epetra_Import` by either referring to the wrapper class and/or replace explicit calls by it

Related to `Epetra_Export` (currently I see 44 overall occurences either as objects, in comments or as string):
- [x] Provide a wrapper class for `Epetra_Export` (`Core::LinAlg::Map` is a good reference)
- [x] Replace all mentions of `Epetra_Export` by either referring to the wrapper class and/or replace explicit calls by it

Post implementation clean up:
- [x] `get_epetra_import()`  
- [ ] `get_epetra_export()`calls
- [ ] Simplify some Map calls : e.g : `get_epetra_block_map()`

maxfirmbach:
@dharinib98 How is the overall progress on this issue? Is this actually done?

dharinib98:
> [@dharinib98](https://github.com/dharinib98) How is the overall progress on this issue? Is this actually done?

There are still a few more raw `Epetra` calls that can be removed only after we have our own implementation of some `Epetra` methods that we call on this. Should be one or two methods. I will get that done.

dharinib98:
There are still three remaining calls to `get_epetra_export`: two on Epetra_FEVector, which I'm leaving as-is for now. The third is on an `Epetra_CrsMatrix`, which can be cleaned up once we have our own import implementation for `Epetra_CrsMatrix`. Once that's done—and after removing the remaining `get_epetra_block_map` calls—this issue can be closed.

maxfirmbach:
@dharinib98 I would propose to move the remaining cleanup tasks to https://github.com/4C-multiphysics/4C/issues/863 and collect them there. This would give us the possibility to close this issue. What do you think?

dharinib98:
@maxfirmbach yes, I thought of something similar too. I will do that.

dharinib98:
Rest of the tasks moved to #863

PR #827, opened 2025-05-25
Implement wrapper for Epetra import

dharinib98:
## Description and Context

This PR implements a wrapper for `Epetra_Import `, and replacing its occurrences in the code. Cleaning up of certain get_epetra_XXX calls will follow.  

#755


src/core/linalg/src/sparse/4C_linalg_graph.hpp
```diff
@@ -12,6 +12,7 @@
 #include "4C_config.hpp"
 
 #include "4C_linalg_map.hpp"
+#include "4C_linalg_transfer.hpp"
 
 #include <Epetra_CrsGraph.h>
 #include <Epetra_Export.h>
```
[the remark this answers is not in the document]
dharinib98: Although we do not call `Epetra Export` class explicitly in the file, there are some smart pointers that needs the full class definition and hence we need to keep this.

src/beamcontact/4C_beamcontact_beam3contact_octtree.cpp
```diff
@@ -18,9 +18,11 @@
 #include "4C_global_data.hpp"
 #include "4C_io_value_parser.hpp"
 #include "4C_linalg_sparsematrix.hpp"
+#include "4C_linalg_transfer.hpp"
 #include "4C_linalg_utils_sparse_algebra_math.hpp"
 #include "4C_rigidsphere.hpp"
 
+#include <Epetra_Export.h>
```
knarfnitram: Thanks providing the Importer @dharinib98 , I have only one small question:
Do we really have to add this epetra header here?

dharinib98: Thanks for pointing it out. Resolved

mayrmt (commented):
I am aware, that the original issue #755 mentions `4C_linalg_transfer.*` as an example for a file name. I don't like the "transfer" notion, since this can be confused with transfer operators in multigrid methods. Why not use two files, `4C_linalg_import.*` and `4C_linalg_export.*`? This would further allow to include import and export functionality independently, if only on of them is needed. Just a thought...

maxfirmbach:
> I am aware, that the original issue #755 mentions `4C_linalg_transfer.*` as an example for a file name. I don't like the "transfer" notion, since this can be confused with transfer operators in multigrid methods. Why not use two files, `4C_linalg_import.*` and `4C_linalg_export.*`? This would further allow to include import and export functionality independently, if only on of them is needed. Just a thought...

To me a name change is okay. I just want to highlight that most people would never relate transfer and multigrid as most either dont know about the theory or it just doesnt matter to them. In `Tpetra` both import and export operations are implemented based on a `Transfer` class, thus the naming suggestion 😉.

dharinib98:
To get things going, I will proceed with this merge right now and incase there is a better naming suggestion for the file, it can be implemented along with the upcoming Epetra_Export PRs

PR #828, opened 2025-05-26
Implement energy embeddedmesh

eulovi:
## Description and Context

In this PR, the evaluation of the energy generated by the coupling of embedded meshes is implemented and tested with a small Hertzian contact problem.


isteinbrecher (approved):
Thanks for adding this feature @eulovi

dharinib98:
Thanks @eulovi  for implementing this feature. The current state looks good to me. Maybe I am not the right person to comment on how to further enhance the implementation especially with regard to the coding part. However just a related remark, would it be useful to have an input flag that controls the computation, so that it is more user friendly and can be controlled from the input files?

src/constraint_framework/4C_constraint_framework_submodelevaluator_embeddedmesh.cpp
```diff
@@ -96,4 +96,20 @@ void Constraints::SubmodelEvaluator::EmbeddedMeshConstraintManager::runtime_outp
...
+{
+  std::map<Solid::EnergyType, double> embedded_mesh_energy;
+  double mortar_manager_energy = mortar_manager_->get_energy();
+
+  // The value we returned here is summed up over all processors. Since we already have the global
+  // energy here, we only return it on rank 0.
+  if (mortar_manager_->get_my_rank() == 0)
+  {
+    embedded_mesh_energy[Solid::embedded_mesh_penalty_potential] = mortar_manager_energy;
+  }
```
sebproell: This seems unnecessarily complicated. Why not return it on all ranks? Of course, you can only write it on rank 0 later.

src/constraint_framework/4C_constraint_framework_embeddedmesh_solid_to_solid_mortar_manager.hpp
```diff
@@ -166,6 +166,16 @@ namespace Constraints::EmbeddedMesh
      */
     void collect_output_integration_points();
 
+    /**
+     * \brief Get the rank where the mortar manager is being called
+     */
+    int get_my_rank();
+
```
sebproell: This is a weird function to have. Can you remove it and just return the result everywhere? If not, I would at least return the `MPI_Comm` and not part of its details.

sebproell (commented):
Looks good, but I wonder if you can make it a bit simpler.

src/constraint_framework/4C_constraint_framework_submodelevaluator_embeddedmesh.cpp
isteinbrecher: The issue here is that the mortar manager already returns the global energy. The result returned in this function is summed up over all processors, so we only return it on processor 0.

sebproell: Okay, so the result is only valid on rank 0? I am not familiar with the mortar code. This code just stood out to me as an unusual MPI pattern. Essentially, I am wondering what the energy value is going to be on rank =/= 0. What would happen if I just do

```
    embedded_mesh_energy[Solid::embedded_mesh_penalty_potential] = mortar_manager_energy;
```

on every rank?

sebproell: Oh, now I got it! The `get_energy` is somehow embedded into an algorithm that performs a sum later. This algorithm expects more of a "local energy contribution" but we already have the global energy here.

eulovi:
@dharinib98 Thanks for your comment! The input flag `RESEVERYERGY` currently controls the output of energy contributions (the naming of this flag is probably not the best). These evaluations usually come from the respective model evaluator I think. You can find all the energy contributions in the resulting .csv file ;).

src/constraint_framework/4C_constraint_framework_submodelevaluator_embeddedmesh.cpp
eulovi: Yes! I think this sum of contributions is done internally by a `Solid::ModelEvaluator::Data` instance in the model evaluator. But it first has to collect the individual energy contributions generated by the  submodel evaluators.

src/constraint_framework/4C_constraint_framework_embeddedmesh_solid_to_solid_mortar_manager.hpp
eulovi: I just applied your second option. Now the `MPI_Comm` its returned from the mortar manager.

eulovi:
@sebproell @dharinib98 if this PR is in a good state for you, could you approve?

amgebauer:
@eulovi your added test fails for the no-dependencies-job due to missing ArborX (see the nightly pipelines of the weekend). Could you please add `if (FOUR_C_WITH_ARBORX)` and `endif (FOUR_C_WITH_ARBORX)` to make sure that the test is not running if ArborX is not installed?

eulovi:
@amgebauer yes, sorry about that! This is being fixed in #918.

Issue #809, opened 2025-05-22
Building in debug results in huge memory consumption


maxfirmbach:
## Description
Building `4C` in debug mode results in an extreme memory usage for creating the test executables.
E.g. building with `make -j 8 full` can use close to 64GB of RAM.

This seems to be not desirable.

## Interested Parties
@RShaw026

RShaw026:
@sebproell Thank you for your recommendations. Now building in debug mode is using significantly less memory. 
@maxfirmbach We can close this issue.

PR #829, opened 2025-05-26
[CMake] Use FOUR_C_BUILD_SHARED_LIBS insted of BUILD_SHARED_LIBS

sebproell:
This adds an entry in the docs. We internally set BUILD_SHARED_LIBS to the same value. Backwards-compatibility is provided with a warning that the new version should be used.

Might help with #809. If people don't know about this option, they can't use it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,7 +175,7 @@ add_library(${FOUR_C_LIBRARY_NAME})
 # Remove the CMake generated prefix "lib" because we already added it manually, to have a distinct target name.
 set_target_properties(${FOUR_C_LIBRARY_NAME} PROPERTIES PREFIX "")
 target_link_libraries(${FOUR_C_LIBRARY_NAME} PRIVATE four_c_private_compile_interface)
-if(NOT BUILD_SHARED_LIBS)
+if(NOT FOUR_C_BUILD_SHARED_LIBS)
   install(
     TARGETS four_c_private_compile_interface
     EXPORT 4CTargets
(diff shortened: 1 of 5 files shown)


davidrudlstorfer (approved):
Looks good - should we create an issue to actually delete the backwards compatibility some day?

sebproell:
Another question would be whether we should make shared libraries the default? I think so. It saves a ton of space in debug. @davidrudlstorfer Do you remember where you posted your performance comparison?

davidrudlstorfer:
> Another question would be whether we should make shared libraries the default? I think so. It saves a ton of space in debug. @davidrudlstorfer Do you remember where you posted your performance comparison?

Unfortunately, I posted it originally on Slack but found a small summary in our private Slack chat:

> **kleine Simulationen (single thread):**
> Release => shared ca. 0-5% langsamer (Unterschiede fokussieren sich im Time Monitor nicht auf einzelne Routinen sondern alles ist quer durch die Bank langsamer)
> Debug => shared ca. 10-20% langsamer (wie oben, durch die Bank langsamer)
> **große Simulationen (single thread and multi thread np=4):**
> Release => shared ca. 0-1% langsamer (schwer messbar wo es langsamer ist)
> Debug => shared ca. 15% langsamer (Unterschiede sind im Löser etwas geringer als im Rest von 4C)
> 
> Bei größeren Simulationen ist der Unterschied in meinen Cases mehr oder weniger vernachlässigbar.

knarfnitram (approved):

PR #834, opened 2025-05-27
Use View mechanism in Vector wrappers

sebproell:
Use our `LinAlg::View` mechanism inside the Vector wrappers. I initially had another view implementation where we synced vectors after dangerous operations. It is easier if we just use `View::sync()` which handles this for us. 

I found another bug in `Epetra_MultiVector`: asking for a vector from a `Epetra_MultiVector` will only create the correct vector once. After a call to `Epetra_MultiVector::ReplaceMap()` this vector will not be updated; see also the test in this PR. I decided not to fix this on our side. Let's just remove `replace_map` from our interface.

### DIFF
--- a/src/core/linalg/tests/4C_linalg_vector_test.np2.cpp
+++ b/src/core/linalg/tests/4C_linalg_vector_test.np2.cpp
@@ -281,6 +281,10 @@ namespace
     Core::LinAlg::Vector<double> a(*map_, true);
     a.put_scalar(1.0);
 
+    const Core::LinAlg::MultiVector<double>& b = a;
+    ASSERT_EQ(b.NumVectors(), 1);
+    const Core::LinAlg::Vector<double>& c = b(0);
+
     // New map where elements are distributed differently
     std::array<int, 5> my_elements;
     if (Core::Communication::my_mpi_rank(comm_) == 0)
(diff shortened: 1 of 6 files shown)


src/core/linalg/tests/4C_linalg_vector_test.np2.cpp
```diff
@@ -289,14 +293,25 @@ namespace
...
-    EXPECT_TRUE(a.get_map().same_as(c.get_map()));
+    {
+      // This highlights a bug in Trilinos: the Epetra_Vector views into a MultiVector are only
+      // set once and never updated, although a map replacement would require an update.
+      const Core::LinAlg::MultiVector<double>& b_new = a;
+      const Core::LinAlg::Vector<double>& c_new = b_new(0);
+      // This is correct.
+      EXPECT_TRUE(b_new.get_map().same_as(new_map));
+      // This is the bug: c_new still has the old map although we just took a new view into b_new.
+      EXPECT_TRUE(c_new.get_map().same_as(*map_));
```
mayrmt, line 313: You could/should perform a more stringent test on map equality:
```suggestion
      // This is correct.
      EXPECT_TRUE(b_new.get_map().point_same_as(new_map));
      // This is the bug: c_new still has the old map although we just took a new view into b_new.
      EXPECT_TRUE(c_new.get_map().point_same_as(*map_));
```

src/core/linalg/tests/4C_linalg_vector_test.np2.cpp
```diff
@@ -289,14 +293,25 @@ namespace
...
-    // A change of the map is reflected to all views
+    EXPECT_EQ(&a.get_ref_of_epetra_vector(), &b.get_epetra_multi_vector());
+    // A change of the map invalidates views, so we need to be careful.
     a.replace_map(new_map);
+    EXPECT_NE(&a.get_ref_of_epetra_vector(), &b.get_epetra_multi_vector());
 
-    EXPECT_TRUE(a.get_map().same_as(b.get_map()));
-    EXPECT_TRUE(a.get_map().same_as(c.get_map()));
+    {
+      // This highlights a bug in Trilinos: the Epetra_Vector views into a MultiVector are only
```
mayrmt, line 306: Do you know, if this is also the case for Tpetra? This should be reported to Trilinos, so that it can be fixed at its root.

mayrmt (commented):
@sebproell, I am mainly wondering, if this can be fixed at its root.

src/core/linalg/tests/4C_linalg_vector_test.np2.cpp
sebproell: I had a quick look at Tpetra and I at least do not see the same problem right away -- although I have looked very little into this package and might simply miss it.

src/core/linalg/tests/4C_linalg_vector_test.np2.cpp
mayrmt: You're probably right. I somehow thought, that `PointSameAs` is more strict than `SameAs`, but I can't confirm this right now in code.

sebproell:
I believe this PR is correct regardless of #835.

PR #837, opened 2025-05-27
History variables multiscale time integration

georghammerl:
## Description and Context
The current implementation of the multiscale approach for solids is buggy with respect to internal variables on element level (e.g. EAS, FBAR, plasticity history variables). In previous times, EAS history was tracked for the micro scale manually. FBAR and plasticity have been introduced later and no adaptions have been made. Hence, FBAR and plasticity internal states on micro scale are not accounted for properly. 

Background: The micro scale discretization is reused for evaluation of all macro elements which hold a micro scale. The states are restored at the beginning of an evaluation and stored after the evaluation for the following evaluation of this element. The advantage of this approach is that only the states and internal variables (e.g. EAS, FBAR, plasticity) of each micro scale need to be stored and not a complete discretization. This approach demands for each element on the micro scale to pack and unpack all its internal states that vary over the simulation. These missing routines are implemented in `MicroMaterialGP`.

## Interested parties
@ischeider


src/core/io/src/4C_io.hpp
```diff
@@ -135,13 +135,13 @@ namespace Core::IO
     double read_double(std::string name);
 
     /// read into the discretization given in the constructor
-    void read_mesh(int step);
+    void read_mesh(const int step);
```
sebproell, line 138: This is noise, IMO. The `const` has no significance in this declaration.

src/mat/4C_mat_micromaterialgp_static.hpp
```diff
@@ -22,6 +22,14 @@ FOUR_C_NAMESPACE_OPEN
...
 // forward declarations
 
+namespace Core
+{
+  namespace Communication
+  {
+    class PackBuffer;
+  }
+}  // namespace Core
+
```
sebproell, line 32: Seems unused (at least it's not in the diff).

src/solid_3D_ele/4C_solid_3D_ele.cpp
```diff
@@ -251,6 +251,27 @@ void Discret::Elements::Solid::unpack(Core::Communication::UnpackBuffer& buffer)
   Discret::Elements::unpack(solid_calc_variant_, buffer);
 }
 
+void Discret::Elements::Solid::pack_history(Core::Communication::PackBuffer& data) const
+{
+  add_to_pack(data, unique_par_object_id());
+
+  // std::cout << "element id in pack_history: " << id() << std::endl;
```
sebproell, line 258:
```suggestion
```

src/core/fem/src/general/element/4C_fem_general_element.hpp
```diff
@@ -215,6 +215,19 @@ namespace Core::Elements
     */
     void unpack(Core::Communication::UnpackBuffer& buffer) override;
 
+    /*!
+    \brief Pack internal data of this class so it can be restored
+
+    */
+    void pack_history(Core::Communication::PackBuffer& data) const;
```
sebproell, line 222: Did you forget `virtual` here? Later on you cast to the Soild element, which would be resolved by virtual.

tests/input_files/sohex8_multiscale_macro_plastic.dat
sebproell, line 1: Please convert dat files to yaml.

sebproell (commented):
@georghammerl I don't want to add more functions to the `Element` interface. I did not understand from the diff and what you wrote why you need the new `pack_history()` call. The operations in there look almost like `pack()`.

> The current implementation of the multiscale approach for solids is buggy with respect to internal variables on element level 

Could you explain which problem is solved by this PR and how?

src/core/fem/src/general/element/4C_fem_general_element.hpp
georghammerl: Yes. Good catch. Thanks!

src/core/fem/src/general/element/4C_fem_general_element.hpp
```diff
@@ -215,6 +215,19 @@ namespace Core::Elements
...
+    */
+    virtual void pack_history(Core::Communication::PackBuffer& data) const;
+
+    /*!
+    \brief Unpack history data of this class to restore internal states
+
+    */
+    virtual void unpack_history(Core::Communication::UnpackBuffer& buffer);
+
+
```
sebproell: I am still against adding these methods here.

src/solid_3D_ele/4C_solid_3D_ele.cpp
```diff
@@ -251,6 +251,26 @@ void Discret::Elements::Solid::unpack(Core::Communication::UnpackBuffer& buffer)
...
 }
 
+void Discret::Elements::Solid::pack_history(Core::Communication::PackBuffer& data) const
+{
+  add_to_pack(data, unique_par_object_id());
+
+  //  material may hold history variables
+  Core::Elements::Element::pack_history(data);
+
+  Discret::Elements::pack(solid_calc_variant_, data);
```
sebproell: I think this call should be in the `Solid` namespace. I was confused what we are doing here until I looked it up in the code. /cc @amgebauer

src/core/fem/src/general/element/4C_fem_general_element.cpp
```diff
@@ -306,6 +306,45 @@ void Core::Elements::Element::unpack(Core::Communication::UnpackBuffer& buffer)
...
+  {
+    add_to_pack(data, true);
+    // pack only first material
+    mat_[0]->pack(data);
+  }
+  else
+  {
+    add_to_pack(data, false);
+  }
+}
```
sebproell: I think this method is basically:

`element->material()->pack()` with some additional checks. Why not simply call this function inside the `MicroMaterialGP` manager class?

georghammerl: Good point, this is basically `(solid_)material()->pack()`. I could call this from outside. Nevertheless, I need to pack the `solid_calc_variant_` which is not exposed outside the element (so far). The latest commit adapts the (un)pack_history in the solid element. To no longer pollute `fem_general_element`, I removed my changes there. Now, a cast to `Discret::Elements::Solid` is added in `MicroMaterialGP`. The latest commit still unions all history of the solid element in `pack_history`.

Alternatively, the `pack_history` in the solid element is changed to `pack_solid_calc_variant()`. Then, it is possible to call within `MicroMaterialGP` first `element->material()->pack()` and afterwards `element->pack_solid_calc_variant()` (with a proper cast). @sebproell Is this what you meant? Do you prefer such a solution?

src/solid_3D_ele/4C_solid_3D_ele.cpp
georghammerl: True, should be worked on in a separate PR.

src/core/fem/src/general/element/4C_fem_general_element.hpp
georghammerl: The latest commit removes those changes. Further alternative to the latest commit is proposed below.

src/core/fem/src/general/element/4C_fem_general_element.cpp
amgebauer: the solid-calc-variant is packed with the pack of the element, so it should be sufficient to pack and unpack the element and the solid material

src/solid_3D_ele/4C_solid_3D_ele.cpp
amgebauer: yes, the namespaces are not perfect in all places. I've hoped that we first collect all solid-related code into its own module.

src/core/fem/src/general/element/4C_fem_general_element.cpp
georghammerl: I am unsure whether this is appropriate because the solid element `unpack()` routine calls the base class `unpack()` which includes ` node_.resize(0);` . This destroys connectivity which I would like to retain. `FillComplete()` is not called during the multiscale simulation. 
 
 Hence, I favor to focus on what I need to restore during multiscale simulation which is exclusively history (and not `celltype`, `solid_ele_property` and all the base class stuff like `nodeids`, ...). Therefore, (un)packing of `solid_calc_variant_` and material is completely enough for the application.

georghammerl: Thanks to @amgebauer for finding a nice solution which now provides the `solid_calc_variant_` to `MicroMaterialgp` which does the (un)pack of the element evaluator and the material.

src/mat/4C_mat_micromaterialgp_static.hpp
```diff
@@ -108,6 +115,9 @@ namespace Mat
     /// my vector of new displacements
     std::shared_ptr<Core::LinAlg::Vector<double>> disn_;
 
+    /// history data for each column element on microscale
+    std::shared_ptr<std::map<int, std::shared_ptr<std::vector<char>>>> history_data_;
```
amgebauer: This type is very weird. 

```suggestion
    std::map<int, std::vector<char>> history_data_;
```

src/mat/4C_mat_micromaterialgp_static.cpp
```diff
@@ -354,12 +410,27 @@ void Mat::MicroMaterialGP::write_restart()
...
+  if (!history_data_->empty())
+  {
+    std::shared_ptr<Core::FE::Discretization> discret =
+        (Global::Problem::instance(microdisnum_))->get_dis("structure");
+
+    Core::Communication::PackBuffer data;
+    for (Core::Elements::Element* actele : discret->my_row_element_range())
+    {
+      add_to_pack(data, (*(*history_data_)[actele->id()]));
+    }
```
amgebauer: with my suggestion, this reduces to

```suggestion
    add_to_pack(data, history_data_);
```

src/mat/4C_mat_micromaterialgp_static.hpp
amgebauer: If you have many micro-elements, it might be better to use an unordered_map.

georghammerl: Good point. Thanks.

src/mat/4C_mat_micromaterialgp_static.cpp
georghammerl: History data during runtime is in element column layout, whereas restart output is written in element row layout. This way will likely lead to trouble.

PR #838, opened 2025-05-28
Unit test to trigger current bug in input spec matching with unused parameters

c-p-schmidt:
## Description and Context
Consider the following situation. You have a spec that only expects parameter `a` and a spec that expects the parameters `a` and `b` and you are reading in the parameters `a` and `b`, i.e. you want the latter spec to match. Currently, both specs are matched because the unused parameter is ignored at the moment.
@sebproell: I introduced a unit test for that behavior as discussed yesterday.


gilrrei:
@c-p-schmidt @sebproell: I thought about this a lot, as this is also a problem in the JSON schemas. The problem here (as most of the time) is the usage of `one_of`. In essence, most usages of `one_of` should be replaced with something different, either group or selection. The problem is that one_of makes the branches of the input anonymous, i.e., the different options have no names. The problem gets even more complicated if, in your testing example, you add a third one_of branch that also requires a and b, and a third optional parameter c. In this case, you can't make the parameter c optional, as otherwise it would match the second branch and third simultaneously.

This highlights the need to check for 'bad' (most of them emerged historically, so no blame here) for new PRs and not introduce them. The same holds for naming parameters, flat parameter setup (everything in one layer instead of nesting), anonymous constructs (i.e., one_of and all_of), implicit stuff (strings and maps with input that can't be checked further) or use of special stuff to trigger different behavior (-1 does something special for example). Of course, updating old input is an important step, which is just too big, so we should increase the quality checks for new input as a means to avoid introducing new 'bad' input.

> Note: Again, no blame to anybody, I am fully convinced that we all do the best we can. Introducing input quality tests just reminds us if we are not and how to improve it,  much the way every linter does.

sebproell:
After #858, this can now be implemented rather easily. I changed the matching logic to report better and nicer errors in case of unmatched data.

sebproell:
I know that this PR is not trivial since I changed a lot of the internals. So maybe read through the new and modified tests, and tell me if you agree with the new error messages.

src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -1991,16 +2017,34 @@ v: [1,2,3]
...
+                {
+                  [ ] Matched parameter 'theta'
+                }
+              [!] The following data remains unused:
+                Special: 
+            }
+          [!] The following data remains unused:
+            Special: 
+            OST:
+              theta: 0.5
```
maxiludwig, line 2039: Why is it not only the OST that is unused here? 
Above it's OST used and special unused. So here, it should be the other way round. Or am I missing something?

src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -2067,4 +2111,53 @@ b: 0.0)",
...
+    ryml::Tree tree = init_yaml_tree_with_exceptions();
+    ryml::parse_in_arena(R"(a: 1
+b: 2
+c: 3)",
+        &tree);
+    const ConstYamlNodeRef node(tree.rootref(), "");
+    InputParameterContainer container;
+    spec.match(node, container);
+    EXPECT_EQ(container.get<int>("a"), 1);
+    EXPECT_EQ(container.get<int>("b"), 2);
```
maxiludwig, line 2161:
```suggestion
    EXPECT_EQ(container.get<int>("b"), 2);
    EXPECT_EQ(container.get<int>("c"), 3);
```
Why not also check c?

src/core/io/tests/4C_io_input_spec_test.cpp
sebproell: This is definitely one of the aspects that could be further improved. @gilrrei would probably say by not using `one_of` ;) 

What you are seeing is the result of the failed `one_of` match. While `OST` matched, the `Special` was left over. and so the `one_of` could not be matched. This in turn makes the `TimeIntegration` group fail as well. Since the content was not used, the group again reports the content as unused, and this time it includes parts that were at least somewhat matching inside the one_of because it does not know that. I would argue that this is technically correct behavior, but it is probably not the best way to print in this case. I don't know a better way to handle this right now except with a special case for a single one_of in a group. Actually, I had such special cases before this PR and they failed to catch some corner cases (thus this PR). That's why I went for a more defensive check that errs on the side of caution and prints out unused data whenever it occurs.

src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -2094,8 +2094,8 @@ b: 0.0)",
       ConstYamlNodeRef node(root, "");
       InputParameterContainer container;
       FOUR_C_EXPECT_THROW_WITH_MESSAGE(spec.match(node, container), Core::Exception, R"({
-  [!] Candidate parameter 'a' (does not pass validation: in_range[0,50])
-  [!] Candidate parameter 'b' (does not pass validation: in_range(0,1.7976931348623157e+308])
+  [!] Candidate parameter 'a' does not pass validation: in_range[0,50]
+  [!] Candidate parameter 'b' does not pass validation: in_range(0,1.7976931348623157e+308]
```
sebproell, line 2098: @gilrrei Thanks for the hint in Slack. I fixed the hard-to-read ranges by removing the parentheses.

davidrudlstorfer (approved):

maxiludwig (approved):

Issue #836, opened 2025-05-27
Convert remaining .dat based input files to .yaml


davidrudlstorfer:
## Description

With #801 the majority of input files was converted from .dat to .yaml.

The following files remain:

### Tests that are currently converted into .yaml during testing (and both the .dat and .yaml files are tested). This way we can also test the converter capability

- [ ] ale2d_laplace_material.dat
- [ ] beam3r_herm2line3_static_beam_to_solid_surface_volume_coupling_mortar.dat
- [ ] particle_nointer_3d_pbc_xyz_initvel_velocityverlet.dat
- [ ] poro_solid_nitsche_contact3D_twoelem_wonopen_new_struc.dat
- [x] tsi_geonln_heatflux_flexoutsurf_monolithic.dat
- [ ] xfsi_3D_boxes.dat

### Tests with POST_ENSIGHT_STRUCTURE` currently fail because the dat reader of the post ensight tests is no longer working.
All addressed in  #847 

- [x] cardiovascular0d_4elementwindkessel_structure_direct_genalpha.dat
- [x] cardiovascular0d_4elementwindkessel_structure_direct_genalpha_new_struc.dat
- [x] cardiovascular0d_4elementwindkessel_structure_direct_ost.dat
- [x] cardiovascular0d_4elementwindkessel_structure_direct_ost_new_struc.dat
- [x] cardiovascular0d_4elementwindkessel_structure_direct_stat.dat
- [x] cardiovascular0d_4elementwindkessel_structure_direct_stat_new_struc.dat
- [x] cardiovascular0d_arterialproxdist_structure_direct_genalpha.dat
- [x] cardiovascular0d_arterialproxdist_structure_direct_genalpha_mor.dat
- [x] cardiovascular0d_arterialproxdist_structure_direct_genalpha_new_struc.dat
- [x] cardiovascular0d_syspulcirculation_0d_heart.dat
- [x] cardiovascular0d_syspulcirculation_2chamber_structure_statprestress.dat
- [x] cardiovascular0d_syspulcirculation_2chamber_structure_direct_genalpha.dat
- [x] cardiovascular0d_syspulcirculation_2chamber_structure_simple_genalpha.dat
- [x] cardiovascular0d_syspulcirculation_2chamber_structurecontactpenalty_direct_genalpha.dat
- [x] homogenized-constrained-mixture-cube-aniso-expl.dat
- [x] homogenized-constrained-mixture-cube-iso-expl.dat
[...]

PR #801, opened 2025-05-21
Convert test input files from .dat to .yaml

davidrudlstorfer:
## Description and Context

@sebproell as discussed today this PR converts the test input files to .4C.yaml

Tests which are not converted are the following:

 - ale2d_laplace_material.dat
 - beam3r_herm2line3_static_beam_to_solid_surface_volume_coupling_mortar
 - particle_nointer_3d_pbc_xyz_initvel_velocityverlet.dat
 - poro_solid_nitsche_contact3D_twoelem_wonopen_new_struc.dat
 - tsi_geonln_heatflux_flexoutsurf_monolithic.dat
 - xfsi_3D_boxes.dat
 
 These are now tested with .dat and the conversion to .yaml as well. This way the .dat input mechanism is still tested as well as the converter mechanism.

## Related Issues and Pull Requests

#598

PR #847, opened 2025-05-30
Remove POST_ENSIGHT_STRUCTURE flag

c-p-schmidt:
## Description and Context
This pull request removes the flag `POST_ENSIGHT_STRUCTURE` from the test suite. This is done as follows. For each "category" of input files listed in #836 for the  `POST_ENSIGHT_STRUCTURE`, there is one commit doing the following:
- Remove the `POST_ENSIGHT_STRUCTURE` from the `list_of_tests`
- Convert all `.dat` files to `.4C.yaml`
- Introduce one `vtk test` to still test the structure output mechanism for the respective "category" of tests

The last commit then removes the `POST_ENSIGHT_STRUCTURE` from CMake and deletes the respective Python utility files.

part of #836 
follows #801


c-p-schmidt:
I assigned @amgebauer as reviewer since mainly files related to his research are changed.

ischeider (dismissed):

amgebauer (dismissed):
I once added these tests since I broke the post-processor in my first year and @sbrandstaeter needed it back then, see https://gitlab.lrz.de/baci/baci/-/issues/504. @sbrandstaeter I think you don't rely on the post-processor anymore, right?

It is actually not necessary to add the runtime output tests since the output quantities that we test are just the default quantities.

sbrandstaeter:
> I once added these tests since I broke the post-processor in my first year and @sbrandstaeter needed it back then, see https://gitlab.lrz.de/baci/baci/-/issues/504. @sbrandstaeter I think you don't rely on the post-processor anymore, right?
> 
> It is actually not necessary to add the runtime output tests since the output quantities that we test are just the default quantities.

Thank you for thinking of me and bringing this old issue to mind.
You’re correct—I no longer use the post_processor.

c-p-schmidt:
> I once added these tests since I broke the post-processor in my first year and @sbrandstaeter needed it back then, see https://gitlab.lrz.de/baci/baci/-/issues/504. @sbrandstaeter I think you don't rely on the post-processor anymore, right?
> 
> It is actually not necessary to add the runtime output tests since the output quantities that we test are just the default quantities.

I tried to add as few reference files as possible (under the assumption that for any subproblem, one VTK test is desired). But I'm also fine with removing these tests as they only have very little value, I agree. So you would prefer not to have them?

amgebauer:
I think we don't need them

PR #850, opened 2025-06-01
Initialize norms in poro scatra to avoid UB

sebproell:
This allows us to convert the problematic case in #836.


davidrudlstorfer (approved):
Thanks for converting the file as well :)

c-p-schmidt (approved):

PR #852, opened 2025-06-01
Add validation of parsed values to InputSpecs

sebproell:
This PR adds a `.validator` option to `InputSpecBuilders::parameter`. This allows to write something like

```cpp
parameter<double>("time_step_size", {..., .validator = positive<double>()});
```

or

```cpp
parameter<int>("my_special_id", {..., .validator = positive<int>() || in_set({-100, -200})});
```

or

```cpp
parameter<double>("mass_fraction", {..., .validator = in_range(excl(0.), 1.0)});
```

If validation fails, the failed check can print itself to help in debugging the input. The checks are deliberately basic (but I think quite powerful already). We can discuss more involved checks if they are truly useful. The goal here is to just do all the stupid checks one tends to forget about; like for positive parameters (I guess this is true for >90% of our doubles :laughing:)

I kind of came up with this during #838 where it will help me to change the internals of the implementation a little more. Anyway, this feature was kind of already requested, so let me know what you think.

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -12,6 +12,7 @@
 
 #include "4C_io_input_parameter_container.templates.hpp"
 #include "4C_io_input_spec.hpp"
+#include "4C_io_input_spec_validators.hpp"
 #include "4C_io_input_types.hpp"
 #include "4C_io_value_parser.hpp"
 #include "4C_io_yaml.hpp"
@@ -538,6 +539,22 @@ namespace Core::IO
       return InputSpec(std::make_unique<InputSpecTypeErasedImplementation<std::decay_t<T>>>(
           std::forward<T>(wrapped), std::move(data)));
     }
+
+    //! Validate if possible. Returns false only if validation was attempted and failed.
+    template <typename T>
+    [[nodiscard]] bool validate_helper(
+        const T& val, const std::optional<InputSpecBuilders::Validators::Validator<T>>& validator)
+    {
+      return !validator.has_value() || (*validator)(val);
+    }
+
+    //! Validate if possible. Returns false only if validation was attempted and failed.
+    template <typename T>
+    [[nodiscard]] bool validate_helper(const std::optional<T>& val,
+        const std::optional<InputSpecBuilders::Validators::Validator<T>>& validator)
+    {
+      return !val.has_value() || validate_helper(*val, validator);
+    }
   }  // namespace Internal
 
   /**
@@ -625,6 +642,14 @@ namespace Core::IO
        * set additional values in the container.
        */
       ParameterCallback on_parse_callback{nullptr};
+
+      /**
+       * An optional validator that is called after the value has been parsed and is guaranteed to
+       * be of the correct type. This validator can then check the value for more specific
+       * conditions such as assuring that a value is within a certain range of values. See the
+       * Validators namespace for some available validators.
+       */
+      std::optional<Validators::Validator<RemoveOptional<T>>> validator{std::nullopt};
     };
 
     template <typename T>
@@ -638,6 +663,8 @@ namespace Core::IO
       DefaultType<T> default_value{};
 
       ParameterCallback on_parse_callback{nullptr};
+
+      std::optional<Validators::Validator<RemoveOptional<T>>> validator{std::nullopt};
 
       /**
        * The size of the vector. This can be a fixed size, #dynamic_size, or a callback that
@@ -657,6 +684,8 @@ namespace Core::IO
       DefaultType<T> default_value{};
 
       ParameterCallback on_parse_callback{nullptr};
+
+      std::optional<Validators::Validator<RemoveOptional<T>>> validator{std::nullopt};
 
       std::array<Size, rank<T>()> size;
     };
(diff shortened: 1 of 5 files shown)


src/core/io/src/4C_io_input_spec_validators.hpp
```diff
@@ -0,0 +1,305 @@
...
+   * // auto validator = in_range(0, 10.0);
+   * @endcode
+   *
+   * @note You can only use this function for types @p int and @p double.
+   *
+   */
+  template <typename Low, typename High,
+      typename T = typename Internal::LowHighCommonType<Low, High>::type>
+    requires((std::integral<T> && !std::same_as<T, bool>) || std::floating_point<T>)
+  [[nodiscard]] auto in_range(const Low& low, const High& high);
```
amgebauer, line 161: Why not define the operators `>`,  `>=`,  `<` and  `<=` as validators and then just use them directly and join them with `&&`? One then does not need the constructs with `incl` and `excl`, which in my eyes unnecessarily  complicate things.

```c++
const auto validator = in_range(incl(0), excl(2));
```
vs
```c++
const auto validator = greater_or_equal(0) && less_than(2);
```

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -625,6 +642,14 @@ namespace Core::IO
...
        */
       ParameterCallback on_parse_callback{nullptr};
+
+      /**
+       * An optional validator that is called after the value has been parsed and is guaranteed to
+       * be of the correct type. This validator can then check the value for more specific
+       * conditions such as assuring that a value is within a certain range of values. See the
+       * Validators namespace for some available validators.
+       */
+      std::optional<Validators::Validator<RemoveOptional<T>>> validator{std::nullopt};
```
amgebauer, line 652: To avoid the optional, you could also define a default `always_valid()`-validator. This might make some implementations easier.

src/core/io/src/4C_io_input_spec_validators.hpp
sebproell: I am open to changing this. However, this would likely complicate the creation of schema validators later. @gilrrei? 
A range is quite natural in schemas. We need to transform all the operator &&, ||, and ! into schema `allOf`, `anyOf`, `not`. But that is doable, and we would need it anyway if we want to combine ranges. So let's do it!

For your proposal, I think it would even make sense to go for the idiomatic `ge`, `lt`, `eq` abbreviations for the comparisons to get a terse notation.

sebproell: > I think it would even make sense to go for the idiomatic ge, lt, eq abbreviations for the comparisons to get a terse notation.

Nvm, @bgoderbauer convinced me that we should not abbreviate this. I will go for the explicit spelling.

src/core/io/src/4C_io_input_spec_builders.hpp
sebproell: At first, I thought this was a good idea. On second thought, I want to be able to discern if we have a validator or not, e.g., to print less noise to the metadata file. So I would keep it with the `std::optional` for now.

amgebauer (dismissed):

sebproell:
@bwirthl I annotated your nice magnet input to try out validation in practice. Please check if I got the validation right!

Example output if validation fails:

```
Could not match this input

FUNCT7:
  - SCATRA_FUNCTION:
      type: cylinder_magnet
      parameters:
        magnet_radius: 0.0
        magnet_length: 0.2
        magnetic_permeability: 1.25663706212
        magnet_position: [0.5,-0.2,0.0]
        magnet_magnetization: 5e3
        dynamic_viscosity_fluid: 1.0e-3
        rotation_around_x_axis: 90
        rotation_around_y_axis: 45
        particle_radius: 100e-6
      particle_magnetization_model:
        type: linear_with_saturation
        saturation_magnetization: 4.78e2
        susceptibility: 100.0


against the given input specification. This was the best attempt to match the input:

[!] Candidate list 'FUNCT7'
  [!] The following list entry did not match:
    [X] Expected exactly one of a few choices but matched:
      [!] Candidate group 'SCATRA_FUNCTION'
        [ ] Matched parameter 'type'
        [!] Candidate group 'parameters'
          [!] Candidate parameter 'magnet_radius' (does not pass validation: in_range(0,1.7976931348623157e+308])
          [ ] Matched parameter 'magnet_length'
          [ ] Matched parameter 'magnetic_permeability'
          [ ] Matched parameter 'magnet_magnetization'
          [ ] Matched parameter 'magnet_position'
          [ ] Matched parameter 'dynamic_viscosity_fluid'
          [ ] Matched parameter 'rotation_around_x_axis'
          [ ] Matched parameter 'rotation_around_y_axis'
          [ ] Matched parameter 'particle_radius'
        [ ] Matched selection 'particle_magnetization_model'
```

sebproell:
After a longer consultation with @gilrrei about which cases should be supported, we propose, for now, to go with a minimal set of useful validators: a range of numeric values, a set of enum values, or a positive value as a special case of the numeric range. We deliberately don't support string, because you should use strong types (usually enums) instead of strings for input. I also removed all the binary operators to create compound validators. If this is really necessary, we can add it back in any time, but we don't see the justification for this added complexity right now.

amgebauer (approved):

PR #853, opened 2025-06-02
Generalize `dot` and `ddot` product for arbitrarily sized tensors

amgebauer:
This PR extends the `dot` and `ddot` product for tensors of arbitrary size. Previously, only rank 1 and rank 2 tensors were possible. The tensor data is reinterpreted in a way so that normal matrix-multiplication routines can be used.

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_tensor.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_tensor.hpp
@@ -61,8 +61,54 @@ namespace Core::LinAlg
   using TensorView = TensorInternal<T, TensorStorageType::view, n...>;  // view onto's other
                                                                         // memory
 
-  // Implementation of tensor operations
+  /*!
+   * @brief Creates a TensorView from a pointer to data and the tensor shape
+   *
+   * This function creates a TensorView from a pointer to data and the tensor shape specified by the
+   * template parameters. The data is expected to be stored in a contiguous memory block of size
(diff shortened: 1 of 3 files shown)


src/core/linalg/tests/4C_linalg_tensor_operations_test.cpp
```diff
@@ -217,6 +217,132 @@ namespace
     EXPECT_EQ(A_ddot_B, 112.0);
   }
 
+  TYPED_TEST(TensorOperationsTest, ThreeTens_ddot_Mat)
```
sebproell: You are not supposed to use underscores in test name as per the Gtest docs. Also, why not `ThreeTensor`? I read this as three "10"s first.

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -61,8 +61,43 @@ namespace Core::LinAlg
   using TensorView = TensorInternal<T, TensorStorageType::view, n...>;  // view onto's other
                                                                         // memory
 
-  // Implementation of tensor operations
+  template <std::size_t... n>
+  constexpr auto make_tensor_view(auto* data)
```
sebproell, line 76: Docs?  I guess there is a convention that this is row-major layout? This is a user-facing function so this would be good to document here.

sebproell (approved):

PR #854, opened 2025-06-02
Implement printing of tensors

amgebauer:
Allows to print tensors to command line, for example during debugging.


src/core/linalg/tests/4C_linalg_tensor_test.cpp
```diff
@@ -198,6 +216,27 @@ namespace
...
+      ss << t1_view;
+      EXPECT_EQ(ss.str(), std::format("TensorView<{}, 2>[0, 0.1]", typeid(double).name()));
+    }
+
+    {
+      Core::LinAlg::Tensor<double, 2, 3> t2 = {{{0.0, 0.1, 0.2}, {1.0, 1.1, 1.2}}};
+      Core::LinAlg::TensorView<double, 2, 3> t2_view(t2);
+      std::stringstream ss;
+      ss << t2_view;
+      EXPECT_EQ(ss.str(), std::format("TensorView<{}, 2, 3>[\n [0, 0.1, 0.2],\n [1, 1.1, 1.2]\n]",
```
sebproell: Just a thought: I've started to use raw string literals to compare expected output, when the output has formatting like indentation and linebreaks. This makes it more obvious how the output looks.

PR #855, opened 2025-06-02
Implement tensor comparison in unit tests

amgebauer:
Allow comparison of tensors in unit tests


sebproell (approved):

amgebauer:
> Would it make sense to document that you only use the absolute tolerance in the comparison? Or optionally to also implement a relative tolerance and also call them atol and rtol?

We have to be consistent with `FOUR_C_EXPECT_NEAR` :grimacing:

PR #856, opened 2025-06-03
Increase number of test chunks for coverage report

amgebauer:
Our coverage report currently runs out of storage. This PR increases the number of test chunks from `60` to `90` to circumvent this problem.

image

https://github.com/4C-multiphysics/4C/actions/runs/15407249162

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           build-directory: ${{ github.workspace }}/build
           source-directory: ${{ github.workspace }}
-          number-of-chunks: 60 # we need fine grained test chunks to avoid running out of storage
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (approved):

amgebauer:
> Looks good to me. But one question. What is the motivation behind going from 60 directly to 90? Has this to do with the available runner capacity that increasing by 30 seems to make sense?

The job currently fails at test 70 of 77. At least, we need to increase the number of chunks by 10 %. I increased it by 50% to have a small buffer. The downside is that every job needs to setup the docker container which takes about 1 minute. For 30 additional jobs, that are 30 minutes in total (or 2 minutes in parallel with our 20 runners). I think that should not hurt us too much.

PR #858, opened 2025-06-03
Use all_of as an internal helper for group and one_of

sebproell:
While working on #838, I found that the implementation of InputSpec can be simplified by using `all_of` as an internal implementation detail in `group` and `one_of`.

Draft. I still need to add it inside `one_of`.


lauraengelhardt (approved):
Nice, looks good! :)

PR #864, opened 2025-06-05
Add C++ concepts for MPI communication

sebproell:
Use C++20 concepts to make it more obvious what we can communicate with the various MPI helpers. I also added a special path for enum communication, which can just use the underlying type in a native MPI call directly instead of going through pack and unpack.


src/core/comm/src/4C_comm_mpi_utils.hpp
```diff
@@ -238,22 +257,29 @@ namespace Core::Communication::Internal
...
+  // If T is an enum, we can broadcast a value of the underlying type.
+  if constexpr (std::is_enum_v<T>)
+  {
+    std::underlying_type_t<T> underlying = static_cast<std::underlying_type_t<T>>(value);
+    broadcast(&underlying, 1, root, comm);
+    value = static_cast<T>(underlying);
+  }
   // Shortcut when T is natively supported by MPI.
-  if constexpr (Internal::is_mpi_type<T>())
+  else if constexpr (Internal::is_mpi_type<T>())
```
georghammerl, line 282: Might it be an idea to switch the order to have this check first, followed by the `is_enum` check? I assume the `is_mpt_type` to be the more standard use case.

src/core/comm/src/4C_comm_mpi_utils.hpp
```diff
@@ -27,6 +27,38 @@ FOUR_C_NAMESPACE_OPEN
...
+  concept IsNativeMpiType = Internal::is_mpi_type<T>();
+
+  /**
+   * Concept to check whether we know of a way to communicate a type over MPI. This is obviously
+   * true for all native MPI types. It is also true for enums, as they can be communicated as
+   * their underlying type. For user-defined types, we require that they can be packed and unpacked.
+   */
+  template <typename T>
+  concept IsCommunicatable =
+      IsNativeMpiType<T> || std::is_enum_v<T> || Packable<T> || Unpackable<T>;
```
amgebauer:
```suggestion
      IsNativeMpiType<T> || std::is_enum_v<T> || (Packable<T> && Unpackable<T>);
```
?

georghammerl:
Very nice changes. Thank you @sebproell

src/core/comm/src/4C_comm_mpi_utils.hpp
sebproell: This is a constexpr check evaluated at compile time, so this branch will not show up in the machine code.

PR #866, opened 2025-06-05
InputField class to supply element-wise parameters

sebproell:
This PR is a first step to generalize existing functionality for element-wise parameter fields with a more usable interface. See #786.

@lauraengelhardt I applied it to the two cases where you read a pattern file. This does not make a lot of sense yet. The parameter will become a lot nicer when @maxiludwig implements a better `InputSpec` for this type of parameter.

### DIFF
--- a/src/core/io/src/4C_io_input_field.hpp
+++ b/src/core/io/src/4C_io_input_field.hpp
@@ -1,0 +1,89 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_IO_INPUT_FIELD_HPP
+#define FOUR_C_IO_INPUT_FIELD_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_utils_exceptions.hpp"
+#include "4C_utils_std23_unreachable.hpp"
+
+#include <unordered_map>
+#include <variant>
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Core::IO
+{
+  /**
+   * @brief A class to represent an input parameter field.
+   *
+   * In its current form, this class can either hold a single value of type T or a map of
+   * element-wise values of type T.
+   */
+  template <typename T>
+  class InputField
+  {
+   public:
+    using IndexType = int;
+    using StorageType = std::variant<T, std::unordered_map<IndexType, T>>;
+
+    /**
+     * Construct an InputField from a single @p const_data. This field will have the same value
+     * for every element index.
+     */
+    explicit InputField(T const_data) : data_(std::move(const_data)) {}
+
+    /**
+     * Construct an InputField from a map of element-wise data. The @p data map contains
+     * element indices as keys and the corresponding values of type T.
+     */
+    explicit InputField(std::unordered_map<IndexType, T> data) : data_(data) {}
+
+    /**
+     * Access the value of the field for the given @p element index. The @p element_id
+     * is not checked for validity and asking for an invalid index will lead to undefined behavior.
+     * Use the `at()` function if you want to check for validity.
+     */
+    [[nodiscard]] const T& operator[](IndexType element_id) const { return get(element_id, false); }
+
+    /**
+     * Access the value of the field for the given @p element index. If the @p element_id
+     * is not a valid index, this function will throw an error.
+     */
+    [[nodiscard]] const T& at(IndexType element_id) const { return get(element_id, true); }
+
+   private:
+    //! Internal getter which can optionally check for the validity of the element index.
+    const T& get(IndexType element_id, bool check) const
+    {
+      if (std::holds_alternative<T>(data_))
+      {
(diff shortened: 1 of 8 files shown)


bgoderbauer (approved):

sebproell:
@maxfirmbach I am only aware of these two. And yes, making this feature more easily available and flexible is our goal. @maxiludwig is working on  it; this PR is just a small preparatory step.

src/mat/4C_mat_muscle_utils.hpp
```diff
@@ -276,22 +277,22 @@ namespace Mat::Utils::Muscle
...
+   * activation_field for a given element id and time. Evaluating the activation_field at the
+   * element id (key) returns a vector of pairs (value). Those vector entries correspond to a
+   * time-"activation value"-pair. The activation at the current time t_current is found via linear
+   * interpolation between the prescribed time-"activation value"-pairs.
    *
    * The time-dependent optimal active stress is obtained by sigma_opt = sigma_max * ft
    *
    * @param[in]     sigma_max Optimal (i.e. maximal) active stress
-   * @param[in]     activation_map Map to be evaluated
+   * @param[in]     activation_field field containing time-activation pairs for each element id
```
lauraengelhardt:
```suggestion
   * @param[in]     activation_field field containing a vector of time-activation pairs for each element id
```

src/mat/4C_mat_muscle_utils.cpp
```diff
@@ -381,17 +382,13 @@ double Mat::Utils::Muscle::evaluate_time_space_dependent_active_stress_by_funct(
 
 double Mat::Utils::Muscle::evaluate_time_space_dependent_active_stress_by_map(
     const double sigma_max,
-    const std::unordered_map<int, std::vector<std::pair<double, double>>>& activation_map,
+    const Core::IO::InputField<std::vector<std::pair<double, double>>>& activation_field,
     const double t_current, const int activation_map_key)
```
lauraengelhardt:
```suggestion
    const double t_current, const int element_id)
```

src/mat/4C_mat_muscle_utils.cpp
```diff
@@ -381,17 +382,13 @@ double Mat::Utils::Muscle::evaluate_time_space_dependent_active_stress_by_funct(
...
   // compute time-dependency ft
-  auto it = activation_map.find(activation_map_key);
 
-  if (it == activation_map.end())
-  {
-    FOUR_C_THROW("Key (element id) {} not found in activation map.", activation_map_key);
-  }
-  const double ft = lineraly_interpolate_between_times(it->second, t_current);
+  const auto& times = activation_field.at(activation_map_key);
+  const double ft = lineraly_interpolate_between_times(times, t_current);
```
lauraengelhardt:
```suggestion
  const auto& time_activation_value_pairs = activation_field.at(element_id);
  const double ft = lineraly_interpolate_between_times(time_activation_value_pairs, t_current);
```

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Core::IO
+{
+  /**
+   * @brief A class to represent an input parameter field.
+   *
+   * In its current form, this class can either hold a single value of type T or a map of
+   * element-wise values.
```
lauraengelhardt:
```suggestion
   * In its current form, this class can either hold a single value of type T or a map of
   * element-wise defined values of type T. 
```

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Access the value of the field for the given @p element index. The @p element_index
+     * is not checked for validity and asking for an invalid index will lead to undefined behavior.
+     * Use the `at()` function if you want to check for validity.
+     */
+    [[nodiscard]] const T& operator[](IndexType element_index) const
+    {
+      return get(element_index, false);
+    }
```
lauraengelhardt:
```suggestion
    /**
     * Access the value of the field for the given @p element id. The @p element_id
     * is not checked for validity and asking for an invalid id will lead to undefined behavior.
     * Use the `at()` function if you want to check for validity.
     */
    [[nodiscard]] const T& operator[](IndexType element_id) const
    {
      return get(element_id, false);
    }
```

lauraengelhardt: And below... ;)

lauraengelhardt: When I think about it, the key doesn't necessarily have to be an element ID. It could just as well be a node ID or any other identifier, depending on how the class is used within, e.g., your material implementation. If you decide your `pattern` file should contain xyz coordinates as keys and you access the `InputField` by these in your material routine, this would also work. So... should it just be the `key`?

lauraengelhardt (changes requested):

src/core/io/src/4C_io_input_field.hpp
sebproell: As discussed: we should differentiate between element-wise and nodal fields in a type-safe manner. This will follow soon.

lauraengelhardt (approved):

PR #868, opened 2025-06-05
Remove various #defines of constants

sebproell:
There are more #defines of constants left, but most of them are inside `_defines.hpp` files, which we should remove in the context of #207.

### DIFF
--- a/src/mat/elast/4C_mat_elast_aniso_structuraltensor_strategy.cpp
+++ b/src/mat/elast/4C_mat_elast_aniso_structuraltensor_strategy.cpp
@@ -15,6 +15,11 @@
 
 FOUR_C_NAMESPACE_OPEN
 
+namespace
+{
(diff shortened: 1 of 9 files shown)


src/mat/elast/4C_mat_elast_aniso_structuraltensor_strategy.cpp
```diff
@@ -15,6 +15,11 @@
 
 FOUR_C_NAMESPACE_OPEN
 
+namespace
+{
+  constexpr int numbgp = 50;
+  constexpr int twice = 100;
```
sebproell, line 21: @amgebauer Do you know what these refer to?

amgebauer: no, I don't know :grimacing:

amgebauer (approved):

PR #862, opened 2025-06-04
Implement `SymmetricTensor`

amgebauer:
This PR introduces the implementation of symmetric tensors including benchmark tests and unit tests. The interface is the same as for normal `Tensor`s.

```cpp
SymmetricTensor<double, 3, 3> tensor{}; // tensor filled with zeros
//SymmetricTensor<double, 3, 4> tensor{}; // ill formed

Tensor<double, 3> t = {{1.0, 2.0, 3.0}};

SymmetricTensor<double, 3, 3> t_dyadic_t = self_dyadic(t); // internal memory storage like our stress-like Voigt notation (like PK2-stress)
SymmetricTensor<double, 3, 3, 3, 3> t_dyadic_t _dyadic_t _dyadic_t = self_dyadic<4>(t); // internal storage like our linearization of a stress-like Voigt vector w.r.t. a strain-like Voigt vector (like our cmat-matrix)
```

The internal data structure is compatible to the stress-like Voigt notation, so that we can create a `SymmetricTensorView` onto an existing stress-like Voigt implementation allowing us to migrate materials step by step.

Most computations are done by computing the full tensor internally and then applying the normal tensor operations and then, if applicable, discard the duplicate computations. The compiler is able to optimize this away for most operations. Large tensors are problematic, so I added an optimization for the `ddot(symmetric 4-tensor, symmetric 2-tensor)`-product. Here are the results:

```
symmetric_tensor_det<2>                             0.318 ns        0.318 ns   2205296314
symmetric_tensor_det<3>                              1.06 ns         1.06 ns    664999210
symmetric_tensor_det<4>                               190 ns          189 ns      3655369
tensor_det<2>                                       0.315 ns        0.315 ns   2202936520
tensor_det<3>                                        1.07 ns         1.07 ns    648854595
tensor_det<4>                                         187 ns          187 ns      3699387
-> equal

symmetric_tensor_inv<2>                             0.870 ns        0.870 ns    794753744
symmetric_tensor_inv<3>                              3.14 ns         3.14 ns    221724024
tensor_inv<2>                                       0.942 ns        0.942 ns    739169915
tensor_inv<3>                                        3.47 ns         3.46 ns    215795189
-> faster

symmetric_matrix_dot_vector<2>                      0.509 ns        0.509 ns   1344275118
symmetric_matrix_dot_vector<3>                       1.14 ns         1.14 ns    610003920
symmetric_matrix_dot_vector<4>                       2.06 ns         2.06 ns    338300377
[...]

PR #869, opened 2025-06-06
Use symmetric tensors in solid element evaluation

amgebauer:
This PR uses the symmetric tensors in the solid element evaluation routine. Draft for now since the changes from #862 are required.

### DIFF
--- a/src/solid_3D_ele/4C_solid_3D_ele_calc_lib.hpp
+++ b/src/solid_3D_ele/4C_solid_3D_ele_calc_lib.hpp
@@ -24,9 +24,13 @@
 #include "4C_fem_nurbs_discretization_utils.hpp"
 #include "4C_global_data.hpp"
 #include "4C_inpar_structure.hpp"
-#include "4C_linalg_fixedsizematrix_generators.hpp"
+#include "4C_linalg_fixedsizematrix.hpp"
 #include "4C_linalg_fixedsizematrix_solver.hpp"
-#include "4C_linalg_utils_densematrix_eigen.hpp"
+#include "4C_linalg_symmetric_tensor.hpp"
+#include "4C_linalg_symmetric_tensor_eigen.hpp"
+#include "4C_linalg_tensor.hpp"
+#include "4C_linalg_tensor_generators.hpp"
+#include "4C_linalg_tensor_matrix_conversion.hpp"
 #include "4C_linalg_vector.hpp"
 #include "4C_mat_so3_material.hpp"
 #include "4C_solid_3D_ele_calc_lib_integration.hpp"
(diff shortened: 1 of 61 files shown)


src/solid_3D_ele/4C_solid_3D_ele_calc_lib.hpp
```diff
@@ -636,19 +634,10 @@ namespace Discret::Elements
...
+      const Core::LinAlg::SymmetricTensor<double, 3, 3>& cauchygreen)
   {
-    Core::LinAlg::Matrix<6, 1> gl_strain;
-
-    gl_strain(0) = 0.5 * (cauchygreen(0, 0) - 1.0);
-    gl_strain(1) = 0.5 * (cauchygreen(1, 1) - 1.0);
-    gl_strain(2) = 0.5 * (cauchygreen(2, 2) - 1.0);
-    gl_strain(3) = cauchygreen(0, 1);
-    gl_strain(4) = cauchygreen(1, 2);
-    gl_strain(5) = cauchygreen(2, 0);
```
davidrudlstorfer, line 649: Just out of interest :D

Why do we not have the `0.5*` in the old version here?

davidrudlstorfer (approved):
Looks good as far as I can judge :D

And again, a general question out of interest. Are these changes of this PR mainly for performance improvements when the symmetric tensors are used? From a code perspective only a handful of places simplified drastically with the new tensors in comparison to the "old" approach. (Although the change of lines is a bad indicator for stuff like this I initially suspected that with tensors the lines of code will drastically decrease and not increase)

Thanks in advance:)

src/solid_3D_ele/4C_solid_3D_ele_calc_lib.hpp
amgebauer: It's the strain like Voigt representation, where the off diagonal entries are multiplied by 2 (so that a dot-product with a stress-like Voigt vector is the same as a ddot product between two symmetric tensors). Lucky you that you never needed to deal with that :grimacing: 

This is the motivation why I introduced the (symmetric) tensors, where all these confusing conversions are unnecessary. This is actually a nice example on how this improves how we write our Gauss point equations:

```c++
    Core::LinAlg::Matrix<6, 1> gl_strain;
    gl_strain(0) = 0.5 * (cauchygreen(0, 0) - 1.0);
    gl_strain(1) = 0.5 * (cauchygreen(1, 1) - 1.0);
    gl_strain(2) = 0.5 * (cauchygreen(2, 2) - 1.0);
    gl_strain(3) = cauchygreen(0, 1);
    gl_strain(4) = cauchygreen(1, 2);
    gl_strain(5) = cauchygreen(2, 0);
    retrun gl_strain;
```
becomes
```c++
    return 0.5 * (cauchygreen - Core::LinAlg::TensorGenerators::identity<double, 3, 3>);
```

davidrudlstorfer: Thanks for the explanation :)

amgebauer:
> And again, a general question out of interest. Are these changes of this PR mainly for performance improvements when the symmetric tensors are used? From a code perspective only a handful of places simplified drastically with the new tensors in comparison to the "old" approach. (Although the change of lines is a bad indicator for stuff like this I initially suspected that with tensors the lines of code will drastically decrease and not increase)

The tensors will simplify the equations on the Gauss point level. Since some interfaces (e.g. the material interface) still relies on the `LinAlg::Matrix`, we have some overhead to create Matrices from tensors. The overhead will vanish once the material interface is based on Tensors as well.

Note also, that rewriting all equations in Tensor-form is not always straight forward and often require a lot of knowledge of the underlying physics to extract the necessary equations from the matrix-implementation. See for example the solid-poro implementation that I mainly left untouched in Matrix-notation.

lkoeglmeier (approved):

PR #870, opened 2025-06-06
Extend some error message while reading exodus geometry inpt

ischeider:
## Description and Context

Just a small modification of error messages, which are thrown if the reading of exodus II does not succeed. I hope that you'll never see them, but the exodus output of the original meshio python package might throw some of them ;-)

In addition, I added some output showing information about the mesh read in. Since the binary file is not readable directly, I find it useful, particularly since the block and nset numbers are given, which have to be used in boundary conditions and material relations. If you think it's too much output, I can delete it, of course.

Maybe it's related to the discussion #840


src/core/io/src/4C_io_exodus.cpp
```diff
@@ -129,14 +129,19 @@ Core::IO::Exodus::Mesh::Mesh(std::filesystem::path exodus_file, MeshParameters m
...
+            << num_node_sets << " node sets, " << num_side_sets << " side sets" << std::endl;
 
   // get nodal coordinates
   {
     std::vector<double> x(num_nodes);
     std::vector<double> y(num_nodes);
     std::vector<double> z(num_nodes);
     error = ex_get_coord(exo_handle, x.data(), y.data(), z.data());
-    if (error != 0) FOUR_C_THROW("exo error returned");
+    if (error != 0) FOUR_C_THROW("error reading exodus node coordinates");
```
maxfirmbach: I would use upper case at the beginning of the message, like `Error reading ...` and end the message with a `.`.

src/core/io/src/4C_io_exodus.cpp
```diff
@@ -129,14 +129,19 @@ Core::IO::Exodus::Mesh::Mesh(std::filesystem::path exodus_file, MeshParameters m
   error = ex_get_init(exo_handle, title, &spatial_dimension_, &num_nodes, &num_elem_, &num_elem_blk,
       &num_node_sets, &num_side_sets);
   title_ = std::string(title);
+  std::cout << "Start reading exodus file " << exodus_file.string() << " entitled <<" << title_
```
maxfirmbach: I like the additional ouput ... but not every time :wink:. Would it make sense to put this under some verbosity? E.g. `Core::IO::cout(Core::IO::verbose)`?

src/core/io/src/4C_io_exodus.cpp
```diff
@@ -215,6 +220,8 @@ Core::IO::Exodus::Mesh::Mesh(std::filesystem::path exodus_file, MeshParameters m
       error = ex_get_name(exo_handle, EX_NODE_SET, npropID[i], mychar);
       // prefer std::string to store name
       std::string nodesetname(mychar);
+      std::cout << "NodeSet " << i << ": name=" << nodesetname << "," << num_nodes_in_set
```
maxfirmbach: Same comment as above.

src/core/io/src/4C_io_exodus.cpp
ischeider: I changed all my occurrences as requested, but I must admit that there seems to be no rule for error output (all variants of output can be found ;-) )

src/core/io/src/4C_io_exodus.cpp
ischeider: I added the new parameter `SHOWINFO` in the `section_name + GEOMETRY` section, which is `false` by default. Output is only given, if it is set to true now. In some of the tutorialss, where the exodus files are read, I set this parameter to have the effect tested.

src/core/io/src/4C_io_exodus.cpp
ischeider: see below...

src/global_legacy_module/4C_global_legacy_module_validparameters.cpp
```diff
@@ -150,6 +150,11 @@ std::map<std::string, Core::IO::InputSpec> Global::valid_parameters()
             parameter<std::filesystem::path>(
                 "FILE", {.description = "Path to the exodus geometry file. Either absolute or "
                                         "relative to the input file."}),
+            parameter<bool>("SHOWINFO",
```
sebproell:
```suggestion
            parameter<bool>("SHOW_INFO",
```

ischeider:
> Printing some stats of the mesh is fine, but I prefer to do this in an isolated manner after reading. There are already some print functions in the ElementBlock, NodeSet, ... classes. Can you use the print function of the mesh itself and call this function in case `SHOW_INFO` is set? Feel free to modify it; I just copied the very old version of it. Then you don't need to interject the print outs into the reading process.

Now, the PR is rather about the exodus mesh output than about the error messages, but I hope the changes are still valid in the context of the described topic.

src/core/io/src/4C_io_exodus.cpp
```diff
@@ -278,43 +282,61 @@ Core::IO::Exodus::Mesh::Mesh(std::filesystem::path exodus_file, MeshParameters m
...
-  if (error < 0) FOUR_C_THROW("error while closing exodus II file");
+  if (error < 0) FOUR_C_THROW("Error while closing exodus II file.");
 }
 
 
 /*----------------------------------------------------------------------*
  |  Print method (public)                                      maf 12/07|
  *----------------------------------------------------------------------*/
-void Core::IO::Exodus::Mesh::print(std::ostream& os, bool verbose) const
+void Core::IO::Exodus::Mesh::print(std::ostream& os) const
```
sebproell: Why isn't the verbosity level a parameter of this function as well? Seems inconsistent with the other functions. Please make it a parameter here and remove the member variable.

src/core/io/src/4C_io_exodus.hpp
```diff
@@ -20,6 +20,27 @@ FOUR_C_NAMESPACE_OPEN
...
+     *         1: output of summary for blocks and sets,
+     *         2: output of summary for each block and set,
+     *         3: detailed output for each block and set
+     *         4: detailed output, even for nodes and elements
+     */
+    none = 0,
+    summary = 1,
+    detailedsummary = 2,
+    detailed = 3,
+    full = 4
```
sebproell:
```suggestion
    /**
     *  level (0: no output,
     *         1: output of summary for blocks and sets,
     *         2: output of summary for each block and set,
     *         3: detailed output for each block and set
     *         4: detailed output, even for nodes and elements
     */
    none = 0,
    summary = 1,
    detailed_summary = 2,
    detailed = 3,
    full = 4
```

Also you can move the description of every value above that value. Would just `none`, `summary` and `full` be enough? I doubt we need all these options.

ischeider: I changed the variable name. However, I have the feeling that sometimes `detailed_summary`might be useful to see which elements belong to a block without having the very detailed `full` output, so I would like to keep the 5 levels (and it does not hurt).

sebproell: @ischeider My experience is that every line of code hurts; that's why I would like to only write the ones we actually use. I am fine with your assessment, though.

PR #872, opened 2025-06-07
Add benchmark tests for 3D solid elements

amgebauer:
This PR adds a simple benchmark test for evaluating a solid element with a simple material.


tests/benchmark_tests/solid_3D_ele/4C_solid_3D_ele_evaluation_benchmark.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include <cstdlib>
+#include <memory>
+#include <mutex>
+#include <vector>
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace
+{
+  std::once_flag setup_global_instance;
```
sebproell: You could move the flag next to `call_once` and make it static.

CMakeLists.txt
```diff
@@ -195,6 +195,7 @@ add_subdirectory(doc EXCLUDE_FROM_ALL)
 
 # Tests
 add_subdirectory(unittests)
+add_subdirectory(tests/benchmark_tests)
```
sebproell, line 198: How about moving the benchmark tests into the `solid_3D_ele` right away? In every module we can have `src`, `tests` and `benchmarks` .

sebproell (commented):
Nice! Can you show the output of the benchmark?

tests/benchmark_tests/solid_3D_ele/4C_solid_3D_ele_evaluation_benchmark.cpp
amgebauer: I think there is a problem. If I add a second benchmark that runs in the same executable, then we set up the global problem twice (which does not work). What about adding a `GlobalProblemScopeGuard` that removes all instances once it goes out of scope?

CMakeLists.txt
amgebauer: I thought about that, but I think this is a separate task since we need to move a lot of other modules into the solid module (that does not yet exist) first.

amgebauer:
Here are the current benchmark results with GCC on my machine:
```
Running ./tests/benchmarktests_solid_3D_ele
Run on (24 X 5000 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x12)
  L1 Instruction 32 KiB (x12)
  L2 Unified 1280 KiB (x12)
  L3 Unified 30720 KiB (x1)
Load Average: 6.39, 18.66, 14.42
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
---------------------------------------------------------------------------------------------------------------------------
Benchmark                                                                                 Time             CPU   Iterations
---------------------------------------------------------------------------------------------------------------------------
evaluate_force_stiff<CellType::hex8>                                                   9311 ns         9309 ns        75896
evaluate_force_stiff<CellType::hex20>                                                128379 ns       128368 ns         5395
evaluate_force_stiff<CellType::hex27>                                                221393 ns       221387 ns         3186
evaluate_force_stiff<CellType::tet4>                                                    876 ns          876 ns       798649
evaluate_force_stiff<CellType::tet10>                                                  7021 ns         7021 ns       100082
evaluate_force_stiff<CellType::wedge6>                                                 4550 ns         4550 ns       155230
evaluate_force_stiff<CellType::pyramid5>                                               4632 ns         4632 ns       151750
evaluate_force_stiff<CellType::hex8, ElementTechnology::none, KinemType::linear>       8341 ns         8340 ns        84093
evaluate_force_stiff<CellType::hex8, ElementTechnology::eas_mild>                     25410 ns        25409 ns        27316
evaluate_force_stiff<CellType::hex8, ElementTechnology::eas_full>                     40957 ns        40956 ns        17106
evaluate_force_stiff<CellType::hex8, ElementTechnology::fbar>                         11685 ns        11684 ns        59619
```

tests/benchmark_tests/solid_3D_ele/4C_solid_3D_ele_evaluation_benchmark.cpp
sebproell: `SingletonOwnerRegistry::ScopeGuard` should do what you need.

tests/benchmark_tests/solid_3D_ele/4C_solid_3D_ele_evaluation_benchmark.cpp
```diff
@@ -0,0 +1,203 @@
...
+    container.add("KINEM", kinem_type);
+    container.add("TECH", element_technology);
+
+    ele->read_element(ele_type, Core::FE::cell_type_to_string(celltype), container);
+
+    return ele;
+  }
+
+  template <Core::FE::CellType celltype>
+  std::unique_ptr<Core::FE::Discretization> make_discretization(
```
sebproell, line 101: We have such helpers in a few places already and should probably merge them somewhere. The deal.II tests and rebalancing tests come to mind, but there are probably more.

sebproell (approved):

PR #873, opened 2025-06-07
InputSpec can store parsed values to structs

sebproell:
When you want to create user-configurable input for an algorithmic component, you ideally encoded the parameters you need in a struct that gets passed to this algorithmic component. This PR enables a connection between specs and statically-typed storage locations inside structs, e.g.

```cpp
   struct Inner
    {
      int a;
      std::string s;
    };

    struct Outer
    {
      double d;
      Inner inner;
    };

    auto spec = group_struct<Outer>(
        "outer", {
                     parameter<double>("d", {.store = in_struct(&Outer::d)}),
                     group_struct<Inner>("inner",
                         {
                             parameter<int>("a", {.store = in_struct(&Inner::a)}),
                             parameter<std::string>("s", {.store = in_struct(&Inner::s)}),
                         },
                         {.store = in_struct(&Outer::inner)}),
                 });
```

When reading in input, we can now call `container.get<Outer>("outer")` and will receive an object of type `Outer`, which will contain the parsed input data in a _statically typed manner_. So basically we can exchange

```cpp
const int& a = container.group("outer").group("inner").get<int>("a"):
```

with 

```cpp
const int& a = container.get<Outer>("outer").inner.a;
```

Note that this is much more impactful than you may think at first glance. It is not just a change in style but a shift from dynamic, string-based lookups, which are run-time checked, to a static world, where only the first access to "outer" is a run-time check. 

#### Open for later PRs

I also want to enable this for `one_of` and `selection`. These will need to store a `variant` with a few options.

### DIFF
--- a/src/core/io/src/4C_io_input_file.cpp
+++ b/src/core/io/src/4C_io_input_file.cpp
@@ -844,7 +844,9 @@ namespace Core::IO
     {
       if (spec.impl().has_default_value())
       {
-        spec.impl().set_default_value(container);
+        // Matching against an empty node will set the default values.
+        auto tree = init_yaml_tree_with_exceptions();
+        spec.match(ConstYamlNodeRef(tree.rootref(), ""), container);
         return;
       }
       else if (spec.impl().required())
--- a/src/core/io/src/4C_io_input_spec_builders.cpp
+++ b/src/core/io/src/4C_io_input_spec_builders.cpp
@@ -122,7 +122,10 @@ namespace
       }
       else if (entry->impl().has_default_value())
       {
-        entry->impl().set_default_value(container);
+        Core::IO::InputSpecBuilders::Storage storage;
+        Core::IO::Internal::init_storage_with_container(storage);
+        entry->impl().set_default_value(storage);
+        container.merge(std::any_cast<Core::IO::InputParameterContainer&&>(std::move(storage)));
       }
     }
   }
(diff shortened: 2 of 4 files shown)


sebproell:
This is ready for review. I am especially interested in feedback on naming for

- `group_struct`: indicates that this is also a group but with the additional struct feature.
- the `.store` option: could also be something like `.bind`
- `in_struct`: could also be anything in this pattern `(in|to)_(struct|member|field)`. Using `struct` again suggests that it is related to `group_struct`.

c-p-schmidt:
Thank you for this nice feature! When looking at your nice example, I was a bit confused that
`{.store = in_struct(&struct_name::component_name)}` 
is defined for each component of a struct, which actually seems to double the definition within the respective group_struct itself. Is this necessary for technical reasons?

sebproell:
@c-p-schmidt This is a pointer-to-member. Cppref:

> A pointer to non-static member object m which is a member of class C can be initialized with the expression &C::m exactly. 

If we want to use it, we need to write exactly this syntax. Why do we need to use it?
We build up the specs from the bottom up. When you define a parameter, you need to say which struct member you want to store it in, and that is precisely what the pointer-to-member does. We need to take care of two potential problems:

- You will need to provide storage for that struct. This is done by group_struct. If the type is not consistent, or you leave it out. you will get an error as shown in the tests.
- Mixing pointers to members of different structs does not work. We also check for that. 

Why does this not happen when writing a struct in C++ code? In that case, the members of the struct receive the struct context from the top down. There cannot be any inconsistency.

One possible improvement would be that the group_struct could infer the type of the struct from its contained specs. However, the current implementation erases the exact type of spec rather early. Maybe this will change in the future.

In C++26, the answer to all this is going to be reflection: https://isocpp.org/files/papers/P2996R4.html

c-p-schmidt (approved):

PR #874, opened 2025-06-07
[LinAlg] Refactor vector sum and replace value methods

maxfirmbach:
## Description and Context

This PR aims to cleanup the vector interface in regards to methods replacing and summing values in the vector:
- Deleting unused methods
- Hard coding the vector index to 0 (as we have a single vector) and simplifying how we call the method from the wrapper
- Refactoring and simplifying a few calls in the physics modules 

Let's see this as preparation step for the switch to `Epetra_FEVector`.

### DIFF
--- a/src/core/comm/tests/4C_comm_utils_test.np3.cpp
+++ b/src/core/comm/tests/4C_comm_utils_test.np3.cpp
@@ -57,10 +57,11 @@ namespace
       int* indices = new int[numMyEles];
       for (int lid = 0; lid < numMyEles; ++lid)
       {
-        indices[lid] = lid;
-        values[lid] = map->gid(lid);
+        const int gid = map->gid(lid);
+        indices[lid] = gid;
+        values[lid] = gid;
       }
-      vector_->replace_local_values(numMyEles, values, indices);
+      vector_->replace_global_values(numMyEles, values, indices);
     }
 
     void TearDown() override { Core::IO::cout.close(); }
--- a/src/fluid/4C_fluid_timint_hdg.cpp
+++ b/src/fluid/4C_fluid_timint_hdg.cpp
@@ -350,7 +350,6 @@ void FLD::TimIntHDG::set_initial_flow_field(
   else
   {
     const Core::LinAlg::Map* dofrowmap = discret_->dof_row_map();
-    const Core::LinAlg::Map* intdofrowmap = discret_->dof_row_map(1);
     Core::LinAlg::SerialDenseVector elevec1, elevec2, elevec3;
     Core::LinAlg::SerialDenseMatrix elemat1, elemat2;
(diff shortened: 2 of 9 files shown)


sebproell (approved):

src/core/linalg/src/sparse/4C_linalg_vector.hpp
```diff
@@ -212,24 +212,20 @@ namespace Core::LinAlg
...
     {
       return vector_->Multiply(ScalarAB, A, B, ScalarThis);
     }
 
-    int replace_global_values(int NumEntries, const double* Values, const int* Indices)
-    {
-      return vector_->ReplaceGlobalValues(NumEntries, Values, Indices);
-    }
+    //! Puts element-wise reciprocal values of input Multi-vector in target.
+    int reciprocal(const Epetra_MultiVector& A) { return vector_->Reciprocal(A); }
```
mayrmt, line 223: This public method asks for an `Epetra_MultiVector`, so it kind of exposes the `Epetra_MultiVector` wrapped in this class to the outside world. Shouldn't it take a `LINALG::Vector` as input?

src/core/linalg/src/sparse/4C_linalg_vector.hpp
```diff
@@ -212,24 +212,20 @@ namespace Core::LinAlg
...
 
-    int replace_global_values(int NumEntries, const double* Values, const long long* Indices)
+    int reciprocal_multiply(double ScalarAB, const Epetra_MultiVector& A,
+        const Epetra_MultiVector& B, double ScalarThis)
```
mayrmt, line 226: This public method asks for an Epetra_MultiVector, so it kind of exposes the Epetra_MultiVector wrapped in this class to the outside world. Shouldn't it take a LINALG::Vector as input?

maxfirmbach: Yeah this is weird, I can check if we even use this function ...

src/core/linalg/src/sparse/4C_linalg_vector.hpp
maxfirmbach: I guess I could just change this to `LinAlg::Vector`, that's what I meant by creating #875, there is still some work to do.

src/core/linalg/src/sparse/4C_linalg_vector.hpp
maxfirmbach: It seems like these things are there, due to `NOX`, which cant be resolved right now.

src/core/linalg/src/sparse/4C_linalg_vector.hpp
maxfirmbach: Same, due to `NOX`.

PR #876, opened 2025-06-07
Quaternion interpolation

abhi-satheesh:
## Description and Context

This PR provides functions and classes for quaternion interpolation, including spherical linear interpolation (SLERP) and generalized SLERP for multiple quaternions with unit tests.

This class is required for tensor interpolation methods , see Satheesh et al., 2024, 10.1002/nme.7373

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
@@ -1926,8 +1926,6 @@ namespace Core::LinAlg
     template <class ValueType, unsigned int i, unsigned int j>
     constexpr inline void normalize(ValueType* vector)
     {
-      FOUR_C_ASSERT_ALWAYS(j == 1, "normalize() is only available for column vectors");
-
       ValueType norm2 = DenseFunctions::norm2<ValueType, i, j>(vector);
       if (norm2 < 1e-10)
       {
@@ -2332,9 +2330,8 @@ namespace Core::LinAlg
     /// Normalize vector
     /*!
       Normalize this matrix, i.e. scale it to have a norm2 of 1.
-      This only works for column vectors, i.e. matrices with one column.
 
-      \note This function will throw an exception if the matrix is not a column vector.
+      \note This function will throw an exception if the matrix norm is zero.
      */
     inline void normalize() { DenseFunctions::normalize<ValueType, rows, cols>(data()); }
 
--- a/src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
@@ -1922,6 +1922,20 @@ namespace Core::LinAlg
       for (unsigned int c = 1; c < i * j; ++c) result += *(++mat);
       return result / (i * j);
     }
+
+    template <class ValueType, unsigned int i, unsigned int j>
+    constexpr inline void normalize(ValueType* vector)
+    {
+      FOUR_C_ASSERT_ALWAYS(j == 1, "normalize() is only available for column vectors");
+
+      ValueType norm2 = DenseFunctions::norm2<ValueType, i, j>(vector);
+      if (norm2 < 1e-10)
+      {
+        FOUR_C_THROW("Cannot normalize vector with norm2 close to zero");
+      }
+
+      DenseFunctions::scale_matrix<ValueType, i, j>(1.0 / norm2, vector);
+    }
   }  // namespace DenseFunctions
 
   namespace Internal
@@ -2314,6 +2328,15 @@ namespace Core::LinAlg
     {
       DenseFunctions::scale_matrix<ValueType, rows, cols>(scalar, data());
     }
+
+    /// Normalize vector
+    /*!
+      Normalize this matrix, i.e. scale it to have a norm2 of 1.
+      This only works for column vectors, i.e. matrices with one column.
+
+      \note This function will throw an exception if the matrix is not a column vector.
+     */
+    inline void normalize() { DenseFunctions::normalize<ValueType, rows, cols>(data()); }
 
     /// Copy: \e this = \e other
     /*!
--- a/src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
+++ b/src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
@@ -40,7 +40,8 @@ Core::LinAlg::Matrix<4, 1> Core::LinAlg::spherical_linear_interpolation(
   {
     // If the quaternions are close, use linear interpolation
     Core::LinAlg::Matrix<4, 1> result;
-    for (int i = 0; i < 4; ++i) result(i, 0) = (1.0 - t) * q1(i, 0) + t * q2_mod(i, 0);
+    for (int i = 0; i < 4; ++i) result(i, 0) = std::lerp(q1(i, 0), q2_mod(i, 0), t);
+
     // Normalize
     result.normalize();
     return result;
--- a/src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
+++ b/src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
@@ -1,0 +1,564 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#include "4C_config.hpp"
+
+#include "4C_linalg_utlis_quaternion_interpolation.hpp"
+
+#include "4C_linalg_fixedsizematrix.hpp"
+#include "4C_linalg_utils_scalar_interpolation.hpp"
+#include "4C_utils_exceptions.hpp"
+
+#include <algorithm>
+#include <cassert>
+
+FOUR_C_NAMESPACE_OPEN
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+Core::LinAlg::Matrix<4, 1> Core::LinAlg::spherical_linear_interpolation(
+    const Core::LinAlg::Matrix<4, 1>& q1, const Core::LinAlg::Matrix<4, 1>& q2, double t)
+{
+  // Compute the dot product
+  double dot = q1.dot(q2);
+
+  // If the dot product is negative, slerp won't take the shorter path.
+  // Fix by reversing one quaternion.
+  Core::LinAlg::Matrix<4, 1> q2_mod = q2;
+  if (dot < 0.0)
+  {
+    for (int i = 0; i < 4; ++i) q2_mod(i, 0) = -q2_mod(i, 0);
+    dot = -dot;
+  }
+
+  const double linear_interpolation_threshold = 0.9995;
+  if (dot > linear_interpolation_threshold)
+  {
+    // If the quaternions are close, use linear interpolation
+    Core::LinAlg::Matrix<4, 1> result;
+    for (int i = 0; i < 4; ++i) result(i, 0) = (1.0 - t) * q1(i, 0) + t * q2_mod(i, 0);
+    // Normalize
+    result.normalize();
+    return result;
+  }
+
+  double theta_0 = std::acos(dot);  // angle between input quaternions
+  double sin_theta_0 = std::sin(theta_0);
+
+  double s0 = std::sin((1.0 - t) * theta_0) / sin_theta_0;
+  double s1 = std::sin(t * theta_0) / sin_theta_0;
+
+  Core::LinAlg::Matrix<4, 1> result;
+  for (int i = 0; i < 4; ++i) result(i, 0) = s0 * q1(i, 0) + s1 * q2_mod(i, 0);
+
+  // Normalize
+  result.normalize();
+
+  return result;
+}
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+template <unsigned int loc_dim>
+Core::LinAlg::GeneralizedSphericalLinearInterpolator<loc_dim>::
+    GeneralizedSphericalLinearInterpolator(
+        const std::vector<Core::LinAlg::Matrix<4, 1>>& unit_quaternions,
+        const std::vector<Core::LinAlg::Matrix<loc_dim, 1>>& ref_locs,
+        const Core::LinAlg::ScalarInterpolation::WeightingFunction weight_func,
+        const Core::LinAlg::ScalarInterpolation::InterpParams& interp_params)
+    : unit_quaternions_(unit_quaternions),
+      ref_locs_(ref_locs),
+      weight_func_(weight_func),
+      interp_params_(interp_params)
+{
+  // Check if reference locations same dimension
+  FOUR_C_ASSERT_ALWAYS(!ref_locs.empty(),
+      "Reference locations must have the same dimension as the interpolation location.");
+
+  // Check if the number of reference locations matches the number of quaternions
+  FOUR_C_ASSERT_ALWAYS(unit_quaternions_.size() == ref_locs.size(),
+      "Number of quaternions must match number of reference locations.");
+
+  // Check if all quaternions are unit quaternions
+  FOUR_C_ASSERT_ALWAYS(std::all_of(unit_quaternions_.begin(), unit_quaternions_.end(),
+                           [](const auto& quat) { return std::abs(quat.norm2() - 1.0) < 1e-6; }),
+      "All quaternions must be normalized (norm should be close to 1).");
+}
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+template <unsigned int loc_dim>
+Core::LinAlg::GeneralizedSphericalLinearInterpolator<loc_dim>::
+    GeneralizedSphericalLinearInterpolator(
+        const std::vector<Core::LinAlg::Matrix<4, 1>>& unit_quaternions,
+        const std::vector<double>& normalized_weights)
+    : unit_quaternions_(unit_quaternions), normalized_weights_(normalized_weights)
+{
+  FOUR_C_ASSERT_ALWAYS(unit_quaternions_.size() == normalized_weights_.size(),
+      "Number of quaternions and weights must match.");
+  FOUR_C_ASSERT_ALWAYS(!unit_quaternions_.empty(), "No quaternions provided for interpolation.");
+  FOUR_C_ASSERT_ALWAYS(!normalized_weights_.empty(), "No weights provided for interpolation.");
+
+  // Check if all quaternions are normalized
+  FOUR_C_ASSERT_ALWAYS(std::all_of(unit_quaternions_.begin(), unit_quaternions_.end(),
+                           [](const auto& quat) { return std::abs(quat.norm2() - 1.0) < 1e-6; }),
+      "All quaternions must be normalized (norm should be close to 1).");
+
+  // Check if weights are non-negative and sum to 1
+  double weight_sum = 0.0;
+  for (const auto& weight : normalized_weights_)
+  {
+    if (weight < 0.0)
+      FOUR_C_THROW("All weights must be non-negative.");
+    else
+      weight_sum += weight;
+  }
+  if (std::abs(weight_sum - 1.0) > 1e-6) FOUR_C_THROW("Weights must sum to 1.");
+}
+
+/*--------------------------------------------------------------------*
+ * @brief Computes the spherical weighted average of points on the unit sphere $\mathbb{S}^3$ using
+ * Algorithm A2 from Buss and Fillmore, 2001.
+ *
+ * @details
+ * **Inputs:** Points $q_1, \ldots, q_n$ on $S^3$ and nonnegative weights $w_1, \dots, w_n$ with
+ * $\sum_{i=1}^n w_i = 1$.
+ *
+ * **Initialization:**
+ * Set $q$ to the normalized weighted sum: $q := \frac{\sum_{i=1}^n w_i q_i}{\left\| \sum_{i=1}^n
+ * w_i q_i \right\|}$.
+ *
+ * **Main Iteration:**
+ * - Construct a local Euclidean coordinate system $C$ at the current $q$.
+ * - For each $i = 1, \ldots, n$:
+ *   - Compute $q_i^* := \log_q(q_i)$ (the logarithmic map at $q$ of $q_i$).
+ *   - Let $\rho_i := \| q_i^* - q \|$.
+ *   - Define a local coordinate system $C_i$ at $q$ with the first axis pointing away from $q_i$.
+ *   - Let $M_i$ be the transformation matrix from $C_i$ to $C$.
+ *   - Let $H_i$ be a $d \times d$ diagonal matrix with first entry 1, others $\rho_i \cot(\rho_i)$.
+ * - Compute:
+ *   - $u := \sum_{i=1}^n w_i (q_i^* - q)$
+ *   - $H := \sum_{i=1}^n M_i H_i M_i^T$
+ *   - $v := H^{-1} u$
+ *   - Update $q := \exp_q(q + v)$ (using the exponential map).
+ * - Repeat until $\|v\|$ is sufficiently small, then return $q$.
+ *--------------------------------------------------------------------*/
+template <unsigned int loc_dim>
+Core::LinAlg::Matrix<4, 1>
+Core::LinAlg::GeneralizedSphericalLinearInterpolator<loc_dim>::get_interpolated_quaternion(
+    Core::LinAlg::Matrix<loc_dim, 1>* interp_loc)
+{
+  if (interp_loc != nullptr and normalized_weights_.empty())
+  {
+    normalized_weights_ = Core::LinAlg::ScalarInterpolation::calculate_normalized_weights(
+        ref_locs_, *interp_loc, weight_func_, interp_params_);
+  }
+
+  FOUR_C_ASSERT_ALWAYS(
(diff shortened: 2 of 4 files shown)


abhi-satheesh:
@c-p-schmidt @sebproell @dragos-ana kindly review the PR. 
I am unable to assign reviewers

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
```diff
@@ -0,0 +1,564 @@
...
+Core::LinAlg::QuaternionInterpolation::GeneralizedSphericalLinearInterpolator<loc_dim>::
+    GeneralizedSphericalLinearInterpolator(
+        const std::vector<Core::LinAlg::Matrix<4, 1>>& unit_quaternions,
+        const std::vector<double>& normalized_weights)
+    : unit_quaternions_(unit_quaternions), normalized_weights_(normalized_weights)
+{
+  if (unit_quaternions_.size() != normalized_weights_.size())
+    FOUR_C_THROW("Number of quaternions and weights must match.");
+  assert(!unit_quaternions_.empty() && "No quaternions provided for interpolation.");
+  assert(!normalized_weights_.empty() && "No weights provided for interpolation.");
```
amgebauer: Could you use FOUR_C_ASSERT_ALWAYS here?
```suggestion
  FOUR_C_ASSERT_ALWAYS(unit_quaternions_.size() == normalized_weights_.size(), "Number of quaternions and weights must match.");
  FOUR_C_ASSERT_ALWAYS(!unit_quaternions_.empty(), "No quaternions provided for interpolation.");
  FOUR_C_ASSERT_ALWAYS(!normalized_weights_.empty(), "No weights provided for interpolation.");
```

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+  // Check if all quaternions are normalized
+  for (const auto& quat : unit_quaternions_)
+    if (std::abs(quat.norm2() - 1.0) > 1e-6)
+      FOUR_C_THROW("All quaternions must be normalized (norm should be close to 1).");
```
amgebauer:
```suggestion
  FOUR_C_ASSERT_ALWAYS(std::ranges::all_of(unit_quaternions_, [](const auto& quat) { return std::abs(quat.norm2() - 1.0) > 1e-6; },"All quaternions must be normalized (norm should be close to 1).");
```

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
```diff
@@ -0,0 +1,274 @@
...
+       * @brief Generalized Spherical Linear Interpolator for quaternions.
+       *
+       * This class implements a generalized spherical linear interpolation (SLERP) method for
+       * quaternions, allowing for interpolation based on reference locations and weights.
+       * It supports different dimensions for reference locations, enabling flexible use cases.
+       *
+       * The interpolation is based on the method (A2) described by Buss and Fillmore, 2001.
+       */
+      template <unsigned int loc_dim = 1>
+      class GeneralizedSphericalLinearInterpolator
```
amgebauer: Couldn't this all be packed into free functions?

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * The main components are:
+     * - `spherical_linear_interpolation`: Computes the SLERP between two unit quaternions for
+     * a given interpolation parameter.
+     * - `GeneralizedSphericalLinearInterpolator`: A template class that enables generalized
+     * quaternion interpolation based on reference locations and custom weighting functions,
+     * supporting both location-based and explicit weight-based interpolation.
+     * The algorithm is based on the algorithm by Buss and Fillmore (2001).
+     */
+    namespace QuaternionInterpolation
```
amgebauer: I think this namespace is not necessary. I think it is clear from the signature of the functions that the spehrical interpolation is based on quaternions.

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
```diff
@@ -0,0 +1,274 @@
...
+     * and stores the result in \p result. All vectors are represented as
+     * 3x1 matrices.
+     *
+     * @param u The first input vector (3x1 matrix).
+     * @param v The second input vector (3x1 matrix).
+     * @param result Reference to a 3x1 matrix where the cross product result will be stored.
+     */
+    void vector_cross_product(const Core::LinAlg::Matrix<3, 1>& u,
+        const Core::LinAlg::Matrix<3, 1>& v, Core::LinAlg::Matrix<3, 1>& result);
+
```
amgebauer: We already have this implemented in `4C_linalg_fixedsizematrix.hpp`

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
```diff
@@ -0,0 +1,274 @@
...
+     * @brief Normalizes the given vector to have unit length.
+     *
+     * This function modifies the input vector in place so that its Euclidean norm becomes 1.
+     *
+     * @tparam loc_dim The dimension of the vector.
+     * @param vector Reference to the vector to be normalized. The vector is represented as a column
+     * matrix.
+     */
+    template <unsigned int loc_dim>
+    void normalize_vector(Core::LinAlg::Matrix<loc_dim, 1>& vector);
```
amgebauer: Could you put this into fixedsizematrix?

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
```diff
@@ -0,0 +1,274 @@
...
+
+        /**
+         * @brief Projects the difference between two quaternions onto the subspace orthogonal to
+         * the second quaternion.
+         *
+         * Given two quaternions, this function computes the vector (quaternion_1 - quaternion_2)
+         * and then removes its component along quaternion_2, effectively projecting the difference
+         * onto the hyperplane orthogonal to quaternion_2.
+         *
+         * @tparam loc_dim The local dimension parameter for the interpolator.
```
c-p-schmidt:
```suggestion
```

src/core/linalg/tests/4C_linalg_utils_quaternion_interpolation_test.cpp
```diff
@@ -0,0 +1,401 @@
...
+  Core::LinAlg::Matrix<4, 1> slerp_result =
+      Core::LinAlg::QuaternionInterpolation::spherical_linear_interpolation(q1, q2, t);
+
+  // Analytical result for SLERP between (0,0,0,1) and (0,0,1,0) at t=0.5:
+  // angle = acos(0) = pi/2
+  // slerp = (sin((1-t)*theta)*q1 + sin(t*theta)*q2) / sin(theta)
+  // theta = pi/2, sin(theta) = 1
+  // sin((1-t)*theta) = sin(0.5*pi/2) = sin(pi/4)
+  // sin(t*theta) = sin(0.5*pi/2) = sin(pi/4)
+  double theta = M_PI / 2.0;
```
c-p-schmidt: Probably we should start using the mathematical constants that are available since C++20. 
```suggestion
  double theta = std::numbers::pi_v<double> / 2.0;
```

src/core/linalg/tests/4C_linalg_utils_quaternion_interpolation_test.cpp
```diff
@@ -0,0 +1,401 @@
...
+  std::vector<double> weights = {0.75, 0.25};
+
+  Core::LinAlg::QuaternionInterpolation::GeneralizedSphericalLinearInterpolator interpolator(
+      quats, weights);
+  Core::LinAlg::Matrix<4, 1> result = interpolator.get_interpolated_quaternion();
+
+  double norm = std::sqrt(result(0, 0) * result(0, 0) + result(1, 0) * result(1, 0) +
+                          result(2, 0) * result(2, 0) + result(3, 0) * result(3, 0));
+  ASSERT_NEAR(norm, 1.0, 1e-12);
+  ASSERT_GT(result(3, 0), result(0, 0));  // w should be greater than x
```
c-p-schmidt, line 134: Is this test specific enough, or should we also test against reference solutions?

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
```diff
@@ -0,0 +1,564 @@
...
+  // If the dot product is negative, slerp won't take the shorter path.
+  // Fix by reversing one quaternion.
+  Core::LinAlg::Matrix<4, 1> q2_mod = q2;
+  if (dot < 0.0)
+  {
+    for (int i = 0; i < 4; ++i) q2_mod(i, 0) = -q2_mod(i, 0);
+    dot = -dot;
+  }
+
+  const double DOT_THRESHOLD = 0.9995;
```
c-p-schmidt: Why this strange case style? You could also directly say what it does.
```suggestion
  const double linear_interpolation_threshold = 0.9995;
```

src/core/linalg/tests/4C_linalg_utils_quaternion_interpolation_test.cpp
abhi-satheesh: Yes. The test is only for that as mentioned in the test description.
I can also couple with the last test and remove this test also. TBH it was an intermediate test.

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
abhi-satheesh: I have a different opinion. 
the namespace introduce to club all the quternion interpolation methods,

it should contain: SLERP, generalized SLERP,  spline SLERP, LERP, .. any other methods.. 
Such that `QuaternionInterpolation::Method`

what’s your opinion?

amgebauer: In my eyes, they are grouped enough by being in a file. The signature is already distinct enough.

`Core::LinAlg::QuaternionInterpolation::GeneralizedSphericalLinearInterpolator`
vs.
`Core::LinAlg::GeneralizedSphericalLinearInterpolator`

In my eyes, the latter is clear enough

Edit: It is of course only my opinion :grimacing:

bgoderbauer: We currently have two other namespaces, `ScalarInterpolation` and `TensorInterpolation`, in `Core::LinAlg`. I'm not familiar with the topic, but wouldn't it be nice for readability if we had a similar structure for all the interpolation cases? Maybe one big namespace `Interpolation`, or three distinct ones as proposed with this PR? Or we remove all three namespaces. But if we keep the existing namespaces and drop the `QuaternionInterpolation` here, it appears a little mixed up.

I personally like the idea of one `Core::LinAlg::Interpolation` namespace.

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
```diff
@@ -0,0 +1,564 @@
...
+    }
+    Core::LinAlg::normalize_vector(e1);
+    Core::LinAlg::normalize_vector(e2);
+    Core::LinAlg::normalize_vector(e3);
+
+    grad.put_scalar(0.);
+    hessian.put_scalar(0.);
+
+    double tt = 0.;
+    // Next approximation approximation
```
dragos-ana:
```suggestion
    // Next approximation 
```

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+      q_interptmp.put_scalar(0.);
+      q_interptmp.update(project_quaternion(qi, q_interp));
+
+      double sin_theta = q_interptmp.norm2();
+      double cos_theta = q_interp.dot(qi);
+      double theta = atan2(sin_theta, cos_theta);
+
+      hessian_local.put_scalar(0.);
+      grad_local.put_scalar(0.);
```
dragos-ana, line 234: Some comments describing what is currently done in each step of the loop iteration would improve the readability

abhi-satheesh:
> Thanks @abhi-satheesh for adding this nice functionality.
> 
> Since it might be relevant for some heart research, do you know how easy your generalized slerp implementation can be adapted for a bi-slerp operation? It takes into account that `+f `and `-f` represent the same fiber direction. A normal slerp operation only needs to be adatped at a single point.

Thanks @amgebauer for the comments.
Did you mean bi-slerp, applying twice slerp for four directions?

I am not sure about the accuracy of that bi-slerp, need to check. You may simply use the gen SLERP with 4 quarterninon with appropriate weights, the resultant quaterniom will be in the geodesic path.

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
abhi-satheesh: > We currently have two other namespaces, `ScalarInterpolation` and `TensorInterpolation`, in `Core::LinAlg`. I'm not familiar with the topic, but wouldn't it be nice for readability if we had a similar structure for all the interpolation cases? Maybe one big namespace `Interpolation`, or three distinct ones as proposed with this PR? Or we remove all three namespaces. But if we keep the existing namespaces and drop the `QuaternionInterpolation` here, it appears a little mixed up.
> 
> I personally like the idea of one `Core::LinAlg::Interpolation` namespace.

Thanks for your views.

In `Core::LinAlg::Interpolation` namespace how do you want to differentiate between the data types(scalar, tensor, vector, quaternion, rotation vector, etc)?

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
abhi-satheesh: I will post the Algorithm from the paper

amgebauer:
> > Thanks @abhi-satheesh for adding this nice functionality.
> > Since it might be relevant for some heart research, do you know how easy your generalized slerp implementation can be adapted for a bi-slerp operation? It takes into account that `+f `and `-f` represent the same fiber direction. A normal slerp operation only needs to be adatped at a single point.
> 
> Thanks @amgebauer for the comments. Did you mean bi-slerp, applying twice slerp for four directions?
> 
> I am not sure about the accuracy of that bi-slerp, need to check. You may simply use the gen SLERP with 4 quarterninon with appropriate weights, the resultant quaterniom will be in the geodesic path.

Sorry, I were not 100% clear about that. Its a small modification of slerp (step 3 here):
Screenshot from 2025-06-11 09-02-50
(taken from the supplementary material of Bayer et al., Ann Biomed Eng 40, 2012.

It is easy for the slerp function (obviously), but what about the generalized version?

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
bgoderbauer: Probably on the innermost level. For example, currently we have `ScalarInterpolation::InterpParams`. This could be exchanged to `Interpolation::ScalarInterpParams`. Only a suggestion :smile:. I'm not an expert, but it appears to me as a logical grouping of similar content.

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
abhi-satheesh: Well, it could be...

I was thinking along these lines:
These functions will be used to interpolate element quantities—such as at Gauss points. In such cases, the data set remains the same for each Gauss point. So, one could instantiate the class with the data set once, and then simply query the class for interpolated values at different Gauss points.

If these were free functions instead, you'd need to supply the same data repeatedly, which would be less efficient.

src/core/linalg/tests/4C_linalg_utils_quaternion_interpolation_test.cpp
abhi-satheesh: @c-p-schmidt  any preference?

abhi-satheesh:
> > > Thanks @abhi-satheesh for adding this nice functionality.
> > > Since it might be relevant for some heart research, do you know how easy your generalized slerp implementation can be adapted for a bi-slerp operation? It takes into account that `+f `and `-f` represent the same fiber direction. A normal slerp operation only needs to be adatped at a single point.
> > 
> > 
> > Thanks @amgebauer for the comments. Did you mean bi-slerp, applying twice slerp for four directions?
> > I am not sure about the accuracy of that bi-slerp, need to check. You may simply use the gen SLERP with 4 quarterninon with appropriate weights, the resultant quaterniom will be in the geodesic path.
> 
> Sorry, I were not 100% clear about that. Its a small modification of slerp (step 3 here): Screenshot from 2025-06-11 09-02-50 (taken from the supplementary material of Bayer et al., Ann Biomed Eng 40, 2012.
> 
> It is easy for the slerp function (obviously), but what about the generalized version?

If you are asking the difference between the SLERP and Generalized SLERP (Buss and Fillmore).

"_SLERP takes two quaternions as input to interpolate between them along the shortest path on the_ $\mathbb{S}^3$ (or $\mathbb{H}^1$), _whereas the Buss & Fillmore generalization extends this to handle two or more quaternions, allowing for a weighted average on the_ $\mathbb{S}^3$.
 _When_ $N = 2$ _(i.e., two quaternions), the generalized SLERP reduces to standard SLERP_: $\text{gen_SLERP}(q_1, q_2) = \text{SLERP}(q_1, q_2)$."

for more explanation see our paper [https://onlinelibrary.wiley.com/doi/abs/10.1002/nme.7373](url)
image

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
abhi-satheesh: @amgebauer @bgoderbauer @c-p-schmidt  
How we should move forward?

amgebauer: I'm a fan of not too many namespaces, but we can go with @bgoderbauer's suggestion

abhi-satheesh: @c-p-schmidt @dragos-ana 
We had a discussion yesterday.
What we converged was remove the namespace and use the class as the way to identify the interpolation strategy.”, ie., 

`Core::LinAlg::Scalarinterpolator`
`Core::LinAlg::Tensorintepolator`
The specific methods will be implemented inside the class.

The quaternion interpolation is now 

`Core::LinAlg::GeneralizedSphericalLinearInterpolator`

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
abhi-satheesh: image
image

what's your opinion on the implementation?

abhi-satheesh:
@c-p-schmidt @amgebauer @dragos-ana 
Please check if your comments are addressed.

@c-p-schmidt please merge it if the above condition is satisfied.

src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
```diff
@@ -1922,6 +1922,20 @@ namespace Core::LinAlg
       for (unsigned int c = 1; c < i * j; ++c) result += *(++mat);
       return result / (i * j);
     }
+
+    template <class ValueType, unsigned int i, unsigned int j>
```
amgebauer, line 1926: instead of the `FOUR_C_ASSERT_ALWAYS`, you can write
```suggestion
    template <class ValueType, unsigned int i, unsigned int j> requires(i==1 || j == 1)
```

src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
```diff
@@ -2315,6 +2329,15 @@ namespace Core::LinAlg
...
     }
 
+    /// Normalize vector
+    /*!
+      Normalize this matrix, i.e. scale it to have a norm2 of 1.
+      This only works for column vectors, i.e. matrices with one column.
+
+      \note This function will throw an exception if the matrix is not a column vector.
+     */
+    inline void normalize() { DenseFunctions::normalize<ValueType, rows, cols>(data()); }
```
amgebauer: Better create a free function instead of a member function. This function is only valid for column-vectors and not for the general matrix.

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
```diff
@@ -0,0 +1,564 @@
...
+    for (int i = 0; i < 4; ++i) q2_mod(i, 0) = -q2_mod(i, 0);
+    dot = -dot;
+  }
+
+  const double linear_interpolation_threshold = 0.9995;
+  if (dot > linear_interpolation_threshold)
+  {
+    // If the quaternions are close, use linear interpolation
+    Core::LinAlg::Matrix<4, 1> result;
+    for (int i = 0; i < 4; ++i) result(i, 0) = (1.0 - t) * q1(i, 0) + t * q2_mod(i, 0);
```
amgebauer: You could make the interpolation here obvious:
```suggestion
    for (int i = 0; i < 4; ++i) result(i, 0) = std::lerp(q1(i, 0), q2(i, 0), t);
```

amgebauer (commented):
I have a few additional comments

src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
abhi-satheesh: I posted the these function in your previous response. 
I believed that you are fine with the changes.

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.hpp
amgebauer: Oh sorry, I have missed your answer here...

dragos-ana (dismissed):
No further comments from me, besides what @amgebauer added

src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
c-p-schmidt: I'm also not 100% sure what you would like to have here. I suppose you would prefer a method, e.g., templated on the row-dimension, that directly takes a reference to a `Core::LinAlg::Matrix<row-dim,1>&`  and then does the normalization?
And then to put it in what namespace? Directly in `Core::LinAlg`?

c-p-schmidt:
I'm curious about @amgebauer's reply, but besides that, LGTM.

abhi-satheesh:
> I have a few additional comments

abhi-satheesh:
> I have a few additional comments

Please try to add your comments earlier, especially on the first commit, to help speed up the merging process.
It would also be really helpful if you could leave comments as early as possible on files or methods that affect most of the repository, since those take a long time to compile.

src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
abhi-satheesh: @amgebauer  If I get you correctly, 
This method was first implemented in `Core::LinAlg` as a template method similar to the explanation by @c-p-schmidt 

Do I now change back to the initial commit?

amgebauer: @c-p-schmidt yes 👍🏼

@abhi-satheesh with my previous comment, I only wanted to see the free function in the file were we define all the other matrix functions (like multiply, dot) and not in the file of tensor interpolation (normalization is general and useful in many situations, not just during slerp)

src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
abhi-satheesh: > instead of the `FOUR_C_ASSERT_ALWAYS`, you can write

I don't think this is even necessary,
In general, this is valid for vector or tensor,  I will remove this lines

src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
abhi-satheesh: yes..

https://github.com/4C-multiphysics/4C/pull/876#issuecomment-2985630367

src/core/linalg/src/dense/4C_linalg_fixedsizematrix.hpp
abhi-satheesh: @amgebauer 
I don't think this should be a free function.

Instead, it should be an internal method, like `invert`, `put_scalar`, or similar.
I will remove the exception for non-column matrices, since `normalize` is valid for general matrices as well:

$T_{\text{normlize}} = \frac{T}{\parallel T \parallel_{L_2}}$

amgebauer: I would argue that `invert` should actually be a free function. But we can move forward here.

abhi-satheesh: Thanks @amgebauer. I agree 'invert'is not a nice example.

I will push the changes.
Now, the method looks like
'tensor.normalize()'

abhi-satheesh:
@c-p-schmidt @amgebauer @dragos-ana 

Please merge it if everyone is fine with the changes

PR #879, opened 2025-06-11
Implement some small improvements to tensor class

amgebauer:
The following improvements are added

 * Add eigen-decomposition for symmetric tensors
 * Add generators for diagonal, identity (2nd and 4th order), symmetric identity (4th order), full tensors
 * Add conversion functions to convert tensors to matrices and vice versa
 * Add in-place operators for symmetric tensors
 * Add a lot of unit tests

This PR will then enable us to use Symmetric tensors during the solid evaluation #869 (and in a later step during solid material evaluation)

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_symmetric_tensor.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_symmetric_tensor.hpp
@@ -189,11 +189,11 @@ namespace Core::LinAlg
     struct IsSymmetricCompressionTypeHelper<SymmetricCompression<n...>> : public std::true_type
     {
     };
-
-    template <typename T>
-    constexpr bool is_symmetric_tensor =
-        IsSymmetricCompressionTypeHelper<TensorCompressionType<std::remove_cvref_t<T>>>::value;
   }  // namespace Internal
+
+  template <typename T>
+  constexpr bool is_symmetric_tensor = Internal::IsSymmetricCompressionTypeHelper<
+      TensorCompressionType<std::remove_cvref_t<T>>>::value;
 
   /*!
    * @brief An owning, dense tensor of arbitrary rank with symmetry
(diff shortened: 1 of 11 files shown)


src/core/linalg/src/dense/4C_linalg_symmetric_tensor.hpp
```diff
@@ -264,7 +286,7 @@ namespace Core::LinAlg
    * cases) identical to the normal tensor
    */
   template <typename Number, TensorStorageType storage_type, std::size_t... n>
-  constexpr Tensor<Number, n...> get_full(
+  constexpr auto get_full(
```
davidrudlstorfer, line 289: Why do you switch to `auto` here?

davidrudlstorfer (dismissed):
@amgebauer I did my best during the review of this PR but I must confess that I do not fully understand it implementation wise 😅

Nice that you covered everything with a lot of unit tests - thanks:)

src/core/linalg/src/dense/4C_linalg_symmetric_tensor.hpp
amgebauer: This also needs to work for const views. In that case, `Tensor<Number, n...>` does not work, but only  `Tensor<std::remove_const_t<Number>, n...>` and at some point it gets hardly readable.

davidrudlstorfer: Thanks for the explanation:)

src/core/linalg/src/dense/4C_linalg_tensor_generators.hpp
```diff
@@ -0,0 +1,123 @@
...
+FOUR_C_NAMESPACE_OPEN
+
+namespace Core::LinAlg
+{
+  /*!
+   * @brief Diagonal 2-tensor with the given value on the diagonal.
+   */
+  template <std::size_t... n>
+    requires(sizeof...(n) == 2 && std::array{n...}[0] == std::array{n...}[1])
+  constexpr auto diagonal(const auto& value)
```
sebproell, line 29: The name `Core::LinAlg::diagonal` may be too unspecific. How about a namespace `TensorGenerators` that one can easily "import" with `using namespace ...` when required.

amgebauer: I'm a fan of short namespaces, but I see that it is not clear whether a Matrix or a Tensor is generated. I moved it into the namespace `TensorGenerators`

PR #881, opened 2025-06-11
Fix failing asan for benchmark tests

amgebauer:
Added the missing asan options to the executable of the address sanitizer.

### DIFF
--- a/tests/benchmark_tests/4C_benchmark_tests_main.cpp
+++ b/tests/benchmark_tests/4C_benchmark_tests_main.cpp
@@ -22,6 +22,7 @@ int main(int argc, char** argv)
   } cleanup;
   // Kokkos should be initialized right after MPI.
   Kokkos::ScopeGuard kokkos_guard{};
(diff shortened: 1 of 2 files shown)


tests/benchmark_tests/4C_benchmark_tests_main.cpp
```diff
@@ -22,6 +22,7 @@ int main(int argc, char** argv)
   } cleanup;
   // Kokkos should be initialized right after MPI.
   Kokkos::ScopeGuard kokkos_guard{};
+  Core::Utils::SingletonOwnerRegistry::ScopeGuard guard;
```
amgebauer, line 25: @sebproell As far as I can see, it is safe to have multiple ScopeGuards, right? I want to have this one globally for all tests, but additionally locally if there are tests that rely on a clean environment.

sebproell: Yes, should be fine. I think we also have this in unit tests.

sebproell (approved):

c-p-schmidt (approved):

PR #882, opened 2025-06-11
Emit internal all_ofs to metadata

sebproell:
The internal InputSpecs create all_ofs to encode a full input object. Our additional tools can also profit from this information if we emit it as metadata.


sebproell:
@gilrrei Will this cause downstream problems in FourCIPP that need to be fixed first, or can we merge this independently?

gilrrei (approved):
LGTM. No changes are required in FourCIPP as long as a JSON schema can be created and the legacy sections do not change :)

davidrudlstorfer (approved):

PR #883, opened 2025-06-11
Remove non-const getter members from `BaseDataSDyn` and remove some obsolete

vovannikov:
## Description and Context
Why dealing with NOX infrastructure, I encountered many members. I tried to remove non-const getters which are not in use, also I refactored the getters of basic type to return value instead of references.

#359

### DIFF
--- a/src/structure_new/src/4C_structure_new_timint_basedatasdyn.hpp
+++ b/src/structure_new/src/4C_structure_new_timint_basedatasdyn.hpp
@@ -75,10 +75,10 @@ namespace Solid
 
      protected:
       /// get the indicator state
-      inline const bool& is_init() const { return isinit_; };
+      inline bool is_init() const { return isinit_; };
 
       /// get the indicator state
-      inline const bool& is_setup() const { return issetup_; };
+      inline bool is_setup() const { return issetup_; };
 
       /// Check if init() has been called, yet.
       inline void check_init() const
@@ -97,14 +97,14 @@ namespace Solid
       ///@{
 
       /// Returns final time \f$t_\text{fin}\f$
-      const double& get_time_max() const
+      double get_time_max() const
       {
         check_init_setup();
         return timemax_;
       };
 
       /// Returns final time step \f$N\f$
-      const int& get_step_max() const
+      int get_step_max() const
       {
         check_init_setup();
         return stepmax_;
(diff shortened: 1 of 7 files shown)


src/structure_new/src/4C_structure_new_timint_basedatasdyn.hpp
```diff
@@ -428,105 +421,46 @@ namespace Solid
...
       ///@}
 
-      /// @name Get mutable general control parameters (read and write access)
+      /// @name Set some general control parameters
       ///@{
-
-      /// Returns final time \f$t_\text{fin}\f$
-      double& get_time_max()
+      /// Set final time step \f$N\f$
+      void set_step_max(double new_stepmax)
```
c-p-schmidt: Shouldn't the step be an `int`?

src/structure_new/src/4C_structure_new_timint_basedatasdyn.hpp
```diff
@@ -428,105 +421,46 @@ namespace Solid
...
-      double& get_pre_stress_displacement_tolerance()
-      {
-        check_init_setup();
-        return prestress_displacement_tolerance_;
+        rand_tsfac_ = new_rand_tsfac;
       }
 
-      [[nodiscard]] int& get_pre_stress_minimum_number_of_load_steps()
+      /// Set level of refinement in case of divercontype_ == adapt_step
+      void set_div_con_refine_level(double new_divconrefinementlevel)
```
c-p-schmidt: This also sounds like an `int` to me.

src/structure_new/src/4C_structure_new_timint_basedatasdyn.hpp
```diff
@@ -428,105 +421,46 @@ namespace Solid
...
-      enum Inpar::Solid::NonlinSolTech& get_nln_solver_type()
-      {
-        check_init_setup();
-        return nlnsolvertype_;
-      };
-
-      /// Returns mid-time energy type
-      enum Inpar::Solid::MidAverageEnum& get_mid_time_energy_type()
+      /// Return number of fine steps in case of in case of divercontype_ == adapt_step
+      void set_div_con_num_fine_step(double new_divconnumfinestep)
```
c-p-schmidt: Same here.

c-p-schmidt (approved):
Thank you for the cleanup.

src/structure_new/src/4C_structure_new_timint_basedatasdyn.hpp
vovannikov: Thank you, this came from some preg replaces I made. I'll fix it.

sebproell (dismissed):
The class `BaseDataSDyn` should just be a struct with public non-const members. Feel free, to change that in the future.

c-p-schmidt (approved):

PR #884, opened 2025-06-12
InputSpecBuilders::selection can store in structs

sebproell:
Since we prefer to use `selection` over `one_of`, it should also be able to store directly to a struct following #873.

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -716,6 +716,7 @@ namespace Core::IO
       std::array<Size, rank<T>()> size;
     };
 
+    template <typename StorageType>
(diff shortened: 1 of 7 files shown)


sebproell:
@bwirthl I took the magnet function as an example of how this PR could be used. The benefit is that we can specify where we want to store the parameters right where we specify the input specs, and we need no more string-based queries later.

To make this work, I had to restructure the input a little bit to have more nested YAML (for input that is a good thing, IMO). Please let me know if this is problematic for some reason or if there are questions. Note that I un-nested the C++ code for the input specs by pulling out a few of the inner specs, so the diff is a little wonky.

bwirthl (approved):

amgebauer (approved):

PR #885, opened 2025-06-12
Extend installation section of the documentation

ischeider:
## Description and Context

This PR includes the yaml settings necessary for CLion and VS code to the installation section.
In addition, a note is added which cmake variables are necessary to create a local documentation.


sebproell:
I would also prefer a documentation more along the lines of: this thing is possible in X, please consult the docs of X.

ischeider:
I just continue what others did before (e.g., the screen shot of VS code). 

If the documentation is too general, it is of no use (except that something should be possible at all). How often do you expect the settings window to change? Since we foresee a growing community using 4C in the the future, changes in setting up the system should become apparent quickly. But of course we can remove the screenshot and shrink down the list to a few points, i.e., the JSON schema mappings using a local mappings file and a file path pattern.

ischeider:
I removed the screenshot and reduced the detailed action points for defining the json schema in CLion slightly. I hope this is a good compromise between a more general description, which still works in future releases, but is still helpful enough for users to create this schema definition.

doc/documentation/src/installation/installation.rst
```diff
@@ -455,6 +478,32 @@ Now you can attach gdb to each process with the following configuration:
 Start it two times and choose in the prompt the respective process id.
 Wait until both instances are connected and then start the computation by pressing any key in the 4C terminal.
 
+.. _vscode_yaml_schema:
+
+**Adding yaml Schema to VS Code**
+
+In order to use the Yaml schema in VS Code, which simplifies editing |FOURC|  input files significantly, is done in two steps.configure
+
+1. install YAML by Red Hat
```
davidrudlstorfer: Maybe

```suggestion
1. Install the official YAML by Red Hat extension within VS Code
```

to highlight that this is a VS Code extension and not something you need to install on your system, i.e., virtual env

davidrudlstorfer (dismissed):
Thanks for the update:)

doc/documentation/src/installation/installation.rst
```diff
@@ -480,17 +529,17 @@ Here's a list of all valid custom target specifiers with a brief explanation:
 
 Executables:
 
-- ``4C`` generate the main |FOURC| executable only
+- ``4C`` generate the main |FOURC| executable only (this is the default)
```
sebproell: That's actually not correct but may be close enough. We also build unit tests,

doc/documentation/src/installation/installation.rst
```diff
@@ -480,17 +529,17 @@ Here's a list of all valid custom target specifiers with a brief explanation:
...
 - ``post_processor`` build the post-filters only
 - ``post_monitor`` build a nodal data extraction application
 - ``full`` generate all executable targets of |FOURC|
 
-Documentation:
+Refer to the variable ``FULL_TARGETS`` IN ``<sourcedir>/CMakeLists.txt`` for a definition of all executable targets.
```
sebproell: I don't think this is useful for users.

sebproell (dismissed):

doc/documentation/src/installation/installation.rst
ischeider: I deleted the sentence completely.

doc/documentation/src/installation/installation.rst
ischeider: I removed it here and put a comment on the default at the first place where ninja is mentioned.

davidrudlstorfer (approved):

sebproell (approved):
Thanks! I will expand this a little with the partial json schema and a general comment on how to create input files (see item in #598).

PR #886, opened 2025-06-12
Modify doxygen comments for problemtype enum

ischeider:
## Description and Context

Just a tiny change for the doxygen documentation of the `enum class ProblemType`, so that the description is also transferred (the explanatory table was empty before). I also changed a few flaws in the descriptions of the `enum ActionType`.


src/core/legacy_enum_definitions/4C_legacy_enum_definitions_problem_type.hpp
```diff
@@ -19,48 +19,48 @@ namespace Core
...
+    red_airways,           /// reduced dimensional airways
+    reduced_lung,          /// New implementation of red_airways with Newton-iteration
+    scatra,                /// scalar transport problem (e.g. convection-diffusion)
+    ssi,                   /// scalar structure interaction
+    ssti,                  /// scalar structure thermo interaction
+    sti,                   /// scalar-thermo interaction
+    structure,             /// structural problem
+    thermo,                /// thermal problem
+    thermo_fsi,            /// thermo-fluid-structure-interaction problem
+    tsi,                   /// thermal structure interaction
```
mayrmt: Are these correctly displayed in Doxygen? Or should this be more like `///<`?

ischeider: Actually it does (at least locally), even though I unintentionally missed the "<" in my replacement action. Anyway, I changed it.

PR #890, opened 2025-06-13
Input field InputSpec to read local parameters

maxiludwig:
This PR is a next step to allow reading parameters elementwise from a file. It adds an additional `InputSpec` `input_field`.
An input file would look like this:
```
 stiffness:
   type: constant
   value: 100.0
```
or
```
 stiffness:
   type: from_file
   value: /some/file.json
```

Co-Authored by @sebproell


src/core/io/src/4C_io_input_field.hpp
```diff
@@ -82,6 +88,22 @@ namespace Core::IO
 
     std::variant<T, std::unordered_map<IndexType, T>> data_;
   };
+
+  template <typename T>
+  void read_input_field_from_json(const std::string& file_name, T& value, const std::string& key)
```
sebproell: - If you look at the arguments of this functions, you can see that it is not really related to the `InputField`. I think it would better fit into the `4C_io_yaml.hpp`. You can also see this from the includes, which now pull in a lot of things that are not relevant if you already have an `InputField`.
- Also, the "json" in the name is unnecessary, as this works for yaml and json.
- This function could be improved with a few checks: does the file exist? Is it the correct format? What If the requested key does not exist?

src/core/io/tests/4C_io_input_local_parameter_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "4C_io_input_spec_builders.hpp"
+#include "4C_io_value_parser.hpp"
+#include "4C_io_yaml.hpp"
+#include "4C_legacy_enum_definitions_conditions.hpp"
+#include "4C_unittest_utils_assertions_test.hpp"
+#include "4C_unittest_utils_support_files_test.hpp"
+#include <browniandyn/4C_browniandyn_input.hpp>
+
+#include <ryml_std.hpp>
+#include <Sacado_trad.hpp>
```
sebproell: Not sure you need these two.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -2313,6 +2378,53 @@ Core::IO::InputSpec Core::IO::InputSpecBuilders::selection(
       common_data);
 }
 
+enum class InputFieldType : std::uint8_t
+{
+  constant,
+  from_file,
+};
```
sebproell: This currently lives at "global scope" (not really, it is inside `FourC` but we should not place it there either). If it is only used internally, put it into `Core::IO::Internal` to hide it from users. Same for the function below

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -1392,6 +1404,45 @@ namespace Core::IO
...
+     *  "stiffness": {
+     *    "1": 2.0,
+     *    "2": 3.5,
+     *    "3": 4.0,
+     *    "4": 5.5
+     *    }
+     * }
+     */
+    template <typename T>
+    [[nodiscard]] InputSpec input_field(std::string name, ParameterDataIn<T> data = {});
```
sebproell, line 1474: I like the docs!

Not sure this function should be inside this header since it is a special case. Currently, we do not have another header for such specs, though. I'll keep it in mind and we can move it later.

sebproell (changes requested):
Just a few things.

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -82,6 +82,8 @@ namespace Core::IO
 
     std::variant<T, std::unordered_map<IndexType, T>> data_;
   };
+
+
```
bgoderbauer:
```suggestion
```

src/core/io/src/4C_io_yaml.hpp
```diff
@@ -226,6 +249,47 @@ namespace Core::IO
...
+    catch (const std::exception& e)
+    {
+      throw YamlException(
+          std::format("Failed to parse file '{}': {}", file_path.string(), e.what()));
+    }
+
+    ryml::NodeRef root = tree.rootref();
+    if (!root.has_child(ryml::to_csubstr(key)))
+    {
+      throw YamlException(std::format("Key '{}' not found in file '{}'.", key, file_path.string()));
```
sebproell:
```suggestion
      throw YamlException(std::format("Key '{}' not found on the top level in file '{}'.", key, file_path.string()));
```

src/core/io/src/4C_io_yaml.hpp
```diff
@@ -226,6 +249,47 @@ namespace Core::IO
...
+    }
+
+    ryml::NodeRef node = root[ryml::to_csubstr(key)];
+    ConstYamlNodeRef json_key_node(node, file_path);
+    read_value_from_yaml(json_key_node, value);
```
sebproell:
```suggestion
    ConstYamlNodeRef key_node(node, file_path);
    read_value_from_yaml(key_node, value);
```

bgoderbauer (approved):
Very nice and clear implementation! I think this will make it very easy to work with field data, especially with all the explicitly handled exceptions.

Just out of curiosity, since I didn't look a lot into the new input spec yet: what is the bridge between the input field and the associated elements/nodes? Where/how do we assign, e.g., ele_id -> param?

maxiludwig:
The `InputField` that @sebproell introduced in #866 is actually doing the mapping from indices to parameter values. Have a look into `4C_io_input_field.hpp`.
It can be used as done in the introduced unittest:
```
container.group("stiffness").get<InputField<double>>("stiffness");
EXPECT_EQ(input_field_stiffness.at(1), 2.0);
```

sebproell:
@maxiludwig Just squash minor fixups into the original commit.

sebproell:
@maxiludwig I took the liberty to fix the failing formatting and squash the commits.

src/core/io/tests/4C_io_input_input_field_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+  {
+    const std::string input_field_file =
+        TESTING::get_support_file_path("test_files/input_field/stiffness_input_field.json");
+    std::unordered_map<int, double> stiffness_map;
+    read_value_from_yaml(input_field_file, "stiffness", stiffness_map);
+    std::unordered_map<int, double> expected_stiffness_map = {
+        {1, 2.0}, {2, 3.5}, {3, 4.0}, {4, 5.5}};
+    EXPECT_EQ(stiffness_map, expected_stiffness_map);
+  }
+  TEST(InputFile, ReadSpecInputField)
```
maxfirmbach, line 40: I think a blank line above would make it a bit more readable and easier to distinguish the different tests, just beauty though.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -571,6 +572,36 @@ namespace Core::IO
     {
       return !val.has_value() || validate_helper(*val, validator);
     }
+
+    enum class InputFieldType : std::uint8_t
+    {
```
maxfirmbach, line 577: Could this also handle a `function`? If yes adding it would cover all of our current use-cases right?

src/core/io/tests/4C_io_input_input_field_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "4C_io_value_parser.hpp"
+#include "4C_io_yaml.hpp"
+#include "4C_legacy_enum_definitions_conditions.hpp"
+#include "4C_unittest_utils_assertions_test.hpp"
+#include "4C_unittest_utils_support_files_test.hpp"
+
+#include <iostream>
+#include <map>
+#include <sstream>
+namespace
```
maxfirmbach, line 23: Can we also do a blank space above, just for separating namespace and includes properly.

src/core/io/src/4C_io_input_spec_builders.hpp
maxiludwig: So far it can't. But yes, specifying a parameter by function is the third relevant use-case for us here. It would require some more adaptions though, so I would opt for implementing it in a follow up step.

maxfirmbach: @maxiludwig Fine for me :+1:.

PR #891, opened 2025-06-13
Remove pseudo geometry type HERM2LINE<X>

sebproell:
The beam elements use a cell type that the discretization does not understand, namely the `HERM2LINE<X>` type. Luckily, what this really means is that the geometry is described by a `LINE<X>` and the `HERM` prefix is just a bool telling us that hermite interpolation of the centerline with only the end nodes of the cell is used. I propose to use a bool `HERMITE_CENTERLINE` to indicate that. Naming suggestions welcome.

This change will enable us to move more of the pure geometrical information out of our elements, since all elements can use the same geometrical primitives. This simplifies input code and allows further work on the discretization.

This is a breaking change, but all beam simulations are done with meshpy (I heard), so this should be easy to adapt to, right, @davidrudlstorfer @isteinbrecher?


davidrudlstorfer:
It would probably also make sense to rename the individual test files @sebproell ?

sebproell:
@davidrudlstorfer Sure, once we agree on the naming and changes in the PR. Which renaming pattern would you suggest? `herm2line<x>` --> `hermite_line<x>` ?

davidrudlstorfer:
> @davidrudlstorfer Sure, once we agree on the naming and changes in the PR. Which renaming pattern would you suggest? herm2line<x> --> hermite_line<x> ?

I would probably go for

`herm2line<x>` --> `line<x>_hermitian`

or to keep it more general

`herm<y>line<x>` --> `line<x>_hermite<y>`

isteinbrecher:
For Reissner type beam elements the naming `HERM2LINE3` means that we have a cubic hermitian interpolation of the centreline (herm2) and a rotational interpolation with 3 nodes (line3) - very similar to Taylor-Hood elements in CFD. I am not sure if it makes sense to force this into a standard element type. Maybe it would be better to have something like this:
```yaml
- "5 BEAM3R GENERAL 9 11 10 MAT 1 TRIADS 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 CENTERLINE herm2 ROTATION line3"
```

Related: A similar question also arises for NURBS with non-equal order, where it is not really feasibly to define a "geometry type" for each possible combination of NURBS degrees. So maybe adding a general element type for all non-standard elements would be an alternative also for other implementations.

sebproell:
A little more context: I was thinking about this problem from the perspective of the mesh/geometry in the discretization. I am assuming you can always specify some kind of geometrical primitive that the discretiztaion method is built on. For the `HERM2LINE3`, we use a 3-noded line as primitve and this is the geometrical information that is stored in the discretization. In the implementation, you then decide that you do not use the middle node for the centerline interpolation. As another example, it would be completely valid to read in an 8-noded hex but use a fifth-order Lagrange polynomial for discretiztaion on it. The numbers of geometrical nodes and discretization compute points do not coincide.

@isteinbrecher I like the suggestion, but wonder how the `GENERAL` element type works, especially to convey the connectivity of the nodes? Is it just a list of nodes with interpretation completely up to the user implementation? This means one cannot visualize this geometry without further knowledge. Second point: I am not really familiar with NURBS. Is my geometrical primitive idea applicable in this case?

TL;DR

Wouldn't

```
- "5 BEAM3R LINE3 9 11 10 MAT 1 TRIADS 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 CENTERLINE herm2 ROTATION line3"
```

work for you?

isteinbrecher:
What exactly do you mean with "geometrical primitive"? 

> Is it just a list of nodes with interpretation completely up to the user implementation?

Yes, with the `GENERAL` my intention was to provide some sort of way to define a custom element that is completely up to the user implementation. Actually, the current beam elements are a good example for this, since they do exactly that. In some sense a 2 noded line but also a 3 noded. They can not use any common functionality for `LINE2` or `LINE3` directly. Everything is implemented in the beam framework as visualisation is dependent on the actually physics implementation.

Or maybe it does not have to be `GENERAL` but something like `GENERAL_LINE` then, we know that this is a 1D element with x number of nodes, which is about all information we need about the elements.

---

> Second point: I am not really familiar with NURBS. Is my geometrical primitive idea applicable in this case?

I am not 100% sure, but I don't think so. I had an idea like this a while back
```
- "1 SOLID NURBS_SURFACE 1 2 3 4 MAT 1 DEGREE_XI 2 DEGREE_ETA 2"
- "2 SOLID NURBS_SURFACE 5 6 7 8 9 10 MAT 1 DEGREE_XI 2 DEGREE_ETA 3" 
```
where you can define that you want a NURBS surface and then specify the polynomial degree in the input line (which affects the number of connected nodes/control points). This avoids having to explicitly provide all possible combinations, e.g., `NURSBp2p2`, `NURSBp2p3`, `NURSBp3p2`, `NURSBp3p3`, ... . But this is just something that came to my mind and is not directly related to this PR.

sebproell:
> What exactly do you mean with "geometrical primitive"?

Evrything we have in `FE::CellType`, so:
point, line, quad, tri, tet, hex, wedge, pyramid (in 4C we also add the node number to this)

It sounds like if what the beams are doing is fully custom, we shouldn't give them a `FE::CellType` that is  `line2`,`line3`, etc., but something like `custom` and a number of nodes? This would also work for what I am trying to achieve:

- all the "primitives" (hex8, tet10, etc.) are general and can be read in. In addition, we could visualize the geometry or assign degrees of freedom on these. The discretization understands the connecitivity.
- the "custom" elements can be read in but they are just an ordered list of nodes. Everything else is up to user implementation. The discretization cannot do anything with the custom element since it doesn't understand the connectivity.

@davidrudlstorfer @isteinbrecher So I think there are two options for beams:

- Do you want to continue treating the beams as e.g. `line3` in the discretization? Then I would like to see the input more or less like I proposed it in this PR. 
  ```
  - "5 BEAM3R LINE3 9 11 10 MAT 1 TRIADS 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 CENTERLINE herm2 ROTATION line3"
  ```
  Especially, we would need the `LINE3 <id> <id> <id>` in the beginning. I don't care about the rest of the line.
- Do you want to use a custom cell type resulting in something like (just one idea):
  ```
  - "5 BEAM3R CUSTOM NODES 3 IDS 9 11 10 MAT 1 TRIADS 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 CENTERLINE herm2 ROTATION line3"
  ```

In the element implementation, both approaches would give almost the same result. One difference I can think of is that `line3` is explicit, but `custom` might need to ask for the number of nodes sometimes.

davidrudlstorfer:
I would go for 

`- "5 BEAM3R LINE3 9 11 10 MAT 1 TRIADS 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 CENTERLINE herm2 ROTATION line3"`

The underlying geometry is still a three noded line element.

isteinbrecher:
I would go with
```
- "5 BEAM3R CUSTOM NODES 3 IDS 9 11 10 MAT 1 TRIADS 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 CENTERLINE herm2 ROTATION line3"
```
yes, the element is still a line with 3 nodes, but that is about everything we have in common with classic `line3` elements.

sebproell:
@isteinbrecher @davidrudlstorfer Could you decide on one of the two options together?

sebproell:
I thought about this a little more and I believe you currently use the `line`-like geometry quite often. I find around 100 hits for the different `line` cell types in `beam3`. Using `line` geometries instead of `custom` seems like the smaller change for now. If you want to isolate beams from the discretization in the future (and I am not saying you should), you could always introduce the `custom` geometry. I don't think you are losing anything by keeping it a `line` element for now. To proceed with a more uniform treatment of the elements which will help the input and FourCIPP, could we go with my change here?

davidrudlstorfer:
> I thought about this a little more and I believe you currently use the line-like geometry quite often. I find around 100 hits for the different line cell types in beam3. Using line geometries instead of custom seems like the smaller change for now. If you want to isolate beams from the discretization in the future (and I am not saying you should), you could always introduce the custom geometry. I don't think you are losing anything by keeping it a line element for now. To proceed with a more uniform treatment of the elements which will help the input and FourCIPP, could we go with my change here?

I would go with the LINE3 approach. In addition I think the LINE3 approach is easier because we also encode the number of nodes into the element. With CUSTOM we need the additional Node identifier to just determine the number of nodes. In my opinion this is somehow redundant. All in all, the beam element has three nodes and is a line, what we do with the nodes in the end is an implementation detail on our side

isteinbrecher:
I think we should go with the `CUSTOM` approach. We are basically using exactly that at the moment and there is no need (nor ist it helpful) to force these elements into standard types that are not applicable.

Can't we use
```
- "5 BEAM3R CUSTOM 9 11 10 MAT 1 TRIADS 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 CENTERLINE herm2 ROTATION line3"
```
without having to specify the number of nodes explicitly? Should be clear from the input.

sebproell:
@isteinbrecher and I just discussed the following plan of attack:

Much of the discussion about how exactly the beam element fits into the discretization stems from the fact that we bunch up the number of nodes/dofs with the geometric primitve. For the pure geometry, it is often enough to know that something is a `hex` element and the number of nodes (which mostly translates to DOFs) is another aspect. Thus, we agree that we eventually want an input line like (names up for debate):

`270 BEAM3R LINE 3 IDS 883 885 884 MAT 2 TRIADS 0 -0.482513295022 0 0 -0.482513295022 0 0 -0.482513295022 0 HERMITE_CENTERLINE true`

Notice the space between `LINE` and `3`.
This PR helps us in its current form because I can then unify all geometry input to a few primitive types. Next, I will introduce the basic primitves `hex`, `tet` etc. and then we can remap the beam input to the style above. We should probably keep `HEX8`, `TET4` etc as valid input, but internally we can represent it with the more flexible structure if this is useful.

Note: we believe this would be a good idea since you can have arbitrary order elements (@isteinbrecher mentioned NURBS elements like `QUAD NODES_X 2 NODES_Y 3 IDS 1 2 3 4 5 6`) where hardcoding the nodes/dofs into the geometric primitive is not feasible/inflexible.

sebproell:
So this PR would be ready :)

davidrudlstorfer (approved):

isteinbrecher:
Thanks for the summation @sebproell!

@davidrudlstorfer regarding the test names, I would stick with the current naming scheme and maybe reconsider this if we refactor the beam implementation at some point - also I would stick with the current object and class names in MeshPy.

Issue #754, opened 2025-05-09
[Teko] Remove `amgnxn` framework


maxfirmbach:
## Description

In the near future we should think about removing the `amgnxn` framework. Most, if not all, functionality is already available open-source in `MueLu` and mostly in `Teko` and if not should be added there. In addition it already proves to be difficult to maintain the code, also with the transition to `Tpetra`.

## Possible Solution and Definition of Done

Port remaining physics using `amgnxn`:
- [x] elch (see #1130)
- [ ] elch (dofsplit)
- [x] sti (see #1143)
- [x] ssi (see #1430)
- [ ] ssi (dofsplit)
- [x] ssti (see #1516)

Delete the `amgnxn` framework:
- [ ] Delete the respective code related to the `amgnxn` implementation   (see #1434)
- [ ] Delete the input procedure and enum classes related to `amgnxn` 

## Interested Parties

@c-p-schmidt

maxfirmbach:
@c-p-schmidt Would you mind documenting how many tests are still using `AMGnxn` for the dofsplit cases here? Thanks!

c-p-schmidt:
> [@c-p-schmidt](https://github.com/c-p-schmidt) Would you mind documenting how many tests are still using `AMGnxn` for the dofsplit cases here? Thanks!

The relevant tests are:
- elch_3D_tet4_s2i_butlervolmer_BGS-AMG_4x4.4C.yaml
- ssi_mono_3D_tet4_elch_s2i_butlervolmer_manifold_BGS-AMG_11x11.4C.yaml

PR #892, opened 2025-06-13
Fix structure new solver factory

c-p-schmidt:
## Description and Context
While working on the usage of Teko in the last remaining `AMGnxn` or `multigrid_nxn`, I believe I have realized an inconsistency in the current usage of `Core::LinearSolver::Parameters::compute_solver_parameters`. The creation of the nullspace can be influenced by a map, e.g., if the source of the block structure of your matrix is a subdivision of the geometry (like we have it in electrochemistry problems, this is why I stumbled over this).

However, so far, only one map has been used, even though the methods that consume this map have different requirements (can be seen from the methods definition and from the default maps that are handed in, if no external map has been provided):
- build_node_coordinates -> requires a node-based map
- Core::FE::compute_null_space -> requires a dof-based map

Now, I provide two maps, one node-based and one dof-based, and consume the respective in the relevant call. There is one test case that uses this functionality: `beam3eb_static_beam_to_solid_volume_meshtying_cube.4C.yaml`. It passes before the change and after the change, maybe because it only subdivides into beam and solid. However, when I did that locally for electrochemistry problems (changes not yet completed), I got the wrong dimensionality of the coordinates, due to the map mismatch I described above.

The relevant changes concerning the described map mismatch are only in the first commit. The second commit just does some additional cleanup.

I'll also mark the relevant lines of code in a diff afterwards.

related #754

### DIFF
--- a/src/core/linear_solver/src/method/4C_linear_solver_method_parameters.cpp
+++ b/src/core/linear_solver/src/method/4C_linear_solver_method_parameters.cpp
@@ -21,21 +21,22 @@ FOUR_C_NAMESPACE_OPEN
 //----------------------------------------------------------------------------------
 //----------------------------------------------------------------------------------
 void Core::LinearSolver::Parameters::compute_solver_parameters(
-    Core::FE::Discretization& dis, Teuchos::ParameterList& solverlist)
+    const Core::FE::Discretization& dis, Teuchos::ParameterList& solverlist)
 {
-  std::shared_ptr<Core::LinAlg::Map> nullspaceMap =
+  const auto nullspace_node_map =
       solverlist.get<std::shared_ptr<Core::LinAlg::Map>>("null space: node map", nullptr);
   auto nullspace_dof_map =
       solverlist.get<std::shared_ptr<Core::LinAlg::Map>>("null space: dof map", nullptr);
 
   int numdf = 1;
   int dimns = 1;
-  int nv = 0;
-  int np = 0;
 
   // set parameter information for solver
   {
-    if (nullspaceMap == nullptr and dis.num_my_row_nodes() > 0)
+    int nv = 0;
+    int np = 0;
+
+    if (nullspace_node_map == nullptr and dis.num_my_row_nodes() > 0)
     {
       // no map given, just grab the block information on the first element that appears
       Core::Elements::Element* dwele = dis.l_row_element(0);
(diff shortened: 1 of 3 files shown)


c-p-schmidt:
By the way, I also do not know why the map (and now maps) are handed in using a parameter-list in a hide-and-seek manner. ;-)

This should also be easily and more clearly reflected by std::optional, if the map is not required by Trilinos internally as well. Maybe you already know if this is the case @maxfirmbach .

src/core/linear_solver/src/method/4C_linear_solver_method_parameters.cpp
```diff
@@ -76,24 +77,24 @@ void Core::LinearSolver::Parameters::compute_solver_parameters(
   // set coordinate information
   {
     std::shared_ptr<Core::LinAlg::MultiVector<double>> coordinates;
-    if (nullspaceMap == nullptr)
+    if (nullspace_node_map == nullptr)
       coordinates = dis.build_node_coordinates();
     else
-      coordinates = dis.build_node_coordinates(nullspaceMap);
+      coordinates = dis.build_node_coordinates(nullspace_node_map);
```
c-p-schmidt, line 83: This method uses a node-based map (see docu, or simply what happens in the method)

src/core/linear_solver/src/method/4C_linear_solver_method_parameters.cpp
```diff
@@ -76,24 +77,24 @@ void Core::LinearSolver::Parameters::compute_solver_parameters(
...
+    if (nullspace_dof_map == nullptr)
     {
       // if no map is given, we calculate the nullspace on the map describing the
       // whole discretization
-      nullspaceMap = std::make_shared<Core::LinAlg::Map>(*dis.dof_row_map());
+      nullspace_dof_map = std::make_shared<Core::LinAlg::Map>(*dis.dof_row_map());
     }
 
-    auto nullspace = Core::FE::compute_null_space(dis, numdf, dimns, *nullspaceMap);
+    const auto nullspace = Core::FE::compute_null_space(dis, numdf, dimns, *nullspace_dof_map);
```
c-p-schmidt, line 97: This method used the same map originally, now a different one, but consumes a dof-based map (see docu, or the dof-row map that is used, if no external map is handed in in line 94).

maxfirmbach:
@c-p-schmidt I will have a look at it tomorrow, but on the first glance it looks good, shame on me ;).

src/core/linear_solver/src/method/4C_linear_solver_method_parameters.cpp
maxfirmbach: Yep, you're right here :+1:.

mayrmt (approved):

PR #894, opened 2025-06-15
Simplify selection in input files

sebproell:
Depends on #890 

Simplify the way a selection looks in the input file by kicking the selector value. Instead, we now require every possible choice to be named exactly like one of the enum constants you are selecting over (this is checked).

Old

```yaml
time_integration:
  type: OST
  OST:
    theta: 0.66
```

New

```yaml
time_integration:
  OST:
    theta: 0.66
```

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.cpp
+++ b/src/core/io/src/4C_io_input_spec_builders.cpp
@@ -304,6 +304,10 @@ namespace
       case InputSpecType::selection:
       {
         print_match_state();
+        if (entry.state == MatchEntry::State::partial && !entry.additional_info.empty())
+        {
+          out << " " << entry.additional_info;
+        }
         out << '\n';
         if (entry.state == MatchEntry::State::partial)
         {
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -579,21 +579,24 @@ namespace Core::IO
       from_file,
     };
 
+    /**
+     * Create an InputField from one of the supported ways.
+     */
     template <typename T>
     void input_field_callback(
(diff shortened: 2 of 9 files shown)


sebproell:
@gilrrei I modified the schema code in a way that seems to work. Please let me know how you would do it (or push it here).

gilrrei:
@sebproell 

> @gilrrei I modified the schema code in a way that seems to work. Please let me know how you would do it (or push it here).

I think it should work, if I understood correctly you remove the selector and when creating the oneOf. Are the specs of the choices groups now with this change?

sebproell:
> Are the specs of the choices groups now with this change?

@gilrrei They must be something with a name/key. So every choices is either a group, a single parameter, or another selection. All of these are complete schema objects.

gilrrei:
> > Are the specs of the choices groups now with this change?
> 
> @gilrrei They must be something with a name/key. So every choices is either a group, a single parameter, or another selection. All of these are complete schema objects.

So this allows something like

```yaml
selection_quantity:
    selected_branch: 0.5
```

If so, this breaks the current idea behind the selection. I guess this makes it more convenient for future work such as the spatially varying parameters.

bwirthl (approved):
@sebproell Do I understand it correctly that the key in the input file has to be the same as the enum in the code? Is this something we want to enforce (comparing the user and developer perspective)? But, to be honest, clean code requires enum names that are understandable "even" to a user. Hence, I am fine with it.

sebproell:
>  Do I understand it correctly that the key in the input file has to be the same as the enum in the code?

@bwirthl Exactly! This is already enforced and you will get a runtime error if you choose wrong names here. Since we construct the full input after building 4C to generate the schema, you will see this error right away when building.

PR #896, opened 2025-06-16
Bugfix in input_field_callback

maxiludwig:
Fixing bug introduced with #890. 

FYI @maxfirmbach

PR #897, opened 2025-06-16
Fix `input_field_callback()`

maxfirmbach:
## Description and Context

I think for the `value` case in the `input_field_callback()` we want to use the template type instead of storing into a hard-coded `double`.

Follows #890


maxfirmbach:
Closing in favor of #896.

PR #898, opened 2025-06-16
Implement SVD for tensors

amgebauer:
A small routine computing the singular value decomposition of 2-tensors (using `Teuchos::LAPACK`)

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_tensor_svd.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_tensor_svd.hpp
@@ -1,0 +1,103 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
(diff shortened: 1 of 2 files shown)


src/core/linalg/src/dense/4C_linalg_tensor_svd.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+namespace Core::LinAlg
+{
+  /*!
+   * @brief Computes the singular value decomposition of a 2-tensor
+   *
+   * @tparam Tensor
+   * @return A tuple of a @p Q, @p s and @p VT, where @p Q and @p VT are orthogonal 2-tensors and @p
+   * s is a std::array containing the singular valurs, such that
+   * t = Q * make_rectangular_diagonal_matrix(s) * VT.
```
sebproell: Maybe create a struct for that return type so you can name the arguments?

sebproell (approved):

src/core/linalg/src/dense/4C_linalg_tensor_svd.hpp
amgebauer: Good point!

sebproell (approved):

PR #899, opened 2025-06-16
[NOX] Remove the derived single step solver class

vovannikov:
## Description and Context
The single step solver was a bit customized in 4C in the output part. These were really unnecessary, so I removed this derived class completely and also cleaned up interfaces a bit when working on this.

Next, I will create a wrapper for the NOX solvers such that the original Trilinos NOX solver classes are not inherited from.

#359


c-p-schmidt (approved):

mayrmt (approved):
Thanks!

PR #901, opened 2025-06-16
Add artery flow rate to function coupling

bwirthl:
This PR adds the flow rate of the reduced-dimensional arteries to the function coupling in the porofluid-elasticity framework with scalar transport.

### DIFF
--- a/src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.cpp
+++ b/src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.cpp
@@ -1634,9 +1634,10 @@ void PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
 
     const double jacobian_determinant_artery = current_segment_length / 2.0;
     evaluate_function_coupling(current_gp_weight, shape_functions_artery,
-        shape_functions_homogenized, jacobian_determinant_artery, *ele_rhs_artery,
-        *ele_rhs_homogenized, *ele_matrix_artery_artery, *ele_matrix_artery_homogenized,
-        *ele_matrix_homogenized_artery, *ele_matrix_homogenized_homogenized, integrated_diameter);
+        shape_functions_artery_deriv, shape_functions_homogenized, jacobian_determinant_artery,
+        *ele_rhs_artery, *ele_rhs_homogenized, *ele_matrix_artery_artery,
+        *ele_matrix_artery_homogenized, *ele_matrix_homogenized_artery,
+        *ele_matrix_homogenized_homogenized, integrated_diameter);
   }
 }
 
@@ -2059,6 +2060,7 @@ template <Core::FE::CellType dis_type_artery, Core::FE::CellType dis_type_homoge
 void PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
     dis_type_homogenized, dim>::evaluate_function_coupling(const double& gp_weight,
     const Core::LinAlg::Matrix<1, num_nodes_artery_>& shape_functions_artery,
+    const Core::LinAlg::Matrix<1, num_nodes_artery_>& shape_functions_artery_deriv,
     const Core::LinAlg::Matrix<1, num_nodes_homogenized_>& shape_functions_homogenized,
     const double& jacobi, Core::LinAlg::SerialDenseVector& ele_rhs_artery,
     Core::LinAlg::SerialDenseVector& ele_rhs_homogenized,
@@ -2071,9 +2073,13 @@ void PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
   std::vector artery_scalar_np_at_gp(num_scalars_artery_, 0.0);
   std::vector homogenized_scalar_np_at_gp(num_scalars_homogenized_, 0.0);
   double artery_pressure_at_gp = 0.0;
+  double artery_pressure_gradient_at_gp = 0.0;
 
   // get artery values at GP
-  get_artery_values_at_gp(shape_functions_artery, artery_pressure_at_gp, artery_scalar_np_at_gp);
+  get_artery_values_at_gp(shape_functions_artery, shape_functions_artery_deriv,
+      artery_pressure_at_gp, artery_pressure_gradient_at_gp, artery_scalar_np_at_gp);
+  const auto artery_element_flow_rate = evaluate_artery_flow_rate(artery_pressure_gradient_at_gp);
+
   // get scatra values at GP
   get_homogenized_scalar_values_at_gp(shape_functions_homogenized, homogenized_scalar_np_at_gp);
   // NOTE: values of fluid phases held by managers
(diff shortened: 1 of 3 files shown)


src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.cpp
```diff
@@ -2703,6 +2718,22 @@ void PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
...
+template <Core::FE::CellType dis_type_artery, Core::FE::CellType dis_type_homogenized, int dim>
+double PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
+    dis_type_homogenized, dim>::evaluate_artery_flow_rate(const double artery_pressure_gradient)
+    const
+{
+  const double radius = artery_diameter_at_gp_ / 2.0;
+
+  // Hagen-Poiseuille equation
+  const double flow_rate =
+      -artery_pressure_gradient * M_PI * pow(radius, 4) / (8.0 * artery_material_->viscosity());
```
sebproell:
```suggestion
      -artery_pressure_gradient * M_PI * radius* radius * radius * radius / (8.0 * artery_material_->viscosity());
```

minor: an integer power is a super simple operation, but going through `std::pow` will use a general algorithm for float exponents.

src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.cpp
```diff
@@ -2097,7 +2103,7 @@ void PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
       double function_value = 0.0;
 
       // evaluate and assemble
-      evaluate_function_and_deriv(*function_vector_[0][i_art], artery_pressure_at_gp,
+      evaluate_function_and_deriv(*function_vector_[0][i_art], artery_pressure_at_gp, flow_rate,
```
lkoeglmeier:
```suggestion
      evaluate_function_and_deriv(*function_vector_[0][i_art], artery_pressure_at_gp, artery_element_flow_rate,
```

tests/input_files/poromultielastscatra_2D_quad4_linebased_artery_coupling_mono.4C.yaml
```diff
@@ -273,13 +273,13 @@ FUNCT8:
 FUNCT9:
   - SYMBOLIC_FUNCTION_OF_SPACE_TIME: "x*y*t"
 FUNCT10:
-  - VARFUNCTION: "pi*D*0.1*(p_art-p3)"
+  - VARFUNCTION: "pi*D*0.1*(p_art-p3)+Q*1.0e-7"
```
lkoeglmeier:
```suggestion
  - VARFUNCTION: "pi*D*0.1*(p_art-p3)+Q_art*1.0e-7"
```
Maybe to make it clear that it is the flow in the arteries and not in the homogenized domain?

src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.cpp
```diff
@@ -2703,6 +2718,22 @@ void PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
...
 }
 
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+template <Core::FE::CellType dis_type_artery, Core::FE::CellType dis_type_homogenized, int dim>
+double PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
+    dis_type_homogenized, dim>::evaluate_artery_flow_rate(const double artery_pressure_gradient)
+    const
+{
+  const double radius = artery_diameter_at_gp_ / 2.0;
```
lkoeglmeier, line 2731: I think using the diameter at the GP will lead to a change in the flow rate inside the element according to the varying diameter. Would it make sense to use the average element diameter instead, as suggested in the thesis of Kremheller in equation (4.120) at page 107? Then the velocity and flow inside the 1D element would still be constant despite the varying diameter?

src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.cpp
```diff
@@ -2459,6 +2468,9 @@ void PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
       FOUR_C_THROW("Unknown coupling type.");
       break;
   }
+  // scale artery pressure gradient with the inverse Jacobian matrix
+  // in case of 1D line elements, this is just 2 / L
+  artery_pressure_gradient *= 2.0 / artery_ele_length_;
```
lkoeglmeier: Should we maybe add the scaling to the calculation of the `artery_pressure_gradient` to make it clearer where the scaling comes from?

src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.cpp
bwirthl: This would indeed be more consistent. But, as it is quite complex, we will post-pone it to a follow-up PR.

src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.cpp
```diff
@@ -2703,6 +2721,22 @@ void PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
...
+ *----------------------------------------------------------------------*/
+template <Core::FE::CellType dis_type_artery, Core::FE::CellType dis_type_homogenized, int dim>
+double PoroPressureBased::PorofluidElastScatraArteryCouplingPair<dis_type_artery,
+    dis_type_homogenized, dim>::evaluate_artery_flow_rate(const double artery_pressure_gradient)
+    const
+{
+  const double radius = artery_diameter_at_gp_ / 2.0;
+
+  // Hagen-Poiseuille equation
+  const double artery_element_flow_rate = -artery_pressure_gradient * M_PI * radius * radius *
```
c-p-schmidt, line 2734: No need to change it in this pull request, but maybe interesting for future development. Several constants are available in the standard library since C++20, see https://en.cppreference.com/w/cpp/numeric/constants.html.

sebproell: @c-p-schmidt True, and this would be a nice candidate for a mass-replace. I prefer to roll out such changes on the whole code base at once, as people keep on copying/modeling the old code, otherwise.

PR #902, opened 2025-06-16
Allow FAD-types in tensors

amgebauer:
Achieving this was actually more involved than hoped.

During review, pay special attention to the new meta-utility header `src/core/utils/src/numerics/4C_utils_fad_meta.hpp`

Our tensors automatically deduce the result type of the operations. This is problematic for the Sacado-Fad types, which build upon lazy evaluation and need to be casted to their base-type. Obtaining the base-type was quite tricky, especially since I needed to avoid loading the heavy Sacado-header in the tensor-header.

A small caveat remains: We require that all scalar-types are the same if a FAD-type is detected. For some reason, automatic differentiation does not work otherwise. I ensured that other usages don't compile

```c++
Tensor<FAD, 2> fad_tensor = {{FAD(2, 0, 1.0), FAD(2, 1, 2.0)}};
// auto result = 2 * fad_tensor; // does not compile
auto result = 2.0 * fad_tensor; // compiles
```

### DIFF
--- a/src/core/utils/src/numerics/4C_utils_fad_meta.hpp
+++ b/src/core/utils/src/numerics/4C_utils_fad_meta.hpp
@@ -1,0 +1,183 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_UTILS_FAD_META_HPP
+#define FOUR_C_UTILS_FAD_META_HPP
+
+#include "4C_config.hpp"
+
+#include <concepts>
+#include <functional>
+#include <type_traits>
+
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Core::FADUtils
+{
+  /*!
+   * @brief A concept that checks whether a type is a Sacado-Fad type (without actually needing to
+   * include the heavy Sacado header).
+   *
+   * @tparam Scalar
+   */
+  template <typename Scalar>
+  concept SacadoFadType = requires(Scalar a) {
+    { a.val() } -> std::convertible_to<typename Scalar::value_type>;
+    { a.dx(0) } -> std::convertible_to<typename Scalar::value_type>;
+  };
+
+  namespace Internal
+  {
+    // Note: A Fad-expression has a ExprT1 or a ExprT2 typedef if it is an Expression (and not a
+    // pure FAD-Type)
+    template <typename T>
+    concept HasExpr1Type = requires { typename T::ExprT1; };
+    template <typename T>
+    concept HasExpr2Type = requires { typename T::ExprT2; };
+
+    /*!
+     * @brief A small helper meta-utility to extract the base type of a FAD expression.
+     *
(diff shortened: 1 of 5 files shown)


src/core/utils/src/numerics/4C_utils_fad_meta.hpp
```diff
@@ -0,0 +1,197 @@
...
+namespace Core::FADUtils
+{
+  /*!
+   * @brief A concept that checks whether a type is a Sacado-Fad type (without actually needing to
+   * include the heavy Sacado header).
+   *
+   * @tparam Scalar
+   */
+  template <typename Scalar>
+  concept AutoDiffType = requires(Scalar a) {
```
sebproell: Should this be called `SacadoFadType` then? The docs state it only works for this case.

src/core/utils/src/numerics/4C_utils_fad_meta.hpp
```diff
@@ -0,0 +1,197 @@
...
+    // Note: A Fad-expression has a ExprT1 or a ExprT2 typedef if it is an Expression (and not a
+    // pure FAD-Type)
+    template <class, class = void>
+    struct HasExpr1Type : std::false_type
+    {
+    };
+    template <class T>
+    struct HasExpr1Type<T, std::void_t<typename T::ExprT1>> : std::true_type
+    {
+    };
```
sebproell:
```suggestion
    template <typename T>
    concept HasExpr1Type = requires { typename T::ExprT1; };
```

I think this should work.

src/core/utils/src/numerics/4C_utils_fad_meta.hpp
```diff
@@ -0,0 +1,197 @@
...
+    template <typename Expr>
+      requires(AutoDiffType<Expr> && !HasExpr1Type<Expr>::value && HasExpr2Type<Expr>::value)
+    struct ExtractAutoDiffBaseType<Expr>
+    {
+      using type = ExtractAutoDiffBaseType<typename Expr::ExprT2>::type;
+    };
+  }  // namespace Internal
+
+  /*!
+   * @brief This template meta extracts the base type of a FAD expression from Sacado.
```
sebproell:
```suggestion
   * @brief This template metafunction extracts the base type of a FAD expression from Sacado.
```
?

src/core/utils/src/numerics/4C_utils_fad_meta.hpp
```diff
@@ -0,0 +1,197 @@
...
+   *
+   * Sacado-Fad internally uses lazy evaluation so that the use of auto as a return type is
+   * dangerous and fails in certain situations. This class can extract the base type of a FAD
+   * expression that can be used as an explicit return type.
+   *
+   * @note This meta-utilities don't rely on the Sacado-header and are, therefore, lightweight by
+   * inclusion.
```
sebproell:
```suggestion
   * @note These meta-utilities don't rely on the Sacado-header and are, therefore, lightweight includes.
```

src/core/utils/src/numerics/4C_utils_fad_meta.hpp
amgebauer: I knew there is a nice c++20 way, but my first try didn't work. Your suggestion works nicely :+1:

sebproell (approved):

PR #903, opened 2025-06-16
WIP: Enable Teko solver for porofluid-elasticity-scatra problems with arteries

bwirthl:
I am still struggling to run our poroelasticity-scattering problems with Teko. In the case of additional coupling with arteries, there was a bug when building the nullspace. I tried to solve this, but I am not at all sure whether my solution is correct. I only know that the previous error has been resolved. This problem type was not tested with an iterative solver so far.

Either way, my new test case `tests/input_files/poromultielastscatra_2D_quad4_linebased_artery_coupling_teko.4C.yaml` is not converging. I don't know whether this is due to the nullspace problem or due to solver settings. 

@maxfirmbach I would be very thankful for your help! Many thanks in advance!


maxfirmbach:
@bwirthl I already did some minor testing at some point, without much success. Even if each field uses a direct solver, thing's don't converge ... so I wonder if there is a more fundamental issue.

maxfirmbach:
@bwirthl Is the input file the exact same one that was also used for `AMGnxn` ?

bwirthl:
@maxfirmbach The test is the same as `tests/input_files/poromultielastscatra_2D_quad4_linebased_artery_coupling_mono.4C.yaml` which is solved with UMFPACK. As far as I remember, we had no test with arteries using `AMGnxn` which is the reason why you did not come across these cases when updating to Teko. 

And yes, I also have the impression that there might be a more fundamental issue because I have tried a lot of different settings and none is even close to converging. I also have example without arteries which I cannot get to convergence at all.

This also concerns @lkoeglmeier and @shervas as all of us are using the same problem type in 4C.

maxfirmbach:
@bwirthl Maybe having a look at the examples without arteries might be easier to begin with.

bwirthl:
@maxfirmbach All simple test cases without arteries that I have tested work with Teko. I also have one larger simulation that converges nicely with Teko, other similar ones (all without arteries) however do not converge at all. Maybe that is because of AMG-BGS vs BGS-AMG or something else. Any suggestions on how I can tell what causes the problem?

shervas:
Thanks @bwirthl. It would be really helpful, I also need the solver for my simulations.

maxfirmbach:
@bwirthl @shervas It's very hard to tell and give a specific reason why certain ones work nice and others don't. Often one parameter, which might change some underlying physics is enough to result in divergence. So even if simulations are **similar**, they might not behave completely identical. So depending on the parameters chosen it might be worth analyzing if there's a change in physics or numerics.

To my experience AMG-BGS vs BGS-AMG shouldn't make a difference for convergence or divergence, but only in the speed of convergence, where AMG-BGS might converge faster at the cost of taking more time.

maxfirmbach:
To give a small example: introducing a penalty parameter is always difficult for the iterative linear solver. The penalty parameter with a value chosen to be high so penalty >> 1 destroys almost all nice features of a linear system and makes things tricky. So if you would change the parameter related how mortar coupling is done from a condensed formulation to a penalty based on, things go wrong very quickly. Now one might think, why did it work before with `AMGnxn`? I would assume it's luck.

bwirthl:
@maxfirmbach I now managed to get my test case to convergence simply by increasing the number of iterations. Probably, it is not converging nicely but it is converging. I just had no idea how to adapt the number of iterations of the linear solver. 

@mayrmt Many thanks again for your talk and your patience in answering our questions in Freising! It was obviously very helpful!

FYI @c-p-schmidt @lkoeglmeier @shervas

tests/input_files/xml/linear_solver/iterative_gmres_porofluid_pressure_based.xml
```diff
@@ -0,0 +1,20 @@
+<ParameterList name="Belos">
+  <!--
+    For iterative solver using the generalized minimal residual method, these are the recommended settings for Belos.
+    -->
+  <ParameterList name="GMRES">
+    <!-- ===========  GENERAL ================ -->
+    <Parameter name="Maximum Iterations" type="int" value="1000"/>
```
c-p-schmidt, line 7: Do you really want 1000 here?

c-p-schmidt (approved):

tests/input_files/xml/linear_solver/iterative_gmres_porofluid_pressure_based.xml
bwirthl: For now, I do as long as I don't have better solver settings. At least, I can now test whether Teko is running at all with arteries (e.g., concerning the nullspace etc. which was not tested before and therefore not detected).

bwirthl:
@maxfirmbach @c-p-schmidt I'd suggest to merge this PR separately and first because without it we cannot run any tests with arteries and teko. As a next step, we can then optimize the solver as we discussed and change the order of blocks/GIDs.

c-p-schmidt:
> @maxfirmbach @c-p-schmidt I'd suggest to merge this PR separately and first because without it we cannot run any tests with arteries and teko. As a next step, we can then optimize the solver as we discussed and change the order of blocks/GIDs.

The suggested approach is fine for me.

tests/input_files/xml/linear_solver/iterative_gmres_porofluid_pressure_based.xml
```diff
@@ -0,0 +1,20 @@
...
+    <Parameter name="Convergence Tolerance" type="double" value="1e-7"/>
```
mayrmt, line 8: In practice, things might work with choosing a less strict tolerance here -- at least in my experience in other coupled problems. For now, I'd leave it as is, but recommend that you try other values in practical examples and then maybe adapt it later.

bwirthl: I totally agree. Thanks for approving.

PR #893, opened 2025-06-14
[thermo] Add heterogenous thermo and material vector option

maxfirmbach:
## Description and Context

This PR is my initial approach to add a heterogeneous material option to Fourier's law by providing a map.pattern file and extract the element-wise conductivity tensor. Still it's possible to provide a constant material vector.
The second addition is to assemble a node-wise material vector for further use in the linear solver.
On the way the StVenant-structure-thermo material is refactored.

Currently this is still a draft as I would like to gather some opinions.
@lauraengelhardt @amgebauer I shamelessly copied some of the mechanism from the muscle material and mixture rule

PR #904, opened 2025-06-16
Allow to store input_field to structs

sebproell:
The last missing piece to make the input_field() as powerful as all other parameters: allow to store it in struct members. Note that I need to make the InputField default constructible to allow a default-constructible struct with such a member. 

@maxiludwig I fixed a few small things along the way. Most importantly, the `input_field` should really take its own set of options and not use the ones from `parameter`. I just made `input_field` and `selection` parameters required for now as I am not sure what the best logic would be to make them optional.

This PR also fixes the duplicated access, i.e. no more `group("stiffness").get<>("stiffness");` but just `get<>("stiffness");`

@maxfirmbach Let's watch out for an git-invisible conflict with #893 since the access will change as indicated above.

### DIFF
--- a/src/core/io/src/4C_io_input_field.hpp
+++ b/src/core/io/src/4C_io_input_field.hpp
@@ -31,7 +31,13 @@ namespace Core::IO
   {
    public:
     using IndexType = int;
-    using StorageType = std::variant<T, std::unordered_map<IndexType, T>>;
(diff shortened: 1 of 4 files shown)


maxfirmbach:
We can merge this PR before mine, if think there will be some minor conflicts.

maxiludwig (approved):

PR #907, opened 2025-06-17
No longer accept .dat input

sebproell:
In line with #598, this PR will disallow .dat input from now on. We still keep the `--to-yaml` around although the user base is probably small (only people who actually live close to main, and for some reason did not convert their files yet). I would remove this flag soonish (~maybe in 2 weeks?~ in the next days). Afterwards, you can still convert to yaml by pulling the docker container `4C:dat-converter` which will be able to convert files compatible with the current state of main.

~Technical detail: conversion from dat to yaml is still kind of covered by unit tests. So I feel confident to remove the tests for conversion through `--to-yaml`. This feature will not work anyway for arbitrary new input files, if people use more and more features that dat does not support (we saw that already). IMO, we would be testing a vaguely specified file, so I suggest simply removing the tests.~


amgebauer (dismissed):

isteinbrecher:
Didn't we agree to do this in the beginning of July and after all points of "Phase1" are completed?

sebproell:
@isteinbrecher We can also wait for two weeks, although I don't see why. I personally interpreted the June 30 as a motivation to pull through with dat deprecation, not a fixed date on which something specific happens. Yes, it is actually allowed to finish something before a deadline :) The important aspect, IMO, is that we agreed on things that should happen before kicking dat. To the best of my knowledge these things have happened, see also the checklist in #598. The open points in "Phase 1" are minor improvements in the documentation that I will add  (and we actually need much more doc improvement than listed there).

I think I have made this point a few times but let me repeat it here: I truly believe that at this point we are actively misleading people by keeping dat around in a convenient manner. So let's make it a bit more inconvenient by forcing users to convert their files, while they still can. This might prevent more drastic surprises for them.

isteinbrecher:
I believe it’s important that we stick to the plan we all agreed on, both in terms of timing and progress (early July **and** after Phase 1 is completed). If there’s a need for re-evaluating that timeline, I think we should reopen the discussion as a group, rather than having individuals act ahead of the agreed plan. Otherwise, we risk undermining the consensus we worked hard to build.

If, as you say, the remaining Phase 1 items are minor, then I don’t see an issue with completing them first as planned before moving ahead with removing dat support.

sebproell:
Let's merge on July 1, then. Until then, all items of phase 1 will be completed (except that you can always write more docs, see e.g. #910, which has nice additions that were not exactly stated like this in the roadmap ).

gilrrei:
Currently in QUEENS, we require dat-files since we have an old but working code to construct one material per element with varying material parameters. Instead of reimplementing this or requiring an additional dependency in QUEENS I would prefer to directly move to the new feature of prescribing materials elementwise with an external file. @sebproell are there already working examples for this? If so, I would use one of them to use it in the checks in QUEENS

maxfirmbach:
> Currently in QUEENS, we require dat-files since we have an old but working code to construct one material per element with varying material parameters. Instead of reimplementing this or requiring an additional dependency in QUEENS I would prefer to directly move to the new feature of prescribing materials elementwise with an external file. @sebproell are there already working examples for this? If so, I would use one of them to use it in the checks in QUEENS

@gilrrei There is a working example e.g. `thermo3D_annulus_elementwise_anisotropic_statics.4C.yaml`, which uses the conductivity field given in `conductivity_input_field.json`.

apps/global_full/4C_global_full_control.cpp
```diff
@@ -59,12 +48,11 @@ void ntam(int argc, char* argv[])
 
   /* input phase, input of all information */
 
-  if (Core::Communication::my_mpi_rank(gcomm) == 0)
+  if (inputfile_name.ends_with(".dat"))
```
maxfirmbach, line 51: How long do we want to keep this message?

sebproell: I don't know. My original idea was to use this PR as an intermediate state before full dat removal (see discussion above) to maybe help a few stragglers realize how nice the new input is and that they need to migrate. We already agreed that dat is going to be removed after June 30. Since we are planning to merge this PR only then, I can also adapt the logic and outright reject dat.

isteinbrecher: > We already agreed that dat is going to be removed after June 30.

That is not what we agreed upon. We agreed that we aim to complete phase1 until June30. Only after phase1 is complete, we agreed to drop `dat` support. As far as I can see, phase1 is not completed.

maxfirmbach: What's missing then? Where and how is "phase1" defined?

mayrmt: @maxfirmbach Phase 1 is defined in #598.

maxfirmbach: So, that seems pretty complete, or am I missing something?

sebproell: I'll rebase this PR later.

sebproell: @maxfirmbach Let's actually keep this message for now. After merging this PR, I will remove more of the dat infrastructure in `InputFile` etc.

sebproell:
Ready.

davidrudlstorfer (approved):
RIP .dat ☠️ 🥳🎊🎉

mayrmt:
@4C-multiphysics/developers I have been asked by a collaborator for a tool to convert dat files to yaml files. Where can I find the docker image `4C:dat-converter` that is mentioned above?

amgebauer:
Here is the last 4C container that still contained the dat-converter: https://github.com/4C-multiphysics/4C/pkgs/container/4c/439828850?tag=dat-converter

ppraegla:
@mayrmt This is a convenient command that directly converts a .dat file in the current folder to yaml
```
docker run --user root --mount type=bind,source=$(pwd),target=/workspace --workdir /workspace ghcr.io/4c-multiphysics/4c:dat-converter /home/user/4C/build/4C --to-yaml my_input.dat
```
The command mounts the current directory into the docker container, runs `4C --to-yaml` and exits the Docker container.

PR #910, opened 2025-06-18
Convert old input file blocks in the analysisguide to yaml

ischeider:
## Description and Context
Since the old input file format is to be removed soon, the analysis guide of our documentation should reflect the new format.
This pull request replaces all occurrences of the old input file blocks to the new format. In addition. Jinja is used to include code blocks directly from input files, and sections from the meta file `4C_metadata.yaml`.

Note that the input files need to be mentioned in a CMakeLists.txt, so that they are copied to the build directory.
The alternative would be to copy all inputs to the build directory, which I find a bit too much.

Extracting sections from 4C_metadata.yaml is just a first step. We will probably extend the capabilities in the future, but the "quick-and-dirty" section grabbing from the metafile at least provides the section names, which can then be used in the documentation.

As of now, only tutorials are given as jinja files. Now we have a new directory `analysis_guide_templates`, where the jinja files for the analysis guide are located.
One might also think about putting all templates in one directory with subdirs. Ideas are welcome.

Part of #598

##Interested parties
@gilrrei


doc/documentation/CMakeLists.txt
```diff
@@ -22,6 +22,37 @@ if(FOUR_C_BUILD_DOCUMENTATION)
...
+      ${CMAKE_COMMAND}
+      -E
+      copy_if_different
+      ${FILE}
+      ${_sphinx_OUT_DIR}/reference_docs/parsed_files/${FILENAME}
+      )
+    message(
+      "COMMAND ${CMAKE_COMMAND} -E copy_if_different ${FILE} ${_sphinx_OUT_DIR}/reference_docs/parsed_files/${FILENAME}"
+      )
+  endforeach()
```
sebproell: Just use `configure_file` with the `COPY_ONLY` option for every file.

doc/documentation/CMakeLists.txt
```diff
@@ -34,13 +65,15 @@ if(FOUR_C_BUILD_DOCUMENTATION)
     documentation ALL
     COMMAND ${CMAKE_COMMAND} -E remove_directory "${PROJECT_BINARY_DIR}/doc/tmp"
     COMMAND ${CMAKE_COMMAND} -E remove_directory "${_sphinx_OUT_DIR}/reference_docs"
+    #    COMMAND ${CMAKE_COMMAND} -E remove_directory "${_sphinx_OUT_DIR}/parsed_files"
```
sebproell:
```suggestion
```

doc/documentation/CMakeLists.txt
```diff
@@ -34,13 +65,15 @@ if(FOUR_C_BUILD_DOCUMENTATION)
...
     COMMAND ${CMAKE_COMMAND} -E remove_directory "${_sphinx_HTML_DIR}"
     # the following command also creates the directory ${_sphinx_OUT_DIR}/reference_docs again
     COMMAND create_rtd
     COMMAND ${CMAKE_COMMAND} -E make_directory "${PROJECT_BINARY_DIR}/doc/tmp"
-    COMMAND ${CMAKE_COMMAND} -E make_directory "${_sphinx_OUT_DIR}/reference_docs/tutorials"
+    COMMAND ${CMAKE_COMMAND} -E make_directory "${_sphinx_OUT_DIR}/reference_docs/parsed_files"
```
sebproell: What does `parsed_files` mean in this context?

doc/documentation/src/analysis_guide/contact.rst
```diff
@@ -33,40 +42,47 @@ for self contact, see also :ref:`DESIGN MORTAR CONTACT CONDITIONS 2D<designlinem
...
+  .. code-block:: yaml
 
-     ---------------------------DESIGN SURF MORTAR CONTACT CONDITIONS 3D
-     DSURF 1
-     E 1 - InterfaceID 1 Side Selfcontact <further parameters>
+     DESIGN SURF MORTAR CONTACT CONDITIONS 3D:
+     - E: 1
+       InterfaceID: 1
+       Side: Selfcontact
+       <further parameters>
```
sebproell: Wouldn't it make more sense to load this from the metadata file? I believe we should almost never list single parameters in the docs as they get outdated all the time.

doc/documentation/src/analysis_guide/preprocessing.rst
```diff
@@ -14,7 +14,7 @@ input file.
 There are not so many means to create a valid input file. At this point, we know of the following
 different ways to create an input file. In general, you'll have two options:
 
-#. Either you create the input file in |FOURC|'s native format directly,
+#. Either you create the input file in |FOURC|'s native yaml format directly,
 #. or you create an input file in a general binary format for finite element information, called ``Exodus II``, develeloped by `Sandia National Laboratories
```
sebproell:
```suggestion
#. Either you create the mesh in |FOURC|'s native format directly,
#. or you create an mesh file in a general binary format for finite element information, called ``Exodus II``, develeloped by `Sandia National Laboratories
```

doc/documentation/src/analysis_guide/preprocessing.rst
```diff
@@ -51,20 +51,22 @@ Geometry as well as element and node sets can be created in any finite element p
 However, the preprocessor should be capable of exporting a file format, which can be converted
 by the python toolset meshio (see <https://pypi.org/project/meshio/>) into an exodus file.
 
+.. admonition:: Warning
+
+    The exodus file that is exported by meshio can currently not read in by |FOURC| directly!
```
sebproell:
```suggestion
    |FOURC| currently cannot read EXODUS files exported by meshio!
```

Please always spell the `exodus` file format as `EXODUS` (see #598).

tests/input_files/xfluid_beltrami_stat_16x16x16_hk_vol_equivalent.4C.yaml
sebproell, line 1: What is happening in these files? I cannot see the diff online. Are you using includes here?

sebproell (commented):
Nice! I'll leave it to you how much more you want to transform the docs here. I think  that long term we don't want manual descriptions of the input but always pull it from an exemplary file or metadata. Of course, this does not all have to happen in one PR.

doc/documentation/CMakeLists.txt
ischeider: parsed_files contains all files that are created from jinja input. Seems not to be intuitive :-/

tests/input_files/xfluid_beltrami_stat_16x16x16_hk_vol_equivalent.4C.yaml
ischeider: These are the input files from which sections are used in the documentation. However, you found the file, which is not yet used in thedocumentation (it will be used, when I created a section on the general structure of input files, which is already in my local repository ;-) ). I'll delete it for now.

ischeider: ... and indeed, I introduced INCLUDES here, which I actually wanted to postpone to a later PR. Shall I remove these changes here?

ischeider: I see that the INCLUDES directive does not pass the schema test. So I will turn it back for now. However, in general it should be possible to run test files with includes, no?

sebproell: Yes, for this we would need the partial json schema. I don't know how much sense the schema verification check makes. @amgebauer Opinions?

doc/documentation/src/analysis_guide/contact.rst
ischeider: I replaced most of the entries by examples from the input_files directory.
A problem exists, for example, in the file struct_dynamics.rst, which contains curly brackets within math equations, which are misinterpreted by Jinja.

doc/documentation/CMakeLists.txt
sebproell: `generated_files` or `templated_files`? This name isn't super important but the path seems to appear in the files for the docs, so it might help to get a better idea where these files come from.

ischeider: Actually, if all files in a directory are jinja-parsed, we could also have the original name there. Currently, for example, the directory tutorials is not used at all, which is not useful.
My proposal: change all file names in analysis_guide, so that they get the suffix j2 (the content does not necessarily change; we did that for tutorials), and then parse them into reference_docs/analysis_guide (same for tutorials).

sebproell: I wondered if it wouldn't be easiest if we unconditionally processed all our documentation files with Jinja first. Would this work? Do we actually need the extension `.rst.j2` (which throws off most linters unless configured separately)?

ischeider: In general it works. It's just my script that selects the .j2 files. The only problem is/was the comment I gave already below that some files have curly brackets in math equations, which jinja tries to convert (and fails of course). However, in the meantime I saw that one can parts of the rst file exclude from jinja parsing by {% raw %} ...  {% endraw %} directives. So, there seems to be no problem!

doc/documentation/python/include_yaml.py
```diff
@@ -10,17 +10,61 @@
...
+        for item in params_avail:
+            parameter_string = ""
+            if "name" not in item:
+                print(
+                    "section ",
+                    section["name"],
+                    " with ",
+                    item.keys(),
+                    " is too complex. Skipping.",
+                )
```
gilrrei: Could you elaborate what is going on here? Why is missing a name problematic?

ischeider: Some sections are nested to deep, the parameters follow after more "spec" hierarchies. cmake produces this warning for the following sections:
```
section  CONTACT CONSTITUTIVE LAWS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  DESIGN S2I KINETICS LINE CONDITIONS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  DESIGN S2I KINETICS POINT CONDITIONS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  DESIGN S2I KINETICS SURF CONDITIONS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  DESIGN SCATRA MULTI-SCALE COUPLING POINT CONDITIONS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  DESIGN SSI MANIFOLD KINETICS SURF CONDITIONS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  ELECTRODE BOUNDARY KINETICS LINE CONDITIONS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  ELECTRODE BOUNDARY KINETICS POINT CONDITIONS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  ELECTRODE BOUNDARY KINETICS SURF CONDITIONS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  FUNCT<n>  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
Parameters in section  INCLUDES  not found. Skipping.
section  MATERIALS  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
section  RESULT DESCRIPTION  with  dict_keys(['type', 'specs'])  is too complex. Skipping.
```
Definitely, when we produce the whole reference from the schema, we'll have the methods to identify all keywords no matter how deep the hierarchy is, but for now, I am fine with a brief scan of the main sections and parameters.

doc/documentation/CMakeLists.txt
ischeider: Now all tutorial files are in `tutorial_templates` dir on the source side, and the html files are can be found in `tutorials`. Accordingly with `analysis_guide_templates/analysis_guide`.

doc/documentation/src/analysis_guide/contact.rst
ischeider: I put thew `{% raw %} ... {% endraw %}` directives to the structu_dynamics.rst and could then also include jinja replacement to input files here.

Note that not all blocks have been replaced. Some give a more general explanation of the section, and the input files would give a quite lengthy output.

doc/documentation/python/include_yaml.py
ischeider: I replaced the output to something more useful. If we read the metafile in many files, we should skip the output completely.

doc/documentation/src/analysis_guide_templates/elementsale.rst
```diff
@@ -5,4 +5,4 @@ ALE elements are used with the following material models in the example input fi
 
 - MAT_ElastHyper
 - MAT_Struct_StVenantKirchhoff
-- MAT_matlist_reactions
+- MAT_matlist_reactions
```
gilrrei, line 8: Is this list created manually? If so, we should definitely move this to the metadata file. (In a followup)

doc/documentation/src/analysis_guide_templates/geometrydefinition.rst.j2
```diff
@@ -31,72 +37,59 @@ See :ref:`here <geometrysets>` how these sets are defined.
...
+meshed by 27 cube-shaped elements by
 
-   ----------STRUCTURE DOMAIN
-   LOWER_BOUND  0 0 0
-   UPPER_BOUND  10 5 3
-   INTERVALS   10 5 3
-   ROTATION 0 0 0
-   ELEMENTS    SOLID HEX8 MAT 1 KINEM nonlinear
-   PARTITION   structured
+{{section_dump(input_file1_content, "STRUCTURE DOMAIN") }}
```
gilrrei:
```suggestion
{{ section_dump(input_file1_content, "STRUCTURE DOMAIN") }}
```
Jinja2 recommends to have spaces between `{{` and the variable name, see [here](https://ttl255.com/jinja2-tutorial-part-3-whitespace-control/) for more information

gilrrei (commented):
I think this improves the documentation a lot. I think that there are further improvements possible which will build upon these changes!

doc/documentation/src/analysis_guide_templates/elementsale.rst
ischeider: Definitely. The elements sections in the analysis guide are awful. But as you said, please let's do it in a separate PR.

sebproell (approved):

gilrrei (approved):

PR #913, opened 2025-06-19
Rework SymbolicExpression for better evaluation performance

sebproell:
This PR adds benchmarks for a few SymbolicExpression examples and performs a couple of performance improvements:

- Remove micro allocations with unique_ptrs. Instead, allocate all memory for the AST in one arena.
- Remove all string lookups during evaluation. Use enum constants for known functions and access variables by index.

Here are the results:

```
-------------------------------------------------------------------------
Benchmark                                             main             PR
-------------------------------------------------------------------------
symbolic_expression_constant_mean                  4.13 ns        5.92 ns
symbolic_expression_basic_mean                      122 ns        65.7 ns
symbolic_expression_first_derivative_mean           383 ns         311 ns
symbolic_expression_second_derivative_mean          452 ns         392 ns
symbolic_expression_functions_mean                  313 ns        87.4 ns
```

Note: the `constant` case is a worse-case dummy where the function is a constant value. In this case, I now store the variables even if they are never used later. The more realistic `basic` case shows that this pays off.

Note 2: I had a quick look if there are nice libraries, but didn't find that many and none where I would know how we could use it with FAD types.


lauraengelhardt (approved):
Thanks for improving this, looks good to me!

amgebauer (approved):

PR #916, opened 2025-06-22
Compile SymbolicExpression to bytecode

sebproell:
Continuing from #913, this PR takes the AST representation of symbolic expressions and compiles it into bytecode instructions.

- Instead of walking the AST during evaluation, we now evaluate very simple bytecode instructions in a virtual machine. 
- This enables optimizations. For now, I implemented constant folding, i.e., all subexpressions involving only constants are evaluated up front (e.g `sqrt(3)`, `3*exp(pi/2)` etc...). I mainly went for this because it makes the instructions simpler (every instruction needs to be a unary or binary operation, no load instructions). In the future, one could implement common subexpression elimination, ...
- A significant part of the cost of evaluation is now related to the variable passing: since we can accept a dynamic number of variables, we need to match them against the allowed variables. For this reason, I now allow a simpler interface to evaluate a symbolic expression, where you just pass key-value pairs. The old interface taking a map of key-values is still possible but slower. The new interface is still rather slow in matching the variables -- at least, much slower than the bytecode execution. This could be further improved in the future.

Performance overview

```
-----------------------------------------------------------------
Benchmark                             main (after #913)        PR 
-----------------------------------------------------------------
symbolic_expression_constant                    6.42 ns   3.61 ns
symbolic_expression_constant_folding            36.8 ns   3.66 ns
symbolic_expression_basic                       61.4 ns   28.2 ns
symbolic_expression_basic_native               0.712 ns  0.715 ns
symbolic_expression_first_derivative             315 ns    137 ns
symbolic_expression_second_derivative            368 ns    248 ns
symbolic_expression_functions                   83.9 ns   59.8 ns
```

--> everything is faster

The diff seems large because I reorganized code in the third commit. The major changes (bytecode compiler + execution) are in the second commit.


mayrmt:
@sebproell Out of curiosity: what's the runtime impact on actual simulations?

sebproell:
> @sebproell Out of curiosity: what's the runtime impact on actual simulations?

@mayrmt Hard to say. I haven't made any specific measurements. In an explicit time stepping scenario with a source term implemented as a SymbolicExpression, the improvement can probably be quite drastic. I have a student working on this. In a typical unoptimized 4C code path you might not see any improvement at all, since this (previously expensive) code is dwarfed by all the other slow code.

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
```diff
@@ -0,0 +1,1046 @@
...
+  template <typename T>
+  struct IsFAD : public std::false_type
+  {
+  };
+
+  template <typename T>
+  struct IsFAD<Sacado::Fad::DFad<T>> : public std::true_type
+  {
+    using value_type = T;
+  };
```
amgebauer: Here you might just use https://github.com/4C-multiphysics/4C/blob/main/src/core/utils/src/numerics/4C_utils_fad_meta.hpp.

If you know that it is a Sacado Fad type, then you get the value type with `typename T::value_type`

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
```diff
@@ -0,0 +1,1046 @@
...
+          lexer.lexan();
+          break;
+        }
+        else
+        {
+          if (name == "acos" or name == "asin" or name == "atan" or name == "cos" or
+              name == "sin" or name == "tan" or name == "cosh" or name == "sinh" or
+              name == "tanh" or name == "exp" or name == "log" or name == "log10" or
+              name == "sqrt" or name == "ceil" or name == "heaviside" or name == "fabs" or
+              name == "floor")
```
amgebauer: instead of listing them here again, you can create this list automatically by creating a constexpr map with a consteval function from the enum definition above.

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
```diff
@@ -0,0 +1,1046 @@
...
+            auto function_name = *EnumTools::enum_cast<UnaryFunctionType>(name);
+            lhs = create_node(
+                NodeType::unary_function, {.unary_function = function_name}, parse_expr(lexer));
+
+            // Consume closing parenthesis
+            if (lexer.tok_ != Lexer::tok_rpar) FOUR_C_THROW("')' expected");
+            lexer.lexan();
+            break;
+          }
+          else if (name == "atan2")
```
amgebauer: I think it also makes sense to define the other functions add, mul, etc additional to the shortcuts +,  *, ... (at least that doesn't hurt). Then you can generate this list automatically (see above)

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
```diff
@@ -0,0 +1,1046 @@
...
+            break;
+          case BinaryFunctionType::sub:
+            dst = lhs - rhs;
+            break;
+          case BinaryFunctionType::mul:
+            dst = lhs * rhs;
+            break;
+          case BinaryFunctionType::div:
+            dst = lhs / rhs;
+            break;
```
amgebauer, line 971: why these special cases? They are also part of `binary_function_table`

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
```diff
@@ -0,0 +1,1046 @@
...
+      }
+
+      if (index < vm_memory_.size())
+      {
+        if constexpr (IsFAD<T>::value)
+        {
+          if constexpr (IsFAD<typename IsFAD<T>::value_type>::value)
+            return std::to_string(vm_memory_[index].val().val());
+          else
+            return std::to_string(vm_memory_[index].val());
```
amgebauer: We could generalize this for arbitrary nesting and move it to https://github.com/4C-multiphysics/4C/blob/main/src/core/utils/src/numerics/4C_utils_fad_meta.hpp.

amgebauer (commented):
Nice implementation. I have a few suggestions.

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
sebproell: Good idea! It's not really the content of this PR (I know the diff is wild because I moved code). I will see if I can add it here or in my next PR.

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
sebproell: Hm, do you mean to allow `add(2,mul(3,x))` as an expression instead of `2 + 3*x`? I would not do that. Again, I don't really want to change the parser in this PR.

amgebauer: yes, I would allow that since it simplifies a lot of the code. I did not took a look at the deleted code, so I didn't know what already existed before.

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
sebproell: Good question: this is to get better assembly code, since these ops are basically instructions and I don't want to jump to a function for them. Makes about 10-20% difference in the benchmarks .

lkoeglmeier (dismissed):

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
sebproell: As discussed offline, I would keep it here, as it is not clear how useful this function would be in general. Btw, this is only debug printing, I could fully remove this (and might in a few PRs).

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
```diff
@@ -0,0 +1,1035 @@
...
+        if (name == "pi")
+        {
+          lhs = create_node(NodeType::number, {.number = M_PI});
+          lexer.lexan();
+          break;
+        }
+        else
+        {
+          auto maybe_unary_function = EnumTools::enum_cast<UnaryFunctionType>(name);
+          if (maybe_unary_function)
```
sebproell, line 518: @amgebauer Thanks to `magic_enum` this was trivial to change. Nice idea!

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
```diff
@@ -0,0 +1,1035 @@
...
+                parse_expr(lexer));
+
+            // Consume closing parenthesis
+            if (lexer.tok_ != Lexer::tok_rpar) FOUR_C_THROW("')' expected");
+            lexer.lexan();
+            break;
+          }
+
+          auto maybe_binary_function = EnumTools::enum_cast<BinaryFunctionType>(name);
+          if (maybe_binary_function)
```
sebproell, line 536: We also accept `add`, `mul` etc. but I won't promote it ;)

amgebauer (approved):

PR #919, opened 2025-06-23
SymbolicExpression with compile-time variables

sebproell:
Allow SymbolicExpression to take its variables at compile-time to allow early checks for correct arguments and much simpler copying of passed variables into internal memory.

```
------------------------------------------------------------------------------
Benchmark                                   fast AST   bytecode  comp-time vars
------------------------------------------------------------------------------
symbolic_expression_constant                 6.42 ns    3.61 ns         2.06 ns   
symbolic_expression_constant_folding         36.8 ns    3.66 ns         2.06 ns
symbolic_expression_basic*                   61.4 ns    28.2 ns          103 ns
symbolic_expression_basic_compile_time_vars        -          -         7.64 ns
symbolic_expression_basic_native            0.712 ns   0.715 ns        0.772 ns 
symbolic_expression_first_derivative          315 ns     137 ns          114 ns
symbolic_expression_second_derivative         368 ns     248 ns          238 ns 
symbolic_expression_functions                83.9 ns    59.8 ns         37.8 ns
```
*: Note that the `basic` benchmark was changed to use the slow map path. The benchmark shows that you should really use the compile-time variables ;)


src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
```diff
@@ -321,24 +262,107 @@ namespace Core::Utils::SymbolicExpressionDetails
...
+        for (const auto& var : variable_names_)
+        {
+          if (!vars.contains(var))
+          {
+            missing_variables += var + " ";
+          }
+        }
+
+        FOUR_C_THROW(
+            "Missing variables {}to evaluate expression '{}'", missing_variables, expression_);
```
amgebauer, line 325:
```suggestion
            "Missing variables {} to evaluate expression '{}'", missing_variables, expression_);
```

amgebauer (approved):

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
sebproell: Yep, that was intentional ;) I know it looks a bit wonky. C++23 will make formatted printing so much nicer.

lauraengelhardt (approved):

PR #920, opened 2025-06-24
IsoNeoHooke: make mu an InputField

sebproell:
Enhance Neo-Hooke material to take `mu` as an InputField . I removed some unused code along the way.

FYI @gilrrei

### DIFF
--- a/src/mat/4C_mat_elasthyper.cpp
+++ b/src/mat/4C_mat_elasthyper.cpp
@@ -189,7 +189,7 @@ int Mat::ElastHyper::mat_id(const unsigned index) const
 
 /*----------------------------------------------------------------------*/
 /*----------------------------------------------------------------------*/
-double Mat::ElastHyper::shear_mod() const
+double Mat::ElastHyper::shear_mod(int ele_gid) const
 {
   // principal coefficients
   bool haveshearmod = false;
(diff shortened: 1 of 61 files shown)


src/mat/4C_mat_elasthyper.cpp
```diff
@@ -208,20 +208,6 @@ double Mat::ElastHyper::shear_mod() const
   return shearmod;
 }
 
-/*----------------------------------------------------------------------*/
-/*----------------------------------------------------------------------*/
-double Mat::ElastHyper::get_young()
```
maxfirmbach: I actually need that function in a PR that is almost done 😬. Would you mind keeping it?

sebproell: Sure, I can keep it :+1:

maxfirmbach: Thanks!

maxiludwig (approved):

ischeider:
You don't add a single example with non-constant MUE, so why do you introduce this parameter as an input_field?

gilrrei:
@ischeider 
> You don't add a single example with non-constant MUE, so why do you introduce this parameter as an input_field?

This was a feature request from QUEENS, as we use this to test and modernize our random field 4C approach for material parameters. You are right, we could have mentioned this.

PR #922, opened 2025-06-24
Clean up group functionality in InputSpec

sebproell:
- I introduced `group_struct` in #904, but meanwhile, the interface has evolved in #884. The first commit removes the `group_struct` function and moves this functionality into the `group` function.
- Whether a group is defaultable or not, is fully deducible. Users can now only specify if the group should be required (the default) or optional. (Idea by @gilrrei)

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -784,28 +784,8 @@ namespace Core::IO
       StoreFunction<DefaultStorage> transform_data{nullptr};
     };
 
-    //! Additional parameters for a group().
-    struct GroupData
-    {
-      /**
-       * An optional description of the Group.
-       */
-      std::string description{};
-
-      /**
-       * Whether the Group is required or optional.
-       */
-      std::optional<bool> required{};
-
-      /**
-       * Whether the Group will store itself and its children with defaulted values, if the Group
-       * is not encountered in the input. This only works if all children have default values.
-       */
-      bool defaultable{};
-    };
-
     template <typename T>
-    struct GroupStructData
+    struct GroupData
     {
       /**
        * An optional description of the Group.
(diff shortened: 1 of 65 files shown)


src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -2190,14 +2185,14 @@ v: [1,2,3]
                                           });
       };
       FOUR_C_EXPECT_THROW_WITH_MESSAGE(
-          construct(), Core::Exception, "Groups can only store to InputParameterContainer.");
+          construct(), Core::Exception, "contains specs that store to");
```
lauraengelhardt, line 2188: Am I missing something, or is this message incomplete?

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -1330,15 +1293,40 @@ namespace Core::IO
...
      *
-     * @note Providing inconsistent types for the store functions and group_struct() will cause
+     * @note Providing inconsistent types for the store functions and group() will cause
      * a runtime error.
+     *
+     * Whether a group can have a default value is determined by the default values of its children.
+     * If all of its children have default values, the group can have a default value. In this case,
+     * you may set the `defaultable` option to guarantee that the default values of the children are
+     * stored under the group name in the container, even if the group is not present in the input.
+     * Obviously, this only makes sense if the group is not required.
```
lauraengelhardt: Just a suggestion to shorten and clarify the need to be non-required.
```
A group can be defaultable if it is not required and all of its children have default values.
In this case, you may set the `defaultable` option to guarantee that the default values of the children are stored under the group name in the container, even if the group is not present in the input.
```

src/core/io/tests/4C_io_input_spec_test.cpp
sebproell: I just test for a fragment of the error message. I cannot include more info, since the type name that is printed is not guaranteed to be portable across compilers.

src/core/io/src/4C_io_input_spec_builders.hpp
sebproell: Yes, good catch! I need to remove `defaultable` here.

gilrrei (approved):

PR #923, opened 2025-06-24
Fix assertion in quaternion interpolation

amgebauer:
Follows #876 

@abhi-satheesh Can you double check? It looks like the `assert` did not fire in the PR checks (but caused a problem locally).

@sebproell Do we define `NDEBUG` in the assertion mode in the PR checks?

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
+++ b/src/core/linalg/src/dense/4C_linalg_utlis_quaternion_interpolation.cpp
@@ -525,7 +525,8 @@ Core::LinAlg::GeneralizedSphericalLinearInterpolator<loc_dim>::gram_schmidt_orth
   gs_basis(2, 3) = -gs_basis(0, 2) * d24 + gs_basis(1, 2) * d14 - gs_basis(3, 2) * d12;
   gs_basis(3, 3) = gs_basis(0, 2) * d23 - gs_basis(1, 2) * d13 + gs_basis(2, 2) * d12;
 
-  assert(gs_basis.determinant() < 1.0 - 1e-8 && gs_basis.determinant() > 1.0 + 1e-8);
+  FOUR_C_ASSERT(
+      std::abs(gs_basis.determinant() - 1.0) < 1e-8, "Determinant of the basis is not 1.");
 
   return gs_basis;
 }


amgebauer:
@abhi-satheesh Note that I changed the assertion from `det != 1` to `det == 1`

sebproell:
@amgebauer CMake defines `NDEBUG` for build type `RELEASE`. Since all our CI tests are based on `RELEASE`, we always defined that.

amgebauer:
Can we undefine it for the assertion job?

sebproell:
Yes, by basing our assertion job on `DEBUG` and increasing the optimization level there.

abhi-satheesh:
@amgebauer 
I meant the following assertion:
`      dsassert(gs_basis.Determinant() < 1.0 - TOLCOMP && gs_basis.Determinant() > 1.0 + TOLCOMP,
          "ERROR gram_schmidt_orth |R| != 1.0");`

amgebauer:
@abhi-satheesh but it should be the other way around, right?

abhi-satheesh:
@amgebauer you are correct:

I switched the sings (+ and -) in the previous response.
it should be the other way round:
`        assert ( 1-TOL < basis.Determinant()
                         &&  basis.determinant < 1 + TOL );`

I wonder why this was not caught in the tests? I guess, all the tests should fail !!!

amgebauer:
> I wonder why this was not caught in the tests? I guess, all the tests should fail !!!

The `assert()` is (currently) not active on our Github checks. The correct way to assert things is with `FOUR_C_ASSERT()` instead of `assert()`.

Can someone approve?

abhi-satheesh:
> > I wonder why this was not caught in the tests? I guess, all the tests should fail !!!
> 
> The `assert()` is (currently) not active on our Github checks. The correct way to assert things is with `FOUR_C_ASSERT()` instead of `assert()`.
> 
> Can someone approve?

I guess it also not active for local runs as well !

amgebauer:
> > > I wonder why this was not caught in the tests? I guess, all the tests should fail !!!
> > 
> > 
> > The `assert()` is (currently) not active on our Github checks. The correct way to assert things is with `FOUR_C_ASSERT()` instead of `assert()`.
> > Can someone approve?
> 
> I guess it also not active for local runs as well !

It is only active in a debug build. I'm using the debug build as default, that's why I stumbled over it 😬

PR #924, opened 2025-06-24
Clang-tidy: enable check for reserved identifiers

sebproell:
I tried to enable this check (https://clang.llvm.org/extra/clang-tidy/checks/bugprone/reserved-identifier.html) in deal.II once, and it was super expensive. Let's see how fast it would run here. I can at least fix whatever it finds.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -18,7 +18,6 @@ Checks: >-
   performance-*,
   bugprone-*,
   boost-*,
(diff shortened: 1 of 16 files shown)


sebproell:
This only takes two more minutes, so let's enable it.

mayrmt (approved):

mayrmt (approved):

PR #925, opened 2025-06-24
CMake: build shared libraries by default

sebproell:
Based on #831, we now build shared libraries by default.


sebproell:
To be clear, this changes the build to shared libs at IMCS. OK @mayrmt ?

mayrmt:
@sebproell, I know. Let's do this!

PR #928, opened 2025-06-25
ReducedLung: Refactor terminal units 

bgoderbauer:
## Description and Context

This is the first step to make the data structures in `ReducedLung` more flexible and easier to handle. The new layout is applied to the terminal units in this PR. The other "elements" in `ReducedLung` will follow. The new layout allows arbitrary model combinations for the viscoelastic-type algebraic equations that are used in terminal units. 

Breakdown:
- Terminal units are saved in model-specific structs with arrays indexed by the different corresponding elements. During assembly, all given terminal unit models are looped one by one.
- Additional models: 
     - Ogden nonlinear elasticity for springs.
     - Four-element Maxwell rheological model.
- Implementation tests of the model functions and their derivatives using finite differences.

Until the new `ReducedLung` input is written, all terminal units remain defaulted to Kelvin-Voigt with linear elasticity.

@maxiludwig

### DIFF
--- a/src/reduced_lung/4C_reduced_lung_main.cpp
+++ b/src/reduced_lung/4C_reduced_lung_main.cpp
@@ -17,9 +17,9 @@
 #include "4C_io_discretization_visualization_writer_mesh.hpp"
 #include "4C_linalg_utils_sparse_algebra_manipulation.hpp"
 #include "4C_linear_solver_method_linalg.hpp"
-#include "4C_mat_maxwell_0d_acinus_NeoHookean.hpp"
 #include "4C_red_airways_elementbase.hpp"
 #include "4C_reduced_lung_helpers.hpp"
+#include "4C_reduced_lung_terminal_unit.hpp"
 #include "4C_utils_function_of_time.hpp"
 
 #include <Teuchos_StandardParameterEntryValidators.hpp>
@@ -61,7 +61,7 @@ namespace ReducedLung
     // Create vectors of local entities (equations acting on the dofs).
     // Physical "elements" of the lung tree introducing the dofs.
     std::vector<Airway> airways;
-    std::vector<TerminalUnit> terminal_units;
+    TerminalUnits terminal_units;
     std::map<int, int> dof_per_ele;  // Map global element id -> dof.
     int n_airways = 0;
     int n_terminal_units = 0;
@@ -81,21 +81,16 @@ namespace ReducedLung
       }
       else if (ele->element_type() == Discret::Elements::RedAcinusType::instance())
       {
-        double E;
-        double eta;
         if (ele->material(0)->material_type() == Core::Materials::m_0d_maxwell_acinus_neohookean)
         {
-          auto* param = ele->material(0)->parameter();
-          auto* mat = static_cast<Mat::PAR::Maxwell0dAcinus*>(param);
-          E = mat->stiffness1_;
-          eta = mat->viscosity1_;
+          // Default terminal unit model until the new input is available.
+          add_terminal_unit_ele<KelvinVoigt, LinearElasticity>(
+              terminal_units, ele, local_element_id);
         }
         else
         {
           FOUR_C_THROW("Material not implemented.");
         }
-        terminal_units.push_back(TerminalUnit{element_id, local_element_id, n_terminal_units,
-            TerminalUnitType::kelvin_voigt, E, eta});
         dof_per_ele[element_id] = 3;
         n_terminal_units++;
       }
@@ -137,13 +132,14 @@ namespace ReducedLung
         airway.global_dof_ids.insert(airway.global_dof_ids.end(), first_dof_gid + i);
       }
     }
-    for (auto& terminal_unit : terminal_units)
+    for (auto& model : terminal_units.models)
     {
-      int first_dof_gid = first_global_dof_of_ele[terminal_unit.global_element_id];
-      int n_dof = dof_per_ele[terminal_unit.global_element_id];
-      for (int i = 0; i < n_dof; i++)
+      for (size_t i = 0; i < model.data.number_of_elements(); i++)
       {
-        terminal_unit.global_dof_ids.insert(terminal_unit.global_dof_ids.end(), first_dof_gid + i);
+        int first_dof_gid = first_global_dof_of_ele[model.data.global_element_id[i]];
+        model.data.gid_p1.push_back(first_dof_gid);
+        model.data.gid_p2.push_back(first_dof_gid + 1);
+        model.data.gid_q.push_back(first_dof_gid + 2);
       }
     }
 
@@ -166,8 +162,9 @@ namespace ReducedLung
       }
       return result;
     };
-    auto global_ele_ids_per_node = Core::Communication::all_reduce<std::map<int, std::vector<int>>>(
-        ele_ids_per_node, merge_maps, comm);
+    FourC::Core::Communication::IsCommunicatable auto global_ele_ids_per_node =
+        Core::Communication::all_reduce<std::map<int, std::vector<int>>>(
+            ele_ids_per_node, merge_maps, comm);
 
     // Create entities with equations connecting elements (acting on "nodes" of the lung tree).
     std::vector<BoundaryCondition> boundary_conditions;
@@ -176,16 +173,15 @@ namespace ReducedLung
     int n_boundary_conditions = 0;
     int n_connections = 0;
     int n_bifurcations = 0;
-    // Loop over all local elements, get their nodes, create associated entity. This way, they are
-    // created on the same ranks as at least one of their connected elements. This reduces the
-    // amount of communication of dofs between ranks.
-
+    // Find all nodes with boundary conditions
     std::vector<const Core::Conditions::Condition*> conditions;
     actdis->get_condition("RedAirwayPrescribedCond", conditions);
     const auto red_airway_prescribed_conditions =
         Core::Conditions::find_conditioned_node_ids_and_conditions(
             *actdis, conditions, Core::Conditions::LookFor::locally_owned_and_ghosted);
-
+    // Loop over all local elements, get their nodes, create associated entity. This way, they are
+    // created on the same ranks as at least one of their connected elements. This reduces the
+    // amount of communication of dofs between ranks.
     for (const auto* ele : actdis->my_row_element_range())
     {
       const auto* nodes = ele->nodes();
@@ -347,21 +343,25 @@ namespace ReducedLung
       }
     }
 
-    // Calculate local and global number of "element" equations
+    // Calculate local and global number of "element" equations and assign local row IDs to define
+    // the structure of the system of equations.
     int n_local_equations = 0;
     for (const auto& airway : airways)
     {
       n_local_equations += airway.n_state_equations;
     }
-    for (const auto& terminal_unit : terminal_units)
+    for (auto& tu_model : terminal_units.models)
     {
-      n_local_equations += terminal_unit.n_state_equations;
+      for (size_t i = 0; i < tu_model.data.number_of_elements(); i++)
+      {
+        tu_model.data.local_row_id.push_back(n_local_equations);
+        n_local_equations++;
+      }
     }
-
     // Assign local equation ids to connections, bifurcations, and boundary conditions.
     for (Connection& conn : connections)
     {
-      // Every connection adds 1 momentum  and 1 mass balance equation.
+      // Every connection adds 1 momentum and 1 mass balance equation.
       conn.first_local_equation_id = n_local_equations;
       n_local_equations += 2;
     }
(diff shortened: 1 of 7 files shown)


src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
```diff
@@ -0,0 +1,632 @@
...
+    NegativeResidualEvaluator negative_residual_evaluator;
+    JacobianEvaluator jacobian_evaluator;
+    InternalStateUpdater internal_state_updater;
+  };
+
+  /**
+   * @brief All terminal unit models together and interface for access.
+   *
+   * Stores all different terminal unit models as distinct building blocks. Acts as interface for
+   * distributing terminal units to the correct models in the input phase amd allows access to all
```
sebproell:
```suggestion
   * distributing terminal units to the correct models in the input phase and allows access to all
```

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
```diff
@@ -0,0 +1,632 @@
...
+      const std::vector<double>& elastic_pressure_grad_dp_el, double dt);
+
+  /**
+   * Creates a function for evaluating the elastic pressure given a specific elasticity model. This
+   * function is given to the NegativeResidualEvaluator.
+   */
+  struct MakeElasticPressureEvaluator
+  {
+    std::function<std::vector<double>&(
+        TerminalUnitData&, const Core::LinAlg::Vector<double>&, double)>
```
sebproell: Maybe a `using` statement for that?

src/reduced_lung/4C_reduced_lung_terminal_unit.cpp
```diff
@@ -0,0 +1,227 @@
...
+    return ogden_hyperelastic_model.elastic_pressure_grad_dp_el;
+  }
+
+  void evaluate_kelvin_voigt_jacobian(Core::LinAlg::SparseMatrix& target,
+      const KelvinVoigt& kelvin_voigt_model, TerminalUnitData& data,
+      const std::vector<double>& elastic_pressure_grad_dp_el)
+  {
+    [[maybe_unused]] int err;
+    if (!target.filled())
+    {
```
sebproell, line 123: I believe it wouldn't be too hard to set up your system matrix with the correct sparsity pattern (in our case, `LinAlg::Graph`) up front. Then you could save this distinction, which interferes with the nice physics implementation.

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
```diff
@@ -0,0 +1,632 @@
...
+    operator()(LinearElasticity& linear_elasticity_model)
+    {
+      return [&linear_elasticity_model](TerminalUnitData& data,
+                 const Core::LinAlg::Vector<double>& dofs, double dt) -> std::vector<double>&
+      { return evaluate_linear_elastic_pressure(linear_elasticity_model, data, dofs, dt); };
```
sebproell:
```suggestion
      return std::bind_front(evaluate_linear_elastic_pressure, linear_elasticity_model);
```

FYI. The lambda is also fine.

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
```diff
@@ -0,0 +1,632 @@
...
+   */
+  void evaluate_four_element_maxwell_jacobian(Core::LinAlg::SparseMatrix& target,
+      const FourElementMaxwell& four_element_maxwell_model, TerminalUnitData& data,
+      const std::vector<double>& elastic_pressure_grad_dp_el, double dt);
+
+  /**
+   * Creates a function for evaluating the elastic pressure given a specific elasticity model. This
+   * function is given to the NegativeResidualEvaluator.
+   */
+  struct MakeElasticPressureEvaluator
```
sebproell: Maybe put this struct into `MakeNegativeResidualEvaluator`. This makes it clear that this is just an internal helper.

Also, I believe @amgebauer has some nice tricks to visit multiple variants. Maybe they can help here.

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
```diff
@@ -0,0 +1,632 @@
...
+   *
+   * @tparam R Rheological model type.
+   * @tparam E Elasticity model type.
+   * @param terminal_units Global container for all terminal unit models.
+   * @return Reference to the matching TerminalUnitModel.
+   *
+   * @note Ensures reuse of model instances for shared material parameter sets.
+   */
+  template <typename R, typename E>
+  TerminalUnitModel& return_or_create_and_return_reference_to_terminal_unit_model(
```
sebproell: I think you may be exposing too much of the implementation with this name. Why not just `register_model`, `access_model` etc.?

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
```diff
@@ -0,0 +1,632 @@
...
+    model.data.local_element_id.push_back(local_element_id);
+    const auto& coords_node_1 = ele->nodes()[0]->x();
+    const auto& coords_node_2 = ele->nodes()[1]->x();
+    const double radius =
+        std::sqrt((coords_node_1[0] - coords_node_2[0]) * (coords_node_1[0] - coords_node_2[0]) +
+                  (coords_node_1[1] - coords_node_2[1]) * (coords_node_1[1] - coords_node_2[1]) +
+                  (coords_node_1[2] - coords_node_2[2]) * (coords_node_1[2] - coords_node_2[2]));
+    const double volume = radius * radius * M_PI;
+    model.data.volume_v.push_back(volume);
+    model.data.reference_volume_v0.push_back(volume);
```
sebproell: I'll put the comment here but it applies to a few places: structs are super nice if you use them with aggregate initialization (the `{}`) and use designated initializers. This basically gives you a constructor with named arguments _and_ you get compiler warnings if you forget to initialize something. And all of that without writing a single line of code :) So I would suggest to use this whenever reasonable. It really helps to construct fully initialized objects. Here, I would define a struct for a single data point, add all the data to it, and create a helper function that puts the single data point into the vectors. This nicely exhibits the AoS to SoA transformation.

src/reduced_lung/tests/4C_reduced_lung_terminal_unit_test.cpp
```diff
@@ -0,0 +1,188 @@
...
+    Vector<double> locally_relevant_dofs(col_map, true);
+    dofs.replace_local_values(9, std::array<double, 9>{1, 1, 1, 1, 1, 1, 1, 1, 1}.data(),
+        std::array<int, 9>{0, 1, 2, 3, 4, 5, 6, 7, 8}.data());
+    export_to(dofs, locally_relevant_dofs);
+
+    double dt = 1e-1;         // Dummy time step size
+    const double eps = 1e-6;  // Perturbation parameter for the FD approximation
+
+    // Model parameters: Kelvin Voigt + Linear Elasticity
+    {
```
sebproell:
```suggestion
    {
    SCOPED_TRACE("Kelvin Voigt + Linear Elasticity")
```

And in the other subtests as well. The unit testing purist might also recommend a fixture class for the setup, so I'll mention it, but I personally would also write it like that, as I don't like to jump around and figure out which setup code belongs to which test.

src/reduced_lung/4C_reduced_lung_main.cpp
```diff
@@ -166,8 +162,9 @@ namespace ReducedLung
       }
       return result;
     };
-    auto global_ele_ids_per_node = Core::Communication::all_reduce<std::map<int, std::vector<int>>>(
-        ele_ids_per_node, merge_maps, comm);
+    FourC::Core::Communication::IsCommunicatable auto global_ele_ids_per_node =
```
sebproell:
```suggestion
    auto global_ele_ids_per_node =
```

I didn't know this was possible! However, I don't think it helps a lot to annotate the type with the concept here.

sebproell (commented):
Nice. Overall, I really like how easy this code is to read.

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
```diff
@@ -0,0 +1,632 @@
...
+    std::vector<double> elasticity_E;
+    // Internal elastic pressure state
+    std::vector<double> elastic_pressure_p_el;
+    std::vector<double> elastic_pressure_grad_dp_el;
+  };
+
+  /**
+   * @brief Ogden-type hyperelastic model. Substitutes a spring in a rheological model.
+   *
+   * Nonlinear stiffening behavior based on a bulk modulus and shape parameter.
```
maxiludwig:
```suggestion
   * Nonlinear stiffening behavior based on a bulk modulus kappa and shape parameter beta.
```

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
```diff
@@ -0,0 +1,632 @@
...
+
+    ElasticityModel& elasticity_model;
+  };
+
+
+  /**
+   * Creates a function for evaluating the elastic pressure gradient w.r.t. inflow q given a
+   * specific elasticity model. This function is given to the JacobianEvaluator.
+   */
+  struct MakeElasticPressureGradientEvaluator
```
maxiludwig: Can also be an internal helper of the `MakeJacobianEvaluator` as suggested in https://github.com/4C-multiphysics/4C/pull/928/files#r2167098873

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
amgebauer: Actually, I think my tricks don't work here.

But couldn't you give `evaluate_linear_elastic_pressure` and `evaluate_ogden_hyperelastic_pressure` the same name, let's say `evaluate_hyperelastic_pressure` and create an overload for each material. Then, you can visit it simply with

```c++
std::visit([&](auto& material) {
    return std::bind_front(evaluate_hyperelastic_pressure, material);
}, variant);
```

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
```diff
@@ -0,0 +1,632 @@
...
+
+    model.data.global_element_id.push_back(ele->id());
+    model.data.local_element_id.push_back(local_element_id);
+    const auto& coords_node_1 = ele->nodes()[0]->x();
+    const auto& coords_node_2 = ele->nodes()[1]->x();
+    const double radius =
+        std::sqrt((coords_node_1[0] - coords_node_2[0]) * (coords_node_1[0] - coords_node_2[0]) +
+                  (coords_node_1[1] - coords_node_2[1]) * (coords_node_1[1] - coords_node_2[1]) +
+                  (coords_node_1[2] - coords_node_2[2]) * (coords_node_1[2] - coords_node_2[2]));
+    const double volume = radius * radius * M_PI;
```
maxiludwig:
```suggestion
    const double volume = (4.0 / 3.0) * M_PI * radius * radius * radius;
```

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
bgoderbauer: This sounds like a nice solution for the material exchange, @amgebauer, thanks! 

One question regarding this: The functions `evaluate_linear_elastic_pressure` and `evaluate_ogden_hyperelastic_pressure` have different variables (linear elasticity doesn't need the strain state, the nonlinear model needs it). Is it okay to give the missing variables to `evaluate_linear_elastic_pressure`, even though they are never used, to be able to use overload and `std::bind_front`?

src/reduced_lung/4C_reduced_lung_main.cpp
bgoderbauer: It is shown as a warning if I don't add it. I will remove it, but it annoys me a little to see a warning in my IDE, which seems unnecessary :smile:

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
amgebauer: if they fit well into the scope, then yes. Passing strains into an evaluation function of the constitutive law is fine. But why does linear elasticity not need the strains?

bgoderbauer: Sorry, wrong function. The gradient of the linear elastic pressure doesn't need the strains.
I'll apply it in this nice and short way :+1: .

bgoderbauer:
Thanks for the initial review! I've added the suggestions so far and discussed a few more points for the next steps with @sebproell:

- Filling the sparsity pattern before starting the assembly is definitely nicer, I will postpone it to after we've restructured the remaining elements (airways, junctions). Then I can just write a plain function that is executed at the beginning.
- The initialization process can be improved with aggregate initialization; however, this is probably easier after changing to the new input.

src/reduced_lung/4C_reduced_lung_terminal_unit.hpp
bgoderbauer: Well, it wasn't so straightforward to apply :disappointed:. I returned to the first suggestion and kept the structs for the elasticity model evaluators inside the residual and `Jacobian` evaluators.

PR #934, opened 2025-06-26
Use tensors in solid material evaluation

amgebauer:
This PR changes the solid material interface from `Matrix<>` to `SymmetricTensor<>`.

I adapted some simple materials along the way, for the more complex one, I just converted to the old voigt-notation (what we previously did in the elements). This enables to write all equations in a bette readable way.

Since I changed the solid-material interface, I also made the `ParameterList` const so that we cannot be misued anymore. A few materials used this to tell the structural time integration about evaluation failure (often if the defgrd is negative). Since we check for this on the element level, we never run into this case. Other materials used it to write some postprocessing stresses into the parameterlist that were never read elsewhere.

fyi @lauraengelhardt @ischeider @dragos-ana @c-p-schmidt @BishrMaradni


amgebauer:
Now this PR is ready for review.

I had to additionally adapt the tolerance of an ehl3d test that has already been identified as brittle in #662.

amgebauer:
Thanks for your reviews! Now since the interface is using tensors, we can work on simplifying the material models one by one independently.

PR #939, opened 2025-06-26
Input: remove redefinition of node counts

sebproell:
Previously, every element defined geometric primitives itself; now we use the primitve name to infer this. 

@gilrrei @davidrudlstorfer this will break FourCIPP (I think). The elements no longer repeat the nodal ID vector since it is obvious from the element type. What would be a useful way to export the different geometries? Or do you code this in FourCIPP?

This is only the next step in the transformation of the element input. In the next steps, the strings will become enums, and I will make the number of nodes more flexible (see https://github.com/4C-multiphysics/4C/pull/891#issuecomment-3008578195).

Follows #891

### DIFF
--- a/apps/global_full/4C_global_full_main.cpp
+++ b/apps/global_full/4C_global_full_main.cpp
@@ -335,8 +335,20 @@ int main(int argc, char* argv[])
   {
     if (Core::Communication::my_mpi_rank(lcomm) == 0)
     {
+      ryml::Tree tree = Core::IO::init_yaml_tree_with_exceptions();
+      ryml::NodeRef root = tree.rootref();
+      root |= ryml::MAP;
+      Core::IO::YamlNodeRef root_ref(root, "");
+
+      // Write the non-user input metadata that is defined globally for 4C.
+      Global::emit_general_metadata(root_ref);
+
+      // Write the user input defined for various physics module.
       Core::IO::InputFile input_file = Global::set_up_input_file(lcomm);
-      input_file.emit_metadata(std::cout);
+      input_file.emit_metadata(root_ref);
+
+      // Finally, dump everything.
+      std::cout << tree;
     }
   }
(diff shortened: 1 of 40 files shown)


gilrrei:
@sebproell thanks for the heads up. This will definitely break fourcipp. How am I supposed to know how many node ids will follow? For example, how should I know that HEX8 requires 8 nodes? Is the naming consistent?

sebproell:
@gilrrei yes the naming is consistent. It wouldn't be a problem to emit the number of nodes into metadata if that helps.

gilrrei:
@sebproell 
> @gilrrei yes the naming is consistent. It wouldn't be a problem to emit the number of nodes into metadata if that helps.

Yes, that would be helpful. If it were possible, directly emitting the node/connectivity ordering compared to the vtk standard (if the cell type exists in vtk), this information would make the life of lnmmeshio, cubitpy, meshpy, and fourcipp easier, as currently these have to be hard-coded and maintained manually.

sebproell:
I restructured the metadata output:

- We now have info about the `cell_types:
  ```yaml
  cell_types:
    QUAD4:
      number_of_nodes: 4
    QUAD6:
      number_of_nodes: 6
    QUAD8:
      number_of_nodes: 8
    TRI3:
      number_of_nodes: 3
    ...
    ```
    This is only a start and will soon be extended with additional info such as vtu cell types. @isteinbrecher we can also enhance this to have nodes per direction, once we go for the `LINE`, `QUAD` etc. cells without fixed numbers.

- The `legacy_element_specs` now report which of the `cell_types` is supported and which additional specs are allowed.

davidrudlstorfer (approved):
This probably breaks FourCIPP, unfortunately I am a bit clueless regarding the necessary changes in FourCIPP, @gilrrei could you soonish create a PR for FourCIPP so the Workflows are not breaking unnecessarily long

gilrrei (approved):

Issue #244, opened 2025-01-13
Remove DbcHandling option from contact conditions


lauraengelhardt:
## Description
The input parameter `DbcHandling` takes the options `DoNothing` and `RemoveDBCSlaveNodes`. The option `RemoveDBCSlaveNodes` is untested. 

There is also a comment by @mayrmt saying: `// ToDo (mayr.mt) Remove? Do not change DBCs at runtime!`

So let's remove the parameter, and code related to the option `RemoveDBCSlaveNodes`.

## Possible Solution and Definition of Done
- [ ] Remove `DbcHandling` input parameter
- [ ] Remove code related to `Inpar::Mortar::DBCHandling::remove_dbc_nodes_from_slave_side`

## Interested Parties
@mayrmt

mayrmt:
@lauraengelhardt I still support to remove this.

maxfirmbach:
@mayrmt @lauraengelhardt Can this be closed?

mayrmt:
Yes, let's close this. Thanks for the reminder!

PR #941, opened 2025-06-27
Remove DbcHandling option from contact conditions

lauraengelhardt:
## Description and Context
Remove `DbcHandling` option from contact conditions as untested and DBCs shouldn't be modified at runtime.

see #244


lauraengelhardt:
> @lauraengelhardt I agree that one should not modify Dirichlet boundary conditions at run time.
> 
> Is there still a check in place, that ensures that the intersection between the slave interface and the Dirichlet boundary is the empty set?

Good point! 

I just ran some checks using a simple example of two cubes adapted from a test case. See the attached screenshot.
[
Screenshot 2025-06-27 at 16 51 11
](url)

Surface 1 and 3 as well as 2 and 3 have overlapping sets of nodes (the bordering lines).
The opposite side of surface 4 is fixed in space. Surface 1 has a DBC pulling in x-direction; fixed in y- and z-direction.
Surface 3 and 4 are the contact pairs (3 being the slave/secondary, 4 the master/primary). 
I used the new time integration where build_interfaces is called. 

1. Running this works on the current main and this branch without any errors. So we do not check if there are nodes that belong to both DBC and CC.
2. The result is equal for the current main and this branch.
3. From the results I can see that the DBC is the "dominant" condition.
4. I printed the result of `detect_dbc_slave_nodes_and_elements` - `dbc_slave_nodes`, `dbc_slave_eles` is always empty. Took me a while to track it down but the slave condition is only added to the vector `sl_conds` for the case `RemoveDBCSlaveNodes`. In consequence `detect_dbc_slave_nodes_and_elements` does only detects dbc slave nodes when `RemoveDBCSlaveNodes` was activated. These lines do nothing unless `RemoveDBCSlaveNodes` is activated:
```
  // skip dbc slave nodes ( if the corresponding option is set for
  // the slave condition )
  if (dbc_slave_nodes.find(node) != dbc_slave_nodes.end()) continue;
```

Long story short: As far as I could see, I think we do not have such a check in place even in the current main. All the functions I removed do nothing unless `RemoveDBCSlaveNodes` is chosen - which is not tested. The name `detect_dbc_slave_nodes_and_elements` is somewhat misleading.

I think we can proceed here, but need to figure out how we want to treat cases where we have an intersection between slave interface and the DBC.

mayrmt:
@lauraengelhardt I have opened #947 to keep track of the missing check. I think this PR can proceed.

lauraengelhardt:
Perfect, thanks for opening the issue @mayrmt!

davidrudlstorfer (approved):

PR #945, opened 2025-06-27
Move input files to physics

davidrudlstorfer:
## Description and Context

Next PR which is part of #306 

Keeping the changes low to maintain ability for reviews.

#306


amgebauer (dismissed):

src/structure_new/src/utils/4C_structure_new_monitor_dbc_input.cpp
```diff
@@ -0,0 +1,75 @@
...
+            // precision for file
+            parameter<int>("PRECISION_FILE",
+                {.description = "precision for written file", .default_value = 16}),
+
+            // precision for screen
+            parameter<int>("PRECISION_SCREEN",
+                {.description = "precision for written screen output", .default_value = 5}),
+
+            // type of written output file
+            deprecated_selection<IOMonitorStructureDBC::FileType>("FILE_TYPE",
```
c-p-schmidt: Would it be so much effort to directly replace this with 

              parameter<Inpar::IOMonitorStructureDBC::FileType>(
                  "FILE_TYPE", {.description = "type of written output file",
                                   .default_value = Inpar::IOMonitorStructureDBC::csv}),

c-p-schmidt (approved):

src/structure_new/src/utils/4C_structure_new_monitor_dbc_input.cpp
davidrudlstorfer: Thanks for the review:)

I'll wait until the workflows are finished so I can see if everything works without any implementation changes. If everything works, I'll add your suggestion:)

src/beam3/4C_beam3_discretization_runtime_output_input.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include "4C_io_input_spec_builders.hpp"
+
+
+FOUR_C_NAMESPACE_OPEN
+
+
+namespace IORuntimeOutput
+{
+  namespace Beam
```
sebproell: These namespaces should be flipped.

src/structure_new/src/utils/4C_structure_new_monitor_dbc_input.cpp
```diff
@@ -0,0 +1,75 @@
...
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#include "4C_structure_new_monitor_dbc_input.hpp"
+
+#include "4C_io_input_spec_builders.hpp"
+
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace IOMonitorStructureDBC
```
sebproell: This namespace is also not allowed. The top-level namespace should be whatever we use for the files in `structure_new`.

src/structure_new/src/utils/4C_structure_new_monitor_dbc_input.cpp
```diff
@@ -0,0 +1,75 @@
...
+    switch (type)
+    {
+      case FileType::csv:
+        return "csv";
+      case FileType::data:
+        return "data";
+      default:
+        FOUR_C_THROW("Unknown file type");
+    }
+  }
```
sebproell: If you want, you can kick this function and use the equivalent functionality from `EnumTools`. Just in case you decide to go for @c-p-schmidt's suggestion above and touch this code part again.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
```diff
@@ -158,8 +160,8 @@ void Solid::MonitorDbc::setup()
       Global::Problem::instance()->output_control_file()->file_name_only_prefix());
   Core::IO::create_directory(full_dirpath, Core::Communication::my_mpi_rank(get_comm()));
   // ... create files paths ...
-  full_filepaths_ =
-      create_file_paths(rconds, full_dirpath, filename_only_prefix, to_string(filetype));
+  full_filepaths_ = create_file_paths(
+      rconds, full_dirpath, filename_only_prefix, std::string(EnumTools::enum_name(filetype)));
```
davidrudlstorfer, line 162: @sebproell one question: why do I need the std::string here as it is already done in other parts of the code? I would have expected to not need it

davidrudlstorfer (commented):
@sebproell @c-p-schmidt thanks for your review, I've incorporated all comments:)

this would be ready for review again

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
sebproell: Apparently this function cannot implicitly take  string_view, which is what is returned from enum_name.

sebproell (approved):

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
davidrudlstorfer: Good to know - thanks for the explanation:)

PR #946, opened 2025-06-27
Doc: recommendations for work on input files

sebproell:
Mention schema files and general format of input files.

Part of #598

PR #948, opened 2025-06-28
Convert a few tests to JSON format

sebproell:
Also adds formatting for these test files.

In reference to #946


davidrudlstorfer (approved):
Would it make sense to activate the .json formatter for all `*.json` files? Do we currently have any unformatted `.json` files in the repo?

PR #952, opened 2025-06-29
InputSpec: write validators to metadata

sebproell:
Write knowledge about valid ranges or sets to the metadata file.

FYI @gilrrei

### DIFF
--- a/utilities/create_json_schema.py
+++ b/utilities/create_json_schema.py
@@ -7,28 +7,29 @@
 
 """Create JSON schema for YAML validation."""
 
-import json
 import argparse
+import json
+
 from pathlib import Path
 
 from metadata_utils import (
-    NotSet,
-    NOTSET,
-    NAMED_TYPES,
-    MISSING_DESCRIPTION,
     FOURC_BASE_TYPES_TO_JSON_SCHEMA_DICT,
-    Primitive,
-    Vector,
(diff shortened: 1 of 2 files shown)


src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -797,6 +799,10 @@ required: true
               - name: C
             required: false
             default: A
+            validator:
+              set:
+                - A
+                - C
```
sebproell: @gilrrei Not sure how we should write the info for enums. The enum already lists all values that are legal for the type in `choices`. The `validator` entry might further constrain that by only allowing some of the entries. It looks weird here; the technical reason is that I don't want to create validators if all enum values are allowed. Let me know what would be desirable for our tools here.

gilrrei: Doesn't this mean that the choices are irrelevant for the schema and only the ones in the validator are sensible choices?

gilrrei (dismissed):

sebproell:
@gilrrei Looks like we need to fix the schema here before we can merge. Would you mind pushing something for the validators?

lauraengelhardt (approved):

amgebauer (approved):

PR #955, opened 2025-06-30
Enable and cleanup vtk based output in poro pressurebased framework

lkoeglmeier:
## Description and Context

The first commit updates the porofluid_pressurebased framwork output, as we are already using the new vtk-based output in the artery framework (see #467).
The second commit adds the runtime csv-writer for domain integrals.

part of #211

### DIFF
--- a/src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm.cpp
+++ b/src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm.cpp
@@ -271,6 +271,18 @@ void PoroPressureBased::PorofluidAlgorithm::init(bool isale, int nds_disp, int n
   // the values of the integrals
   domain_integrals_ = std::make_shared<Core::LinAlg::SerialDenseVector>(num_domainint_funct_);
 
+  if (num_domainint_funct_ > 0)
+  {
+    // setup csv writer for domain integrals
+    runtime_csvwriter_domain_integrals_.emplace(
+        myrank_, *Global::Problem::instance()->output_control_file(), "domain_integrals");
+    for (int funct_i = 0; funct_i < num_domainint_funct_; funct_i++)
+    {
+      runtime_csvwriter_domain_integrals_->register_data_vector(
+          "Function ID " + std::to_string(domainint_funct_[funct_i]), 1, 14);
+    }
+  }
+
   set_element_general_parameters();
 
   // build mesh tying strategy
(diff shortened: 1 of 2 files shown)


src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm.cpp
```diff
@@ -271,6 +271,17 @@ void PoroPressureBased::PorofluidAlgorithm::init(bool isale, int nds_disp, int n
   // the values of the integrals
   domain_integrals_ = std::make_shared<Core::LinAlg::SerialDenseVector>(num_domainint_funct_);
 
+  // setup csv writer for domain integrals
+  runtime_csvwriter_domain_integrals_.emplace(
```
c-p-schmidt: Did you maybe forget to initialize this csvwriter, only if the num_domainint_funct_ is larger than 0?

Because then:
- The std::optional would make sense, currently, you always initialize a csvwriter
- If I did not miss anything, you only add data to the file in those cases, so only setting up the writer, and with that, the file in those cases looks reasonable to me.

src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm.cpp
```diff
@@ -1553,42 +1563,17 @@ void PoroPressureBased::PorofluidAlgorithm::evaluate_domain_integrals()
...
 
-    if (step_ == 0)
-    {
-      // inform user that function output has been requested
-      std::cout << "\nINFO for domain integrals:\nUser requested " << num_domainint_funct_
-                << " function(s) which will be integrated over the entire domain" << std::endl;
-      std::cout << "The result will be written into " << filename << "\n" << std::endl;
-
-      // open file and write header
-      file.open(filename, std::fstream::trunc);
```
c-p-schmidt, line 1571: This section looks way nicer now!

c-p-schmidt (commented):
Thank you for your cleanup!

src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm.cpp
lkoeglmeier: yes, thanks! I added it.

lkoeglmeier:
Formatting should be back in place now, sorry!

bwirthl (approved):

PR #959, opened 2025-07-01
Fix: Update test file check for yaml/json files

davidrudlstorfer:
## Description and Context

Following the .4C.yaml/4C.json switch we also need to update the pre-commit check to test for orphaned test files.

I removed the check for empty sections. I think it would make more sense to just validate the input files? But then we would somehow need a metadata file to compare against.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
         name: Format C/C++
         # Only use on C and C++ files and not on other files that clang could format (e.g. json).
         types_or: [c++, c]
-        exclude: 'mm$'
+        exclude: "mm$"
 
   - repo: https://github.com/cheshirekow/cmake-format-precommit
     rev: v0.6.13
@@ -33,7 +33,7 @@ repos:
     hooks:
       - id: xml-formatter
         name: Format xml
-        args: [--indent, '2', --indent-char, ' ', --selfclose, --overwrite]
+        args: [--indent, "2", --indent-char, " ", --selfclose, --overwrite]
         # Only run on "pure" XML files and not on files that internally use XML (e.g. svg)
         files: \.xml$
 
@@ -55,7 +55,7 @@ repos:
       - id: trailing-whitespace
         name: Trim trailing whitespace
         # Do not modify reference vtu files, since we generate these and do not hand-edit them.
-        exclude: 'vtu$'
+        exclude: "vtu$"
         # Do not trim trailing whitespace in C++ files. It is not trivial to trim whitespace
         # correctly, e.g., when inside a string literal. Clang-format will do this correctly.
         exclude_types: [c, c++]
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -116,50 +116,51 @@ repos:
...
         entry: utilities/python-venv/bin/check-preprocessor
 
       - id: input-files
         name: Test input files
-        description: 'Check input files conform to expectations'
+        description: "Check input files conform to expectations"
         language: python
-        files: 'tests/input_files/.*\.dat$'
+        files: 'tests/input_files/.*\.4C\.(yaml|json)$'
+        exclude: '.*\.mat\.4C\.yaml$'
```
amgebauer: Probably I missed the discussion. What are `.mat.4C.yaml`?

utilities/four_c_utils/src/four_c_utils/check_test_files.py
```diff
@@ -88,16 +63,16 @@ def main():
...
     # check input file tests
-    errors += check_inputtests(args.filenames, allerrors)
+    errors = check_inputtests(args.filenames)
 
-    utils.pretty_print_error_report("", allerrors, errfile)
-    return errors
+    utils.pretty_print_error_report("", errors, errfile)
+
+    if errors:
+        return 1
```
amgebauer:
```suggestion
        raise RuntimeError("Input files contain errors (see above)")
```
or, better, put the error message from `pretty_print_error_report` in the exception.

amgebauer (commented):
> I think it would make more sense to just validate the input files?

We validate the input files, but we can only do after the buildtest to have the schema-file.

.pre-commit-config.yaml
davidrudlstorfer: No clue :D

 in the elch tests they are used as "microfiles", the switch to yaml was done in https://github.com/4C-multiphysics/4C/pull/906
 
 @sebproell knows more regarding this

sebproell: @c-p-schmidt and @georghammerl use these files as additional input files.

c-p-schmidt: In the past, we distinguished between .mat and .dat, where the .mat file e.g., defined the parameters on the microscale and is not a complete input file but has large parts of an input file.
So the current naming convention that was introduced from changing from dat to yaml is .mat -> .mat.4C.yaml .dat -> .4C.yaml

amgebauer (approved):

.pre-commit-config.yaml
c-p-schmidt: Currently, there are no such JSON files in the repo, but since we can read them, we could already add such a check?
```suggestion
        exclude: '.*\.mat\.4C\.(yaml|json)$'
```

davidrudlstorfer: Good point! Added.

c-p-schmidt (approved):

PR #961, opened 2025-07-01
Let 4C B free*

amgebauer:
Don't use the B-operator for evaluation of nonlinear displacement based solid mechanics. The underlying equations get much more visible than before.

Draft for now

`*` I stole this nice title from [Planas et al (2011), Comput. Methods Appl. Mech. Eng.](https://www.sciencedirect.com/science/article/abs/pii/S0045782512000308?via%3Dihub), though I'm not using their ansatz.

### DIFF
--- a/tests/input_files/poro_scatra_contact3D_two_cubes.4C.yaml
+++ b/tests/input_files/poro_scatra_contact3D_two_cubes.4C.yaml
@@ -403,7 +403,7 @@ RESULT DESCRIPTION:
       DIS: "structure"
       NODE: 23
       QUANTITY: "dispz"
-      VALUE: 1.096025534030374e-09
+      VALUE: 1.096025e-09
       TOLERANCE: 1e-15
   - FLUID:
       DIS: "porofluid"
--- a/tests/input_files/ssti_mono_3D_tet4_batt_with_anode_plate_and_current_collectors_elch_s2i_butlervolmer_BGS_AMG_11x11.4C.yaml
+++ b/tests/input_files/ssti_mono_3D_tet4_batt_with_anode_plate_and_current_collectors_elch_s2i_butlervolmer_BGS_AMG_11x11.4C.yaml
@@ -374,7 +374,7 @@ RESULT DESCRIPTION:
       DIS: "scatra"
       NODE: 4
       QUANTITY: "phi2"
-      VALUE: 6.85349056391118278e-03
+      VALUE: 0.006853490381277293
       TOLERANCE: 6.9e-11
   - STRUCTURE:
       DIS: "structure"


amgebauer:
Here is the benchmark comparison:



Evaluation/assembly of element stiffness matrix (1 GP):

| Element 	| B-op 	| B-free 	| speedup 	| |
|---	|---	|---	|---	|-|
| hex8 	| 756 ns 	| 552 ns 	| 1.4 	| :white_check_mark: |
| hex20 	| 4055 ns 	| 1803 ns 	| 2.2 	| :white_check_mark: |
| hex27 	| 7013 ns 	| 3676 ns 	| 1.9 	| :white_check_mark: |
| tet4 	| 235 ns 	| 348 ns 	| 0.7 	| :x: |
| tet10 	| 1143 ns 	| 696 ns 	| 1.64 	| :white_check_mark: |

Evaluation of element (default Gauss rule):

| Element 	| B-op 	| B-free 	| speedup 	| |
|---	|---	|---	|---	|-|
| hex8 	| 8 795 ns 	| 7 708 ns 	| 1.1 	| :white_check_mark: |
| hex20 	| 123 466 ns 	| 64 900 ns 	| 1.9 	| :white_check_mark: |
| hex27 	| 213 839 ns 	| 119 997 ns 	| 1.8 	| :white_check_mark: |
| tet4 	| 835 ns 	| 1 125 ns 	| 0.7 	| :x: |
| tet10 	| 6 467 ns 	| 5 312 ns 	| 1.2 	| :white_check_mark: |

amgebauer:
This PR is ready for review

src/solid_3D_ele/4C_solid_3D_ele_calc_lib.hpp
```diff
@@ -834,6 +841,47 @@ namespace Discret::Elements
     return stress;
   }
 
+  template <Core::FE::CellType celltype>
+  static inline void add_nodal_contribution(std::size_t element_node_id,
```
sebproell, line 854: Maybe a small doc since this is a rather central function that you use often?

src/solid_3D_ele/4C_solid_3D_ele_calc_lib.hpp
```diff
@@ -856,6 +904,50 @@ namespace Discret::Elements
         integration_fac, Bop, Core::LinAlg::make_stress_like_voigt_view(stress.pk2_), 1.);
   }
 
+  template <Core::FE::CellType celltype>
+  void add_internal_force_vector(const JacobianMapping<celltype>& jacobian_mapping,
+      const Core::LinAlg::Tensor<double, Core::FE::dim<celltype>, Core::FE::dim<celltype>>& F,
+      const Core::LinAlg::SymmetricTensor<double, Core::FE::dim<celltype>, Core::FE::dim<celltype>>&
+          pk2,
```
sebproell, line 940: You could consider using stronger types here. It seems very easy to mess up by giving the wrong tensors. E.g. with

```
struct DefGrad
{
const Core::LinAlg::Tensor<double, Core::FE::dim<celltype>, Core::FE::dim<celltype>>& f;
};
```
you would change this to

```suggestion
      const DefGrad& F,
      const Core::LinAlg::SymmetricTensor<double, Core::FE::dim<celltype>, Core::FE::dim<celltype>>&
          pk2,
```

and at the call site use `add_internal_force_vector(..., DefGrad{F}, ...)`.

 I am not sure how general this function is. If it is only an internal helper, this may not be as useful.

src/solid_3D_ele/4C_solid_3D_ele_calc_lib.hpp
```diff
@@ -885,41 +977,27 @@ namespace Discret::Elements
...
-    for (int inod = 0; inod < Internal::num_nodes<celltype>; ++inod)
+    for (std::size_t i = 0; i < Core::FE::num_nodes(celltype); ++i)
     {
-      SmB_L[0] = stress.pk2_.data()[0] * N_XYZ[inod](0) + stress.pk2_.data()[3] * N_XYZ[inod](1) +
-                 stress.pk2_.data()[5] * N_XYZ[inod](2);
-      SmB_L[1] = stress.pk2_.data()[3] * N_XYZ[inod](0) + stress.pk2_.data()[1] * N_XYZ[inod](1) +
-                 stress.pk2_.data()[4] * N_XYZ[inod](2);
-      SmB_L[2] = stress.pk2_.data()[5] * N_XYZ[inod](0) + stress.pk2_.data()[4] * N_XYZ[inod](1) +
-                 stress.pk2_.data()[2] * N_XYZ[inod](2);
+      const auto dNidX_PK2w = jacobian_mapping.N_XYZ[i] * stress.pk2_ * integration_fac;
```
sebproell, line 1024: Wow! It is actually readable now 🤩

unittests/solid_3D_ele/4C_solid_3D_ele_calc_lib_test.cpp
```diff
@@ -234,4 +239,106 @@ namespace
 
     FOUR_C_EXPECT_NEAR(x_centroid, x_centroid_ref, 1e-15);
   }
+
+  namespace
+  {
+    template <Core::FE::CellType celltype>
+    Discret::Elements::ElementNodes<celltype> get_element_nodes()
+    {
+      std::srand(0);
```
sebproell, line 248: Why not write out the arbitrary numbers instead of hiding them behind random? Would make it easier to debug or test elsewhere.

src/solid_3D_ele/4C_solid_3D_ele_calc_lib.hpp
```diff
@@ -856,6 +904,50 @@ namespace Discret::Elements
...
+        Core::LinAlg::einsum<"bi", "aijd", "cj">(F, stress.cmat_, F * integration_fac);
+
+    for (std::size_t i = 0; i < Core::FE::num_nodes(celltype); ++i)
+    {
+      const auto dNidX_FCFTw = jacobian_mapping.N_XYZ[i] * FCFTw;
+      const auto dNidX_PK2w = jacobian_mapping.N_XYZ[i] * stress.pk2_ * integration_fac;
+
+      for (std::size_t j = 0; j < Core::FE::num_nodes(celltype); ++j)
+      {
+        add_nodal_contribution<celltype>(
```
lauraengelhardt, line 971:
```suggestion
        // material stiffness
        add_nodal_contribution<celltype>(
```

lauraengelhardt (commented):
Only one minor comment. I also think this is much more readable now! :)

unittests/solid_3D_ele/4C_solid_3D_ele_calc_lib_test.cpp
amgebauer: I left it for now like this. Since I test it for different celltypes, this is easier scalable instead of writing out the random numbers (Depending on the celltype, I need more or less of these numbers).

src/solid_3D_ele/4C_solid_3D_ele_calc_lib.hpp
amgebauer: I would postpone it to another PR and then do it generally for the whole file to make it consistent.

lauraengelhardt (approved):

sebproell (approved):

PR #963, opened 2025-07-02
Remove all raw epetra import calls

dharinib98:
## Description and Context

With this PR all the physics modules are rid of `get_epetra_import` and are contained only within `linalg`. Additionally our own implementation for obtaining source and target map is included.

#755

### DIFF
--- a/src/core/fem/src/discretization/4C_fem_discretization.cpp
+++ b/src/core/fem/src/discretization/4C_fem_discretization.cpp
@@ -540,8 +540,8 @@ void Core::FE::Discretization::set_state(
     }
     // (re)build importer if necessary
     if (stateimporter_[nds] == nullptr or
-        not stateimporter_[nds]->sourcemap().same_as(state.get_map().get_epetra_block_map()) or
-        not stateimporter_[nds]->targetmap().same_as(colmap->get_epetra_block_map()))
+        not stateimporter_[nds]->source_map().same_as(state.get_map().get_epetra_block_map()) or
+        not stateimporter_[nds]->target_map().same_as(colmap->get_epetra_block_map()))
     {
       stateimporter_[nds] = std::make_shared<Core::LinAlg::Import>(*colmap, state.get_map());
     }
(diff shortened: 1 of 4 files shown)


src/core/fem/src/discretization/4C_fem_discretization.cpp
```diff
@@ -540,10 +540,8 @@ void Core::FE::Discretization::set_state(
     }
     // (re)build importer if necessary
     if (stateimporter_[nds] == nullptr or
-        not stateimporter_[nds]->get_epetra_import().SourceMap().SameAs(
-            state.get_map().get_epetra_block_map()) or
-        not stateimporter_[nds]->get_epetra_import().TargetMap().SameAs(
-            colmap->get_epetra_block_map()))
+        not stateimporter_[nds]->sourcemap().same_as(state.get_map().get_epetra_block_map()) or
+        not stateimporter_[nds]->targetmap().same_as(colmap->get_epetra_block_map()))
```
maxfirmbach: I think it would make sense to call it `source_map()` and `target_map()`. Just to be consistent.

src/core/linalg/src/sparse/4C_linalg_transfer.cpp
```diff
@@ -18,7 +18,9 @@ FOUR_C_NAMESPACE_OPEN
 //! Standard constructor using source and target maps
 Core::LinAlg::Import::Import::Import(const Map& targetMap, const Map& sourceMap)
```
maxfirmbach: Same here, let's do snake style as `target_map` and `source_map`.

src/core/linalg/src/sparse/4C_linalg_transfer.hpp
```diff
@@ -25,10 +25,14 @@ namespace Core::LinAlg
     Import(const Map& targetMap, const Map& sourceMap);
 
     const Epetra_Import& get_epetra_import() const;
+    const Map& sourcemap() const;
+    const Map& targetmap() const;
 
 
    private:
     Utils::OwnerOrView<Epetra_Import> import_;
+    Map sourceMap_;
```
maxfirmbach: Snake style.

PR #966, opened 2025-07-03
Remove dat from InputFile

sebproell:
- Removes `--to-yaml` options
- Removes the `dat` as an option in `InputFile`
- Removes associated tests

There is still a little more dat-related or dat-inspired code which we currently use to parse the legacy strings.

Part of #598


davidrudlstorfer (approved):

mayrmt:
https://github.com/4C-multiphysics/4C/issues/598#issuecomment-2979753532 indicates, that one could leave this option around for some time. Is that time already over? Is there urgent need to progress with maximum speed?

sebproell:
@mayrmt Nobody mentioned any interest in that and nobody suggested any time. This comment was written two weeks ago and was meant in the spirit of the original state of #907.  I don't like to keep code around that we don't use and even less so without a good reason or time frame. So I suggest to just remove this code and better reflect the current state of 4C (where dat does no longer work).

maxfirmbach:
I would also vote for removing all the unused code as soon as possible, especially if it can't be used anymore.

PR #967, opened 2025-07-03
Fix visualization of distributed moments for the SBIP approach (potential interaction between beams)

davidrudlstorfer:
## Description and Context

Until now it was unclear if the SBIP approach has a more fundamental flaw or if just the visualization of the resulting residual force vector is flawed. As it turned out the problem lies within the visualization and the underlying implementation is correct.

This PR now fixes the visualization of the distributed moments for both (simple and full) implemented SBIP approaches implemented by Max Grill. 

Noteworthy, the visualization now is:

- discretization independent. Distributed moments do no longer scale with the discretization size, i.e., element size.
- correct regarding the magnitude of the moment. Previously, this was off by a constant factor even if the element size was 1
- correctly oriented. Previously the moment pointed into the wrong direction.

I also added exhaustive explanations on how to derive this distributed moment (as this is not at all straight forward to compute).

Additionally, I added comments to the two implemented SSIP approaches (small and large separations) which explain that no moment is present in these approaches.

Big thanks to Max Grill and Christoph Meier for long and in-depth discussions regarding this topic.


src/beaminteraction/src/potential/4C_beaminteraction_potential_pair_beam_to_beam.cpp
```diff
@@ -3449,21 +3466,59 @@ bool BeamInteraction::BeamToBeamPotentialPair<numnodes, numnodalvalues,
...
 
-  moment_pot_tmp.multiply(1.0 / Core::FADUtils::cast_to_double(norm_r_xi_master),
-      spin_pseudo_moment_tmp, Core::FADUtils::cast_to_double<T, 3, 1>(t_master));
+  m.cross_product(Core::FADUtils::cast_to_double<T, 3, 1>(r_xi_slave), m_pseudo);
+
+  vtk_moment_pot_slave_GP.update(prefactor_visualization_data, m, 1.0);
+
+
+  // distributed moment on master
+  m_pseudo.update_t(Core::FADUtils::cast_to_double(pot_red_fac_deriv_xi_master) *
```
c-p-schmidt, line 3512: At some locations, you first perform the multiplication and then the `cast_to_double`, whereas here, you do it the other way around. You could think about unifying it.

davidrudlstorfer:
@c-p-schmidt & @isteinbrecher thanks a lot for your review of the changes

@c-p-schmidt good catch regarding the inconsistent use of the `cast_to_double` conversions, I changed it accordingly to be consistent

@isteinbrecher good catch - there actually was a test for this exact property but the test failed, this led to #993 and #994. Now I updated the quantities in the reference files.

Regarding the overall state of this PR: there still is some inconsistency within the approach, namely the force vector still scales with the discretization (but I am still not sure if this is really the case). This needs further and deeper investigations regarding our entire approach. To prevent that the changes of this PR get further outdated I would opt to merge this PR in its current form.

isteinbrecher (approved):

PR #929, opened 2025-06-25
Update porofluid header to better use new input format

bwirthl:
This PR updates the entire pressure-based porofluid header input to make better use of the capabilities of the new input format. Unfortunately, the input of the different subtypes is so entangled that I had to do everything together, which makes this PR so huge. Plus, since it uses structure and scatra input, I had to add several "translators" to translate my new input to the old naming. 

Here you find a list of all possible input parameters for easier overview (since the diff of the PR is huge)
```yaml
PROBLEM TYPE:
  PROBLEMTYPE: porofluid_pressure_based
porofluid_dynamic:
  total_simulation_time: 1000.0
  time_integration:
    number_of_time_steps: 10
    time_step: 0.1
    theta: 0.5
    scheme: one_step_theta
  nonlinear_solver:
    maximum_number_of_iterations: 10
    absolute_tolerance_residual: 1.0E-14
    linear_solver_id: 1
    residual:
      vector_norm: l2
      tolerance: 1.0E-6
    increment:
      vector_norm: l2
      tolerance: 1.0E-6
    convergence_criteria_adaptivity:
      active: true
      nonlinear_to_linear_tolerance_ratio: 0.1
  initial_condition:
    type: by_function
    function_id: 1
  output:
    restart_data_every: 1
    results_every: 1
    saturation_and_pressure: true
    solid_pressure: true
    porosity: true
    phase_velocities: true
  fd_check:
    active: true
    epsilon: 1.0E-6
    tolerance: 1.0E-6
  calculate_error_to_analytical_solution:
    active: true
    function_id: 2
  skip_initial_time_derivative: false
  biot_stabilization:
    active: true
    scaling_factor: 1
  divergence_action: continue_anyway
  flux_reconstruction:
    active: true
    solver_id: 2
  domain_integrals_function_ids: "3 4"
  starting_DBC:
    active: "0 1"
    time_end: 2
    function_ids: "0 5"
  artery_coupling_active: true
porofluid_dynamic/artery_coupling:
  coupling_method: mortar_penalty
  penalty_parameter: 1000
  coupled_dofs:
    artery: "1"
    homogenized: "3"
  reaction_terms:
    artery_function_ids: "1 0"
    artery_scaling: "1 0"
    homogenized_function_ids: "0 0 4"
    homogenized_scaling: "0 0 1"
  maximum_number_of_segments_per_artery_element: 5
  evaluate_in_reference_configuration: false
  lateral_surface_coupling: true
  integration_patches:
[...]

PR #971, opened 2025-07-07
Add check for illegal enum names in input

davidrudlstorfer:
- adds check for illegal enum names
- fixes "inf" in poro input

## Description and Context

Adds a check for illegal enum names in the input.

Big thanks to @sebproell for the pair programming and help :)

@bwirthl is it ok for you to rename the norm from "inf" to "max"?

In the near future a workflow should be added which tests the latest fourcipp version with the latest metadata file in each PR

Part of #420 
Closes https://github.com/4C-multiphysics/fourcipp/issues/77


src/porofluid_pressure_based/4C_porofluid_pressure_based_input.hpp
```diff
@@ -37,7 +37,7 @@ namespace PoroPressureBased
     l1_scaled,  //!< L1/linear norm scaled by length of vector
     l2,         //!< L2/Euclidean norm
     rms,        //!< root mean square (RMS) norm
-    inf         //!< Maximum/infinity norm
+    max         //!< Maximum/infinity norm
```
mayrmt: Mathematicians call such a norm an inf-norm, so I prefer to keep "inf" to be consistent with established language.

src/core/io/src/4C_io_input_types.hpp
```diff
@@ -28,10 +30,40 @@ namespace Core::IO
...
 
+    constexpr std::array illegal_enum_names = {
+        // "yes", // TODO https://github.com/4C-multiphysics/4C/issues/420
+        // "no", // TODO https://github.com/4C-multiphysics/4C/issues/420
+        "on",
+        "off",
+        "null",
+        "inf",
+        "infinity",
+    };
```
mayrmt: Why should these be illegal? Where does this list come from?

src/porofluid_pressure_based/4C_porofluid_pressure_based_input.hpp
bwirthl: As far as I can tell (but I am no mathematician) it is called maximum norm or infinity norm. I would be fine with `max` or `maximum`, but then this should be consistent in 4C. In `LinAlg`, it is called inf-norm which calls `NormInf` from Trilinos.

src/core/io/src/4C_io_input_types.hpp
bwirthl: I wonder the same thing. @davidrudlstorfer or @sebproell Can you explain why these should be illegal? Because they are reserved by c++?

sebproell: I think the correct list of identifiers we should rather not use as enums (only in the input of course) is given here: https://yaml.org/spec/1.2.2/#1032-tag-resolution

@davidrudlstorfer As mentioned during lunch: I think we may be fixing things in the wrong place, if we disallow `inf`. The spec linked above says that special meaning is assigned to `.inf`. Isn't something wrong with the roundtripping of yaml->json->python in FourCIPP?

davidrudlstorfer:
As it turned out this was a little bit more complex than initially involved. We should exclude the yaml reserved keywords (see https://yaml.org/spec/1.2.2/#1032-tag-resolution) which is done in this PR.

The remaining keywords need more attention => see #420 

The initial `inf` problem is solved with https://github.com/4C-multiphysics/fourcipp/pull/78

src/porofluid_pressure_based/4C_porofluid_pressure_based_input.hpp
davidrudlstorfer: This change was reverted and the problem is now solved differently in fourcipp

src/core/io/src/4C_io_input_types.hpp
davidrudlstorfer: As  @sebproell mentioned these illegal enum names emerge from the yaml standard

sebproell (approved):

src/core/io/src/4C_io_input_types.hpp
```diff
@@ -28,10 +30,41 @@ namespace Core::IO
...
+    // used as enum names anyways
+    constexpr std::array illegal_enum_names = {
+        "True",
+        "TRUE",
+        "False",
+        "FALSE",
+        "null",
+        "Null",
+        "NULL",
+    };
```
mayrmt, line 44: @davidrudlstorfer The comment mentions, that "inf" and "nan" are illegal, but they are not listed in the `illegal_enum_names`.

davidrudlstorfer: The comment mentions that various combinations of a `.` (dot) and `inf`/`nan` are illegal not the keywords alone. These names are not able to be used as enum names anyways because we have other checks in place, i.e., clang tidy.

@sebproell am I missing something here?

mayrmt: Ok, then I just misunderstood the comment. 

So, is `inf` now an allowed input value?

davidrudlstorfer: Yes `inf` is allowed, otherwise the already merged changes from https://github.com/4C-multiphysics/4C/pull/929 would not work because since then we use `inf` as an input variable (via an enum)

mayrmt: Thanks!

PR #974, opened 2025-07-08
Introduce `NOX::Nln::GroupBase` and remove `NOX::Epetra::Group`

vovannikov:
## Description and Context
I am working now on our own implementation of `NOX::Abstract::Group`, ultimately, `NOX::Epetra::Group` should be removed. As an intermediate step, I have decided to create base class `GroupBase` for our current group implementations.

Most of appearances of `NOX::Epetra::Group` across the code (except from the groups and `NOX::FSI::SDRelaxation`) have been removed.

#359, #379

### DIFF
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_group_base.hpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_group_base.hpp
@@ -1,0 +1,37 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_SOLVER_NONLIN_NOX_GROUP_BASE_HPP
+#define FOUR_C_SOLVER_NONLIN_NOX_GROUP_BASE_HPP
+
+/*----------------------------------------------------------------------------*/
+/* headers */
+#include "4C_config.hpp"
+
+#include <NOX_Epetra_Group.H>  // base class
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace NOX
+{
+  namespace Nln
+  {
+    class GroupBase : public ::NOX::Epetra::Group
+    {
+     public:
+      GroupBase(Teuchos::ParameterList& printParams,
+          const Teuchos::RCP<::NOX::Epetra::Interface::Required>& i, const ::NOX::Epetra::Vector& x,
+          const Teuchos::RCP<::NOX::Epetra::LinearSystem>& linSys);
+
+      GroupBase(const NOX::Nln::GroupBase& source, ::NOX::CopyType type);
+    };  // class GroupBase
+  }  // namespace Nln
+}  // namespace NOX
+
+FOUR_C_NAMESPACE_CLOSE
+
+#endif
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
@@ -112,9 +112,7 @@ bool NOX::Nln::LineSearch::Backtrack::compute(::NOX::Abstract::Group& grp, doubl
   // want to access the jacobian, we have to use the oldGrp
   // (target of the copy process), instead.                hiermeier 08/15
   // ----------------------------------------------------------------------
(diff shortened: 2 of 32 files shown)


src/solver_nonlin_nox/4C_solver_nonlin_nox_group_base.hpp
```diff
@@ -0,0 +1,37 @@
...
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_SOLVER_NONLIN_NOX_GROUP_BASE_HPP
+#define FOUR_C_SOLVER_NONLIN_NOX_GROUP_BASE_HPP
+
+/*----------------------------------------------------------------------------*/
+/* headers */
```
sebproell:
```suggestion
```

not terribly important, but this comment serves no point.

sebproell (approved):
LGTM! 

Another thought: Maybe we should name our `NOX` namespace something like `NoxWrappers`? It's always super-hard to read if something is ours or truly NOX.

vovannikov:
> LGTM!
> 
> Another thought: Maybe we should name our `NOX` namespace something like `NoxWrappers`? It's always super-hard to read if something is ours or truly NOX.

That's a good idea, I fully agree, the current namespaces are so confusing.

maxfirmbach:
> LGTM!
> 
> Another thought: Maybe we should name our `NOX` namespace something like `NoxWrappers`? It's always super-hard to read if something is ours or truly NOX.

Or should we just put that stuff in the namespace `Nonlinear`?

sebproell:
@maxfirmbach @vovannikov I think that `NoxWrappers` tells more clearly what the namespace is about. `Nonlinear` is too general; it should at least be `NonlinearSolver`. I like the explicit reference to `NOX` since there are countless other packages one could use for the solution of nonlinear systems.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
```diff
@@ -112,9 +112,7 @@ bool NOX::Nln::LineSearch::Backtrack::compute(::NOX::Abstract::Group& grp, doubl
   // want to access the jacobian, we have to use the oldGrp
   // (target of the copy process), instead.                hiermeier 08/15
   // ----------------------------------------------------------------------
-  const ::NOX::Epetra::Group* epetraOldGrpPtr = dynamic_cast<const ::NOX::Epetra::Group*>(&oldGrp);
-  if (not epetraOldGrpPtr->isJacobian())
-    throw_error("compute()", "Ownership changed unexpectedly!");
+  if (not oldGrp.isJacobian()) throw_error("compute()", "Ownership changed unexpectedly!");
```
mayrmt, line 115: Shouldn't we use our own `FOURC_*` macro to throw?

src/structure/4C_structure_timint_noxgroup.hpp
```diff
@@ -34,12 +35,12 @@ namespace NOX
   {
     /*==================================================================*/
     /*!
-     * \brief Special ::NOX::Epetra::Group that always //ToDo (mayr) That's not correct anymore.
+     * \brief Special Group that always //ToDo (mayr) That's not correct anymore.
      *        sets Jacobian and RHS at the same time.
```
mayrmt: Apparently, I left a comment on this some time ago, but I don't remember what it means and if it is still relevant.

mayrmt (commented):
LGTM.

Regarding `NOX::FSI::SDRelaxation`, this might go away, cf. #900.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
vovannikov: We should I guess, but I would prefer to do that in a different PR, as `throw_error` is used not only here, but all across this file. Let's keep it consistent.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group_base.hpp
vovannikov: Yes, I have removed that.

src/structure/4C_structure_timint_noxgroup.hpp
vovannikov: I have changed the comment such that it better reflects the reality.

sebproell (approved):

PR #980, opened 2025-07-09
Create wrapper for Epetra FE Vector

knarfnitram:
## Description and Context

First draft for the wrapper of the `Epetra_FEVector` wrapper. Current implementation is originated from the `linalg_vector`. 
Let' see how many test are failing and how we proceed from here (check if some functionality may be removed, cleaned up, combined etc...)

https://gitlab.lrz.de/baci/baci/-/issues/1319

### DIFF
--- a/src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
+++ b/src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
@@ -18,11 +18,10 @@
 #include "4C_fem_discretization.hpp"
 #include "4C_geometry_pair.hpp"
 #include "4C_global_data.hpp"
+#include "4C_linalg_fevector.hpp"
 #include "4C_linalg_utils_sparse_algebra_manipulation.hpp"
 #include "4C_linalg_utils_sparse_algebra_math.hpp"
 #include "4C_utils_exceptions.hpp"
-
-#include <Epetra_FEVector.h>
 
 FOUR_C_NAMESPACE_OPEN
 
@@ -224,7 +223,8 @@ void BeamInteraction::BeamToSolidMortarManager::setup()
   set_global_maps();
 
   // Create the global coupling matrices.
-  constraint_ = std::make_shared<Epetra_FEVector>(lambda_dof_rowmap_->get_epetra_block_map());
+  constraint_ =
+      std::make_shared<Core::LinAlg::FEVector<double>>(lambda_dof_rowmap_->get_epetra_block_map());
   constraint_lin_beam_ = std::make_shared<Core::LinAlg::SparseMatrix>(
       *lambda_dof_rowmap_, 30, true, true, Core::LinAlg::SparseMatrix::FE_MATRIX);
   constraint_lin_solid_ = std::make_shared<Core::LinAlg::SparseMatrix>(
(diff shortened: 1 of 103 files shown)


src/core/linalg/src/sparse/4C_linalg_fevector.hpp
```diff
@@ -0,0 +1,323 @@
...
+
+    explicit FEVector(const Epetra_Vector& Source);
+
+
+    // Rule of five: We currently need to take care to make a deep copy of the Epetra_FEVector.
+    FEVector(const FEVector& other);
+
+    FEVector& operator=(const FEVector& other);
+
+    ~FEVector() = default;
```
sebproell, line 53: To really follow the rule of five, can you default the move constructor and move assignment?

src/core/linalg/src/sparse/4C_linalg_fevector.hpp
```diff
@@ -0,0 +1,323 @@
...
+
+    //! Map from Epetra_FEVector
+    mutable View<const Map> map_;
+
+    //! MultiVector view of the FEVector. This is used to allow implicit conversion to MultiVector.
+    mutable View<MultiVector<T>> multi_vector_view_;
+
+    friend class MultiVector<T>;
+  };
+
```
sebproell, line 305:
```suggestion

  template <>
  struct EnableViewFor<Epetra_FEVector>
  {
    using type = FEVector<double>;
  };
```

Something like this is missing to enable the `View<>` mechanism.

sebproell (commented):
@knarfnitram Thanks! It's looking quite good. Can you be more specific in telling us what is not working with the views? Maybe write a small test, similar like the tests we have for `Vector`.

src/core/linalg/src/sparse/4C_linalg_fevector.hpp
knarfnitram: @sebproell  thanks for the suggestion. I added it including some adapted tests from the vector, but I think that there's still something wrong with the conversion to our multi vector interface.

src/core/linalg/src/sparse/4C_linalg_fevector.hpp
knarfnitram: If I add the move constructor and move assignment, the implicit conversion to the multivector fails. I propose to  skip that.

src/core/linalg/src/sparse/4C_linalg_fevector.hpp
knarfnitram: I think now all test should pass again.

src/core/linalg/src/sparse/4C_linalg_fevector.hpp
sebproell: I saw that we also miss it for the `Vector`. I think we should fix it in both places. We can merge this PR, but I will have a look what needs to be done for `Vector`. The same will likely apply here.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
```diff
@@ -548,11 +551,10 @@ void BeamInteraction::BeamToSolidMortarManager::evaluate_and_assemble_global_cou
...
 
   // Complete the global scaling vector.
-  if (0 != kappa_->GlobalAssemble(Add, false))
-    FOUR_C_THROW("Failed to perform FE assembly of kappa_.");
-  if (0 != lambda_active_->GlobalAssemble(Add, false))
+  if (0 != kappa_->global_assemble()) FOUR_C_THROW("Failed to perform FE assembly of kappa_.");
+  if (0 != lambda_active_->global_assemble())
     FOUR_C_THROW("Failed to perform FE assembly of lambda_active_.");
-  if (0 != constraint_->GlobalAssemble(Add, false))
+  if (0 != constraint_->global_assemble())
```
maxfirmbach: We might want to call this `complete()` at some point to be in line with the other data structures and the respective calls there.

src/core/linalg/src/sparse/4C_linalg_fevector.hpp
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    int sum_into_local_value(int MyRow, int FEVectorIndex, double ScalarValue)
+    {
+      return vector_->SumIntoMyValue(MyRow, FEVectorIndex, ScalarValue);
+    }
+
+
+    int sum_into_local_value(
+        int MyBlockRow, int BlockRowOffset, int FEVectorIndex, double ScalarValue)
```
maxfirmbach: I'm pretty sure we don't use this, right?

knarfnitram: Let's find out.

isteinbrecher (approved):

PR #989, opened 2025-07-11
Exodus reader always checks for errors

sebproell:
We previously ignored some errors and warnings that EXODUS could generate.

### DIFF
--- a/src/core/io/src/4C_io_exodus.cpp
+++ b/src/core/io/src/4C_io_exodus.cpp
@@ -8,6 +8,7 @@
 #include "4C_io_exodus.hpp"
 
 #include "4C_fem_general_utils_local_connectivity_matrices.hpp"
+#include "4C_io_pstream.hpp"
 #include "4C_utils_enum.hpp"
 
 #include <exodusII.h>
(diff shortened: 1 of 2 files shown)


src/core/io/src/4C_io_exodus.cpp
```diff
@@ -20,6 +21,24 @@
...
+    {                                                                                          \
+      /* Use EnumTools for a direct interpretation of the error value. */                      \
+      auto error_enum = static_cast<ex_error_return_code>(error);                              \
+      FOUR_C_THROW("Exodus II failed with error code {}: {}", error_enum, #call);              \
+    }                                                                                          \
+    if (error > 0)                                                                             \
+    {                                                                                          \
+      Core::IO::cout << std::format("Exodus II returned a warning code {}: {}", error, #call); \
+    }                                                                                          \
+  } while (false)
```
davidrudlstorfer, line 39:
```suggestion
#define CHECK_EXODUS_CALL(call)
  do
  {
    int error = call;
    if (error < 0)
    {
      /* Use EnumTools for a direct interpretation of the error value. */
      auto error_enum = static_cast<ex_error_return_code>(error);
      FOUR_C_THROW("Exodus II failed with error code {}: {}", error_enum, #call);
    }
    if (error > 0)
    {
      Core::IO::cout << std::format("Exodus II returned a warning code {}: {}", error, #call);
    }
  } while (false)
```

can these be removed? I've never seen something like this before

davidrudlstorfer (approved):
Just a minor question/improvement

src/core/io/src/4C_io_exodus.cpp
ischeider: In fact it seems one has to have a "\" at the end of multi-line #define macros, since the pre-processor assumes a define statement to be a single line. In 4C you'll find many occurrences of such constructs.

davidrudlstorfer: Thanks for the explanation:)

Issue #935, opened 2025-06-26
Further improvements for InputField


sebproell:
We added the new `InputField` feature in #866 and its input in #890. A lot of people are interested in this feature and quickly found some issues that we still need to fix:

- [ ] Support for large meshes: this wastes too much memory right now since we cannot partition the field alongside the discretizaion.
 (e.g. @maxfirmbach 5e6 elements, with 9 values per element, not readable on 40 ranks -> out of memory)
- [ ] Hard to use if element ids are inside an EXODUS file or generated from a `domain` generator (@gilrrei)

sebproell:
@maxfirmbach One possible way to improve the large mesh situation: the `InputField` is probably tied more closely to a discretization than anticipated. If we manage the input field alongside the discretization, we could wait until the discretization is partitioned, then read the file and partition the data accordingly. However, this would require a different way to access the data e.g. inside a material (the one use case we currently had).

gilrrei:
Global ID questions:
- Do they start at 0? (I think so)
- Are they continuous if in the input file they are not continuous? (I don't think so)
- Is there a particular order to read them?

sebproell:
@gilrrei I would say that all of the above are unspecified. In fact, we should try not to rely on any of the implementation answers to these questions. See e.g. #733 where I tried to change something and (so far) failed.

maxfirmbach:
> [@maxfirmbach](https://github.com/maxfirmbach) One possible way to improve the large mesh situation: the `InputField` is probably tied more closely to a discretization than anticipated. If we manage the input field alongside the discretization, we could wait until the discretization is partitioned, then read the file and partition the data accordingly. However, this would require a different way to access the data e.g. inside a material (the one use case we currently had).

@sebproell I'm fine with any first solution that makes the input field work on large scale.

lauraengelhardt:
Summary from our discussion @sebproell @maxfirmbach on how to tackle supporting input fields for large meshes
- Have a top level input field section (similar to functions)
  - Define fields with a name that should be read from json files (`name: file_path`)
  - In the future reading the input field from a mesh could also be supported
  - Example:
    ```
    fields:
      - young_field: 
          from_file: "field_path.json"
      - young_field_stored_in_mesh: 
          from_mesh: ... (maybe in the future)
    ```
- Refer to fields in that section within the input_field spec
  - Example:
    ```
      YOUNGS: 
        constant: 3 
      YOUNGS:
        field: "young_field"
      YOUNGS:
[...]

PR #991, opened 2025-07-12
Redistribute an InputField according to element column map

sebproell:
This should fix one of the issues in #935.

@maxfirmbach Could you give this a try on your example? I adapted the thermo annulus test to the new style with global fields.

I have to say this code has a few places where I hack around in global state, but I don't see a reasonable way that will work with all our convoluted legacy code paths (e.g. calling down into a material). This only affects the input and setup phase. Importantly, I tried to keep the InputField free from any global references so that new code can use this cleanly without pulling in global state if not desired.


src/core/io/src/4C_io_input_field.hpp
```diff
@@ -49,7 +121,40 @@ namespace Core::IO
...
+    /**
+     * Construct an InputField that refers to a centrally registered field. The necessary @p ref
+     * may be obtained by calling the InputFieldRegistry::register_field_reference() function.
+     * The resulting InputField will not be useable until the `set_up()` function is called with
+     * the desired target map. When using the global input field registry and 4C's standard main
+     * function this is done automatically.
+     */
+    explicit InputField(InputFieldReference ref) : ref_(ref)
+    {
+      ref.registry->attach_input_field(ref, std::bind_front(&InputField::set_up, this), this);
```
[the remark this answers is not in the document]
sebproell: This is checked in all special member functions.

maxfirmbach:
@sebproell Did several test runs on our cluster. Currently I run into following issue:
```terminal
InputField index 0 is less than 1. All indices must be >= 1.
```

maxfirmbach:
> @sebproell Did several test runs on our cluster. Currently I run into following issue:
> 
> ```
> InputField index 0 is less than 1. All indices must be >= 1.
> ```

Okay, my bad, wrong index start in the input_field file.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -578,24 +579,39 @@ namespace Core::IO
...
+              case InputFieldType::constant:
+              {
+                T data = in.get<T>("constant");
+                auto field = IO::InputField<T>(data);
+                return store(out, std::move(field));
+                break;
+              }
+              case InputFieldType::field:
+              {
+                std::string field_name = in.get<std::string>("field");
```
maxiludwig: Can we find a better name for this? Essentially both, `field` and `from_file`, read from a json/yaml file.
Something like `parallelized_field` came into my mind but I'm not super happy with it.

sebproell: `field_reference`, `reference` or just `ref` come to mind.

maxiludwig: I would opt for `field_reference`

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -10,16 +10,87 @@
...
+
+  /**
+   * Refer to an input field by a name. This name is used to look up the input field in a registry
+   * of known fields.
+   */
+  struct InputFieldReference
+  {
+    //! The name which is used to uniquely identify this input field.
+    std::string ref_name;
+    InputFieldRegistry* registry;
```
lauraengelhardt, line 39:
```suggestion
    //! The registry of known input fields
    InputFieldRegistry* registry;
```

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -49,7 +121,40 @@ namespace Core::IO
      * Construct an InputField from a map of element-wise data. The @p data map contains
      * element indices as keys and the corresponding values of type T.
```
lauraengelhardt:
```suggestion
     * Construct an InputField from a map of element-wise data. The @p data map contains
     * element indices as keys and the corresponding values of type T. Element indices are converted to follow zero-based indexing.
```

tests/input_files/thermo3D_annulus_elementwise_anisotropic_statics.4C.yaml
```diff
@@ -31,7 +31,14 @@ MATERIALS:
...
       CAPA: 420
       CONDUCT:
-        from_file: conductivity_input_field.json
+        field: spatially_varying_conductivity
+
+fields:
+  - name: spatially_varying_conductivity
+    discretization: "thermo"
+    file: conductivity_input_field.json
+    key: "CONDUCT"
```
lauraengelhardt, line 40: Is this key necessary because you may want to define multiple fields in the json file?

sebproell: Yes.

src/core/io/src/4C_io_input_spec_builders.hpp
lauraengelhardt: If I get it right we currently support both the options
```
  YOUNGS: 
    from_file: "abc.json"
```
and
```
  YOUNGS:
    field: "young_field"
```

Shouldn't we just entirely switch to option 2?

sebproell: I don't think so. Option 2 requires a global registry that you may not want to pull in (in new code). However,  we should probably extend Option 1 with a lazy reading at some point, so one can distribute the field as done automatically for Option 2.

lauraengelhardt: Ok, agree.

I just think that from a user perspective it's not really intuitive whether to use `from_file` or `field_reference`...but I also don't have a better idea right now.

lauraengelhardt:
@sebproell is on vacation, so I suggest we merge the current state. We can still change the naming if we come up with a better solution in a follow-up PR.

PR #992, opened 2025-07-14
[IMCS] Update workstation preset

maxfirmbach:
As the title states. Building with shared libraries set to `off`. Build the debug version with asan, assertions and failing on compiler warnings.

### DIFF
--- a/presets/imcs/workstation/CMakePresets.json
+++ b/presets/imcs/workstation/CMakePresets.json
@@ -19,7 +19,8 @@
         "FOUR_C_BUILD_DOXYGEN": "ON",
         "FOUR_C_WITH_GOOGLETEST": "ON",
         "FOUR_C_WITH_ARBORX": "ON",
-        "FOUR_C_WITH_MIRCO": "ON"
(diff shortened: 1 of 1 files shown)


presets/imcs/workstation/CMakePresets.json
```diff
@@ -19,7 +19,8 @@
         "FOUR_C_BUILD_DOXYGEN": "ON",
         "FOUR_C_WITH_GOOGLETEST": "ON",
         "FOUR_C_WITH_ARBORX": "ON",
-        "FOUR_C_WITH_MIRCO": "ON"
+        "FOUR_C_WITH_MIRCO": "ON",
+        "FOUR_C_BUILD_SHARED_LIBS": "OFF"
```
sebproell, line 23: You are, of course, free to do this, but I would be interested if there is a specific reason you like static libs here.

maxfirmbach: With our current IT infrastructure and how we handle TPL's it's unfortunately not working without. At some point we might be able to switch back.

PR #995, opened 2025-07-14
Don't append timestamp to ccache key

amgebauer:
I don't really understand why the ccache action per default appends the timestamp to the cache key. The result is that we have a lot of redundant caches for single branches and might lose the (important) caches from the main branch. Without the timestap, new cache entries replaces old ones from the same branch.

@davidrudlstorfer


davidrudlstorfer (approved):
Thanks for looking into this 🙏🏻

PR #997, opened 2025-07-15
Remove undefined behaviour for access of empty std::vectors

marksidoro:
Below, I have updated some places at which we currently have code of the form `&vec[0]`, which is undefined behaviour if the vector is empty, as the access operator is not defined in that case. My gcc asserts this in debug mode, that's how I came across the issue. 
My proposed solution is to use std::vector::data() plus pointer arithmetics, as data() is guaranteed to return null for an empty vector. However this takes out the assertion in debug mode completely so maybe the approach depicted in `src/core/rebalance/src/4C_rebalance_graph_based.cpp` is more suitable.

### DIFF
--- a/src/core/comm/src/4C_comm_pack_buffer.hpp
+++ b/src/core/comm/src/4C_comm_pack_buffer.hpp
@@ -111,7 +111,7 @@ namespace Core::Communication
 #ifdef FOUR_C_ENABLE_ASSERTIONS
       // Check that the type matches the type that was packed.
       std::size_t hash;
-      std::memcpy(&hash, &data_[position_], sizeof(hash));
+      std::memcpy(&hash, data_.data() + position_, sizeof(hash));
       position_ += sizeof(hash);
       FOUR_C_ASSERT(hash == typeid(T).hash_code(),
           "Type mismatch during unpacking. Tried to extract type {}", typeid(T).name());
@@ -119,7 +119,7 @@ namespace Core::Communication
 
       // We know that stuff is trivially copyable, so we can safely use memcpy. We cast to void* to
       // circumvent a GCC warning
-      std::memcpy(static_cast<void*>(&stuff), &data_[position_], sizeof(T));
+      std::memcpy(static_cast<void*>(&stuff), data_.data() + position_, sizeof(T));
       position_ += sizeof(T);
     }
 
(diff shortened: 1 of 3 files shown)


src/core/comm/src/4C_comm_pack_buffer.hpp
```diff
@@ -137,16 +137,17 @@ namespace Core::Communication
...
           "Type mismatch during unpacking. Tried to extract type {}", typeid(T).name());
 #endif
 
-
+      FOUR_C_ASSERT(position_ + stuff_size <= data_.size(),
+          "UnpackBuffer: Trying to extract more data than available in the buffer.");
       // We know that stuff is trivially copyable, so we can safely use memcpy. We cast to void* to
       // circumvent a GCC warning
-      std::memcpy(static_cast<void*>(stuff), &data_[position_], stuff_size);
+      memcpy(static_cast<void*>(stuff), data_.data() + position_, stuff_size);
```
sebproell:
```suggestion
      std::memcpy(static_cast<void*>(stuff), data_.data() + position_, stuff_size);
```

src/core/comm/src/4C_comm_pack_buffer.hpp
```diff
@@ -162,12 +163,13 @@ namespace Core::Communication
...
-      std::memcpy(&hash, &data_[position_], sizeof(hash));
+      std::memcpy(&hash, data_.data() + position_, sizeof(hash));
       position += sizeof(hash);
       FOUR_C_ASSERT(hash == typeid(T).hash_code(), "Type mismatch during unpacking.");
 #endif
-
-      memcpy(&stuff, &data_[position], sizeof(T));
+      FOUR_C_ASSERT(position < data_.size(),
+          "UnpackBuffer: Trying to extract more data than available in the buffer.");
+      memcpy(&stuff, data_.data() + position, sizeof(T));
```
sebproell:
```suggestion
      std::memcpy(&stuff, data_.data() + position, sizeof(T));
```

sebproell:
> However this takes out the assertion in debug mode completely so maybe the approach depicted in src/core/rebalance/src/4C_rebalance_graph_based.cpp is more suitable.

I don't understand this. What does 

>  takes out the assertion in debug mode completely 

mean?

sebproell (commented):
Can you add a test for the empty vector case  in `4C_comm_pack_unpack_test.np3.cpp`?

marksidoro:
Currently, the vector performs an index check during the `operator[](size_t i)` call i.e. if i < vector.size(). Using the data() + i approach this Assertion is not done. 
But during a call like  `vector.empty() ? null_ptr :   &vector[i]` the assertion would remain.

src/core/rebalance/src/4C_rebalance_graph_based.cpp
```diff
@@ -301,7 +301,8 @@ std::shared_ptr<const Core::LinAlg::Graph> Core::Rebalance::build_graph(
     }
     Core::Communication::broadcast(&size, 1, proc, dis.get_comm());
     if (proc != myrank) recvnodes.resize(size);
-    Core::Communication::broadcast(&recvnodes[0], size, proc, dis.get_comm());
+    Core::Communication::broadcast(
+        recvnodes.empty() ? nullptr : &recvnodes[0], size, proc, dis.get_comm());
```
amgebauer: Is it allowed to broadcast the nullptr?


```suggestion
    if (!recvnodes.empty()) Core::Communication::broadcast(&recvnodes[0], size, proc, dis.get_comm());
```

amgebauer (commented):
> My proposed solution is to use std::vector::data() plus pointer arithmetics, as data() is guaranteed to return null for an empty vector.

Isn't that still undefined behavior? Then you are adding something to the nullptr

marksidoro:
> Isn't that still undefined behavior? Then you are adding something to the nullptr

Not if we add zero as `nullptr +/- 0` is well defined. 
For i =nonzero case it is indeed undefined but so is vec[i] in the first place, so we don't "lose" anything

marksidoro:
After talking to @sebproell we decided to check the `stuff_size` int to make sure we only call `std::memcpy` if we should have valid vectors. 
I also did a global search and updated all places where we used `&vec[0]` and vec could feasibly be empty to use vec.data() instead

amgebauer (approved):

PR #999, opened 2025-07-17
Implement Einstein's summation convention

amgebauer:
This PR adds the Einstein's sum convention to 4C tensors. It can evaluate arbitrary einstein sum convention expressions. All operations are unfolded at compile time.

A few examples:

```c++
auto trace = einsum<"ii">(A); // trace of A = A_ii
auto A_t = einsum<"ji">(A); // transpose of A = A_ji
auto Ab = einsum<"ij", "j">(A, b); // A * b = A_ij b_j
auto bb = einsum<"i","j">(a, b); // b otimes b = b_i b_j
auto FTAF = einsum<"ji", "jk","kl">(F, A, F); // F^T * A * F = F_ji A_jk F_kl
```

There is also a symmetric version that results in a symmetric tensor (note that symmetry will not be checked but only assumed!)

```c++
auto bb = einsum_sym<"i","j">(a, b); // b otimes b = b_i b_j
auto FTAF = einsum_sym<"ji", "jk","kl">(F, A, F); // F^T * A * F = F_ji A_jk F_kl
```

Note that this will unroll all loops at compile time. Sometimes, it still might be faster to split the operations to save a few floating point operations, e.g. F^T * A * F in 3D has `54m+36a` in tensor notation and `81m+81a` in einstein sum notation.


amgebauer:
I changed a little the interface. Now, the PR is ready for review. I use this in #961 and it will also simplify the implementations in https://github.com/4C-multiphysics/4C/blob/main/src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.cpp a lot.

amgebauer:
Can someone please have a look?

src/core/linalg/src/dense/4C_linalg_tensor_einstein.hpp
```diff
@@ -0,0 +1,564 @@
...
+   */
+  template <EinsteinHelper::FixedString... einstein_indexes, typename... Tensor>
+    requires(sizeof...(einstein_indexes) == sizeof...(Tensor))
+  auto einsum(const Tensor&... t)
+  {
+    using TensorCompressionTypes = std::tuple<TensorCompressionType<Tensor>...>;
+
+    constexpr auto min_index = EinsteinHelper::smallest_used_index<einstein_indexes...>;
+
+    constexpr auto all_indexes = EinsteinHelper::all_used_indexes<einstein_indexes...>;
```
sebproell:
```suggestion
    constexpr auto all_indices = EinsteinHelper::all_used_indices<einstein_indices...>;
```

Search and replace `indexes` with `indices`.

src/core/linalg/src/dense/4C_linalg_tensor_einstein.hpp
```diff
@@ -0,0 +1,564 @@
...
+    {
+      return [&]<std::size_t... i>(std::integer_sequence<std::size_t, i...> seq)
+      {
+        return TensorCompression::template flatten_index<Internal::TensorBoundCheck::no_check>(
+            i...);
+      }(IntegerSequenceFromArray<index>());
+    }();
+
+    template <std::size_t rank, std::size_t num_contraction_indexes, std::size_t num_dyadic_indexes>
+    consteval std::array<std::size_t, rank> get_tensor_index(std::array<std::size_t, rank> indexes,
```
sebproell: I think a short doc would help for these longer helper functions.

sebproell:
Super nice! I really like the user-facing code.

src/core/linalg/src/dense/4C_linalg_tensor_einstein.hpp
```diff
@@ -0,0 +1,619 @@
...
+     * The contraction indices are [k], the dyadic indices are [i, j].
+     *
+     * Let's say, we want to evaluate the above equation for i=1, j=2, k=3, the given parameters
+     * are:
+     *
+     * indices = [0, 1] # corresponds to i, j
+     * contraction_indices = [2] # corresponds to k
+     * dyadic_indices = [0, 1] # corresponds to i, j
+     * contraction_index = [3]
+     * dyadic_index = [1, 2]
```
sebproell: Thanks for the docs! What I don't get here:

- What does `k=3` mean? Don't we need to sum over all `k`s to compute `A_12`? Or is this just one summand? 
- I also don't understand why the indices are shifted by 1 in the list of different indices, i.e., contraction_indices is 2 but contraction_index is 3.

These are all internals, so I let you be the judge of whether this (internal) doc needs to be improved.

sebproell (dismissed):

src/core/linalg/src/dense/4C_linalg_tensor_einstein.hpp
amgebauer: Maybe I'm not clear enough with the documentation here.

indices, contraction indices, and dyadic indices correspond to the axis (not a specific component), i.e., i, j, k, as an integer type (i=0, j=1, ...). For A_ij=B_ik C_kj, the contraction indices are [k] and the dyadic indices are [i, j]. We convert them to an integer type so that we can more easily handle them, i.e., [2] and [0, 1], respectively. These remain the same for all iterations of the Einstein summation.

The contraction index and dyadic index are the specific components for each of the given contraction/dyadic indices. These change at each iteration since we loop over all contraction/dyadic indices.

Maybe I can make this a bit clearer if I convert the first three arguments into a template argument and instantiate this function outside the for loop.

amgebauer: I changed a little the names, basically, I refer to "ij" as index_names instead of indices to not get confused with the indices (0,1)

sebproell: Sorry, I still don't get it. I understand what you mean with contraction and dyadic, but I am still confused why there is a second set of the same indices that are shifted by one in the example.

amgebauer: ok, now it should be clearer (I restructured it a little more).

The first and the second ones are structurally different. The first one corresponded to the name of the axis (i, j) represented as an integer index, and the second one corresponded to the respective value. I have now introduced a new type of EinsteinIndex, which makes a few things simpler and also makes this function here more obvious.

src/core/linalg/src/dense/4C_linalg_tensor_einstein.hpp
```diff
@@ -0,0 +1,628 @@
...
+   *
+   * @throws Compilation error if index constraints are violated.
+   */
+  template <EinsteinHelper::FixedString... einstein_indices, typename... Tensor>
+    requires(sizeof...(einstein_indices) == sizeof...(Tensor))
+  auto einsum(const Tensor&... t)
+  {
+    using TensorCompressionTypes = std::tuple<TensorCompressionType<Tensor>...>;
+
+    constexpr auto all_einstein_indexes =
```
sebproell: Now you have "indexes" again a few times.

src/core/linalg/src/dense/4C_linalg_tensor_einstein.hpp
```diff
@@ -0,0 +1,628 @@
...
+     *
+     * A_ij = B_ik C_kj.
+     *
+     * The contraction indices are [k], the dyadic indices are [i, j].
+     *
+     * Let's say, we want to evaluate the above equation for i=1, j=2, k=3, the given parameters
+     * are:
+     *
+     * my_tensor_einstein_index_names = [i, j]
+     * contraction_einstein_indices contraction_index_names = [k]
```
sebproell:
```suggestion
     * contraction_einstein_indices = [k]
```
?

src/core/linalg/src/dense/4C_linalg_tensor_einstein.hpp
```diff
@@ -0,0 +1,628 @@
...
+    /*!
+     * @brief Returns the tensor indices of a component for a tensor in an Einstein summation.
+     *
+     * Take the matrix-matrix product as an example:
+     *
+     * A_ij = B_ik C_kj.
+     *
+     * The contraction indices are [k], the dyadic indices are [i, j].
+     *
+     * Let's say, we want to evaluate the above equation for i=1, j=2, k=3, the given parameters
```
sebproell: I think my confusion might stem from `i=1, j=2, k=3`. What does this mean? I don't know how to interpret `k=3` since this is a contracted index. I'll drop by in your office tomorrow and you can enlighten me ;)

src/core/linalg/src/dense/4C_linalg_tensor_einstein.hpp
```diff
@@ -0,0 +1,618 @@
...
+#include "4C_config.hpp"
+
+#include "4C_linalg_tensor.hpp"
+#include "4C_linalg_tensor_internals.hpp"
+#include "4C_linalg_tensor_meta_utils.hpp"
+#include "4C_utils_compile_time_string.hpp"
+#include "4C_utils_exceptions.hpp"
+#include "4C_utils_fad_meta.hpp"
+
+#include <bits/utility.h>
```
sebproell:
```suggestion
```

sebproell (commented):
Thanks for iterating over the docs (of the internals)! I think the explanation is now very helpful.

PR #1000, opened 2025-07-17
Clean up scatra for pressure-based porofluid framework

bwirthl:
This PR cleans up the `scatra_ele_calc` class for scatra as part of the pressure-based porofluid framework.


src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -255,34 +255,35 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
-                {.description = "relative mobility function ID", .default_value = 0}),
-        },
-        {.description =
-                "advanced reaction material for multiphase porous flow (species in fluid)"});
+    known_materials[Core::Materials::m_scatra_in_fluid_porofluid_pressure_based] =
+        group("MAT_scatra_multiporo_fluid",
+            {
+                parameter<double>("DIFFUSIVITY", {.description = "kinematic diffusivity"}),
+                parameter<int>(
+                    "PHASEID", {.description = "ID of fluid phase the scalar is associated with"}),
```
lkoeglmeier:
```suggestion
                    "PHASEID", {.description = "ID of fluid phase the scalar is associated with. Staring with zero"}),
```
At least I was confused the first time I used it :)

lkoeglmeier (dismissed):
Looks good to me! I only have one suggestion. Thanks a lot @bwirthl!

src/core/legacy_enum_definitions/4C_legacy_enum_definitions_materials.cpp
```diff
@@ -43,13 +43,13 @@ std::string_view Core::Materials::to_string(Core::Materials::MaterialType materi
       return "MAT_scatra_reaction_poro";
     case m_scatra_reaction:
       return "MAT_scatra_reaction";
-    case m_scatra_multiporo_fluid:
+    case m_scatra_in_fluid_porofluid_pressure_based:
```
c-p-schmidt, line 46: I was a bit astonished that the naming seems not to be consistent to me. But I asked @lkoeglmeier about that, and she explained that this is still an intermediate step, which is totally fine for me.
Thus, approving.

c-p-schmidt (dismissed):

src/core/legacy_enum_definitions/4C_legacy_enum_definitions_materials.cpp
bwirthl: Yes, it is scalar transport in a fluid phase of the multiphase pore space as opposed to scalar transport in a fluid phase of the additional pore space (which we call `volfrac`). Honestly, I cannot come up with concise names for these and I do not want to go down this rabbit hole with my current clean-ups. Hence, I leave this to @lkoeglmeier at some point.

lkoeglmeier (approved):

c-p-schmidt (approved):

PR #1006, opened 2025-07-25
Enable use of spatial input field for mixture massfractions

lauraengelhardt:
## Description and Context
We used to have a separate mixture rule `MIX_Rule_Map` that would take element-wise defined mass fractions of the mixture constituents from a `pattern` file. With the introduction of the `InputField` this is now obsolete and can be and can be handled directly by the `Mix_Rule_Simple`.

This PR is organised as follows:
- 1st commit removes the  obsolete `MIX_Rule_Map` and the related test.
- 2nd commit: Clean up docs. 
  @amgebauer: If you would rather keep these lines, I can also convert them to yaml. I just believe they become outdated rather quickly and with the schema file, we have enough documentation about the input anyhow. 
- 3rd commit: `MIX_Rule_Simple` used to take the parameter `NUMCONST` that was solely used to determine the length of the massfractions vector. With yaml input this is not needed anymore. 
  @amgebauer: I willl clean this up for all remaining mixture materials in a separate PR.
- 4rd commit: Equip the InputField with a method `for_each_value(operation)`, allowing a function to be uniformly applied to all stored values, whether the values are single constant or element-wise data. In this case, this is required to validate that the mass fractions sum up to 1. @maxiludwig @bgoderbauer @sebproell Tell me what you think!
- 5th commit: Finally make the `MASSFRAC` an `InputField`. As you can see the code changes are really just a few lines - so making a parameter optionally spatially varying is really really easy :) 

#991 #866 #890


src/core/io/src/4C_io_input_field.hpp
```diff
@@ -171,6 +171,34 @@ namespace Core::IO
      */
     [[nodiscard]] const T& at(IndexType element_id) const { return get(element_id, true); }
 
+    /**
+     * Apply a function to each value in the InputField. If the InputField is a map, the function
+     * will be called for each value in the map. If the InputField is a single value, the function
+     * will be called once for that value.
+     */
+    template <typename F>
+    void for_each_value(F&& operation) const
```
sebproell: - The function assumes that there is a finite number of domain values. Currently, this is true, since the `InputField` can only depend on element IDs. However, if you think about a field that depends on arbitrary coordinates, the domain is infinite and I am not sure what this function should do in that case. 
- I think you can achieve what you need to do by just iterating over all values yourself? E.g., you may use all entries in the element row map and perform the check.

As a result, I think you don't need this function.

amgebauer (commented):
Awesome!

src/core/io/src/4C_io_input_field.hpp
lauraengelhardt: Good point!

My issue was that during construction of the mixture rule, I don’t yet know which element IDs it will be evaluated on, and I cannot access the full internal field data map without exposing it. I didn't want to perform the check in the evaluation routine either. I do it in the setup now.

lauraengelhardt:
Removing the remaining NUMCONST was actually very easy, so I just updated 44fda5c9fa67bff5da776866bad6cff42f3497bc.

Ready for another review.

maxiludwig (approved):

sebproell (approved):

Issue #1009, opened 2025-07-25
Caching of clang install job does not work properly


amgebauer:
## Bug Description: What's going wrong?

The idea is that the install job can build upon the cache from the previous stage. Unfortunately, in many cases, the correct cache is not found and an older cache is used instaed resulting in a long rebuild of the job.

Maybe it takes some time until the cache is updated so that the clang-install job only sees an older one?

## Possible Solution
Maybe explicitly transfer the cache with the artifacts?

amgebauer:
It seems like a bug that appeared very recently (within the last month). From the Github action performance metrices:

| Period | Runtime (average)|
|----------|----------|
| last month    | 3m20s   |
| this month    | 11m3s   |
| this week    |  43m 46s  |

davidrudlstorfer:
Could this https://github.com/4C-multiphysics/4C/pull/995 be a reason?

amgebauer:
> Could this [#995](https://github.com/4C-multiphysics/4C/pull/995) be a reason?

I don't know how this could be the reason. And there are still examples that work correctly after #995. I don't think that it is related. I suspect that Github has changed something in the background with their caching.

amgebauer:
We see this error during saving the cache:

> Failed to save: Unable to reserve cache with key ccache-clang18_build-docker_clang-, another job may be creating this cache.

PR #1012, opened 2025-07-28
Revert "Don't append timestamp to ccache key"

amgebauer:
The Github cache is immutable, so a cache-key can only be used once. So, I think it is sensible to append the timestamp.

Instead, we might manually delete old cache entries of a branch.


amgebauer:
It does not fix #1009, but is still correct (unfortunately)

PR #1013, opened 2025-07-28
Transfer ccache via artifacts for clang18 install

amgebauer:
Unfortunately, our previous approach does not work anymore. The ccache from the clang-build job is not found reliably. Now, we transfer the ccache directly via artifacts for the clang18_build -> clang18_build_install test.


amgebauer:
Now it works again. Unfortunately, the changes shown on Github look quite complicated, but they are quite simple:

 * Generalization of upload/download_4C_build: Can now upload/download any directory
 * clang18 uploads their ccache as an artifact
 * clang18_install uses the ccache from the artifact

.github/workflows/coverage.yml
```diff
@@ -5,7 +5,7 @@ on:
     - cron: '0 1 * * *'
 
 jobs:
-  gcc13_coverage_build:
+  clang18_coverage_build:
```
ppraegla, line 8: This is just a clean up of the name of the job, right? The coverage job was already running with clang before?

ppraegla (approved):

.github/workflows/coverage.yml
amgebauer: yes, the name was wrong. Sorry for polluting the changes in this PR 😬

.github/workflows/nightly_tests.yml
```diff
@@ -417,10 +425,12 @@ jobs:
           build-targets: full
           build-directory: ${{ github.workspace }}/build
           use-ccache: "false" # no ccache since asan is only tested once per day
-      - uses: ./.github/actions/upload_4C_build
+      - uses: ./.github/actions/
```
ppraegla, line 428:
```suggestion
      - uses: ./.github/actions/upload_directory
```
@amgebauer @sebproell Sorry, I did not catch that during review.

PR #1014, opened 2025-07-28
Add all_elements validator

davidrudlstorfer:
## Description and Context

Adds a validator to apply individual validators to elements in a vector.

@sebproell thanks for the pair-programming:)
@gilrrei as far as @sebproell and me can tell this should not break fourcipp?


src/core/io/src/4C_io_input_spec_validators.hpp
```diff
@@ -220,6 +220,17 @@ namespace Core::IO::InputSpecBuilders::Validators
...
   template <Internal::Numeric T>
   [[nodiscard]] auto positive();
+
+
+  /**
+   * A validator that checks if all elements in a vector satisfy the given validator.
+   *
+   * @note This is useful for validating lists of values, e.g., when you want to check if all
+   * elements in a list are positive integers.
+   */
```
sebproell:
```suggestion
  /**
   * A validator that checks if all elements in a vector satisfy the given @p validator, e.g., 
   *   
   * @code
   * // check that all vector elements are positive
   * all_elements(positive<int>()):
   * @endcode
   *
   */
```

gilrrei:
@davidrudlstorfer 
> @gilrrei as far as @sebproell and me can tell this should not break fourcipp?

This will break the creation of the schema file in 4C, see [here](https://github.com/4C-multiphysics/4C/blob/25880a602174319d8ada67b98ab2fee5e17437d8/utilities/metadata_utils.py#L465). We could remove the error, but actually, I think it is much more interesting to invest the time in this [python function](https://github.com/4C-multiphysics/4C/blob/25880a602174319d8ada67b98ab2fee5e17437d8/utilities/create_json_schema.py#L299) pushing the validator to the item, i.e. pass the validator to the type of the vector items itself. This would mean that fourcipp and your IDE checks each item of the vector is it fulfills the desired condition (for example ranges)

sebproell:
@gilrrei 

>  pushing the validator to the item, i.e. pass the validator to the type of the vector items itself.

I was wondering if that is possible. Sounds good. I think this will be a bit more involved for the general case. For instance, I am thinking about `std::vector<std::optional<int>>` or even `std::optional<std::vector<std::optional<int>>>` where `optional` adds the ability to have `null` in yaml (currently called `noneable` in metadata). It seems we need to restructure the metadata a little bit to enable (simple) schema generation for these cases. We already export the nested type with the `noneable` info; I am only talking about an additional validator. Just as we saw it for the enums, where the possible values are related to the validator, `null` may also be related to the validators. 

I guess what I am trying to say: Can we delay better validators in the schema for a few PRs? I have a follow-up PR that enables validators for the optional types mentioned above. This gives a fuller picture of what we can encounter in schemas.

gilrrei:
@sebproell 

> We already export the nested type with the `noneable` info; I am only talking about an additional validator. Just as we saw it for the enums, where the possible values are related to the validator, `null` may also be related to the validators.

I don't fully get your point. If I understand correctly, you want to do the `null`/optional check through the validator. I don't fully see how this makes things easier :sweat_smile:  but I trust your opinion on this. Nevertheless, I think the metadata export of optional parameters should still be with a `noneable` boolean variable (name up to discussion), not a validator expression. This makes things easier and, more importantly, clearer to understand.

> I guess what I am trying to say: Can we delay better validators in the schema for a few PRs? I have a follow-up PR that enables validators for the optional types mentioned above. This gives a fuller picture of what we can encounter in schemas.

I am not a big fan of follow-ups, as IMO, in general, they tend not to get done. However, since it is backward compatible, I think delaying this is justified. Just a heads-up, I won't be able to provide the changes in the schema generation, but I would be available for PR reviews of the changes.

sebproell:
@gilrrei I think one should distinguish "follow-up PRs" based on whether they fix a problem that is introduced in the current PR in the first place (I don't like these), or whether they just add on top of the PR with more features (totally fine). Here, we have the latter case.

My reasoning for the follow-up: let me first add another validator (related to `null`), then I will better understand how to enable the new validators in schemas. And yes, my explanation is not super clear, this is simpler to see in code, I think. Also, it is probably not super clear, since I am not exactly sure how the solution will turn out, e.g., I am not sure whether `null` will be done through the validators; I just realized that `nullability` somehow appears in there.

davidrudlstorfer:
so @gilrrei @sebproell how should we continue here?

sebproell:
@davidrudlstorfer just adress the one comment, remove the error print in the python script and I will add more validators and adapt the schema.

davidrudlstorfer:
@sebproell now the remaining points are addressed - sorry for the delay

utilities/metadata_utils.py
```diff
@@ -463,10 +463,6 @@ class RangeValidator(Validator):
 
 
 def validator_from_dict(validator_dict):
-    if len(validator_dict) > 1:
-        raise ValueError(
-            f"Currently only a single validator function is allowed, you provided {validator_dict}"
-        )
```
sebproell: I think this error should stay. It is the one below (in the switch) that you need to drop.

davidrudlstorfer: Sorry - changed it

bgoderbauer (approved):

PR #1020, opened 2025-07-31
[structure] Hookup of forcing term methods to the nonlinear Newton solver

maxfirmbach:
## Description and Context

The nonlinear Newton solver in the new structure time integration always used a constant forcing term so far, prescribing a constant linear solver tolerance for each step. For highly nonlinear problems and/or difficulties in converging it makes sense to relax the linear solver tolerance and let the Newton method decide how accurate the linear system should be solved.

This often results in a great wall clock time improvement, as the first few steps of the Newton method are using a rather inaccurate solution of the linear solve, but improve accuracy where necessary. This is especially interesting with reusing the preconditioner across several iterations.

I adapted a test where we already check for linear and nonlinear iterations. Also the details of the NOX internals are printed in addition resulting in e.g. the following output (consider, the linear solver tolerance in the input file is given by 1e-10):
```terminal
-- Nonlinear Solver Step 2 -- 
||F|| = 9.73190e-07  step = 1.00000e+00  dx = 9.27563e-02

       CALCULATING FORCING TERM
       Method: Type 2
       Forcing Term Norm: Using L-2 Norm.
       Residual Norm k-1 =             0.000211234
       Residual Norm k =               9.7319e-07
       Calculated eta_k (pre-bounds) = 0.000281444
       Forcing Term: 0.000281444

*******************************************************
***** Belos Iterative Solver:  Block Gmres 
***** Maximum Iterations: 1000
***** Block Size: 1
***** Residual Test: 
*****   Test 1 : Belos::StatusTestImpResNorm<>: (2-Norm Res Vec) / (2-Norm Res0), tol = 0.000281444
*******************************************************
Iter    0, [ 1] :    1.000000e+00
Iter    5, [ 1] :    1.220747e-04
```

In this case the forcing term relaxes the linear solver tolerance to 0.000281444, which is then used for the solve.

### DIFF
--- a/tests/input_files/embeddedmesh_cantilever_beam.4C.yaml
+++ b/tests/input_files/embeddedmesh_cantilever_beam.4C.yaml
@@ -21,10 +21,8 @@ STRUCTURAL DYNAMIC:
   PREDICT: "TangDis"
   LINEAR_SOLVER: 1
 SOLVER 1:
-  SOLVER: "Belos"
-  IFPACK_XML_FILE: "xml/preconditioner/ifpack.xml"
-  SOLVER_XML_FILE: "xml/linear_solver/iterative_gmres_template.xml"
   NAME: "Struct_Solver"
+  SOLVER: "UMFPACK"
 XFEM GENERAL:
   GMSH_DEBUG_OUT: true
   GMSH_CUT_OUT: true


maxfirmbach:
@lauraengelhardt At some point this might be interesting for you.

mayrmt (approved):

mayrmt (approved):

vovannikov (dismissed):
Nice feature!

isteinbrecher (dismissed):

davidrudlstorfer:
I have relatively little knowledge regarding solvers and these concepts so far. Is this method also helpful for setups where you do not use a preconditioner at all? I am struggling with convergence for a long time with my highly nonlinear problem (beam potential interactions) so could this help? Thanks in advance!

mayrmt:
@davidrudlstorfer This is not related to preconditioning, only to iterative solvers for the linear system.

The proposed change lets the simulation adapt the tolerance of the linear solver based on the current state of the nonlinear solver. In essence, it changes the required accuracy of the solution of the linear problem in each Newton step (i.e. the accuracy of the search direction / Newton increment).

The question, whether you use a preconditioner and which one, does not affect the accuracy of the linear solver, only its convergence behavior (i.e. speed / number of iterations until convergence).

Note: If you use a direct solver, then these chances are irrelevant, since a direct solver does not use a convergence tolerance.

davidrudlstorfer:
@mayrmt thanks for the explanation 🙏🏻

lauraengelhardt:
> @lauraengelhardt At some point this might be interesting for you.

That point may be now 😆 Very nice, thanks for mentioning me!

mayrmt (approved):

maxfirmbach:
@isteinbrecher @eulovi I switched the failing test related to embedded mesh to `UMFPACK`. The behavior before was also not the best (many iterations).

isteinbrecher (approved):

PR #1024, opened 2025-08-01
Reword title of 4C citation

amgebauer:
I think we are missing the important open-source aspect in our citation title. What do you think?

@4C-multiphysics/maintainer

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ in the [GitHub discussions](https://github.com/4C-multiphysics/4C/discussions).
 Please cite 4C as follows:
 
 ```
-4C: A Comprehensive Multiphysics Simulation Framework, https://www.4c-multiphysics.org
+4C: An Open-Source Framework for Comprehensive Multiphysics Simulations, https://www.4c-multiphysics.org
 ```
 
 You could use the following BibTeX entry:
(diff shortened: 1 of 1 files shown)


ppraegla (approved):

c-p-schmidt:
I would be fine with the changes. But we should consistently rename the website and the citation file in the repo as well.

mayrmt:
I am hesitant to change this without further coordination. 

I can see that it might make sense to add "open-source" to explicitly mention it, since it still feels new. This will change over time, so we might not need to emphasize it anymore in the future.

sebproell:
I personally would not have come up with the idea to put "open-source" in the citation. Is that a common thing for other codes to do? I have no strong opinion, but as @mayrmt said, it can feel a bit weird further down the line, and we probably should rarely/never change this title. If we go this route, we might as well think about the "community" aspect, which is even in the unabbreviated project name. Also, the PMM might be interested in choosing a good title.

amgebauer:
Open-source is a crucial piece of information. The title plays a decisive role in whether someone chooses to explore a paper or a website further. Making it clear that our framework is open source significantly increases interest. Who is interested in yet another in-house simulation framework? That's why I would directly clarify in the title that we are open-source.

amgebauer:
Closing for now. See https://github.com/4C-multiphysics/maintainer-meeting/issues/23#issuecomment-3163336719

PR #1025, opened 2025-08-01
Remove duplicate parameter definition

mayrmt:
## Description and Context

By accident, we did define the minimum size of an aggregate twice in the aggregation sublist of MueLu. This PR fixes that.

### DIFF
--- a/tests/input_files/fsi_dc_mono_fs_ga_ga.4C.yaml
+++ b/tests/input_files/fsi_dc_mono_fs_ga_ga.4C.yaml
@@ -2958,7 +2958,7 @@ RESULT DESCRIPTION:
       NODE: 841
       QUANTITY: "dispx"
       VALUE: 3.0312271145639657e-05
-      TOLERANCE: 1e-12
+      TOLERANCE: 1e-11
   - ALE:
       DIS: "ale"
       NODE: 841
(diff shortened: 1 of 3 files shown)


mayrmt:
@maxfirmbach, alternatively, we could just not specify a maximum aggregate size at all (and just delete these lines). Do you have an opinion / preference?

sebproell (approved):
Shouldn't we have seen an error for duplicated parameters? I guess it is Trlinos' fault, but maybe we are not using a function correctly?

mayrmt:
I am surprised as well, but this problem is not part of 4C. We just pass the filename to MueLu and they read and process the xml file. So, yes, probably an issue in Trilinos.

I'll follow up with the MueLu team.

maxfirmbach:
@mayrmt I think we should delete these parameters completely from the input file and depend on the defaults, which are pretty good.

mayrmt:
@maxfirmbach I've removed the constraints on the aggregate size. Let's see how many tests are affected.

sebproell (approved):

c-p-schmidt:
@mayrmt: We have these settings in several other XML files as well. So I guess they should be removed there as well?

mayrmt:
@c-p-schmidt It depends on the problem. This PR mainly tackles the fact, that the same parameter has been defined twice. I checked all xml files and didn't find this issue anywhere else. In that sense, this PR is self-contained and ready to be merged. 

If others want to improve other preconditioner templates, they can do so, of course.

c-p-schmidt (approved):

PR #1028, opened 2025-08-01
Refactor beam potential interactions input

davidrudlstorfer:
## Description and Context

This PR now finally completely refactors the beam potential interactions input with all features of our new input spec.
In addition:

- it remove interaction potential capabilities from the old beam contact manager (unused + untested code which is replaced by the current beaminteractions module, see #1022)
- it enables one to directly get a struct from the input

@sebproell big thanks for everything input related, the new features are just amazing.


src/beamcontact/4C_beamcontact_beam3contact_manager.cpp
```diff
@@ -1474,115 +1304,6 @@ void CONTACT::Beam3cmanager::fill_contact_pairs_vectors(
...
-    // is sufficient to check only one node
-    Core::Nodes::Node** nodes1;
-    Core::Nodes::Node** nodes2;
-    nodes1 = ele1->nodes();
-    nodes2 = ele2->nodes();
-
-    FOUR_C_ASSERT(nodes1 != nullptr and nodes2 != nullptr, "pointer to nodes is nullptr!");
-    FOUR_C_ASSERT(nodes1[0] != nullptr and nodes2[0] != nullptr, "pointer to nodes is nullptr!");
-
-    nodes1[0]->get_condition("BeamPotentialLineCharge", conds1);
```
sebproell, line 1537: Nice to see this line gone! Related to #795

davidrudlstorfer:
And one code design question from my side which I do not fully understand:

https://github.com/4C-multiphysics/4C/blob/ac3ab52b526f242649afe068bfd87b36f6d02442/src/beaminteraction/src/potential/4C_beaminteraction_potential_pair_base.hpp#L192-L194

https://github.com/4C-multiphysics/4C/blob/ac3ab52b526f242649afe068bfd87b36f6d02442/src/beaminteraction/src/potential/4C_beaminteraction_potential_submodel_evaluator.hpp#L201-L203

Does it make sense to always save these parameters like this? I would probably just store them in the base sub-model evaluator (without making them shared) and just hand over a pointer to this data to each interaction pair.

But I am unsure on the effects of that

sebproell:
@davidrudlstorfer Yes, `shared_ptr` (or `Teuchos::RCP`)  is almost never what you want. If you can state to which object's lifetime these parameters are tied, then let that object own the paramters and just pass a reference/pointer to whichever objects need to use it. Of course, these objects may not live longer than the owner object. If you do not know enough about lifetime of your objects, you likely have more fundamental problems, but `shared_ptr` still only patches over the problem.

src/beaminteraction/src/potential/4C_beaminteraction_potential_pair_beam_to_sphere.cpp
```diff
@@ -121,7 +120,9 @@ bool BeamInteraction::BeamToSpherePotentialPair<numnodes, numnodalvalues>::evalu
     if (chargeconds[0]->type() == Core::Conditions::BeamPotential_LineChargeDensity)
       chargeconds_[0] = chargeconds[0];
     else
-      FOUR_C_THROW("Provided condition is not of correct type BeamPotential_LineChargeDensity!");
+      FOUR_C_THROW(
+          "Provided condition is not of correct type "
+          "BeamInteraction::Potential::_LineChargeDensity!");
```
sebproell: Something is off here.

davidrudlstorfer: Good catch - this was a wrong search and replace with the Namespace change

davidrudlstorfer:
> @davidrudlstorfer Yes, shared_ptr (or Teuchos::RCP) is almost never what you want. If you can state to which object's lifetime these parameters are tied, then let that object own the paramters and just pass a reference/pointer to whichever objects need to use it. Of course, these objects may not live longer than the owner object. If you do not know enough about lifetime of your objects, you likely have more fundamental problems, but shared_ptr still only patches over the problem.

@sebproell thanks for your explanation :) I've simplified everything in the second commit so the model evaluator now owns the data and the individual beam pairs only have a regular pointer to that data. Now this makes sense in my opinion

src/beaminteraction/src/potential/4C_beaminteraction_potential_submodel_evaluator.cpp
```diff
@@ -869,9 +868,10 @@ void BeamInteraction::SubmodelEvaluator::BeamPotential::create_beam_potential_el
       ele_ptrs[1] = *secondeleiter;
 
       std::shared_ptr<BeamInteraction::BeamPotentialPair> newbeaminteractionpair =
-          BeamInteraction::BeamPotentialPair::create(ele_ptrs, beam_potential_params());
+          BeamInteraction::BeamPotentialPair::create(ele_ptrs, beam_potential_parameters());
 
-      newbeaminteractionpair->init(beam_potential_params_ptr(), ele_ptrs[0], ele_ptrs[1]);
+      // change this pointer to just use the data in a first place
```
sebproell: I don't get this comment. What is "this pointer"? Is this a ToDo comment?

davidrudlstorfer: Thanks for the catch again - this was a comment which should have been deleted

sebproell (approved):

tests/input_files/beam3r_herm2line3_backweuler_browndyn_repLJ_singlelengthspec_smallsepapprox_twocrossedbeams.4C.yaml
```diff
@@ -44,18 +44,21 @@ BROWNIAN DYNAMICS:
...
-  TYPE: volume
-  STRATEGY: single_length_specific_small_separations
-  N_INTEGRATION_SEGMENTS: 2
-  N_GAUSS_POINTS: 50
-  AUTOMATIC_DIFFERENTIATION: true
-BEAM POTENTIAL/RUNTIME VTK OUTPUT:
-  VTK_OUTPUT_BEAM_POTENTIAL: true
-  FORCES: true
-  MOMENTS: true
+beam_potential:
```
isteinbrecher, line 47: Is there a reason you chose lower case names for the parameters here?

isteinbrecher (approved):

tests/input_files/beam3r_herm2line3_backweuler_browndyn_repLJ_singlelengthspec_smallsepapprox_twocrossedbeams.4C.yaml
davidrudlstorfer: This change was based on the improved readability (in my opinion at least) of snake_case. Because the input changed drastically for the beam_potential within this PR I took the chance and also moved to lower case here.

This was also briefly proposed in https://github.com/4C-multiphysics/4C/discussions/930

And also the input was changed to snake_case for a lot of poro stuff https://github.com/4C-multiphysics/4C/pull/929

PR #1031, opened 2025-08-01
[solver] Recompute preconditioner due to linear solver stall

maxfirmbach:
## Description and Context

If a preconditioner is reused over several nonlinear iterations the linear solver might begin to stall, due to a change in nonlinearity / physical behavior. Stalling in this case means a rapid increase in linear iterations, which can be reduced by triggering a recomputation of the preconditioner.

This PR introduces a new input parameter, which can be used to define an upper bound for linear iterations. If the number of linear iterations go over this value, we assume it to be stalling, thus triggering a recomputation of the preconditioner.

Example of the adapted test case:

```terminal
*******************************************************
***** Belos Iterative Solver:  Block Gmres 
***** Maximum Iterations: 1000
***** Block Size: 1
***** Residual Test: 
*****   Test 1 : Belos::StatusTestImpResNorm<>: (2-Norm Res Vec) / (2-Norm Res0), tol = 2.64275e-06
*******************************************************
Iter    0, [ 1] :    1.000000e+00
Iter    5, [ 1] :    4.054261e-06
Iter    6, [ 1] :    4.780254e-07
-- Status Test Results --
**...........OR Combination -> 
  **...........AND Combination -> 
    **...........AND Combination -> 
      **...........Structure-Update-Norm = 2.861e-05 < 1e-10
                   (Unscaled 2-Norm, Absolute)
                   (Min Step Size:  1.000e+00 >= 1)
                   (Max Lin Solv Tol:  0.000e+00 < 0.5)
      Converged....Structure-F-Norm = 9.879e-15 < 1.000e-08
                   (Unscaled 2-Norm, Absolute)
  **...........Number of Iterations = 4 < 50

-- Nonlinear Solver Step 4 -- 
||F|| = 9.87920e-15  step = 1.00000e+00  dx = 2.86051e-05

*******************************************************
Recomputation due to linear solver stalling.
Compute preconditioner 
*******************************************************
```

This example is of course super artificial, but shows the principle.

### DIFF
--- a/src/core/linear_solver/src/method/4C_linear_solver_method_input.cpp
+++ b/src/core/linear_solver/src/method/4C_linear_solver_method_input.cpp
@@ -91,8 +91,15 @@ namespace Core::LinearSolver
                             .default_value = 0}),
 
         parameter<int>("AZREUSE",
-            {.description = "The number specifying how often to recompute some preconditioners",
+            {.description = "Update preconditioner after this many nonlinear iterations. The "
+                            "preconditioner is recomputed at every start of a nonlinear solve.",
                 .default_value = 0}),
+
+        parameter<int>("REUSE_STALL_ITER",
+            {.description =
+                    "Maximum number of linear iterations that triggers a nonlinear iteration to "
+                    "be declared stalled and thus force recomputation of the preconditioner.",
+                .default_value = 50}),
 
         parameter<int>(
             "AZSUB", {.description = "The maximum size of the Krylov subspace used with \"GMRES\" "
--- a/src/core/linear_solver/src/method/4C_linear_solver_method_iterative.cpp
+++ b/src/core/linear_solver/src/method/4C_linear_solver_method_iterative.cpp
@@ -22,7 +22,6 @@
 #include <BelosLinearProblem.hpp>
 #include <BelosPseudoBlockCGSolMgr.hpp>
 #include <BelosPseudoBlockGmresSolMgr.hpp>
-#include <Epetra_CrsMatrix.h>
 #include <Teuchos_RCPStdSharedPtrConversions.hpp>
 #include <Teuchos_TimeMonitor.hpp>
 #include <Teuchos_XMLParameterListHelpers.hpp>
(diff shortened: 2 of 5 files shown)


src/core/linear_solver/src/method/4C_linear_solver_method_input.cpp
```diff
@@ -91,9 +91,16 @@ namespace Core::LinearSolver
...
 
         parameter<int>("AZREUSE",
-            {.description = "The number specifying how often to recompute some preconditioners",
+            {.description = "Update preconditioner after this many nonlinear iterations. The "
+                            "preconditioner is recomputed at every start of a nonlinear solve.",
                 .default_value = 0}),
 
+        parameter<int>("REUSE_STALL_ITER",
+            {.description =
+                    " Maximum number of linear iterations that triggers a nonlinear iteration to "
```
mayrmt:
```suggestion
                    "Maximum number of linear iterations that triggers a nonlinear iteration to "
```

src/core/linear_solver/src/method/4C_linear_solver_method_input.cpp
```diff
@@ -91,9 +91,16 @@ namespace Core::LinearSolver
                             .default_value = 0}),
 
         parameter<int>("AZREUSE",
-            {.description = "The number specifying how often to recompute some preconditioners",
+            {.description = "Update preconditioner after this many nonlinear iterations. The "
+                            "preconditioner is recomputed at every start of a nonlinear solve.",
```
mayrmt, line 95: I don't understand the second sentence in the description. 

Do you mean "Independent of this parameter, the preconditioner is always recomputed at the beginning of a time/load step."?

maxfirmbach: Example: AZREUSE is set to 10, but the nonlinear solve is done after 6. The next nonlinear solve will start with a recomputation of the preconditioner, despite the 4 steps left as prescribed by the parameter.

mayrmt: That's the right approach in my opinion. I just was not sure about the exact wording.

PR #1035, opened 2025-08-02
Clarify normal distribution parameters

sebproell:
Found by #1033, Depends on #1030

Assert that stddev > 0 and change the naming and docs.

FYI @ppraegla @davidrudlstorfer

### DIFF
--- a/src/particle_interaction/4C_particle_interaction_dem_adhesion_surface_energy.cpp
+++ b/src/particle_interaction/4C_particle_interaction_dem_adhesion_surface_energy.cpp
@@ -86,7 +86,7 @@ void ParticleInteraction::DEMAdhesionSurfaceEnergyDistributionNormal::adhesion_s
     const double& mean_surface_energy, double& surface_energy) const
 {
   // initialize random number generator
-  Global::Problem::instance()->random()->set_mean_variance(mean_surface_energy, variance_);
+  Global::Problem::instance()->random()->set_mean_stddev(mean_surface_energy, variance_);
 
   // set normal distributed random value for surface energy
   surface_energy = Global::Problem::instance()->random()->normal();
@@ -106,8 +106,7 @@ void ParticleInteraction::DEMAdhesionSurfaceEnergyDistributionLogNormal::adhesio
     const double& mean_surface_energy, double& surface_energy) const
 {
   // initialize random number generator
-  Global::Problem::instance()->random()->set_mean_variance(
-      std::log(mean_surface_energy), variance_);
+  Global::Problem::instance()->random()->set_mean_stddev(std::log(mean_surface_energy), variance_);
 
   // set log-normal distributed random value for surface energy
   surface_energy = std::exp(Global::Problem::instance()->random()->normal());
--- a/tests/input_files/beam3r_rigidsphere_activecrosslinking_truss_browndyn_truss_catchslipbond.4C.yaml
+++ b/tests/input_files/beam3r_rigidsphere_activecrosslinking_truss_browndyn_truss_catchslipbond.4C.yaml
@@ -49,6 +49,10 @@ BROWNIAN DYNAMICS:
   BROWNDYNPROB: true
   VISCOSITY: 0.001
   TIMESTEP: 0.01
+  # Setting a very small KT ensures that random forces are sampled.
+  # This became necessary when an internal change would have lead to non-convergence due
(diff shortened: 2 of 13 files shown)


src/particle_interaction/4C_particle_interaction_dem_adhesion_surface_energy.cpp
```diff
@@ -86,7 +86,7 @@ void ParticleInteraction::DEMAdhesionSurfaceEnergyDistributionNormal::adhesion_s
     const double& mean_surface_energy, double& surface_energy) const
 {
   // initialize random number generator
-  Global::Problem::instance()->random()->set_mean_variance(mean_surface_energy, variance_);
+  Global::Problem::instance()->random()->set_mean_stddev(mean_surface_energy, variance_);
```
sebproell, line 89: @ppraegla This looks suspicious. The function expected a standard deviation (hence the rename), but we are passing a variance here. The tests are passing, since this has been wrong (?) before.

sebproell:
@davidrudlstorfer @isteinbrecher The tests are not passing since `beaminteraction` seems to sometimes generate distributions with zero standard deviation (which is not allowed by C++). The problem is that `kT` is defaulted to zero, leading to a zero standard deviation. If `kT=0` makes physical sense, we can add the degenerate case of a point-wise distribution to our `Random` class.

davidrudlstorfer:
> @davidrudlstorfer @isteinbrecher The tests are not passing since beaminteraction seems to sometimes generate distributions with zero standard deviation (which is not allowed by C++). The problem is that kT is defaulted to zero, leading to a zero standard deviation. If kT=0 makes physical sense, we can add the degenerate case of a point-wise distribution to our Random class.

Thanks for looking into this. TLDR; `kT = 0` makes sense

Long answer: `kT` is the Boltzmann constant multiplied with the Temperature. The main case of the Brownian Dynamics framework is to get the "random" background noise in molecular simulations - and these depend on the Temperature of the surroundings. In addition to that feature, the Brownian Dynamics framework also supports the use of a _quiescent background_ fluid.  I, for example, use this to enable the simulation of my fibers in a fluid which in turn adds dampening to my simulations. Because I have no interest in the random force fluctuations I use `kT = 0` to just use the other utils of the framework. There would definitely be the option to just enable and disable these fluctuations without getting the probability function but I guess this would result in a larger overhaul of the framework

tests/input_files/beam3r_rigidsphere_activecrosslinking_truss_browndyn_truss_catchslipbond.4C.yaml
```diff
@@ -49,6 +49,7 @@ BROWNIAN DYNAMICS:
   BROWNDYNPROB: true
   VISCOSITY: 0.001
   TIMESTEP: 0.01
+  KT: 1e-14
```
sebproell, line 55: This is a bit of a hack: I choose a very small `kT` to sample the same number of random numbers as before. Otherwise, the nonlinear solver does not converge due to the sampled random crosslinking forces, AFAICT. Not pretty, so if anyone is knowledgeable here, feel free to suggest something better (or make the tests robust).

davidrudlstorfer: As discussed, currently I have no clue regarding the cross linking framework. If I need it in the future I have to look a little bit closer anyways.

An explaining comment on why you chose 1e-14 is fine for me

src/particle_interaction/4C_particle_interaction_dem_adhesion_surface_energy.cpp
ppraegla: Thanks for uncovering this bug. This seems to be wrong. This should be the standard deviation. Also, further down, we adjust the surface energy to some calculated bounds (`mean_surface_energy - cutofffactor_ * variance_`). I think something like `3*sigma` instead of `3*sigma^2` was meant here.

It does not matter for the test. And as far as I know, we have not used this "feature" so far.
I will open a new PR to remove the lognormal distribution for the surface energy `DEMAdhesionSurfaceEnergyDistributionLogNormal`. It is unused and the way we adjust the bounds does not make sense for a lognormal distribution. I can then also take care of renaming `variance_`.

PR #1057, opened 2025-08-05
Fix small bug in solid-poro Brinkmann flow off-diagonal block

amgebauer:
To me this looks like a bug.


c-p-schmidt (approved):
Or at least the naming is misleading. ;-)

lkoeglmeier (approved):

Issue #863, opened 2025-06-04
[LinAlg] Refactor sparse operator, matrix and block-matrix wrapper


maxfirmbach:
## Description

Currently our `LinAlg::SparseOperator` inherits from `Epetra_Operator`, which brings a few difficulties with it. I would propose to remove this inheritance and provide just an abstract `LinAlg::SparseOperator` interface with necessary methods we need. The actual implementations, namely `LinAlg::SparseMatrix` and `LinAlg::BlockSparseMatrix` should internally hold the respective `Epetra` related object (similar to our other wrappers).

## Possible Solution and Definition of Done

- [x] Remove the inheritance to `Epetra_Operator` of `LinAlg::SparseOperator` (see #1524).
- [ ] Remove the inhertiance to to `Epetra_Operator` of `LinAlg::BlockSparseMatrix`. The only (??) thing that prevents this is the `epetra_operator()` call in the iterative solver that needs the inheritance to be able to overwrite `Apply()` and `ApplyInverse()`. With #1400 resolved, we should be able to cut the inheritance.
- [ ] Move member functions related to matrix addition to free functions (see #136)

Refactor `LinAlg::SparseMatrix` wrapper 
- [x] Think about how the wrapper could hold an `Epetra_CrsMatrix` as private member.
- [x] Remove as many `Epetra_CrsMatrix` occurences in the code and replace by `Core::LinAlg::SparseMatrix`
- [x] One more instance of `get_epetra_export` call remains: need our own `Import` implementation for `Epetra_CrsMatrix`.  (see #1107)
- [x] Add a conversion utility function, which returns a `Thyra::LinearOp` wrapping the `Epetra_CrsMatrix` 
- [ ] Turn on `StaticProfile` to `true` (see #1544)

Refactor `LinAlg::BlockSparseMatrix` implementation (see #1525)
- [x] Remove the method `Core::LinAlg::block_matrix2x2()`, which is only used once in `xfluid`
- [x] Remove the method `Core::LinAlg::BlockSparseMatrixBase::get_partial_extractor()`, which is not used
- [x] Add a conversion utility function, which fills a `Thyra::PhysicallyBlockedLinearOp` and returns a `Thyra::LinearOp`

## Interested Parties

@sebproell @knarfnitram @dharinib98

PR #1062, opened 2025-08-06
Remove some raw Epetra calls

dharinib98:
## Description and Context

This PR cleans up all the remaining `get_epetra_block_map` calls and some more `get_epetra_export` calls. 

#863

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_fevector.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_fevector.hpp
@@ -233,6 +233,12 @@ namespace Core::LinAlg
       return vector_->Export(A, Importer.get_epetra_import(), CombineMode, Indexor);
     }
 
+    int export_to(const Epetra_SrcDistObject& A, const Core::LinAlg::Export& Exporter,
+        Epetra_CombineMode CombineMode, const Epetra_OffsetIndex* Indexor = nullptr)
+    {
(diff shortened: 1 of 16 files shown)


src/core/linalg/src/sparse/4C_linalg_fevector.hpp
```diff
@@ -233,6 +233,12 @@ namespace Core::LinAlg
       return vector_->Export(A, Importer.get_epetra_import(), CombineMode, Indexor);
     }
 
+    int export_to(const Epetra_SrcDistObject& A, const Core::LinAlg::Export& Exporter,
+        Epetra_CombineMode CombineMode, const Epetra_OffsetIndex* Indexor = nullptr)
```
mayrmt, line 237: I don't understand, why the name `Indexor` is chosen (here and at other places in this file).

mayrmt (approved):
@dharinib98, thanks for looking into this!

src/core/linalg/src/sparse/4C_linalg_fevector.hpp
maxfirmbach: @mayrmt This is just a copy of the `Epetra` naming scheme. We decided to work on the naming at some other point, mainly when all functionality is wrapped.

maxfirmbach:
@dharinib98 Thanks for looking into this 👍.

PR #1063, opened 2025-08-07
Remove redundant stress and strain output from new structure integration

c-p-schmidt:
## Description and Context
As part of the unification of the output, this pull request removes the output of stress and strain for the new structure algorithm and replaces the functionality with respective vtk-based output methods. The main steps are:
    - Remove custom binary stress-strain output from the new structure
    - Remove post-processing test and respective reference file
    - Enable vtk-based output of coupling stresses
    - Add a vtk-test to test this feature

part of #211


c-p-schmidt:
I'm aware that the added reference files are not very short; however, I already tried to reduce the amount of added reference lines by only comparing against the last time step.  
This was the only test that used the coupling stress output, so it was the only file available to enable this check with reasonable effort.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
```diff
@@ -978,129 +1008,144 @@ void Solid::ModelEvaluator::Structure::output_runtime_structure_postprocess_stre
...
-    // Set dummy evaluation vectors and matrices.
-    std::array<std::shared_ptr<Core::LinAlg::Vector<double>>, 3> eval_vec = {
-        nullptr, nullptr, nullptr};
-    std::array<std::shared_ptr<Core::LinAlg::SparseOperator>, 2> eval_mat = {nullptr, nullptr};
+  evaluate_internal_specified_elements(
+      eval_mat.data(), eval_vec.data(), discret().element_row_map());
 
-    evaluate_internal_specified_elements(
-        eval_mat.data(), eval_vec.data(), discret().element_row_map());
+  auto DoPostprocessingOnElement = [](const Core::Elements::Element& ele)
```
sebproell: Why is this not `snake_case`? Same for all other function objects.

sebproell (commented):
I really dislike the `couplstress` approach. You need to know (and remember) that the `stress` quantity in the output is not the stress unless you add `couplstress`. Ofc, you don't need to fix this here, but I think we should change this if we work on the output anyway.

amgebauer:
> I really dislike the `couplstress` approach. You need to know (and remember) that the `stress` quantity in the output is not the stress unless you add `couplstress`. Ofc, you don't need to fix this here, but I think we should change this if we work on the output anyway.

Is there any usecase for the couplstress? I also think that it does not fit well into the structure model evaluator data since it is a quantity that only occurs in coupled problems. We might just get rid of it completely?

c-p-schmidt:
> > I really dislike the `couplstress` approach. You need to know (and remember) that the `stress` quantity in the output is not the stress unless you add `couplstress`. Ofc, you don't need to fix this here, but I think we should change this if we work on the output anyway.
> 
> Is there any usecase for the couplstress? I also think that it does not fit well into the structure model evaluator data since it is a quantity that only occurs in coupled problems. We might just get rid of it completely?

As I have written, there was a test case for that, so I migrated. I can't judge whether that is relevant or not, as it is in a different kind of module.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
c-p-schmidt: I just used what was already there in the file. But as a rename is easy, I renamed all those lambdas right away.

amgebauer:
In TSI, we removed it during the migration from old to new solid elements.  I think it's easier to coordinate with the few poro people whether this quantity is really needed. Of course, we would need to add the coupl-stress to the "normal" stress output so that this is the consistent stress response of the solid-poro material.

lkoeglmeier:
Maybe to briefly summarise what we discussed today: the `coupling_stress` is only the fluid contribution to the total stress tensor ( $-p^s \boldsymbol{I}$ for cauchy stress and the pull back for the 2nd pKS $-  p^s J \boldsymbol{F}^{-1} \boldsymbol{F}^{-T} $). It is only implemented in the porofluid pressure and velocity based formulation. 
Personally, I do not use this formulation and do not have a strong opinion on that. As it calculates nothing wrong, I would also be fine with keeping it. Do the other (poro)people have an opinion on that? @bwirthl @shervas @knarfnitram

tests/input_files/poro_3D_hex8_new_struc.4C.yaml
```diff
@@ -23,6 +23,13 @@ IO:
   STRUCT_STRESS: "Cauchy"
   STRUCT_COUPLING_STRESS: "Cauchy"
   STRUCT_STRAIN: "GL"
```
isteinbrecher: I think at some point we should move these options to the VTK section. But maybe that should be done once the post-processor is removed.

isteinbrecher (approved):

bwirthl:
> Maybe to briefly summarise what we discussed today: the `coupling_stress` is only the fluid contribution to the total stress tensor. It is only implemented in the porofluid pressure and velocity based formulation. Personally, I do not use this formulation and do not have a strong opinion on that. As it calculates nothing wrong, I would also be fine with keeping it. Do the other (poro)people have an opinion on that? @bwirthl @shervas @knarfnitram

Neither do I use it. Hence, I also do not have a strong opinion on that. Following your discussion above, I would vote for deleting it.

lkoeglmeier (approved):

c-p-schmidt:
Based on the feedback obtained I removed the coupling stress. This is done in the second commit in this PR.

amgebauer (approved):

lkoeglmeier (approved):

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_pressure_based.cpp
```diff
@@ -252,19 +252,6 @@ void Discret::Elements::SolidPoroPressureBasedEleCalc<celltype>::coupling_stress
     const Core::Elements::Element& ele, const Core::FE::Discretization& discretization,
```
lkoeglmeier: Can we also delete the entire function here?

c-p-schmidt: Yes, this must be possible. Good catch!

c-p-schmidt:
I removed the redundant method as pointed out by @lkoeglmeier. Furthermore, I saw that some post-processing methods and one file were still there, which I also removed.

PR #1066, opened 2025-08-07
Generalize norms evaluation for NOX vectors and reduce the use of `get_ref_of_epetra_vector()`

vovannikov:
## Description and Context
I have removed some obsolete creation of NOX vector views when computing the norms since our `LinAlg::Vector` can compute them itself. I also removed some other unnecessary calls to `get_ref_of_epetra_vector()`.

#359

### DIFF
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_aux.cpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_aux.cpp
@@ -571,6 +571,33 @@ std::string NOX::Nln::Aux::get_direction_method_list_name(const Teuchos::Paramet
   }
 }
 
+template <typename T>
+T NOX::Nln::Aux::calc_vector_norm(const Core::LinAlg::Vector<T>& vector,
+    const ::NOX::Abstract::Vector::NormType type, const bool is_scaled)
+{
+  T value;
+
+  switch (type)
+  {
+    case ::NOX::Abstract::Vector::OneNorm:
+      vector.norm_1(&value);
+      break;
+    case ::NOX::Abstract::Vector::TwoNorm:
+      vector.norm_2(&value);
+      break;
+    case ::NOX::Abstract::Vector::MaxNorm:
+      vector.norm_inf(&value);
+      break;
+    default:
+      FOUR_C_THROW("The norm type is not supported");
+  }
+
+  // do the scaling if desired
+  if (is_scaled) value /= static_cast<T>(vector.global_length());
+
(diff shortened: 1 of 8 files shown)


src/solver_nonlin_nox/4C_solver_nonlin_nox_aux.cpp
```diff
@@ -571,6 +571,33 @@ std::string NOX::Nln::Aux::get_direction_method_list_name(const Teuchos::Paramet
...
+      break;
+    case ::NOX::Abstract::Vector::MaxNorm:
+      vector.norm_inf(&value);
+      break;
+    default:
+      FOUR_C_THROW("The norm type is not supported");
+  }
+
+  // do the scaling if desired
+  if (is_scaled) value /= static_cast<T>(vector.global_length());
```
mayrmt, line 596: NOX distinguishes for the scaling between two-norm and the other norms. See https://github.com/trilinos/Trilinos/blob/ae38582c14290406d3829298783238bc7d547929/packages/nox/src/NOX_StatusTest_NormF.C#L125-151

vovannikov: As you can see, the code I am replacing evaluates norm via `::NOX::Epetra::Vector::notm()`, that delegates it itself, effectively, to the following function: https://github.com/trilinos/Trilinos/blob/ae38582c14290406d3829298783238bc7d547929/packages/nox/src-epetra/NOX_Epetra_VectorSpace_L2.C#L31-L47. It is somewhat very similar to what I have implemented via a very similar switch. Of course, NOX Epetra also offers the so-called scaled vector space, for which the norms computation differs, but we do not use them.

So, NOX Epetra does not compute norm itself but delegates it to the underlying Epetra vector, similarly like done in this PR.

vovannikov: Yes, indeed, the scaling is different in NOX. But here I just refactored the existing code and transferred the repetitive code into a single function: on the 4C side this kind of scaling repeats itself in many places. Probably, we should reconsider scaling too, but I would do that in a different PR, since the point of the current one was mainly to remove the uses of `get_ref_of_epetra_vector()` where they are not needed.

mayrmt: Sure! Let's look at this separately.

mayrmt (approved):
LGTM.

Issue #723, opened 2025-04-30
[Solver] Use SparseOperator instead of Epetra_Operator in linear solver


maxfirmbach:
## Description

We should refactor the `setup()` and `solve()` calls from `Core::LinAlg::Solver` to use a `Core::LinAlg::SparseOperator` instead of a `Epetra_Operator`. The conversion to the later one can happen inside the linear solver itself. This hides away the explicit `Epetra` based call `epetra_operator()`.

## Possible Solution and Definition of Done

- [x]  Replace `Epetra_Operator` with `Core::LinAlg::SparseOperator` in the `setup()` and `solve()` methods from `Core::LinAlg::Solver` and fix internals (see #1155).
- [x] Refactor the code base to use the modified methods correctly.
  - Problematic case is in `4C_solver_nonlinear_nox_linearsystem.cpp` as we use the `Epetra_LinearProblem`, scale it and return the `Epetra_Operator` ... very unfortunate

maxfirmbach:
This is ready to be worked on.

maxfirmbach:
This issue is done, thanks again @vovannikov for helping 👍.

PR #1071, opened 2025-08-11
Reduce the use of `epetra_operator()`

vovannikov:
## Description and Context
I modified the interface of `Core::LinAlg::Solver` such that it works with `Core::LinAlg::SparseOperator` rather than with `Epetra_Operator`. This then reduces the fallback calls to `SparseOperator::epetra_operator()` all over the code. In my opinion, this is again an intermediate solution towards removing `Epetra`, since currently the fallbacks to `Epetra_Operator` are not completely hidden everywhere, see the modifications of some more complex block cases.

#359, #379


mayrmt:
Also related to https://github.com/4C-multiphysics/4C/issues/723

src/core/linear_solver/src/amgnxn/4C_linear_solver_amgnxn_smoothers.hpp
```diff
@@ -297,6 +297,7 @@ namespace Core::LinearSolver::AMGNxN
    private:
     std::shared_ptr<Core::LinAlg::Solver> solver_;
     std::shared_ptr<Epetra_Operator> a_;
```
mayrmt: Is `a_` still required? What needs to be done to get rid of it entirely?

vovannikov: Indeed, this pointer is no longer needed as a member. I have removed it.

vovannikov:
It seems some preliminary refactoring regarding `Epetra_LinearProblem` is better to be performed prior to this PR. I will prepare a new one (maybe a few even), an then this one can be addressed.

maxfirmbach:
Thanks for removing the `epetra_operator()` calls! This was overdue :+1:.

vovannikov:
This is now ready to be reviewed.

PR #1072, opened 2025-08-11
[cmake] Cleanup python dependency

sebproell:


cmake/functions/four_c_testing_functions.cmake
```diff
@@ -58,6 +58,15 @@ function(set_timeout name_of_test)
   endif()
 endfunction()
 
+# Mark a test as skipped. This register a test printing an informational message.
```
mayrmt:
```suggestion
# Mark a test as skipped. This registers a test to must print an informational message instead of actually executing the test.
```

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -58,6 +58,15 @@ function(set_timeout name_of_test)
...
+# Running ctest will display that a test was skipped, to make it clear that this test exists
+# but is not executed. The reason may be explained in the message.
+function(skip_test name_of_test message)
+  set(dummy_command "echo \"${message}\"; exit 42")
+  add_test(NAME ${name_of_test} COMMAND bash -c "${dummy_command}")
+  set_tests_properties(${name_of_test} PROPERTIES SKIP_RETURN_CODE 42)
+endfunction()
```
mayrmt: 42 might be an arbitrary number here, and I probably can guess its origin. Yet, is it clear to everyone that it does not convey any deeper meaning (other than being non-zero)?

mayrmt (dismissed):
@sebproell, I think this nicely exposes and cleans-up our dependency on Python. I see value in being able to configure without Python support, so it is good that we now have a clear strategy and useful error messages for this case.

i left some comments on comments... just as an idea.

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -58,6 +58,16 @@ function(set_timeout name_of_test)
   endif()
 endfunction()
 
+# Mark a test as skipped. This registers a dummy test printing an informational message.
+# Running ctest will display that test as skipped, to make it clear that this test exists
+# but is not executed. The reason may be explained in the message.
+function(skip_test name_of_test message)
+  # The dummy test needs to report a arbitrary error code that ctest interprets as "skipped".
```
mayrmt, line 65:
```suggestion
  # The dummy test needs to report an arbitrary error code that ctest interprets as "skipped".
```

mayrmt (approved):

amgebauer (approved):

cmake/configure/configure_Python.cmake
```diff
@@ -49,20 +49,28 @@ endfunction()
...
-if(Python3_FOUND)
-  message(STATUS "Using python executable: ${Python3_EXECUTABLE} (V${Python3_VERSION})")
+find_package(Python REQUIRED COMPONENTS Interpreter)
+if(Python_FOUND)
+  message(STATUS "Using python executable: ${Python_EXECUTABLE} (V${Python_VERSION})")
 endif()
 
-if(Python3_VERSION VERSION_LESS "3.8")
-  message(FATAL_ERROR "Python version must be at least 3.8, but found ${Python3_VERSION}")
+if(Python_VERSION VERSION_LESS "3.10")
```
sebproell, line 57: Sorry, I forgot to mention that I increased the version to simplify some checks. 3.9 will become end-of-life in October, so this needed to happen soon.

PR #1074, opened 2025-08-11
Allow documentation of enum values in input

sebproell:
Requested by and implemented together with @ischeider 

FYI @c-p-schmidt, you also requested this feature a while ago

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validparameters.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validparameters.cpp
@@ -154,9 +154,12 @@ std::map<std::string, Core::IO::InputSpec> Global::valid_parameters()
                 "FILE", {.description = "Path to the exodus geometry file. Either absolute or "
                                         "relative to the input file."}),
             parameter<Core::IO::Exodus::VerbosityLevel>("SHOW_INFO",
-                {.description =
-                        "Print element, node and set info for the exodus file after reading.",
-                    .default_value = Core::IO::Exodus::VerbosityLevel::none}),
+                {
+                    .description = "Choose verbosity of reporting element, node and set info for "
+                                   "the exodus file after reading.",
+                    .enum_value_description = Core::IO::Exodus::describe,
+                    .default_value = Core::IO::Exodus::VerbosityLevel::none,
+                }),
 
             // Once we support more formats, we should add a "TYPE" parameter for the file format.
             list("ELEMENT_BLOCKS",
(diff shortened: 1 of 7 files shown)


amgebauer (commented):
Can you also adapt the schema file generation?

sebproell:
> Can you also adapt the schema file generation?

Yes, but I need to see if we can directly provide docs for the values or we need to generate a description on the parameter.

src/global_legacy_module/4C_global_legacy_module_validparameters.cpp
```diff
@@ -154,9 +154,12 @@ std::map<std::string, Core::IO::InputSpec> Global::valid_parameters()
                 "FILE", {.description = "Path to the exodus geometry file. Either absolute or "
                                         "relative to the input file."}),
             parameter<Core::IO::Exodus::VerbosityLevel>("SHOW_INFO",
-                {.description =
+                {
+                    .description =
                         "Print element, node and set info for the exodus file after reading.",
```
mayrmt: This comment should also indicate, that one has the option to select a verbosity level.

sebproell:
> > Can you also adapt the schema file generation?
> 
> Yes, but I need to see if we can directly provide docs for the values or we need to generate a description on the parameter.

We can either

- generate an adapted description where the documentation of every enum value is added to the general description of the parameter
- treat every enum value as a `const` json schema and add description there.

Ping @gilrrei

gilrrei:
@sebproell @amgebauer 

>     * generate an adapted description where the documentation of every enum value is added to the general description of the parameter

Easy and quick, we just add the description of the options to the description of the enum itself.
 
>     * treat every enum value as a `const` json schema and add description there.

Should lead to better autocomplete, we need to transform the enum into an object, where the options are `const` as you mentioned. It is a bit more complex, but nothing major.

I think the latter is more complete and consistent with the `selection`.  However, I think for now, let's do it the easy way.

sebproell:
@gilrrei I implemented the version with `oneOf` and `const` since it wasn't too difficult. The problem is that depending on the editor, I do not get what I would like to see. With CLion (not super relevant), I always see the general description of the parameter, regardless of the enum value. With VSCode (more relevant), I always see the description of the enum value I chose, and I no longer see the general description. Due to the inconsistency, I think we are better off if we just create a unified description for the purpose of schemas. Ofc, it is still useful to have the info in metadata separately for other tooling.

sebproell:
I now implemented the simple option. Ready for another review.

mayrmt (approved):

PR #1075, opened 2025-08-11
Add pybind11 as optional dependency

sebproell:
Provides [pybind11](https://github.com/pybind/pybind11) as an optional dependency

Requested by @mayrmt, who will contribute some actual code using this dependency


amgebauer (approved):

sebproell:
This PR reminds me that I need to tackle #618. It nicely points out the duplication of information that could be generated.

PR #1077, opened 2025-08-11
Replace `NOX::Epetra::Scaling` with a custom class

vovannikov:
## Description and Context
I do not like the last commit in #1071 (https://github.com/4C-multiphysics/4C/pull/1071/commits/9ba0d5afb205a418680e6dedbc39cc175f9c11db) as it introduces not a very elegant solution. It gets worse for the cases when scaling is actually used. A better solution would be to replace `Epetra_LinearProblem` with our own class. To do that, firstly, we need to replace `Epetra::Scaling` with our own implementation that provides a similar interface (though, reduced). That is what this PR doing.

#1071, #723

### DIFF
--- a/src/fsi/src/monolithic/nonlinear_solver/4C_fsi_nox_linearsystem_gcr.cpp
+++ b/src/fsi/src/monolithic/nonlinear_solver/4C_fsi_nox_linearsystem_gcr.cpp
@@ -29,7 +29,7 @@ NOX::FSI::LinearSystemGCR::LinearSystemGCR(Teuchos::ParameterList& printParams,
     const Teuchos::RCP<::NOX::Epetra::Interface::Required>& iReq,
     const Teuchos::RCP<::NOX::Epetra::Interface::Jacobian>& iJac,
     const Teuchos::RCP<Epetra_Operator>& jacobian, const ::NOX::Epetra::Vector& cloneVector,
-    const Teuchos::RCP<::NOX::Epetra::Scaling> s)
+    const std::shared_ptr<NOX::Nln::Scaling> s)
     : utils(printParams),
       jacInterfacePtr(iJac),
       jacType(EpetraOperator),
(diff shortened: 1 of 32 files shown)


src/fsi/src/monolithic/nonlinear_solver/4C_fsi_nox_linearsystem_gcr.cpp
```diff
@@ -136,7 +131,7 @@ bool NOX::FSI::LinearSystemGCR::applyJacobianInverse(
   }
 
   // Unscale the linear system
-  if (!!(scaling)) scaling->unscaleLinearSystem(Problem);
+  if (!!(scaling)) scaling->unscale_linear_system(Problem);
```
sebproell:
```suggestion
  if (scaling) scaling->unscale_linear_system(Problem);
```

src/solver_nonlin_nox/4C_solver_nonlin_nox_scaling.hpp
```diff
@@ -0,0 +1,48 @@
...
+  {
+   public:
+    //! Virtual destructor
+    virtual ~Scaling() = default;
+
+    /**
+     * @brief Compute the scaling for the linear system.
+     *
+     * This method is mainly to provide compatibility with the current pattern how scaling applied:
+     * is flag manualScaling is set to true, then this method is called.
```
sebproell:
```suggestion
     * This method is mainly to provide compatibility with the current pattern how scaling is applied:
     * if flag manualScaling is set to true, then this method is called.
```

sebproell (approved):

src/fsi/src/monolithic/nonlinear_solver/4C_fsi_nox_linearsystem_gcr.cpp
vovannikov: Yep, there were a couple of other places like that, I fixed them too,

sebproell:
@vovannikov A general remark: we usually like to squash minor fixes from review into the appropriate commits. We are not interested in the exact steps it took to achieve a change, but rather want to have commits that perform a self-contained transformation of the code.

vovannikov:
> @vovannikov A general remark: we usually like to squash minor fixes from review into the appropriate commits. We are not interested in the exact steps it took to achieve a change, but rather want to have commits that perform a self-contained transformation of the code.

Sure, no problem on this. Do you mean that it is preferable to squash _all_ the commits of a PR into one?

sebproell:
> Sure, no problem on this. Do you mean that it is preferable to squash all the commits of a PR into one?

It depends: often, yes, one commit seems fine. If you do some cleanup work that emerged from the actual change, feel free to do this as a separate commit (or even PR), to make it easy to grasp the relevant changes in review and have a useful log and blame.

PR #1081, opened 2025-08-12
Support real YAML as element input for EXODUS meshes

sebproell:
Previously, one needed to specify for an EXODUS mesh:

```yaml
      ELEMENT_NAME: SOLID
```

Now this becomes

```yaml
      ELEMENT_DATA:
        SOLID:
          HEX8:
            MAT: 1
            KINEM: nonlinear
```

The reason why `HEX8` is required is related to some elements that have different options for different cell types (e.g., EAS only works for certain elements). 

**NOTE**: this does _not_ affect the legacy string input of elements, only the direct mesh input (#653).

#


sebproell:
@danielwolff1 you expressed interest in that.

sebproell:
I will need to adapt more tests, but I would first like some feedback on the new input structure. Do you like it or would you like something else? Note that I went for quite a bit of nesting, since every level of nesting naturally corresponds to a choice you make that further constrains the remaining parameters.

ischeider:
Since all elements have a physics (here: SOLID), a cell type (HEX8) and a material (MAT 1), why not having a slightly flatter hierarchy like
```
ELEMENT_PHYSICS: SOLID
CELL_TYPE: HEX8
MAT: 1
ELEMENT_DATA: 
- KINEM nonlinear
```
If not all elements have a material, `MAT` could be moved to the element_data as well.

sebproell:
@ischeider The problem with such a flat hierarchy is that the content inside `ELEMENT_DATA` depends on `ELEMENT_PHYSICS` and `CELL_TYPE`. I think we can generate a schema for that, but it can be a bit hard to understand what you need to enter next. The flattened structure hides dependencies between parameters. Nesting conveys the dependency of different fields a lot better. @gilrrei Please correct me if I am wrong here.

Example:

```yaml
ELEMENT_PHYSICS: SOLID
CELL_TYPE: TET10
MAT: 1
ELEMENT_DATA: 
- EAS: full
```

This is not valid input, but it is difficult to pinpoint (for the schema!) why. It can just tell you that the whole `SOLID` + `TET10` + `ELEMENT_DATA` (with EAS) does not work.

```yaml
SOLID:
  TET10:
     ELELEMENT_DATA:
       MAT: 1
       EAS: full
```

Here, it is clear that the `EAS` property does not exist for the previous two choices you made, namely `SOLID` and `TET101.

This was my initial thought process but I will give it a try and see how your proposition works in practice :eyes:

sebproell:
@ischeider, I tried your proposed structure, and it has exactly the problems I anticipated. Completion is not great, since you essentially need to choose multiple parameters at once.

gilrrei:
@ischeider one important point that was not obvious to me but @sebproell showed it is that some element data depends on the cell type, for example, `LINE2` vs `LINE3` cells for `BEAM3K` require for the `TRIADS` parameter 6 or 9 entires respectively. So by nesting:

```yaml
        SOLID:
          LINE2:
            TRIADS: [1, 2, 3, 4, 5, 6]
```
and

```yaml
        SOLID:
          LINE3:
            TRIADS: [1, 2, 3, 4, 5, 6, 7 ,8 ,9]
```
It becomes much easier to do validation, autocompletion, and schemas, as we know that within the `LINEx` group, the `TRIADS` parameter has a specific length depending on the type of cell.

Note: I'm not saying this is a good example, but it was something I wasn't aware/thinking of.

tests/tutorials/tutorial_battery.4C.yaml
```diff
@@ -673,17 +673,37 @@ STRUCTURE GEOMETRY:
   SHOW_INFO: "summary"
   ELEMENT_BLOCKS:
     - ID: 1
-      ELEMENT_NAME: SOLIDSCATRA
-      ELEMENT_DATA: "MAT 12 KINEM nonlinear TYPE ElchElectrode"
+      ELEMENT_DATA:
```
gilrrei: I'm not fully up-to-date, but is there something else besides `ELEMENT_DATA` possible or in the making? Just asking to know if we can remove the group.

sebproell: I think the decision whether we want `ELEMENT_DATA` is independent of any other entries. I have no concrete plans to add something else, but I wouldn't exclude the possibility. I think you are right that we can just remove `ELEMENT_DATA`, though.

gilrrei: I can live with both, I'll leave it up to you :)

gilrrei (approved):

PR #1082, opened 2025-08-12
Use array of Tensors for material derivative of shape functions

amgebauer:
It is more convinient to use a `std::array<Tensor<double, 3>, num_nodes>` to represent the material derivatives of the shape functions. This PR also provides a function that creates a Matrix-view of the corresponding data.

I will use this in #961

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_tensor_matrix_conversion.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_tensor_matrix_conversion.hpp
@@ -15,6 +15,8 @@
 #include "4C_linalg_tensor.hpp"
 #include "4C_linalg_tensor_internals.hpp"
 
+#include <ranges>
+#include <tuple>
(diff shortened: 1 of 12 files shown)


src/core/linalg/src/dense/4C_linalg_tensor_matrix_conversion.hpp
```diff
@@ -143,6 +145,21 @@ namespace Core::LinAlg
     return Core::LinAlg::Matrix<n1, n2, ValueType>{tensor.data(), true};
   }
 
+  /*!
+   * @brief Creates a matrix view from a contiguous range of tensors.
```
sebproell: Maybe a small comment that explains how the outer extent previously managed by `array` translates into the respective matrix extent? Is it row or col? Should I be able to specify that?

tuchpaul (approved):

PR #1085, opened 2025-08-12
Rename CMake function for clarification

marksidoro:
After encountering the issue that I added my own external dependency through this function, resulting in a linker error, I renamed the function to clarify that it is used only for internal dependencies in accordance with a discussion with @sebproell.


sebproell (approved):

sebproell:
Thanks!

mayrmt (approved):

PR #1086, opened 2025-08-12
Add btb contact with btb coupling support and enhance runtime output

knarfnitram:
## Description and Context

So far it was not possible to couple two beam elements with point or rotation coupling and additionally let more beams interact with them. The general capability was already there, however due an error message this was not possible. Within this MR we are now only restricting the use of beam to beam contact with beam to beam coupling per node.

Additionally, I have added  two more quantities to the runtime output:
-  `CONTACT_ANGLE` to write the contact angle between two beams
- `CONTACT_TYPE` - to indicate which formulation(point, line or endpoint) is currently contributing


src/beamcontact/4C_beamcontact_input.cpp
```diff
@@ -218,7 +218,21 @@ void BeamContact::set_valid_parameters(std::map<std::string, Core::IO::InputSpec
           // whether to write visualization output for gaps
           parameter<bool>(
               "GAPS", {.description = "write visualization output for gap, i.e. penetration",
-                          .default_value = false})},
+                          .default_value = false}),
+          // whether to write visualization output for the contact angle of beams
+          parameter<bool>("CONTACT_ANGLE",
+              {.description = "write visualization output for gap, i.e. penetration",
```
mayrmt: The description seems wrong.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.hpp
```diff
@@ -165,6 +166,14 @@ namespace BeamInteraction
      */
     void print_summary_one_line_per_active_segment_pair(std::ostream& out) const override;
 
+    /**
+     * \brief Returns the type of this beam point couplint pair.
```
mayrmt:
```suggestion
     * \brief Returns the type of this beam point coupling pair.
```

src/beaminteraction/src/4C_beaminteraction_contact_pair.hpp
```diff
@@ -51,6 +51,15 @@ namespace BeamInteraction
   class BeamInteractionConditions;
   class BeamToSolidMortarManager;
 
+  enum class ContactPairType
+  {
+    unspecified_base,
+    beam_to_beam_contact,
+    beam_to_beam_point_coupling,
+    beam_to_solid_base,
```
mayrmt, line 59: Why does this one have the suffix `_base`? What does it refer to?

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
```diff
@@ -336,15 +336,32 @@ void Solid::ModelEvaluator::BeamInteraction::set_sub_model_types()
...
+    {
+      for (auto node : *nodes)
+      {
+        if (bpcc->contains_node(node))
+        {
+          FOUR_C_THROW(
+              "It is not possible to use beam-to-beam contact in combination with "
+              "beam-to-beam point coupling for the same node. Please reconsider boundary "
+              "conditions with the ids {} and {}, since they are intersection at the node with id "
+              "{}",
```
mayrmt: Should one really re-consider "boundary" conditions or rather "beam-to-beam interaction model / condition"? I am unsure, but when reading "boundary condition" I immediately thought of Dirichlet or Neumann, but they are unrelated here.
```suggestion
          FOUR_C_THROW(
              "It is not possible to use beam-to-beam contact in combination with "
              "beam-to-beam point coupling for the same node. Please reconsider boundary "
              "conditions with the ids {} and {}, since they are intersecting at the node with id "
              "{}",
```

mayrmt (changes requested):
@knarfnitram, just some thoughts... and a typo.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -5189,30 +5189,39 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::
  *----------------------------------------------------------------------------*/
 template <unsigned int numnodes, unsigned int numnodalvalues>
 void BeamInteraction::BeamToBeamContactPair<numnodes,
-    numnodalvalues>::get_all_active_contact_forces(std::vector<double>& forces) const
+    numnodalvalues>::get_all_active_contact_forces(std::vector<double>& forces,
```
isteinbrecher: Can we unify this with `get_all_active_contact_gaps` and have one function `get_all_active_beam_to_beam_visualization_values`?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.hpp
```diff
@@ -165,6 +166,14 @@ namespace BeamInteraction
...
+     */
+    ContactPairType get_type() const override final
```
isteinbrecher: The current way you are implementing the new features, this does not make sense. This function allows you to check wether an object is on beam to beam type, but you are still not able to cast to that type, this you still have to implement the visualization function in the base class and the overwrite it. Let’s talk about some possible alternative solutions.

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
```diff
@@ -336,15 +336,32 @@ void Solid::ModelEvaluator::BeamInteraction::set_sub_model_types()
...
-        "It is not yet possible to use beam-to-beam contact in combination with beam-to-beam point "
-        "coupling because every coupling point is also interpreted as a point of contact between 2 "
-        "beams.");
+  // extract beam to beam contact conditions
+  std::vector<const Core::Conditions::Condition*> beamtobeamcontactconditions;
+  discret_ptr_->get_condition("BeamToBeamContact", beamtobeamcontactconditions);
+
+  // ensure that no beam to beam contact condition is specified with a penalty point coupling
+  // condition
+  for (auto btbcc : beamtobeamcontactconditions)
```
isteinbrecher: Very nice extension of the current implementation!

I would only suggest a slight change in ordering, I would first check if there are point coupling pairs, if so get the beam to beam conditions and then do the check as proposed. In most cases there are no point coupling pairs so it makes sense to put the whole check inside the if.

src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_beamcontact.cpp
```diff
@@ -600,11 +618,22 @@ void BeamInteraction::SubmodelEvaluator::BeamContact::write_output_runtime_beam_
...
   }
 
+  if (beam_contact_params().beam_contact_runtime_visualization_output_params()->is_write_angles())
+  {
+    visualization_manager_ptr_->get_visualization_data().set_point_data_vector("angles", angles, 1);
+  }
+
+  if (beam_contact_params().beam_contact_runtime_visualization_output_params()->is_write_types())
+  {
+    visualization_manager_ptr_->get_visualization_data().set_point_data_vector("types", types, 1);
```
isteinbrecher: Types is not right, because we output a single type for each point, so type. Same for angleS.

isteinbrecher (commented):
Very nice changes! I suggest to skip the changes with the implemented enums and talk about if this actually makes sense (maybe I was to quick to suggest this yesterday) and if so do it in a follow up PR.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.hpp
knarfnitram: Thanks for catching the wrong description and error, I'll update it in the following next version.

src/beaminteraction/src/4C_beaminteraction_contact_pair.hpp
knarfnitram: The object to which it refers is called `BeamToSolidPairBase`, I tried to align the names to the existing objects.

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
knarfnitram: I rephrased it a little bit and additionally added the print command of the two conditions, therefore it should now be better possible to also understand where the error is coming from.

mayrmt (approved):
Looks good to me, but I let @isteinbrecher have the final say in this.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -5189,60 +5189,47 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::
  *----------------------------------------------------------------------------*/
 template <unsigned int numnodes, unsigned int numnodalvalues>
 void BeamInteraction::BeamToBeamContactPair<numnodes,
-    numnodalvalues>::get_all_active_contact_forces(std::vector<double>& forces) const
+    numnodalvalues>::get_all_active_beam_to_beam_visualization_values(std::vector<double>& forces,
```
isteinbrecher, line 5192: I like that we combined this!

PR #1089, opened 2025-08-13
Attempt to join pull requests of dependabot

c-p-schmidt:
## Description and Context
Yesterday, Dependabot opened several pull requests to update a single dependency (see #1076, #1078, #1079,#1080). According to the documentation https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups-- it should be possible to join those pull requests to a single one using `groups`, which is suggested in this PR.

Follows #1048, #1042, #327

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,12 +1,12 @@
 version: 2
 updates:
   - package-ecosystem: "github-actions"
-    directory: ".github/workflows"
-    schedule:
-      interval: "weekly"
-
-  - package-ecosystem: "github-actions"
     directories:
       - ".github/actions/**/*"
+      - ".github/workflows"
     schedule:
       interval: "weekly"
+    groups:
+      create-combined-pull-request:
+        patterns:
+          - "*"


.github/dependabot.yml
```diff
@@ -1,12 +1,12 @@
...
   - package-ecosystem: "github-actions"
     directories:
       - ".github/actions/**/*"
+      - ".github/workflows"
     schedule:
       interval: "weekly"
+    groups:
+      create-combined-pull-request:
+        patterns:
+          - "*"
```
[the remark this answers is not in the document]
c-p-schmidt: image

The documentation https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory-- seems to have a different opinion on that, or do I misunderstand?

amgebauer:
```suggestion
```

I think having the dependencies separated makes sense since we have to decide case by case whether we can just apply the fix or have to adapt some things. Combining tge workflows and the actions directory makes sense, though.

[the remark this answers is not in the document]
amgebauer: probably copilot is wrong here :grimacing:

amgebauer (dismissed):

.github/dependabot.yml
c-p-schmidt: If we do not put it in groups, it won't combine the changes concerning one dependency into one PR as far as I understand it.
One potential, but rather verbose solution would be to create a group for each dependency that can be updated by the Dependabot.
But I would also be fine with obtaining several PRs for one dependency. So I'm fine with reverting the groups part.

c-p-schmidt: I briefly talked to @sebproell, and we both do not like to explicitly spell out all dependencies in a group's section.
Thus, I decided to revert the group part.

We can still think about other solutions if we are annoyed by the granular pull requests of the Dependabot.

sebproell (approved):

PR #1091, opened 2025-08-13
Add infrastructure for python bindings in 4C

amgebauer:
This PR adds the infrastructure for python bindings with 4C. Python bindings can be used to call functions from 4C within python. The python bindings can simply be tested with pytest.

The folder python bindings integrate into the module folder structure like so:

```
src/
├── module/
│   ├── python/
│   │   ├── bindings/
│   │   └── tests/
│   ├── src/
│   ├── tests/
```

Python bindings can be build with the option `FOUR_C_ENABLE_PYTHON_BINDINGS` which requires `FOUR_C_WITH_PYBIND11`.  The bindings can then be installed in your python project with `pip install ~/path/to/build/py4C` and used in python with

```python
import py4C

py4C.module.func(...)
```

I implemented Python bindings for the G&R-fibers in the mixture rule.

### DIFF
--- a/src/mixture/python/bindings/4C_mixture.cpp
+++ b/src/mixture/python/bindings/4C_mixture.cpp
@@ -1,0 +1,291 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#include "4C_config_submodule_registry.hpp"
+
+#include "4C_io_input_parameter_container.hpp"
+#include "4C_mixture_constituent_remodelfiber_material_exponential.hpp"
+#include "4C_mixture_full_constrained_mixture_fiber.hpp"
+#include "4C_mixture_remodelfiber-internal.hpp"
+
+#include <pybind11/functional.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <cstddef>
+#include <memory>
+#include <numeric>
+#include <stdexcept>
+
+namespace py = pybind11;
+
+FOUR_C_NAMESPACE_OPEN
+namespace
+{
+  struct FiberMaterial
+  {
+    double k1;
+    double k2;
+    bool compression;
+  };
+
+  struct LinearGrowthPoissonTurnover
+  {
+    double growth_constant;
+    double decay_time;
+  };
+
+  FourC::Core::Mat::PAR::Parameter::Data create_material(const FiberMaterial& fiber_material)
+  {
+    auto container = FourC::Core::IO::InputParameterContainer();
+    container.add("K1", fiber_material.k1);
+    container.add("K2", fiber_material.k2);
+    container.add("COMPRESSION", fiber_material.compression);
+    return {.parameters = container};
+  }
+
+  class FullConstrainedMixtureFiberBinder
+  {
+   public:
+    FullConstrainedMixtureFiberBinder(const FiberMaterial& fiber_material,
+        const LinearGrowthPoissonTurnover& turnover_law, double lambda_pre,
+        FourC::Mixture::HistoryAdaptionStrategy adaptive_strategy, bool growth_enabled)
+        : mat_data_(create_material(fiber_material)),
+          fiber_(
+              std::make_shared<FourC::Mixture::RemodelFiberMaterialExponential<double>>(&mat_data_),
+              {turnover_law.growth_constant, turnover_law.decay_time, true}, lambda_pre,
+              adaptive_strategy, growth_enabled)
+    {
+    }
+    void reinitialize_history(double lambda_f, double time)
+    {
+      fiber_.reinitialize_history(lambda_f, time);
+    }
+    void recompute_state(double lambda_f, double time, double delta_time)
+    {
+      fiber_.recompute_state(lambda_f, time, delta_time);
+    }
+    void update() { fiber_.update(); }
+    double get_cauchy_stress() { return fiber_.computed_sigma_; }
+    double get_growth_scalar() { return fiber_.computed_growth_scalar_; }
+
+    double get_d_cauchy_stress_d_lambda()
+    {
+      return 2 * fiber_.computed_dsigma_dlambda_f_sq_ * fiber_.current_state_.lambda_f;
+    }
+
+    double get_d_growth_scalar_d_lambda()
+    {
+      return 2 * fiber_.computed_dgrowth_scalar_dlambda_f_sq_ * fiber_.current_state_.lambda_f;
+    }
+
+    double get_adaptive_tolerance()
+    {
+      if (fiber_.adaptive_history_strategy_ !=
+              FourC::Mixture::HistoryAdaptionStrategy::model_equation &&
+          fiber_.adaptive_history_strategy_ !=
+              FourC::Mixture::HistoryAdaptionStrategy::higher_order_integration)
+        FOUR_C_THROW(
+            "This property is only defined if the history adaption strategy is model_equation or "
+            "higher_order_integration");
+      return fiber_.adaptive_tolerance_;
+    }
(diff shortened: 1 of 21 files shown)


cmake/functions/four_c_auto_define_pybind11_module.cmake
```diff
@@ -0,0 +1,86 @@
...
+# GNU Lesser General Public License v3.0 or later.
+#
+# See the LICENSE.md file in the top-level for license information.
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+# Automatically create a library for the sources and headers in the current directory. The target
+# will be named based on the folder name. If this function is called recursively inside an already
+# defined module, the sources are appended to the already defined module. The module name is returned
+# in the variable AUTO_DEFINED_MODULE_NAME which is set at the call site.
```
sebproell: Copy-paste mistake. Please adapt.

cmake/functions/four_c_auto_define_pybind11_module.cmake
```diff
@@ -0,0 +1,86 @@
...
+      )
+  endif()
+
+  set(_bindings_name_base "unittests_${_bindings_for_module}")
+
+  file(GLOB_RECURSE _sources CONFIGURE_DEPENDS *.cpp)
+
+  foreach(_source ${_sources})
+    list(APPEND _four_c_python_bindings_sources ${_source})
+  endforeach()
```
sebproell: Where is this used?

cmake/functions/four_c_auto_define_pybind11_module.cmake
```diff
@@ -0,0 +1,86 @@
...
+
+  set(_target pybind11_${_bindings_name_base})
+  add_library(${_target}_objs OBJECT ${PROJECT_SOURCE_DIR}/cmake/dummy.cpp)
+  target_sources(${_target}_objs PRIVATE ${_sources})
```
sebproell: It could also be a separate target for clearer separation, but this would require more manual switching in different places. I need to think about this.

cmake/functions/four_c_auto_define_pybind11_module.cmake
```diff
@@ -0,0 +1,86 @@
...
+  target_link_libraries(${_target}_objs PUBLIC pybind11_config)
+  target_link_libraries(${_target}_objs PRIVATE ${_bindings_for_module}_unit_test_deps)
+  set_target_properties(${_target}_objs PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
sebproell: Is this required unconditionally? `four_c_private_compile_interface` would pull this in for shared lib builds.

cmake/functions/four_c_auto_define_pybind11_module.cmake
```diff
@@ -0,0 +1,86 @@
...
+
+  foreach(_source ${_sources})
+    list(APPEND _four_c_python_bindings_sources ${_source})
+  endforeach()
+
+  set(_target pybind11_${_bindings_name_base})
+  add_library(${_target}_objs OBJECT ${PROJECT_SOURCE_DIR}/cmake/dummy.cpp)
+  target_sources(${_target}_objs PRIVATE ${_sources})
+  target_link_libraries(${_target}_objs PUBLIC pybind11_config)
+  target_link_libraries(${_target}_objs PRIVATE ${_bindings_for_module}_unit_test_deps)
```
sebproell: I don't understand why you need this. Can you comment a little more? CMake is hard enough to understand as a language that I frequently have to remind myself why we do something 😆

pybind11/mixture/4C_mixture.cpp
sebproell, line 1: I think I will need to ask you some questions about pybind in person tomorrow 😄 Afterwards, I will review this file.

pybind11/config/4C_config_submodule_registry.hpp
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+  inline std::vector<SubmoduleRegistryEntry>& get_submodule_registry()
+  {
+    static std::vector<SubmoduleRegistryEntry> registry;
+    return registry;
+  }
+}  // namespace Py4C
+
+#define FOUR_C_PYBIND_REGISTER_SUBMODULE(name, doc, init_func)          \
```
sebproell, line 51: Docs.

cmake/functions/four_c_auto_define_pybind11_module.cmake
amgebauer: Yes, we need this (but we would need this for the whole codebase actually). I think we need some additional check that we can only enable python bindings if we actually use shared libs for 4C.

cmake/functions/four_c_auto_define_pybind11_module.cmake
amgebauer: If I understand the `_unit_test_deps`-suffix correctly, it contains all dependencies of the respective module that are required to build a unit test for this module. These should be the same dependencies that we need for the bindings. Probably we should rename the suffix to `_deps`

amgebauer:
I will try to look into this again to get this merged in the upcoming weeks. The only thing that is missing is a way to reliably test the bindings

amgebauer:
Now this PR is ready

src/mixture/python/bindings/4C_mixture.cpp
```diff
@@ -0,0 +1,293 @@
...
+    double get_d_cauchy_stress_d_lambda()
+    {
+      return fiber_.evaluate_d_current_cauchy_stress_d_lambda_f_sq() * 2 * current_lambda_f_;
+    }
+    double get_d_growth_scalar_d_lambda()
+    {
+      return fiber_.evaluate_d_current_growth_scalar_d_lambda_f_sq() * 2 * current_lambda_f_;
+    }
+    double get_current_lambda_r() { return fiber_.evaluate_current_lambda_r(); }
+    // double get_current_lambda_e() { return fiber_.evaluate_current_lambda_e(); }
```
c-p-schmidt: Probably a leftover from development?.

src/mixture/python/bindings/4C_mixture.cpp
```diff
@@ -0,0 +1,293 @@
...
+                 fiber_.evaluate_current_lambda_r(), 1.0, fiber_.evaluate_current_growth_scalar()) /
+             fiber_.evaluate_current_growth_scalar();
+    }
+    double get_d_cauchy_stress_d_lambda()
+    {
+      return fiber_.evaluate_d_current_cauchy_stress_d_lambda_f_sq() * 2 * current_lambda_f_;
+    }
+    double get_d_growth_scalar_d_lambda() { return 0; }
+    double get_current_lambda_r() { return fiber_.evaluate_current_lambda_r(); }
+    // double get_current_lambda_e() { return fiber_.evaluate_current_lambda_e(); }
```
c-p-schmidt: Same here.

src/mixture/python/bindings/4C_mixture.cpp
```diff
@@ -0,0 +1,293 @@
...
+    {
+      return fiber_.evaluate_growth_evolution_equation_dt(current_lambda_f_,
+                 fiber_.evaluate_current_lambda_r(), 1.0, fiber_.evaluate_current_growth_scalar()) /
+             fiber_.evaluate_current_growth_scalar();
+    }
+    double get_d_cauchy_stress_d_lambda()
+    {
+      return fiber_.evaluate_d_current_cauchy_stress_d_lambda_f_sq() * 2 * current_lambda_f_;
+    }
+    double get_d_growth_scalar_d_lambda() { return 0; }
```
c-p-schmidt, line 209: Here, you explicitly do not want to call `fiber_.evaluate_d_current_cauchy_stress_d_lambda_f_sq() * 2 * current_lambda_f_` as above?

amgebauer: This is the explicit version of the fiber, so the derivative is actually 0. I only have this function to be consistent with the implicit version.

amgebauer:
@c-p-schmidt Thanks for your comments. It's now ready for re-review

PR #1092, opened 2025-08-13
Use proper YAML for generated domain input

sebproell:
Part of #571 

The ability to generate a domain directly in the input file is nice. This PR makes the existing implementation actually useful by moving the string-based input to YAML. In the future, we probably want to enhance this further, e.g., by providing better boundary condition support.

@gilrrei Probably breaks FourCIPP?


sebproell:
@gilrrei I noticed that #1081 and this PR combined blow up the schema to 230k lines (16MB). An input for a domain section alone is 15k lines. Is this going to be a problem? Do we need to factor out these repeated blocks? The schema verification job is also failing on this PR.

gilrrei:
@sebproell 

> @gilrrei Probably breaks FourCIPP?

Yes, but in a good sense, less stuff hard coded :)

gilrrei:
@sebproell 

> @gilrrei I noticed that #1081 and this PR combined blow up the schema to 230k lines (16MB). An input for a domain section alone is 15k lines. Is this going to be a problem?

I think it is possible that some implementations are not able to handle the JSON schema; however, up until now, I haven't encountered a problem. VSCode still manages it effortlessly. I tried it with `jsonschema_rs`, which is what we use in FourCIPP, and it is also not a problem:
```bash
Traceback (most recent call last):
  File "/home/rei/data/papers/particle_optimization/paper/test.py", line 80, in <module>
    validate_using_json_schema(
  File "/home/rei/data/papers/particle_optimization/paper/test.py", line 77, in validate_using_json_schema
    validator.validate(data)
jsonschema_rs.ValidationError: "3a" is not of type "integer"

Failed validating "type" in schema["properties"]["FLUID DOMAIN"]["properties"]["subdivisions"]["items"]

On instance["FLUID DOMAIN"]["subdivisions"][1]:
    "3a"
```

> Do we need to factor out these repeated blocks?

I think we should or at least should look into it. I think there is a lot of redundant information currently, for the result description (I think). Or solver sections 1 to 10 or element definitions. If the information is identical, the information can be reduced by referencing a separate schema. This makes things cleaner as well.

>The schema verification job is also failing on this PR.

I think this is due to the fact that package used for it is not able to handle it. We should switch to a different validator and we should be fine

src/global_legacy_module/4C_global_legacy_module_validparameters.cpp
```diff
@@ -176,11 +177,20 @@ std::map<std::string, Core::IO::InputSpec> Global::valid_parameters()
             .required = false});
   };
 
+  const auto add_domain_section = [spec = Core::IO::GridGenerator::RectangularCuboidInputs::spec()](
+                                      auto& specs, const std::string& field_identifier)
+  {
+    specs[field_identifier + " DOMAIN"] = group(field_identifier + " DOMAIN", {spec},
```
lauraengelhardt, line 183: No in this PR but I think we should account for the fact that the domain is a cuboid in the naming (`CUBOID DOMAIN` or similar).

src/core/io/src/4C_io_input_file_utils.cpp
```diff
@@ -224,34 +224,35 @@ void Core::IO::read_design(InputFile& input, const std::string& name,
       const Core::FE::Discretization& actdis = get_discretization(disname);
 
       std::vector<double> box_specifications;
+      box_specifications.reserve(9);
```
lauraengelhardt: Can we add a comment here, why 9?

lauraengelhardt: For consistency, what do you think about calling this cuboid_specifications? 
Now we use 'box', 'cuboid', and 'domain' essentially referring to the same thing.

lauraengelhardt (commented):
Nice work, thanks for tackling this. 

We need to update the documentation here as well, right?: https://4c-multiphysics.github.io/4C/documentation/analysis_guide/geometrydefinition.html#finite-element-discretization

Maybe we could also update the docu of this function here https://github.com/4C-multiphysics/4C/blob/a4e78aa0a7d619071469ee87cfb80c065f57b4fa/src/core/io/src/4C_io_meshreader.hpp#L92 and mention that we read meshes from the three different sections?

ischeider (approved):
While there may be many ideas for extension, I am happy to have this new style.
One major concern: It's good that the <..> DOMAIN now appear In the documentation (currently it isn't), but it's a bit overwhelming - each DOMAIN description has about 1700 lines and contains all element types, not only those applicable to the respective physics. I am afraid, it's probably hard to change this at this point, but I would like to keep that in mind.

sebproell:
@ischeider, I think the `refs` in #1103 might help us with the documentation.

src/core/io/src/4C_io_input_file_utils.cpp
sebproell: I added the comment, but we need to do another pass over the boundary conditions anyway, so I would not start renaming these variables now.

sebproell:
@lauraengelhardt I updated the docs. Note that the input file content itself is already updated due to the awesome mechanism that extracts examples from files :+1:

lauraengelhardt (approved):

PR #1097, opened 2025-08-14
Fix beam-to-solid mortar maps

isteinbrecher:
## Description and Context

This PR fixes a bug that occurs when using beam-to-solid mortar methods in combination with point couplings (the classical ones, not penalty point couplings). There it can happen that multiple nodes point to the same DOFs which resulted in some missing entries in the global stiffness matrix. We now ensure that the maps stored in the mortar manager have unique entries.


maxfirmbach:
I guess consecutive increasing GIDs is always a very good thing to have :wink:.

isteinbrecher:
@maxfirmbach yes, this only affects these two maps.

PR #1098, opened 2025-08-14
Replace `Epetra_LinearProblem` with a custom class

vovannikov:
## Description and Context
The next step to proceed with #1071 is to get rid of `Epetra_LinearProblem`. It is now only used in `NOX::FSI::LinearSystemGCR` (see comment below) an in `Core::LinearSolver::DirectSolver` for older version of Trilinos.

## Exception `NOX::FSI::LinearSystemGCR`
I had also to bring back the use of `NOX::Epetra::Scaling` in `NOX::FSI::LinearSystemGCR` since this system implements its own GCR and GMRES solution algorithms, this was earlier removed in #1077. Surely, these manual algos should be replaced by the suitable Trilinos packages. So since this class requires significant refactoring anyway, I think it is OK to have both `NOX::Epetra::Scaling` and `Epetra_LinearProblem` still used there. Alternative solutions would be to either disable the scaling feature (in fact, it is not used currently) or to create a `SparseOperator`-like view wrapper around `Epetra_Operator` as similarly done for `Core::Linalg::Vector` and its view. However, I would prefer not to use this option if other alternatives are available.

#723, #1071, #1077

### DIFF
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_linearproblem.hpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_linearproblem.hpp
@@ -1,0 +1,31 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_SOLVER_NONLIN_NOX_LINEARPROBLEM_HPP
+#define FOUR_C_SOLVER_NONLIN_NOX_LINEARPROBLEM_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_linalg_sparseoperator.hpp"
+#include "4C_linalg_vector.hpp"
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace NOX::Nln
+{
+  class LinearProblem
+  {
+   public:
+    std::shared_ptr<Core::LinAlg::SparseOperator> jac;
+    std::shared_ptr<Core::LinAlg::Vector<double>> lhs;
+    std::shared_ptr<Core::LinAlg::Vector<double>> rhs;
+  };
+}  // namespace NOX::Nln
+
+FOUR_C_NAMESPACE_CLOSE
+
+#endif
--- a/src/contact/4C_contact_nox_nln_contact_linearsystem.cpp
+++ b/src/contact/4C_contact_nox_nln_contact_linearsystem.cpp
@@ -149,23 +149,20 @@ Core::LinAlg::SolverParams NOX::Nln::CONTACT::LinearSystem::set_solver_options(
 
 /*----------------------------------------------------------------------*
  *----------------------------------------------------------------------*/
-void NOX::Nln::CONTACT::LinearSystem::set_linear_problem_for_solve(
-    Epetra_LinearProblem& linear_problem, Core::LinAlg::SparseOperator& jac,
-    Core::LinAlg::Vector<double>& lhs, Core::LinAlg::Vector<double>& rhs) const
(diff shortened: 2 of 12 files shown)


maxfirmbach:
@mayrmt How important is `NOX::FSI::LinearSystemGCR` can this be removed (in an easy way)?

mayrmt:
@maxfirmbach, I have to look into it in the code in detail to voice an educated opinion. Will try to find the time today...

src/solver_nonlin_nox/4C_solver_nonlin_nox_linearproblem.hpp
```diff
@@ -0,0 +1,31 @@
...
+FOUR_C_NAMESPACE_OPEN
+
+namespace NOX::Nln
+{
+  class LinearProblem
+  {
+   public:
+    std::shared_ptr<Core::LinAlg::SparseOperator> jac;
+    std::shared_ptr<Core::LinAlg::Vector<double>> lhs;
+    std::shared_ptr<Core::LinAlg::Vector<double>> rhs;
```
sebproell, line 25: Maybe just a ref or raw pointer? It seems you aren't sharing the ptr with anything and you are also not owning it, if I read the code correctly. Also, I would write `struct` and drop `public:`.

vovannikov: Using refences was my first implementation (https://github.com/4C-multiphysics/4C/commit/e1270bb7bca082d4c8a9dc18d45a4e410b7d284a#diff-3623af079a47601db35a9b5fcdb4f5350fe33458dde5adf10d45572c9b1bb090), but then the test `stc-mltube-sosh8` fails since the Jacobian inside of the `LinearProblem` object needs to get replaced without altering the original Jacobian:
https://github.com/4C-multiphysics/4C/blob/432a1301d92ec917469bf61e1b4486b74f4da287/src/structure_new/src/nonlinear_solver/4C_structure_new_nln_linearsystem_scaling.cpp#L313-L329C23

Using raw pointers could be also an option, but then there is no way to guaranty life time of the Jacobian and the vectors. Currently, this is not an issue, but the code part above could have been (and probably should be) reimplemented in a different way such that the scaled Jacobian is assigned not to a member shared pointer `stiff_scaled_` but to a pointer inside the `LinearProblem` object.

As a sidenote, I think shared pointers are overused across the code, too much shared state between the classes. For this reason, I think that `LinearProblem` class should not exist at all, so I would deem it a transient solution on the way to remove Epetra.

vovannikov: I would also write `struct` and that again was my first option to go :) https://github.com/4C-multiphysics/4C/commit/e1270bb7bca082d4c8a9dc18d45a4e410b7d284a But then all the other forward declarations for this class should be made with the same keyword (due to `-Wmismatched-tags`), which I find a bit less intuitive because of that:

1. All other classes are forward declared with `class`
2. One then would need to always check which keyword is used for the class declaration.

I think if `-Wmismatched-tags` is applied, then it is better to use the same keyword across the code otherwise it can easily become a headache: should I forward declare a particular class with `class` or `struct`?

See the result: https://github.com/4C-multiphysics/4C/actions/runs/16960666106/job/48072466399

vovannikov:
@maxfirmbach @mayrmt What if we merge this PR without removing `NOX::FSI::LinearSystemGCR`? I would create a separate PR for that purpose anyway.

mayrmt:
@vovannikov  Fine for me.

maxfirmbach:
@vovannikov That is fine for me as well. I'll have a look at this PR now.

mayrmt (approved):
I saw some RCPs and I'm not sure if they could be `std::shared_ptr`.

maxfirmbach:
> I saw some RCPs and I'm not sure if they could be `std::shared_ptr`.

That's most likely due to interfacing with Trilinos functions directly ... there we need RCPs.

PR #1099, opened 2025-08-14
Allow for monolithic partitioning on a single MPI rank

isteinbrecher:
## Description and Context

Currently setting
```yaml
MESH PARTITIONING:
  METHOD: monolithic
```
and running the simulation on a single MPI rank produces an error. This PR fixes this, by simply skipping the partitioning in case of a single MPI rank.

### DIFF
--- a/src/core/io/src/4C_io_meshreader.cpp
+++ b/src/core/io/src/4C_io_meshreader.cpp
@@ -601,6 +601,12 @@ namespace
         colmap = std::make_shared<Core::LinAlg::Map>(
             -1, graph->col_map().num_my_elements(), graph->col_map().my_global_elements(), 0, comm);
 
+        if (Core::Communication::num_mpi_ranks(comm) == 1)
+        {
+          // In the case of a single MPI rank we don't have to perform the redistribution.
+          break;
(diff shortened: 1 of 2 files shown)


mayrmt:
The proposed change circumvents the problem. Can it be fixed?

sebproell:
> The proposed change circumvents the problem. Can it be fixed?

I agree; the fix seems to be in the wrong place. I assume you would need this check in front of every call to `redistribute`. Can we make the `redistribute` function robust enough to accept the 1-rank scenario?

isteinbrecher:
I don't know what the problem is. @maxfirmbach can you have a look?

maxfirmbach:
I don't like these skips ... already discussed with @georghammerl at some point :wink:. I can check ...

maxfirmbach:
Seems like the issue is somewhere in `FillComplete()` not yet sure why this appears though ...

isteinbrecher:
Thanks to @maxfirmbach for pointing out where the error happens and thanks to my old self for fixing this in some unrelated branch a while ago. Let's see if the pipeline passes.

PR #1100, opened 2025-08-14
Fix beam output issue

isteinbrecher:
## Description and Context

Using monolithic partitioning and Reissner beam output with stress/strain data leads to an error in the output mechanism of the first step. This is because the output requires certain Gauss point vectors to have the correct size, but these vectors are not correctly set because they are not packed / unpacked. This PR fixes this issue by communicating and initialising these vectors with the correct size.


maxfirmbach:
Is this only a problem in the monolithic partitioning case? I don't get why this doesn't appear for other partitioning types.

isteinbrecher:
> Is this only a problem in the monolithic partitioning case? I don't get why this doesn't appear for other partitioning types.

I only tried the monolithic case. But I am pretty sure that we test all other options with a single core. Maybe the partitioning is right but there is an issue in the distributed search?

maxfirmbach:
I don't see how this is related to the search itself :thinking:.

isteinbrecher:
Sorry, I thought this was a question on the other PR...

I think this is a problem in all cases where we do a redistribution and then evaluate the beam output without evaluating the residual and or stiffness first.

maxfirmbach:
@isteinbrecher Alright, this makes sense :+1:.

knarfnitram (approved):

mayrmt (commented):
Does the test need to be that large?

isteinbrecher:
> Does the test really need to be that large?

There likely is a smaller model setup that triggers the bug, but the runtime is only 1s, so I would be fine with the current version.

PR #1103, opened 2025-08-14
Condense metadata by sharing duplicated information

sebproell:
Inspired by #1092, Depends on #1102 

Introduce a `$references` section in the metadata file that contains duplicated specs. These can be referred to by `$ref$. 

Just to be clear: this has nothing to do with how your input file looks. This is only useful for internal tooling.


sebproell:
Can I get a second look here please?

PR #1110, opened 2025-08-16
Draft: Include new material for volume fractions in porofluid-pressure-based framework

lkoeglmeier:
## Description and Context
This pr adds a new material for the additional porous network (volfracs). 
- We now have two closing relations implemented: "blood lung" and "homogenized_vasculature_tumor"(existing)
- In the new material (blood lung), the volume fractions are not solved as primary variables but dependent on the deformation of the mixture and pressures of air and blood
- tests (2D/3D) are included


mayrmt:
@lkoeglmeier What is the run time of the input file tests? At least one of the meshes seems to be very large for a test of a material.

lkoeglmeier:
This PR is not yet ready for review. After a discussion with @bwirthl, some changes may be made to the framework, and it will be split into more commits for easier review.

lkoeglmeier:
> @lkoeglmeier What is the run time of the input file tests? At least one of the meshes seems to be very large for a test of a material.

Thank you for the comment @mayrmt. I will reduce the mesh in the 2D example.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -3073,10 +3073,15 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 "NUMSCAL", {.description = "number of scalars coupled with this problem"}),
             parameter<int>("TOTALNUMDOF", {.description = "total number of multiphase-dofs"}),
             parameter<int>("NUMVOLFRAC", {.description = "number of volfracs"}),
+            parameter<std::string>("VOLFRAC_CLOSING_RELATION",
```
c-p-schmidt: You could define an `enum class` for the closing relation and then use this instead of the string as an input.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -3073,10 +3073,15 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                {.description = "type of closing relation for volume fraction material: "
+                                "blood_lung, homogenized_vasculature_tumor (default)",
+                    .default_value = "homogenized_vasculature_tumor"}),
             parameter<std::vector<int>>("SCALE", {.description = "advanced reaction list",
                                                      .size = from_parameter<int>("TOTALNUMDOF")}),
             parameter<std::string>("COUPLING",
-                {.description = "type of coupling: scalar_by_function, no_coupling (default)"}),
+                {.description = "type of coupling: scalar_by_function, no_coupling (default)",
```
c-p-schmidt, line 3120: Same for the COUPLING here.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -3136,6 +3141,28 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+  /*----------------------------------------------------------------------*/
+  // one volume fraction pressure material for vascular units in the lungs for multiphase flow in a
+  // poroelastic material
+  {
+    known_materials[Core::Materials::m_fluidporo_volfrac_pressure_blood_lung] =
+        group("MAT_FluidPoroVolFracPressureBloodLung",
+            {
+                parameter<double>("DENSITY", {.description = "density of phase"}),
+                parameter<double>("PERMEABILITY", {.description = "permeability of phase"}),
+                parameter<int>("VISCOSITYLAWID", {.description = "ID of viscosity law"}),
```
c-p-schmidt: For better readability, maybe?
```suggestion
                parameter<int>("VISCOSITY_LAW_ID", {.description = "ID of viscosity law"}),
```
Would also apply to the `INITIALVOLFRAC`

src/mat/4C_mat_fluidporo_multiphase.cpp
```diff
@@ -55,22 +56,84 @@ void Mat::PAR::FluidPoroMultiPhase::initialize()
...
+  // fractions with closing relation "blood_lung" or "homogenized_vasculature_tumor"
+  // It is not possible to have both volume fractions in the same problem so far.
+  if ((int)matids_.size() > numfluidphases_)
+  {
+    const int volfracmatid = (matids_)[numfluidphases_];
+    std::shared_ptr<Core::Mat::Material> volfracmat = material_by_id(volfracmatid);
+    if (volfracmat->material_type() == Core::Materials::m_fluidporo_volfrac_pressure_blood_lung)
+    {
+      closing_relation_volfrac_ = Mat::PAR::PoroFluidPressureBased::cr_evolutionequation_blood_lung;
+      numvolfrac_ = (int)(((int)matids_.size() - numfluidphases_));
```
c-p-schmidt:
```suggestion
      numvolfrac_ = static_cast<int>(matids_.size()) - numfluidphases_;
```

You could also replace the other c-style casts `(type)` by `static_cast<type>`

src/mat/4C_mat_fluidporo_multiphase.cpp
```diff
@@ -55,22 +56,84 @@ void Mat::PAR::FluidPoroMultiPhase::initialize()
...
+    }
+    else if (volfracmat->material_type() == Core::Materials::m_fluidporo_singlevolfrac)
+    {
+      closing_relation_volfrac_ =
+          Mat::PAR::PoroFluidPressureBased::cr_evolutionequation_homogenized_vasculature_tumor;
+      numvolfrac_ = (int)(((int)matids_.size() - numfluidphases_) / 2);
```
c-p-schmidt: I have seen that this was already there like this. But is this doing what is desired, i.e. dividing by 2 and then casting to an `int` always works as desired?

src/mat/4C_mat_fluidporo_multiphase.cpp
```diff
@@ -55,22 +56,84 @@ void Mat::PAR::FluidPoroMultiPhase::initialize()
...
+    }
+  }
+  else
+  {
+    numvolfrac_ = 0;
+  }
+
+  if (numvolfrac_)
+  {
+    if (closing_relation_volfrac_ ==
```
c-p-schmidt: Aren’t those if, else if, else cases exactly the same as some lines above?

Why not moving the logic directly up there? Then you would not have to double up the branch logic here.

src/mat/4C_mat_fluidporo_multiphase_singlereaction.cpp
```diff
@@ -339,6 +340,28 @@ Mat::PAR::FluidPoroSingleReaction::set_coupling_type(const Core::Mat::PAR::Param
   }
 }
 
+int Mat::PAR::FluidPoroSingleReaction::
+    get_numfluidphases_from_closing_relation_of_additional_porous_network_type(
+        const Core::Mat::PAR::Parameter::Data& matdata)
+{
+  if ((matdata.parameters.get<std::string>("VOLFRAC_CLOSING_RELATION")) == "blood_lung")
```
c-p-schmidt: If you move this to an enum class as suggested above you can switch the different cases here.

src/mat/4C_mat_fluidporo_multiphase_singlereaction.hpp
```diff
@@ -107,6 +107,11 @@ namespace Mat
       Mat::PAR::FluidPoroSingleReaction::PorofluidReactionCoupling set_coupling_type(
           const Core::Mat::PAR::Parameter::Data& matdata);
 
+      // get number of fluid phases in multiphase porespace from closing relation type of volfrac
```
c-p-schmidt, line 110: You need e.g. `//!` or `///` otherwise it is not consumed by doxygen.

src/mat/4C_mat_fluidporo_singlephase.cpp
```diff
@@ -514,4 +514,118 @@ void Mat::FluidPoroVolFracPressure::initialize()
...
+ *----------------------------------------------------------------------*/
+void Mat::FluidPoroVolFracPressureBloodLung::pack(Core::Communication::PackBuffer& data) const
+{
+  // pack type of this instance of ParObject
+  int type = unique_par_object_id();
+  add_to_pack(data, type);
+
+  // matid
+  int matid = -1;
+  if (params_ != nullptr) matid = params_->id();  // in case we are in post-process mode
```
c-p-schmidt: Is this still needed, or is the comment outdated? Because post-processing should not be necessary, right?

src/mat/4C_mat_fluidporo_singlephase.hpp
```diff
@@ -147,6 +147,65 @@ namespace Mat
...
+      //@}
+
+     private:
+      bool isinit_;
+
+    };  // class FluidPoroVolFracPressure
+
+    namespace PoroFluidPressureBased
+    {
+      enum ClosingRelation
```
c-p-schmidt:
```suggestion
      enum class ClosingRelation
```
Then it would be automatically scoped and the cr_ prefix would be irrelevant.

src/mat/4C_mat_fluidporo_singlephase.hpp
```diff
@@ -551,6 +610,134 @@ namespace Mat
...
+
+    //! @name Packing and Unpacking
+
+    /*!
+     \brief Return unique ParObject id
+
+     every class implementing ParObject needs a unique id defined at the
+     top of parobject.H (this file) and should return it in this method.
+     */
+    int unique_par_object_id() const override
```
c-p-schmidt, line 648: All these methods that `override` other methods inherit their documentation, meaning that you just need to spell out things that are specific to this derived version, if there is anything.

This way you can save time on writing docu for inherited methods that is better invested into improved docu of the base. ;-)

And your file is less cluttered.

src/mat/4C_mat_fluidporo_singlephase.hpp
```diff
@@ -551,6 +610,134 @@ namespace Mat
...
+    /// return copy of this material object
+    std::shared_ptr<Core::Mat::Material> clone() const override
+    {
+      return std::make_shared<FluidPoroVolFracPressureBloodLung>(*this);
+    }
+
+    /// Return quick accessible material parameter data
+    Core::Mat::PAR::Parameter* parameter() const override { return params_; }
+
+    /// return minimum volume fraction
```
c-p-schmidt: Is this correct? Density vs volume fraction?

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_phasemanager.hpp
```diff
@@ -664,6 +691,24 @@ namespace Discret
...
           return 0.0;
         };
+        //! get relative diffusivity of volume fraction pressure
+        double dyn_viscosity_vol_frac_pressure_blood_lung(
+            int volfracpressnum, double abspressgrad, int matnum = 0) const override
+        {
+          FOUR_C_THROW(
+              "Dynamic Viscosity (VolFracPressure in the lungs) not available for this phase "
+              "manager!");
+          return 0.0;
```
c-p-schmidt: You don’t need these returns after the FOUR_C_THROWS, FOUR_C_THROW is handling that for you.

src/scatra_ele/4C_scatra_ele_calc_porofluid_pressure_based.cpp
```diff
@@ -1840,6 +1860,19 @@ void Discret::Elements::ScaTraEleInternalVariableManagerPorofluidPressureBased<n
   volume_fractions_ = phase_manager_->vol_frac();
   volume_fractions_pressure_ = phase_manager_->vol_frac_pressure();
 
+  // set closing relation type volfrac
+  auto closing_relation = std::invoke(
```
c-p-schmidt, line 1864: Nice that you hide such expert knowledge (i.e. which number of dofs etc result from which closing relation) in this lambda!
Could also be applicable in other files I saw before.

src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm.cpp
```diff
@@ -724,6 +736,30 @@ void PoroPressureBased::PorofluidAlgorithm::collect_runtime_output_data()
     visualization_writer_->append_result_data_vector_with_context(
         *porosity_multi, Core::IO::OutputEntity::node, {"porosity"});
   }
+
+  //! determinant of deformation gradient
+  if (output_det_def_grad_)
+  {
+    //! convert dof-based vector into node-based multi-vector for postprocessing
+    std::shared_ptr<Core::LinAlg::MultiVector<double>> det_def_grad_multi =
+        PoroPressureBased::convert_dof_vector_to_node_based_multi_vector(
```
c-p-schmidt: This method is not poro-related, so it could/should be moved out of poro to core, if there is not already a method available there doing this.
But not part of this PR but a follow up.

src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm.cpp
```diff
@@ -724,6 +736,30 @@ void PoroPressureBased::PorofluidAlgorithm::collect_runtime_output_data()
     visualization_writer_->append_result_data_vector_with_context(
         *porosity_multi, Core::IO::OutputEntity::node, {"porosity"});
   }
+
+  //! determinant of deformation gradient
+  if (output_det_def_grad_)
+  {
+    //! convert dof-based vector into node-based multi-vector for postprocessing
+    std::shared_ptr<Core::LinAlg::MultiVector<double>> det_def_grad_multi =
```
c-p-schmidt:
```suggestion
    auto det_def_grad_multi =
```
Maybe?

tests/input_files/porofluid_pressure_based_elast_scatra_2D_quad4_nodetopoint_artery_airway_coupling_volfrac_closingrelation_lung_mono.4C.yaml
```diff
@@ -0,0 +1,528 @@
...
+    linear_solver_id: 1
+STRUCTURAL DYNAMIC:
+  INT_STRATEGY: "Standard"
+  DYNAMICTYPE: "OneStepTheta"
+  LINEAR_SOLVER: 1
+STRUCTURAL DYNAMIC/ONESTEPTHETA:
+  THETA: 1
+porofluid_dynamic:
+  output:
+    volfrac_blood_lung: true
```
c-p-schmidt, line 37: Here you add the new added output quantities. Did you also add a vtk-test, to test whether they are written out properly?

src/w1/4C_w1_poro_evaluate.cpp
```diff
@@ -731,6 +734,39 @@ void Discret::Elements::Wall1Poro<distype>::gauss_point_loop(Teuchos::ParameterL
...
+          solidporo_fluid_properties.number_of_volfracs_)
+    return Mat::PAR::PoroFluidPressureBased::cr_evolutionequation_blood_lung;
+  else if (solidporo_fluid_properties.number_of_fluid_dofs_per_node_ ==
+           solidporo_fluid_properties.number_of_fluid_phases_in_multiphase_porespace_ +
+               solidporo_fluid_properties.number_of_volfracs_ * 2)
+    return Mat::PAR::PoroFluidPressureBased::cr_evolutionequation_homogenized_vasculature_tumor;
+  else
+  {
+    FOUR_C_THROW("The Number of volume fractions is not correct in the solid evaluator!");
+    return Mat::PAR::PoroFluidPressureBased::cr_undefined;
```
c-p-schmidt: Not necessary, `FOUR_C_THROW` handles that for you.
```suggestion
```

src/w1/4C_w1_poro_evaluate.cpp
```diff
@@ -765,12 +801,17 @@ void Discret::Elements::Wall1Poro<distype>::gauss_point_loop_pressure_based(
...
+  // Initialize variables of multiphase porous medium flow
+  const SolidPoroFluidProperties solidporo_fluid_properties = evaluate_porofluid_properties();
+
+  const Mat::PAR::PoroFluidPressureBased::ClosingRelation type_volfrac_closingrelation =
+      get_volfrac_closing_relation_type(solidporo_fluid_properties);
+
+  const bool hasvolfracs =
+      (solidporo_fluid_properties.number_of_fluid_dofs_per_node_ >
+          solidporo_fluid_properties.number_of_fluid_phases_in_multiphase_porespace_);
+  std::vector<double> phiAtGP(solidporo_fluid_properties.number_of_fluid_dofs_per_node_);
```
c-p-schmidt:
```suggestion
  std::vector<double> phi_at_gp(solidporo_fluid_properties.number_of_fluid_dofs_per_node_);
```
Maybe better for consistency with the rest and readability.

src/w1/4C_w1_poro_evaluate.cpp
```diff
@@ -2239,29 +2299,138 @@ template <Core::FE::CellType distype>
...
+                           get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
+                               *fluidmulti_mat_, numfluidphases)
+                               .scaling_parameter_pressure());
+          }
+        });
+
+    double dvolfrac_multiphaseporespace_dDetDefGrad = dporosity_dJ - dvolfrac_dDetDefGrad;
+
+    double dps_dDetDefGrad =
+        dvolfrac_multiphaseporespace_dDetDefGrad * pow(porosity, -1.0) * fluidpress +
```
c-p-schmidt: I would not use pow(x, -1.0) or pow(x, -2.0), both for performance of the calculation (according to ChatGPT) but also for readability and use 1.0 / x or 1.0 / (x*x) instead.

src/w1/4C_w1_poro_evaluate.cpp
```diff
@@ -2239,29 +2299,138 @@ template <Core::FE::CellType distype>
...
+                                   .scaling_parameter_deformation()) *
+                    pow((phiAtGP[0] / volfracpressure),
+                        PoroPressureBased::ElementUtils::
+                            get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
+                                *fluidmulti_mat_, numfluidphases)
+                                .scaling_parameter_pressure()));
+          }
+        });
+
+    // p_s = (porosity - sumaddvolfrac)/porosity * fluidpress
```
c-p-schmidt: Wouldn’t these lines be better suited below, where you actually do the calculation?

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
```diff
@@ -1533,16 +1533,1203 @@ namespace Discret
...
+            int phasetoadd,     //!< index of current phase
+            int numdofpernode,  //!< total number of DOFs/phases
+            const PoroFluidManager::PhaseManagerInterface& phasemanager,  //!< phase manager
+            const PoroFluidManager::VariableManagerInterface<nsd, nen>&
+                variablemanager,  //!< variable manager
+            double timefacfac,    //!< domain-integration factor
+            double fac,           //!< domain-integration factor times time-integration factor
+            double det) override;
+
+        //! evaluate off-diagonal coupling matrix with structure
```
c-p-schmidt:
```suggestion
        //! evaluate off-diagonal coupling matrix with scatra
```
I guess a copy-paste error from above according to the name (also below). Some more details e.g. linearization of what residual contributions w.r.t. scatra could be helpful.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
```diff
@@ -1533,16 +1533,1203 @@ namespace Discret
...
+            int phasetoadd,     //!< index of current phase
+            int numdofpernode,  //!< total number of DOFs/phases
+            const PoroFluidManager::PhaseManagerInterface& phasemanager,  //!< phase manager
+            const PoroFluidManager::VariableManagerInterface<nsd, nen>&
+                variablemanager,  //!< variable manager
+            double timefacfac,    //!< domain-integration factor
+            double fac            //!< domain-integration factor times time-integration factor
+            ) override;
+
+        //! cast to VarExp-function
```
c-p-schmidt, line 2347: How does this comment fit to the function?

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
```diff
@@ -1533,16 +1533,1203 @@ namespace Discret
...
+        [[nodiscard]] const Core::Utils::FunctionOfAnything& function(int functnum) const;
+
+        //! vector holding the functions to be integrated
```
c-p-schmidt, line 2350: You probably store the function id’s as it’s a vector of int’s. Maybe it would be better to actually really store the functions directly as indicated by the comment.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
```diff
@@ -1533,16 +1533,1203 @@ namespace Discret
...
+      };
+
+      /*----------------------------------------------------------------------*
+       * **********************************************************************
+       *----------------------------------------------------------------------*/
+      /*!
+      \brief helper class for reconstruction of flux
+
+      This class implements the linearization of the flux reconstruction matrix (L_2 projection).
+      Only the matrix! For RHS see class ReconstructFluxRHS.
```
c-p-schmidt, line 2364: Below it looks as if you are also implementing the method to evaluate and assemble the RHS, or is this just a dummy?
Same applies to the class below where you calculate the RHS according to the docu.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
```diff
@@ -1533,16 +1533,1203 @@ namespace Discret
...
+      /*----------------------------------------------------------------------*
+       * **********************************************************************
+       *----------------------------------------------------------------------*/
+      /*!
+      \brief class for evaluation of additional volume fraction terms in fluid equations
+      (instationary terms) into the element matrix (for models with additional porous network with
+      closing relation "homogenized_vasculature_tumor")
+
+
+      This class implements the term $( w, \frac{-\sum^volfrac \phi_volfrac) }{K_s} \frac{\partial
```
c-p-schmidt, line 2721: Nice! Something like this would also be cool for other classes in this file.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
```diff
@@ -1615,22 +2802,38 @@ namespace Discret
             double timefacfac,    //!< domain-integration factor
             double fac            //!< domain-integration factor times time-integration factor
             ) override;
+
+        //! get transient term for rhs and OD
```
c-p-schmidt, line 2806: Comment and function name only partially match (also below). Can you double check what is done and adapt either the comment, or function name accordingly?

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
```diff
@@ -3728,28 +4270,128 @@ void Discret::Elements::PoroFluidEvaluator::EvaluatorVolFracAddDivVelTerm<nsd,
...
+  const std::vector<double>& phi = *variablemanager.phinp();
+  const std::vector<double>& phidt = *variablemanager.phidtnp();
+  double pA = phi[0];
+  double pB = phi[numfluidphases];
+  double dpAdt = phidt[0];
+  double dpBdt = phidt[numfluidphases];
+  double ratio_pApB = phi[0] / phi[numfluidphases];
 
-  Core::LinAlg::Matrix<nsd, nsd> gridvelderiv(Core::LinAlg::Initialization::zero);
+  static Core::LinAlg::Matrix<nsd, nsd> gridvelderiv(Core::LinAlg::Initialization::zero);
```
c-p-schmidt: We usually do not use `static` anymore for such matrices.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
```diff
@@ -3795,43 +4518,269 @@ void Discret::Elements::PoroFluidEvaluator::EvaluatorVolFracAddInstatTermsSat<ns
...
+            mymat(fvi, fui) += vfunct;
+          }
+        }
+      }
+    }
+  }
+}
+
+/*----------------------------------------------------------------------*
+ | evaluate RHS vector                                 kremheller 09/17 |
```
c-p-schmidt: Please remove such unnecessary information, that soon gets outdated and might even be wrong. Git does that for us. ;-)

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
sebproell: This matrix should probably be a `Tensor`.

src/mat/4C_mat_fluidporo_singlephase.hpp
lkoeglmeier: What do you mean here exactly?

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
lkoeglmeier: As already discussed offline, this will be done in a follow-up PR, as this code already existed and was not changed in this pr.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
lkoeglmeier: same as above.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
lkoeglmeier: Yes, you are absolutely right. We could use tensors in many places in this file to avoid the less intuitive calculations. Since I would have to change several lines here to use a tensor, I would open a cleanup issue  (see #1293) to use tensors throughout the file, if this is ok for you?

tests/input_files/porofluid_pressure_based_elast_scatra_2D_quad4_nodetopoint_artery_airway_coupling_volfrac_closingrelation_lung_mono.4C.yaml
lkoeglmeier: Thanks. I included a vtk-test.

src/mat/4C_mat_fluidporo_multiphase.cpp
lkoeglmeier: Thanks for the hint. I changed it!

lkoeglmeier:
Thank you very much @c-p-schmidt for taking the time to review this long and pr! I have learned a lot from your suggestions!

src/mat/4C_mat_fluidporo_multiphase.cpp
```diff
@@ -86,14 +104,11 @@ void Mat::PAR::FluidPoroMultiPhase::initialize()
         FOUR_C_THROW(
             "You have chosen {} fluidphases, however your material number {} is no poro "
             "singlephase material\n"
-            "Your Input should always look like (for example: 4 fluid phases, 2 volume "
-            "fractions):\n"
-            "MAT 0 MAT_FluidPoroMultiPhase LOCAL No PERMEABILITY 1.0 NUMMAT 8 MATIDS    1 2 3 4 5 "
-            "6 7 8 NUMFLUIDPHASES_IN_MULTIPHASEPORESPACE 4 END\n"
+            "Your Input should always look like (for example: 4 fluid phases \n"
+            "MAT 0 MAT_FluidPoroMultiPhase LOCAL No PERMEABILITY 1.0 NUMMAT 8 MATIDS    1 2 3 4 "
+            " NUMFLUIDPHASES_IN_MULTIPHASEPORESPACE 4 END\n"
```
sebproell: Outdated. We use YAML now.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
c-p-schmidt: As discussed offline, I think it is reasonable to do this in a follow-up as you suggested.

src/mat/4C_mat_fluidporo_multiphase.cpp
```diff
@@ -113,38 +128,74 @@ void Mat::PAR::FluidPoroMultiPhase::initialize()
...
+              "You have chosen {} fluid phases and {} volume fractions, however your material "
+              "number "
+              "{} is no poro volume fraction material\n"
+              "Your Input should always look like (for example: 4 fluid phases, 1 volume "
+              "fraction):\n"
+              "MAT 0 MAT_FluidPoroMultiPhase LOCAL No PERMEABILITY 1.0 NUMMAT 8 MATIDS    1 2 3 4 "
+              "5  NUMFLUIDPHASES_IN_MULTIPHASEPORESPACE 4 END\n"
+              "with: 4 fluid phases in multiphase pore space: materials have to be "
+              "MAT_FluidPoroSinglePhase \n"
+              "     1 volume fractions: materials have to be MAT_FluidPoroVolFracPressureBloodLung",
```
sebproell: Also here. And a few more times.

src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_input.hpp
```diff
@@ -36,6 +36,13 @@ namespace PoroPressureBased
     twoway_monolithic
   };
 
+
+  enum class FluidporoReactionCoupling
+  {
+    scalar_by_function,
+    no_coupling
+  };
```
sebproell, line 44: Docs missing.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
```diff
@@ -2888,22 +2904,1052 @@ namespace Discret
...
+                variablemanager,  //!< variable manager
+            double timefacfac,    //!< domain-integration factor
+            double fac            //!< domain-integration factor times time-integration factor
+            ) override;
+      };
+
+
+      /*----------------------------------------------------------------------*
+       * **********************************************************************
+       *----------------------------------------------------------------------*/
```
sebproell, line 3016:
```suggestion
```

lkoeglmeier: Thanks a lot for your review @sebproell!  Is there a reason why you chose this particular separator? I use it in several places and personally find it quite useful for better readability.

sebproell: I personally use blank lines since I find these headers very noisy. In hpp files, I often don't find separators necessary, since you should ideally have a nice doc block above every symbol ;) In cpp files, I tend to just use 2-3 empty lines to separate function bodies.

src/mat/4C_mat_fluidporo_multiphase.cpp
sebproell: Sorry, I could have been clearer, I guess. I would remove the whole comment, since this just gets outdated all the time. Instead, can you state what needs to happen logically, irrespective of the exact input specification? You may leave it as is; just something to consider in the future.

lkoeglmeier: Ah, you are right, of course! I changed it! Thanks!

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
lkoeglmeier: Thank you very much for sharing your opinion with us. However, I would prefer to keep them, at least for the moment.

PR #1119, opened 2025-08-18
Remove unused code from partitioned FSI

mayrmt:
## Description and Context

This PR removes some unused and untested code paths from the partitioned FSI solvers. ~At its core, it deletes the `NOX::FSI::LinearSystemGCR` class.~

- Related to findings in PR #1098


maxfirmbach:
Seems like one test is using something of what has been deleted. I can't comment on if it is worth to keep this test however.

mayrmt:
I checked the coverage report before deletion. It did not show this.

mayrmt:
This has to be investigated in detail. For now, this PR has been changed to not touch the `NOX::FSI::LinearSystemGCR`, but only removes truly unused code paths.

PR #1123, opened 2025-08-18
Switch ordering of blocks in block systems

c-p-schmidt:
## Description and Context
While working on removing the AMGnxn within the SSI module, we found together with @maxfirmbach that the ordering of dofs within the blocks of the block system needs to be ascending. This was not the case and has been changed within this PR as a preparation for the switch to Teko (which internally uses Thyra, where this requirement seems to originate from).

part of #754

### DIFF
--- a/src/ssi/4C_ssi_utils.cpp
+++ b/src/ssi/4C_ssi_utils.cpp
@@ -619,7 +619,7 @@ SSI::Utils::SSIMaps::SSIMaps(const SsiMono& ssi_mono_algorithm)
 
 /*--------------------------------------------------------------------------------------*
  *--------------------------------------------------------------------------------------*/
-std::vector<int> SSI::Utils::SSIMaps::get_block_positions(Subproblem subproblem) const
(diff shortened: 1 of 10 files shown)


mayrmt:
@c-p-schmidt Is the assumed ordering of the fields in the matrix documented somewhere?

c-p-schmidt:
> @c-p-schmidt Is the assumed ordering of the fields in the matrix documented somewhere?

Good point. Do you mean in SSI? Or are you rather referring to the requirement from Teko that the dof gids in the blocks of block systems need to be ascending, i.e.
Block 1 dofgids 1 to m
Block 2 dofgids m+1 to n
...

Which should be documented in our TekoPreconditioner class, I suppose. What do you think, @maxfirmbach?

For SSI, I will do it in a follow-up, where I actually introduce the usage of Teko. For Teko itself, I'd suggest putting a note in the documentation of the TekoPreconditioner class that such an ordering is necessary for Teko to work properly.

maxfirmbach:
I think this is a thing that should be documented in general for block matrices, as this is how it should be. A warning/requirement statement in `Teko` makes sense too.

PR #1128, opened 2025-08-19
CI: own schema validation script for input files

sebproell:
Replaces the action with our own script, which works for larger schema files and is much faster. Should help to proceed with #1092.


gilrrei (dismissed):

.github/workflows/buildtest.yml
```diff
@@ -232,17 +232,13 @@ jobs:
...
-          yaml_extension: .4C.yaml
-          yaml_extension_short: .4C.yml
-          json_extension: .4C.json
+      - name: Validate all input files against schema
+        run: |
+          cd $GITHUB_WORKSPACE
+          ${{ github.workspace }}/utilities/set_up_dev_env.sh
+          source ${{ github.workspace }}/utilities/python-venv/bin/activate
+          shopt -s globstar
+          python ${{ github.workspace }}/utilities/validate_schema.py --schema ${{ github.workspace }}/4C_schema.json  **/*.4C.yaml **/*.4C.yml **/*.4C.json
```
sebproell: @ppraegla My shell hackery is not working and not pretty. Should I do this in python?

.github/workflows/buildtest.yml
```diff
@@ -232,17 +232,13 @@ jobs:
...
-          json_schema_version: draft-2020-12
-          yaml_extension: .4C.yaml
-          yaml_extension_short: .4C.yml
-          json_extension: .4C.json
+      - name: Validate all input files against schema
+        run: |
+          cd $GITHUB_WORKSPACE
+          ${{ github.workspace }}/utilities/set_up_dev_env.sh
+          source ${{ github.workspace }}/utilities/python-venv/bin/activate
+          shopt -s globstar
```
amgebauer:
```suggestion
          shopt -s globstar nullglob
```

amgebauer (changes requested):

utilities/validate_schema.py
```diff
@@ -0,0 +1,72 @@
...
+    for yaml_file in args.yaml_files:
+        try:
+            data = load_yaml(yaml_file)
+            validator.validate(data)
+        except jsonschema_rs.ValidationError as e:
+            print(f"Validation error in {yaml_file}: {e}")
+            total_errors += 1
+        except Exception as e:
+            print(f"An error occurred while validating {yaml_file}: {e}")
+            total_errors += 1
```
amgebauer: Personally, I would make the output a little more verbose, similar to the `ctest --output-on-failure` option:

```suggestion
        try:
            data = load_yaml(yaml_file)
            validator.validate(data)
            print(format_dotted(yaml_file, "passed"))
        except jsonschema_rs.ValidationError as e:
            print(f"Validation error in {yaml_file}: {e}")
            print(format_dotted(yaml_file, "failed"))
            print(f"{e}")
            
            files_with_errors.append(yaml_file)
        except Exception as e:
            print(f"Validation error in {yaml_file}: {e}")
            print(format_dotted(yaml_file, "error"))
            print(f"{e}")
            
            files_with_errors.append(yaml_file)
      
```

with

```
def format_dotted(left: str, right: str) -> str:
    dots = '.' * (300 - len(left) - len(right))
    return f"{left}{dots}{right}"
```

sebproell:
@amgebauer Thanks for the hints!

tests/input_files/sohex8_distributed-pointcoupling.4C.yaml
```diff
@@ -19,7 +19,7 @@ STRUCTURAL DYNAMIC:
   DYNAMICTYPE: "Statics"
   RESTARTEVERY: 1000
   NUMSTEP: 2
-  TIMESTEP: 5.
+  TIMESTEP: 5.0
```
sebproell: Note that a trailing decimal separator causes issues when using our `yaml->json` strategy for extremely fast reading of files. I just converted the two offending cases. We now check for this in CI, so we will immediately see if we get problems with it. However, it is still something that might be worth fixing in rapidyaml.

@gilrrei Additionally, I could also add a regex-replace next to the `inf` case. WDYT?

ppraegla (approved):

tests/input_files/sohex8_distributed-pointcoupling.4C.yaml
gilrrei: In the specification of [YAML tags](https://yaml.org/spec/1.2.2/#24-tags) they only give these examples:

```yaml
canonical: 1.23015e+3
exponential: 12.3015e+02
fixed: 1230.15
negative infinity: -.inf
not a number: .nan
```

I am not sure if yaml is suppose to read `5.` as a float in the first place. We should check the specification and if it is supposed to be interpreted as string, then the schema validation should definitely fail as it currently is. If not your regex fix seems fine.

sebproell: It is allowed in the [core schema](https://yaml.org/spec/1.2.2/#core-schema):

`[-+]? ( \. [0-9]+ | [0-9]+ ( \. [0-9]* )? ) ( [eE] [-+]? [0-9]+ )?`

vs. JSON

`-? [1-9] ( \. [0-9]* [1-9] )? ( e [-+] [1-9] [0-9]* )?`

amgebauer: Interestingly, the json regex would also not allow `5.0`, but it works obviously

utilities/validate_schema.py
```diff
@@ -0,0 +1,88 @@
...
+
+    files_with_errors = []
+    for yaml_file in args.yaml_files:
+        try:
+            data = load_yaml(yaml_file)
+            validator.validate(data)
+            print(format_dotted(yaml_file, "passed"))
+
+        except jsonschema_rs.ValidationError as e:
+            print(f"Validation error in {yaml_file}: {e}")
```
amgebauer:
```suggestion
```

utilities/validate_schema.py
```diff
@@ -0,0 +1,88 @@
...
+            print(format_dotted(yaml_file, "failed"))
+            print(f"{e}")
+
+            files_with_errors.append(yaml_file)
+        except Exception as e:
+            print(f"Validation error in {yaml_file}: {e}")
```
amgebauer:
```suggestion
```

tests/input_files/sohex8_distributed-pointcoupling.4C.yaml
sebproell: True. Since it seems more complicated to exclude the case of a single trailing zero than just accept it, I assume most json parsers accept it. I can also imagine that there is disagreement over not allowing this in the first place, and in practice, people will write numbers like `5.0`, I think. Looks very natural to me.

amgebauer (approved):

tests/input_files/sohex8_distributed-pointcoupling.4C.yaml
gilrrei: Then let's go with the regex approach. I think this problem would disappear if we add a yaml formatter that would add the trailing zero.

gilrrei (approved):

PR #1130, opened 2025-08-19
Enable teko for elch problems

c-p-schmidt:
## Description and Context
As part of #754, this PR enables the usage of Teko for electrochemistry problems. The AMG-BGS preconditioning is covered by MueLu directly, whereas the BGS-AMG preconditioning is performed with Teko.
However, at the moment, the AMGnxn implementation is not yet deleted for the elch module, as the dof-splitting, i.e., assigning the different dofs of nodes to different blocks and thereby violating the Teko requirement that dof gids have to be ascending for blocks with increasing block ids, is not yet possible to cover within Teko. For these tests, the AMGnxn is kept for the moment, and we will have to work on a different solution for that.

part of #754 
follows #1123 #1124


src/core/linear_solver/src/method/4C_linear_solver_method_linalg.cpp
```diff
@@ -291,6 +293,18 @@ Teuchos::ParameterList translate_four_c_to_belos(const Teuchos::ParameterList& i
...
+    Teuchos::ParameterList belos_parameters;
+    Teuchos::updateParametersFromXmlFileAndBroadcast(xml_file_name,
+        Teuchos::Ptr<Teuchos::ParameterList>(&belos_parameters),
+        *Xpetra::toXpetra(Core::Communication::as_epetra_comm(comm)));
+
+    // required for adaptive linear solver tolerance
+    beloslist.set("Convergence Tolerance",
+        belos_parameters.sublist("GMRES").get<double>("Convergence Tolerance"));
+    beloslist.set("Implicit Residual Scaling",
+        belos_parameters.sublist("GMRES").get<std::string>("Implicit Residual Scaling"));
```
maxfirmbach, line 307: At some point we have to do the whole linear solver adaptivity in a different way, but not part of this PR of course.

c-p-schmidt:
> At some point we also have to rethink this whole parameterlist business ... :sweat_smile:.

I totally agree, I had countless headaches the recent days concerning that 😮‍💨

PR #1131, opened 2025-08-20
[CMake] Unclutter installation and document it

sebproell:
Removes one of the places (former `4CSettings.cmake.in`) that duplicate information about dependencies. While there, I realized that we had no docs for installing and using 4C separately, so I added them.

Part of #618


amgebauer (approved):

sebproell:
This is ready for review.

PR #1133, opened 2025-08-20
Print field name in input field error

maxiludwig:
This pull requests adds the option to give the name of the input field when calling the `at()` method. This name is printed in the error message when no value can be found for a index, improving error tracing, especially, if multiple `input_fields` are used.

### DIFF
--- a/src/core/io/src/4C_io_input_field.hpp
+++ b/src/core/io/src/4C_io_input_field.hpp
@@ -167,13 +167,19 @@ namespace Core::IO
 
     /**
      * Access the value of the field for the given @p element index. If the @p element_id
-     * is not a valid index, this function will throw an error.
+     * is not a valid index, this function will throw an error which contains the optional @p
+     * field_name for informational purposes.
      */
-    [[nodiscard]] const T& at(IndexType element_id) const { return get(element_id, true); }
+    [[nodiscard]] const T& at(
+        IndexType element_id, std::string_view field_name = "unknown field") const
+    {
+      return get(element_id, true, field_name);
+    }
 
    private:
     //! Internal getter which can optionally check for the validity of the element index.
-    const T& get(IndexType element_id, bool check) const
+    const T& get(
+        IndexType element_id, bool check, std::string_view field_name = "unknown field") const
     {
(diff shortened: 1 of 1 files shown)


src/core/io/src/4C_io_input_field.hpp
```diff
@@ -169,11 +169,14 @@ namespace Core::IO
      * Access the value of the field for the given @p element index. If the @p element_id
      * is not a valid index, this function will throw an error.
      */
-    [[nodiscard]] const T& at(IndexType element_id) const { return get(element_id, true); }
+    [[nodiscard]] const T& at(IndexType element_id, std::string_view field_name = "") const
```
sebproell: Please document the new parameter.

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -168,12 +168,18 @@ namespace Core::IO
     /**
      * Access the value of the field for the given @p element index. If the @p element_id
      * is not a valid index, this function will throw an error.
+     * @param element_id The index of the element for which to get the value.
+     * @param field_name Optional name of the field for error messages. This is particularly useful
+     * if multiple input fields are used.
```
sebproell: I am not against this checklist style per se, but in my experience, documentation for short functions reads a lot better if it is written in full sentences. So, how about you add after the original text: 


```suggestion
     * which contains the optional @p field_name for informational purposes.
```

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -168,12 +168,18 @@ namespace Core::IO
...
      */
-    [[nodiscard]] const T& at(IndexType element_id) const { return get(element_id, true); }
+    [[nodiscard]] const T& at(IndexType element_id, std::string_view field_name = "") const
```
sebproell: Sorry for only realizing this now: 
Not printing a name inside the '' '' in the error message might be confusing; thus, the default value should likely be something like `"unknown field"`, or you make the parameter non-defaulted.

sebproell (approved):

Issue #1073, opened 2025-08-11
[CMake] Consistently name our configuration options


sebproell:
Our CMake setup is in a good state. I claim this based on the fact that it did not change a lot in the past year (and not in any breaking way). It is time to clean up a few relics. My goal is to make [this table](https://4c-multiphysics.github.io/4C/documentation/developer_guide/developer_cmake.html#reference-of-all-cmake-variables) more useful and consistent.

- [x] Name all options that enable a dependency `FOUR_C_WITH_<dependency>`
- [x] Name all options that enable a feature `FOUR_C_ENABLE_<feature>` (e.g. `FOUR_C_ENABLE_DOCUMENTATION`)
- [x] Properly document `FOUR_C_CXX_FLAGS` in the doc table
- [x] Remove deprecated names after the next release

*Do these steps backwards-compatible with warnings for the wrong variable names.*

sebproell:
This is done, except that we should remove the deprecated names after the next release.

PR #1135, opened 2025-08-20
CMake: document FOUR_C_CXX_ variables

sebproell:
Part of #1073

I decided to remove the `FOUR_C_CXX_DEBUG_FLAGS` etc. since they aren't useful if we don't do multi-configuration builds. Technically, a breaking change, but I doubt that anybody uses these.


ppraegla (approved):
Do you want to apply the breaking change label anyway? This also helps for generating the changelog for a release

sebproell:
This is ready for review.

amgebauer (approved):

PR #1136, opened 2025-08-20
Add pattern validator for input strings

sebproell:
Allows specifying a regex pattern for strings. I don't anticipate a lot of uses for this case, but it is exactly what we need for validating the version number within #74. It is trivial to add the pattern in a general way, so I went for this option rather than special hackery for certain fields.

### DIFF
--- a/src/inpar/4C_inpar_scatra.cpp
+++ b/src/inpar/4C_inpar_scatra.cpp
@@ -166,9 +166,12 @@ void Inpar::ScaTra::set_valid_parameters(std::map<std::string, Core::IO::InputSp
                   .default_value = true}),
 
           parameter<std::string>("WRITEFLUX_IDS",
-              {.description = "Write diffusive/total flux vector fields for these scalar "
-                              "fields only (starting with 1)",
-                  .default_value = "-1"}),
+              {
+                  .description = "Write diffusive/total flux vector fields for these scalar "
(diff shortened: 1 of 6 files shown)


src/inpar/4C_inpar_scatra.cpp
```diff
@@ -166,9 +166,12 @@ void Inpar::ScaTra::set_valid_parameters(std::map<std::string, Core::IO::InputSp
...
 
           parameter<std::string>("WRITEFLUX_IDS",
-              {.description = "Write diffusive/total flux vector fields for these scalar "
-                              "fields only (starting with 1)",
-                  .default_value = "-1"}),
+              {
+                  .description = "Write diffusive/total flux vector fields for these scalar "
+                                 "fields only (starting with 1)",
+                  .default_value = "-1",
+                  .validator = Validators::pattern(R"(^(-?\d+)(\s+-?\d+)*$)"),
```
sebproell, line 173: This is not a fix for this parameter (one should use a vector<int>), but it allows us to test that the metadata and schema work. I will add the `input_version` next, which will also use this new validator.

sebproell:
@mayrmt Well, this PR is useful regardless of any outcome in #74. It may have been inspired by #74, but it does not implement #74. So, I think we can proceed with this PR?

PR #1137, opened 2025-08-20
Introduce an optional input version

sebproell:
Introduces a top-level key `input_version` and records the respective version using a SemVer-style triple. For now, this is not doing much, only providing the infrastructure and printing a warning if no version is given or it is "wrong".

### Points to discuss

These do not need to be resolved in this PR, but some related questions are:

- Should the behavior of what happens if no or a wrong version is given be configurable on the command line? 
- Do we like the keyword `input_version` and want to use a SemVer triple? (For schemas, a triple is great, since `1.0.0` cannot be mistaken as a float.)
- Should we then roll out `input_version: 1.0.0` to all our test input files?

For anyone who prefers not to include `input_version`, could you share specific concerns with this proposal?

Depends on #1136 

First step for #74

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -1761,12 +1761,7 @@ bool Core::IO::Internal::ParameterSpec<T>::match(ConstYamlNodeRef node,
 {
   auto spec_name = ryml::to_csubstr(name);
 
-  // If we are not even in a map, we refuse to do anything and let the MatchTree handle this case.
-  // Setting a default would confuse the user, since something fundamental must be wrong in the
-  // input file.
-  if (!node.node.is_map()) return false;
-
-  if (!node.node.has_child(spec_name))
+  if (!node.node.is_map() || !node.node.has_child(spec_name))
   {
     // It is OK to not encounter an optional parameter
     if (data.default_value.index() == 1)
@@ -2029,12 +2024,7 @@ bool Core::IO::Internal::DeprecatedSelectionSpec<T>::match(ConstYamlNodeRef node
 {
   auto spec_name = ryml::to_csubstr(name);
 
-  // If we are not even in a map, we refuse to do anything and let the MatchTree handle this case.
-  // Setting a default would confuse the user, since something fundamental must be wrong in the
-  // input file.
-  if (!node.node.is_map()) return false;
-
-  if (!node.node.has_child(spec_name))
+  if (!node.node.is_map() || !node.node.has_child(spec_name))
   {
     // It is OK to not encounter an optional parameter
     if (data.default_value.index() == 1)


mayrmt (changes requested):
Until the discussion in #74 has been resolved, I would not proceed with this PR.

mayrmt:
Given the open discussion in $74, this PR cannot proceed now. Review are thus not necessary at the moment. Hence, I'll mark it as draft.

sebproell:
> Given the open discussion in https://github.com/4C-multiphysics/4C/issues/74, I think we should not move forward here, until we have a common understanding of a way forward.

I agree that we should not move forward on the open points of 

- how to upgrade files (technical solution that is easy and user-friendly)
- mass-adapt all parameters
- interplay of 4C and FourCIPP

Since none of this is part of this PR, we are good to go here, I think.

input_version/INPUT_VERSION
```diff
@@ -0,0 +1 @@
+1.0.0
```
amgebauer, line 1: I don't like having an extra directory for the input version. Can we even put this into `VERSION` and specify the 4C and the input version there? (I raise this point, but I'm not super convinced whether it is superior than separating it into two files)

```yaml
version: 2025.3.0-dev
input_version: 1.0.0
```

ppraegla:
> I agree that we should not move forward on the open points

Introducing the input file version itself seems to be also an open point, judging from the comments in #74. So, we should wait with this PR until after the proposed meeting in #74.

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -474,17 +504,40 @@ namespace Core::IO
...
+      match_section(input_version_key, version_data);
+
+      if (auto opt_version = version_data.get<std::optional<std::string>>(input_version_key);
+          opt_version.has_value())
       {
-        if (!pimpl_->is_section_known(section_name))
+        if (*opt_version != VersionControl::input_version)
+          std::cout << std::format(
+              "Warning: version mismatch: "
+              "The input file requires input version {}, but the input version of 4C is {}.\n",
```
amgebauer:
```suggestion
              "The input file requires input version {}, but the input version of 4C is {}. Your input file may be incompatible with 4C.\n",
```

CMakeLists.txt
```diff
@@ -69,6 +69,23 @@ set(FOUR_C_VERSION_PATCH ${CMAKE_MATCH_3})
 
 message(STATUS "Configuring 4C version ${FOUR_C_VERSION_FULL}")
 
+# The input version is used to determine the compatibility of input files.
+# Syntax: major.minor
```
amgebauer:
```suggestion
# Syntax: major.minor.patch
```

CMakeLists.txt
```diff
@@ -69,6 +69,23 @@ set(FOUR_C_VERSION_PATCH ${CMAKE_MATCH_3})
...
+  )
+string(REGEX MATCH "([0-9]+)\\.([0-9]+)\\.([0-9]+)" _dummy ${FOUR_C_INPUT_VERSION_FULL})
+if(NOT _dummy)
+  message(
+    FATAL_ERROR
+      "The input version string ${FOUR_C_INPUT_VERSION_FULL} does not match the expected format MAJOR.MINOR.PATCH"
+    )
+endif()
+set(FOUR_C_INPUT_VERSION_MAJOR ${CMAKE_MATCH_1})
+set(FOUR_C_INPUT_VERSION_MINOR ${CMAKE_MATCH_2})
```
amgebauer, line 86:
```suggestion
set(FOUR_C_INPUT_VERSION_MINOR ${CMAKE_MATCH_2})
set(FOUR_C_INPUT_VERSION_PATCH ${CMAKE_MATCH_3})
```

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -474,17 +504,40 @@ namespace Core::IO
...
+          std::cout << "Compatible input file version " << *opt_version << " detected.\n";
         }
       }
+      else
+      {
+        std::cout << std::format(
+            "Warning: Your input file does not specify an input version and may "
+            "be incompatible with the input version {} of 4C.\n",
+            VersionControl::input_version);
+      }
```
mayrmt: Throwing a warning, that the optional specification of an input version is missing, does not make it optional anymore, since it implies that one has to add an input version. That's the reason, why I find this PR to be part of the discussion in #74.

mayrmt (commented):
What's the benefit of this change? I can put a number into the input file and it is compared to a number somewhere in the 4C configuration and I get a warning if

- I do not put the number into the input file
- the two numbers do not match

I do not see, how this helps with anything. Am I missing something?

Again, there are many open questions in #74. Before they are answered, I do not see, how we can move forward towards any meaningful l improvements.

src/core/io/src/4C_io_input_file.cpp
sebproell: As I said in the description, we could make the printouts configurable on the command line. I wouldn't say the warning forces you to do anything; it is just factual information you are free to ignore. This is what we have been doing implicitly. Now it is explicit.

Side note: #937 and proper logging would make such status output nicer.

isteinbrecher: An optional feature should not cause a warning if it is not enabled.

sebproell:
@mayrmt Correct, as the description states, this PR does not provide much apart from the basic input version. It paves the way to implement (and discuss!) the actual features that many of us want _afterwards_. We are free to modify details of the version (e.g. where we store it) along the way, once we have more knowledge of the technical details.

isteinbrecher:
I agree with @ppraegla that we have to wait with any of the proposed advances until the proposed meeting and a commonly agree decision on how to proceed.

input_version/INPUT_VERSION
sebproell: - I would not merge the two version files, since this makes it hard to read out the version. Currently, these are UNIX-style files with info that is readily interpreted.
- I am open to placing this file somewhere different. I placed it here, since this was a suggestion in #74.

src/core/io/src/4C_io_input_file.cpp
sebproell: I think my proposal is fine, but let's not get caught up in the discussion. I am fine with removing the `else` branch for now. Ofc, any warning will be more useful once the version is bumped and we have established tools that help you to act on the warning.

mayrmt: If there is the need for a discussion, then we have to have the discussion. Since this PR has been announced as a place for discussion, let's discuss.

sebproell: I removed the warning for now. 

> let's discuss.

Does this mean you want me to restore the warning, but in a different form? Please tell me which part of this thread is not resolved.

isteinbrecher:
This is not yet ready for review, thus marking as draft.

amgebauer (approved):

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -474,17 +504,34 @@ namespace Core::IO
...
+      if (auto opt_version = version_data.get<std::optional<std::string>>(input_version_key);
+          opt_version.has_value())
       {
-        if (!pimpl_->is_section_known(section_name))
+        if (*opt_version != VersionControl::input_version)
+          std::cout << std::format(
+              "Warning: version mismatch: "
+              "The input file requires input version {}, but the input version of 4C is {}. "
+              "Your input file may be incompatible with 4C.\n",
+              *opt_version, VersionControl::input_version);
```
ppraegla, line 521: Do you want to mention that the input version is experimental?
```suggestion
              "Warning: version mismatch for experimental input version: "
              "The input file requires input version {}, but the input version of 4C is {}. "
              "Your input file may be incompatible with 4C.\n",
              *opt_version, VersionControl::input_version);
```

ppraegla (approved):
IMO, you can proceed as long as the input version remains purely optional so that nobody who does not care is affected (This is the case with the current state of the PR). 

For your open questions
> Should the behavior of what happens if no or a wrong version is given be configurable on the command line?

I see no need to introduce a commandline option at this stage.

> Should we then roll out input_version: 1.0.0 to all our test input files?

No. To test the input version one/few input file is enough. This also allows you to easily increment the version without creating too much noise for everyone by touching every input file.

isteinbrecher:
https://github.com/4C-multiphysics/4C/pull/1137#issuecomment-3210665338

src/core/io/src/4C_io_input_file.cpp
sebproell: I am not against it, but the version is already optional, and you are not going to introduce it into your file, and thus won't see this output unless you are beta-testing or developing this feature.

gilrrei:
Our recent behavior has been childish, myself included. I apologize for the bluntness, but our current behavior doesn't reflect our standards and the way we normally interact. Besides, it negatively impacts the community, which I know nobody wants. Let's remember that **we are all passionate nerds who love to code and care deeply about this project**. We've collaborated successfully in the past, and that's the spirit we want to maintain. We were literally drinking beers together some weeks ago! It seems there are some unresolved issues beyond just introducing version numbers. So, let’s cool down, enjoy our hopefully sunny weekends, and come back next week ready to get back on track, **together**.

isteinbrecher:
I agree, we all want what is best for 4C. Also, I agree that we should work on a common solution **together**. So let's schedule a meeting as proposed in #74.

davidrudlstorfer (approved):

PR #1138, opened 2025-08-21
Adjust path for vtk files on restart

knarfnitram:
## Description and Context

Adjust the path of the `VtkWriterBase` such that Paraview can now display a restarted simulation correctly (independent if  the restart path was specified absolute or relative).

Detailed description of the issue can be found in non-ported issue: 
https://gitlab.lrz.de/baci/baci/-/issues/1309

### DIFF
--- a/src/core/io/src/4C_io_vtk_writer_base.cpp
+++ b/src/core/io/src/4C_io_vtk_writer_base.cpp
@@ -724,22 +724,22 @@ void VtkWriterBase::create_restarted_initial_collection_file_mid_section(
 
       if (readtime <= (restart_time + 1e-12))
       {
-        std::string filename = get_xml_option_value(line, "file");
         std::filesystem::path p_restart(restartfilename);
+        std::filesystem::path p_filename(get_xml_option_value(line, "file"));
 
-        if (p_restart.is_absolute())
-        {
-          // Do not modify the path if the restart is given with an absolute path (we don't want to
-          // have absolute paths in the collection file)
-          write_master_file_and_time_value_into_given_vtk_collection_file_stream(
-              collection_file_midsection_cumulated_content_, filename, readtime);
-        }
-        else
-        {
-          std::filesystem::path p_new_filename = "." / p_restart.parent_path() / filename;
-          write_master_file_and_time_value_into_given_vtk_collection_file_stream(
-              collection_file_midsection_cumulated_content_, p_new_filename.string(), readtime);
-        }
+        // Choose base directory: on absolute restart select its parent directory
+        // otherwise keep relative with ".")
+        const std::filesystem::path base_dir =
+            p_restart.is_absolute() ? p_restart.parent_path()
+                                    : (std::filesystem::path(".") / p_restart.parent_path());
+
+        // Resolve final path: select absolute file paths, otherwise anchor at base directory
+        const std::filesystem::path resolved =
+            p_filename.is_absolute() ? p_filename : (base_dir / p_filename);
+
+        write_master_file_and_time_value_into_given_vtk_collection_file_stream(
+            collection_file_midsection_cumulated_content_,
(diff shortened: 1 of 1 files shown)


knarfnitram:
So I am not 100% familiar with the `four_c_testing_functions.cmake`, but I think we can currently restart only from the same folder, so i think we'd need to rewrite a custom cmake function for that.

c-p-schmidt (commented):
I'm also fine with the changes; however, I agree that a test would be good to ensure it won't break.

knarfnitram:
So far the restart function from a different folder has not been tested, hence the introduced test function creates two subfolders and performs  a restart within `simulation_2` from `simulation_1` given the absolute path. 
As last the `vtk-files` are compared within the second folder to ensure that  the paths are set correctly.

By dropping the first commit, I verified that the changes are really covered by the introduced testing function.

davidrudlstorfer (commented):
Just one question, isn't it possible to just extend the regular `four_c_test` function? This one already supports restarts and I thought it should be doable to add a new variable with the path to the restart

knarfnitram:
@davidrudlstorfer I had a look at it this afternoon and yes it can be done, but from my point of view it just complicates things, because you have to adapt `four_c_test_vtk` as well and then with the changing folder names it gets messy. 
So since this is a relative special test case, I want to tackle here, I'd be in favor of merging it with this separate test function.

tests/list_of_tests.cmake
```diff
@@ -2150,4 +2150,12 @@ four_c_test_vtk(structure_solid_beam_element_gid_runtime_output-vtk-solid.4C.yam
...
 four_c_test_vtk(thermo3D_annulus_elementwise_anisotropic_statics_vtk_thermo.4C.yaml thermo3D_annulus_elementwise_anisotropic_statics.4C.yaml 3 xxx-thermo.pvd ref/thermo3D_annulus_elementwise_anisotropic_statics-thermo.pvd 1e-08)
 four_c_test_vtk(beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_line3_vtk_lambda.4C.yaml beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_line3.4C.yaml 3 xxx-beam-to-solid-volume-mortar-continuous.pvd ref/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_line3-beam-to-solid-volume-mortar-continuous.pvd 1e-08)
+four_c_test_restarted_vtk(TESTNAME correct-restarted-vtk-files-for-solid_runtime_tet4
+        FILE solid_runtime_tet4.4C.yaml
+        RESTART_STEP 1
+        PVD_RESULTFILENAME xxx-structure.pvd
+        PVD_REFERENCEFILENAME ref/solid_runtime_tet4-vtk/structure.pvd
+        TOLERANCE 1e-06
+        TIMEOUT 10
+        TIME_STEPS 0.5)
```
davidrudlstorfer: Can we put this into one line like all other tests? As far as I know we do not have any maximum line length setup so this should work.

Let's stay consistent here.

davidrudlstorfer (approved):
> @davidrudlstorfer I had a look at it this afternoon and yes it can be done, but from my point of view it just complicates things, because you have to adapt four_c_test_vtk as well and then with the changing folder names it gets messy.
> So since this is a relative special test case, I want to tackle here, I'd be in favor of merging it with this separate test function.

@knarfnitram thanks for looking into it. I do not want to block any progress here so let's just stay with your current approach.

Would you be able to comment this decision and the reason for it in #56? Maybe also add that in the future these could be combined.

Let's get this merged once the test passes ✅

davidrudlstorfer (approved):

PR #1139, opened 2025-08-21
Use json input field for muscle activation and support std::pair as an InputSpec type

lauraengelhardt:
## Description and Context
We now prescribe muscular activation in one of our skeletal muscle material models using JSON input fields instead of the old *pattern files (second commit).

Since activations are defined as a vector of (time, activation) pairs, the InputSpecs have been extended to support std::pair as a supported type. This includes extending the schema generation functionalities and YAML read/write toolings to support std::pair (first commit). 

CC: @sebproell @gilrrei


src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -22,6 +22,8 @@
 #include "4C_utils_string.hpp"
 #include "4C_utils_symbolic_expression.hpp"
 
+#include <Sacado_tradvec.hpp>
```
sebproell:
```suggestion
```

utilities/metadata_utils.py
```diff
@@ -128,6 +128,22 @@ def __post_init__(self):
             self.value_type = _metadata_object_from_dict(self.value_type)
 
 
+@dataclass
+class Pair(Primitive):
+    """Pair parameter."""
+
+    size: int = None
```
sebproell: We know a pair has two items, so this field seems unnecessary.

src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -211,6 +229,20 @@ namespace Core::IO
...
 
+    template <typename T1, typename T2>
+    struct YamlTypeEmitter<std::pair<T1, T2>>
+    {
+      void operator()(ryml::NodeRef node, size_t*)
+      {
+        FOUR_C_ASSERT(node.is_map(), "Expected a map node.");
+        node["type"] = "pair";
+        node["first_type"] |= ryml::MAP;
+        YamlTypeEmitter<T1>{}(node["first_type"], nullptr);
```
sebproell: That's not quite correct. Depending on if pair itself is "sized" it should either be


```suggestion
        YamlTypeEmitter<T1>{}(node["first_type"], size + 1);
```

or

```suggestion
        YamlTypeEmitter<T1>{}(node["first_type"], size);
```

utilities/create_json_schema.py
```diff
@@ -286,6 +287,31 @@ def array_schema(parameter, items):
...
+        items={
+            "first": get_schema(pair.first_type),
+            "second": get_schema(pair.second_type),
+        },
+        maxItems=2,
+        minItems=2,
+        noneable=pair.noneable,
+    )
+
+    return schema
```
gilrrei: Pairs do not exist in the JSON schema language. Instead, we have to use arrays or tuple in order to validate each entry differently, see [here](https://json-schema.org/understanding-json-schema/reference/array#tupleValidation):
```json
{  "type": "array",  "prefixItems": [    { "type": "number" },    { "type": "string" },    { "enum": ["Street", "Avenue", "Boulevard"] },    { "enum": ["NW", "NE", "SW", "SE"] }  ]}
```
So I  think the concrete changes are removing `maxItems` and `minItems` and changing to `prefixItems` instead of `items`

lauraengelhardt: Perfect, thanks for the link!

lauraengelhardt:
I have moved all things related to the std::pair support to https://github.com/4C-multiphysics/4C/pull/1148 (first commit with the requested changes).
Once #1148 is merged, I'll update this PR.

lauraengelhardt:
With #1148 merged, I updated this PR. It is now ready for review.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -507,11 +484,10 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 parameter<int>("FUNCTID",
                     {.description =
                             "function id for time- and space-dependency of muscle activation"}),
-                parameter<std::filesystem::path>("MAPFILE",
+                input_field<std::vector<std::pair<double, double>>>("MAP",
```
amgebauer: Is `MAP` still a good name? The input field also supports being constant. If I see it correctly, you don't need the `ACTEVALTYPE` anymore and just call the parameters in the `one_of` as `ACTIVATION_FUNCTION_ID` and `ACTIVATION`

lauraengelhardt: Good point - I wanted to get rid of `ACTEVALTYPE` in a separate PR anyhow (and use the InputSpec Validators) :smiley: 
I updated the PR as you suggested.

PR #1132, opened 2025-08-20
Fix nightly-pipline by

knarfnitram:
## Description and Context

Try to fix the nightly pipeline by 
- removing failing test case `beam3r_herm2line3_static_contact_penalty_with_positional_coupling.yaml` for now
- move failing test `beam3r_herm2line3_static_beam_to_solid_surface_contact_mortar_line2_gap_variation_reference_segmentation_penalty_lin_quad_point_coupling.4C.yaml` into arborx testing scope

## Related Issues and Pull Requests

Related to #1125
Closes #1125 #1120

PR #1140, opened 2025-08-21
Restore the beam interaction with beam coupling test

knarfnitram:
## Description and Context

Restores and adapts the test case which has been removed in  #1132 . 
The `FE_Overflow` within the clang pipeline was caused by a the regularization parameter close to `0`. 
The test case uses now slightly more elements and another value for the regularization parameter. 
Additionally, the penalty law's parameter check is enhanced to catch values numerically close to `0`. 

Related to #1132 and #1086


src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -1527,19 +1527,18 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::calc_pena
     }
     case BeamContact::pl_lpqp:  // quadratic regularization for positive gaps
     {
-      if (g0 == -1.0)
+      if (g0 == -1.0 || abs(g0) < 1e-16)
```
maxfirmbach: Shouldn't we use something like `std::numeric_limits<T>::epsilon()`?

maxfirmbach: Also `g0 == -1.0` doesn't look that great ...

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -1527,19 +1527,18 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::calc_pena
...
       double gbar = params()->beam_to_beam_contact_params()->gap_shift();
       gap = gap + gbar;
-
       double f0 = g0 * pp / 2.0;
-      double factor_a = pp / (g0)-f0 / (g0 * g0);  //=pp/(2*g0)
       double factor_b = -pp;
       double factor_c = f0;
       if (gap > 0)
       {
+        double factor_a = pp / (g0)-f0 / (g0 * g0);  //=pp/(2*g0)
```
maxfirmbach: Is the comment really correct? The way how the brackets are set is weird especially `(g0)`.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
knarfnitram: thanks for the `std::numeric_limits`, I was searching for that. 
`-1` is the initial value from the input file in case the parameter is not specified.
Both cases are now adapted.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
knarfnitram: I removed it.

mayrmt (approved):

PR #1145, opened 2025-08-22
Initialize more 2D-3D beam-to-solid pairs

isteinbrecher:
## Description and Context

Add more variants to the 2D-3D beam-to-solid volume mortar pairs. We can now create coupling pairs for hex8 and hex27 and up to 10 Fourier modes.

### DIFF
--- a/src/beaminteraction/src/4C_beaminteraction_beam_to_solid_volume_meshtying_pair_2d-3d_mortar.cpp
+++ b/src/beaminteraction/src/4C_beaminteraction_beam_to_solid_volume_meshtying_pair_2d-3d_mortar.cpp
@@ -600,38 +600,76 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPair2D3DMortar<Beam, Solid,
 /**
  *
  */
-std::shared_ptr<BeamInteraction::BeamContactPair>
+std::unique_ptr<BeamInteraction::BeamContactPair>
 BeamInteraction::create_beam_to_solid_volume_pair_mortar_cross_section(
     const Core::FE::CellType shape,
     const Inpar::BeamToSolid::BeamToSolidMortarShapefunctions mortar_shape_function,
     const int n_fourier_modes)
 {
-  switch (mortar_shape_function)
+  using namespace GeometryPair;
+
+  // Function to create the mortar pair for a given solid type
+  auto make_mortar_pair =
+      []<typename Solid>(Inpar::BeamToSolid::BeamToSolidMortarShapefunctions mortar_shape_function,
+          int n_fourier_modes) -> std::unique_ptr<BeamInteraction::BeamContactPair>
   {
-    case Inpar::BeamToSolid::BeamToSolidMortarShapefunctions::line2:
+    switch (mortar_shape_function)
     {
-      switch (n_fourier_modes)
-      {
-        case 0:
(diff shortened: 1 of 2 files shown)


src/beaminteraction/src/4C_beaminteraction_beam_to_solid_volume_meshtying_pair_2d-3d_mortar.hpp
```diff
@@ -70,36 +73,26 @@ namespace GeometryPair
...
-  template <>
-  struct IsFourierElement<t_line2_fourier_2>
-  {
-    static const bool value_ = true;
-  };
-  template <>
-  struct IsFourierElement<t_line2_fourier_3>
+  template <typename ElementType>
+  struct IsFourierElement<ElementType, std::void_t<typename ElementType::fourier_tag>>
+      : std::true_type
```
sebproell: Instead of SFINAE, you could use a concept roughly like this

```
template <typename ElementType>
concept FourierElement = requires {
   typename ElementType::fourier_tag;
}; 
```

vovannikov:
> Is there a way to reduce/replace the switch-case? Seems kind of repetitive.

Given that shared pointers are being created here, I guess a solution using `std::integer_sequance` could be used, something like https://stackoverflow.com/questions/68839163/is-it-impossible-to-pass-a-run-time-integer-as-a-template-argument.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_volume_meshtying_pair_2d-3d_mortar.cpp
```diff
@@ -606,32 +606,70 @@ BeamInteraction::create_beam_to_solid_volume_pair_mortar_cross_section(
...
+            return std::make_shared<BeamToSolidVolumeMeshtyingPair2D3DMortar<
+                GeometryPair::t_hermite, Solid, GeometryPair::t_line2_fourier_9>>();
+          case 10:
+            return std::make_shared<BeamToSolidVolumeMeshtyingPair2D3DMortar<
+                GeometryPair::t_hermite, Solid, GeometryPair::t_line2_fourier_10>>();
+          default:
+            FOUR_C_THROW("Got wrong number of fourier mortar modes {}.", n_fourier_modes);
+        }
+      default:
+        FOUR_C_THROW("Wrong mortar shape function.");
```
vovannikov: In general, it is considered to be a better practice for factory methods to return `std::unique_ptr`s instead of the shared ones (uptrs can be implicitly converted to the shared ), unless there is a strong reason to do the other way around (e.g., you create many small objects with relatively short lifetime, probably this is your case).

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_volume_meshtying_pair_2d-3d_mortar.hpp
isteinbrecher: Thanks for the hint, that is actually a very nice feature. I added this.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_volume_meshtying_pair_2d-3d_mortar.cpp
isteinbrecher: Makes sense, I changed the factory to a `unique_ptr`.

PR #1147, opened 2025-08-22
Consistently name CMake options

sebproell:
This PR 

- improves the `four_c_process_global_option` function to use argument parsing
- provides backward compatible warnings that an options is now named differently
- renames the options that are not in the `FOUR_C_ENABLE` style

Q: What would be a good time frame to remove the deprecated option names? Maybe leave it in the next release, and remove it after the one after that?

Addresses #1073


sebproell:
This is ready for review.

sebproell:
We could use release milestones and add the issue there.

PR #1148, opened 2025-08-22
Support std::pair as an InputSpec type

lauraengelhardt:
## Description and Context
Extend InputSpecs to support std::pair as a supported type. This includes extending the schema generation functionalities and YAML read/write toolings to support std::pair. The first commit adds a compile-time check in SizeChecker to catch missing template implementations for types where rank != 0. 

Preparation for #1139

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.hpp
+++ b/src/core/io/src/4C_io_input_spec_builders.hpp
@@ -26,6 +26,7 @@
 #include <functional>
 #include <optional>
 #include <ostream>
+#include <tuple>
 #include <variant>
 
 FOUR_C_NAMESPACE_OPEN
@@ -83,6 +84,21 @@ namespace Core::IO
           out << " ";
         }
       }
+
+      template <typename... Ts>
+      void operator()(std::ostream& out, const std::tuple<Ts...>& vals) const
+      {
+        std::apply([&](const auto&... vals) { (((*this)(out, vals), out << " "), ...); }, vals);
+      }
+
+      template <typename T1, typename T2>
+      void operator()(std::ostream& out, const std::pair<T1, T2>& val) const
+      {
+        (*this)(out, val.first);
+        out << " ";
+        (*this)(out, val.second);
+        out << " ";
+      }
     };
 
     template <typename T>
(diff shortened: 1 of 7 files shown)


src/core/io/src/4C_io_input_spec_builders.hpp
```diff
@@ -211,6 +229,20 @@ namespace Core::IO
...
+    struct YamlTypeEmitter<std::pair<T1, T2>>
+    {
+      void operator()(ryml::NodeRef node, size_t* size)
+      {
+        FOUR_C_ASSERT(node.is_map(), "Expected a map node.");
+        node["type"] = "pair";
+        node["first_type"] |= ryml::MAP;
+        YamlTypeEmitter<T1>{}(node["first_type"], size);
+        node["second_type"] |= ryml::MAP;
+        YamlTypeEmitter<T2>{}(node["second_type"], size);
```
sebproell: I wonder whether we will want to support tuples. If we do, we might want to change this here to `type: tuple` and then add a list of the types inside the tuple. `pair` is just a special case then. 

I can definitely see that `tuple` invites misuse, i.e., you combine a bunch of fields that should have separate names. However, I find it more consistent to allow a tuple if we allow a pair.

sebproell: Documenting discussion between @gilrrei, @lauraengelhardt, and me: we should write the metadata as if a pair were a tuple since the metadata and other tools do not know a pair anyway.

lauraengelhardt: With my current implementation, the schema section for a pair (treated as a tuple of size 2) looks like this:


```
"title": "(tuple)",
  "description": "No description yet.",
  "type": "array",
  "prefixItems": [
    {
      "title": "(double)",
      "description": "No description yet.",
      "type": "number"
    },
    {
      "title": "(double)",
      "description": "No description yet.",
      "type": "number"
    }
  ] 
```

and the corresponding metadata section like this:
```
type: tuple
  size: 2
  value_types:
    - type: double
    - type: double
```

What do you think? @gilrrei @sebproell

sebproell: Make sense! The indentation is off, though. I guess you mean:

```
type: tuple
size: 2
value_types:
  - type: double
  - type: double
```

lauraengelhardt: Yes, sorry, messed that up when copying!

src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -1127,6 +1127,88 @@ required: true
     }
   }
 
+  TEST(InputSpecTest, MatchYamlComplicatedParameterTuplePairMap)
+  {
+    using ComplicatedType =
+        std::tuple<int, std::pair<std::vector<double>, std::map<std::string, bool>>,
+            std::tuple<std::string, int, double>>;
```
sebproell, line 1134: That's complicated indeed! Nice that this works so generically.

sebproell (dismissed):
LGTM

src/core/io/src/4C_io_value_parser.hpp
```diff
@@ -247,7 +248,20 @@ namespace Core::IO
       }
     }
 
-    template <typename U, typename... SizeInfo>
+    template <typename... Ts>
+    void read_internal(std::tuple<Ts...>& value)
+    {
+      std::apply([this](auto&... val) { (read_internal(val), ...); }, value);
```
sebproell: @lauraengelhardt I think you run into [this bug](https://github.com/llvm/llvm-project/issues/36228) with clang. Try:


```suggestion
      std::apply([this](auto&... val) { (this->read_internal(val), ...); }, value);
```

gilrrei (approved):

src/core/io/src/4C_io_value_parser.hpp
lauraengelhardt: Thanks! 👍

PR #1151, opened 2025-08-23
[LinAlg] Refactor matrix complete call

### LINKED ISSUE #764: [Epetra] Remove calls to `epetra_matrix()`
## Description

There are quite a few calls to `epetra_matrix()` to call respective methods. Most of those should be already available in our wrapper class, thus we can remove `epetra_matrix()` as it pollutes our code. If a method is not available, we have to add it to our wrapper.

- [x] Remove as many calls as possible to `epetra_matrix()` outside of `core`

maxfirmbach:
## Description and Context

This PR refactors the matrix `complete()` call by introducing a struct for the different options to handle. This also removes some specific `epetra_matrix()->FillComplete(false)` calls. Also some unused `complete()` calls based on `Epetra` input variables are removed.

Closes #764.

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
@@ -67,6 +67,17 @@ namespace Core::LinAlg
                             each 'original' block is divided into 2 blocks. */
   };
 
+  //! \brief Options for matrix completion
+  struct OptionsMatrixComplete
+  {
+    //! enforce a lightweight fill_complete() even though the matrix might already have been filled
+    bool enforce_complete = false;
+
+    //! make consecutive row index sections contiguous, minimize internal storage used for
+    //! constructing graph
+    bool optimize_data_storage = true;
+  };
+
   /// Linear operator interface enhanced for use in FE simulations
   /*!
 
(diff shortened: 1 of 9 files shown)


src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
```diff
@@ -67,6 +67,19 @@ namespace Core::LinAlg
                             each 'original' block is divided into 2 blocks. */
   };
 
+  /*!
+\brief Options for matrix completion
+*/
```
mayrmt:
```suggestion
  /*!
   * \brief Options for matrix completion
   */
```
of just one line...

src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
```diff
@@ -67,6 +67,19 @@ namespace Core::LinAlg
...
+  struct OptionsMatrixComplete
+  {
+    //! enforce fill_complete() even though the matrix might already be filled
```
mayrmt: Not sure, if this comment already clearly conveys that an expensive operation is re-done.

maxfirmbach: Actually the `ExpertStaticFillComplete` is rather lightweight, as the matrix already has been `FillComplete` `ed once before.

src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
maxfirmbach: Done.

Issue #51, opened 2024-11-29
Rework namespaces to avoid ambiguous lookups


sebproell:
_Issue ported from Gitlab. Original issue [here](https://gitlab.lrz.de/baci/baci/-/issues/1160)_

## Description

As already described in !1910, we have some ambiguity in our namespaces. Take, e.g., the identifier `CONTACT`. This word exists as part of `FourC::CONTACT`, `FourC::NOX::NLN::CONTACT`, `INPAR::CONTACT`, and possibly others. The problem here is that 

- For a human, `CONTACT` right now is itself not very informative and the full path is necessary to identify the exact namespace. I think it is logical to use `CONTACT` only on the highest level when referring to general `CONTACT` functionality. 
- For the compiler, the lookup fails in this example

```c++
namespace FourC::NOX::NLN::CONTACT
{
  void f()
  {
     // This path is not enough as the compiler will look inside `FourC::NOX::NLN::CONTACT`
     // CONTACT::Algorithm algo;

     // The correct version reads:
     FourC::CONTACT::Algorithm algo;
     ...
  }
}
```

We should not have to add `FourC::` to code within BACI. Therefore, @ssinzig and me suggest to do the following: 

- Have top-level `FourC` via macro as done in !1906. This namespace is "invisible" inside BACI and never needs to be mentioned inside BACI code.

To achieve this:

## Possible Solution and Definition of Done

- [ ] "Reserve" the names of the second-level namespaces such as "CORE", "SOLID", "THERMO", ...: These names may not appear in any nested namespaces. This should resolve any ambiguity. (footnote 1) If cases would require arbitrary renaming just to achieve this effect (e.g. `CONTACT2`) then let's reconsider.
- [x] Remove "DRT" as this has no meaning. Replace with an appropriate name depending on case. (!1933)
- [x] In unit tests, use `using namespace FourC;`
- [ ] If we can remove `FourC::`: consider adding a pre-commit and CI check ensuring we do reintroduce `FourC::` in the code base.

## Related Issues and Merge Requests

* Follows !1906

## Additional Information

Footnote 1: Very technical comment: for a lookup to be unambiguous, no sub-namespace may have the same name as any sibling of its parent. For deeper namespace levels, this rule would also hold, but it becomes less and less likely to hit ambiguities the further down you descend into the namespace abyss.

We believe it is acceptable to reserve the use of certain namespaces ("CORE", "SOLID", "THERMO", ...) as these are the names of 4C's tbd modules.

[...]

PR #1159, opened 2025-08-25
Introduce new reduced_lung input

maxiludwig:
This PR adds a new input mechanism for the `reduced_lung` module. Employing the new `input_field` feature, we can selectively compose airway and terminal unit models from different sub-models. So far, the new input is only used for the Terminal Units. It will be extended to the Airways soon.


src/reduced_lung/4C_reduced_lung_main.cpp
```diff
@@ -83,9 +93,42 @@ namespace ReducedLung
...
+            {
+              add_terminal_unit_ele<FourElementMaxwell, LinearElasticity>(
+                  terminal_units, ele, local_element_id, tu_params);
+            }
+            else if (elasticity_model_name == IO::ElasticityModel::Ogden)
+            {
+              add_terminal_unit_ele<FourElementMaxwell, OgdenHyperelasticity>(
+                  terminal_units, ele, local_element_id, tu_params);
+            }
+          }
```
maxiludwig: This can probably be done much easier but I'm not sure how,  @sebproell @bgoderbauer

src/reduced_lung/4C_reduced_lung_input.cpp
```diff
@@ -0,0 +1,150 @@
...
+                          group("VISCOELASTIC_RLC",
+                              {input_field<double>("GAMMA",
+                                   {.description = "Van Ertbruggen's generation dependent "
+                                                   "turbulence factor defining turbulent onset."}),
+                                  group("WALL_MODEL",
+                                      {
+                                          input_field<IO::WallModel>("WALL_MODEL_TYPE",
+                                              {.description = "Type of wall model for the "
+                                                              "viscoelastic airway."}),
+                                          group("STANDARD_WALL_MODEL",
```
maxiludwig: Do we have a better name for this model? @bgoderbauer

src/reduced_lung/4C_reduced_lung_input.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "4C_io_input_spec.hpp"
+
+#include <map>
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace ReducedLung
+{
+  namespace IO
```
sebproell: According to our plans in #51, please don't use such generic names as nested namespaces. Do you need another namespace inside `ReducedLung` due to name clashes?

src/reduced_lung/4C_reduced_lung_input.cpp
```diff
@@ -0,0 +1,150 @@
...
+      {
+          group("DYNAMICS",
+              {
+                  parameter<IO::ReducedLungDyn>(
+                      "DYNAMIC_TYPE", {.description = "Time integration scheme.",
+                                          .default_value = IO::ReducedLungDyn::BackwardEuler}),
+                  parameter<double>("TIMESTEP", {.description = "Time increment dt."}),
+                  parameter<int>("NUM_STEP", {.description = "Number of time steps."}),
+                  parameter<int>("RESTART_EVERY",
+                      {.description = "Increment for writing restart.", .default_value = 1}),
```
sebproell: "increment" or "number of steps"?

src/reduced_lung/4C_reduced_lung_input.cpp
```diff
@@ -0,0 +1,150 @@
...
+              },
+              {.required = true}),
+
+          group("LUNG_TREE",
+              {
+                  input_field<IO::ElementType>(
+                      "ELEMENT_TYPE", {.description = "Type of reduced lung elements."}),
+                  input_field<int>(
+                      "GENERATION", {.description = "Generation of the airway elements"}),
+                  group("AIRWAYS",
```
maxiludwig: Both, `AIRWAYS` and `TERMINAL_UNITS` are required groups. This leads to the fact that we need to specify both even if we have a tree without TUs like in the test case. But this is rather a test case and not our classical use-case, so i would opt to leave both groups required.

sebproell (commented):
Nicely substructured input! I think you could directly store your parameters into a struct to remove the duplication of key names.

src/reduced_lung/4C_reduced_lung_input.hpp
maxiludwig: Yes, my intention was to separate all these enum definition from the model structs.

sebproell: Would `Parameters` work?

maxiludwig: If this is already sufficiently non-generic, yes :sweat_smile:  :+1:

sebproell: I find it at least somewhat more specific than `IO`

src/reduced_lung/4C_reduced_lung_main.cpp
```diff
@@ -40,9 +42,17 @@ namespace ReducedLung
...
-    const Teuchos::ParameterList& rawdyn =
-        Global::Problem::instance()->reduced_d_airway_dynamic_params();
-    const int linear_solver_number = rawdyn.get<int>("LINEAR_SOLVER");
+    const Teuchos::ParameterList& redlung_params =
+        Global::Problem::instance()->reduced_d_lung_params();
+
+    const Teuchos::ParameterList& dyn_params = redlung_params.sublist("DYNAMICS");
+    const Teuchos::ParameterList& tu_params =
+        redlung_params.sublist("LUNG_TREE").sublist("TERMINAL_UNITS");
+    const Teuchos::ParameterList& [[maybe_unused]] aw_params =
```
sebproell:
```suggestion
    [[maybe_unused]] const Teuchos::ParameterList& aw_params =
```

If you use structs to store the parameters, this will become unnecessary.

src/reduced_lung/4C_reduced_lung_input.cpp
bgoderbauer: We could also name it after a rheological model? Just as an idea:

- `rigid` without compliance
- `kelvin_voigt`: Current model default with linear elasticity and linear viscosity (which can also be set to zero for a purely elastic case). 
- ...

Then we can compose the models again based on a flow model (linear resistive, all non-linear resistive cases, resistive + inertia) in the axial direction and the rheological model for wall viscoelasticity. This would be very similar to the current Terminal Unit composition. The only difference comes from the fact that each model has its own equation in the airway (or none in the case of a rigid wall). They're not merged into one equation like the terminal unit models.

src/reduced_lung/4C_reduced_lung_input.hpp
```diff
@@ -0,0 +1,86 @@
...
+  {
+    /**
+     * Enum class for the reduced lung dynamics.
+     */
+    enum class ReducedLungDyn
+    {
+      BackwardEuler,
+    };
```
bgoderbauer: Should we keep this? We have a lot of different sets of algebraic equations, and at least I don't see a benefit to specifying this globally for all equations, especially when there is only one type of discretization available. IMO, the discretization scheme is very dependent on the model (for example, for fractional models, we will likely use exponential integrator schemes), so I would rather hide this internally.

src/reduced_lung/4C_reduced_lung_main.cpp
bgoderbauer: I think we can write a function doing this in the terminal unit file, then it at least isn't in the main file. At some point, we need to loop over the possible combinations. As far as I understand, your solution, a switch statement, or a visitor similar to the model generation in the terminal unit file, are all good ways to do this. Maybe we will find a way to merge this with the model generation process, but it's fine for me as it is 😃.

src/reduced_lung/4C_reduced_lung_input.cpp
```diff
@@ -0,0 +1,150 @@
...
+                                                      {.description = "Viscous time constant"}),
+                                                  input_field<double>("VISCOUS_PHASE_SHIFT",
+                                                      {.description = "Viscous phase shift"}),
+                                              },
+                                              {.description = "TODO find a name and description",
+                                                  .required = false}),
+                                      })},
+                              {.description = "Wall model of the viscoelasitic airway",
+                                  .required = false})}),
+                  group("TERMINAL_UNITS",
```
bgoderbauer: The reference volume is only created from the line element length as equivalent radius, right?

bgoderbauer:
I really like this new input a lot :heart_eyes:. This is so much nicer than before, and you can finally understand how it is working!

src/reduced_lung/4C_reduced_lung_input.cpp
maxiludwig: I would then propose the following hierarchy. What do you think? 
```
airway/
│
├── flow_model/
│   ├── resistance_type/ 
│   │   ├── linear
│   │   └── non_linear
│   │
│   ├── resistance_model/
│   │   ├── linear/
│   │   │   └── (no parameters)
│   │   └── non_linear_resistance/
│   │       └── gamma (double)
│   │
│   └── inertia (bool)
│
├── wall_model_type/
│   ├── rigid
│   └── kelvin_voigt
│
└── wall_model/
    ├── rigid/
    │   └── (no parameters)
    │
    └── kelvin_voigt/
        ├── elasticity/
        │   ├── poisson_ratio (double)
        │   ├── wall_elasticity (double)
        │   └── wall_thickness (double)
        │
        └── viscosity/
            ├── time_constant (double)
            └── phase_shift (double)
```
The `_type` parameters would be input fields of enums. So that we can have different models in different parts of the AW tree

bgoderbauer: Looks good :+1:. Let's go ahead with this. If we think of new things in the future, we can easily change things. :smile:

src/reduced_lung/4C_reduced_lung_input.cpp
maxiludwig: Exactly

src/reduced_lung/4C_reduced_lung_main.cpp
maxiludwig: I would then leave it as it is for now :)

bgoderbauer (approved):

src/reduced_lung/4C_reduced_lung_input.cpp
```diff
@@ -0,0 +1,417 @@
...
+                                                                  viscosity_kelvin_voigt_eta),
+                                              }),
+                                      },
+                                      {
+                                          .description = "Kelvin-Voigt model of the terminal unit.",
+                                          .required = false,
+                                          .store = in_struct(&ReducedLungParameters::LungTree::
+                                                  TerminalUnits::RheologicalModel::kelvin_voigt),
+                                      }),
+                                  group<ReducedLungParameters::LungTree::TerminalUnits::
```
sebproell: FYI: When the nesting gets too crazy, you can always write an inner spec before defining the outer spec and just insert it into the outer spec.

maxiludwig:
I just pulled the inner specs for the airways and terminal unit models out as suggested by @sebproell

bgoderbauer (approved):

sebproell:
@maxiludwig Could you rebase instead of merging `main` into your branch?

bgoderbauer (approved):

PR #1161, opened 2025-08-25
Minor cleanup in Knotvectors

sebproell:
The `KNOTVECTORS` sections are legacy sections that should become nested YAML. While looking through the code to see how this can be achieved, I performed a little cleanup.

### DIFF
--- a/src/core/fem/src/nurbs_discretization/4C_fem_nurbs_discretization_knotvector.cpp
+++ b/src/core/fem/src/nurbs_discretization/4C_fem_nurbs_discretization_knotvector.cpp
@@ -79,34 +79,6 @@ Core::FE::Nurbs::Knotvector::Knotvector(const int dim, const int npatches)
   return;
 }  // Core::FE::Nurbs::Knotvector::Knotvector (standard)
 
-/*----------------------------------------------------------------------*
- |  copy ctor (public)                                       gammi 05/08|
- *----------------------------------------------------------------------*/
-Core::FE::Nurbs::Knotvector::Knotvector(const Core::FE::Nurbs::Knotvector& old)
-    : ParObject(old),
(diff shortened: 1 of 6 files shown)


src/core/fem/src/nurbs_discretization/4C_fem_nurbs_discretization_knotvector.cpp
```diff
@@ -1078,8 +1030,8 @@ void Core::FE::Nurbs::Knotvector::print(std::ostream& os) const
     for (int direction = 0; direction < dim_; ++direction)
     {
       os << "  direction " << direction << ": ";
-      for (std::size_t i = 0; i < knot_values_[patch][direction]->size(); ++i)
-        os << (*(knot_values_[patch][direction]))[i] << ", ";
+      for (std::size_t i = 0; i < knot_values_[patch][direction].size(); ++i)
+        os << ((knot_values_[patch][direction]))[i] << ", ";
```
isteinbrecher, line 1034: Are the double `((` needed here?

isteinbrecher (approved):
What is the plan for the yaml representation of the knot vectors? The data format currently used in FourCIPP?

sebproell:
@isteinbrecher 

> What is the plan for the yaml representation of the knot vectors? The data format currently used in FourCIPP?

Yes, I would try to use that format and see if we like the outcome.

src/core/fem/src/nurbs_discretization/4C_fem_nurbs_discretization_knotvector.cpp
sebproell: No, it's not needed. If I need to touch this PR again, I can fix it.

PR #1163, opened 2025-08-25
Improve beaminteraction checks

isteinbrecher:
## Description and Context

#1086 introduces a nice check for the compatibility of beam-to-beam point couplings and beam-to-beam contact pairs. This check is extended in such a way, that we don't allow point couplings between two beams that are potentially in contact with each other. Point couplings between two beams that are possibly in contact with other beams are OK.

### DIFF
--- a/src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
+++ b/src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
@@ -342,24 +342,24 @@ void Solid::ModelEvaluator::BeamInteraction::set_sub_model_types()
 
   // Ensure that no point coupling condition connects two beams that are possibly in contact with
   // each other.
-  for (auto bpcc : beampenaltycouplingconditions)
+  for (auto beam_point_coupling_condition : beampenaltycouplingconditions)
   {
     // get all nodes of beam to beam point coupling condition
-    const std::vector<int>& nodes = *(bpcc->get_nodes());
+    const std::vector<int>& nodes = *(beam_point_coupling_condition->get_nodes());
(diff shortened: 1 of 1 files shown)


knarfnitram (approved):

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
```diff
@@ -340,28 +340,26 @@ void Solid::ModelEvaluator::BeamInteraction::set_sub_model_types()
...
-          bpcc->print(std::cout);
-          FOUR_C_THROW(
-              "It is not possible to use beam-to-beam contact in combination with "
-              "beam-to-beam point coupling for the same node. Please reconsider the specified "
-              "interaction conditions with the ids {} and {}, since they are intersecting at the "
-              "node with id "
-              "{}",
-              btbcc->id(), bpcc->id(), node);
-        }
+        btbcc->print(std::cout);
```
maxfirmbach: Do we want this plain print-out? What is `btbcc` a coupling pair? How does the output look like? Do we print the two ID's?

knarfnitram: Yes, in case of overlapping nodes, I decided to print the condition names and ids, such that you get immediately a good idea, which conditions needs to be adapted.

isteinbrecher: Good point, I think the `btbcc->print(std::cout);` can be removed as we print the ID in the error anyway. I also renamed the variables to a more verbose name. Should be ready now.

PR #1164, opened 2025-08-25
Add logical operators to symbolic expressions

sebproell:
In order to implement https://github.com/4C-multiphysics/4C/issues/571#issuecomment-3214966458 but also as a general extension, this PR adds comparison operators (`>`, `>-=`, `<`, `<=`, `==`, `!=`) and logical operators (`!`, `&&` and `||`) to the SymbolicExpression.

I think this basically makes the `heaviside` function that was implemented before obsolete, because theses conditions are more powerful and readable. Compare e.g.

```
f(x) = 2*H(3 - x) + (2 + (x - 3))*(H(x - 3) - H(x - 7)) + 6*H(x - 7)
```

vs.

```
f(x) = (x < 3)*2 + ((x >= 3) && (x < 7))*(2 + (x - 3)) + (x >= 7)*6
```

### DIFF
--- a/src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
+++ b/src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
@@ -67,13 +67,14 @@ namespace Core::Utils::SymbolicExpressionDetails
     sqrt,
     heaviside,
     fabs,
+    logical_not,
   };
 
   template <typename T>
   using UnaryFunction = T (*)(const T&);
 
   template <typename T>
-  constexpr std::array<UnaryFunction<T>, 15> unary_function_table = {
+  constexpr std::array<UnaryFunction<T>, 16> unary_function_table = {
       [](const T& arg) -> T { return acos(arg); },
       [](const T& arg) -> T { return asin(arg); },
       [](const T& arg) -> T { return atan(arg); },
@@ -89,6 +90,7 @@ namespace Core::Utils::SymbolicExpressionDetails
       [](const T& arg) -> T { return sqrt(arg); },
       [](const T& arg) -> T { return arg > 0 ? 1.0 : 0.0; },
       [](const T& arg) -> T { return fabs(arg); },
+      [](const T& arg) -> T { return arg == 0.0 ? 1.0 : 0.0; },
   };
 
   enum class BinaryFunctionType : std::uint8_t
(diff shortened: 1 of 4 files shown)


sebproell:
I will also add `>=` and `<=`.

src/core/utils/tests/functions/4C_utils_symbolic_expression_test.cpp
```diff
@@ -229,6 +229,42 @@ namespace
...
+        symbolicexpression.value(Core::Utils::var<"x">(0.0), Core::Utils::var<"y">(0.0)), 1.0);
+    EXPECT_DOUBLE_EQ(
+        symbolicexpression.value(Core::Utils::var<"x">(0.0), Core::Utils::var<"y">(1.0)), 0.0);
+  }
+
+  TEST(SymbolicExpressionTest, LogicalOperators)
+  {
+    // Note that || has lower precedence than &&
+    Core::Utils::SymbolicExpression<double, "x", "y"> symbolicexpression(
+        "(x > 1.0 - 1.0) && (y > sin(0)) || (x <= !2) && !(y > 0)");
```
amgebauer, line 265: What does `!2` mean?

src/core/utils/src/functions/4C_utils_symbolic_expression_details.hpp
```diff
@@ -247,12 +266,26 @@ namespace Core::Utils::SymbolicExpressionDetails
...
+    //! A comparison expression.
+    IndexType parse_comparison(Lexer& lexer);
+    //! Addition and subtraction.
+    IndexType parse_term(Lexer& lexer);
+    //! Multiplication and division.
+    IndexType parse_factor(Lexer& lexer);
+    //! Power operator.
+    IndexType parse_pow(Lexer& lexer);
+    //! Number, variable, parentheses, unary operators, function call.
+    IndexType parse_primary(Lexer& lexer);
```
sebproell: > And where is the order of operations defined? 0.0 > x * y && x > x+1 is probably (0.0 > (x * y)) && (x > (x+1)), but I don't find this in the code 🤔

Good question! What we have implemented so far is a handcoded recursive descent parser and you can find the precedence of operations by starting at `parse` and then traversing the recursive calls.  In the marked section, I have ordered these functions from low to high precedence.

Related: I already thought about changing the internals to something known as a "Pratt parser" which abstracts away the duplicated code you might have noticed in many of the `parse_` functions and explicitly spells out precedence.

src/core/utils/tests/functions/4C_utils_symbolic_expression_test.cpp
sebproell: The same thing it means in C or C++, i.e, `0`.

amgebauer (approved):

gilrrei:
@sebproell, not saying it's easy, but _if_ we can find a regex string to check the syntax, we could, in theory, use that in a validator and for the schemas. The same holds for the function definition in `FUNT x`

Not sure it's worth it

sebproell:
> not saying it's easy, but if we can find a regex string to check the syntax, 

@gilrrei I love this question and hope I can answer correctly: a regular expression in the purest sense can only match a regular grammar, however, the expressions we have here are from a context-free grammar (as I believe, is typical for programming languages) (see also https://en.wikipedia.org/wiki/Chomsky_hierarchy) I belive it is provable that no such regex exists. A practical argument is the fact that you need balanced parenthese which regex cannot handle.

PR #1166, opened 2025-08-26
Fix: python root directory on bruteforce

davidrudlstorfer:
## Description and Context

Fix the CMake variable to find the correct python root dir on bruteforce


sebproell (approved):

davidrudlstorfer:
https://gitlab.lrz.de/baci/baci-config-lnm/-/pipelines/2685419 works as expected

PR #1167, opened 2025-08-26
Clean up implementation of Lexer and Parser in symexpr

sebproell:
I improved the backend of SymbolicExpression a lot in #913 and #916. This PR improves the frontend (lexing and parsing) with a more idiomatic implementation with better error reporting on lexical and syntactical errors. This PR is probably hard to review, since parsing tends to be recursive and all over the place, and I changed some internals. The good thing is that we have an excessive test suite for this functionality which makes these changes possible. 

The main user-facing change is that you now get nice errors like

```
Error while parsing:
2*4 - (3 + 1
           ^
           Expected closing parenthesis.
```


src/core/utils/src/functions/4C_utils_symbolic_expression.cpp
```diff
@@ -31,237 +23,593 @@ namespace Core::Utils::SymbolicExpressionDetails
...
+        return create_node(NodeType::binary_function, {.binary_function = BinaryFunctionType::mul},
+            prefactor, rhs);
+      }
+    }
+    else if (match(tok_name))
+    {
+      auto token = previous();
+      if (token.text == "pi")
+      {
+        return create_node(NodeType::number, {.number = M_PI});
```
amgebauer, line 481: It's more standard
```suggestion
        return create_node(NodeType::number, {.number = std::numbers::pi_v<double>});
```

amgebauer (dismissed):

src/core/utils/src/functions/4C_utils_symbolic_expression.cpp
sebproell: Good idea, but let's do it everywhere -> #1168

sebproell:
Can I get a second look here?

c-p-schmidt (approved):

Issue #973, opened 2025-07-08
Add testing workflow with fourcipp to ensure compatible 4C metadata file


davidrudlstorfer:
## Description

In the last few weeks/months changes in 4C unintentionally broke fourcipp and all downstream projects (4C-webviewer, beamme(meshpy), ...).

We should add a simple workflow which checks out the latest fourcipp version, uses the metadata file which is created during each PR and run the fourcipp tests.

This way we can directly see if some processes break.

One thing we should think about is what happens if a change requires changes in fourcipp because then we could end up in a dead lock.

## Possible Solution and Definition of Done

- [ ] Create a workflow which
    - [ ] checks out the latest fourcipp version
    - [ ] downloads the current metadata file for each PR
    - [ ] runs the base test suite of fourcipp

## Interested Parties

@sebproell @gilrrei

mayrmt:
@davidrudlstorfer I am not sure, if we should test a downstream package in 4C to ensure compatibility with all downstream packages (similar to you concerns that downstream changes might be necessary to avoid a dead lock). I understand, that such breaking changes are very very annoying, but testing across projects -- in particular including downstream tests in to an upstream project -- raises some concerns for me.

davidrudlstorfer:
> @davidrudlstorfer I am not sure, if we should test a downstream package in 4C to ensure compatibility with all downstream packages (similar to you concerns that downstream changes might be necessary to avoid a dead lock). I understand, that such breaking changes are very very annoying, but testing across projects -- in particular including downstream tests in to an upstream project -- raises some concerns for me.

@mayrmt unfortunately I do not really understand your points.

1. Which concerns does it raise for you, unfortunately you did not add them.

2. I do think there are a lot of solutions do avoid such a dead lock - for example we could simply allow this test to "fail" in PR's (so it is not a requirement to merge). This way we see and know that something fails but this does not hinder progress for PR's in 4C. Additionally, the pipeline would take a few seconds, so the computational effort is minimal as well compared to all of our other workflows.

3. https://github.com/4C-multiphysics/website/pull/22 here we recently discussed that fourcipp is a part of the 4C software stack - therefore, we should also ensure that the software stack works together.

If there are some valid points why adding such tests is a bad idea I am curious about them and I am totally open to dismiss this idea:)

mayrmt:
[...]

PR #1171, opened 2025-08-27
Use nested YAML for KNOTVECTORS sections

sebproell:
Use nicely nested YAML instead of strings. I tried to make this close to what FourCIPP uses internally, but it does not match perfectly.


gilrrei:
@sebproell matching the fourcipp structure is not important, once it moves to the nonlegacy sections, it works automagically.

Btw, kudos for the work and motivation to improve/fix less-used features. Great community spirit :people_hugging:

tests/input_files/thermo3D_meshtying_nurbs.4C.yaml
```diff
@@ -493,73 +493,72 @@ DVOL-NODE TOPOLOGY:
...
-  - "+1.000000000000"
-  - "END NURBSPATCH"
+  NURBS_DIMENSION: 3
+  PATCHES:
+    - ID: 1
+      KNOT_VECTORS:
+        - DEGREE: 2
+          TYPE: Interpolated
+          KNOTS:
+            - +0.000000000000
```
gilrrei: We could remove the `+`, it's not super common

tests/input_files/thermo3D_meshtying_nurbs.4C.yaml
```diff
@@ -493,73 +493,72 @@ DVOL-NODE TOPOLOGY:
...
-  - "DEGREE 2"
-  - "TYPE Interpolated"
-  - "+0.000000000000"
-  - "+0.000000000000"
-  - "+0.000000000000"
-  - "+1.000000000000"
-  - "+1.000000000000"
-  - "+1.000000000000"
-  - "END NURBSPATCH"
+  NURBS_DIMENSION: 3
```
isteinbrecher: This should be moved inside of patches, i.e., next to the keys "ID" and "KNOT_VECTORS". I am not 100% sure if we currently can handle different dimensions, but I think we should allow for this from the input side and if necessary fix it on the implementation side.

src/core/fem/src/nurbs_discretization/4C_fem_nurbs_discretization_knotvector.cpp
```diff
@@ -1038,4 +1029,62 @@ void Core::FE::Nurbs::Knotvector::print(std::ostream& os) const
...
+
+  for (unsigned int p = 0; p < patches.size(); ++p)
+  {
+    const auto& patch = patches[p];
+    const auto& id = patch.get<int>("ID");
+
+    const auto& knot_vectors = patch.get_list("KNOT_VECTORS");
+
+    FOUR_C_ASSERT_ALWAYS(knot_vectors.size() == dim,
+        "Every patch needs as many knot vector entries as the NURBS_DIMENSION (={})", dim);
```
eulovi: Thanks for taking care of this! Here, I would rather write:
"Every patch should have as many knot vectors as the NURBS_DIMENSION (={})".

tests/input_files/thermo3D_meshtying_nurbs.4C.yaml
eulovi: I tried to run a simulation with a NURBS surface and volume and this didn't work. In my older version of 4C, the error comes from the reading of the knot vector type in `4C_fem_nurbs_discretization_knotvector.cpp`.

src/core/fem/src/nurbs_discretization/4C_fem_nurbs_discretization_knotvector.cpp
```diff
@@ -1038,4 +1029,62 @@ void Core::FE::Nurbs::Knotvector::print(std::ostream& os) const
...
+                  })),
+          })),
+  });
+}
+
+
+Core::FE::Nurbs::Knotvector Core::FE::Nurbs::Knotvector::from_input(
+    const Core::IO::InputParameterContainer& data)
+{
+  unsigned dim = data.get<int>("NURBS_DIMENSION");
```
isteinbrecher: I think we should not specify a fixed NURBS dimension in the input file, but instead allow patches to have different dimensions.

The dimension of each NURBS patch should simply be determined by the number of knot vectors it contains.

It looks like we currently have some issues when reading patches of different sizes, but in principle the code itself should handle this. For now we should simply throw an error when inconsistent patch dimensions are found (i.e., not all have the same dimension) - but avoid having a global NURBS dimension.

@eulovi and myself can have a look on how to then get that working in the code in the future (should be a simple fix).

tests/input_files/thermo3D_meshtying_nurbs.4C.yaml
```diff
@@ -493,73 +493,72 @@ DVOL-NODE TOPOLOGY:
...
+  PATCHES:
```
isteinbrecher, line 496: Thanks for using the same data structure @gilrrei and I came up with for FourCIPP, this should make the adaption effort for BeamMe minimal.

sebproell:
Here is the vibe-coded script I used to upgrade the existing KNOTVECTORS sections: 
[update_knotvectors.py](https://github.com/user-attachments/files/22042144/update_knotvectors.py)

tests/input_files/solid_ele_nurbs27_Standard_Static_multielement.4C.yaml
```diff
@@ -30,43 +30,42 @@ STRUCTURAL DYNAMIC:
...
-  - "0.0"
-  - "0.25"
-  - "0.5"
-  - "0.75"
-  - "1.0"
-  - "1.0"
-  - "1.0"
-  - "END NURBSPATCH"
+  PATCHES:
+    - KNOT_VECTORS:
```
gilrrei, line 34: If the `NURBS_DIMENSION` is not required anymore, is this key necessary?

gilrrei: Nevermind, I misread

src/core/fem/src/nurbs_discretization/4C_fem_nurbs_discretization_knotvector.cpp
sebproell: Removed since the check is now don earlier.

tests/input_files/beam3r_herm2line3_static_beam_to_solid_surface_coupling_gpts_reference_forced_FAD_nurbs.4C.yaml
```diff
@@ -702,75 +702,73 @@ DVOL-NODE TOPOLOGY:
...
-  - "TYPE Interpolated"
-  - "+0.000000000000"
-  - "+0.000000000000"
-  - "+0.000000000000"
-  - "+0.500000000000"
-  - "+1.000000000000"
-  - "+1.000000000000"
-  - "+1.000000000000"
-  - "END NURBSPATCH"
+  PATCHES:
```
isteinbrecher, line 705: Maybe one more comment/question, we now have `PATCHES` as a key in `STRUCTURE KNOTVECTORS`. As far as I can see there is no need to ever add anything else than that key. So we could remove it entirely (and rename `FIELD KNOTVECTORS` to `FIELD NURBS PATCHES`). I am not sure of all the consequences of such a change with the input spec. But my opinion on that is not very strong, I am also fine with the current approach.

sebproell: I would like to avoid more fundamental changes in this PR. NURBS input is quite brittle and easy to mess up and we need to gradually improve here. I have no strong opinion how this should be named. Feel free to change it to your liking with other NURBS enthusiasts.

isteinbrecher (approved):

davidrudlstorfer:
Just a side note from me; it could be beneficial to prepare the necessary changes in FourCIPP before merging this

Merges to "production" on a Friday afternoon are always tricky 😆

sebproell:
Good point, let's wait for the FourCIPP draft :+1: Who is going to prepare that?

isteinbrecher:
> Just a side note from me; it could be beneficial to prepare the necessary changes in FourCIPP before merging this

Good point, if that would be done today, I can update BeamMe on Monday.

sebproell:
Side note: this highlights that we might want to look into #973

isteinbrecher:
Changes in FourCIPP should be ready, https://github.com/4C-multiphysics/fourcipp/pull/92 so I think we can go ahead an merge this. If we manage to get the FourCIPP PR merged on the weekend (hopefully simply rerunning the pipeline), fixing BeamMe should be no problem on Monday (I will be on vacation starting Tuesday, so if we manage this on Monday would be nice).

PR #1172, opened 2025-08-27
Remove shared_ptr on Communicators

sebproell:
... and some unused functions in Communicators.

### DIFF
--- a/apps/global_full/4C_global_full_main.cpp
+++ b/apps/global_full/4C_global_full_main.cpp
@@ -287,7 +287,7 @@ int main(int argc, char* argv[])
   // Initialize our own singleton registry to ensure we clean up all singletons properly.
   Core::Utils::SingletonOwnerRegistry::ScopeGuard singleton_owner_guard{};
 
-  std::shared_ptr<Core::Communication::Communicators> communicators =
+  Core::Communication::Communicators communicators =
       Core::Communication::create_comm(std::vector<std::string>(argv, argv + argc));
 
   CommandlineArguments arguments{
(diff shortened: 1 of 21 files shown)


davidrudlstorfer (dismissed):

src/global_data/4C_global_data.hpp
```diff
@@ -200,10 +200,10 @@ namespace Global
     /// @name Communicators and their parallel groups
 
     /// set communicators
-    void set_communicators(std::shared_ptr<Core::Communication::Communicators> communicators);
+    void set_communicators(Core::Communication::Communicators communicators);
 
     /// return communicators
-    std::shared_ptr<Core::Communication::Communicators> get_communicators() const;
+    Core::Communication::Communicators& get_communicators() const;
```
georghammerl, line 206:
```suggestion
    const Core::Communication::Communicators& get_communicators() const;
```
Necessary/useful?

apps/global_full/4C_global_full_main.cpp
```diff
@@ -421,36 +421,37 @@ int main(int argc, char* argv[])
...
     {
-      Core::Communication::barrier(arguments.comms->global_comm());
+      Core::Communication::barrier(arguments.comms.global_comm());
       printf("processor %d finished normally\n",
-          Core::Communication::my_mpi_rank(arguments.comms->local_comm()));
+          Core::Communication::my_mpi_rank(arguments.comms.local_comm()));
     }
   }
 
+  communicators.finalize();
```
georghammerl, line 454: Out of curiosity, here the communicators are freed, while `global_data` still holds the `unique_ptr` to it. `global_data` will be destroyed later. How was the freeing of the comm done previously?

src/global_data/4C_global_data.hpp
sebproell: Not possible since there are places that modify the comms.

apps/global_full/4C_global_full_main.cpp
sebproell: Only the internal MPI_Comms are freed, not the Communicators object. Previously they were not freed at all.

georghammerl: okay. Thanks.

PR #1173, opened 2025-08-27
Remove some EpetraCrs matrix calls

dharinib98:
## Description and Context

This PR replaces some of the few `Epetra_CrsMatrix` calls with a `SparseMatrix`

#863

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
+++ b/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
@@ -181,7 +181,7 @@ void Core::LinAlg::matrix_put(const Core::LinAlg::SparseMatrix& A, const double 
   if (Aprime == nullptr) FOUR_C_THROW("Cast failed");
 
   // Loop over Aprime's rows, extract row content and replace respective row in sysmat
-  const int MaxNumEntries = EPETRA_MAX(Aprime->max_num_entries(), B.max_num_entries());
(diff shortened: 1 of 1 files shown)


src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -177,11 +177,11 @@ void Core::LinAlg::matrix_put(const Core::LinAlg::SparseMatrix& A, const double
   if (A.get_matrixtype() != Core::LinAlg::SparseMatrix::CRS_MATRIX)
     FOUR_C_THROW("Please check code and see whether it is save to apply it to matrix type {}",
         A.get_matrixtype());
-  Epetra_CrsMatrix* Aprime = const_cast<Epetra_CrsMatrix*>(&(*(A.epetra_matrix())));
+  Core::LinAlg::SparseMatrix* Aprime = const_cast<Core::LinAlg::SparseMatrix*>(&A);
   if (Aprime == nullptr) FOUR_C_THROW("Cast failed");
 
   // Loop over Aprime's rows, extract row content and replace respective row in sysmat
-  const int MaxNumEntries = EPETRA_MAX(Aprime->MaxNumEntries(), B.max_num_entries());
+  const int MaxNumEntries = EPETRA_MAX(Aprime->max_num_entries(), B.max_num_entries());
```
mayrmt: What does `EPETRA_MAX` do? Can we get rid of it right away? If it only computes the max of both arguments, then it should be easy to replace it with standard C++.

PR #1175, opened 2025-08-28
Remove beamcontact

davidrudlstorfer:
## Description and Context

Temporary PR to remove beamcontact as discussed in #1022

I temporarily kept the input from the old framework because it is used everywhere...

I also just commented out some stuff in the old structure framework. As far as I know all the beam3 and beaminteraction stuff uses the new structure framework? @isteinbrecher

Locally it built and the minimal test suite ran through

### DIFF
--- a/doc/removed_code/major_changes.txt
+++ b/doc/removed_code/major_changes.txt
@@ -1,5 +1,13 @@
 List of major features that were removed from 4C, including date and revision where code was removed
 
+Old beam contact framework
+- Removal of the old beam contact framework
+- reasons: last unported feature of enhanced gap function was implemented in the new beam interaction
+           framework (only unported feature is beam tangent smoothing => only applicable for rarely
+           used LINE2 elements)
+- removed in commit 886a76e3a3060911090ef62a754e88a4e0a0d842
+- removed by David Rudlstorfer
+
 Guarded code in beam3, beamcontact and beaminteraction
 - While cleaning up Core::LinAlg::MultiMapExtractor, some guarded code was found related to untested functions. The
   code was guarded by the flags:


maxfirmbach:
@davidrudlstorfer `beaminteraction` is soley used in `structure_new`, yes.

davidrudlstorfer:
I ran all "beam" tests locally and the following failed

 - `beam3eb_static_contact_beamtosolidcontact_solidcontact.4C.yaml`
 - `beam3eb_static_contact_penalty_newgap_twocrossedbeams.4C.yaml`
 - `beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_smoothed_twocrossedbeams.4C.yaml`
 - `beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_twocrossedbeams.4C.yaml`

These perfectly coincide with the ones from here https://github.com/4C-multiphysics/4C/discussions/1022#discussioncomment-14242105

Let's see how the pipeline compares

maxfirmbach:
@davidrudlstorfer I also checked at some point (over a year ago), I think, and as @knarfnitram already mentioned it's only related to this one parameter that is missing in the new implementation.

davidrudlstorfer:
So I now looked a little bit closer into the `BEAMS_NEWGAP` functionality

It appears here

https://github.com/4C-multiphysics/4C/blob/7ed682bf13cf83307c0fec6db2ed5287e3b1b76f/src/beamcontact/4C_beamcontact_beam3contact_manager.cpp#L1736-L1751

where it actually does nothing at all because it is commented out for more than 10 years

And the second appearance is here

https://github.com/4C-multiphysics/4C/blob/7ed682bf13cf83307c0fec6db2ed5287e3b1b76f/src/beamcontact/4C_beamcontact_beam3contactnew.cpp#L3146-L3157

As far as I can tell this feature should be pretty straightforward to port to the new implementation if it is necessary at all. But this needs to be determined by the people who actually use beam contact

@isteinbrecher @maxfirmbach @knarfnitram

knarfnitram:
So far I have not used it, but it could be interesting for me. 
I briefly checked it yesterday and interesting question is, where `gap` and `gap_original_` is then additionally used within the remaining calculations (and that's where I stopped).

maxfirmbach:
Seems like we have to put it into `4C_beaminteraction_beam_to_beam_contact_pair.cpp` in `compute_normal()`. The method there looks nearly identical to the one from `4C_beamcontact_beam3contactnew.cpp`.

davidrudlstorfer:
@knarfnitram 

`gap_original_` is used here

https://github.com/4C-multiphysics/4C/blob/7ed682bf13cf83307c0fec6db2ed5287e3b1b76f/src/beamcontact/4C_beamcontact_beam3contactnew.cpp#L3515-L3528

and this function is then only used here

https://github.com/4C-multiphysics/4C/blob/7ed682bf13cf83307c0fec6db2ed5287e3b1b76f/src/beamcontact/4C_beamcontact_beam3contact_manager.cpp#L1761-L1777

So the usage is not widespread. I still think this could be done pretty quickly.

davidrudlstorfer:
If the feature should be ported, the easiest way would probably be to either use

`beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_smoothed_twocrossedbeams.4C.yaml`

or

`beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_twocrossedbeams.4C.yaml`

to reproduce the results/features. These tests only use 2 beam elements each.

Just let me know how we should continue here. If all other tests (hopefully) pass, I would close this PR.

maxfirmbach:
@davidrudlstorfer I have no problem with keeping this open as draft and merge the fix in later.

davidrudlstorfer:
Who will decide on how we continue here?

Add the new definition of the gap function or simply remove it?

maxfirmbach:
@davidrudlstorfer We (me, @knarfnitram and @isteinbrecher) think it makes sense to port this functionality. Who of us will actually do the porting and the more detailed testing is not clear yet :wink:.

isteinbrecher:
I will have a look at `beam3eb_static_contact_beamtosolidcontact_solidcontact.4C.yaml` - but only in a couple of weeks. But we don't have to wait for that test to remove the old framework, as this is simply a test that checks that beam-to-beam, beam-to-solid and solid-to-solid contact work at the same time. It is already clear that no beam-to-solid contact functionality from the old framework will be ported, so this is simply a matter of making the test work with the new framework and adapting the results (hopefully).

Thanks for investigating the new-gap functionality @davidrudlstorfer I didn't know this was so "contained".

isteinbrecher:
I had a look at `beam3eb_static_contact_beamtosolidcontact_solidcontact`, the `NEW_GAP` feature is activated, but not used, so I think porting it should be doable. I will have a closer look in October.

isteinbrecher:
I had a look at the test `beam3eb_static_contact_beamtosolidcontact_solidcontact` and ported it to the new framework (https://github.com/isteinbrecher/4C/tree/port-beam-to-solid-contact-from-beamcontact, 8b61f8e031a2f1b36a4234f620f9ba7e666d5202). Since the underlying formulation is changed, the results had to be adapted.

mayrmt (dismissed):

davidrudlstorfer:
@knarfnitram I propose to remove the old framework like the current status of this PR

Locally, two tests fail for me:

     beam3eb_static_contact_penalty_newgap_twocrossedbeams.4C.yaml
     beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_smoothed_twocrossedbeams.4C.yaml-p1
     
Can both of these be removed? In #1022 only the second one is mentioned if I am not mistaken.

(And as already discussed, I would keep the old beam contact input for now, and move it to the new implementation in a follow-up PR)

knarfnitram:
I'll take a look at `beam3eb_static_contact_penalty_newgap_twocrossedbeams` .

maxfirmbach (approved):

davidrudlstorfer:
The last test was ported by @knarfnitram - see the first commit. Thanks a lot for porting the test @knarfnitram!

The test regarding beam tangent smoothing was removed as decided in #1022 

Therefore, this should now work and is ready for review and merge.

isteinbrecher (approved):

Issue #413, opened 2025-02-20
[BUG] Memory leaks inside RCPs


vovannikov:
There are memory leaks in the new structure solver. The debug version of Trilinos clearly reveals it, see the attached log file. valgrind also detects them, see another log.

I looked at those location indicated by Trilinos, and it seems the leaks are related for those objects which get added to a shared `ParametersList` in multiple places. At least, the pattern preceding the appearances of this leaks is always the same: something gets plugged into an arbitrary branch of a large params list. I assume, there is a cycle dependency appearing somewhere. I could not identify the source of the problem with a glance look at the code since the hierarchy between the shared pointers is actually quite complicated + a lot of objects behaving like global variables. I can try to debug it a bit further.

```
valgrind --leak-check=full --log-file=valgrind.log \
    ~/development/4C-debug/4C ~/development/4C/tests/input_files/structure_new_nln_solver_singlestep_cube_linear.dat test1
```

[valgrind.log](https://github.com/user-attachments/files/18892414/valgrind.log)
[4c-log.txt](https://github.com/user-attachments/files/18892415/4c-log.txt)

amgebauer:
Thanks for reporting the issue. That reminds me of #63

vovannikov:
Just another small remark, I think we should not abuse `ParametersList` for transferring objects all over the code, this is nasty. The pattern like this appears in multiple places:
https://github.com/4C-multiphysics/4C/blob/0c76853271366d23c585a675bead2af5fa200bfe/src/structure_new/src/utils/4C_structure_new_dbc.cpp#L99-L108

sebproell:
FYI, I have been trying to reduce the number of RCPs/shared_ptrs inside 4C for precisely this reason. And fully agree on the ParameterList thing. It's really meant for input and should not be used as dynamic container to make bad inheritance work, see #61.

vovannikov:
Actually, the code example above is exactly a typical example that creates a cyclic dependency, we need just to a few more lines above to recognize it:
https://github.com/4C-multiphysics/4C/blob/0c76853271366d23c585a675bead2af5fa200bfe/src/structure_new/src/utils/4C_structure_new_dbc.cpp#L94-L109
`timint_ptr_` is a shared pointer inside `Solid::Dbc`, it contains the `ParametersList`, to which we ultimately store the `RCP` to the `this` object itself.

PR #1177, opened 2025-08-28
ASAN: only suppress report on specific memory leaks

sebproell:
Way back when I introduced AddressSanitizer (ASAN) years ago on Gitlab, we had enough to do to fix the serious memory errors. Memory leak reports were always suppressed since then. This PR makes the various suppressions explicit. Draft for now, maybe I can fix a few leaks right away.

Related to #413


sebproell:
Suppressing only what I have in this PR reveals 1,100 indirect leaks in our code or related to mistakes that we make (test  beam3r_line2_genalpha_liegroup_contact_penalty_linpen_twocrossedbeams.4C.yaml-p2). Interestingly, there are no direct leaks. This means that all leaks are caused by reference counting cycles as reported e.g. in #413.

[leaks.log](https://github.com/user-attachments/files/22021264/leaks.log)

PR #1178, opened 2025-08-28
Draft: Enable std::array as a supported InputSpec type

lauraengelhardt:
## Description and Context
Currently, we only support `std::vector`. If you want to store a variable in an `std::array`, a conversion step is needed after reading the input. This PR hence adds the functionalities to support `std::array` as an input type, such that you can directly specify a parameter as, e.g., 
```
parameter<std::array<double, 3>>("name", {.description = "..."})
```

I applied this to the STRUCTURE/FLUID DOMAIN sections. 

@gilrrei As discussed, we treat arrays in the schema file as of type "vector". 

#1148 #1092


src/core/io/src/4C_io_input_types.hpp
```diff
@@ -111,6 +117,12 @@ namespace Core::IO
       static constexpr std::size_t value = 1 + RankHelper<T>::value;
     };
 
+    template <typename T, std::size_t n>
+    struct RankHelper<std::array<T, n>>
+    {
+      static constexpr std::size_t value = RankHelper<T>::value;
```
sebproell, line 123: I think we should now rename `rank()` to `dynamic_rank()` and mention in the docs that this is the number of dynamic extents a type has. Would you mind adding a commit here that does that?

src/core/io/src/4C_io_input_spec_validators.hpp
```diff
@@ -227,12 +227,13 @@ namespace Core::IO::InputSpecBuilders::Validators
...
-   * all_elements(positive<int>()):
+   * all_elements<std::vector<int>>(positive<int>()):
    * @endcode
    *
    */
-  template <typename T>
-  [[nodiscard]] Validator<std::vector<T>> all_elements(Validator<T> validator);
+  template <typename Container>
+  [[nodiscard]] Validator<Container> all_elements(
+      Validator<typename Container::value_type> validator);
```
sebproell: As just discussed in person, let's use `std::span` here.

src/core/io/src/4C_io_input_types.hpp
lauraengelhardt: Agree, I'll do!

lauraengelhardt:
For documentation: Enabling the validators for arrays has been proven to be more complicated than we expected, if we want to do it right. @sebproell will take a deeper look at the Validator class. This PR will thus only introduce `std::array` as an InputSpec. Once we have a working solution for the validators, I will switch some parameter InputSpecs from `std::vector`  to `std::array`, for e.g., the `DOMAIN` sections.

src/core/io/src/4C_io_input_spec_validators.hpp
lauraengelhardt: I'll resolve this conversation as not applicable for the PR anymore.

sebproell (approved):

amgebauer (approved):

PR #1182, opened 2025-08-29
Remove shared_ptr on TimIntMStep

sebproell:
No reason to have this in a shared_ptr.

### DIFF
--- a/src/timestepping/4C_timestepping_mstep.hpp
+++ b/src/timestepping/4C_timestepping_mstep.hpp
@@ -74,6 +74,13 @@ namespace TimeStepping
       return Core::Utils::shared_ptr_from_ref(state_[index_by_step(step)]);
     }
 
+    std::shared_ptr<const STATE> operator()(const int step  //!< inquiry step
+    ) const
+    {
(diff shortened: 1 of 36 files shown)


sebproell:
FYI, I reverted the changes to the old structure time integration since I got non-obvious wrong results for explicit time integration schemes. Ready for review.

src/timestepping/4C_timestepping_mstep.hpp
```diff
@@ -74,6 +74,13 @@ namespace TimeStepping
       return Core::Utils::shared_ptr_from_ref(state_[index_by_step(step)]);
     }
 
+    std::shared_ptr<const STATE> operator()(const int step  //!< inquiry step
+    ) const
+    {
+      if (not step_exists(step)) FOUR_C_THROW("Step {} is not admissible", step);
+      return Core::Utils::shared_ptr_from_ref(state_[index_by_step(step)]);
+    }
```
amgebauer, line 82: Where does this come from?

sebproell: It was missing before. It is the correct const-qualified equivalent to the non-const accessor. Before this PR, the shared_ptrs masked problems in const-correctness.

PR #1185, opened 2025-08-29
Add workflow to test fourcipp

davidrudlstorfer:
## Description and Context

First part of #973.

Some points to discuss:

- We definitely need the metadata/schema file from the clang18 build of each PR. Currently, the `fourcipp_test` is executed after `buildtest` completed. Ideally, it should start after the clang18 build is finished and not the entire workflow. Unfortunately, as far as I know it is not possible to start a workflow based on an individual job from another workflow. We could move the `fourcipp_test` to the `buildtest` but I do not really like this idea. @amgebauer any tips/tricks/ideas?
    
- We should also lint the 4C input files with FourCIPP as @gilrrei stated in https://github.com/4C-multiphysics/4C/issues/973#issuecomment-3056277376. @gilrrei what is your vision there? Is this necessary if we already validate all input files against the latest schema in the clang18 build?

#973

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -309,3 +309,43 @@ jobs:
         uses: re-actors/alls-green@release/v1
         with:
           jobs: ${{ toJSON(needs) }}
+
+  test_fourcipp:
+    runs-on: ubuntu-latest
+    needs: clang18_build
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          repository: 4C-multiphysics/fourcipp
+      - name: Set up virtual environment
+        uses: conda-incubator/setup-miniconda@v3.2.0
+        with:
+          auto-update-conda: true
+          conda-remove-defaults: true
+          activate-environment: fourcipp
+          python-version: 3.12
+      - name: Install requirements from FourCIPP
+        shell: bash -el {0}
+        run: |
+          pip install .
+      - name: Download schema file from current pull request
+        uses: actions/download-artifact@v5
+        with:
+          name: clang18_build-schema
+          path: ${{ github.workspace }}/latest_metadata_schema/
+
+      - name: Download metadata file from current pull request
+        uses: actions/download-artifact@v5
+        with:
+          name: clang18_build-metadata
+          path: ${{ github.workspace }}/latest_metadata_schema/
+      - name: Move current metadata and schema file into FourCIPP (continue to use default profile)
+        shell: bash -el {0}
+        run: |
+          mv ${{ github.workspace }}/latest_metadata_schema/4C_metadata.json ${{ github.workspace }}/src/fourcipp/config/4C_metadata.json
+          mv ${{ github.workspace }}/latest_metadata_schema/4C_schema.json ${{ github.workspace }}/src/fourcipp/config/4C_schema.json
+      - name: Run pytest
+        shell: bash -el {0}
+        run: |
+          pytest --color=yes -v --no-cov


amgebauer:
> We definitely need the metadata/schema file from the clang18 build of each PR. Currently, the fourcipp_test is executed after buildtest completed. Ideally, it should start after the clang18 build is finished and not the entire workflow. Unfortunately, as far as I know it is not possible to start a workflow based on an individual job from another workflow. We could move the fourcipp_test to the buildtest but I do not really like this idea. @amgebauer any tips/tricks/ideas?

You could add a step in the build-test workflow after uploading the schema file that triggers the fourcipp-job. But I think it will then not be part of the PR and will not be shown in the PR checks.

davidrudlstorfer:
Do we have any other magic in place in 4C that prevents this workflow from being run? I do not really understand why it did not start

amgebauer:
> Do we have any other magic in place in 4C that prevents this workflow from being run? I do not really understand why it did not start

It only starts if the workflow is on the main branch.

mayrmt:
@davidrudlstorfer As already voiced in #973, there are concerns, that have not been resolved over the last two months. 

I have said this in #973, so I say it loud and clear again: I strongly object to cross-test across repository boundaries. This interconnects two independent repositories. And yes, 4C is independent of FourCIPP. That might not be true for FourCIPP, but for sure is and must be true for 4C.

sebproell:
@mayrmt Could you explain these objections? Maybe with an example? What are the negative consequences you fear?

I think we will need to see how exactly FourCIPP and 4C interact in the future. Basically, the question is whether FourCIPP is the repo for 4C's python tools which could just as well be part of 4C itself or whether it is something more clearly separated. That's not clear (to me) yet. For now, I don't see a problem but rather a benefit in testing that FourCIPP works ahead of time.

mayrmt:
As @davidrudlstorfer pointed out already in the original issue #973, testing FourCIPP inside 4C’s PR pipeline creates a coordination deadlock: each change requires a cross-repo merge. This result in a deadlock, where neither repository can't merge. (Yes, one could configure the workflow to not block a PR even if it is failing. Still, this yields a non-clean set of PR tests in 4C. And what's the point of a failing, but accepted workflow? What is supposed to happen, if any PR results in a failing FourCIPP test?). For 4C, cross-repository testing slows down delivery and complicates rollbacks.

Again: **4C is independent of FourCIPP**, so it should not include FourCIPP into its testing. We also do not include other "associated projects" into 4C. If FourCIPP wants an early detection for changes in 4C, then it can implement in its own workflows. Or define a compatible 4C hash, that is update automatically, if acceptance testing of a new is successful (similar to deal.ii updates in the "original" 4C repository on GitLab). If acceptance testing fails, changes have to be made. Any approach along these lines maintains the independence of projects also in the PR testing.

I am not aware of any other established project, that performs cross-repository testing -- for good reasons. Why should we deviate from such standard practices?

davidrudlstorfer:
> As @davidrudlstorfer pointed out already in the original issue https://github.com/4C-multiphysics/4C/issues/973, testing FourCIPP inside 4C’s PR pipeline creates a coordination deadlock: each change requires a cross-repo merge. This result in a deadlock, where neither repository can't merge. (Yes, one could configure the workflow to not block a PR even if it is failing. Still, this yields a non-clean set of PR tests in 4C. And what's the point of a failing, but accepted workflow? What is supposed to happen, if any PR results in a failing FourCIPP test?). For 4C, cross-repository testing slows down delivery and complicates rollbacks.

This is not really a problem in my opinion - as you've written yourself and anticipated with this PR => we can still merge everything in 4C althought the test would fail. This would act as an early warning detection that all of our downstream projects (FourCIPP => BeamMe, 4C Webviewer, Queens, ...) will fail.

> Again: 4C is independent of FourCIPP, so it should not include FourCIPP into its testing. We also do not include other "associated projects" into 4C. If FourCIPP wants an early detection for changes in 4C, then it can implement in its own workflows. Or define a compatible 4C hash, that is update automatically, if acceptance testing of a new is successful (similar to deal.ii updates in the "original" 4C repository on GitLab). If acceptance testing fails, changes have to be made. Any approach along these lines maintains the independence of projects also in the PR testing.
> 
> I am not aware of any other established project, that performs cross-repository testing -- for good reasons. Why should we deviate from such standard practices?

I am also not aware of any other established project which is able to do all these things we are currently capable of doing with the input of 4C. In my opinion with FourCIPP we tried many new approaches (kudos to @gilrrei) and I must confess that these are pretty successful. For me it is hard to justify to not test out something "because no other established projects does this" is not a really good reason for myself.

As far as I can see it is not a real problem that all downstream projects break from time to time and then everybody is in a hurry - therefore, closing this PR

PR #1186, opened 2025-08-29
Remove a bunch of unnecessary smart pointers

sebproell:
A while back, I wrote a clang tool to replace unnecessary smart pointers with values/references. This is still experimental and unpublished, but 4C is a nice code base to test it on :D  

One of the minor issues the tool has is that it introduces explicit mentions of the `FourC` namespace. Since we have accumulated a few of these anyway, I will add a second commit that cuts down on `FourC::`.


sebproell:
Could we merge this trivial change so I don't have to keep rebasing? :pray:

sebproell:
Caching gives me a build that fails in the linking stage. I removed the clang18 cashes of this branch. This is the second time this has happened to me since we are on GitHub. Not terribly often, but maybe there is a setting in ccache that needs to be adapted for these edge cases.

amgebauer (approved):

PR #1188, opened 2025-08-29
Use std::numbers instead of C macros

### LINKED ISSUE #1168: Replace preprocessor constants with std::numbers equivalents
Since C++20 we can use [these constants](https://en.cppreference.com/w/cpp/numeric/constants.html) instead of the respective preprocessor defines `M_PI`, `M_E`, etc. 

## Replace the following

- [ ] `M_PI` -> `std::numbers::pi`
- [ ] feel free to add more

sebproell:
Closes #1168


mayrmt (approved):
Looks good! I did not check every change manually, but looked at many places. The passing pipeline takes care the rest.

Issue #825, opened 2025-05-25
Allow GMSH (.msh) file format for input


PhilipOesterlePekrun:
## Description
Given that 4C is now open source and not everyone has access to Cubit, I think it is useful to consider adding support for more standard mesh formats, such as Gmsh (.msh). While Exodus II (.e) is not closed source, there are not exactly many tools other than Cubit that allow one to go from geometry to an Exodus II file that works for 4C.

The documentation states that meshio can be used to convert from Gmsh to Exodus II--the issue here is that meshio (and all other tools I have found) converts 1- and 2-Dimensional* physical groups into sidesets*, while our 4C input files use only nodesets for the definition of boundary conditions.

Of course, anybody could modify meshio or make a tool based on SEACAS which also converts 1D and 2D* physical groups from Gmsh into nodesets, but this is not exactly user-friendly (even if, for example, I made such a tool, people would have to find it/know about it in the first place). I think it would be nicer if Gmsh support was directly built into 4C. As I see it, there are two options:

- We could consider supporting adding support to recognize sidesets for DESIGN LINE and DESIGN SURF* conditions, i.e.
  ```yaml
  DESIGN LINE DIRICH CONDITIONS: # or SURF
    - ENTITY_TYPE: side_set_id
      ...
  ```
  This would allow the physical groups from Gmsh that are converted to sidesets by meshio to be recognized. However, I am not familiar with how sidesets fundamentally differ from nodesets and what implications this might have.
  (also mentioned in #653)

- We could alternatively allow a reference to a .msh file in `FILE` under `STRUCTURE GEOMETRY`. This might take more work to integrate into the reader. `Core::IO::Gmsh` already exists, but this seems to be purely for writing to a Gmsh postprocessing file.
  (also mentioned in #60)

*...in 3D. For purely 2D geometry, 2D physical groups are of course converted into element blocks

## Interested Parties
@mayrmt

sebproell:
We really should use side sets in 4C. Node sets can not always correctly identify faces of a cell and even if they can, it is a much more indirect way to specify faces. I am currently working a lot on the conditions as a side effect of refactoring our discretization-related functionality. This should make side sets feasible soon.

mayrmt:
@PhilipOesterlePekrun, thanks for opening this feature request!

I agree with @sebproell in preferring proper support of side sets.

PhilipOesterlePekrun:
Thanks for the feedback, and great to hear that it is a planned change.

[...]

PR #1189, opened 2025-08-29
Generalize external mesh input

amgebauer:
These changes generalize the exodus intermediate input representation so that it can also be used for other mesh formats. The data structure now contains all the data, the mesh reader expects to build the 4C discretization.

I have a small draft for vtu file input, which are a couple of lines then (using the vtk c++ library).

Part of #825

### DIFF
--- a/src/core/io/src/4C_io_exodus.cpp
+++ b/src/core/io/src/4C_io_exodus.cpp
@@ -128,33 +128,16 @@ namespace
 }  // namespace
 
 
-std::string Core::IO::Exodus::describe(VerbosityLevel level)
-{
-  switch (level)
-  {
-    case VerbosityLevel::none:
-      return "no output";
-    case VerbosityLevel::summary:
-      return "output of summary for blocks and sets";
-    case VerbosityLevel::detailed_summary:
-      return "output of summary for each block and set";
-    case VerbosityLevel::detailed:
-      return "detailed output for each block and set";
-    case VerbosityLevel::full:
-      return "detailed output, even for nodes and element connectivities";
-  }
-  std23::unreachable();
-}
 
-
-Core::IO::Exodus::Mesh::Mesh(std::filesystem::path exodus_file, MeshParameters mesh_parameters)
-    : mesh_parameters_(mesh_parameters)
+Core::IO::MeshInput::Mesh Core::IO::Exodus::read_exodus_file(
+    const std::filesystem::path& exodus_file, MeshParameters mesh_parameters)
 {
+  Core::IO::MeshInput::Mesh mesh{};
+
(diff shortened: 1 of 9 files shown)


src/core/io/src/4C_io_mesh.hpp
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * The ID of the first node in the mesh. This defaults to 1, since this is the default in
+     * the Exodus II mesh format.
+     */
+    int node_start_id{1};
+  };
+
+  struct Mesh
+  {
+    MeshParameters mesh_parameters_;
```
sebproell:
```suggestion
    MeshParameters mesh_parameters;
```

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -0,0 +1,153 @@
...
+  struct MeshParameters
+  {
+    /**
+     * The ID of the first node in the mesh. This defaults to 1, since this is the default in
+     * the Exodus II mesh format.
+     */
+    int node_start_id{1};
+  };
+
+  struct Mesh
```
sebproell, line 54: Needs some docs.

src/core/io/src/4C_io_exodus.cpp
```diff
@@ -128,33 +128,16 @@ namespace
...
-  std23::unreachable();
-}
-
 
-Core::IO::Exodus::Mesh::Mesh(std::filesystem::path exodus_file, MeshParameters mesh_parameters)
-    : mesh_parameters_(mesh_parameters)
+std::unique_ptr<Core::IO::MeshInput::Mesh> Core::IO::Exodus::read_exodus_file(
+    const std::filesystem::path& exodus_file, MeshParameters mesh_parameters)
 {
+  std::unique_ptr<Core::IO::MeshInput::Mesh> mesh = std::make_unique<Core::IO::MeshInput::Mesh>();
```
sebproell: Why do you heap-allocate this object instead of returning by value? You can still move it into a `unique_ptr` in the relevant places.

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -0,0 +1,167 @@
...
+  struct CellBlock;
+  struct SideSet;
+  struct LineSet;
+  struct PointSet;
+
+  /*!
+   * @brief An intermediate representation of finite element meshes
+   *
+   * 4C will read meshes into this basic representation of the mesh and generate its internal
+   * Discretization from it.
```
sebproell, line 51: Nice! If we make this a little more flexible, we could probably do all our mesh input via this struct. We would need some partitioning support for this to also work for legacy string meshes in the input file.

Issue #1190, opened 2025-08-30
[LinAlg] Move FOUR_C_THROW into respective wrapper classes


maxfirmbach:
## Description

Currently many methods of the linear algebra wrapper classes return error integers and outside a error check happens. I propose to move these error checks (e.g. FOUR_C_THROW) into the wrapper classes itself.

Example:
```cpp
int err = my_graph->insert_global_indices(1, &index_main, 1, &index);
if (err != 0)
FOUR_C_THROW("Core::LinAlg::Graph::InsertGlobalIndices returned {} for global row {}", err, node_main->id());
```
This is a common pattern that appears all over the code. This can be simplified by just throwing inside the wrapper class itself, which would make the other parts of the code a bit more clearer.

This is part of #875.

Actually, `Tpetra` doesn't return error codes anymore, but throws internally, so this will also help with the conversion.

## Possible Solution and Definition of Done

- [ ] Add `CHECK_EPETRA_CALL()` inside the methods of the wrapper classes (see #1191 #1390 #1411 #1412)
- [ ] Change the return value of these methods to `void`
- [ ] Remove the error checks in the other parts of the code

## Interested Parties

@knarfnitram @dharinib98 @mayrmt

maxfirmbach:
Well, we don't catch errors on a regular basis with these methods either ... some parts check the integer return, others don't ... I wonder if some parts of the code will actually error out with these changes 😅.

mayrmt:
I assume, that errors will arise. But it's better to actually throw than to silently continue.

maxfirmbach:
With #1546 merged, this issue can be considered done.

PR #1191, opened 2025-08-30
[LinAlg] Refactor error throw in Graph

maxfirmbach:
## Description and Context

This PR refactors the error handling and throwing inside the `Graph` wrapper class. The error handling is moved into the wrapper class and some refactoring is done.

Part of #1190

### DIFF
--- a/src/core/utils/src/exceptions/4C_utils_epetra_exceptions.hpp
+++ b/src/core/utils/src/exceptions/4C_utils_epetra_exceptions.hpp
@@ -12,6 +12,15 @@
 
 #include "4C_utils_exceptions.hpp"
 
+/**
+ * Epetra has a C-style interface returning integer error codes. This macro is meant to be wrapped
+ * around a call to an Epetra function and asserts that the integer return value is equal to zero.
+ * Otherwise, an exception is thrown.
+ *
+ * @code
+ *    CHECK_EPETRA_CALL(graph_->insert_global_value(...));
+ * @endcode
+ */
 #define CHECK_EPETRA_CALL(expr)                                     \
   do                                                                \
   {                                                                 \


src/core/linalg/src/sparse/4C_linalg_graph.cpp
```diff
@@ -57,28 +56,92 @@ Core::LinAlg::Graph& Core::LinAlg::Graph::operator=(const Graph& other)
...
   if (graphtype_ == CRS_GRAPH)
-    FOUR_C_THROW("This type of insert_global_indices() only available for FE_GRAPH type.");
+    err = graph_->FillComplete(domain_map.get_epetra_block_map(), range_map.get_epetra_block_map());
   else if (graphtype_ == FE_GRAPH)
     err = static_cast<Epetra_FECrsGraph*>(graph_.get())
-              ->InsertGlobalIndices(numRows, rows, numCols, cols);
+              ->GlobalAssemble(domain_map.get_epetra_map(), range_map.get_epetra_map());
 
-  return err;
+  if (err != 0) FOUR_C_THROW("Graph::fill_complete() returned {}.", err);
```
mayrmt: Technically, it does not return. Rather an error occurred.

This comment also applies to the other methods in the wrapper.

maxfirmbach: I updated the error message. Now it is: "Graph::xxx() was not successful and returned error code {}.".

sebproell: The name of the function is already reported in the stack trace.

maxfirmbach:
> You could write a macro like `CHECK_EPETRA_CALL(expr)` that roughly looks like
> 
> ```
> #define CHECK_EPETRA_CALL(expr) \
>     do { \
>        int err = (expr); \
>        FOUR_C_ASSERT_ALWAYS(err == 0, "Epetra error (code {}) while calling '#expr'.", err); \
>     } while(0)
> ```

Good idea! Seems like someone actually had that one before, just found something in `4C_utils_epetra_exceptions.hpp` ... seems like it barely got used, maybe I can work with that.

src/core/utils/src/exceptions/4C_utils_epetra_exceptions.hpp
```diff
@@ -27,20 +27,20 @@ void set_trace_back_mode(int tracebackmode);
...
-    }                                                                                             \
-    else if (epetra_error > 0)                                                                    \
-    {                                                                                             \
-      std::cout << __FILE__ ", " << __LINE__ << ": EPETRA_WARNING " << epetra_error << " caught!" \
-                << std::endl;                                                                     \
-    }                                                                                             \
-    set_trace_back_mode(0);                                                                       \
+#define CATCH_EPETRA_ERROR(expr)                                                    \
+  {                                                                                 \
+    set_trace_back_mode(2);                                                         \
```
sebproell: Ah nice. It is a pretty standard thing to have these macros.

- The name is wrong since this throws rather than catches.
- Could you go with my suggestion unless you understand what this `trace_back` stuff does? Seems overengineered on first glance. If you change it, be sure to also remove the `set_trace_back_mode` function.
- Also, the warning should just be an error, right? Maybe if we make progress on #937 we can revisit such cases. A simple printout will get overlooked, though.

maxfirmbach: Yes, I can go with your thing (and in theory delete the current one). I think the warning should also be treated as error.

maxfirmbach:
I think now it's looking right.

src/core/utils/src/exceptions/4C_utils_epetra_exceptions.hpp
```diff
@@ -12,36 +12,11 @@
...
-    }                                                                                             \
-    else if (epetra_error > 0)                                                                    \
-    {                                                                                             \
-      std::cout << __FILE__ ", " << __LINE__ << ": EPETRA_WARNING " << epetra_error << " caught!" \
-                << std::endl;                                                                     \
-    }                                                                                             \
-    set_trace_back_mode(0);                                                                       \
-  }
-
+#define CHECK_EPETRA_CALL(expr)                                     \
```
sebproell, line 24: Small doc would be nice.

"Epetra has a C-style interface returning integer error codes. This macro is meant to be wrapped around a call to an Epetra function and asserts that the integer return value is equal to zero. Otherwise, an exception is thrown."

maxfirmbach: Ah ... true ...

knarfnitram (approved):

PR #1192, opened 2025-08-30
Add optional dependency to VTK

amgebauer:
This will allow us to use VTK for reading meshes.

Installable on ubuntu via `sudo apt install libvtk9-dev`. Totally optional at the moment and off by default.


mayrmt (dismissed):
Looks good!

If reading meshes from a different format than exodus works nicely, we could think about making all types of mesh input a configurable option.

cmake/configure/configure_VTK.cmake
```diff
@@ -0,0 +1,27 @@
...
+
+if(VTK_FOUND)
+  message(STATUS "VTK component libraries: ${VTK_LIBRARIES}")
+
+  target_compile_definitions(
+    VTK::CommonCore
+    INTERFACE "-DVTK_MAJOR_VERSION=${VTK_MAJOR_VERSION}"
+              "-DVTK_MINOR_VERSION=${VTK_MINOR_VERSION}"
+              "-DVTK_BUILD_VERSION=${VTK_BUILD_VERSION}"
+    )
```
sebproell: I think we also have this hack in other places but now I am wondering if this can ever work when using an installed version 🤔 It seems we are not adding these definitions in this case.

amgebauer: I think this sneaked in accidentally. I removed it.

cmake/configure/configure_VTK.cmake
```diff
@@ -0,0 +1,20 @@
...
+# See the LICENSE.md file in the top-level for license information.
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+find_package(VTK REQUIRED COMPONENTS CommonCore CommonDataModel IOXML)
+
+if(VTK_FOUND)
+  message(STATUS "VTK component libraries: ${VTK_LIBRARIES}")
+
+  target_link_libraries(four_c_all_enabled_external_dependencies INTERFACE ${VTK_LIBRARIES})
```
sebproell: Can we use an actual target here? What hides behind `VTK_LIBRARIES`?

amgebauer: That are the components that we searched for in `find_package` (`CommonCore CommonDataModel IOXML`)

cmake/configure/configure_VTK.cmake
```diff
@@ -0,0 +1,23 @@
...
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+find_package(VTK REQUIRED COMPONENTS CommonCore CommonDataModel IOXML)
+
+if(VTK_FOUND)
+  message(STATUS "VTK component libraries: ${VTK_LIBRARIES}")
+
+  target_link_libraries(
+    four_c_all_enabled_external_dependencies
```
sebproell, line 14: I doubt that this works. Isn't this in a namespace, i.e.


```suggestion
    four_c_all_enabled_external_dependencies INTERFACE VTK::CommonCore VTK::CommonDataModel VTK::IOXML
```

amgebauer: yes it is, I already fixed it.

sebproell (approved):

PR #1195, opened 2025-08-31
Allow vtu mesh input

amgebauer:
This PR enables vtu input if 4C is compiled with `FOUR_C_WITH_VTK`.

Note, vtu does not have a native support for node-sets and cell-blocks. The provided mesh must therefore come with the following data-arrays:

 * cell-array `block_id`: Defines the id of the block a cell belongs to. A block can only contain cells of the same type.
 * point-array `point_set_#`: Defines whether a point is part of a point-set with id `#` 

These arrays can very easily be added in python with the [meshio](https://github.com/nschloe/meshio)-package. This pacage can also be used to convert meshes from many other standard mesh formats.

Part of #60 #825


src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -0,0 +1,215 @@
...
+            cell_type, std::span{connectivity, static_cast<std::size_t>(number_of_points)});
+  }
+
+  return mesh;
+}
+
+#else
+Core::IO::MeshInput::Mesh Core::IO::VTU::read_vtu_file(const std::string& vtu_file)
+{
+  FOUR_C_THROW("You have to enable VTK to support vtu mesh file input (option FOUR_C_WITH_VTK)");
```
sebproell:
```suggestion
  FOUR_C_THROW("You have to enable VTK to support vtu mesh file input. Reconfigure 4C with the CMake option 'FOUR_C_WITH_VTK' set to 'ON".");
```

tests/list_of_tests.cmake
```diff
@@ -1922,6 +1922,10 @@ four_c_test(TEST_FILE xfsi_push_1D_1st_order_dens5.4C.yaml)
 four_c_test(TEST_FILE xfsi_push_1D_1st_order.4C.yaml)
 four_c_test(TEST_FILE xfsi_push_1D_2nd_order.4C.yaml)
 
+if (FOUR_C_WITH_VTK)
+    four_c_test(TEST_FILE solid_vtu_input.4C.yaml NP 2 RESTART_STEP 5)
+endif (FOUR_C_WITH_VTK)
```
sebproell: I will add a utility to make these dependency handling a bit more obvious.

src/core/io/src/4C_io_element_vtk_cell_type_register.hpp
```diff
@@ -10,77 +10,279 @@
...
-        FOUR_C_THROW("VTK cell type not implemented for element: {}",
-            Core::FE::cell_type_to_string(four_c_ele_shape_type));
+      reverse_mapping[vtk_connectivity_mapping<cell_type>[i]] = i;
     }
+
+    return reverse_mapping;
+  }();
+
+  /*!
+   * @brief Mapping of the 4C celltype to the vtk celltype for output
```
sebproell: Maybe create `using VtkCellType = uint8_t;`?

src/core/io/src/4C_io_element_vtk_cell_type_register.hpp
```diff
@@ -10,77 +10,279 @@
...
+  template <>
+  static constexpr uint8_t vtk_cell_type<Core::FE::CellType::point1> = 1;
+
+  /*!
+   * @brief This is a list of all known mapping from vtk-celltypes to 4C CellTypes
+   *
+   * @note This map only contains 1-to-1 mappings (bijective) of the celltype.
+   */
+  constexpr std::array vtk_celltype_mapping = {
+      std::make_pair(Core::FE::CellType::quad4, 9),
```
sebproell: Isn't this duplicated info? Can you generate it from the listing above?

src/core/io/src/4C_io_vtu_reader.hpp
```diff
@@ -0,0 +1,42 @@
...
+   *
+   *  - point-arrays "point_set_##": Integer arrays telling whether a point is part of a point-set.
+   * `##` is the id of the point-set
+   *  - cell-array "block_id": The id of the block of the respective cell. Note that every block can
+   * only contain elements of the same cell-type.
+   *
+   * @param vtu_file (in) : Path to the file
+   * @return Core::IO::MeshInput::Mesh
+   */
+  Core::IO::MeshInput::Mesh read_vtu_file(const std::string& vtu_file);
```
sebproell: String or filesystem? Please check the includes as well.

src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -0,0 +1,215 @@
...
+FOUR_C_NAMESPACE_OPEN
+
+#ifdef FOUR_C_WITH_VTK
+namespace
+{
+  // Returns a reference to the specified array in the given vtkDataArrayCollection
+  vtkDataArray& get_array(auto* data, const std::string& name)
+  {
+    vtkDataArray* data_array = data->GetArray(name.c_str());
+    FOUR_C_ASSERT_ALWAYS(data_array, "Array {} not found!", name);
```
sebproell: The error could be improved to tell users that 4C requires special arrays (as you say in the docs) and what they should contain.

src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -0,0 +1,215 @@
...
+              using ArrayT = std::decay_t<decltype(*typedArray)>;
+              using ValueT = typename ArrayT::ValueType;
+              static_assert(std::is_integral_v<ValueT>, "Expecting only integer arrays");
+
+              ValueT value = typedArray->GetComponent(tupleIdx, compIdx);
+              result = static_cast<T>(value);
+            }))
+    {
+      FOUR_C_THROW(
+          "Array is of type {}, but expecting an integer-type array!", array.GetDataTypeAsString());
```
sebproell: Can we print the array name?

src/core/io/src/4C_io_vtu_reader.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Core::IO::VTU
+{
+  /*!
+   * @brief Read a vtu-mesh into the 4C intermediate representation
+   *
+   * The vtu-format does not have a natural representation of cell- and point-sets. To define these
+   * sets, we assume the following point-/cell-data arrays to be present:
+   *
+   *  - point-arrays "point_set_##": Integer arrays telling whether a point is part of a point-set.
```
sebproell:
```suggestion
   *  - zero or more point-arrays named "point_set_##": Integer arrays telling whether a point is part of a point-set.
```

sebproell: Should the "point set" be called "node set"? We seem to use it later with that name.

sebproell (commented):
Very nice and not intrusive at all! Apart from my comments: how about adding a unit test for the mesh reading? Not sure it is absolutely required, but it could be a nice way to ensure that all vtu data is present in the `MeshInput`.

src/core/io/src/4C_io_element_vtk_cell_type_register.hpp
amgebauer: Unfortunately not, because the list above (meant for output) is not bijective. We map cell-types like `hex16` or `hex18` to `hex8` during vtu output.

sebproell: Ah right! How about bundling information about one of our cell types into a struct then

```
struct VtkCellInfo
{
  VtkCellType vtk_cell_type_in;
  VtkCellType vtk_cell_type_out;
  std::vector<> reordering;
  
  ...
}
```

and store the complete information in some form of mapping?

I find the separate template specializations a bit hard to parse. Also, don't you fundamentally need to make run time decisions about the types? I believe the remapping from template metaprogramming world to dynamic world show  this as well. I am wondering how well-suited templated info is for this task.

tests/list_of_tests.cmake
sebproell: #1196

src/core/io/src/4C_io_element_vtk_cell_type_register.hpp
amgebauer: I separated it into those that have a direct mapping, and those that need special handling for vtk output. Now the magic during vtk output is more visible.

> I find the separate template specializations a bit hard to parse. Also, don't you fundamentally need to make run time decisions about the types? I believe the remapping from template metaprogramming world to dynamic world show this as well. I am wondering how well-suited templated info is for this task.

I need the templated info to generate the reverse mapping at compile time.

amgebauer:
I now added unit-tests for all supported 3d elements. Note that our previous vtk connectivity mapping of our wedge elements was wrong. It lead to a negative jacobian during output.

src/core/io/src/4C_io_vtu_reader.hpp
amgebauer: I use `point` instead of `node` because it is more common in geometry/mesh generation and that's also the way vtk refers to it. Similarily, I use `cell` instead of `element`.

amgebauer:
Now ready for review

src/core/io/src/4C_io_vtu_reader.hpp
```diff
@@ -0,0 +1,43 @@
...
+   * @brief Read a vtu-mesh into the 4C intermediate representation
+   *
+   * The vtu-format does not have a natural representation of cell- and point-sets. To define these
+   * sets, we assume the following point-/cell-data arrays to be present:
+   *
+   *  - zero or more point-arrays named "point_set_#": Integer arrays telling whether a point is
+   * part of a point-set.
+   * `#` is the id of the point-set
+   *  - cell-array "block_id": The id of the block of the respective cell. Note that every block can
+   * only contain elements of the same cell-type.
```
sebproell: Formatting.

src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -0,0 +1,208 @@
...
+namespace
+{
+  // Returns a reference to the specified array in the given vtkDataArrayCollection
+  vtkDataArray& get_array(auto* data, const std::string& name)
+  {
+    vtkDataArray* data_array = data->GetArray(name.c_str());
+    FOUR_C_ASSERT_ALWAYS(data_array,
+        "Array {} not found!\n\n4C requires: \n\n * Zero or more integer-type point-arrays "
+        "'point_set_#' that define the points in a set with id #. A point is in the set if the "
+        "respective value is not zero.\n * Integer-typed cell-array `block_id` that define blocks "
```
sebproell:
```suggestion
        "respective value is not zero.\n * Exactly one integer-typed cell-array `block_id` that defines blocks "
```
?

c-p-schmidt (approved):
Nice feature!

Issue #594, opened 2025-04-02
[Trilinos] Switch from Isorropia/Zoltan to Zoltan2/Zoltan


maxfirmbach:
## Description

Currently we use the `Zoltan` package and it's `Epetra` wrapper `Isorropia` from Trilinos as interface to partitioning methods, mostly for distributing things regarding the discretization. Being able to switch to `Tpetra means we need to switch to the newer package `Zoltan2, which can handle `Tpetra` data structures.
There might be some slight changes in parameter names, but the overall behavior should not change significantly.

With the switch to `Zoltan2`, we can deactivate `Isorropia`, but still need the dependency to `Zoltan`.

## Possible Solution and Definition of Done

**Preparation**
- [x] Place all partitioning functionality and calls to `Isorropia` into a dedicated place
Currently: `4C/src/core/rebalance/src/4C_rebalance_graph_based.{hpp, cpp}`
- [x] Is this necessary? Make testing robust regarding parallel partitioning independent of the method used (see #138)
- [x] Activate `Zoltan2` in our Trilinos configure script

**Switching**
- [x] Write a `Zoltan2` interface to replace the 'Isorropia' one (see #596)
- [x] Solve possible issues on the way: 
  - There's an inconsistency, means the hypergraph partitioning results from `Isorropia` and `Zoltan2` differ slightly (see https://github.com/trilinos/Trilinos/issues/14010)
- [x] Check if it works correctly by using our current `ctest` routines and fix things on the way

**Further clean-up**
- [x] Deactivate `Isorropia` from our Trilinos configure script (see #1590)

## Interested Parties

@mayrmt @isteinbrecher @eulovi @m-frey @dharinib98

maxfirmbach:
Currently I'm observing issues with ParMETIS in the testing pipeline, which for now I can't reproduce locally.

maxfirmbach:
This issue is done and therefore can be closed. Thanks again for everyone helping out!

PR #1197, opened 2025-08-31
[Rebalance] Add unit tests for some edge cases

maxfirmbach:
## Description and Context

This PR adds some unit testing for edge cases that might happen during graph rebalancing, e.g.:
- Serial case: input graph should be equal to output graph (we should not guard with myrank == 0)
- One processor holds the full graph with others beeing empty: after rebalancing the empty procs should be populated
- Empty graph: should throw before rebalancing
- Empty discretization: should throw before building a monolithic graph

Part of #594

### DIFF
--- a/src/core/rebalance/tests/4C_rebalance_graph_test.cpp
+++ b/src/core/rebalance/tests/4C_rebalance_graph_test.cpp
@@ -1,0 +1,77 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#include <gtest/gtest.h>
+
+#include "4C_config.hpp"
+
+#include "4C_fem_discretization.hpp"
+#include "4C_rebalance_graph_based.hpp"
+
+#include <Teuchos_ParameterList.hpp>
+
+namespace
+{
+  using namespace FourC;
+
+  class RebalanceGraph : public ::testing::Test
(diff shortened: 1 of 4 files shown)


maxfirmbach:
This is just an initial effort to have unit testing for the rebalancing at all. Other cases will be added in the near future.

amgebauer (approved):

src/core/rebalance/tests/4C_rebalance_graph_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+    graph.optimize_storage();
+
+    Teuchos::ParameterList rebalance_params;
+    rebalance_params.set("partitioning method", "hypergraph");
+
+    auto rebalanced_graph = Core::Rebalance::rebalance_graph(graph, rebalance_params);
+
+    EXPECT_EQ(rebalanced_graph->num_global_nonzeros(), graph.num_global_nonzeros());
+    EXPECT_EQ(rebalanced_graph->num_local_nonzeros(), graph.num_local_nonzeros());
+    EXPECT_EQ(rebalanced_graph->num_global_nonzeros(), rebalanced_graph->num_local_nonzeros());
```
mayrmt, line 75: Can't' you do a `same_as()` query to compare both graphs?

maxfirmbach: Does that exist for graphs? I have never seen this method.

mayrmt: I thought that it exists. But you are right: it does not. Sorry.

PR #1198, opened 2025-09-01
Fix some destructors

amgebauer:
Some were not yet defauled.

### DIFF
--- a/src/constraint/4C_constraint_multipointconstraint.hpp
+++ b/src/constraint/4C_constraint_multipointconstraint.hpp
@@ -46,7 +46,7 @@ namespace Constraints
     /*!
         \brief Destructor
     */
-    virtual ~MPConstraint() { ; };
+    virtual ~MPConstraint() = default;
 
     /// Set state of the underlying constraint discretization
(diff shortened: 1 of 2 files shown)


src/particle_algorithm/4C_particle_algorithm_timint.hpp
```diff
@@ -64,7 +64,7 @@ namespace PARTICLEALGORITHM
      * \note At compile-time a complete type of class T as used in class member
      *       std::unique_ptr<T> ptr_T_ is required
      */
```
sebproell: If your change works indeed, please fix this comment. I believe it will not work tough, since the class has members `unique_ptr<T>` where T's are incomplete. Thus, the compiler cannot generate a default destructor without these. Maybe the comment could explain this a bit more clearly.

amgebauer: I didn't read the comment :grimacing:

sebproell (approved):

c-p-schmidt (approved):

PR #1204, opened 2025-09-01
Fix vtk node ordering for wedge6 elements

amgebauer:
This fixes the vtk node ordering in 4C. Previously, our output produced elements with negative volumes in Paraview.

I verified it by integrating 1 over an wedge6 element in Paraview. Previously the result was negative, now it is positive.

Part of #1195 

@isteinbrecher @eulovi


eulovi (approved):
Thanks for the fix @amgebauer!

PR #1207, opened 2025-09-02
Read cell and point ids from external mesh file

amgebauer:
* Allow reading external ids for cells and points
 * Read names of cell-blocks and point-sets
 * Remove side- and linesets (we can use `cell_blocks` with 2d or 1d cells)
 * use contiguous memory for the large arrays (points and cell connectivity) so that we can reserve the space upfront (and maybe use memcopy)


src/core/io/src/4C_io_mesh.hpp
```diff
@@ -51,13 +53,13 @@ namespace Core::IO::MeshInput
...
+  template <unsigned dim>
   struct Mesh
   {
     /**
-     * The points in the mesh. The keys are the point IDs, and the values are the coordinates of the
-     * points.
+     * The points in the mesh.
      */
-    std::map<int, std::vector<double>> points;
+    std::vector<std::array<double, dim>> points;
```
sebproell: I am in favor of this change but do we only have contiguous IDs? I am assuming we implicitly rely on this fact somewhere anyway (see my troubles in #733). I would like to also use the `Mesh` structure as the intermediate representation for meshes read from a legacy string.

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -89,73 +81,84 @@ namespace Core::IO::MeshInput
...
+          connectivity.size(), FE::cell_type_to_string(cell_type), FE::num_nodes(cell_type));
+
+      cells_connectivity_.insert(
+          cells_connectivity_.end(), connectivity.begin(), connectivity.end());
+    }
+
+    /*!
+     * @brief Returns a range for iterating over the cells in this block
+     */
+    [[nodiscard]] auto cells() const
```
sebproell:
```suggestion
     * @brief Returns a range for iterating over the cells in this block.
     * 
     *  Every range element contains a the point indices forming the cell.
     *
     */
    [[nodiscard]] auto cell_connectivities() const
```

src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -163,23 +163,17 @@ Core::IO::MeshInput::Mesh Core::IO::VTU::read_vtu_file(const std::filesystem::pa
...
-    {
-      // verify that the cell type is the same as the ones from all cells in this block
-      const auto existing_cell_type = mesh.cell_blocks[block_id].cell_type;
 
-      FOUR_C_ASSERT_ALWAYS(existing_cell_type == cell_type,
-          "Cell block {} has mixed cell types: {} and {}.", block_id,
-          Core::FE::cell_type_to_string(existing_cell_type),
-          Core::FE::cell_type_to_string(cell_type));
-    }
+    // Try to emplace the cell block into the mesh
```
sebproell:
```suggestion
```

src/global_data/4C_global_data_read.cpp
```diff
@@ -1477,6 +1478,9 @@ namespace
         {
           const auto& set = node_set.point_ids;
           node_sets[id] = std::vector<int>(set.begin(), set.end());
+
+          // 4C relies on a sorted node-set
+          std::ranges::sort(node_sets[id]);
```
sebproell: Seeing this, I definitely think the sorting should move into the node set itself rather than doing it here.

see https://github.com/4C-multiphysics/4C/pull/1208#discussion_r2316000376

src/core/io/src/4C_io_mesh.hpp
amgebauer: A simple solid-mechanics example runs with non-contiguous id. But do we need to do it? Especially with #733, it looks like we don't want such freedom in our code-base. Many formats don't provide the node-id anyway and you have to determine it youself anyway.

sebproell: I agree. I am not trying to say that they _should_ be non-contiguous. It was more the question if they _need_ to be non-contiguous for some reason. The numbering should be an implementation detail.

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -39,25 +43,36 @@ namespace Core::IO::MeshInput
...
   struct PointSet;
 
+  /*!
+   * @brief A point in the mesh
+   */
+  template <unsigned dim>
+  struct Point
+  {
+    /// External ID of the point as defined in the mesh
+    int external_id{};
```
sebproell: 1. Should we typedef this? I can see that this might be larger than an `int` someday.
2. Should the index be inside the `Point` or should this be an additional (optional) array in the `Mesh` struct?

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -67,97 +82,171 @@ namespace Core::IO::MeshInput
...
+      external_ids_.push_back(external_id);
+    }
+
+    /*!
+     * @brief Returns a range for iterating over the cell connectivities in this block
+     */
+    [[nodiscard]] auto cell_connectivities() const
+    {
+      auto indices = std::views::iota(std::size_t(0), size());
+      return indices | std::views::transform(
```
sebproell: The simplicity of this is appealing. I know you also suggested this construct to me instead of my recently introduced `NodeRef` and `ElementRef` together with `IteratorRange`. I think it makes sense to go for the std ranges in this PR, since we only use this in internal code, right? For the Discretization, I believe the extra types make things more readable. No idea, if there is a performance difference. We can continue monitoring this as we rework more Discretization internals 👍

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -67,97 +82,171 @@ namespace Core::IO::MeshInput
...
+      FOUR_C_ASSERT(i < size(),
+          "You are trying to access external ID of cell {} in a block with {} cells.", i, size());
+      return external_ids_[i];
+    }
+
+   private:
+    /*!
+     * Cells in this block. The cell connectivity is flattened to a 1D array.
+     */
+    std::vector<int> cells_connectivity_{};
```
sebproell: ~~Should we enhance `FlatVectorVector` to support a templated constant stride?~~ I see: a constexpr stride would not work here. So, it would need to be a constant.

PR #1208, opened 2025-09-02
Add some debug check

amgebauer:
We do a binary search on the nodes, thus, requiring that the nodes of a conditions are sorted. Let's assert that in debug mode.


src/core/fem/src/condition/4C_fem_condition.hpp
```diff
@@ -117,7 +117,12 @@ namespace Core::Conditions
     /*!
     \brief Set vector of my global node ids
     */
-    void set_nodes(const std::vector<int>& nodes) { nodes_ = nodes; }
+    void set_nodes(const std::vector<int>& nodes)
+    {
+      FOUR_C_ASSERT(
+          std::ranges::is_sorted(nodes), "4C requires the nodes of a condition to be sorted!");
```
sebproell: Alternative: sort the nodes in this function? If this object does the binary search, it seems like this is an internal detail that is not necessarily required outside.

mayrmt: ... and inform the user about it via screen output / warning.

sebproell: @mayrmt If we do the sorting here, there is no point in printing a warning, right?

mayrmt: You still might want to inform the user, that there is an operation, that you'd be avoided by pre-sorting during input generation. Yes, not a warning in the strict sense.

sebproell: The contract of this function can be either “give me a sorted array or I fail” or “give me anything ( I’ll sort it for my own use; you won't even know about this)”. The first errors on unsorted input, the second always passes (maybe errors if empty). I’d choose the second. Sortedness is either in the contract or it isn’t. What you suggest makes it not part of the contract, but still leaks internals.

amgebauer: The user can't do anything since whether the nodes end up sorted here or not is determined by the input (using `std::unordered_set` vs `std::set`). I'm now sorting the nodes in this function.

PR #1210, opened 2025-09-02
Enforce node numbering being consecutive starting at 1 during dat-style mesh input

amgebauer:
Control points needed to be consecutive anyways. Let's see whether we can enforce this for all nodes.

Related to #733

### DIFF
--- a/src/core/io/src/4C_io_meshreader.cpp
+++ b/src/core/io/src/4C_io_meshreader.cpp
@@ -505,51 +505,53 @@ namespace
     const int myrank = Core::Communication::my_mpi_rank(input.get_comm());
     if (myrank > 0) return;
 
-    int line_count = 0;
+    int expected_node_id = 0;
     for (const auto& node_line : input.in_section_rank_0_only(node_section_name))
     {
       Core::IO::ValueParser parser{
           node_line.get_as_dat_style_string(), {.user_scope_message = "While reading node data: "}};
       auto type = parser.read<std::string>();
 
+      int id = parser.read<int>() - 1;
+      FOUR_C_ASSERT_ALWAYS(id == expected_node_id,
+          "4C internally expects node IDs to be consecutive starting at 1! Expecting node id {} "
+          "but got {}",
+          expected_node_id + 1, id + 1);
+      ++expected_node_id;
+
       if (type == "NODE")
       {
-        int nodeid = parser.read<int>() - 1;
         parser.consume("COORD");
         auto coords = parser.read<std::vector<double>>(3);
 
-        max_node_id = std::max(max_node_id, nodeid) + 1;
+        max_node_id = std::max(max_node_id, id) + 1;
         std::vector<std::shared_ptr<Core::FE::Discretization>> dis =
-            find_dis_node(element_readers, nodeid);
+            find_dis_node(element_readers, id);
 
         for (const auto& di : dis)
         {
           // create node and add to discretization
           std::shared_ptr<Core::Nodes::Node> node =
-              std::make_shared<Core::Nodes::Node>(nodeid, coords, myrank);
+              std::make_shared<Core::Nodes::Node>(id, coords, myrank);
           di->add_node(node);
         }
       }
       // this node is a Nurbs control point
       else if (type == "CP")
       {
-        int cpid = parser.read<int>() - 1;
         parser.consume("COORD");
         auto coords = parser.read<std::vector<double>>(3);
         double weight = parser.read<double>();
 
-        max_node_id = std::max(max_node_id, cpid) + 1;
-        if (cpid != line_count)
-          FOUR_C_THROW(
-              "Reading of control points {} failed: They must be numbered consecutive!!", cpid);
+        max_node_id = std::max(max_node_id, id) + 1;
         std::vector<std::shared_ptr<Core::FE::Discretization>> diss =
-            find_dis_node(element_readers, cpid);
+            find_dis_node(element_readers, id);
 
         for (auto& dis : diss)
         {
           // create node/control point and add to discretization
           std::shared_ptr<Core::FE::Nurbs::ControlPoint> node =
-              std::make_shared<Core::FE::Nurbs::ControlPoint>(cpid, coords, weight, myrank);
+              std::make_shared<Core::FE::Nurbs::ControlPoint>(id, coords, weight, myrank);
           dis->add_node(node);
         }
       }
@@ -569,10 +571,9 @@ namespace
         std::vector<std::array<double, 3>> fibers;
         std::map<Core::Nodes::AngleType, double> angles;
 
-        int nodeid = parser.read<int>() - 1;
         parser.consume("COORD");
         auto coords = parser.read<std::vector<double>>(3);
-        max_node_id = std::max(max_node_id, nodeid) + 1;
+        max_node_id = std::max(max_node_id, id) + 1;
 
         while (!parser.at_end())
         {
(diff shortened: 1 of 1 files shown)


amgebauer:
I fixed all tests. I think it is worth restricting ourselves like that to make reading a mesh from any format as efficient as possible. Any thoughts on this?

sebproell:
This might break a lot of custom workflows, not sure. I totally agree that we need to move away from users having "control" over the IDs we use internally. Others are probably in a better situation to judge whether this can be merged.

amgebauer:
I don't think that it will break a lot of workflows. Adapted tests appeared to be manually crafted. Any script-based workflow in production likely generates contiguous point IDs anyways.

davidrudlstorfer (approved):

georghammerl:
Not sure whether this is important, our `grid_generator` does not provide consecutive node numbers if I get the implementation right, see https://github.com/4C-multiphysics/4C/blob/6b238ca9c3f8cc18dbd649569aa475c7324a2d3d/src/core/io/src/4C_io_gridgenerator.cpp#L391

Node numbers are consecutive for the `HEX27` layout, while for `HEX20` and `HEX8` some node ids are skipped.

Honestly, to me it feels like a step backwards to enforce consecutive node numbering. Depending on which tool is used to create meshes, it is not guaranteed that this requirement is fulfilled. I personally tend to provide a "node id renumberer" to clean meshes if our requirement is not fulfilled.

amgebauer:
> Not sure whether this is important, our `grid_generator` does not provide consecutive node numbers if I get the implementation right, see
> 
> https://github.com/4C-multiphysics/4C/blob/6b238ca9c3f8cc18dbd649569aa475c7324a2d3d/src/core/io/src/4C_io_gridgenerator.cpp#L391
> 
> Node numbers are consecutive for the `HEX27` layout, while for `HEX20` and `HEX8` some node ids are skipped.

The node ids of our grid generator are purely internal. So we can just make them consecutive.

> Honestly, to me it feels like a step backwards to enforce consecutive node numbering. Depending on which tool is used to create meshes, it is not guaranteed that this requirement is fulfilled. I personally tend to provide a "node id renumberer" to clean meshes if our requirement is not fulfilled.

Most mesh formats don't even allow to give points an id. And I think we should move in that direction, i.e., removing the id completely from the input. This is an intermediate step enforcing that we don't allow doing weird things with the id while still being (mostly) compatible with the original input.

Don't get me wrong, it is super easy to also allow specifying the id from the outside, but I don't know in which scenario this simplifies things.

amgebauer:
Let me add a little background: We want to unify our mesh input so that every format goes through the same interface. Currently, we basically have 3 mechanisms: dat-style input, vtu/ensight and grid generator. It would be nice if we can merge all three into one so that creating a mesh is more or less just creating the object `MeshInput::Mesh`. 

Currently we use a lot of `std::map` for our discretization to store nodes and elements. If we can replace these with `std::vector`, the underlying memory is much more flat (and if we have consecutive ids, the index is the id, so accessing elements by their id is super fast).

georghammerl:
>  Most mesh formats don't even allow to give points an id. And I think we should move in that direction, i.e., removing the id completely from the input. This is an intermediate step enforcing that we don't allow doing weird things with the id while still being (mostly) compatible with the original input.

I work a lot with Ansys, Nastran and Abaqus mesh files. All files that I have seen so far include node numbers. I see an import of those file formats as an important entrance step into 4C (for potential future users). Hence, it would be of importance to take care of node numbers.

I get the intention of your idea with using the more performant `std::vector` access. In the end it boils down to performance vs. flexibility. I tend not to loose the latter.

mayrmt:
> Most mesh formats don't even allow to give points an id.

I do not understand how you then identify nodes to assign them to elements and build up a mesh connectivity?

amgebauer:
@mayrmt 

>> Most mesh formats don't even allow to give points an id.
>
>I do not understand how you then identify nodes to assign them to elements and build up a mesh connectivity?

```xml
<?xml version="1.0"?>
<VTKFile type="UnstructuredGrid" version="0.1" byte_order="LittleEndian">
  <UnstructuredGrid>
    <Piece NumberOfPoints="8" NumberOfCells="1">
      <Points>
        <DataArray type="Float32" NumberOfComponents="3" format="ascii">
          0 0 0
          1 0 0
          1 1 0
          0 1 0
          0 0 1
          1 0 1
          1 1 1
          0 1 1
        </DataArray>
      </Points>
      <Cells>
        <DataArray type="Int32" Name="connectivity" format="ascii">
          0 1 2 3 4 5 6 7
        </DataArray>
        <DataArray type="Int32" Name="offsets" format="ascii">
          8
        </DataArray>
        <DataArray type="UInt8" Name="types" format="ascii">
          12
        </DataArray>
      </Cells>
    </Piece>
  </UnstructuredGrid>
</VTKFile>
```

Here is a simple vtu file (a single hex element). You cannot define the id of a point (node). You refer to it with its position in the list.


@georghammerl 

> I work a lot with Ansys, Nastran and Abaqus mesh files. All files that I have seen so far include node numbers. I see an import of those file formats as an important entrance step into 4C (for potential future users). Hence, it would be of importance to take care of node numbers.

The question is what we want to do with the ids. If we are only using it internally, we can just ignore any non-contiguous node-ids from the ansys mesh and go with our own id ordering. Then it is a pure implementation detail.

What do you have in mind where it is important that the node-ids in 4C must match with the ones provided in the mesh?

georghammerl:
> What do you have in mind where it is important that the node-ids in 4C must match with the ones provided in the mesh?

For example, according to the FKM guideline, stresses at the component's surface as well as stress gradients into the solid are needed. That means the user would likely write a script to extract values at the (known) node numbers, in order to perform further calculations. It would be cumbersome if the user would need to work with spatial positions in order to extract relevant values in case he/she cannot rely on the node numbers to be in original order.

georghammerl:
If we switch internally to consecutive node numbering, this is fine for me because it is hidden to the user. I think it is only important that a user can access all the simulation data with the originally provided node ids (and additionally any (error) output needs to use the original numbering). In other words: Any interaction of the user with the code/results should rely on the original numbering.

mayrmt:
> I work a lot with Ansys, Nastran and Abaqus mesh files. All files that I have seen so far include node numbers.

Same applies to the standardized ExodusII format, that we use in 4C, as far as I know.

amgebauer:
> Same applies to the standardized ExodusII format, that we use in 4C.

We have actually never used it and no-one complained so far. Only because a mesh format provides the id does not necessarily mean that we also use it.

If there are use-cases, of course, we can do that. Obviously in the 20 years history we also did not care very much about the ids. Printing the ids was never consistent because sometimes we print the raw internal ids and sometimes we translate it to the original ids (they are off by `1`, fixing it is not straigt forward, see #733).

If we at some point can do adaptive mesh refinement, we probably care even less about the node ids.

Also note, that some discretizations already now enforce this (NURBS). So we are even inconsistent.

mayrmt:
@amgebauer I'm fine with merging. Just wanted to add a format, that we use, to @georghammerl's list.

sebproell:
Sounds like:

- internally we want to freely renumber as we see fit, especially contiguously for good data layout
- but we need to remember the numbering that was given by users (if it is different)

I agree with both points. For now, all meshes are continuously numbered so we can rework some internals without losing anything in terms of functionality. If meshes with different numbering start to appear, we can implement a renumbering that you can query anytime you interact with user-facing errors, info, output etc. Does that sound about right?

mayrmt:
Thanks for the summary! Sounds good to me.

ischeider:
Sorry, I just noticed this PR. When having meshes from various sources, they may have arbitrary node numbering, also noncontinuous, and starting from arbitrary values. How do we deal with these? If read by meshio, the renumbering is done automatically (including node sets), same has to be done (manually) with vtu, if I am not mistaken. It seems that our input file options always will use node lists (not dicts), so that the PR is a logical consequence. However, users might be happy, if they find their original node numbers in the post processor.

amgebauer:
As discussed, we keep the info of the original ids within 4C and don't discard them during input. We might use different ids internally if that makes things easier for us, but we have a mapping from our internal ids to the ids for the output. Closing.

PR #1211, opened 2025-09-02
Write proper csv as DBC monitor output

c-p-schmidt:
## Description and Context
In preparation for the 4C workshop, I realized that the DBC monitor does not write proper CSV files, but some strange custom format. This is changed within this PR by:
- Use the RuntimeCSVWriter instead of a custom implementation
- remove unused option FILE_TYPE, as the second option data was never unused -> always csv now
- remove weird option WRITE_HEADER that is inconsistent with CSV format


sebproell:
@c-p-schmidt I believe I remember that @isteinbrecher depends on this.


edit: On second glance, maybe I misremembered that. I just remember we used to have some output related to monitoring that abused the `InputParameterContainer` for its print function :smile:

isteinbrecher:
Thanks for the mention @sebproell. Especially @knarfnitram and myself are using the DBC output in the current form.

Reformatting of the output file should not be an issue as we would only have to adapt one function. But as far as I can see, with the proposed changes, the node ID information is missing in the output file, which we require in our workflow.

I am not sure if this can easily be added to csv. Maybe writing a yaml file makes more sense here?

davidrudlstorfer:
@c-p-schmidt first of all - a big thanks for the work you put into the new form of the DBC monitor. This was a pain for myself for a long time, the old wannabe csv file was horrendous.

I also use the DBC monitor on a daily basis but I only need the information which is kept in the new csv file.

@isteinbrecher @knarfnitram wouldn't it just be possible to get the node ID information from the input file? There it is available and present.

EDIT: in the current state of the code (in beamme) it should easily be possible to just supply the input file besides the DBC monitor file. Then the node information is available as well.

knarfnitram:
My current workflows rely on the fact that I use [def read_dbc_monitor_file():](https://github.com/beamme-py/beamme/blob/1a5c6668b04703a2e54470e3e619b6a5a14cd4c0/src/beamme/four_c/dbc_monitor.py#L103C1-L103C38)  and do not have to manage or search again for the node ID's from previous created input file.
So @davidrudlstorfer , yes we could try to either reload it from a previous generated input file, but I think this can get error prune. So I think it would makes sense to include the information regarding the node IDs here. 
@c-p-schmidt would it be possible to add the node IDs as an additional line in the csv file ?

c-p-schmidt:
> My current workflows rely on the fact that I use [def read_dbc_monitor_file():](https://github.com/beamme-py/beamme/blob/1a5c6668b04703a2e54470e3e619b6a5a14cd4c0/src/beamme/four_c/dbc_monitor.py#L103C1-L103C38) and do not have to manage or search again for the node ID's from previous created input file. So @davidrudlstorfer , yes we could try to either reload it from a previous generated input file, but I think this can get error prune. So I think it would makes sense to include the information regarding the node IDs here. @c-p-schmidt would it be possible to add the node IDs as an additional line in the csv file ?

You mean you need this information:
`Nodes of this condition: 0 2 4 6 8 10 12 14 16 18 20`
for your workflow. Or also this part?
`Condition 100 Tagged elementsonoff : 1 1 1 1 1 1 1 1 1`

I'm a bit sceptical about adding an additional line to the CSV file because then loading the CSV file with standard tooling can't work properly (because the columns would most likely not match, and if the data included would not fit the description). What would certainly work is adding additional columns, but then you would have "number of nodeIDs" additional columns, with each containing one node ID. I'm not sure if you really want that?

Maybe a better solution would be to write out an additional CSV file with the same base-name but added _node_ids that only contains node ids, comma separated?

I'm not involved in BeamMe, so I can't judge whether it would be better suited there or in 4C itself.

c-p-schmidt:
The second commit adds a rather straightforward solution that you can obtain the info by just reusing the functionality that is already there and was also used for the file output before.

You now obtain a second file if the flag `WRITE_CONDITION_INFORMATION` is activated (used to be `WRITE_HEADER`), but as it is not really a header, I found the name too misleading to keep it.

The file names look as follows:
- xxx-101_monitor_dbc.csv (contains the data)
- xxx-101_monitor_dbc_condition.txt (contains the condition information)

The content of the second file looks as it used to look:

```bash
Condition 100 Tagged elements: ONOFF : 1 1 1 1 1 1 1 1 1
Nodes of this condition: 0 2 4 6 8 10 12 14 16 18 20
```

But obviously, I can't judge whether that is what you need or not, so please comment. ;-)

tests/input_files/beam3r_herm2line3_static_vdW_singlelengthsspec_smallsepapprox_simple_end_point_potential_reduction_strategy.4C.yaml
```diff
@@ -29,7 +29,7 @@ IO:
 IO/MONITOR STRUCTURE DBC:
   INTERVAL_STEPS: 1
   PRECISION_SCREEN: 3
-  WRITE_HEADER: true
+  WRITE_CONDITION_INFORMATION: true
```
davidrudlstorfer: I think you could remove this option here because the new additional file that is written is not actually tested.

(and the test was setup by myself - back then I did not know what this option did and it can be disabled for this tests :) )

davidrudlstorfer (dismissed):
Thanks a lot for the changes and going the extra mile and implementing the additional file output for the nodes.

@knarfnitram I think the additional file can be easily implemented?

isteinbrecher:
I am still wondering why we can’t write this to a yaml/json file, where one file contains all the relevant information, seems much cleaner to me.

If we want to go with this two file approach, we have to test the second file `xxx-101_monitor_dbc_condition.txt` during the pipeline, otherwise, we would decrease the test coverage of the dbc monitor with this PR.

Also, writing two files complicates some things. There is no way to know if a condition file “belongs” same named dbc file. It could be from a previous run and the subsequent run has WRITE_CONDITION_INFORMATION=false. We should make sure that no “condition” file from a previous run exists with the same name if WRITE_CONDITION_INFORMATION=false (either by throwing an error or removing that file). Otherwise, this could lead to some very hard to find bugs in complex workflows.

tests/input_files/beam3r_herm2line3_static_vdW_singlelengthsspec_smallsepapprox_simple_end_point_potential_reduction_strategy.4C.yaml
c-p-schmidt: Is changed in the last push.

c-p-schmidt:
Now, either YAML format or CSV format can be written. If the additional condition information is desired, a check is included that it has to be set to the YAML output format.

tests/input_files/ref/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_restart-101_monitor_dbc.yaml
```diff
@@ -0,0 +1,136 @@
+dbc monitor condition:
+  - condition information: "Condition 100 Tagged elements: ONOFF : 1 1 0 0 0 0 0 0 0 \nNodes of this condition:
```
isteinbrecher, line 2: We can think about changing this to full yaml structure in the future.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
```diff
@@ -13,17 +13,21 @@
...
 #include "4C_structure_new_monitor_dbc_input.hpp"
 #include "4C_structure_new_timint_basedataglobalstate.hpp"
 #include "4C_structure_new_timint_basedataio.hpp"
 #include "4C_structure_new_timint_basedataio_monitor_dbc.hpp"
-#include "4C_utils_enum.hpp"
 
-#include <Teuchos_StandardParameterEntryValidators.hpp>
+#include <c4/yml/parse.hpp>
+#include <c4/yml/tree.hpp>
+#include <ryml.hpp>
```
sebproell: These are unnecessary. I think.


```suggestion
```

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
```diff
@@ -149,35 +152,112 @@ void Solid::MonitorDbc::setup()
...
+        dbc_monitor_csvwriter_.back()->register_data_vector("curr_area", 1, of_precision_);
+        dbc_monitor_csvwriter_.back()->register_data_vector("f", DIM, of_precision_);
+        dbc_monitor_csvwriter_.back()->register_data_vector("m", DIM, of_precision_);
+        break;
+      }
+      case IOMonitorStructureDBC::FileType::yaml:
+      {
+        // handle restart
+        if (Global::Problem::instance()->restart() and
+            Core::Communication::my_mpi_rank(get_comm()) == 0)
```
isteinbrecher, line 169: What happens in the case that we don't write DBC output in the original simulation but in the restart one? I actually don't know if this was supported until now.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.hpp
```diff
@@ -15,19 +15,25 @@
 #include "4C_linalg_map.hpp"
 
 #include <mpi.h>
+#include <ryml.hpp>
```
sebproell:
```suggestion
#include "4C_io_yaml.hpp"
```

isteinbrecher (commented):
Thanks a lot @c-p-schmidt for the additional effort! I really like the possibility to select yaml files as well and having everything contained in one file.

I actually had a close look because I wanted to implement the nodes in yaml format and found two issues:
- The FILE_PRECSISION parameter does not affect the yaml file (I would be fine with just removing that parameter and always output the full precision to yaml/csv files).
- I think our comparison script for yaml files is wrong. I changed the condition information part in the yaml, but the test still passed.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
```diff
@@ -285,38 +311,78 @@ void Solid::MonitorDbc::execute(Core::IO::DiscretizationWriter& writer)
...
+        output_data["f"] = rforce_vec;
+        output_data["m"] = rmoment_vec;
+        dbc_monitor_csvwriter_[condition_counter]->write_data_to_file(
+            gstate_ptr_->get_time_n(), gstate_ptr_->get_step_n(), output_data);
+
+        break;
+      }
+      case IOMonitorStructureDBC::FileType::yaml:
+      {
+        if (Core::Communication::my_mpi_rank(get_comm()) == 0)
```
sebproell: I would use an early return here and invert the check.

isteinbrecher:
@c-p-schmidt you can have a look at my changes here https://github.com/4C-multiphysics/4C/commit/f9189b73c0b20f0b6fb162976fa740a1d0b1c656, they did not break the test that checks `tests/input_files/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_restart-101_monitor_dbc.yaml` and I think they should have.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
```diff
@@ -285,38 +311,78 @@ void Solid::MonitorDbc::execute(Core::IO::DiscretizationWriter& writer)
...
+          FILE* f = std::fopen(file_name.c_str(), "w");
+          if (!f)
+          {
+            std::perror("Failed to open file for writing");
+          }
+          else
+          {
+            emit_yaml(dbc_monitor_yaml_file_trees_[condition_counter], f);
+            std::fclose(f);
+          }
```
sebproell: Can you use C++ IO streams instead of C-style IO here? Alsom use `FOUR_C_THROW` instead of `perror` please.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.hpp
```diff
@@ -15,19 +15,25 @@
...
 namespace Core::FE
 {
   class Discretization;
 }  // namespace Core::FE
 namespace Solid
 {
+  namespace IOMonitorStructureDBC
+  {
+    enum class FileType : std::uint8_t;
+  }
```
sebproell: Just include the respective file.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
c-p-schmidt: Maybe I'm missing something, but all emit_yaml implementations take a FILE*. Is there another option?

sebproell: You are using the performant internal interface of `ryml`. Just do `stream << tree` which will call the function you are trying to call here.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
c-p-schmidt: It did not work before. It also throws an error now, see some lines later:
```bash
std::ifstream in(filename, std::ios::binary);
if (!in)
{
    FOUR_C_THROW("Could not open file {}", filename);
}
```

c-p-schmidt:
It is now possible to also compare yaml files, similar to what we already could with csv files.

We now have an additional check to verify that the file endings of the files to be compared are actually the same. If the file ending is not recognized, i.e. is neither .csv nor .yaml we also obtain an error.

For the comparison of yaml files I largely relied on FourCIPP capabilities, tailored to what we need here.

I added a third commit with wrong results in a csv and a yaml reference file to verify that it works.

c-p-schmidt:
The pipeline yesterday failed with the expected errors, thus the checking works.

I pushed without the last commit to remove the intentionally wrong changes to the reference files.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
isteinbrecher: Thanks for the clarification!

isteinbrecher (approved):
Thanks again for going the extra mile and adding the yaml output. Looks good to me.

isteinbrecher:
As for the required changes in BeamMe, I would first add the node IDs as a true yaml list in 4C and then adapt BeamMe.

PR #1212, opened 2025-09-02
Generalize `all_elements` validator to work on arbitrary ranges

sebproell:
Depends on #1178 and will be rebased

Prototyped together with @lauraengelhardt

### DIFF
--- a/src/core/io/src/4C_io_input_spec_validators.hpp
+++ b/src/core/io/src/4C_io_input_spec_validators.hpp
@@ -33,14 +33,78 @@ namespace Core::IO::InputSpecBuilders::Validators
 {
   namespace Internal
   {
+    template <typename T>
+    struct RangeLike
+    {
+      using value_type = T;
+    };
+
+    /**
+     * A type trait that abstracts certain types into a more general form. For example,
+     * a std::vector<T> is abstracted into RangeLike<T> to indicate that a validator may be
+     * applied to all elements of the vector.
+     */
+    template <typename T, typename Enable = void>
+    struct GeneralizedType;
+
+    template <typename T>
+    struct GeneralizedType<T, std::enable_if_t<IO::Internal::SupportedTypePrimitives<T>>>
+    {
+      using type = T;
+    };
+
+    template <typename T>
+    struct GeneralizedType<RangeLike<T>>
+    {
+      using type = RangeLike<typename GeneralizedType<T>::type>;
+    };
+
+    template <typename T>
+    struct GeneralizedType<std::vector<T>>
+    {
+      using type = RangeLike<typename GeneralizedType<T>::type>;
+    };
+
+    template <typename T, std::size_t n>
+    struct GeneralizedType<std::array<T, n>>
+    {
+      using type = RangeLike<typename GeneralizedType<T>::type>;
+    };
+
+    template <typename T>
+    struct GeneralizedType<std::optional<T>>
+    {
+      using type = std::optional<typename GeneralizedType<T>::type>;
+    };
+
+    template <typename T>
+    struct GeneralizedType<std::map<std::string, T>>
+    {
+      using type = std::map<std::string, typename GeneralizedType<T>::type>;
+    };
+
+    template <typename T1, typename T2>
+    struct GeneralizedType<std::pair<T1, T2>>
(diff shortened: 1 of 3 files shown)


src/core/io/src/4C_io_input_spec_validators.hpp
```diff
@@ -30,78 +31,202 @@ FOUR_C_NAMESPACE_OPEN
...
-   */
-  template <typename T>
-  class Validator
+  namespace Internal
   {
-    using Predicate = std::function<bool(const T&)>;
-    using Describe = std::function<void(std::ostream&)>;
-    using EmitMetadata = std::function<void(YamlNodeRef)>;
+    template <typename T>
+    struct RangeLike
```
sebproell, line 37: The tag type that we use instead of anything that can be a range...

src/core/io/src/4C_io_input_spec_validators.hpp
```diff
@@ -30,78 +31,202 @@ FOUR_C_NAMESPACE_OPEN
...
+    };
 
-    struct Concept
+    /**
+     * A type trait that abstracts certain types into a more general form. For example,
+     * a std::vector<T> is abstracted into RangeLike<T> to indicate that a validator may be
+     * applied to all elements of the vector.
+     */
+    template <typename T, typename Enable = void>
+    struct GeneralizedType;
```
sebproell, line 48: ... and this helper does the recursive conversion of something like `std::vector<std::vector<int>>` into
`RangeLike<RangeLike<int>>`.

src/core/io/src/4C_io_input_spec_validators.hpp
```diff
@@ -30,78 +31,202 @@ FOUR_C_NAMESPACE_OPEN
...
+    {
+      using type = std::tuple<typename GeneralizedType<Ts>::type...>;
+    };
+
+    template <typename T>
+    using GeneralizedTypeT = typename GeneralizedType<T>::type;
+
+
+    template <typename T, typename Enable = void>
+    class ValidatorImpl
```
sebproell, line 104: `Validator` is specialized for different cases. I moved this into `Internal` and now call it `ValidatorImpl`.

src/core/io/src/4C_io_input_spec_validators.hpp
```diff
@@ -146,6 +271,23 @@ namespace Core::IO::InputSpecBuilders::Validators
...
 
+  /**
+   * A generic validator providing the interface to check an input value against a predicate.
+   * This class is heavily tied to the InputSpec system and is used to validate input values in a
+   * type-erased manner. Usually, there is nothing you should do with this class directly. Use the
+   * various factory functions in this namespace to create instances of this class that can then be
+   * passed to one of the InputSpecBuilders functions.
+   */
+  template <typename T>
+  using Validator = Internal::ValidatorImpl<Internal::GeneralizedTypeT<T>>;
```
sebproell, line 282: @lauraengelhardt, here the magic happens. This alias would probably have saved us a ton of time if we had realized this earlier.

 Validator is now just an alias forwarding to the implementation class but generalizing the type! This means that `Validator<std::vector<int>> == Validator<std::array<int, 2>> == Validator<std::array<int, 1000>> == Validator<std::span<int>>`, i.e., they are all the same type without requiring any extra steps. Notably, we do not need to touch the `InputSpec` at all.

sebproell (commented):
@lauraengelhardt Thanks for your patience in trying to get this to work in #1178! We were really close this afternoon. As I suspected, starting from scratch, equipped with all we learned from our attempts, it was quite simple to finally get this working in a rather elegant way, if I may say so. Here is how:

src/core/io/src/4C_io_input_spec_validators.hpp
lauraengelhardt: Uff, yes 😆 Thanks for all the explanations and also for taking the time to look into this!
Nice to see this working out without too many changes!

lauraengelhardt (approved):
Great! As I already said in my comment above - thanks again for taking the time to work on this :) Looks good to me!

maxiludwig (approved):
Nothing to add from my side 👍

PR #637, opened 2025-04-09
Beaminteraction using Lagrange multipliers

dharinib98:
## Description and Context

This branch implements the functionality to solve the beam interaction  using the Lagrange Multiplier with the saddle point behaviour.

PR #1214, opened 2025-09-03
[beaminteraction] Small cleanup of the saddle point case

maxfirmbach:
## Description and Context

Removes methods specific  to the beam-contact module evaluator from the generic class. We should not put random methods in "almost" abstract classes. Does some tiny cleanup on the way.

Follows #637


dharinib98:
Thanks! This is also something we wanted to change as part of a cleaner layout and where to store the Lagrange maps.

mayrmt (approved):

PR #1216, opened 2025-09-03
Replace deprecated cmake-flags with current ones

c-p-schmidt:
## Description and Context
After #1147 there have been several occurrences of the deprecated cmake flags in our doc and presets. These have been changed to the current ones in this PR to not guide users to deprecated things.

However, the deprecated flags are still accepted, so this behavior is unchanged compared to #1147.

follows #1147


sebproell (approved):
Thank you for fixing!

PR #1219, opened 2025-09-03
[beam] Update error message in neuman boundary evaluation

maxfirmbach:
## Description and Context

Somehow the neumann load for `beam3eb` and `beam3k` can only be evaluated, when load linearization is done. The error message was quite unspecific, thus this PR adds a hint that `LOADLIN: true` has to be set.

I don't understand why a load linearization is necessary ... but otherwise the new displacement state is not retrievable. In the new structure model evaluator the new displacement state is explicitly only set for load linearization, otherwise not.

Actually `beam3r` calculates the neumann load completely different ... which is also strange ...

### DIFF
--- a/src/beam3/4C_beam3_euler_bernoulli_evaluate.cpp
+++ b/src/beam3/4C_beam3_euler_bernoulli_evaluate.cpp
@@ -235,7 +235,10 @@ int Discret::Elements::Beam3eb::evaluate_neumann(Teuchos::ParameterList& params,
   // get element displacements
   std::shared_ptr<const Core::LinAlg::Vector<double>> disp =
       discretization.get_state("displacement new");
-  if (disp == nullptr) FOUR_C_THROW("Cannot get state vector 'displacement new'");
+  if (disp == nullptr)
+    FOUR_C_THROW(
+        "Cannot get state vector 'displacement new'. "
+        "No linearization provided for load (add 'LOADLIN: true' to input file)");
   std::vector<double> mydisp = Core::FE::extract_values(*disp, lm);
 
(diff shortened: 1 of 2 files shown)


src/beam3/4C_beam3_euler_bernoulli_evaluate.cpp
```diff
@@ -235,7 +235,10 @@ int Discret::Elements::Beam3eb::evaluate_neumann(Teuchos::ParameterList& params,
   // get element displacements
   std::shared_ptr<const Core::LinAlg::Vector<double>> disp =
       discretization.get_state("displacement new");
-  if (disp == nullptr) FOUR_C_THROW("Cannot get state vector 'displacement new'");
+  if (disp == nullptr)
+    FOUR_C_THROW(
+        "Cannot get state vector 'displacement new'. "
+        "No linearization provided for load (add 'LOADLIN: true' to input file)");
```
mayrmt, line 241: This is a possible fix. But is it the actual and only reason? I just want to avoid, that the error message steers the user into the possibly wrong direction, if the actual reason is a different one.

mayrmt (commented):
Is it guaranteed that this is the only

src/beam3/4C_beam3_euler_bernoulli_evaluate.cpp
maxfirmbach: From a pure code perspective the answer would be: yes. I can't comment on the why though. The state `displacement new` is only set in two different other places related to load linearization, otherwise it doesn't appear.

mayrmt (approved):

Issue #1228, opened 2025-09-04
Move geometry representation from Node and Element into Discretization


sebproell:
# Context

The Discretization class consists of Elements and Nodes. Unfortunately, these classes 

- own the geometric representation instead of referring to data
- provide a ton of other functionality, notably all "element evaluation" derives directly from the `Element` class.

This is a meta issue to gather refactorings to move the geometric representation out of the fragmented object-oriented world and into data that the Disretization understands. Of course, it can still be interesting to loop over the nodes and elements of a Discretization, so this functionality will be kept but provided differently.

The benefit of this work should be a much simpler Discretization which we can give new features (deal.II conversion, more meshes as input, adaptivity, dynamic rebalancing, easier InputFields etc.)

# New data layout

_Current plan, may evolve further_

- data in `Discretization`:
  - node_coords: array of coordinate triples (size: n_local_nodes (owned+ghosted!))
    implictly defines a local(!) index for the nodes
  - element_connectivity: array of local(!) node indices (size: n_local_elements (owned+ghosted))
     ElementRef can trivially provide a node range by iterating over these indices
  - global_node_ids: associate IDs with node_coords (size n_local_nodes)
  - global_element_ids: associated GIDs for element_connectivity (size n_local_elements)
  - element_user_data: user pointers to data (size: n_local_elements)
     This enables to still use the old `Element` base class
  - node_user_data: user pointers to data (size: n_local_nodes)
     This enables to still use the old `Node` base class
- index arrays in `Discretization` to build iterator ranges
  - local_col_nodes_ids: array from `[0, n_local_nodes-1]` (implictly the position in node_coords, but need explicit indices for iterators)
  - local_row_node_ids: array containing only indices for nodes that are locally owned
  - local_col_element_ids: array from `[0, n_local_elements-1]` (implictly the position in element_connectivity but needed explicitly for iterators)
  - local_row_element_ids: array containing only indices for elements that are locally owned
  - adjacent_element_indices: array containing arrays with element indices (size: n_local_nodes)
    `NodeRef::adjacent_elements`will return range to iterate over these indices

- `NodeRef`:
  - provides acces to coord, lid, gid, user_data
  - `adjacent_elements()`

- `ElementRef`:
  - lid, gid, user_data
  - `nodes()`: iterate over `element_connectivity`

# Roadmap
- [x] Provide `DiscretizationIterator` capabilities to refer to a range of nodes or elements (#1229)
[...]

PR #1229, opened 2025-09-04
DiscretizationIterators for index-based iteration: apply to node ranges

sebproell:
Introduce a `DiscretizationIterator` which can walk over index arrays of the `Discretization`. The iterator gives you access to entities in the Discretization through a reference type. The PR demonstrates that this works by applying it to `my_row_node_range()` and `my_col_node_range()`, now giving a range of iterators that return `NodeRef` or `ConstNodeRef`. This reference type is meant to only provides basic info on nodes (coordinates, id, maybe adjacent elements). It is not meant to be inherited from. To stay backwards compatible, the old type (still called `Node`) is accessible via `user_data()`. I will need more PRs to roll this out for all node accesses.

This fixes a few const violations that were uncovered by the new implementation.

#1228

### DIFF
--- a/src/ale/4C_ale.cpp
+++ b/src/ale/4C_ale.cpp
@@ -138,12 +138,8 @@ void ALE::Ale::set_initial_displacement(const ALE::InitialDisp init, const int s
     {
       const Core::LinAlg::Map* dofrowmap = discret_->dof_row_map();
 
-      // loop all nodes on the processor
-      for (int lnodeid = 0; lnodeid < discret_->num_my_row_nodes(); lnodeid++)
+      for (auto lnode : discret_->my_row_node_range())
       {
-        // get the processor local node
-        Core::Nodes::Node* lnode = discret_->l_row_node(lnodeid);
-
         // the set of degrees of freedom associated with the node
         std::vector<int> nodedofset = discret_->dof(0, lnode);
 
@@ -156,7 +152,7 @@ void ALE::Ale::set_initial_displacement(const ALE::InitialDisp init, const int s
           // evaluate component d of function
           double initialval = Global::Problem::instance()
                                   ->function_by_id<Core::Utils::FunctionOfSpaceTime>(startfuncno)
-                                  .evaluate(lnode->x().data(), 0, d);
+                                  .evaluate(lnode.x().data(), 0, d);
 
           int err = dispn_->replace_local_value(doflid, initialval);
           if (err != 0) FOUR_C_THROW("dof not on proc");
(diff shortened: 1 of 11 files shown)


src/core/fem/src/general/node/4C_fem_general_node.hpp
```diff
@@ -341,6 +341,77 @@ namespace Core::Nodes
...
+    NodeRefImpl(DiscretizationType* dis, int local_id) : discretization_(dis), local_id_(local_id)
+    {
+      FOUR_C_ASSERT(dis, "NodeRef needs a valid discretization");
+    }
+
+    [[nodiscard]] std::span<const double, 3> x() const
+    {
+      const auto& coord = discretization_->nodecolptr_[local_id_]->x();
+      FOUR_C_ASSERT(coord.size() == 3, "NodeRef only supports 3D coordinates");
+      return std::span<const double, 3>(coord.data(), 3);
```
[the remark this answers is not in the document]
sebproell: I'd rather have this stated here explicitly since it is a fact of the current implementation.

src/core/fem/src/general/node/4C_fem_general_node.hpp
```diff
@@ -341,6 +341,117 @@ namespace Core::Nodes
...
 
   };  // class Node
+
+
+
+  /**
+   * Implementation helper for NodeRef and ConstNodeRef.
+   */
+  template <bool is_const>
+  class NodeRefImpl
```
amgebauer, line 351: Maybe move this into an `Internal` namespace?

sebproell: I am unsure what the best approach is. This class provides the actual methods you are supposed to call as a user and the two deriving classes just "specialize" the constness -- a standard technique, AFAIK.

The question whether this is `Internal` boils down to what exactly `Internal`  is meant to convey. For me, it mostly means "no stable interfaces in here, don't try to use this directly". In that case, one could argue that `NodeRef` and `ConstNodeRef` should also move into `Internal`, since you should never create them yoursefl. You will however interact with all three classes (mostly through `auto` types of course).

If this makes sense to you, would it be OK if I leave it as is for now? I will do a lot more iterations (possibly changing the implementation) and as we learn more I hope it becomes clearer where this goes.

amgebauer: I'm fine with that.

I typically use `Internal` to indicate you should not use this class. I might define a `using`-statement outside of the `Internal` namespace for the special cases meant to be used.

bgoderbauer (approved):

PR #1231, opened 2025-09-05
Add FlatVectorVector<T> data structure

sebproell:
A performant vector<vector<T>> storing data and offsets

### DIFF
--- a/src/core/utils/src/stl_extension/4C_utils_flat_vector_vector.hpp
+++ b/src/core/utils/src/stl_extension/4C_utils_flat_vector_vector.hpp
@@ -1,0 +1,112 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_UTILS_FLAT_VECTOR_VECTOR_HPP
+#define FOUR_C_UTILS_FLAT_VECTOR_VECTOR_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_utils_exceptions.hpp"
+
+#include <span>
+#include <vector>
+
+FOUR_C_NAMESPACE_OPEN
+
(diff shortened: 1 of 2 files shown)


src/core/utils/src/stl_extension/4C_utils_flat_vector_vector.hpp
```diff
@@ -0,0 +1,112 @@
...
+    //! Default constructor. Creates an empty FlatVectorVector.
+    FlatVectorVector() = default;
+
+    //! @brief Construct from a vector-of-vectors
+    explicit FlatVectorVector(const std::vector<std::vector<T>>& nested) { from_nested(nested); }
+
+    //! @brief Replace contents from a vector-of-vectors
+    void from_nested(const std::vector<std::vector<T>>& nested)
+    {
+      offsets_.resize(nested.size() + 1);
```
amgebauer, line 47: You could also skip the `0` as the first entry

amgebauer (approved):

src/core/utils/src/stl_extension/4C_utils_flat_vector_vector.hpp
sebproell: The last one can be dropped with the same argument. The reason the obvious `0` and `size()` entries are included is to enable `span(offset[i], offset[i+1] - offset[i])` for every valid `i` without additional conditionals.

mayrmt:
I am curious:

- What's the actual performance benefit?
- how is such a design related to a two-dimensional `Kokkos::View`?

sebproell:
I believe these questions should, to some extent, be answered by the docs.

> What's the actual performance benefit?

The design just acknowledges that it is better to store a single index per inner vector compared to three pointers encoding the inner `std::vector` of a `vector<vector>`. Furthermore, the data has better locality. I made no concrete measurements here, but it is hard to think about a case where the performance is worse. I know: always measure.

> how is such a design related to a two-dimensional Kokkos::View?

`FlatVectorVector` is "ragged" and not multi-dimensional with fixed extents.

lauraengelhardt (approved):

PR #1232, opened 2025-09-05
Fix constness of SparseMatrix

amgebauer:
The sparse matrix reveiled the underlying non-const Epetra Matrix via `epetra_matrix()`.

I had to add two const_cast in the interface to Trilinos.

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_create.cpp
+++ b/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_create.cpp
@@ -10,6 +10,7 @@
 #include "4C_fem_discretization.hpp"
 #include "4C_utils_exceptions.hpp"
 
+#include <Epetra_CrsMatrix.h>
 #include <MueLu_AmalgamationFactory.hpp>
 #include <MueLu_CoalesceDropFactory.hpp>
 #include <MueLu_CoarseMapFactory.hpp>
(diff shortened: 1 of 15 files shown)


src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_create.cpp
```diff
@@ -72,8 +73,8 @@ Core::LinAlg::SparseMatrix Core::LinAlg::create_interpolation_matrix(const Spars
           Teuchos::rcp(new Epetra_MultiVector(nullspace.get_epetra_multi_vector()))));
   const int number_of_equations = params.get<int>("PDE equations");
 
-  Teuchos::RCP<Xpetra::CrsMatrix<SC, LO, GO, NO>> mueluA =
-      Teuchos::rcp(new EpetraCrsMatrix(Teuchos::rcpFromRef(*matrix.epetra_matrix())));
+  Teuchos::RCP<Xpetra::CrsMatrix<SC, LO, GO, NO>> mueluA = Teuchos::rcp(new EpetraCrsMatrix(
+      Teuchos::rcpFromRef(const_cast<Epetra_CrsMatrix&>(matrix.epetra_matrix()))));
   Teuchos::RCP<Xpetra::Matrix<SC, LO, GO, NO>> xpetra_matrix =
       Teuchos::rcp(new Xpetra::CrsMatrixWrap<SC, LO, GO, NO>(mueluA));
   xpetra_matrix->SetFixedBlockSize(number_of_equations);
```
amgebauer, line 80: Probably Trilinos is not going to modify the matrix. @maxfirmbach Could you double check?

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -296,7 +296,9 @@ std::shared_ptr<Core::LinAlg::SparseMatrix> Core::LinAlg::matrix_transpose(const
   EpetraExt::RowMatrix_Transpose transposer;
   std::shared_ptr<Core::LinAlg::SparseMatrix> matrix = nullptr;
 
-  Epetra_CrsMatrix* a_prime = &(dynamic_cast<Epetra_CrsMatrix&>(transposer(*A.epetra_matrix())));
+  // Transposer does not modify the matrix but only provides a non-const interface
+  Epetra_CrsMatrix* a_prime = &(dynamic_cast<Epetra_CrsMatrix&>(
+      transposer(const_cast<Epetra_CrsMatrix&>(A.epetra_matrix()))));
```
amgebauer, line 301: Same here, as far as I know, the original matrix will not be touched.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_create.cpp
maxfirmbach: Correct, original matrix is not touched.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
maxfirmbach: Yes, also here the original matrix is not modified.

mayrmt (dismissed):

PR #1234, opened 2025-09-05
Enhance DiscretizationIterator functionality 

sebproell:
Since we iterate over an index array with this Iterator, we can

- make it a random access iterator
- enable views on the iterator range
- provide a make_iterator_range helper

#1228

### DIFF
--- a/src/core/fem/src/discretization/4C_fem_discretization_iterator.hpp
+++ b/src/core/fem/src/discretization/4C_fem_discretization_iterator.hpp
@@ -10,21 +10,14 @@
 
 #include "4C_config.hpp"
 
-#include <concepts>
 #include <iterator>
+#include <span>
 
 FOUR_C_NAMESPACE_OPEN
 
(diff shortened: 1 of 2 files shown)


amgebauer (approved):
Nice!

src/core/fem/src/discretization/4C_fem_discretization_iterator.hpp
```diff
@@ -10,22 +10,15 @@
...
 namespace Core::FE
 {
   class Discretization;
 
-  /**
-   * Check whether T is a type that can be constructed from a Discretization pointer and an int.
-   */
-  template <typename T>
-  concept DiscretizationReferenceType = std::constructible_from<T, Discretization*, int> ||
-                                        std::constructible_from<T, const Discretization*, int>;
```
sebproell, line 27: @bgoderbauer I removed the unused concept, as you suggested offline.

PR #1236, opened 2025-09-05
Add AAA (Adaptive Antoulas-Anderson) rational approximation implementation to ReducedLung

bgoderbauer:
## Description

This PR introduces an implementation of the original AAA algorithm for rational approximation, introduced by Nakatsukasa, Sète, and Trefethen (SIAM SISC, 2018) ([Link](https://epubs.siam.org/doi/10.1137/16M1106122)).

The AAA algorithm computes a rational approximation in barycentric form of a function evaluated at a given set of support points. I need it in later implementations of material models. I will also extend it to work with complex values and arbitrary precision in the future. At the current stage, there are also a few utility functions for further utilizing the resulting barycentric approximation. I tried to test all critical steps to show that it's working, since there is no direct usage in 4C yet at the moment.

- The first commit moves all implementation in `ReducedLung` into a subfolder `src` (I've forgotten it in the past).
- The second commit refactors the return type of the generalized eigenvalue implementation with `SerialDenseMatrix` a little, since I need the full eigenvalue vector.
- The third commit is then the full AAA implementation.


src/core/linalg/src/dense/4C_linalg_utils_densematrix_eigen.cpp
```diff
@@ -275,21 +270,40 @@ double Core::LinAlg::generalized_eigen(
...
-      // Eigenvalues:
-      // std::cout << "lambda " << i << ":  " <<  ALPHAR[i]/BETA[i] << std::endl;
-      maxlambda = std::max(ALPHAR[i] / BETA[i], maxlambda);
+      ++num_inf;
+      continue;  // skip infinite eigenvalue
     }
     if (ALPHAI[i] > 1e-12)
     {
-      std::cout << " Warning: you have an imaginary EW " << ALPHAI[i] << std::endl;
+      std::cout << " Warning: you have an imaginary EW " << ALPHAI[i] << '\n';
```
sebproell, line 276: What ist `EW`?

src/core/linalg/src/dense/4C_linalg_utils_densematrix_eigen.cpp
```diff
@@ -275,21 +270,40 @@ double Core::LinAlg::generalized_eigen(
    std::cout << std::setprecision(16) << "Q 2 " << Q_2 << std::endl;
    std::cout << std::setprecision(16) << "Z 2 " << Z_2 << std::endl;*/
```
sebproell: Would you mind removing these commented lines?

src/core/linalg/src/dense/4C_linalg_utils_densematrix_eigen.cpp
```diff
@@ -275,21 +270,40 @@ double Core::LinAlg::generalized_eigen(
...
     if (ALPHAI[i] > 1e-12)
     {
-      std::cout << " Warning: you have an imaginary EW " << ALPHAI[i] << std::endl;
+      std::cout << " Warning: you have an imaginary EW " << ALPHAI[i] << '\n';
     }
+    eigenvalues.emplace_back(ALPHAR[i] / BETA[i], ALPHAI[i] / BETA[i]);
+  }
+  if (num_inf > 0)
+    std::cout << "Generalized eigenvalue computation: " << num_inf
+              << " infinite eigenvalue(s) discarded." << '\n';
```
sebproell: Printing warnings in such low-level algorithms is not great since it limits usability. Can you return the data in a sorted way and remove the warning? Maybe you can have the real, imaginary, "infinite" etc. eigenvalues in separate vectors?

src/reduced_lung/src/4C_reduced_lung_aaa_approx.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace ReducedLung
+{
+  /**
+   * @brief Options for the AAA (Adaptive Antoulas-Anderson) rational approximation.
+   *
+   * Controls the stopping tolerance and the maximum number of support points
+   * used by the greedy AAA iteration.
```
sebproell:
```suggestion
```

Just document the options inside. You will for sure forget to update this header; at least I always do ;)

src/reduced_lung/src/4C_reduced_lung_aaa_approx.hpp
```diff
@@ -0,0 +1,131 @@
...
+    int mmax = 100;
+  };
+
+  /**
+   * @brief Result of an AAA approximation in barycentric form.
+   *
+   * Holds the support points, function values, and barycentric weights that
+   * define the rational approximant, plus a per-iteration error history.
+   * Utility methods allow evaluation, pole/zero extraction, and gain.
+   *
```
sebproell: Not necessary. You don't need to restate what is written below. If there is other information about the interplay of these parameters, you can of course add it here.


```suggestion
```

src/reduced_lung/src/4C_reduced_lung_aaa_approx.hpp
```diff
@@ -0,0 +1,131 @@
...
+    std::vector<double> w;
+
+    /// Max-norm error per AAA iteration on the sample grid.
+    std::vector<double> errvec;
+
+    /**
+     * @brief Evaluate the rational approximant at given points.
+     * Exact support-point queries return the stored f_j.
+     */
+    std::vector<double> operator()(const std::vector<double>& zz) const;
```
sebproell:
```suggestion
    std::vector<double> operator()(std::span<const double> zz) const;
```

src/reduced_lung/src/4C_reduced_lung_aaa_approx.hpp
```diff
@@ -0,0 +1,131 @@
...
+   * @param w       Barycentric weights (real).
+   * @param tiny_abs Absolute fallback threshold for declaring the denominator
+   *                 derivative effectively zero (0 = derive a relative threshold).
+   * @param eps_rel  Relative proximity threshold to treat a pole as colliding
+   *                 with a support point.
+   * @return Residues in the same order as @p poles.
+   *
+   */
+  std::vector<std::complex<double>> compute_residues(const std::vector<std::complex<double>>& poles,
+      const std::vector<double>& z, const std::vector<double>& f, const std::vector<double>& w,
```
sebproell: You have three parameters of the same type in a row, which makes it easy to pass parameters in the wrong order. You could think about a struct to give them names.

src/reduced_lung/src/4C_reduced_lung_aaa_approx.hpp
```diff
@@ -0,0 +1,131 @@
...
+      const double tiny_abs = 0.0, const double eps_rel = 1e-14);
+
+  using EvaluationFunction = std::function<std::vector<double>(const std::vector<double>& x)>;
```
sebproell:
```suggestion
  using EvaluationFunction = std::function<std::vector<double>(std::span<const double> x)>;
```

src/reduced_lung/src/4C_reduced_lung_aaa_approx.cpp
```diff
@@ -0,0 +1,287 @@
...
+      B(i, i) = 1.0;
+    }
+
+    return Core::LinAlg::generalized_eigen(A, B);
+  }
+
+  double AAAResult::compute_gain() const
+  {
+    double num = 0.0;
+    double den = 0.0;
```
sebproell, line 95:
```suggestion
    double den = 0.0;
    FOUR_C_ASSERT(f.size() == w.size(),...);
```

src/reduced_lung/src/4C_reduced_lung_aaa_approx.cpp
```diff
@@ -0,0 +1,287 @@
...
+      for (int i = 0; i < M; ++i) err = std::max(err, std::abs((F[i] - R[i]) / f_max));
+      errvec.push_back(err);
+      if (err <= opts.tol) break;
+    }
+
+    AAAResult result;
+    result.z = std::move(z);
+    result.f = std::move(f);
+    result.w = std::move(w);
+    result.errvec = std::move(errvec);
```
sebproell: You could also create the `result` at the top and fill the values in there during the algorithm.

src/reduced_lung/tests/4C_reduced_lung_aaa_approx_test.cpp
```diff
@@ -0,0 +1,458 @@
...
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace
+{
+
+  // Helper: compute max abs error between two vectors
+  double max_error(const std::vector<double>& a, const std::vector<double>& b)
+  {
+    assert(a.size() == b.size());
```
sebproell:
```suggestion
    ASSERT_EQ(a.size(), b.size());
```

src/reduced_lung/tests/4C_reduced_lung_aaa_approx_test.cpp
```diff
@@ -0,0 +1,458 @@
...
+      return y;
+    };
+
+    const auto result = ReducedLung::aaa(Z, F);
+
+    // Check interpolation: r(z_k) == f_k
+    const auto val = result(result.z);
+    for (size_t k = 0; k < result.z.size(); ++k)
+    {
+      EXPECT_NEAR(abs(val[k] - result.f[k]), 0.0, 1e-14)
```
sebproell:
```suggestion
      EXPECT_DOUBLE_EQ(abs(val[k] - result.f[k]), 0.0)
```

src/reduced_lung/tests/4C_reduced_lung_aaa_approx_test.cpp
```diff
@@ -0,0 +1,458 @@
...
+    const double disc = B * B - 4 * A * C;
+    ASSERT_GT(disc, 0.0);
+    const double r1 = (-B + std::sqrt(disc)) / (2 * A);
+    const double r2 = (-B - std::sqrt(disc)) / (2 * A);
+
+    const auto poles = R.compute_poles();
+    ASSERT_EQ(poles.size(), 2u);
+
+    std::vector<std::complex<double>> got = poles, exp = {std::complex<double>(r1, 0),
+                                                       std::complex<double>(r2, 0)};
```
sebproell: Please don't put two declarations in one line. This is hard to read.

sebproell (commented):
Interesting feature!

src/core/linalg/src/dense/4C_linalg_utils_densematrix_eigen.cpp
bgoderbauer: Thanks for the hint, I just read over it. It should be called eigenvalue.

src/core/linalg/src/dense/4C_linalg_utils_densematrix_eigen.cpp
bgoderbauer: I don't think infinite eigenvalues have any use atm since we don't calculate the respective eigenvectors. I'll just remove the warning. The returned eigenvalues are complex, so I think there is no need to distinguish real and imaginary eigenvalues.

src/reduced_lung/src/4C_reduced_lung_aaa_approx.hpp
bgoderbauer: I changed it to directly take the `AAAResult` struct with all necessary vectors inside. This should be more convenient (ultimately, you'll always need to call `aaa()` first before computing anything else :smile:)

src/reduced_lung/tests/4C_reduced_lung_aaa_approx_test.cpp
bgoderbauer: I removed it completely since it is only a helper function and `ASSERT_EQ` has an interfering `void` return type

bgoderbauer:
Thanks for the review and the helpful comments @sebproell! I addressed all your suggestions. Looking forward to further opinions! :smiley:

sebproell (approved):

src/reduced_lung/src/4C_reduced_lung_aaa_approx.hpp
```diff
@@ -0,0 +1,122 @@
...
+   * @param aaa_result  result from an AAA approximation with support points z, function values at
+   * support points f, and barycentric weights w.
+   * @param tiny_abs Absolute fallback threshold for declaring the denominator
+   *                 derivative effectively zero (0 = derive a relative threshold).
+   * @param eps_rel  Relative proximity threshold to treat a pole as colliding
+   *                 with a support point.
+   * @return Residues in the same order as @p poles.
+   */
+  std::vector<std::complex<double>> compute_residues(const std::vector<std::complex<double>>& poles,
+      const AAAResult& aaa_result, const double tiny_abs = 0.0, const double eps_rel = 1e-14);
```
amgebauer, line 94: These two consts don't have any effect
```suggestion
      const AAAResult& aaa_result, double tiny_abs = 0.0, double eps_rel = 1e-14);
```

PR #1237, opened 2025-09-05
Add ElementRef to Discretization

sebproell:
Add `(Const)ElementRef` similar to `(Const)NodeRef` #1229. I decided to move the `Ref` classes to the Discretization header, since they actively refer into the Discretization (and not into `Node` or `Element`).

Part of #1228 

Depends on #1231 #1234


src/core/fem/src/discretization/4C_fem_discretization.hpp
```diff
@@ -76,6 +77,217 @@ namespace Core::IO
...
+     */
+    [[nodiscard]] int global_id() const { return discretization_->nodecolptr_[local_id_]->id(); }
+
+    /**
+     * Return a pointer to user data of type Node. This may be nullptr.
+     */
+    [[nodiscard]] MaybeConst<Nodes::Node>* user_node() const
+    {
+      return discretization_->nodecolptr_[local_id_];
+    }
```
amgebauer, line 139: Don't you want to have this method non-const if it is returning a non-const pointer?
```suggestion
    [[nodiscard]] MaybeConst<Nodes::Node>* user_node()
    {
      return discretization_->nodecolptr_[local_id_];
    }
    
    [[nodiscard]] const Nodes::Node* user_node() const
    {
      return discretization_->nodecolptr_[local_id_];
    }
```

sebproell: No, not really. This type has the semantics of a reference. Think about a reference in C++. Const-qualifying a reference has no effect (or you can't even do it, depending on the context) since a reference is always immutable. To be clear, I am talking about something like `T& const`, which is not even allowed, rather than `const T&`. Alternatively, look at the implementation of iterators in STL (e.g. https://en.cppreference.com/w/cpp/iterator/reverse_iterator/operator%252A.html) where the accessors are unconditionally `const`.

amgebauer: yes, you are right :+1:

sebproell:
Can I get another review here?

bgoderbauer (approved):

PR #1238, opened 2025-09-06
Evaluate initial acceleration with full RHS vector

amgebauer:
During the setup of the new structural framework, we evaluate the RHS to compute the initial accelerations. The vector we created previously was only large enough to accommodate the solid RHS. However, for cardiovascular0D-BCs and Lagrange constraints, this vector was too short to assemble their RHS. In these modules, we had a peculiar check to determine whether to assemble their RHS, which depended on the node numbering at input.

This issue has now been resolved: we create an RHS of the appropriate size and assemble the entire RHS.

I also removed unnecessary pointers on the way.

Related to #733

### DIFF
--- a/src/structure_new/src/4C_structure_new_integrator.cpp
+++ b/src/structure_new/src/4C_structure_new_integrator.cpp
@@ -177,14 +177,8 @@ void Solid::Integrator::compute_mass_matrix_and_init_acc()
   check_init();
 
   // temporary right-hand-side
-  std::shared_ptr<Core::LinAlg::Vector<double>> rhs_ptr =
-      std::make_shared<Core::LinAlg::Vector<double>>(*global_state().dof_row_map_view(), true);
-
-  // initialize a temporal structural stiffness matrix
-  Teuchos::RCP<Core::LinAlg::SparseOperator> stiff_ptr =
-      Teuchos::make_rcp<Core::LinAlg::SparseMatrix>(
-          *global_state().dof_row_map_view(), 81, true, true);
-
+  Core::LinAlg::Vector<double> rhs_full =
+      Core::LinAlg::Vector<double>(global_state().global_problem_map(), /*zero_out=*/true);
 
   // overwrite initial state vectors with Dirichlet BCs
   // note that we get accelerations resulting from inhomogeneous Dirichlet conditions here
@@ -204,7 +198,7 @@ void Solid::Integrator::compute_mass_matrix_and_init_acc()
   eval_data().set_total_time(gstate_ptr_->get_time_n());
 
   // initialize the mass matrix and the Rayleigh damping matrix (optional)
-  if (not model_eval().initialize_inertia_and_damping(*disnp_ptr, *stiff_ptr))
+  if (not model_eval().initialize_inertia_and_damping(*disnp_ptr))
     FOUR_C_THROW("initialize_inertia_and_damping failed!");
 
   /* If we are restarting the simulation, we do not have to calculate a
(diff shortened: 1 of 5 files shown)


src/structure_new/src/4C_structure_new_integrator.cpp
```diff
@@ -177,13 +177,11 @@ void Solid::Integrator::compute_mass_matrix_and_init_acc()
...
-  std::shared_ptr<Core::LinAlg::Vector<double>> rhs_ptr =
-      std::make_shared<Core::LinAlg::Vector<double>>(*global_state().dof_row_map_view(), true);
+  Core::LinAlg::Vector<double> rhs_full =
+      Core::LinAlg::Vector<double>(global_state().global_problem_map(), /*zero_out=*/true);
 
-  // initialize a temporal structural stiffness matrix
-  Teuchos::RCP<Core::LinAlg::SparseOperator> stiff_ptr =
-      Teuchos::make_rcp<Core::LinAlg::SparseMatrix>(
-          *global_state().dof_row_map_view(), 81, true, true);
+  // initialize a temporal matrix
```
mayrmt: "temporal" probably means "temporal" or "auxiliary", doesn't it?

mayrmt (dismissed):

src/structure_new/src/4C_structure_new_integrator.cpp
amgebauer: The matrix itself was obviously more or less a leftover from an older state. I removed it.

mayrmt (approved):

src/structure_new/src/4C_structure_new_integrator.cpp
amgebauer: Obviously this matrix was actually needed (I should have read the comment :grimacing: )

src/structure_new/src/4C_structure_new_integrator.cpp
```diff
@@ -177,14 +177,8 @@ void Solid::Integrator::compute_mass_matrix_and_init_acc()
...
-  std::shared_ptr<Core::LinAlg::Vector<double>> rhs_ptr =
-      std::make_shared<Core::LinAlg::Vector<double>>(*global_state().dof_row_map_view(), true);
-
-  // initialize a temporal structural stiffness matrix
-  Teuchos::RCP<Core::LinAlg::SparseOperator> stiff_ptr =
-      Teuchos::make_rcp<Core::LinAlg::SparseMatrix>(
-          *global_state().dof_row_map_view(), 81, true, true);
-
+  Core::LinAlg::Vector<double> rhs_full =
+      Core::LinAlg::Vector<double>(global_state().global_problem_map(), /*zero_out=*/true);
```
davidrudlstorfer, line 181:
```suggestion
      Core::LinAlg::Vector<double>(global_state().global_problem_map(), true);
```

Is this wanted?

amgebauer: yes, makes it clear what this magic boolean parameter is doing.

mayrmt (approved):

src/structure_new/src/4C_structure_new_integrator.cpp
sebproell: Yes, since the interface is not very telling. An enum constant would be nicer but I prefer this comment. Clang warns if the comment name is not equal to the parameter name.

davidrudlstorfer: Thanks for the explanations - didn't know about the clang warnings :)

PR #1239, opened 2025-09-06
Validate muscle and St. Venant Kirchhoff material input via InputSpecBuilders::Validators

lauraengelhardt:
## Description and Context
- Use `Core::IO::InputSpecBuilders::Validators` to validate:
  - **St. Venant–Kirchhoff material model** (commit 1)  
  - **All muscle materials** (commits 2 + 3, to be squashed/merged together)  
- Fix a documentation error in `Core::IO::InputSpecBuilders::deprecated_selection` (commit 4).  

## Open Question
I often use:  
```cpp
using namespace Core::IO::InputSpecBuilders::Validators;
auto positive_or_zero_validator =
    in_range<double>(0.0, std::numeric_limits<double>::max());
```

Should we introduce a `positive_or_zero()` helper just as we have the `positive()` helper function?

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
@@ -1044,10 +1044,14 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
   /*--------------------------------------------------------------------*/
   // St.Venant--Kirchhoff
   {
+    using namespace Core::IO::InputSpecBuilders::Validators;
+
     known_materials[Core::Materials::m_stvenant] = group("MAT_Struct_StVenantKirchhoff",
         {
-            parameter<double>("YOUNG", {.description = "Young's modulus"}),
-            parameter<double>("NUE", {.description = "Poisson's ratio"}),
+            parameter<double>(
+                "YOUNG", {.description = "Young's modulus", .validator = positive<double>()}),
+            parameter<double>("NUE",
+                {.description = "Poisson's ratio", .validator = in_range<double>(-1.0, excl(0.5))}),
             parameter<double>("DENS", {.description = "mass density"}),
         },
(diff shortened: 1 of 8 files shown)


sebproell:
> Should we introduce a positive_or_zero() helper just as we have the positive() helper function?

Yes. It might also be called `non_negative()`. Not sure which of the two I like better.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -454,28 +472,39 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
                 "KAPPA", {.description = "material parameter for coupled volumetric contribution"}),
             parameter<double>(
-                "OMEGA0", {.description = "weighting factor for isotropic tissue constituents"}),
-            parameter<double>("POPT", {.description = "tetanised optimal (maximal) active stress"}),
-            parameter<double>("LAMBDAMIN", {.description = "minimal active fiber stretch"}),
+                "OMEGA0", {.description = "weighting factor for isotropic tissue constituents",
+                              .validator = in_range<double>(0.0, 1.0)}),
+            parameter<double>("POPT",
+                {.description = "tetanised optimal (maximal) active stress",
+                    .validator = in_range<double>(0.0, std::numeric_limits<double>::max())}),
```
amgebauer: Why do you here not use `positive_or_zero<double>()`?

(also in a few other places)

amgebauer (approved):

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
lauraengelhardt: I forgot to replace those, there is no reason why not to use `positive_or_zero<double>()` :smiley:  
Thanks for the hint, I'll change.

Issue #1206, opened 2025-09-01
Add more tests for beaminteraction using Lagrange Multipliers


dharinib98:
## Description

With #637 merged, adding a few more tests would be necessary. Currently the implemented feature of Lagrange Multipliers for beam interaction have two tests, both using a Simo-Reissner beam with only positional coupling. It would be interesting to look at how an Euler-Bernoulli beam model would behave with Lagrange Multipliers. Additionally, an output test, testing whether the Lagrange multipliers are output correctly must be added.

## Possible Solution and Definition of Done

- [ ] Add some tests with Euler Bernoulli beams
- [ ] Add a test to check if LMs are output correctly. 
## Interested Parties

@isteinbrecher @mayrmt @maxfirmbach

mayrmt:
What about rotational coupling?

maxfirmbach:
We should add all three kinds of tests as soon as possible. This ensures that things are done correctly in the code.

maxfirmbach:
@dharinib98 and me encountered some challenges in the `beam3eb` case related to #1219.

dharinib98:
I would close this issue since we added all tests that test the current state of the implementation.

PR #1243, opened 2025-09-08
[beaminteraction] Add output test for Lagrange Multiplier

dharinib98:
## Description and Context

Following the implementation of Lagrange Multipliers for beaminteraction in #637  This PR adds an output test to check the Lagrange Multipliers for the added saddle point test case. 

#1206


maxfirmbach:
Great :+1: , I have just one minor comment. Maybe we should write the output in `ascii` to make it less cryptic. @dharinib98 as you are well trained in creating these kind of tests, this shouldn't be a big deal :wink:.

dharinib98:
@isteinbrecher @mayrmt We need to decide if we need an output test for the Lagrange Multiplier regularized variant as well.

mayrmt:
@dharinib98, does the regularized version output the same variable? If yes, then we do not necessarily need an additional test. If it is a different variable, then we better add another test.

dharinib98:
> @dharinib98, does the regularized version output the same variable? If yes, then we do not necessarily need an additional test. If it is a different variable, then we better add another test.

Yes. Lambda is the same for both. Only difference in the code between these two methods is the fact that we set penalty terms in the lambda model block. So I think one test should suffice.

PR #1244, opened 2025-09-09
Rename DataAccess::View to DataAccess::Share in SparseMatrix (and some smaller fixes)

amgebauer:
The name View confused me a lot. We have a lot of constructs like this:

```c++
  // --- Kzd - block - already scaled correctly by 0D model !-----------
  block_ptr = Core::LinAlg::matrix_transpose(*cardvasc0dman_->get_mat_dcardvasc0d_dd());
  global_state().assign_model_block(jac, *block_ptr, type(), MatBlockType::lm_displ, Core::LinAlg::DataAccess::View);
  block_ptr = nullptr;
```
The view-assign model block gets a reference of the SparseMatrix that will be destroyed directly afterwards. Seems wrong, but the "view"-assignment to the block matrix takes co-ownership of the underlying epetra objects. It is therefore clearer to call it `DataAccess::Share` instead of `DataAccess::View`

On the way, I fixed a few other things (second commit):

 * Some wrong consts in SparseMatrix
 * Removed unused destroy
 * Make defaulting to `View` (now Share) in `BaseDataGlobalState::assign_model_block` visible in the signature.


src/core/linalg/src/sparse/4C_linalg_sparsematrix.cpp
```diff
@@ -1631,47 +1589,47 @@ int Core::LinAlg::SparseMatrix::extract_global_row_view(
...
   return sysmat_->ReplaceGlobalValues(global_row, num_entries, values, indices);
 }
 
 /*----------------------------------------------------------------------*
  *----------------------------------------------------------------------*/
 int Core::LinAlg::SparseMatrix::insert_global_values(
-    int global_row, int num_entries, const double* values, const int* indices) const
+    int global_row, int num_entries, const double* values, const int* indices)
 {
   return sysmat_->InsertGlobalValues(global_row, num_entries, values, indices);
```
sebproell, line 1626: The constant `const` violations in our wrappers prompted me to create `OwnerOrView` (https://github.com/4C-multiphysics/4C/blob/main/src/core/utils/src/stl_extension/4C_utils_owner_or_view.hpp) which wraps underlying objects and propagates const-ness. It might be interesting to apply it here as well.

amgebauer: I also wanted to use this (or rather our Box). However, our "view" is not a view but takes co-ownership of the underlying epetra objects (we rely on that in many places). So this is not yet possible.

davidrudlstorfer (approved):

PR #1247, opened 2025-09-09
Use ElementRef in discretization loops

sebproell:
Roll out the new iterators to `my_(row|col)_element_range()`.

Part of #1228


src/core/fem/src/discretization/4C_fem_discretization_fillcomplete.cpp
```diff
@@ -234,7 +251,8 @@ void Core::FE::Discretization::build_element_to_node_pointers()
     const int nnode = ele->num_node();
     element_connectivity[ele->lid()].resize(nnode);
     const int* nodes = ele->node_ids();
-    for (int j = 0; j < nnode; ++j) element_connectivity[ele->lid()][j] = nodes[j];
+    for (int j = 0; j < nnode; ++j)
+      element_connectivity[ele->lid()][j] = noderowmap_->lid(nodes[j]);
```
sebproell: Indexing mistake I made in #1237. We should stop using `id()`  and always use something like `local_id` or `global_id`. I try to stick to this notation for all new features.

sebproell:
Now ready.

PR #1248, opened 2025-09-09
Don't pass input parameters of element to material

amgebauer:
The input parameters of the element were passed so that the material can read in their anisotropy information. Now, we pass the information about anisotropy directly to the material.

This is mainly a refactoring. The goal is that we can (in a future PR) read in the fiber directions also as an input field so that we can use inhomogeneous fiber directions with vtu/exodus based input.

Draft for now to check whether there are failing tests.


amgebauer:
Now ready for review

davidrudlstorfer (approved):
I am curious on why so many results changed with the refactoring?😅

amgebauer:
>I am curious on why so many results changed with the refactoring?😅

I was also curious. The cardiovascular-tests were super sensitive to the way we normalize the given fibers. Previously, we divided each entry by the norm, and now we multiply each entry with the inverse of the norm. Took me a while until I found that.

These cardiovascular tests have already proven to be not super stable to small changes :grimacing: I plan to do a rework of the cardiovascular framework and make these tests (hopefully) more stable.

PR #1249, opened 2025-09-09
[solver] Add diagonal pertubation to SPAI

maxfirmbach:
## Description and Context

A well-known disadvantage of incomplete factorizations is that the two factors can be unstable (e.g. L and U, or Q and R). This may occur, for instance, if matrix A is badly scaled or extremely ill-conditioned beyond machine precision, or if one of the pivotal elements occurs to be very small. In this case, a-priori diagonal perturbations may be effective as it shifts the eigenvalues just a tiny bit into the right direction to make an inversion feasible.

Note: This does of course not work for a singular matrix.


src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -370,6 +390,7 @@ std::shared_ptr<Core::LinAlg::SparseMatrix> Core::LinAlg::matrix_sparse_inverse(
     Teuchos::SerialQRDenseSolver<int, double> qrSolver;
     qrSolver.setMatrix(Teuchos::rcpFromRef(localA.base()));
     qrSolver.setVectors(Teuchos::rcpFromRef(localX), Teuchos::rcpFromRef(ek));
+    // qrSolver.factorWithEquilibration(true);
```
maxfirmbach: Not sure if we should turn this on or not.

mayrmt: I assume, that equilibration makes it more robust in case of small values. Did you run some experiments on that?

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.hpp
```diff
@@ -135,6 +135,26 @@ namespace Core::LinAlg
...
+   * A_perturbed(i,i) = alpha*sign(A(i,i))+rho*A(i,i)
+   *
+   */
+  struct OptionsSparseMatrixInverse
+  {
+    //! Absolute diagonal threshold
+    double alpha = 0.0;
+
+    //! Relative diagonal threshold
+    double rho = 1.0;
```
mayrmt: 1. Is "threshold" the right terminology? Isn't this rather like a "scaling factor"?
1. The documentation of the individual values is only useful in combination with the documentation of the `struct`. Just an observation. Might be fine.

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
```diff
@@ -182,6 +182,88 @@ namespace
           A_thresh->norm_frobenius(), expected_frobenius_norm, expected_frobenius_norm * 1e-10);
     }
   }
+
+  /** The test setup is based on a beam discretization with Euler-Bernoulli beam elements.
+   * The underlying discretized operator is singular as it stemms from a pure Neumann problem.
+   * A normal calculation of a sparse inverse on such kind of matrix is ill-defined and thus throws.
+   * By projecting the operator in a space explicitly not containing the rigid body modes / null
+   * space the problem can be shifted to be non-singular, but highly ill-conditioned. By introducing
+   * a a-priori diagonal perturbation, the Eigenvalues are shifted minimally to provide better
```
mayrmt:
```suggestion
   * an a-priori diagonal perturbation, the Eigenvalues are shifted minimally to provide better
```

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
```diff
@@ -182,6 +182,88 @@ namespace
...
+    }
+
+    // Try to invert pure Neumann problem, this should succeed as we use a projected operator and
+    // a-priori diagonal perturbation
+    {
+      Epetra_CrsMatrix* A;
+
+      int err = EpetraExt::MatrixMarketFileToCrsMatrix(
+          TESTING::get_support_file_path("test_matrices/beamII_projected.mm").c_str(),
+          Core::Communication::as_epetra_comm(comm_), A);
```
mayrmt, line 232: This reads a projected matrix. Wouldn't it make sense to read an unprojected matrix an perform the projection within this test?

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
maxfirmbach: No, for the cases I did there was no major difference.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.hpp
maxfirmbach: Right, they are scaling factors, will rename it.

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
maxfirmbach: Doing the projection inside the test is rather tricky, at least to this point. The projection is build upon a condition given in the input file ... so currently I don't see a straight forward way in doing this.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
mayrmt: Hmmm... I don't know. Pick one option. We can change it later.

maxfirmbach: I turned it on for know.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -370,6 +390,7 @@ std::shared_ptr<Core::LinAlg::SparseMatrix> Core::LinAlg::matrix_sparse_inverse(
     Teuchos::SerialQRDenseSolver<int, double> qrSolver;
     qrSolver.setMatrix(Teuchos::rcpFromRef(localA.base()));
     qrSolver.setVectors(Teuchos::rcpFromRef(localX), Teuchos::rcpFromRef(ek));
+    qrSolver.factorWithEquilibration(true);
```
isteinbrecher, line 393: How is this related to the other changes here?

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
```diff
@@ -182,6 +182,88 @@ namespace
...
+      if (err != 0) FOUR_C_THROW("Matrix read failed.");
+      std::shared_ptr<Epetra_CrsMatrix> A_crs = Core::Utils::shared_ptr_from_ref(*A);
+      Core::LinAlg::SparseMatrix A_sparse(A_crs, Core::LinAlg::DataAccess::Copy);
+
+      const double tol = 1e-14;
```
isteinbrecher, line 237: Why is the top here different from the previous case where we expect a fail?

isteinbrecher (commented):
Looks good to me, just some minor comments/questions

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
maxfirmbach: Makes the factorization a bit more stable. We should've activated it already before. Other than that not related.

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
maxfirmbach: Good point. I can make them identical.

PR #1252, opened 2025-09-10
Refactor set_valid_parameters

maxiludwig:
This PR refactors the occurences of `set_valid_parameters` to return `InputSpecs` or `std::vector<InputSpec>`. 
Previously these functions worked with a map of a string-valued section name and InputSpec, meaning that we defined the section name twice. Once, in the map, and once, as the name of the `group`.

### DIFF
--- a/src/ale/4C_ale_input.cpp
+++ b/src/ale/4C_ale_input.cpp
@@ -14,7 +14,7 @@ FOUR_C_NAMESPACE_OPEN
 
 
 
-Core::IO::InputSpec ALE::set_valid_parameters()
+Core::IO::InputSpec ALE::valid_parameters()
 {
   using namespace Core::IO::InputSpecBuilders;
   Core::IO::InputSpec spec = group("ALE DYNAMIC",
--- a/src/ale/4C_ale_input.hpp
+++ b/src/ale/4C_ale_input.hpp
@@ -60,7 +60,7 @@ namespace ALE
   };
 
   /// Defines all valid parameters for ale problem
-  Core::IO::InputSpec set_valid_parameters();
+  Core::IO::InputSpec valid_parameters();
 
   /// Defines ale specific conditions
   void set_valid_conditions(std::vector<Core::Conditions::ConditionDefinition>& condlist);
--- a/src/art_net/4C_art_net_input.cpp
+++ b/src/art_net/4C_art_net_input.cpp
@@ -14,7 +14,7 @@
 FOUR_C_NAMESPACE_OPEN
 
 
-Core::IO::InputSpec ArtDyn::set_valid_parameters()
+Core::IO::InputSpec ArtDyn::valid_parameters()
 {
   using namespace Core::IO::InputSpecBuilders;
   Core::IO::InputSpec spec = group("ARTERIAL DYNAMIC",
@@ -67,7 +67,7 @@ Core::IO::InputSpec ArtDyn::set_valid_parameters()
 
 
 
-Core::IO::InputSpec ArteryNetwork::set_valid_parameters()
+Core::IO::InputSpec ArteryNetwork::valid_parameters()
 {
   using namespace Core::IO::InputSpecBuilders;
   Core::IO::InputSpec spec = group("COUPLED REDUCED-D AIRWAYS AND TISSUE DYNAMIC",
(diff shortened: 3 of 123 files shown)


maxiludwig:
With this change, `set_valid_parameters` is actually not setting anything. I thought of changing this to `valid_parameters`

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -322,11 +322,24 @@ namespace Core::IO
...
 
 
-  InputFile::InputFile(std::map<std::string, InputSpec> valid_sections,
+  InputFile::InputFile(const std::vector<InputSpec>& valid_sections,
       std::vector<std::string> legacy_section_names, MPI_Comm comm)
       : pimpl_(std::make_unique<Internal::InputFileImpl>(comm))
   {
-    pimpl_->valid_sections_ = std::move(valid_sections);
+    std::map<std::string, InputSpec> section_map;
+    for (const auto& spec : valid_sections)
```
sebproell:
```suggestion
    for (auto&& spec : valid_sections)
```

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -322,11 +322,24 @@ namespace Core::IO
...
+    {
+      const auto name = spec.name();
+      FOUR_C_ASSERT_ALWAYS(!name.empty(),
+          "You are trying to add an unnamed InputSpec. "
+          "You can only add a named spec, not an all_of or one_of spec.");
+      FOUR_C_ASSERT_ALWAYS(section_map.find(name) == section_map.end(),
+          "You are trying to add multiple InputSpecs with the same name '{}'. "
+          "Each InputSpec must have a unique name.",
+          name);
+      section_map[name] = spec;
```
sebproell:
```suggestion
      section_map.emplace(name, std::move(spec));
```

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -322,11 +322,24 @@ namespace Core::IO
...
       std::vector<std::string> legacy_section_names, MPI_Comm comm)
       : pimpl_(std::make_unique<Internal::InputFileImpl>(comm))
   {
-    pimpl_->valid_sections_ = std::move(valid_sections);
+    std::map<std::string, InputSpec> section_map;
+    for (const auto& spec : valid_sections)
+    {
+      const auto name = spec.name();
+      FOUR_C_ASSERT_ALWAYS(!name.empty(),
+          "You are trying to add an unnamed InputSpec. "
```
sebproell:
```suggestion
          "You are trying to add an unnamed InputSpec for a top-level section. "
```

src/core/io/src/4C_io_input_file.cpp
```diff
@@ -322,11 +322,24 @@ namespace Core::IO
...
+          "You can only add a named spec, not an all_of or one_of spec.");
+      FOUR_C_ASSERT_ALWAYS(section_map.find(name) == section_map.end(),
+          "You are trying to add multiple InputSpecs with the same name '{}'. "
```
sebproell:
```suggestion
          "You are trying to add multiple InputSpecs with the same name '{}' on the top level. "
```

src/core/io/src/4C_io_input_spec.hpp
```diff
@@ -94,6 +95,11 @@ namespace Core::IO
      */
     [[nodiscard]] const Internal::InputSpecImpl& impl() const;
 
+    /**
+     * Get the name of this InputSpec. The name is the identifier used in the input file.
+     */
```
sebproell:
```suggestion
    /**
     * Get the name of this InputSpec. The name is used as a key in the input file.
     * The name may be an empty string in case the spec does not have a name.
     */
```

src/core/io/src/4C_io_input_spec.hpp
```diff
@@ -94,6 +95,11 @@ namespace Core::IO
...
+    [[nodiscard]] std::string name() const;
```
sebproell:
```suggestion
    [[nodiscard]] const std::string& name() const;
```

Actually, `std::string_view` would be even better, but 4C is generally quite std::string reliant, so let's stick with this.

sebproell (commented):
Thank you for cleaning this up! I would rename the function `valid_parameters` as you suggested.

Issue #1256, opened 2025-09-11
[constraint framework] Unify input parameters, straighten class hierarchy and outsource functions


maxfirmbach:
## Description

I had a look at the `constraint_framework` module and it seems a bit unintuitive to me. I would like to add a submodel-evaluator there myself to enforce certain constraints, but it's difficult to implement an easy solution in the current state.

There are a lot of duplicated member variables that do the same thing, the sub-model evaluator base class is cluttered with implementation specifics of RVEs, the actual RVE evaluator implements functionality that should live in free functions outside of the whole evaluator pattern and the embedded mesh evaluator seems like a weird attachment to all of this.

Not only is it difficult to add new things to the `constraint_framework` (due to the points mentioned above), it also actively encourages to find mediocre ways to use these parts of `4C`.

In addition there are other small things that should be cleaned up:
Why are there two parameter lists pointing to `EMBEDDED MESH COUPLING` in `4C_global_data.hpp`?
Why is are the input parameters for the `constraint_framework` distributed over several different files and not inside sublists like `CONSTRAINT/EMBEDDED MESH COUPLING` or `CONSTRAINT/MULTI POINT CONSTRAINTS`?

Resolving all these points would highly improve the code and the lower the bar for adding new evaluators. I'm open for helping with this issue, as I really like the concept of the `constraint_framework` and want to use it myself.

## Possible Solution and Definition of Done

Fix the points mentioned above:
- [x] Unify input into sublists and merge into one file (like it's done for `beaminteraction`) (see #1258 #1262)
- [ ] Unify member variables in the respective classes that do the same thing (see #1258 #1262)
- [ ] Move functionality irrelevant to the sub-model evaluators into dedicated files and use them as free functions
- [ ] Straighten the class hierarchy of the `constraint_framework`, a base class should contain no implementation specifics, move them to the respective sub-model evaluators

## Interested Parties

@mayrmt @isteinbrecher @eulovi @m-frey

m-frey:
I'll look into the rve and base class related things.

maxfirmbach:
> I'll look into the rve and base class related things.

Thank you!

maxfirmbach:
I will have a look at input related things, at least unifying the input parameters.

m-frey:
[...]

PR #1258, opened 2025-09-11
[constraint framework] Unify input and move into respective module

maxfirmbach:
## Description and Context

This PR unifies the `constraint_framework` by doing the following things:
1. Unify files and parameterlists related to the constraint framework
2. Unify enum and parameters of the input
3. Move input to the constraint framework

Part of #1256


src/constraint_framework/4C_constraint_framework_embeddedmesh_params.hpp
```diff
@@ -23,20 +23,18 @@ namespace Constraints::EmbeddedMesh
...
-            Inpar::Constraints::EmbeddedMeshConstraintEnforcement::none;
+    EnforcementStrategy embedded_mesh_constraint_enforcement_ = EnforcementStrategy::none;
 
     //! Penalty parameter for coupling enforcement
     double embedded_mesh_constraint_penalty_parameter_ = 0.0;
 
     //! Shape function for the mortar Lagrange-multiplicators
-    enum Inpar::Constraints::SolidToSolidMortarShapefunctions embedded_mesh_mortar_shape_function_ =
-        Inpar::Constraints::SolidToSolidMortarShapefunctions::none;
+    enum Constraints::EmbeddedMesh::SolidToSolidMortarShapefunctions
```
sebproell: Using the `enum` tag is unnecessary in C++.


```suggestion
    Constraints::EmbeddedMesh::SolidToSolidMortarShapefunctions
```

src/constraint_framework/4C_constraint_framework_input.cpp
```diff
@@ -5,44 +5,79 @@
...
 #include "4C_fem_condition_definition.hpp"
 #include "4C_io_input_spec_builders.hpp"
+
 FOUR_C_NAMESPACE_OPEN
-// set the mpc specific parameters
-void Inpar::RveMpc::set_valid_parameters(std::map<std::string, Core::IO::InputSpec>& list)
+
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+void Constraints::set_valid_parameters(std::map<std::string, Core::IO::InputSpec>& list)
```
sebproell: This will need to be rebased since #1252 was merged.

src/constraint_framework/4C_constraint_framework_input.cpp
```diff
@@ -5,44 +5,79 @@
...
-                  .default_value = automatic}),
+                  .default_value = MultiPoint::RveReferenceDeformationDefinition::automatic}),
 
-          deprecated_selection<Inpar::RveMpc::EnforcementStrategy>("ENFORCEMENT",
+          deprecated_selection<EnforcementStrategy>("ENFORCEMENT",
               {
-                  {"penalty_method", penalty},
-                  {"lagrange_multiplier_method", lagrangeMultiplier},
+                  {"penalty_method", EnforcementStrategy::penalty},
+                  {"lagrange_multiplier_method", EnforcementStrategy::lagrange},
```
sebproell: If you name the enum constants like the string, you can use the nicer `parameter` with the enum instead of `deprecated_selection`.

src/constraint_framework/4C_constraint_framework_input.hpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+namespace Constraints
+{
+  /// type of the submodel for constraintmodels
+  enum class SubModelType
+  {
+    submodel_undefined,    ///< default
+    submodel_pbc_rve,      ///< apply periodic displacement bcs on
+    submodel_embeddedmesh  ///< apply embedded mesh bcs
```
sebproell:
```suggestion
    undefined,    ///< default
    pbc_rve,      ///< apply periodic displacement bcs on
    embeddedmesh  ///< apply embedded mesh bcs
```

The scoped enum already serves as a prefix.

src/constraint_framework/4C_constraint_framework_input.hpp
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  /// type of employed constraint enforcement strategy
+  enum class EnforcementStrategy
+  {
+    //! Default value.
+    none,
```
sebproell: I think saying "default value" here is weird. Do you mean in the input? What would `none` do then? That is what should be documented here. Note that with `InputSpec` you do not need a default for parameters if you don't want one.

src/constraint_framework/4C_constraint_framework_input.cpp
```diff
@@ -5,44 +5,79 @@
...
+
+  /*----------------------------------------------------------------------*/
+  /* parameters for embedded mesh constraint submodel */
+
+  list["EMBEDDED MESH COUPLING"] = group("EMBEDDED MESH COUPLING",
+      {
+
+          parameter<EmbeddedMesh::CouplingStrategy>("COUPLING_STRATEGY",
+              {.description = "Strategy to couple background and overlapping mesh",
+                  .default_value = EmbeddedMesh::CouplingStrategy::none}),
```
sebproell: Does it make sense to have a `none` default? The same holds for a few other parameters.

maxfirmbach: No, I don't think so ... actually all of the `none` can most likely be removed.

src/constraint_framework/4C_constraint_framework_input.cpp
maxfirmbach: Did a rebase to get the new changes :+1:.

src/constraint_framework/4C_constraint_framework_embeddedmesh_params.hpp
```diff
@@ -23,20 +23,18 @@ namespace Constraints::EmbeddedMesh
...
     //! Strategy for coupling the embedded meshes
-    enum Inpar::Constraints::EmbeddedMeshCouplingStrategy embedded_mesh_coupling_strategy_ =
-        Inpar::Constraints::EmbeddedMeshCouplingStrategy::none;
+    CouplingStrategy embedded_mesh_coupling_strategy_ = CouplingStrategy::none;
 
     //! Constraint enforcement method
-    enum Inpar::Constraints::EmbeddedMeshConstraintEnforcement
-        embedded_mesh_constraint_enforcement_ =
-            Inpar::Constraints::EmbeddedMeshConstraintEnforcement::none;
+    EnforcementStrategy embedded_mesh_constraint_enforcement_ = EnforcementStrategy::none;
```
mayrmt: One could even drop the `embedded_mesh_` prefix from the members... they are part of a `struct` for embedded mesh inside a file for embedded mesh.

src/constraint_framework/4C_constraint_framework_embeddedmesh_solid_to_solid_utils.cpp
```diff
@@ -339,22 +339,22 @@ void Constraints::EmbeddedMesh::change_gauss_rule_of_cut_elements(
...
+    case Constraints::EmbeddedMesh::SolidToSolidMortarShapefunctions::quad4:
     {
       n_lambda_node = 1 * 3;
       return;
     }
-    case Inpar::Constraints::SolidToSolidMortarShapefunctions::nurbs9:
+    case Constraints::EmbeddedMesh::SolidToSolidMortarShapefunctions::nurbs9:
     {
       n_lambda_node = 1 * 3;
       return;
```
mayrmt, line 355: Both cases perform the same action, so they could be combined. And both perform a multiplication with `1`, which could be dropped. Maybe beyond the scope of this PR, but @eulovi might want to look at this at some point.

src/constraint_framework/4C_constraint_framework_input.hpp
maxfirmbach: Same here ... `none` can be removed and we just don't set a default value that actually does nothing.

src/constraint_framework/4C_constraint_framework_input.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+
+          parameter<EmbeddedMesh::SolidToSolidMortarShapefunctions>("MORTAR_SHAPE_FUNCTION",
+              {.description = "Shape functions that should be use in case of coupling using the "
+                              "Mortar/Lagrange  Multiplier method ",
+                  .default_value = EmbeddedMesh::SolidToSolidMortarShapefunctions::none}),
```
mayrmt: No part of this PR, but "mortar" and "Lagrange multiplier method" are two distinct topics, that are often done in combination, but do not mean the same thing. Hence, the documentation here is not very accurate. 

/cc @eulovi

maxfirmbach:
Regarding the new input spec features I would need to have a more detailed look at them. I'm not very familiar with the recent changes.

davidrudlstorfer:
> Regarding the new input spec features I would need to have a more detailed look at them. I'm not very familiar with the recent changes.

@maxfirmbach maybe it helps you - I used (as far as I know) all the new input features in this PR to update the input for the beam potential https://github.com/4C-multiphysics/4C/pull/1028/files

And also most (if not all) new featuers are used in the new lung input, see https://github.com/4C-multiphysics/4C/pull/1159

src/constraint_framework/4C_constraint_framework_embeddedmesh_params.hpp
maxfirmbach: I updated the member variable names.

src/constraint_framework/4C_constraint_framework_embeddedmesh_solid_to_solid_utils.cpp
maxfirmbach: Let's do this in another PR.

src/constraint_framework/4C_constraint_framework_input.cpp
maxfirmbach: Let's do this in another PR.

maxfirmbach:
I would try to get this PR merged. In a following PR I will further work on the input and see how to incorporate the new features.

PR #1260, opened 2025-09-12
Move Proper Orthogonal Decomposition (POD) to Core

amgebauer:
I want to enable MOR for the new structural time integration. Therefore, I now moved the general utilities to core. In a second step, one could generalize it for any problem type.


amgebauer:
That depends on how we want to enable POD for all problem types. I thought of implementing it as a solver for a linear system. So you get a large sparse matrix, then pod will be applied and then your sub-solver solves a much smaller linear system. In this case, it should directly be available for all problem types.

mayrmt:
Sounds like algebraic multigrid with the POD basis as ~nullspace vectors~ transfer operators...

src/core/linear_solver/src/mor/4C_linear_solver_mor_pod.cpp
```diff
@@ -0,0 +1,344 @@
...
+    file1.seekg(0, std::ifstream::beg);
+
+    // read the first 8 byte into the memory-block
+    file1.read(sizeblock, std::streamoff(8));
+
+    // close the file
+    file1.close();
+
+    // union for conversion of 4 bytes to int or float
+    union CharIntFloat
```
sebproell: This kind of type punning via a union is UB, although neither clang nor gcc dares to exploit it. Please use `memcpy` instead.

src/core/linear_solver/src/mor/4C_linear_solver_mor_pod.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    // open binary file
+    std::ifstream file1(absolute_path_to_pod_file.c_str(),
+        std::ifstream::in | std::ifstream::binary | std::ifstream::ate);
+
+    FOUR_C_ASSERT_ALWAYS(
+        file1.good(), "File containing the matrix could not be opened. Check Input-File.");
+
+    // allocation of a memory-block to read the data
+    char* sizeblock = new char[8];
```
sebproell:
```suggestion
    std::array<char,8> sizeblock;
```

and remove the corresponding `delete` below.

maxfirmbach:
Regarding the second point of generalizing this feature. We have a `projector` object that can be given to the linear solver. Wouldn't it make sense to use that? Yes currently it's a very specific thing, but in the end this does exactly the same.

src/core/linear_solver/src/mor/4C_linear_solver_mor_pod.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    // calculate a number of bytes that are needed to be reserved in order to fill the multivector
+    int mysize = projection_matrix.NumVectors() * projection_matrix.MyLength() * formatfactor;
+
+    // open binary file (again)
+    std::ifstream file2(absolute_path_to_pod_file.c_str(),
+        std::ifstream::in | std::ifstream::binary | std::ifstream::ate);
+
+    // allocation of a memory-block to read the data
+    char* memblock = new char[mysize];
```
sebproell: Depending on how often you allocate this, either use a unique_ptr (if only used once) or keep scratch memory around and pass it to this function.

maxfirmbach:
In addition, this is more like a `linalg` thing e.g. see krylov projector.

src/core/linear_solver/src/mor/4C_linear_solver_mor_pod.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace Core::LinAlg
+{
+  class SparseMatrix;
+}  // namespace Core::LinAlg
+namespace Core::LinearSolver::MOR
+{
+  class ProperOrthogonalDecomposition
```
sebproell: Some documentation would be good here.

src/core/linear_solver/src/mor/4C_linear_solver_mor_pod.cpp
```diff
@@ -0,0 +1,344 @@
...
+        "Projection matrix does not match discretization.");
+
+    // check orthogonality
+    FOUR_C_ASSERT_ALWAYS(
+        is_pod_basis_orthogonal(projection_matrix), "Projection matrix is not orthogonal.");
+
+    return projection_matrix;
+  }
+
+  void multiply_multi_vectors(Core::LinAlg::MultiVector<double>& multivect1, char multivect1Trans,
```
maxfirmbach: This would nicely fit into `linalg`, e.g. `4C_linalg_utils_sparse_algebra_math.{hpp,cpp}`.

src/core/linear_solver/src/mor/4C_linear_solver_mor_pod.cpp
```diff
@@ -0,0 +1,344 @@
...
+    int err =
+        multivect_temp.Multiply(multivect1Trans, multivect2Trans, 1.0, multivect1, multivect2, 0.0);
+    FOUR_C_ASSERT_ALWAYS(!err, "Multiplication failed.");
+
+    // import the result to a Core::LinAlg::MultiVector<double> whose elements/rows are distributed
+    // over all procs
+    result.Import(multivect_temp, impo, Insert, nullptr);
+  }
+
+  Core::LinAlg::SparseMatrix multi_vector_to_linalg_sparse_matrix(
```
maxfirmbach: This would nicely fit into linalg, e.g. `4C_linalg_utils_sparse_algebra_manipulation.{hpp,cpp}`.

src/core/linear_solver/src/mor/4C_linear_solver_mor_pod.cpp
amgebauer: I now use the standard matrix market format and can reuse some functionality from Trilinos and don't need the custom parser.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_manipulation.hpp
```diff
@@ -423,6 +423,21 @@ namespace Core::LinAlg
...
+   * represented in sparse matrix form. The rangemap defines the mapping of rows,
+   * while the domainmap (if provided) defines the mapping of columns.
+   *
+   * @param multivect The multivector containing the data to be converted into a sparse matrix.
+   * @param rangemap The map specifying the row mapping for the sparse matrix.
+   * @param domainmap Optional map specifying the column mapping for the sparse matrix.
+   * @return Core::LinAlg::SparseMatrix The constructed sparse matrix.
+   */
+  Core::LinAlg::SparseMatrix make_sparse_matrix(const Core::LinAlg::MultiVector<double>& multivect,
+      const Core::LinAlg::Map& rangemap, const std::optional<Core::LinAlg::Map>& domainmap);
```
mayrmt, line 440: From the code, I understand that a MultiVector with 3 vectors and 100 entries per vector will result in a rectangular 100x3 matrix. The matrix is dense, but stored in a sparse format. If that is true, then I find the comments about the domain map a bit misleading.

If the domain map just decribes, how a vector `x` in a mat-vec `Ax` needs to be layed out, such that the mat-vac can be computed, then the domain map just follows standard Petra terminology.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.hpp
```diff
@@ -155,6 +155,41 @@ namespace Core::LinAlg
...
+   * - After multiplication, the result is redistributed according to `importer` using the
+   *   Insert combine mode.
+   * - This approach ensures that multiplication can be performed without worrying about
+   *   parallel distribution of input MultiVectors, at the cost of temporary memory usage.
+   *
+   * @throws Throws an assertion if the internal Multiply operation fails.
+   */
+  void multiply_multi_vectors(const Core::LinAlg::MultiVector<double>& v1, bool trans_v1,
+      const Core::LinAlg::MultiVector<double>& v2, bool trans_v2, const Core::LinAlg::Map& red_map,
+      const Core::LinAlg::Import& importer, Core::LinAlg::MultiVector<double>& result);
```
mayrmt, line 191: `red_map` -> `redundant_map` to reflect its meaning in the naming?

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_manipulation.hpp
maxfirmbach: If the multivector contains zeros, one could potentially filter them out.

mayrmt: True. I was thinking about the column map, which is "dense". In the end, it's a sparse matrix format that is likely to have a large fill-in. Not a big deal, though.

amgebauer:
Obsolete with #1627

PR #1196, opened 2025-08-31
CMake: allow to specifiy required dependenices for tests

sebproell:
#1072 added the option to mark some of our tests as skipped. This has the benefit that the test number is ideally a constant and not dependent on your configuration. You will see that there would be tests, that are not executable due to your configuration. This PR adds a convenience feature to enable this check for the `four_c_test` function.

PR #1261, opened 2025-09-12
Fix invalid ref in TU when having multiple models

maxiludwig:
With the new lung input introduced in #1159 we enabled an easier and more adaptable model generation. This should also allow different types of terminal unit and airway models in different elements. When trying this out I realized that this leads to invalid references to the internal vectors of the `TerminalUnitModel. 

This PR resolve this issue by deferring the creation of some evaluators (negative_residual_evaluator`, `jacobian_evaluator`, internal_state_updater)  until all models are registered. This routine now runs after the DOF assignment.


tests/list_of_tests.cmake
```diff
@@ -1944,14 +1946,14 @@ four_c_test(TEST_FILE embeddedmesh_parallel_bending_beam.4C.yaml NP 3 REQUIRED_D
...
-    four_c_test_vtk(embeddedmesh_parallel_bending_beam-vtk-structure.4C.yaml embeddedmesh_parallel_bending_beam.4C.yaml 3 xxx-structure.pvd ref/embeddedmesh_parallel_bending_beam_np_3-structure.pvd 1e-06)
-    four_c_test_vtk(embeddedmesh_parallel_bending_beam_fine-vtk-structure.4C.yaml embeddedmesh_parallel_bending_beam_fine.4C.yaml 3 xxx-structure.pvd ref/embeddedmesh_parallel_bending_beam_fine-structure.pvd 1e-06)
-endif ()
+if(FOUR_C_WITH_ARBORX)
+        # only execute these vtk tests with ArborX
+        four_c_test_vtk(beam3eb_static_beam_to_solid_volume_meshtying_beam_to_beam_contact_boundingvolume-vtk-beam-interaction-geometric-search.4C.yaml beam3eb_static_beam_to_solid_volume_meshtying_beam_to_beam_contact_boundingvolume.4C.yaml 2 xxx-beam-interaction-geometric-search.pvd ref/beam3eb_static_beam_to_solid_volume_meshtying_beam_to_beam_contact_boundingvolume-vtk/beam-interaction-geometric-search.pvd 1e-06 0.0 1.0)
```
bgoderbauer: Do you have different pre-commit settings?

davidrudlstorfer: Could it be the case that the branch is outdated?

With https://github.com/4C-multiphysics/4C/pull/1196/files these sections are no longer in the latest main? Or am I missing something here

davidrudlstorfer: Whops sorry, the vtk tests do not yet support this feature, as @bgoderbauer said the formatting is just off. Sorry!

PR #1262, opened 2025-09-12
[constraint framework] Refactor input parameters and lists

maxfirmbach:
## Description and Context

This PR reduces the constraint framework to one parameterlist entry point given by `CONSTRAINT`, which defines general parameters like `ENFORCEMENT_STRATEGY` and e.g. the penalty parameter if penalty is used. Each submodel evaluator has it's one sublist.

This makes it easier to add new parameters, if a new submodel evaluator is added.

The changes are highly invasive as it changes the way things are done in the input sections!

Part of #1256


src/constraint_framework/4C_constraint_framework_input.cpp
```diff
@@ -20,47 +20,39 @@ std::vector<Core::IO::InputSpec> Constraints::valid_parameters()
 
   std::vector<Core::IO::InputSpec> spec;
 
+  /*----------------------------------------------------------------------*/
+  /* parameters for constraint model */
+
+  spec.push_back(group("CONSTRAINT",
+      {parameter<EnforcementStrategy>("CONSTRAINT_ENFORCEMENT",
+           {.description =
+                   "Apply a constraint enforcement in the embedded mesh coupling strategy"}),
```
mayrmt: This description is probably outdated now. It should be more general, right?

maxfirmbach: Yes, I will update this.

maxfirmbach:
I don't understand why the clang pipeline fails, nothing really changed in regards to includes etc.

amgebauer:
> I don't understand why the clang pipeline fails, nothing really changed in regards to includes etc.

It happens from time to time, but we have not yet understood why. I deleted the cache and restarted the job. https://github.com/4C-multiphysics/4C/pull/1186#issuecomment-3241402705

maxfirmbach:
@amgebauer Thank's a lot! I use `clang` locally too, so this was confusing ... thanks for the clarification.

tests/input_files/embeddedmesh_2d_hertzian_contact.yaml
```diff
@@ -162,11 +162,12 @@ CUT GENERAL:
   DO_SELFCUT: false
   BOUNDARYCELL_CUBATURDEGREE: 1
   INTEGRATE_INSIDE_CELLS: false
-EMBEDDED MESH COUPLING:
+CONSTRAINT:
+  CONSTRAINT_ENFORCEMENT: penalty
+  PENALTY_PARAM: 10000
+CONSTRAINT/EMBEDDED MESH COUPLING:
```
sebproell, line 168: Since you break backward compatibilty anyway: why not make this a nested section inside `CONSTRAINT`?

maxfirmbach: Currently this follows the logic of how things are done in `beaminteraction`. I like the idea of doing a nested section, but due to consistency I would rather opt for an intermediate step here. I think it would make sense to find a general solution for all structure submodel evaluators at some point.

dharinib98 (approved):

m-frey (approved):
Thanks again for these changes. As discussed offline it is a good idea to update the input paramters.

PR #1263, opened 2025-09-13
Make SparseMatrix movable and move-assignable

amgebauer:
We currently fall back to a deep copy, which is quite expensive.

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_sparsematrix.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_sparsematrix.hpp
@@ -151,6 +151,9 @@ namespace Core::LinAlg
     /// Assignment operator. Makes a deep copy.
     SparseMatrix& operator=(const SparseMatrix& mat);
 
+    SparseMatrix(SparseMatrix&&) noexcept = default;
+    SparseMatrix& operator=(SparseMatrix&&) noexcept = default;
+
     /// Assignment method. Deep copy or view on matrix.
     /*!
       Explicit method for the assignment operator. You can make an explicit
--- a/src/core/linalg/src/sparse/4C_linalg_view.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_view.hpp
@@ -132,8 +132,8 @@ namespace Core::LinAlg
     View(const View& other) = default;
     View& operator=(const View& other) = default;
 
-    View(View&& other) = default;
-    View& operator=(View&& other) = default;
+    View(View&& other) noexcept = default;
+    View& operator=(View&& other) noexcept = default;
 
     ~View() = default;
 


amgebauer:
@sebproell Why do we rely on this? With a true move-operation, this will never work.

```c++
Core::LinAlg::Vector<double> a(*map_, true);
a.put_scalar(1.0);

Core::LinAlg::MultiVector<double>& mv = a;
mv.PutScalar(2.0);
EXPECT_EQ(means_multi_vector(a)[0], 2.0);

// Reassigning to a must keep mv valid: move assign
a = Core::LinAlg::Vector<double>(*map_, true);
EXPECT_EQ(means_multi_vector(mv)[0], 0.0);
```

sebproell:
> @sebproell Why do we rely on this? With a true move-operation, this will never work.
> 
> ```c++
> Core::LinAlg::Vector<double> a(*map_, true);
> a.put_scalar(1.0);
> 
> Core::LinAlg::MultiVector<double>& mv = a;
> mv.PutScalar(2.0);
> EXPECT_EQ(means_multi_vector(a)[0], 2.0);
> 
> // Reassigning to a must keep mv valid: move assign
> a = Core::LinAlg::Vector<double>(*map_, true);
> EXPECT_EQ(means_multi_vector(mv)[0], 0.0);
> ```

Historical reason. In Epetra, Vector derives from MultiVector. Thus, it is perfectly valid there to take a pointer mv of type MV pointing to a V. If you assign to the underlying vector, the mv pointer should still be usable and we relied on that for historical reasons, so I disallowed moving. Our wrappers are not derived from each other and instead work with views if converted to other types. Maybe we no longer need the behavior that is tested here, since we improved the robustness of views.

amgebauer:
@sebproell The solution now is to only treat the `multi_vector_view_` differently (move the contents there). In all other cases, we move the view instead of the content.

What might be inconsistent is

```c++
Core::LinAlg::Vector<double> a(*map_, true);
a.put_scalar(1.0);

const Core::LinAlg::Map& map = a.get_map();

// Reassigning to a must keep mv valid: move assign
a = Core::LinAlg::Vector<double>(*map_, true);

// map is now invalid
```
But that was also the case before.

sebproell (dismissed):

amgebauer:
Vector and MultiVector cannot be savely made move-assignable. It all boils down that the underlying Epetra structures don't support moves.

At least for the SparseMatrix, these operations are possible.

amgebauer:
I realize that we can also not allow moving for the SparseMatrix since it can also be both, a view (or rather sharing) and owning (and assigning to a shared matrix would split the share).

I think the cleanest solution would be to implement a thin View similar to the std::span and define all operations based on a cheap View.

PR #1267, opened 2025-09-15
Generalize projection for solving linear systems

amgebauer:
This PR generalizes projectors. Currently, only a Krylov-Projector was possible. Now, I generalized the interface so that one can implement more projectors.

I want to use this for a POD-projector to enable MOR in the new structural time integration scheme, see #1260.


amgebauer:
@maxfirmbach 
> In a lot of places apply_pt is changed to apply, but shouldn't it be apply_transpose?

I tried to make it consistent:

`projector.apply(...)` -> Projects from the normal to space to the projected space
`projector.apply_transpose(...)` -> Projects from the projected space to the normal space.

```c++
A_p = projector.apply(A);
b_p = projector.apply(b); // previously, this was projector.apply_pt(b);

/// solve linear system ...

x = projector.apply_transpose(x_p); // previously, this was projector.apply_p(b);
```

I don't really understand why the naming was switched before.

mayrmt:
The question is, which direction we call a "projection". Full to reduced basis or reduced basis to full basis?

amgebauer:
> The question is, which direction we call a "projection". Full to reduced basis or reduced basis to full basis?

Both are projections of course. However, I think it is clear in this context. I pass a projector to the linear solver, so this projector should project from the un-projected bases (full) to the projected basis (might be reduced) before solving the system, and then inverse-project the solution vector.

mayrmt:
Makes sense. 

Would it help to use function names such as `project_full_to_reduced_space()`? This might be more self-descriptive and avoids math notations such as "apply" and "apply_transpose".

src/core/linear_solver/src/method/4C_linear_solver_method_projector.hpp
```diff
@@ -0,0 +1,71 @@
...
+    [[nodiscard]] virtual Core::LinAlg::Vector<double> apply(
+        const LinAlg::Vector<double>& b) const = 0;
+
+    /*!
+     * @brief Apply the transpose of the projector to a vector, i.e., transforming the vector from
+     * the projected space back to the original space.
+     *
+     * @param b The input vector.
+     */
+    [[nodiscard]] virtual Core::LinAlg::Vector<double> apply_transpose(
```
mayrmt: I still find this naming not optimal, since it implies knowledge about the formulas of projections. This could be improved by expressing the direction of the projection within the function name.

The same holds for `apply()` declared above.

mayrmt (approved):

PR #1268, opened 2025-09-15
Fix Gauss point data output manager MPI implementation

rjoussen:
## Description and Context

This avoids using the same send and receive addresses for MPI_Allreduce calls in the gauss point data output manager, which is not in line with MPI standards and doesn't work for intel-mpi. 

none

### DIFF
--- a/src/structure_new/src/output/4C_structure_new_gauss_point_data_output_manager.cpp
+++ b/src/structure_new/src/output/4C_structure_new_gauss_point_data_output_manager.cpp
@@ -180,7 +180,8 @@ void Solid::ModelEvaluator::GaussPointDataOutputManager::distribute_quantities(M
   const Core::Communication::Exporter exporter(comm);
 
   int max_quantities = quantities_.size();
-  Core::Communication::max_all(&max_quantities, &max_quantities, 1, comm);
+  int max_quantities_local = quantities_.size();
+  Core::Communication::max_all(&max_quantities_local, &max_quantities, 1, comm);
 
   if (max_quantities == 0)
   {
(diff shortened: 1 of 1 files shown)


amgebauer (approved):

sebproell (changes requested):
@rjoussen First, thank you for contributing to 4C!

I can see that your fix works but would prefer a different one. To fully fix this issue, you would now need to do this kind of change in every place we call our `max_all`. Instead, let us fix the `max_all` function itself. I would propose a signature like `T max_all(const T& local, int root, MPI_Comm comm)`. The return value is created inside the functions and will thus always be different from the source value.

rjoussen:
@sebproell I understand what you mean. As far as I can tell, the Gauss point data manager is the only place where this is incorrectly used, which is why I changed it there. I can apply the proposed change globally, but then I would suggest doing that not only for `Core::Communication::max_all`, but also for `Core::Communication::min_all` and `Core::Communication::sum_all`, which wrap `MPI_Allreduce` as well. What do you think?

sebproell:
> What do you think?

@rjoussen That would be even better :) I did not mention that to not overwhelm you with requests for changes. You may also merge this PR and do this cleanup in the next one. If you cannot do it in the next days (no problem, if not!), it would be good to open an issue documenting this problem.

rjoussen:
Okay, in this case I will merge the PR request for now and start changing all the occurrences of `sum_all`, `min_all` and `max_all`.

davidrudlstorfer (approved):

sebproell (approved):

PR #1270, opened 2025-09-16
Extract rebalancing from mesh input

sebproell:
Rebalancing is independent of mesh input and useful in its own right.

### DIFF
--- a/src/core/rebalance/src/4C_rebalance.hpp
+++ b/src/core/rebalance/src/4C_rebalance.hpp
@@ -10,7 +10,20 @@
 
 #include "4C_config.hpp"
 
+#include <mpi.h>
+#include <Teuchos_ParameterList.hpp>
+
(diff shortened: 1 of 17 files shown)


src/core/rebalance/src/4C_rebalance.hpp
```diff
@@ -10,8 +10,21 @@
...
 
+namespace Core::FE
+{
+  class Discretization;
+}
+
+namespace Core::LinAlg
+{
+  class Map;
+}
```
davidrudlstorfer, line 26: Do we need these forward declarations here?

sebproell: Yes, we refer to these types in l. 72

amgebauer (approved):

maxfirmbach:
Looks good! Thank you for moving it to `rebalance`.

PR #1273, opened 2025-09-16
Allow Discretization to fill from Mesh

sebproell:
Next step to remove incremental allocation of elements and nodes. The new function still does the same thing, but we are now in a better place to do something different internally.

Part of #1228


src/core/fem/src/discretization/4C_fem_discretization.cpp
```diff
@@ -68,6 +71,69 @@ void Core::FE::Discretization::add_node(std::shared_ptr<Core::Nodes::Node> node)
...
+      const std::string cell_type_string = Core::FE::cell_type_to_string(eb.cell_type);
+
+      FOUR_C_ASSERT_ALWAYS(cell_type == eb.cell_type,
+          "Element block '{}' has cell type '{}' but your given element definition for '{}' has "
+          "cell type '{}'.",
+          eb_id, eb.cell_type, element_name, cell_type);
+
+      for (const auto& cell : eb.cells())
+      {
+        // Do not yet use the external cell ID. 4C is not yet prepared to deal with this!
```
davidrudlstorfer, line 104: Should we create a dedicated issue for this (and further down below)?

davidrudlstorfer (approved):

src/core/fem/src/discretization/4C_fem_discretization.cpp
sebproell: I don't think that's necessary. @amgebauer and I are working on this.

Issue #1274, opened 2025-09-17
Avoid using the same send and receive addresses for MPI_Allreduce calls


rjoussen:
## Description

The current `MPI_Allreduce` wrapper functions allow inputting identical sender and receiver addresses, which can cause problems (cf. #1268). As proposed by @sebproell, the signature of `MPI_Allreduce` wrappers should look like this:

`T MPI_Allreduce_wrapper(const T& local, MPI_comm communicator)`

The (potential) size/length of `T` is no longer needed as it can be deduced from the reference to local.

## Possible Solution and Definition of Done

The three MPI_Allreduce wrappers must be refactored:
- [x] `Core::Communication::max_all` (cf. #1275 )
- [x] `Core::Communication::min_all` (cf #1275 )
- [x] `Core::Communication::sum_all` (cf. #1324 )

`sum_all` is trickier as it is called with pointers to a variety of different data types (could be handled by overloading `sum_all` often), or the pointers used so far are (in my opinion) cryptic and not defined in the vicinity of the `sum_all` call, e.g. in:
`Core::Communication::sum_all(mymoments.data(), ((*liftdragvals)[label]).data() + 3, 3, trueresidual.get_comm());`

PR #1275, opened 2025-09-17
Cleanup MPI_Allreduce wrappers

rjoussen:
## Description and Context

Refactor `Core::Communication::max_all` and `Core::Communication::min_all` as described in #1274.
Marked as draft since it probably makes sense to merge this with the refactoring of `Core::Communication::sum_all` combined.

#1274


sebproell (approved):
LGTM. One could make the functions more generic by taking a `contiguous_range`, although that would likely require different return types with dynamic extent. Let's go with what you have for now. Regarding the draft state: you can add `sum_all` here or do it in another PR.

amgebauer (approved):

PR #1279, opened 2025-09-17
Pass node coordinates explictly to Discretization

sebproell:
An intermediate step towards making `Node*` obsolete in most cases.

#1228


src/core/fem/src/general/node/4C_fem_general_node.hpp
```diff
@@ -136,7 +136,7 @@ namespace Core::Nodes
     /*!
     \brief Return coordinates vector
     */
-    inline const std::vector<double>& x() const { return x_; }
+    inline std::span<const double, 3> x() const { return std::span<const double, 3>(x_.data(), 3); }
```
isteinbrecher, line 139: Does this have to be a hardcoded 3? What about 2D simulations or higher order, e.g., space-time fomulations.

sebproell: Currently, it is de facto a hard-coded dimension 3 throughout 4C (as you can see from the fact that the tests still pass). 2D is treated as 3D with the last coordinate zero or ignored. Of course, our Discretization should really be able to have different dimensions. @amgebauer WDYT: should we make the span here be dynamically sized?

amgebauer: We could also add a template parameter `dim` to the Discretization, which currently only supports (and defaults to 3). If we at some point want to support true 2D cases, or space-time nodes, we of course need to adapt all the dependent places. But we anyway need to do it since we inherently rely on that.

src/core/fem/src/discretization/4C_fem_discretization_fillcomplete.cpp
```diff
@@ -181,7 +180,7 @@ static void make_map_and_local_pointers(MPI_Comm comm,
 
 void Core::FE::Discretization::build_node_row_map()
 {
-  make_map_and_local_pointers(get_comm(), node_, noderowptr_, noderowmap_, /*local_only=*/true);
+  make_map_and_local_pointers(get_comm(), node_, nodecolptr_, noderowmap_, /*local_only=*/true);
```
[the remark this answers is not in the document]
sebproell: This comment is technically correct but irrelevant since I just discard the result. I will continue working on this and would leave it like this.

src/beamcontact/4C_beamcontact_beam3contact_manager.cpp
```diff
@@ -468,14 +468,15 @@ void CONTACT::Beam3cmanager::init_beam_contact_discret()
     Core::Nodes::Node* node = problem_discret().l_col_node(i);
     if (!node) FOUR_C_THROW("Cannot find node with local id {}", i);
     std::shared_ptr<Core::Nodes::Node> newnode = std::shared_ptr<Core::Nodes::Node>(node->clone());
+
     if (BeamInteraction::Utils::is_beam_node(*newnode))
     {
-      bt_sol_discret().add_node(newnode);
+      bt_sol_discret().add_node(node->x(), node->id(), newnode);
```
c-p-schmidt: Why not 
```suggestion
      bt_sol_discret().add_node(newnode->x(), newnode->id(), newnode);
```
as in the other calls?

src/beamcontact/4C_beamcontact_beam3contact_manager.cpp
```diff
@@ -468,14 +468,15 @@ void CONTACT::Beam3cmanager::init_beam_contact_discret()
...
       nodedofs[node->id()] = problem_discret().dof(0, node);
     }
     else if (BeamInteraction::Utils::is_rigid_sphere_node(*newnode))
     {
-      bt_sol_discret().add_node(newnode);
+      bt_sol_discret().add_node(node->x(), node->id(), newnode);
```
c-p-schmidt: Also here.

src/beamcontact/4C_beamcontact_beam3contact_manager.cpp
sebproell: Doesn't really matter but I changed it. I need to revisit this in subsequent PRs anyway, once the coordinates fully move out of the `Node` class.

src/core/fem/src/general/node/4C_fem_general_node.hpp
sebproell: I now assert that we indeed always pass three coordinates. There was only one place in the code where this was wrong in Mortar (where a temporary node was created with 2 coordinates since the third one was not used).

isteinbrecher: I know that we don't use it at the moment, but currently it is very easy to have variable dimensions of the nodes. I am afraid that with this change we will loose flexibility. 

For context, we are currently working on space time formulations where we need 4 positions for each node.

sebproell: I see. I am sure we can find a solution for that. I am not too familiar with space-time (never implement it) but I assume you also start with a 3D mesh but have additional dofs for the temporal component of the solution field? IMO, a node is only a spatial position in a mesh. Anything additional is either done by dofs + shape functions or custom stuff that you can implement on a custom type. Nurbs for example want to add a weight to to the nodes/control points and can still do this.

If I am completely wrong here, we can also make the `x()` call more flexible again, i.e. return a dynamically-sized span.

PR #1286, opened 2025-09-19
Allow data input from mesh file

amgebauer:
This allows data input from the input-file. The data is stored inside the IMR. Only possible for vtu mesh input.

Supported types are:

- scalars: `int`, `double`
- vectors: `Tensor<int, dim>`, `Tensor<double, dim>`
- symmetric Tensors: `SymmetricTensor<int, dim, dim>`, `SymmetricTensor<double, dim, dim>`
- Tensors: `Tensor<int, dim, dim>`, `Tensor<double, dim, dim>`

In a follow-up, I will enable reading `InputField`s directly from the vtu-file. This will avoid having to maintain the mapping between the geometry and the corresponding (material) parameters in separate files. And you can directly visualize it in ParaView.

fyi @gilrrei


src/core/linalg/src/dense/4C_linalg_symmetric_tensor.hpp
```diff
@@ -193,9 +193,15 @@ namespace Core::LinAlg
...
-                          TensorCompressionType<std::remove_cvref_t<T>>>::value;
+  constexpr bool is_symmetric_tensor = []() consteval
+  {
+    if constexpr (is_tensor<T>)
+    {
+      return Internal::IsSymmetricCompressionTypeHelper<
+          TensorCompressionType<std::remove_cvref_t<T>>>::value;
+    }
+    return false;
+  }();
```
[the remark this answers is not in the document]
amgebauer: This is required here since `TensorCompressionType<std::remove_cvref_t<T>>` cannot be instantiated for non-tensor types.

amgebauer:
Now ready for review.

src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -73,29 +77,239 @@ namespace
...
+  {
+    const int n_components = array.GetNumberOfComponents();
+
+    auto extract_entry = [&](auto& typed_array) -> T
+    {
+      switch (n_components)
+      {
+        case 1:
+        {
+          // scalar
```
davidrudlstorfer, line 207: Maybe add these types to the other cases as well?

src/core/io/tests/test_files/vtu/hex8_point_data.vtu
davidrudlstorfer, line 1: Would it make sense to provide a smaller test file (<< 800 lines)?

davidrudlstorfer (dismissed):
Nice!

src/core/io/tests/test_files/vtu/hex8_point_data.vtu
amgebauer: It is already more or less minimal size with 3 hex elements and all implemented types. However,  meshio does not format the arrays properly by default, i.e., each component of a tensor is written into its own line. Now, I formatted so that the components of vectors and tensors are grouped in each line.

davidrudlstorfer (approved):
Nice!

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -65,9 +96,14 @@ namespace Core::IO::MeshInput
      */
     std::vector<std::array<double, dim>> points{};
 
+    /**
+     * Point data associated with each point in the mesh.
+     */
+    std::unordered_map<std::string, FieldDataVariantType<dim>> point_data{};
```
sebproell, line 106: I think this needs more docs.

```
The key refers to the name of the data set. The value is a variant containing a vector of one of the eligible data types.
```

and possibly more.

src/core/io/tests/4C_io_vtu_test.cpp
```diff
@@ -60,6 +66,228 @@ namespace
     }
   }
 
+  TEST(VTU, MeshCubeHex8PointDataInput)
+  {
+#ifndef FOUR_C_WITH_VTK
+    GTEST_SKIP() << "Skipping test: 4C vtu-input requires VTK support";
+#endif
```
sebproell, line 73: Let's create a macro like `FOUR_C_TEST_REQUIRES(...)`. Not in this PR though.

davidrudlstorfer (approved):

PR #1287, opened 2025-09-19
Refactor metadata and schemas

gilrrei:
This PR:
- Refactors the metadata python objects to handle all the current input types and their validation
- Introduces a JSON schema draft 2020-12 module, such that we can introduce more schema functionality (it was a bit crazy before)
- All the files are gathered in a new `four_c_metadata` folder in the utilities

Users don't have to change anything.

This work is needed for the:
- New documentation build for the input data (@ischeider, soon)
- Tests for the metadata objects (soon)
- Diagnostics of the input (number of parameters, parameters with the wrong type etc, soonish) 

#1285 found this bug


utilities/four_c_metadata/requirements.txt
```diff
@@ -0,0 +1,15 @@
...
+numpy
+vtk
+
+# for sphinx documentation:
+matplotlib
+myst-parser
+nbsphinx
+scipy
+sphinx-multibuild
+sphinx-rtd-theme
```
sebproell, line 10: Isn't this in the wrong place? It is not related to metadata but rather the test/build. Or maybe the package should be called `four_c_build_test` rather than `four_c_metadata`?

davidrudlstorfer:
Just a basic question:

Would it make sense to somehow unify the Python capabilites in 4C in one folder? Currently we have:

- `utilities/four_c_utils`
- the python files in `utilities`
- the various python files in `tests/`
- and I am sure there are more floating around

@c-p-schmidt and me looked through various of them today

@gilrrei this is absolutely not a task for this PR - but would it make sense to move this stuff into `utilities/four_c_utils`?

utilities/four_c_metadata/requirements.txt
gilrrei: Yes, sure. Renaming it sounds good to me

gilrrei:
@davidrudlstorfer 

> Would it make sense to somehow unify the Python capabilities in 4C in one folder? 

I talked about this with @sebproell, and we came to the conclusion that there are two different types of requirements/Python scripts:
- related to **development/workflows**: various checks, precommit stuff etc (acts on the **source**)
- related to **build/testing**: from schema build, documentation, testing (acts on the **build**)

Hence, I think it makes sense to split it up into two distinct packages with different requirements.

davidrudlstorfer:
@gilrrei I just briefly discussed with @sebproell and what do you think of the following approach:

- move all python capability into one python project, i.e., `four_c_python`
- this project could still contain different modules, i.e., `src/metadata` & `src/testing_building` & `src/develop`
- split the requirements into separate files, i.e., `requirements_metadata.txt` & `requirements_testing_building.txt` & `requirements_develop`
- install correct requirements into the develop environment (located in the src folder) install the build/test/metadata requirements into the build/test/metadata environment in the build folder

this way we would have one single "python repo" in 4C but still be able to install different "versions" of the package

Similar to the approach you did in Queens with 4C and the FourCIPP install.

I am still a bit hesitant to provide multiple python "repos" in 4C when we can still combine all of them into a single one

gilrrei:
@davidrudlstorfer 

> I am still a bit hesitant to provide multiple python "repos" in 4C when we can still combine all of them into a single one

From a Python POV, there is no big difference. I don't mind either way since it's not really a Python project that people use as a library, but rather a simple collection of tools.

gilrrei:
Additional changes in this PR (thanks @sebproell and @davidrudlstorfer)

- unified python project for 4C
- different requirements for build/testing and development
- reduced number of python scripts (@davidrudlstorfer is ready to remove the missing ones)
- removed all mismatching python versions 

cleanup for more will come :)

cmake/configure/configure_Python.cmake
```diff
@@ -43,21 +42,20 @@ function(_execute_process)
...
-
 message(STATUS "Setting up virtual Python environment for building and testing 4C")
 find_package(Python REQUIRED COMPONENTS Interpreter Development)
+
 if(Python_FOUND)
   message(STATUS "Using python executable: ${Python_EXECUTABLE} (V${Python_VERSION})")
 endif()
 
-if(Python_VERSION VERSION_LESS "3.10")
+if(Python_VERSION VERSION_LESS "3.12")
```
ppraegla, line 55: @gilrrei Why to we need a newer Python version? Does Python 3.12 provide some essential features that 3.10 does not have?

Python 3.10 is the default on Ubuntu 22.04 which makes it easy to just install via the apt package.

gilrrei: Python 3.12 is the standard of Ubuntu 24.04, which was already used in the build environment before.  That decision was made some time ago.  Btw 3.10 EOL is in a year

ppraegla: We also want to support the oldest supported Ubuntu LTS version (22.04 in this case). Though this does probably not hold for Python if 3.10 will be EOL. A newer Python version is easier to install than a new compiler.

Thanks.

PR #1288, opened 2025-09-21
Refactor input arguments of nodal nullspace

maxfirmbach:
## Description and Context

The nodal nullspace matrix dimension is usually hardcoded, therefore the number of degrees of freedom and nullspace dimension input arguments are not relevant.


src/beam3/4C_beam3_reissner.cpp
```diff
@@ -163,9 +140,10 @@ Core::LinAlg::SerialDenseMatrix Discret::Elements::Beam3rType::compute_null_spac
   rotTangTwo.cross_product(e2, tangent);
   rotTangThree.cross_product(e3, tangent);
 
-  Core::LinAlg::SerialDenseMatrix nullspace(numdof, dimnsp);
   if (beam3r->is_centerline_node(node))
   {
+    Core::LinAlg::SerialDenseMatrix nullspace(9, 6);
```
mayrmt, line 145: If dimensions are hardcoded now, does it make sense to switch to our fixed-size matrix with compile-time dimensions?

maxfirmbach: That is a good point. Yes, most likely it's worth thinking about that.

sebproell: You would need to return the value differently in this case _or_ remove the inheritance that mandates identical return types.

maxfirmbach: Yes, let's leave this for another PR.

maxfirmbach:
There's still something wrong. I thought it would be easier ... I will open a seperate PR for some initial modifications.

maxfirmbach:
In a first step I will try to get rid of the prior known nullspace dimension and get this merged. Afterwards I'll think about the number of degrees of freedom. There are also some inconsistencies e.g. xwall is given 4 degrees of freedom per node, but when I print it out, it has 8 ...

mayrmt:
> There are also some inconsistencies e.g. xwall is given 4 degrees of freedom per node, but when I print it out, it has 8 ...

The x in xwall stands for "extended". It enriches the function space close to the wall, essentially adding degrees of freedom.

maxfirmbach:
Most things I cared about are merged, thus closing for now.

PR #1289, opened 2025-09-21
[LinAlg] Move sparse linear algebra methods out of projector classes

maxfirmbach:
## Description and Context

As the title states. This PR aims to move all sparse linear algebra related methods out of projector classes to the respective files in `LinAlg/sparse`.

This also touches some places already modified in #1260.

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
+++ b/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
@@ -430,7 +430,7 @@ Core::LinAlg::SparseMatrix Core::LinAlg::multiply_multi_vector_multi_vector(
     const int id, const bool fill)
 {
   // compute information about density of P^T A P
-  const Core::LinAlg::MultiVector<double>& temp = [&]()
+  const Core::LinAlg::MultiVector<double>& get_multi_vector = [&]()
   {
     if (id == 1)
       return mv1;
@@ -440,8 +440,8 @@ Core::LinAlg::SparseMatrix Core::LinAlg::multiply_multi_vector_multi_vector(
       FOUR_C_THROW("id must be 1 or 2");
   }();
 
-  Core::LinAlg::Vector<double> prod(temp(0));
-  for (int i = 1; i < mv1.NumVectors(); ++i) prod.multiply(1.0, temp(i), prod, 1.0);
+  Core::LinAlg::Vector<double> prod(get_multi_vector(0));
+  for (int i = 1; i < mv1.NumVectors(); ++i) prod.multiply(1.0, get_multi_vector(i), prod, 1.0);
   int numnonzero = 0;
   for (int i = 0; i < prod.local_length(); ++i)
     if (prod[i] != 0.0) numnonzero++;
@@ -476,31 +476,25 @@ Core::LinAlg::SparseMatrix Core::LinAlg::multiply_multi_vector_multi_vector(
   //--------------------------------------------------------
   // compute mat by multiplying upright mv1 with lying mv2^T:
   //--------------------------------------------------------
-  // loop over all proc-rows
   for (int rr = 0; rr < nummyrows; ++rr)
   {
-    // get global row id of current local row id
     const int grid = mat.global_row_index(rr);
 
-    // vector of all row values - prevented from growing in following loops
     std::vector<double> rowvals;
     rowvals.reserve(numvals);
 
-    // vector of indices corresponding to vector of row values
     std::vector<int> indices;
     indices.reserve(numvals);
 
-    // loop over all entries of global w
     for (int mm = 0; mm < numvals; ++mm)
     {
       double sum = 0.0;
-      // loop over all kernel/weight vector
+
       for (int vv = 0; vv < mv1.NumVectors(); ++vv)
       {
         sum += mv1(vv)[rr] * mv2glob(vv)[mm];
       }
 
-      // add value to vector only if non-zero
       if (sum != 0.0)
       {
         rowvals.push_back(sum);
@@ -508,7 +502,6 @@ Core::LinAlg::SparseMatrix Core::LinAlg::multiply_multi_vector_multi_vector(
       }
     }
 
-    // insert values in mat
     int err = mat.insert_global_values(grid, indices.size(), rowvals.data(), indices.data());
     if (err < 0)
     {
@@ -527,12 +520,12 @@ Core::LinAlg::SparseMatrix Core::LinAlg::multiply_multi_vector_multi_vector(
  *----------------------------------------------------------------------*/
 void Core::LinAlg::multiply_multi_vectors(Core::LinAlg::MultiVector<double>& multivect1,
     char multivect1Trans, Core::LinAlg::MultiVector<double>& multivect2, char multivect2Trans,
-    Core::LinAlg::Map& redmap, Core::LinAlg::Import& impo,
+    Core::LinAlg::Map& redundant_map, Core::LinAlg::Import& impo,
     Core::LinAlg::MultiVector<double>& result)
 {
-  // initialize temporary Core::LinAlg::MultiVector<double> (redmap: all procs hold all
+  // initialize temporary Core::LinAlg::MultiVector<double> (redundant_map: all procs hold all
   // elements/rows)
-  Core::LinAlg::MultiVector<double> multivect_temp(redmap, multivect2.NumVectors(), true);
+  Core::LinAlg::MultiVector<double> multivect_temp(redundant_map, multivect2.NumVectors(), true);
 
   // do the multiplication: (all procs hold the full result)
   int err =


src/core/linalg/src/sparse/4C_linalg_krylov_projector.hpp
```diff
@@ -97,19 +97,6 @@ namespace Core::LinAlg
...
-        const Core::LinAlg::MultiVector<double>& mv,
-        const Core::LinAlg::SerialDenseMatrix& dm) const;
-
-    //! outer product of two MultiVectors
-    Core::LinAlg::SparseMatrix multiply_multi_vector_multi_vector(
-        const Core::LinAlg::MultiVector<double>& mv1,  //! first MultiVector
-        const Core::LinAlg::MultiVector<double>& mv2,  //! second MultiVector
-        const int id = 1,  //! id of MultiVector form which sparsity of output matrix is estimated
-        const bool fill = true  //! bool for completing matrix after computation
-    ) const;
```
mayrmt, line 111: I know, that these two methods are linear algebra operations, so they could be moved. However, I do not see general applicability. They only exist due to the choice of data structures in POD.

I do not object to moving them, but I want to make sure, that they do not spread across the code base.

mayrmt (approved):

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_manipulation.cpp
```diff
@@ -793,6 +793,30 @@ void Core::LinAlg::multi_vector_to_std_vector(const Core::LinAlg::MultiVector<do
...
+    std::shared_ptr<Core::LinAlg::Map> domainmap, Core::LinAlg::SparseMatrix& sparsemat)
+{
+  double* Values;
+  Values = multivect.Values();
+
+  for (int i = 0; i < multivect.NumVectors(); i++)
+  {
+    for (int j = 0; j < multivect.MyLength(); j++)
+    {
+      sparsemat.assemble(Values[i * multivect.MyLength() + j], rangemap.gid(j), i);
```
amgebauer: You could add a check whether the value is zero.

amgebauer (approved):

src/core/linalg/src/sparse/4C_linalg_krylov_projector.hpp
maxfirmbach: I would argue that it's better to have these methods in a central place. They are not specific to the Krylov projector or POD at all. Actually those methods exist in a similar version in `Belos` too.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_manipulation.cpp
maxfirmbach: I now check for a zero value and continue if true.

mayrmt (dismissed):

maxfirmbach:
I refactored the `multi_vector_to_sparse_matrix` function a little bit and added it to the nullspace constraint submodel evaluator. Reduces some code and bundles functionality.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -402,4 +403,145 @@ std::shared_ptr<Core::LinAlg::SparseMatrix> Core::LinAlg::matrix_sparse_inverse(
...
+}
+
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+Core::LinAlg::SparseMatrix Core::LinAlg::multiply_multi_vector_multi_vector(
+    const Core::LinAlg::MultiVector<double>& mv1, const Core::LinAlg::MultiVector<double>& mv2,
+    const int id, const bool fill)
+{
+  // compute information about density of P^T A P
+  const Core::LinAlg::MultiVector<double>& temp = [&]()
```
mayrmt: Can we find another name? `temp` is very generic and it took me a while to figure out, where this is used and what it does.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -402,4 +403,145 @@ std::shared_ptr<Core::LinAlg::SparseMatrix> Core::LinAlg::matrix_sparse_inverse(
...
+  for (int i = 0; i < prod.local_length(); ++i)
+    if (prod[i] != 0.0) numnonzero++;
+
+  int glob_numnonzero = 0;
+  Core::Communication::sum_all(&numnonzero, &glob_numnonzero, 1, prod.get_comm());
+
+  Core::LinAlg::Map mv1map(mv1.get_map().num_global_elements(), mv1.get_map().num_my_elements(),
+      mv1.get_map().my_global_elements(), 0, mv1.get_map().get_comm());
+  // initialization of mat with map of mv1
+  Core::LinAlg::SparseMatrix mat(mv1map, glob_numnonzero, false);
```
mayrmt, line 455: What about a check that the maps of both vectors match?

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -402,4 +403,145 @@ std::shared_ptr<Core::LinAlg::SparseMatrix> Core::LinAlg::matrix_sparse_inverse(
...
+  if (fill) mat.complete();
+
+  return mat;
+}
+
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+void Core::LinAlg::multiply_multi_vectors(Core::LinAlg::MultiVector<double>& multivect1,
+    char multivect1Trans, Core::LinAlg::MultiVector<double>& multivect2, char multivect2Trans,
+    Core::LinAlg::Map& redmap, Core::LinAlg::Import& impo,
```
mayrmt: `redmap` can sound like both `reduced_map` or `redundant_map`. Explicitly spelling it out might improve readability.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -402,4 +403,145 @@ std::shared_ptr<Core::LinAlg::SparseMatrix> Core::LinAlg::matrix_sparse_inverse(
...
+
+    // vector of indices corresponding to vector of row values
+    std::vector<int> indices;
+    indices.reserve(numvals);
+
+    // loop over all entries of global w
+    for (int mm = 0; mm < numvals; ++mm)
+    {
+      double sum = 0.0;
+      // loop over all kernel/weight vector
```
mayrmt: This comment sounds physics-specific.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -402,4 +403,145 @@ std::shared_ptr<Core::LinAlg::SparseMatrix> Core::LinAlg::matrix_sparse_inverse(
...
+      for (int vv = 0; vv < mv1.NumVectors(); ++vv)
+      {
+        sum += mv1(vv)[rr] * mv2glob(vv)[mm];
+      }
+
+      // add value to vector only if non-zero
+      if (sum != 0.0)
```
mayrmt: Compare to an exact `0.0` or numeric limits?

maxfirmbach:
I just want to point out that this PR is not about redoing / refactoring all these methods. I mainly moved them to a different place, such that it makes sense for other parts of the code to facilitate them.

mayrmt (dismissed):

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
maxfirmbach: Right, I will delete it.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
maxfirmbach: I think we could compare to the numeric limit.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
maxfirmbach: Do the maps need to match? Or just the size of the vectors?

mayrmt: I don't know. 

Maps must match in the usual multiplication scenario in Trilinos. This one is not the usual case, so I don't know.

mayrmt: If we don't know, then let's leave it for now. Time will show if further checks are necessary.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
maxfirmbach: Yes, I think `get_multi_vector` or similar would be more fitting.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
maxfirmbach: I renamed it to `redundant_map`.

maxfirmbach:
@mayrmt I don't know for sure if a redundant map is necessary. I would say no, but I didn't look into the function workings in detail.

mayrmt (dismissed):

PR #1290, opened 2025-09-22
Add node IDs as yaml data in DBC monitor output

isteinbrecher:
## Description and Context

Following #1211 , this PR adds the node ids as a yaml list instead of a string.


tests/input_files/ref/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_restart-101_monitor_dbc.yaml
```diff
@@ -1,6 +1,5 @@
 dbc monitor condition:
-  - condition information: "Condition 100 Tagged elements: ONOFF : 1 1 0 0 0 0 0 0 0 \nNodes of this condition:
```
isteinbrecher, line 2: As far as I know, no-one is using this string. So I decided to remove it and replace it with only the node ids. If more information is needed in the future, this can be easily done without (or minimally) modifying the existing data structure.

Issue #198, opened 2025-01-08
[NOX] Replace NOX::Epetra::Group with Tpetra counterpart


eulovi:
> This issue was originally created by @mayrmt in https://gitlab.lrz.de/baci/baci/-/issues/949 on 2022-12-23.

## Description

Due to the upcoming deprecation of `Epetra`, we need to replace our use of the `NOX::Epetra::Group` with a `Tpetra`-based construct. There is no direct counterpart (inparticular no `NOX::Tpetra::Group`), so we need to figure out how to do that. 

If we want to keep `NOX`, the way to go is to use the `ModelEvaluator` concept of `Thyra` properly. This stands in extreme contrast to the usage of `NOX` in the new structure time integration (here `NOX` for nonlinear solves, time integration and model evaluation seem to be very tightly connected), thus mentioning [890](https://gitlab.lrz.de/baci/baci/-/issues/890).

## Possible Solution and Definition of Done
To be even able to start working on `Tpetra` and `NOX` we should centralize the calls to `NOX`. The implementation of `NOX` in the old structure time integration and in the fluid-structure interaction seem to be quite similar. Of course this does not count for the new-structure time integration ...
- [ ] Unify calls to `NOX` in a central place and make them more abstract and also usable for other physics modules
- [ ] Identify the mechanism to hook-up `Tpetra` and `NOX`
- [ ] Replace `NOX::Epetra::Group` with its `Tpetra` counterpart in all `NOX` instances within Baci (list incomplete so far):
   - [ ] Solid 
   - [ ] Monolithic FSI
   - [ ] Partitioned FSI 

## Related Issues and Merge Requests

* Part of [870](https://gitlab.lrz.de/baci/baci/-/issues/870) 

## Interested Parties

@baci/team_structure @baci/team_fsi @maxfirmbach

maxfirmbach:
Highlighting https://github.com/trilinos/Trilinos/pull/14084 and thus moving this issue to blocked.

sebproell:
While we might be able to inject the internals from our linear algebra wrappers into a Tpetra group (once they are Tpetra internally), this will not solve the problem that we should not code against a specific `Group` implementation. What I mean by this is that we should not just replace the 100+ occurrences `Epetra::Group/Vector` with `Tpetra::Group/Vector`  and still have this type showing up all over the place. We may do this to make progress on the Tpetra conversion side; I just want to mention that this is not a useable NOX interface und our control.

[...]

PR #1292, opened 2025-09-23
Implement `GroupBase` to replace `NOX::Epetra::Group`

vovannikov:
## Description and Context
This PR implements our own group class `GroupBase` that replaces `NOX::Epetra::Group` from Trilinos. Effectively, it reproduces functionality of the latter with the following essential modifications:
- Remove method `getNormLastLinearSolveResidual()` - it is mainly used in the single place of NOX, inside `NOX::Direction::ModifiedNewton::rescueBadNewtonSolve()` which attempts to proceed to the next time step even if the convergence tests failed but the "accuracy" measure was not so bad. We have never used this option, and in general, it does not look robust to me.
- Remove everything related to preconditioners since we build them outside of the group.
- Remove method computeNormNewtonSolveResidual() as it mainly provides information for `getNormLastLinearSolveResidual()`.
- Remove data fields related to normNewtonSolveResidual for the above reason.
- Remove `SharedObject` wrapper for the linear system and required interface objects. In general, this pattern is very confusing, especially if one tries to guess why it is even needed provided we use shared pointers. To my understanding, this wrapper was used in the original `NOX::Epetra::Group` in order to keep the object status with the flags `isValidSolverJacOp` and `isValidJacobian` such that they would mean meaningful things. It looks that for our use cases this concern is not a problem. However, if later we need something similar, a different strategy should be implemented. For instance, the status flags should be stored together with the linear system object inside some other wrapper.
- Remove flag `isValidSolverJacOp` which gets obsolete after removing the aforementioned shared object pattern.

Later this class will be further modified to remove `NOX::Epetra::Vector` and switch to `Core::LinAlg::Vector`.

#359, #198

### DIFF
--- a/src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_sd.cpp
+++ b/src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_sd.cpp
@@ -11,11 +11,11 @@
 #include "4C_global_data.hpp"
 #include "4C_io_control.hpp"
 #include "4C_linalg_vector.hpp"
+#include "4C_solver_nonlin_nox_group.hpp"
 
 #include <NOX_Abstract_Group.H>
 #include <NOX_Abstract_Vector.H>
 #include <NOX_Common.H>
-#include <NOX_Epetra_Group.H>
 #include <NOX_Epetra_Interface_Required.H>
 #include <NOX_Epetra_Vector.H>
 #include <NOX_GlobalData.H>
@@ -54,7 +54,7 @@ bool NOX::FSI::SDRelaxation::compute(::NOX::Abstract::Group& newgrp, double& ste
   }
 
   const ::NOX::Abstract::Group& oldgrp = s.getPreviousSolutionGroup();
-  ::NOX::Epetra::Group& egrp = dynamic_cast<::NOX::Epetra::Group&>(newgrp);
+  auto& grp = dynamic_cast<NOX::Nln::GroupBase&>(newgrp);
 
   // Perform single-step linesearch
 
@@ -63,7 +63,7 @@ bool NOX::FSI::SDRelaxation::compute(::NOX::Abstract::Group& newgrp, double& ste
 
   double numerator = oldgrp.getF().innerProduct(dir);
   double denominator =
-      compute_directional_derivative(dir, *egrp.getRequiredInterface()).innerProduct(dir);
+      compute_directional_derivative(dir, *grp.get_required_interface()).innerProduct(dir);
 
   step = -numerator / denominator;
   utils_->out() << "          RELAX = " << std::setw(5) << step << "\n";
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
@@ -59,24 +59,13 @@ Teuchos::RCP<::NOX::Abstract::Group> NOX::Nln::Group::clone(::NOX::CopyType type
  *----------------------------------------------------------------------------*/
 ::NOX::Abstract::Group& NOX::Nln::Group::operator=(const ::NOX::Abstract::Group& source)
 {
-  return operator=(dynamic_cast<const ::NOX::Epetra::Group&>(source));
-}
-
-/*----------------------------------------------------------------------------*
- *----------------------------------------------------------------------------*/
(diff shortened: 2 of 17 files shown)


src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_sd.cpp
```diff
@@ -54,7 +54,7 @@ bool NOX::FSI::SDRelaxation::compute(::NOX::Abstract::Group& newgrp, double& ste
   }
 
   const ::NOX::Abstract::Group& oldgrp = s.getPreviousSolutionGroup();
-  ::NOX::Epetra::Group& egrp = dynamic_cast<::NOX::Epetra::Group&>(newgrp);
+  auto& egrp = dynamic_cast<NOX::Nln::GroupBase&>(newgrp);
```
mayrmt:
```suggestion
  auto& grp = dynamic_cast<NOX::Nln::GroupBase&>(newgrp);
```

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
```diff
@@ -59,25 +59,14 @@ Teuchos::RCP<::NOX::Abstract::Group> NOX::Nln::Group::clone(::NOX::CopyType type
...
 {
-  return operator=(dynamic_cast<const ::NOX::Epetra::Group&>(source));
-}
-
-/*----------------------------------------------------------------------------*
- *----------------------------------------------------------------------------*/
-::NOX::Abstract::Group& NOX::Nln::Group::operator=(const ::NOX::Epetra::Group& source)
-{
-  ::NOX::Epetra::Group::operator=(source);
+  NOX::Nln::GroupBase::operator=(source);
```
mayrmt, line 62: Not part of this PR, but I still find it confusing to have our own `NOX::` namespace. It only means that we have to be extras careful to distinguish between `NOX::` and `::NOX::`.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group_base.cpp
```diff
@@ -7,18 +7,286 @@
...
+  lastLinearSolveConverged = source.lastLinearSolveConverged;
+  lastNumIterations = source.lastNumIterations;
+  lastAchievedTol = source.lastAchievedTol;
+
+  return *this;
+}
+
+Teuchos::RCP<const ::NOX::Abstract::Vector> NOX::Nln::GroupBase::getXPtr() const
+{
+  FOUR_C_THROW("not implemented");
```
mayrmt: Just a matter of style...
```suggestion
  FOUR_C_THROW("Not implemented.");
```

src/solver_nonlin_nox/4C_solver_nonlin_nox_linearsystem_base.hpp
```diff
@@ -136,6 +136,8 @@ namespace NOX
        * sharedLinearSystem.getObject(this)->setJacobianOperatorForSolve(
        *   sharedLinearSystem.getObject(this)->getJacobianOperator());
        *
+       * In GroupBase a similar call is completely removed.
+       *
```
mayrmt: Is this comment relevant? I do not understand it.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
maxfirmbach: I agree. We could just call it `NonlinearSolver::` similar to `LinearSolver`.

mayrmt: Good idea! Probably something for another PR, since it affects much more files that just this of the group.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linearsystem_base.hpp
vovannikov: I agree, a bit confusing. At the end, I have removed the entire lengthy comment around these lines since it is not longer relevant after removing `Epetra::Group`.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
vovannikov: I agree, surely worth doing that.

mayrmt (approved):
Thanks!

PR #1294, opened 2025-09-23
Move build_nodal_coordinates() tests to core

sebproell:
Since we have a helper function to create a test hypercube discretization, we can use it for more tests to make them independent of `Global::Problem`. This PR moves the test for `build_nodal_coordinates()`.

### DIFF
--- a/src/core/fem/tests/discretization/4C_fem_discretization_nodal_coordinates_test.cpp
+++ b/src/core/fem/tests/discretization/4C_fem_discretization_nodal_coordinates_test.cpp
@@ -82,41 +82,14 @@ namespace
       nodal_test_coordinates->ExtractCopy(coords.data(), nodal_test_coordinates->MyLength());
 
       // first coordinate
-      EXPECT_NEAR(coords[0], 0.0, 1e-14);
-      EXPECT_NEAR(coords[4], 0.0, 1e-14);
-      EXPECT_NEAR(coords[8], 0.0, 1e-14);
+      EXPECT_DOUBLE_EQ(coords[0], 0.0);
+      EXPECT_DOUBLE_EQ(coords[4], 0.0);
+      EXPECT_DOUBLE_EQ(coords[8], 0.0);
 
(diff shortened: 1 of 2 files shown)


sebproell:
@maxfirmbach The tests do not yet pass and I need your help here. Can you tell me what the tests `NodalCoordinatesPartialMap` are supposed to test? I assume you only want to extract coordinates of nodes with a GID given in a map.

However, I do not understand what the second case with ` std::array<int, 3> nodeList{50, 62, 114};  // random GIDs` is supposed to achieve when there are only 27 nodes in the test discretization. In my new version, all coordinates are returned as zero which looks correct when reading the implementation. What am I missing?

sebproell:
Github does not let me comment on unchanged lines, so for clarity: I am talking about this subcase: https://github.com/4C-multiphysics/4C/blob/main/unittests/io/4C_discretization_nodal_coordinates_test.cpp#L130-L152

How did this work before?

src/core/fem/tests/discretization/4C_fem_discretization_nodal_coordinates_test.cpp
```diff
@@ -120,9 +87,9 @@ namespace
       EXPECT_NEAR(coords[8], 0.0, 1e-14);
 
       // last coordinate
-      EXPECT_NEAR(coords[3], 0.0, 1e-14);
-      EXPECT_NEAR(coords[7], 0.5, 1e-14);
-      EXPECT_NEAR(coords[11], 0.0, 1e-14);
+      EXPECT_NEAR(coords[3], 0.5, 1e-14);
```
maxfirmbach: Is the last coordinate different now?

sebproell: I think I understand the issue now. GridGenerator labels GIDs very differently and they do not form a contiguous set. This specific test is correct; I checked the value. I will just adapt all tests now

PR #1297, opened 2025-09-24
Let Function framework take spatial position as a span

sebproell:
A few of our linear algebra data structures now provide their data as spans as well. This is safer than a raw pointer and works nicely with the functions taking such a span.

Needed for #1283 but useful on its own.


mayrmt (approved):

sebproell:
@maxfirmbach I already talked to @amgebauer about this: we could always return span rather than raw pointers in our linear algebra data structures or wrappers. You _could_ still call `.data()` on the span to get a raw pointer, but only if you really mean it, e.g. if you interface with C. Maybe something to consider for the linalg wrappers.

PR #1298, opened 2025-09-24
Feat: update Python version on thought

davidrudlstorfer:
## Description and Context

Update to a newly provided Python version on thought (following #1287)


davidrudlstorfer:
Works as expected https://gitlab.lrz.de/baci/baci-config-lnm/-/pipelines/2718022

Issue #1296, opened 2025-09-23
Move or use functionality to convert a dof-vector into a node-based multi-vector to or from core


lkoeglmeier:
## Description

In the porofluid code part we have a method implemented `convert_dof_vector_to_node_based_multi_vector(...) ` that converts a dof-based vector into a node-based multi-vector for post processing. As suggested 
As suggested [here](https://github.com/4C-multiphysics/4C/pull/1110#discussion_r2339699374), we should move or use  this functionality to or from core.

## Possible Solution and Definition of Done

[ ] Check if we already have a method in core available dof-based vectors into node-based multi-vectors
[ ] If not, move functionality to core 

## Interested Parties

@c-p-schmidt

mayrmt:
@lkoeglmeier What is special about poro, that such a transformation is required? How do other fields handle a similar situation? Do they also convert between vector formats or is there another solution?

lkoeglmeier:
> [@lkoeglmeier](https://github.com/lkoeglmeier) What is special about poro, that such a transformation is required? How do other fields handle a similar situation? Do they also convert between vector formats or is there another solution?

I would say that there is nothing special in poro. We aim to output some additional gp quantities here. These are already projected to the nodes and have to be a NODE quantity and therefore have to use a multivector for the output. We have also seen this functionality implemented in scatra. For the output of strains and stresses in structure, they use a function from core (`Core::FE::extrapolate_gauss_point_quantity_to_nodes(...)`). I just talked with @c-p-schmidt  and we will have a closer look and decide how we can unify this on Friday during our clean-up afternoon.

PR #1299, opened 2025-09-24
Remove unused function in scatra framework

lkoeglmeier:
This pr removes an unused function in the scatra framework.

Following #1296


davidrudlstorfer (approved):

c-p-schmidt (approved):
Thank you for removing the unnecessary remnant.

PR #1301, opened 2025-09-24
Hard wire the nodal nullspace sizes and options

maxfirmbach:
## Description and Context

This PR hard wires the nodal null space sizes and options for fluid style null spaces. It is a bit over constructed, but it helps to understand in which cases the nodal nullspace has a certain dimension. This is an intermediate step in a bigger cleanup effort.


src/fluid_ele/4C_fluid_ele.cpp
```diff
@@ -63,7 +63,23 @@ void Discret::Elements::FluidType::nodal_block_information(
 Core::LinAlg::SerialDenseMatrix Discret::Elements::FluidType::compute_null_space(
     Core::Nodes::Node& node, const double* x0, const int numdof, const int dimnsp)
 {
-  return FLD::compute_fluid_null_space(numdof, dimnsp);
+  switch (numdof)
+  {
+    case 3:
+      // 2D fluid
+      return FLD::compute_fluid_null_space<3>();
```
sebproell, line 70: Why not simply return an identity matrix here? If you believe it helps to have this function that is OK with me. Although it seems the helper is not really doing anything other than constructing an identity matrix.

src/fluid_ele/4C_fluid_ele.cpp
```diff
@@ -63,7 +63,23 @@ void Discret::Elements::FluidType::nodal_block_information(
...
+      // 3D fluid
+      return FLD::compute_fluid_null_space<4>();
+    case 8:
+      // 3D enriched fluid
+      return FLD::compute_fluid_null_space<8>();
+    default:
+      FOUR_C_THROW(
+          "The null space computation of a fluid element of dimension {} is not yet "
+          "implemented",
+          numdof);
```
mayrmt, line 81: At first glance, this error message could sound like the "dimension" refers to the fluid element, not to the null space. Maybe:
```suggestion
      FOUR_C_THROW(
          "The computation of a {}-dimensional null space is not yet implemented for the fluid element.",
          numdof);
```
The same could apply to other error messages of similar type.

mayrmt (dismissed):
Just an idea for rewording the error messages...

src/fluid_ele/4C_fluid_ele.cpp
maxfirmbach: Yes, that is true :sweat_smile:.

maxfirmbach: Actually I think we don't have any helper to generate an identity matrix for the `SerialDenseMatrix`, so I need one anyway.

maxfirmbach: So I would keep the helper for now. I'm pretty sure it will go away in further PR's related to this topic.

src/fluid_ele/4C_fluid_ele.cpp
maxfirmbach: I updated the error messages according to your recommendations.

src/fluid_ele/4C_fluid_ele.cpp
sebproell: But wouldn't it be more logical to provide this helper in the matrix class, rather than calling it `compute_fluid_nullspace` :laughing:

sebproell (approved):

PR #1303, opened 2025-09-24
Refactor nodal nullspace input types

maxfirmbach:
## Description and Context

This PR replaces the `const double*` with a `std::array` for the discretization centroid coordinates. In addition the unused `dimns` input parameter is removed from the nodal null space.

### DIFF
--- a/src/ale/4C_ale_ale2.cpp
+++ b/src/ale/4C_ale_ale2.cpp
@@ -70,7 +70,7 @@ void Discret::Elements::Ale2Type::nodal_block_information(
 /*----------------------------------------------------------------------------*/
 /*----------------------------------------------------------------------------*/
 Core::LinAlg::SerialDenseMatrix Discret::Elements::Ale2Type::compute_null_space(
-    Core::Nodes::Node& node, const double* x0, const int numdof, const int dimnsp)
+    Core::Nodes::Node& node, std::span<const double> x0, const int numdof)
 {
   return compute_solid_null_space<2>(node.x(), x0);
 }
--- a/src/ale/4C_ale_ale2.hpp
+++ b/src/ale/4C_ale_ale2.hpp
@@ -54,7 +54,7 @@ namespace Discret
       void nodal_block_information(Core::Elements::Element* dwele, int& numdf, int& dimns) override;
 
       Core::LinAlg::SerialDenseMatrix compute_null_space(
-          Core::Nodes::Node& node, const double* x0, const int numdof, const int dimnsp) override;
+          Core::Nodes::Node& node, std::span<const double> x0, const int numdof) override;
 
       void setup_element_definition(
           std::map<std::string, std::map<Core::FE::CellType, Core::IO::InputSpec>>& definitions)
@@ -491,7 +491,7 @@ namespace Discret
       }
 
       Core::LinAlg::SerialDenseMatrix compute_null_space(
-          Core::Nodes::Node& node, const double* x0, const int numdof, const int dimnsp) override
+          Core::Nodes::Node& node, std::span<const double> x0, const int numdof) override
       {
         Core::LinAlg::SerialDenseMatrix nullspace;
         FOUR_C_THROW("method ComputeNullSpace not implemented!");
--- a/src/core/fem/src/discretization/4C_fem_discretization_nullspace.cpp
+++ b/src/core/fem/src/discretization/4C_fem_discretization_nullspace.cpp
@@ -20,100 +20,92 @@ FOUR_C_NAMESPACE_OPEN
 namespace Core::FE
 {
   std::shared_ptr<Core::LinAlg::MultiVector<double>> compute_null_space(
-      const Core::FE::Discretization& dis, const int numdf, const int dimns,
-      const Core::LinAlg::Map& dofmap)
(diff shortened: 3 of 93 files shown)


mayrmt (dismissed):

maxfirmbach:
I had to keep the early return in the nullspace calculation for `dimns == 0`. There are two scatra dof-split cases that see to explicitly need this due to a mismatch in dofs given by the discretization and the one given by the element, somehow weird.

src/ale/4C_ale_ale2.cpp
```diff
@@ -70,7 +70,7 @@ void Discret::Elements::Ale2Type::nodal_block_information(
 /*----------------------------------------------------------------------------*/
 /*----------------------------------------------------------------------------*/
 Core::LinAlg::SerialDenseMatrix Discret::Elements::Ale2Type::compute_null_space(
-    Core::Nodes::Node& node, const double* x0, const int numdof, const int dimnsp)
+    Core::Nodes::Node& node, std::array<double, 3> x0, const int numdof)
```
amgebauer: Please use std::span instead of std::array.
```suggestion
    Core::Nodes::Node& node, std::span<double, 3> x0, const int numdof)
```

And is the 3 correct for ale2? It might need to be coordinated with #1283

maxfirmbach: Good point. Currently it was always calculated for 3 dimensions, even though the last one is zero.

eulovi (approved):

src/ale/4C_ale_ale2.cpp
maxfirmbach: @amgebauer I'm now using `std::span`. I did not adjust the dimension though.

mayrmt (approved):

amgebauer (approved):
Thanks for the cleanup

src/ale/4C_ale_ale2.cpp
amgebauer: @sebproell Any issues that might arise with https://github.com/4C-multiphysics/4C/pull/1283?

src/core/fem/src/discretization/4C_fem_discretization_nullspace.cpp
```diff
@@ -20,97 +20,89 @@ FOUR_C_NAMESPACE_OPEN
...
-      std::array<double, 3> x0send = {0.0, 0.0, 0.0};
-      for (int i = 0; i < dis.num_my_row_nodes(); ++i)
-        for (int j = 0; j < 3; ++j) x0send[j] += dis.l_row_node(i)->x()[j];
 
-      std::array<double, 3> x0;
-      Core::Communication::sum_all(x0send.data(), x0.data(), 3, dis.get_comm());
+    // for rigid body rotations compute nodal center of the discretization
+    std::array<double, 3> x0send = {0.0, 0.0, 0.0};
+    for (int i = 0; i < dis.num_my_row_nodes(); ++i)
+      for (int j = 0; j < 3; ++j) x0send[j] += dis.l_row_node(i)->x()[j];
```
sebproell: This is problematic with #1283

src/ale/4C_ale_ale2.hpp
```diff
@@ -54,7 +54,7 @@ namespace Discret
       void nodal_block_information(Core::Elements::Element* dwele, int& numdf, int& dimns) override;
 
       Core::LinAlg::SerialDenseMatrix compute_null_space(
-          Core::Nodes::Node& node, const double* x0, const int numdof, const int dimnsp) override;
+          Core::Nodes::Node& node, std::span<const double, 3> x0, const int numdof) override;
```
sebproell: I don't think it makes a lot of sense to hard-code this to dim `3`. With #1283, we will have correct dynamic coordinate numbers. This span should probably be


```suggestion
          Core::Nodes::Node& node, std::span<const double> x0, const int numdof) override;
```

Ofc, you can can assert inside the derived classes that the number of coordinates is as you expect.

sebproell (commented):
@maxfirmbach Could you wait for #1283 and base your changes on that?

maxfirmbach:
@sebproell I can wait, yes.

mayrmt (dismissed):

src/core/fem/src/discretization/4C_fem_discretization_nullspace.cpp
```diff
@@ -20,100 +20,92 @@ FOUR_C_NAMESPACE_OPEN
...
+    if (dimns == 1)
     {
-      // compute nullspace for simple case: vector of ones
       nullspace->PutScalar(1.0);
+      return nullspace;
     }
-    else
+
+    // for rigid body rotations compute nodal center of the discretization
+    std::array<double, 3> x0send = {0.0, 0.0, 0.0};
```
sebproell: FYI

```suggestion
    std::array<double, 3> x0send{};
```

sebproell (dismissed):

src/core/fem/src/discretization/4C_fem_discretization_nullspace.cpp
maxfirmbach: I changed it.

mayrmt (approved):

PR #1306, opened 2025-09-25
[solver] Remove leftovers of `ML`

maxfirmbach:
## Description and Context

As the title states. This PR removes leftover parameterlists related to `ML`. At some point we also started to use `ML` based parameters as they were our own, which is not good. These are removed in addition.

In addition a test using an `Ifpack` preconditioner for a block system from contact mechanics is removed. It only worked on one proc and never turned out to be a thing anyway.


tests/list_of_tests.cmake
```diff
@@ -318,7 +318,6 @@ four_c_test(TEST_FILE contact2D_twobeamsvertical.4C.yaml NP 2)
 four_c_test(TEST_FILE contact3D_adhesion_fixedbound_new_struc.4C.yaml NP 2)
 four_c_test(TEST_FILE contact3D_bin_strat.4C.yaml NP 3)
 four_c_test(TEST_FILE contact3D_bin_strat_new_struct.4C.yaml NP 3)
-four_c_test(TEST_FILE contact3D_blocked_system_ilu_precon.4C.yaml NP 1)
```
mayrmt, line 321: Why is this test deleted?

maxfirmbach: "In addition a test using an Ifpack preconditioner for a block system from contact mechanics is removed. It only worked on one proc and never turned out to be a thing anyway."

mayrmt: Sorry, I missed that.

maxfirmbach: Some time ago I tested this with Chris to have an "easy" option to solve the block system with an iterative solver ... it never worked well though.

mayrmt (approved):

maxfirmbach:
Of course one test had to fail :sweat_smile:.

sebproell (approved):

PR #1305, opened 2025-09-25
Fix: include of span in serial dense vector

davidrudlstorfer:
## Description and Context

Does this already solve the problem of #1304 ?

## Related Issues and Pull Requests

Closes #1304

PR #1307, opened 2025-09-25
Check for internal standard library includes

sebproell:
It happens surprisingly often that one accidentally includes an internal standard library header. Instead of having to watch out during review, let's use a pre-commit check to catch this.

See e.g. #1305


sebproell:
I am checking for `#include <c++/.*` and `#include <bits/.*`. If we see other wrong includes, this is easy to extend.

davidrudlstorfer (approved):

utilities/four_c_python/src/four_c_development/check_includes.py
```diff
@@ -68,10 +68,17 @@ def check_and_fix_includes(file: Path) -> Tuple[list, bool]:
...
+
+        # External includes must use angle brackets
+        if not "<" in line:
+            invalid_include_lines.append([line_no + 1, line])
+            continue
 
-        else:
-            if not "<" in line:
-                invalid_include_lines.append([line_no + 1, line])
+        match = re.search(r"#include <(c\+\+.*|bits/)>", line)
```
amgebauer:
```suggestion
        match = re.search(r"#include <(c\+\+|bits)/.*>", line)
```

sebproell: Ah right, good catch!

PR #1310, opened 2025-09-25
Add some vtu preprocessing information to documentation

amgebauer:
This adds a small section to the preprocessing section.

We would also need a tutorial showing how to do it step by step. I will do it in another PR.

### DIFF
--- a/doc/documentation/src/analysis_guide_templates/preprocessing.rst
+++ b/doc/documentation/src/analysis_guide_templates/preprocessing.rst
@@ -28,9 +28,30 @@ Creating meshes for |FOURC|
 
 #. Either you create the mesh in |FOURC|'s native format directly. Refer to the :ref:`tools` section
    to find tools that can help you to create |FOURC| input files.
-#. or you create a mesh file in a general binary format for finite element information, called EXODUS II, develeloped by `Sandia National Laboratories
-   <https://www.sandia.gov/files/cubit/15.8/help_manual/WebHelp/finite_element_model/exodus/exodus2_file_specification.htm>`_.
-   This can be read into |FOURC| via its input file.
+#. or you directly read the mesh from standardized mesh formats. |FOURC| supports reading vtu-files,
+   or EXODUS II.
+
+Generating ``VTU`` files
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+VTU is a standard file format for unstructured grids and is part of the VTK library. VTU files
+can be generated from various mesh formats using the Python library `meshio <https://pypi.org/project/meshio/>`_,
+or they can often be exported directly from meshing software.
+
+|FOURC| uses element blocks to assign physical fields and their properties to subsets of elements
+in the mesh. The VTU file must contain a scalar cell data array called ``block_id``, which assigns
+an integer ID to each element.
+
+.. admonition:: Warning
+
+   Each block may only contain elements of the same type (e.g., all hexahedra or all tetrahedra).
+   For mixed meshes, multiple block IDs must be defined. Multiple blocks can be assigned to the same
+   physical field in the |FOURC| input file.
+
+To define boundary conditions, |FOURC| relies on the definition of point sets. |FOURC| creates
+point sets from integer-type point data arrays named ``point_set_<id>``. A point is interpreted as
+belonging to a set if the corresponding entry in the point data array is nonzero.
+
 
 Generating ``EXODUS II`` files
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


doc/documentation/src/analysis_guide_templates/preprocessing.rst
```diff
@@ -28,9 +28,28 @@ Creating meshes for |FOURC|
...
+in the mesh. The VTU file must define a cell data array called ``block_id``, which assigns an integer
+ID to each element.
+
+.. admonition:: Warning
+   Each block may only contain elements of the same type (e.g., all hexahedra or all tetrahedra).
+   For mixed meshes, multiple block IDs must be defined. Multiple blocks can be assigned to the same
+   physical field in the |FOURC| input file.
+
+To define boundary conditions, |FOURC| relies on the definition of point sets. |FOURC| creates
+point sets from integer-type point data arrays named ``point_set_<id>``. A point is interpreted as
```
ischeider, line 52: To me this documentation seems too short. Maybe a bit more information like this is useful:

```suggestion
point sets from integer-type point data arrays named ``point_set_<id>``, that is, a point data array has to be defined in the vtu file file each node set with a length of the number of nodes. The id in the name should start with 1. A point is then interpreted as
```

doc/documentation/src/analysis_guide_templates/preprocessing.rst
```diff
@@ -28,9 +28,28 @@ Creating meshes for |FOURC|
...
+Generating ``VTU`` files
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+VTU is a standard file format for unstructured grids and is part of the VTK library. VTU files
+can be generated from various mesh formats using the Python library `meshio <https://pypi.org/project/meshio/>`_,
+or they can often be exported directly from meshing software.
+
+|FOURC| uses element blocks to assign physical fields and their properties to subsets of elements
+in the mesh. The VTU file must define a cell data array called ``block_id``, which assigns an integer
+ID to each element.
```
georghammerl:
```suggestion
|FOURC| uses element blocks to assign physical fields and their properties to subsets of elements
in the mesh. The VTU file needs to contain a cell data array called ``block_id``, which assigns an integer ID to each element. This ID can then be referenced in the ``ELEMENT_BLOCKS`` ID in the |FOURC| input file.
```

georghammerl: Not sure, tried to make the link with the ID more clear.

sebproell: While I agree that this kind of link is useful, we should always try to keep it in sync with the actual input name. Can we use the metadata linking here?

georghammerl:
@amgebauer Thank you for the vtu description in the docu here.
Is it possible to name helper tools/tricks to assign the `block_id` and `point_set_<id>` in the vtu file efficiently?

doc/documentation/src/analysis_guide_templates/preprocessing.rst
ischeider:
```suggestion
in the mesh. The user must define a cell data array in the VTU file called ``block_id``, which assigns an integer
ID to each element, starting from 1.
```

ischeider (commented):
For me, the explanations seem a bit short, I tried to explain a bit more; the tutorial will make things much clearer I guess.

doc/documentation/src/analysis_guide_templates/preprocessing.rst
amgebauer: I also think that we should cover this in a tutorial so that it does not get outdated and is part of testing.

doc/documentation/src/analysis_guide_templates/preprocessing.rst
amgebauer: It also works if you start you id with `0` (at least the file that I tested locally)

amgebauer:
I slightly extended the docu, but I think we should put more details in the tutorial. Then the documentation does not get outdated and we can include the relevant sections automatically.

doc/documentation/src/analysis_guide_templates/preprocessing.rst
ischeider: Can you really assign a 0 as boundary condition set?

ischeider:
I don't see that any automation can happen soon, since no tool will provide point or block sets. In the tutorial you will probably show how the user can include the arrays, but this manual editing will be necessary for a long time, so I still would like to see the extension that both  @georghammerl and me found useful.

What kind of "automation" do you have in mind?

amgebauer:
> I don't see that any automation can happen soon, since no tool will provide point or block sets. In the tutorial you will probably show how the user can include the arrays, but this manual editing will be necessary for a long time, so I still would like to see the extension that both @georghammerl and me found useful.

> What kind of "automation" do you have in mind?

Here I just meant to automatically include 4C.yaml-files within the docu so that we don't have to manually update the docu if things change in the input file #694.

I think defining here exactly how one can assign blocks to physics and point-sets to boundary condition does not really fit into the documentation here. The remaining docs in this page (for exodus ii), does also not define that. It is basically anyway the same for all mesh input formats. We have to document that once centrally, i.e., how one can assign physics and materials to element blocks and BCs to point-sets, and then, only specify for each formats how one needs to define blocks and point-sets.

What automation are you thinking of? Directly using the generated mesh from the meshing software? I think that can easily be enabled by allowing input from the Gmsh-format. It has physical entities (points, edges, surfaces and volumes) that can (out of the box) be used to assign physics and boundary conditions. For me personally, I will in most cases anyway have to do some more steps than meshing before I can run simulations. I will do these preprocessing steps with our new python package [biomesh](https://github.com/TUM-LNM/biomesh), which will be equipped with all necessary tools to generate complex patient-specific geometries with vtu-output in the way 4C can understand and work with. I think many scenarios will need such an extra step between meshing and simulation for computing fibers, parameter fields, etc.

doc/documentation/src/analysis_guide_templates/preprocessing.rst
amgebauer: Yes, I tested it on the vtu test that we have checked in in the repo.

georghammerl:
I see the point, that manual adaptions are okay for you when dealing with vtu files.
I was not aware of that fact when discussing with you @amgebauer the vtu import. 
I seek for a mesh format (which is then likely the Gmsh-format) which can directly be imported from many widely-used formats via conversion within meshio (without the need of manual adaptions). Hence, I it is fine for me to keep this short docu here for vtu.

c-p-schmidt (approved):

PR #1312, opened 2025-09-25
Improve Discretization::redistribute and fill_complete

sebproell:
- Make the meaning of arguments more obvious at the call site by using option structs.
- Prefer to to use redistribute() instead of manual `export_xxx` calls.

Part of #1228

### DIFF
--- a/src/art_net/4C_art_net_dyn_drt.cpp
+++ b/src/art_net/4C_art_net_dyn_drt.cpp
@@ -112,7 +112,11 @@ std::shared_ptr<Adapter::ArtNet> dyn_art_net_drt(bool CoupledTo3D)
     if (scatradis->add_dof_set(arterydofset) != 1)
       FOUR_C_THROW("unexpected dof sets in scatra field");
 
-    scatradis->fill_complete(true, false, false);
+    scatradis->fill_complete({
+        .assign_degrees_of_freedom = true,
+        .init_elements = false,
+        .do_boundary_conditions = false,
+    });
   }
   else
   {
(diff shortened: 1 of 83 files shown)


src/art_net/4C_art_net_dyn_drt.cpp
```diff
@@ -112,7 +112,7 @@ std::shared_ptr<Adapter::ArtNet> dyn_art_net_drt(bool CoupledTo3D)
     if (scatradis->add_dof_set(arterydofset) != 1)
       FOUR_C_THROW("unexpected dof sets in scatra field");
 
-    scatradis->fill_complete(true, false, false);
+    scatradis->fill_complete({true, false, false});
```
mayrmt: What's the difference here in readability? I still have to manually look up the meaning of each of the booleans?

maxfirmbach: In theory one could also write {.option_1 = true, .option_2 = true, ...} if I'm not mistaken + if one uses a proper IDE it will be shown in a nicer way. In general I like these changes, because they give the possibility to make the code more readable, which is not the case if one uses plain input arguments.

maxfirmbach:
For readability reasons one could actually think about writing the options explicitly in the struct initialization.

src/art_net/4C_art_net_dyn_drt.cpp
sebproell: @maxfirmbach Exactly. Even "just the braces" help.

@mayrmt You have a point here. Since I touch all places anyway, I should really use the best syntax right away. I regex replaced these occurrences and forgot to use designated initializers in the process. I will update it.

mayrmt (approved):

src/art_net/4C_art_net_dyn_drt.cpp
mayrmt: @sebproell It mostly felt weird to me, since there was again an implicit ordering of the arguments and I do not now which one. Is it the ordering of the three options in side the struct? What would happen if the struct's declaration is reordered, but the call sites just use a sequence of booleans. Will they still map correctly to the internal options?

sebproell: Technical details: https://en.cppreference.com/w/cpp/language/aggregate_initialization.html (scroll down to "deisgnated initializers"). 

I think what you are asking is whether the struct version w/o designated initializers is any better than the non-struct version. I would say not really, except that many editors insert useful hints. You really should use the designated initializers. Then, a change in the struct member ordering will no longer compile, which is exactly what we want.

PR #1314, opened 2025-09-26
Refactor beam-to-beam point couplings

isteinbrecher:
## Description and Context

Refactor the beam-to-beam point couplings. The code produces the same results, but is adapted such that the actual evaluation of the coupling constraints is moved to the methods `evaluate_positional_coupling` and `evaluate_rotational_coupling`, allowing for an easy extension to different beam theories. Where easily possible, analytic derivatives are used. Furthermore, the penalty regularisation is refactored such that it is easy to use these couplings directly with a Lagrange multiplier method.


knarfnitram:
I am fine with the changes, I just have some comments regarding the documentation:
At multiple comments the `.` is missing and I think some equations to elaborate more the implemented changes could be helpful for understanding.

isteinbrecher:
> At multiple comments the . is missing 

Do you mean the period at the end? Do we have a guideline for that?

> I think some equations to elaborate more the implemented changes could be helpful for understanding.

In principle I agree, however, I think just some equations here will not really be useful as the whole coupling terms are quite complicated and warrant more explanation than a simple equation.

mayrmt:
> Do you mean the period at the end? Do we have a guideline for that?

For one-line comments such as `// Get stuff` or `// Initialize some object`, we usually do not use periods to delimit the comment. I am fine with that. In fact, I prefer that style.

PR #1316, opened 2025-09-26
Add remapping function for AD types

isteinbrecher:
## Description and Context

Add functionality to remap the indices of depended FAD derivatives.

Basically, one can compute AD values with a certain number and ordering of dependent variables and then convert that variable to another AD variable with a different number and ordering of dependent variables.

This is needed to enable a simple averaged nodal normals / tangent smoothing implementation.

### DIFF
--- a/src/core/utils/src/numerics/4C_utils_fad.hpp
+++ b/src/core/utils/src/numerics/4C_utils_fad.hpp
@@ -266,6 +266,105 @@ namespace Core::FADUtils
     static double apply(const int n, const int i, const double x) { return x; }
   };
 
+  /**
+   * \brief Get the depth of a possibly nested FAD scalar type.
+   */
+  template <class ScalarType>
+  constexpr std::size_t get_nested_fad_type_depth()
+  {
+    if constexpr (Sacado::IsADType<ScalarType>::value)
+    {
+      using Inner = typename Sacado::ValueType<ScalarType>::type;
+      return 1 + get_nested_fad_type_depth<Inner>();
+    }
+    else
+    {
+      return 0;
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


src/core/utils/src/numerics/4C_utils_fad.hpp
```diff
@@ -266,6 +266,105 @@ namespace Core::FADUtils
...
+    }
+  }
+
+  /**
+   * \brief Store the information needed to remap the indices of FAD derivatives.
+   */
+  struct RemapInformation
+  {
+    //! Map the local IDs of the incoming FAD type to the local IDs of the outgoing type.
+    std::vector<int> old_local_id_to_new_local_id_;
```
amgebauer:
```suggestion
    std::vector<int> old_local_id_to_new_local_id;
```
It's better readable if you remove the trailing `_` in structs. They are only required (and meaningful) for private member variables.

src/core/utils/src/numerics/4C_utils_fad.hpp
```diff
@@ -266,6 +266,105 @@ namespace Core::FADUtils
...
+
+    //! New derivative space length for this level. This is needed to correctly initialize the
+    //! remapped variables.
+    int n_dependent_variables_;
```
amgebauer: same here


```suggestion
    int n_dependent_variables;
```

amgebauer (dismissed):

src/core/utils/src/numerics/4C_utils_fad.hpp
isteinbrecher: Makes sense, I adapted it.

PR #1320, opened 2025-09-26
Add full variant of beam-to-beam point couplings

isteinbrecher:
## Description and Context

Add a "full" variant of the beam-to-beam point coupling. This now works also for cases where the two beams are offset from each other in the reference configuration. This requires more input arguments to the evaluation of the positional constraints and leads to moments arising from only the positional coupling.


tests/input_files/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_elbow_offset.4C.yaml
```diff
@@ -0,0 +1,169 @@
...
+  INTEGRATION_POINTS: true
+  AVERAGED_NORMALS: true
+BEAM INTERACTION/BEAM TO SOLID VOLUME MESHTYING/RUNTIME VTK OUTPUT:
+  WRITE_OUTPUT: true
+  NODAL_FORCES: true
+  MORTAR_LAMBDA_DISCRET: true
+  MORTAR_LAMBDA_CONTINUOUS: true
+  MORTAR_LAMBDA_CONTINUOUS_SEGMENTS: 5
+  SEGMENTATION: true
+  INTEGRATION_POINTS: true
```
isteinbrecher: Just saw that these are here, I will remove them.

mayrmt (approved):

PR #1321, opened 2025-09-26
Allow reading input fields directly from the (VTU-)mesh

amgebauer:
This PR allows to directly read input fields from the VTU-mesh.

The first commit generalizes the reference based input field and separates the initialization and the redistribution part.

The second commit then allows to read in the data directly from the vtu file instead of the json file.

The nice thing about this is, that we don't have to care about the mapping in the separate file. We can directly check with ParaView whether the input field is correct. It works at all places where 4C accepts input fields (which can be easily be enabled where needed).

Accepted input types are scalars, vectors, symmetric tensors and tensors (with scalar types `int` and `double`).

```yaml
MATERIALS:
  - MAT: 1
    MAT_ElastHyper:
      NUMMAT: 2
      MATIDS: [11, 12]
      DENS: 1
  - MAT: 11
    ELAST_IsoNeoHooke:
      MUE:
        from_mesh: mue
# ...
fields:
  - name: mue
    discretization: "structure"
    source:
      from_mesh: {}
```


amgebauer:
@maxfirmbach Since there are now two options to define the input field, I have to adapt the way we define the fields. Four your thermo3D example, it is now

```yaml
fields:
  - name: spatially_varying_conductivity
    discretization: "thermo"
    source:
      separate_file:
        file: conductivity_input_field.json
        key: "CONDUCT"
```

tests/input_files/solid_vtu_field_input.4C.yaml
```diff
@@ -0,0 +1,136 @@
...
+      SOLID:
+        TET10:
+          MAT: 1
+          KINEM: nonlinear
+  FILE: solid_vtu_field_input.vtu
+fields:
+  - name: mue
+    discretization: "structure"
+    source:
+      from_mesh: {}
```
gilrrei: This looks weird. Why pass an empty map?

amgebauer: The `from_mesh` option has an optional `key` to define how the variable is called in the mesh-file (and it is necessary since `source` is a selection). But I agree, it is weird.

I anyway want to add a mandatory option to specify whether it is cell or point-data. Then it should be less weird.

amgebauer: Now it is

```yaml
  - name: mue
    discretization: "structure"
    source:
      from_mesh:
        basis: cells
```

We do not yet support points as basis, but I hope it is coming soon.

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -91,6 +104,66 @@ namespace Core::IO
    */
   InputFieldRegistry& global_input_field_registry();
 
+
+  /**
+   * Mesh data reference. This name is used initialize the field from the mesh data.
```
sebproell:
```suggestion
   * Mesh data reference. This name is used to initialize the field from the mesh data.
```

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -261,6 +263,53 @@ namespace Core::IO::MeshInput
...
+  template <typename IndexType, typename T, unsigned dim>
+    requires(std::is_integral_v<IndexType>)
+  void read_value_from_element_data(
+      const Mesh<dim>& mesh, const std::string& key, std::unordered_map<IndexType, T>& value)
+  {
+    unsigned ele_id = 0;
+    for (const auto& [block_id, block] : mesh.cell_blocks)
+    {
+      // This block is not part of any physics, hence ignore it
+      if (!block.specific_data) continue;
```
sebproell: Will clash with my changes in #1328

amgebauer: I would need the already filtered mesh here

maxfirmbach:
> @maxfirmbach Since there are now two options to define the input field, I have to adapt the way we define the fields. Four your thermo3D example, it is now
> 
> ```yaml
> fields:
>   - name: spatially_varying_conductivity
>     discretization: "thermo"
>     source:
>       separate_file:
>         file: conductivity_input_field.json
>         key: "CONDUCT"
> ```

@amgebauer That is fine for me :+1:.

lauraengelhardt:
Very nice and much more usable (at least for most cases) than the json files :+1: 
Out of curiosity: If the mesh file is not vtu do we get an error, or how is this handled?

amgebauer:
@lauraengelhardt 

> Out of curiosity: If the mesh file is not vtu do we get an error, or how is this handled?

There is an error that the field data was not found in the mesh.

amgebauer:
The PR is ready for review

c-p-schmidt (approved):

PR #1323, opened 2025-09-26
Allow to filter Mesh input

sebproell:
This PR introduces a distinctions between `RawMesh` which is close to a mesh file, and `Mesh` which is an abstraction on top of `RawMesh`. This enables us to implement filtering operations on a mesh, e.g., filtering by the `cell_block_id`. Filtering is useful for all kinds of coupled physics, where one mesh file contains all the info, but we do not constantly want to check which pieces we need from the mesh. Instead, we filter the mesh once and then process the whole (filtered) mesh.

### DIFF
--- a/src/core/io/src/4C_io_mesh.hpp
+++ b/src/core/io/src/4C_io_mesh.hpp
@@ -14,6 +14,7 @@
 #include "4C_io_input_parameter_container.hpp"
 #include "4C_linalg_symmetric_tensor.hpp"
 #include "4C_utils_exceptions.hpp"
+#include "4C_utils_owner_or_view.hpp"
 
 #include <cstddef>
 #include <map>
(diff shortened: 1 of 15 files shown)


src/core/io/src/4C_io_mesh.hpp
```diff
@@ -237,11 +242,196 @@ namespace Core::IO::MeshInput
...
+  namespace Internal
+  {
+    /**
+     * A lightweight reference to a point in a Mesh used to provide a nicer interface.
+     *
+     * @note This class does not own any data and only refers to data in the RawMesh. Thus, it can
+     * only be used as long as the corresponding RawMesh is alive.
+     */
+    template <unsigned dim, bool is_const>
+    struct PointReference
```
amgebauer, line 254: We also need access to the `point_data` of the mesh.

sebproell: Done.

amgebauer (approved):

PR #1327, opened 2025-09-29
Limit number of 4C invocations per batch in coverage

### LINKED ISSUE #1325: [BUG] Coverage is failing
We again see the known failure of going out of disk space.

`System.IO.IOException: No space left on device`

amgebauer:
Each invocation of 4C during the coverage report generates a lot of temporary raw coverage data. For test chunks with small test runtimes but large number of 4C invocations (i.e., chunks with small test runtimes), this is problematic since we run out of storage.

Now I additionally add the feature to limit the number of tests*processors per chunk.

In my fork, I test whether this fixes the nightly pipeline: https://github.com/amgebauer/4C/actions/runs/18091843309

Closes #1325


sebproell:
@amgebauer Does it work?

amgebauer:
> @amgebauer Does it work?

not yet

amgebauer:
The coverage report is passing again after my recent changes: https://github.com/amgebauer/4C/actions/runs/18121775619

For our chunking to work correctly, dependent tests should ideally be placed directly after the tests they depend on. If they end up in different chunks, the dependent test will have to run again. However, this repeated execution time is not counted, which can cause us to exceed certain limits (in this case, the storage limit).

To address this, I placed the postprocessing tests immediately after their base simulations so they are likely to end up in the same chunk.

Ready for review.

sebproell (approved):

PR #1328, opened 2025-09-29
DiscretizationBuilder for incremental construction

sebproell:
Instead of repeatedly calling `add_node` and `add_element` on the `Discretization` we should gather all data first, ensure it is consistent and then build a discretization all at once. This PR provides `DiscretizationBuilder` as a first step to achieve this. Currently, building happens fully on rank 0; so this is not yet ready to be rolled out for the old legacy input, or at least not without checking the implications. Thus, we use the builder only for unit tests and thereby already test it.

Depends on #1323 

Part of #1228

### DIFF
--- a/unittests/common/unittest_utils/4C_unittest_utils_create_discretization_helper_test.hpp
+++ b/unittests/common/unittest_utils/4C_unittest_utils_create_discretization_helper_test.hpp
@@ -161,10 +161,10 @@ namespace TESTING
     const double imbalance_tol(1.1);
 
     Core::Rebalance::RebalanceParameters rebalance_parameters;
-    rebalance_parameters.mesh_partitioning_parameters.set("MIN_ELE_PER_PROC", total_ranks);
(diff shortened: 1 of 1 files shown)


src/core/io/src/4C_io_mesh.hpp
```diff
@@ -237,11 +235,222 @@ namespace Core::IO::MeshInput
...
+        return std::visit([this](auto& variant_vector) -> EligibleFieldTypes<dim>
+            { return variant_vector[index_]; }, raw_mesh_->point_data.at(field_name));
+      }
+
+      template <typename T>
+      // requires eligible
+      [[nodiscard]] const T& data_as(const std::string& field_name) const
+      {
+        const auto* vector = std::get_if<std::vector<T>>(&raw_mesh_->point_data.at(field_name));
+        FOUR_C_ASSERT_ALWAYS(vector, "bal");
```
amgebauer: Could you fix these ones here?

amgebauer (approved):

PR #1344, opened 2025-10-02
Move inpar particle and pasi

ppraegla:
Part of #59


davidrudlstorfer:
Nice - thanks for moving :)

PR #1346, opened 2025-10-06
Rework of the solid tutorial

### LINKED ISSUE #1337: Update solid mechanics tutorial
## Description

Based on https://github.com/mayrmt/UKACM_GACM_Tutorial_4C_QUEENS, let's update the solid mechanics tutorial located at `tests/tutorials/solid`.

## Interested Parties

@c-p-schmidt

c-p-schmidt:
## Description and Context
For the UKACM GACM workshop, where both 4C and QUEENS have been presented the solid tutorial was reworked to have some more information and with the intention to guide people step-by-step through a solid mechanics analysis.

This is now ported to the main repo.

Closes #1337


c-p-schmidt:
Note: Along the way, I switched from rst to md.

tests/tutorials/solid/tutorial_solid.4C.yaml
```diff
@@ -1,17 +1,58 @@
...
         HEX27:
           MAT: 1
           KINEM: nonlinear
-
+STRUCTURAL DYNAMIC:
+  INT_STRATEGY: Standard
+  DYNAMICTYPE: Statics
+  TIMESTEP: 0.2
+  RESTARTEVERY: 60
+  NUMSTEP: 60
```
c-p-schmidt: Some more information on the tutorial test.

In the pipeline, the complete simulation runs on the coarsest mesh with the iterative solver. This takes about 65 s on my workstation. Do we accept this, or should we reduce the number of time steps for the pipeline and a comment in this line that for the complete simulation this number has to be increased to 60?

tests/tutorials/solid/tutorial_solid.4C.yaml
```diff
@@ -1,17 +1,58 @@
...
+      SATHARDENING: 1000
+      HARDEXPO: 5
+      VISC: 0.01
+      RATE_DEPENDENCY: 1
+SOLVER 1:
+  SOLVER: Superlu
+  NAME: direct_solver
+SOLVER 2:
+  SOLVER: Belos
+  SOLVER_XML_FILE: iterative_gmres_porofluid_pressure_based.xml
```
c-p-schmidt: I do not really like this, but with the standard XML file, it does not work as more than 100 iterations are needed.

How should we proceed here?

tests/tutorials/solid/tutorial_solid.4C.yaml
amgebauer: I would reduce the number of timesteps or increase the mesh size. Our long testing times are super annoying.

tests/tutorials/solid/tutorial_solid.4C.yaml
amgebauer: I would suggest to change the max number of iterations in our template xml file

tests/tutorials/solid/tutorial_solid.4C.yaml
ischeider: Wouldn't it make more sense to remove the tutorials from the testing? These should not test a feature, but rather enable the users to run 4C simulations. We could also use the testing flags more, it's possible to exclude them if they get the "tutorial" flag (`ctest -LE tutorial`).

c-p-schmidt: I would not remove them from testing, as it won't take long until they will not run anymore with the current version of 4C.
Since these tutorials should be the low-effort entry point for new users, we can't expect them to resolve such issues before they can actually start playing around with 4C.

tests/tutorials/solid/tutorial_solid.4C.yaml
maxfirmbach: I agree, we can increase the value.

tests/tutorials/solid/tutorial_solid.4C.yaml
ischeider: Sorry, my text was a bit misleading. Of course, I did not mean to remove them completely, just to remove them from the test of each single commit. Wouldn't it be sufficient to test them nightly? It could happen that people change the input without noticing immediately that the tutorials don't work, but it will pop up in due time, and the errors are easy to fix.

Anyway, in the old version of the tutorial the number of time steps was significantly reduced:
```
Note that the number of time steps is currently set to 10, 
in order to keep the simulation time short for the CI/CD tests. 
If you want to run the whole simulation including necking of the specimen, 
you have to change the parameter to at least NUMSTEP: 200.
```
So, I am fine with @amgebauer 's comment.

c-p-schmidt:
One more thought. I accidentally removed the `RESULT DESCRIPTION` section, meaning that no result checks are performed anymore.
After thinking about it, I think that this is better (at least for this tutorial, but probably for most, if not all), as the people are supposed to play around with the settings, which would result in failing result checks, obviously. Again, in the spirit of a low-effort entry barrier, inexperienced people might be deterred by that. And in the spirit of what @ischeider was saying above. The tutorials are not meant to test physics, so I would argue that we do not really lose anything.
What do you think?

sebproell:
It could make sense to introduce a `--check-results` flag on `4C` which needs to be set for the result check to run.

c-p-schmidt:
This test `fsi_dc_mono_mtrss_ada_ab2_none.4C.yaml-p2` failed two times now. Did anyone else experience this in recent times?

The only thing that was changed related to this test is that the `Maximum Iterations` in the linear solver XML file has been increased. So if this was the problem, it should have failed before and not now. ;-)

maxfirmbach:
@mayrmt Should we adapt the `fsi` test tolerances of the failing test?

amgebauer:
This test has the setting `THROW_IF_UNCONVERGED: false`, which probably now causes the problem (as the solution is now better)

mayrmt:
@maxfirmbach @c-p-schmidt @amgebauer

Yes, let's loosen the tolerances of the FSI test a little bit. They seem very tight, especially in comparison to solver tolerances. Furthermore, can we switch to `THROW_IF_UNCONVERGED: true` now, since the max number of iterations has been increased? (One could just remove that line, since `true` should be the default value.)

c-p-schmidt:
I have now removed the `THROW_IF_UNCONVERGED: false` as it does not have an effect anymore and updated the results and tolerances accordingly, as other results are to be expected as outlined above by @amgebauer.

mayrmt (approved):

c-p-schmidt:
Can someone have a look, please?

tests/tutorials/solid/tutorial_solid.4C.yaml
```diff
@@ -37,73 +78,6 @@ DESIGN SURF DIRICH CONDITIONS:
...
-      VALUE: 0.4561235910743636
-      TOLERANCE: 1e-12
-      NAME: "elongation"
-  - STRUCTURE:
-      DIS: "structure"
-      NODE: 6
-      QUANTITY: "dispx"
-      VALUE: -0.04256648855679243
-      TOLERANCE: 1e-12
-      NAME: "necking"
```
amgebauer, line 109: Do we really not want to test the results?

ischeider (commented):
I hope it's not too late, but you have two links to the documentation in your text, which point to the absolute URL of the documentation. With {ref}`Link text <link_anchor>` you can point to the documentation without an absolute URL. I highly recommend this (e.g., since one can also use it locally).

ischeider:
I wonder whether the cubit journal files should really be placed in the `tests/tutorial` directory, since they are not tested at all. 

In my opinion they rather belong to the documentation. Shouldn't they be placed in `_assets` and provided as a download link in the documentation text?

c-p-schmidt:
> I hope it's not too late, but you have two links to the documentation in your text, which point to the absolute URL of the documentation. With {ref}`Link text <link_anchor>` you can point to the documentation without an absolute URL. I highly recommend this (e.g., since one can also use it locally).

It would not be too late, but it does not work for me. What you posted is probably rst style and it did not work in markdown. I also tried different markdown style links but it never worked. Maybe because these are not links within the same file?

c-p-schmidt:
> I wonder whether the cubit journal files should really be placed in the `tests/tutorial` directory, since they are not tested at all.
> 
> In my opinion they rather belong to the documentation. Shouldn't they be placed in `_assets` and provided as a download link in the documentation text?

You can open an issue with that. I'm open to either solution. But the files should be moved at once.

PR #1347, opened 2025-10-07
Redefine graph enrichment method

maxfirmbach:
## Description and Context

We might want to calculate things as we did in the paper. This calculates $G(A^{2^l})$ instead of $G(A^l)$.


reginabuehler:
After an offline discussion, we are closing this PR for now, but might open it again in the future.

PR #1348, opened 2025-10-07
Let shape functions take a span

amgebauer:
Instead of the vector type itself, the shape functions now take a span so that it can easier be used with other types, for example `std::vector`.


davidrudlstorfer (approved):
🚀

PR #1349, opened 2025-10-07
Enable point based input fields

amgebauer:
This PR suggests to enable point-based input fields. Requires #1348 (first commit)

Point-based input fields need to interpolate the quantity from the nodes of the mesh to a point in the domain. This is solved by passing the element id and the reference coordinates to an interpolate function. The way how to interpolate a quantity can be changed with a template parameter. The default is to simply use the shape functions.

A runtime-error is thrown if one tries to access a point-based field by the element id.

@sebproell wdyt?


src/core/io/src/4C_io_mesh.hpp
```diff
@@ -282,8 +282,15 @@ namespace Core::IO::MeshInput
         return raw_mesh_->external_ids ? (*raw_mesh_->external_ids)[index_] : invalid_external_id;
       }
 
+      /**
+       * Check if the point has data for the given field name.
+       */
+      [[nodiscard]] bool contains_data(const std::string& field_name) const
+      {
+        return raw_mesh_->point_data.contains(field_name);
+      }
```
sebproell: Having data is a property of the whole mesh and not just a point reference. We always assume that data is available for all points. I would move this function to the `Mesh` class.

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -517,6 +524,45 @@ namespace Core::IO::MeshInput
...
+   */
+  template <typename IndexType, typename T, unsigned dim>
+    requires(std::is_integral_v<IndexType>)
+  void read_value_from_point_data(
+      const Mesh<dim>& mesh, const std::string& key, std::unordered_map<IndexType, T>& value)
+  {
+    for (const auto& point_ref : mesh.points_with_data())
+    {
+      FOUR_C_ASSERT_ALWAYS(point_ref.contains_data(key),
+          "The mesh does not contain point data with the name '{}'.", key);
```
sebproell: Related to the other comment, this check only needs to happen once outside the loop and on the mesh.

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -273,6 +331,99 @@ namespace Core::IO
...
+     * template parameter of the interpolator. The optional @p field_name is only used for error
+     * messages.
+     *
+     * @note The @p xi coordinates are expected to be in the reference element of the element's
+     * cell type.
+     */
+    template <typename InterpolatorType = LinearInterpolator<T>>
+      requires InputFieldInterpolator<InterpolatorType, T, std::vector<double>,
+          decltype(std::declval<std::vector<T>>() | std::ranges::views::transform(std::identity{}))>
+    [[nodiscard]] T interpolate(IndexType element_id, std::span<const double> xi,
```
sebproell: Passing a `ConstElementRef` here would be better. You would not need to store the `Discretization*` yourself in this case. I guess it is not easily possible since we like to throw around the "naked" IDs as weak types. Maybe something more for the future.

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -316,6 +446,13 @@ namespace Core::IO
         }
         return it->second;
       }
+      if (const PointMapType* map = std::get_if<PointMapType>(&data_))
+      {
+        FOUR_C_THROW(
+            "Accessing a point-based InputField is only possible via the interpolate() function. "
+            "Either provide an element-based field or use the interpolate() function to obtain the "
+            "value.");
```
sebproell: Technically, this is correct. However, the error is instructing the user on two completely different levels of abstraction: either change the input _or_ change the implementation. The fact that this error exists, might indicate that we should separate the InputField types more strongly. We could have something like a cell-wise and an interpolated InputField. Then the types are strong enough to only allow one of the two accessors and remove the runtime error.

sebproell (commented):
I think it would be good to test this. It is now quite easy to create a dummy discretization with the test helper. I wonder if the Registry stuff works in a test, but finding out that it doesn't would probably reveal design problems.

I am not happy with the way we bundle up different interfaces into one class. I kind of saw it coming that this will happen and think it is time to come up with a proper design for the InputField, now that we know the different requirements :)

amgebauer:
> I think it would be good to test this. It is now quite easy to create a dummy discretization with the test helper. I wonder if the Registry stuff works in a test, but finding out that it doesn't would probably reveal design problems.

I added a unit test.

> I am not happy with the way we bundle up different interfaces into one class. I kind of saw it coming that this will happen and think it is time to come up with a proper design for the InputField, now that we know the different requirements :)

True, I'm also not super happy with this. Let's discuss this once you are back.

amgebauer:
@sebproell We cannot read `Tensors` as input, which would be required to allow (interpolated) tensor fields.

The context is that we can only read mesh data as scalar, `Vector`, `SymmetricTensor` or `Tensor` since many formats specify a certain ordering of the components (e.g., vtk with symmetric tensors). So, we cannot read a raw `std::array` of data.

Can I just add Tensor to the supported input types?

src/core/io/tests/4C_io_input_field_test.np2.cpp
```diff
@@ -90,4 +96,189 @@ namespace
...
+      {
+        EXPECT_NEAR(
+            input_field.interpolate(0, std::array{0.0, 0.0, 0.0}), expected_conductivity[0], 1e-12);
+        EXPECT_NEAR(
+            input_field.interpolate(1, std::array{0.0, 0.0, 0.0}), expected_conductivity[1], 1e-12);
+        EXPECT_NEAR(
+            input_field.interpolate(2, std::array{0.0, 0.0, 0.0}), expected_conductivity[2], 1e-12);
+      }
+      else
+      {
```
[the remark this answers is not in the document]
amgebauer: @sebproell and I don't understand why the redistribution of the mesh does not really work. All elements end up on the same processor. `dis.fill_from_mesh(...)` already distributes the discretization, right?

amgebauer:
Now ready for review (I will add tensor-based input in a separate PR since it is independent of this PR)

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -307,204 +529,232 @@ namespace Core::IO
...
-  }
+  }  // namespace Internal
 
 
+  /**
+   * @brief A class to represent an element-wise or constant input parameter field.
+   *
+   * In its current form, this class can either hold a single value of type T or a map of
+   * element-wise values of type T. It can be accessed via the `operator[]` or `at()` functions with
+   * global element ids.
```
sebproell:
```suggestion
   * global element ids. If you wish to interpolate the field with a cell, see InterpolatedInputField.
```

PR #1351, opened 2025-10-07
Unify functionality to convert a dof-vector into a node based multi-vector and move it to core

### LINKED ISSUE #1296: Move or use functionality to convert a dof-vector into a node-based multi-vector to or from core

lkoeglmeier:
To output additional gp data, we need to convert a dof-vector into a node-based multivector. The functionality has been implemented multiple times (poro and scatra) and has now been unified and moved to `Core::IO`.
Thanks @c-p-schmidt for your help!

Closes #1296


src/core/io/src/4C_io_discretization_visualization_writer_mesh.hpp
```diff
@@ -29,6 +29,24 @@ namespace Core::IO
     node      ///< The quantity to be written for output is node-based
   };
 
+  /// convert a dof based vector to a node based multi vector
+  /*!
```
c-p-schmidt, line 32: Can you please fix the docu?

c-p-schmidt:
Thank you for this cleanup. I only have one remark concerning the documentation.

src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
```diff
@@ -26,6 +26,36 @@ FOUR_C_NAMESPACE_OPEN
...
+  /*-----------------------------------------------------------------------------------------------*
+   *-----------------------------------------------------------------------------------------------*/
+  std::shared_ptr<Core::LinAlg::MultiVector<double>> convert_dof_vector_to_node_based_multi_vector(
+      const Core::FE::Discretization& dis, const Core::LinAlg::Vector<double>& input_vector,
+      const int number_of_dofset, const int number_of_dofs_per_node)
+  {
+    // initialize multi vector
+    std::shared_ptr<Core::LinAlg::MultiVector<double>> multi =
+        std::make_shared<Core::LinAlg::MultiVector<double>>(
+            *dis.node_row_map(), number_of_dofs_per_node, true);
```
amgebauer: It's better to return a unique_ptr instead of a shared_ptr


```suggestion
  std::unique_ptr<Core::LinAlg::MultiVector<double>> convert_dof_vector_to_node_based_multi_vector(
      const Core::FE::Discretization& dis, const Core::LinAlg::Vector<double>& input_vector,
      const int number_of_dofset, const int number_of_dofs_per_node)
  {
    // initialize multi vector
    auto multi =
        std::make_unique<Core::LinAlg::MultiVector<double>>(
            *dis.node_row_map(), number_of_dofs_per_node, true);
```

src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
```diff
@@ -26,6 +26,36 @@ FOUR_C_NAMESPACE_OPEN
...
+    const Core::LinAlg::Map& vectormap = input_vector.get_map();
+
+    // loop over nodes of the discretization
+    for (int inode = 0; inode < dis.num_my_row_nodes(); ++inode)
+    {
+      // get current node
+      Core::Nodes::Node* node = dis.l_row_node(inode);
+      // copy each dof value of node
+      for (int idof = 0; idof < number_of_dofs_per_node; ++idof)
+        (*multi)(idof).get_values()[inode] =
```
amgebauer:
```suggestion
        (*multi)(idof)[inode] =
```

amgebauer (dismissed):

src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
sebproell: I agree, but this was disabled in a recent PR. I think we should return a span of local values from the vector wrappers. @knarfnitram @maxfirmbach ?

knarfnitram: yes, so we started to remove element access operator (see https://github.com/4C-multiphysics/4C/pull/1170)

maxfirmbach: @sebproell Returning a span of local values should be fine. The direct element access is not ideal though :wink:.

lkoeglmeier: As suggested, I used the `std::span`. Thanks @amgebauer for the explanation.

PR #1353, opened 2025-10-08
Remove instantiation of `::NOX::Epetra::LinearSystemAztecOO`

vovannikov:
## Description and Context
It looks like that the default scenario for `FSI::Partitioned` is such that if a child class does not specify `create_linear_system()` directly, then the partitioning scheme does not need _this_ particular linear system object at all and we can return `null`.

Let's see if this assumption is correct and how many test break.


maxfirmbach:
Interesting ... would be nice if it works!

maxfirmbach:
Closes #1355.

maxfirmbach:
@vovannikov I took the freedom to set some labels and create an issue (which actually was already existing on our old gitlab instance of 4C) :wink:.

mayrmt (approved):

PR #1354, opened 2025-10-08
Remove references to `Aztec`

maxfirmbach:
## Description and Context

This PR removes all references to `Aztec`.


knarfnitram (approved):

maxfirmbach:
@mayrmt The input parameters should be renamed at some point. This is a breaking change ... so not sure when to do it.

PR #1356, opened 2025-10-08
Allow to lint multi vector

knarfnitram:
## Description and Context

Up to now the `MultiVector` class has not been formatted according to the 4C-Style guides.
Within this PR the class functions are renamed to match the clang-tidy format.


maxfirmbach:
Looking good! Thanks for putting the correct style into place.

src/core/linalg/src/sparse/4C_linalg_multi_vector.hpp
```diff
@@ -69,75 +64,75 @@ namespace Core::LinAlg
...
     //! Returns a view of the EpetraMap as type Map for this multi-vector.
     const Map& get_map() const { return map_.sync(vector_->Map()); };
 
     //! Returns the MPI_Comm for this multi-vector.
-    [[nodiscard]] MPI_Comm Comm() const;
+    [[nodiscard]] MPI_Comm get_comm() const;
 
     //! Returns true if this multi-vector is distributed global, i.e., not local replicated.
-    bool DistributedGlobal() const { return (vector_->Map().DistributedGlobal()); };
+    bool distributed_global() const { return (vector_->Map().DistributedGlobal()); };
```
mayrmt:
```suggestion
    bool is_distributed_global() const { return (vector_->Map().DistributedGlobal()); };
```

mayrmt (approved):

knarfnitram:
> Are there other wrapper classes, where this as to be done too?

I just searched for  `// NOLINTBEGIN` and maybe `4C_linalg_serialdensematrix.hpp` it can be adjusted. 
The remaining 4 occurrences are used for fortran and forward declarations.

sebproell (approved):

PR #1357, opened 2025-10-08
Update imcs cmake preset to enable debugging on cluster

m-frey:
## Description and Context

Debugging on the IMCS cluster is currently not working with CLion due to environment issues.
I added some required local paths to the base preset and added a dedicated debug preset.

### DIFF
--- a/presets/imcs/cluster/CMakePresets.json
+++ b/presets/imcs/cluster/CMakePresets.json
@@ -11,19 +11,7 @@
         "FOUR_C_ENABLE_NATIVE_OPTIMIZATIONS": "ON",
         "FOUR_C_WITH_ARBORX": "ON",
         "FOUR_C_WITH_FFTW": "ON",
-        "FOUR_C_WITH_GOOGLETEST": "OFF",
-        "FOUR_C_TRILINOS_ROOT": "/opt/trilinos/2025-03/",
-        "MPI_C_COMPILER": "/opt/openmpi/4.1.8/gcc/bin/mpicc",
-        "MPI_CXX_COMPILER": "/opt/openmpi/4.1.8/gcc/bin/mpicxx",
-        "MPIEXEC_EXECUTABLE": "/opt/openmpi/4.1.8/gcc/bin/orterun",
-        "FOUR_C_QHULL_ROOT": "/opt/spack/v0.23.1/opt/spack/linux-ubuntu24.04-sapphirerapids/gcc-13.3.0/qhull-2012.1-52nebni444m47tb7wxonn23bbbvlcapb",
-        "QHULL_INCLUDE_DIR": "/opt/spack/v0.23.1/opt/spack/linux-ubuntu24.04-sapphirerapids/gcc-13.3.0/qhull-2012.1-52nebni444m47tb7wxonn23bbbvlcapb/include",
-        "QHULL_LIBRARY": "/opt/spack/v0.23.1/opt/spack/linux-ubuntu24.04-sapphirerapids/gcc-13.3.0/qhull-2012.1-52nebni444m47tb7wxonn23bbbvlcapb/lib/libqhull.so"
-      },
-      "environment": {
-        "PATH": "/opt/trlinos/2025-03/bin:/opt/openmpi/4.1.8/gcc/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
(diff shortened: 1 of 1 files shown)


presets/imcs/cluster/CMakePresets.json
```diff
@@ -11,7 +11,19 @@
         "FOUR_C_ENABLE_NATIVE_OPTIMIZATIONS": "ON",
         "FOUR_C_WITH_ARBORX": "ON",
         "FOUR_C_WITH_FFTW": "ON",
-        "FOUR_C_WITH_GOOGLETEST": "OFF"
+        "FOUR_C_WITH_GOOGLETEST": "OFF",
+        "FOUR_C_TRILINOS_ROOT": "/opt/trilinos/2025-03/",
```
maxfirmbach: Is it really necessary to hard code all these paths?

maxfirmbach: Is there no other way to fix things?

presets/imcs/cluster/CMakePresets.json
```diff
@@ -11,7 +11,19 @@
...
+        "MPI_C_COMPILER": "/opt/openmpi/4.1.8/gcc/bin/mpicc",
+        "MPI_CXX_COMPILER": "/opt/openmpi/4.1.8/gcc/bin/mpicxx",
+        "MPIEXEC_EXECUTABLE": "/opt/openmpi/4.1.8/gcc/bin/orterun",
+        "FOUR_C_QHULL_ROOT": "/opt/spack/v0.23.1/opt/spack/linux-ubuntu24.04-sapphirerapids/gcc-13.3.0/qhull-2012.1-52nebni444m47tb7wxonn23bbbvlcapb",
+        "QHULL_INCLUDE_DIR": "/opt/spack/v0.23.1/opt/spack/linux-ubuntu24.04-sapphirerapids/gcc-13.3.0/qhull-2012.1-52nebni444m47tb7wxonn23bbbvlcapb/include",
+        "QHULL_LIBRARY": "/opt/spack/v0.23.1/opt/spack/linux-ubuntu24.04-sapphirerapids/gcc-13.3.0/qhull-2012.1-52nebni444m47tb7wxonn23bbbvlcapb/lib/libqhull.so"
```
sebproell: If this is truly required, you either have a bug in your library setup, or 4C's configure_Qhull is not working as it should. Could you try removing these two lines and only specifying the root?

presets/imcs/cluster/CMakePresets.json
```diff
@@ -25,6 +37,19 @@
...
+    },
+    {
+      "name": "imcs_charon_debug",
+      "displayName": "Debug build for charon at IMCS",
+      "inherits": [
+        ".imcs_cluster"
+      ],
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "DEBUG",
+        "FOUR_C_ENABLE_ADDRESS_SANITIZER": "ON",
```
sebproell: This combination is likely going to run incredibly slowly. I think the ASAN docs strongly recommend always using some optimization.

How about:

```suggestion
        "CMAKE_BUILD_TYPE": "RELWITHDEBINFO",
        "FOUR_C_ENABLE_ADDRESS_SANITIZER": "ON",
```

knarfnitram:
Thanks for you messages @sebproell @maxfirmbach 
I'd call it a special Clion preset. In this shell no modules are loaded or paths are set. Therefore, we add them with the hard-coded paths over cmake. If anyone has an idea how to simple setup all environment flags in clion, we're happy to hear about that (we tried several options and integrating them over cmake directly worked out most robust).
@m-frey can you please seperate this from the general cluster set up?

knarfnitram:
@m-frey - I removed some unnecessary paths and split it into a separate preset `imcs_charon_debug_clion `.
Can you verify if it works for you? To launch applications or test cases out of cmake we currently have to set the 
`LD_LIBRARY_PATH=/opt/trilinos/2025-04/lib` within Clion.

presets/imcs/cluster/CMakePresets.json
m-frey: It was not required.

presets/imcs/cluster/CMakePresets.json
m-frey: Thanks for this hint!

m-frey:
@knarfnitram Thanks you, yes it works. From my side we can merge this.

Issue #93, opened 2024-11-30
Post-processing tests do not work with Paraview 5.13.1


ppraegla:
Copied from https://gitlab.lrz.de/baci/baci/-/issues/1371

The decision was that it is not worth the effort to fix the `comparison.py` script because the ensight output should be replace by runtime output anyway. I'm just documenting this here on Github in case someone stumbles across the same problem.

## Bug Description: What's going wrong?
The tests using `tests/post_processing_test/comparison.py` fail when using `pvpython` from `ParaView-5.13.1-MPI-Linux-Python3.10-x86_64`. Error:
```
2792:   File "/path/to/4C/tests/post_processing_test/comparison.py", line 53, in <module>
2792:     csv_reader1 = csv.reader(open(sys.argv[4] + "/xxx_par0.csv", "r"), delimiter=",")
2792: FileNotFoundError: [Errno 2] No such file or directory: '/path/to/4C-build/framework_test_output/poro_3D_hex8-p2/xxx_par0.csv'
```

I found that there are two things why this does not work:
1. Up-to Paraview Version 5.6.2. The CSV writer appends a `0` to the filename such that the subsequent csv reader has to read `xxx_par0.csv` instead of `xxx_par.csv`,
2. The script still contains python 2 syntax as `pvpython` 5.5.2 uses Python 2.7.14

## Expectations
- [ ] Fix the tests such that a newer Paraview version can be used. The version 5.5.2 from the tests is from 2018.

## Possible Solution
~~Two options~~

~~1. Update the Paraview version and adapt the script~~
~~2. Replace Paraview with pyvista (<- I'm working on that)~~
- [ ] Use or implement the runtime output for the tests
- [ ] Remove the tests

## Steps to Reproduce
- Run a post-processing test e.g. `ctest -VV -R poro_3D_hex8.dat_porofluid-p2-pp`

## Your Environment
- **Relevant repo SHA1s:** main: 33d612370e605057f911715d139a689f3b449b71
- **Relevant configure flags or configure script:**
- **Operating system and version:**
- **Compiler and TPL versions:**

ppraegla:
This is fixed in #1358 by replacing pvpython with pyvista.

PR #1358, opened 2025-10-08
Remove dependency of our legacy post-processing tests on pvpython

amgebauer:
instead, the comparison script is now based on pyvista.

Note, that previously the comparison script did not complain if a quantity was not found in the ensight file. Now, I removed all quantities that are not written anymore and the test now complains if a quantity is not written.

In a second step, we can remove paraview from our docker image.


amgebauer:
Closes #93 fyi @ischeider

davidrudlstorfer (approved):
Really nice!!

Let's not forget to also remove the entire `FOUR_C_PVPYTHON` stuff in cmake

amgebauer:
> Let's not forget to also remove the entire FOUR_C_PVPYTHON stuff in cmake

Should already be removed. The only thing left is paraview in the docker container which I will remove in a follow-up.

amgebauer:
> Should already be removed. 

Sorry, forgot to stage the changes. Now we have it removed

utilities/four_c_python/src/four_c_building_testing/post_processing_comparison.py
```diff
@@ -4,303 +4,143 @@
 # See the LICENSE.md file in the top-level for license information.
 #
 # SPDX-License-Identifier: LGPL-3.0-or-later
+"""
+Script for comparing result of the post_ensight filter with reference data provided in a csv file.
 
-# Script for comparing result of the parallel and serial post processing by the post_ensight filter
+Point-wise quantities are compared by first finding the closest point in the output data set
+using a KDTree and then comparing the values of the quantities at that point.
```
gilrrei, line 11: Why not use `find_closest_point` from pyvista?

amgebauer: Yes, would have been easier, but it will also find any closest point. I only want to have the closed point within a tolerance (and throw an error otherwise). Anyways, the script is legacy and can hopefully be removed in the not too long future.

PR #1359, opened 2025-10-08
[thermo] Remove `output_state()`

maxfirmbach:
## Description and Context

This PR removes the `ouput_state()` method from the thermo module. The relevant output can already be written during run-time. In addition the related input file parameter is deleted.

Related to #211

### DIFF
--- a/src/thermo/src/implicit/4C_thermo_timint.cpp
+++ b/src/thermo/src/implicit/4C_thermo_timint.cpp
@@ -456,12 +456,6 @@ void Thermo::TimInt::output_step(bool forced_writerestart)
     if ((writerestartevery_ and (step_ % writerestartevery_ == 0)) or
         step_ == Global::Problem::instance()->restart())
       return;
-    // if state already exists, add restart information
-    if (writeglobevery_ and (step_ % writeglobevery_ == 0))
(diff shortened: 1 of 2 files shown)


src/thermo/src/implicit/4C_thermo_timint.cpp
```diff
@@ -491,85 +492,43 @@ void Thermo::TimInt::output_step(bool forced_writerestart)
...
-  return;
-
-}  // output_state()
-
+}
 
 /*----------------------------------------------------------------------*
- | add restart information to OutputStatewrite restart      ghamm 10/13 |
  *----------------------------------------------------------------------*/
 void Thermo::TimInt::add_restart_to_output_state()
```
c-p-schmidt: You deleted the `output_state` method, so this method should not be necessary anymore, if the name is not misleading ;-)

maxfirmbach: The code construct calling this method is very confusing ... so I'm not sure xD

maxfirmbach: @c-p-schmidt I removed the method.

PR #1363, opened 2025-10-09
[xfluid] Refactor unstable test case

maxfirmbach:
## Description and Context

This PR tackles a `xfluid` test that behaved a bit weird, due to the iterative solver on the fluid field. Therefore the fluid field linear solver is changed to `SuperLU` and the test timeout increased.


maxfirmbach:
@knarfnitram Yes, to speed up the testing time.

PR #1364, opened 2025-10-09
Fix bug in beam-to-surface pair creation

isteinbrecher:
## Description

Up until now there was a but in the beam interaction framework that resulted in wrong beam-to-surface pairs being created for shell formulations. This bug only materialised when the coupled shell elements in the coupling set are not starting from GID 0. An existing test case was adapted to account for this.


tests/input_files/beam3r_herm2line3_static_beam_to_solid_surface_contact_mortar_line2_gap_variation_reference_segmentation_penalty_lin_quad_kl_shell.4C.yaml
```diff
@@ -242,10 +242,6 @@ STRUCTURE KNOTVECTORS:
             - 1.0
       ID: 1
 DSURF-NODE TOPOLOGY:
-  - NODE 1 DSURFACE 1
-  - NODE 2 DSURFACE 1
-  - NODE 3 DSURFACE 1
-  - NODE 4 DSURFACE 1
```
davidrudlstorfer, line 248: Can these just be deleted completely? Doesn't this change anything in the simulation setup?

isteinbrecher: Good question, by removing the nodes from this set, there are fewer possible contact surface elements for the beam-to-solid condition (which triggered the previous existing bug), but none of those elements are actually in contact, thus the result does not change.

PR #1366, opened 2025-10-09
Reduce file size of pngs in solid tutorial

amgebauer:
Reduce the file size of the pngs. Previously, the pngs had a resolution of about 6k pixels in width. Now, I reduce them to roughly 600.


amgebauer:
Our current style of the documentation limits the width to barely above 600px.

amgebauer:
@ischeider now it is 1000px in width

PR #1368, opened 2025-10-09
Create new reference documentation layout

ischeider:
## Description and Context

This PR introduces the new layout for the reference sections. The basic layout has been provided by @gilrrei (thanks!), now they are integrated into the documentation. The new style is very informative which unfortunately means that the reference documentation becomes very large and complex. Hence, the built-up of the page was very slow, and I split the reference documentation in different chapters. 

As of now we have:
- Cell types (was already available previously)
- Spatial discretization
- Material information
- Boundary conditions
- general parameters

The differentiation is done by simple string search, which is done in the routine `create_markdown_documentation` within https://github.com/ischeider/4C/blob/17ef391e65c36254bf9ed350a1bdf68bb1c23430/utilities/four_c_python/src/four_c_documentation/make_documentation.py . Particularly, please check whether we should use a different mechanism or split the chapters differently.

In addition, this PR removes large parts of `create_rtd`. Only the resources for the cell type definitions are still contained in this tool. If anybody knows how to get the information about node, line and face numbering of the different elements in a different way, we can dump the whole tool (but deleting this information is not the solution IMO).


doc/documentation/src/analysis_guide_templates/boundaryconditions.rst.j2
```diff
@@ -133,7 +133,7 @@ Of course, the applied Dirichlet boundary condition may depend on time and on th
 of the node. This is achieved by a function definition, after the keyword ``FUNCT``.
 The number of the function (for each component) can be specified in
 order to define a spatial or temporal dependence. The proper definition
-of functions is given in the :ref:`functions <functiondefinitions>` section.
+of functions is given in the :ref:`functions <SECfunctn>` section.
```
sebproell, line 136: Are links case insensitive? It is not consistent with the link above

ischeider:
One additional comment: 

A particular feature introduced here is the listing of all test files that contain a particular section (which by default is collapsed) - thanks again, @gilrrei .
However, some section are used so often that I think a listing is not useful (e.g., `SOLVER 1` or `MATERIALS`). Therefore, in `make_documentation.py` I introduced a constant `TOO_MANY_TESTS_TO_SHOW`, which is set to 200 by now. Any discussion whether this is meaningful is welcome.

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+# THE SOFTWARE.
+import pathlib
+import textwrap
+import re
+import time
+from dataclasses import dataclass, field
+
+from four_c_common_utils.io import load_yaml
+
+# from four_c_input import sort_by_section_names
```
sebproell:
```suggestion
```

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+    Vector,
+    NotSetString,
+)
+from four_c_metadata.not_set import check_if_set
+
+DESCRIPTION_MISSING = '<span style="color:grey">*no description yet*</span>'
+TOO_MANY_TESTS_TO_SHOW = 200
+
+
+# Global variables to store which sections belong to which chapter (and thus to which file)
```
sebproell: I don't understand the comment. Which global variables are you referring to here? What follows in the code is a class.

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+    reference_directory, tests_directory, fourc_metadata_yaml_path
+):
+    """
+    Args:
+        reference_directory: Directory to write the reference files to
+        tests_directory: Directory containing the tests (to check which sections are used in tests)
+        fourc_metadata_yaml_path: Path to the fourc_metadata.yaml file (containing the metadata for the 4C input file format)
+
+    Here we create markdown documentation for the 4C input file format (called from conf.py.in)
+    Note particularly how the reference sections are split into different files by the definition of the reference_documents variable.
```
sebproell: I find it very confusing that you document this detail here in the function description. I had to do a search for `reference_documents` to understand what you mean. I would leave out the details here and maybe move them next to the relevant code.

```suggestion
    Note particularly that the reference sections are split into different files.
```

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+            "condition_reference",
+            "condition_reference",
+            ["^DESIGN ", ".*CONDITIONS$"],
+        ),
+        ReferenceChapter(
+            "General parameters", "general_reference", "general_reference", [".*"]
+        ),
+    ]
+
+    # -- Additional scripts to generate more fancy output -------------------------------------------------
```
sebproell:
```suggestion
```
?

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+            case _:
+                raise ValueError(type(entry))
+
+        entries += (
+            "\n\n"
+            + indent * " "
+            + "- "
+            + string_entry
+            + "\n"
+            + textwrap.indent(description_entry, prefix=(indent + 2) * " ")
```
gilrrei, line 483: Could we add another new line here? Then it would be an 'open' list, this way we have more space between the bullet points here:
image

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+        tests_directory: Directory containing the tests (to check which sections are used in tests)
+        fourc_metadata_yaml_path: Path to the fourc_metadata.yaml file (containing the metadata for the 4C input file format)
+
+    Here we create markdown documentation for the 4C input file format (called from conf.py.in)
+    Note particularly how the reference sections are split into different files by the definition of the reference_documents variable.
+    This list of ReferenceChapter objects defines the different files and which sections go into which file.
+    The sections are matched by regular expressions in the pattern attribute of each ReferenceChapter object.
+    """
+
+    global section_in_tests, section_in_tutorials
```
gilrrei: Whenever possible avoiding global is good practice. I know they are convenient, but why not pass them as argument?

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+)
+from four_c_metadata.not_set import check_if_set
+
+DESCRIPTION_MISSING = '<span style="color:grey">*no description yet*</span>'
+TOO_MANY_TESTS_TO_SHOW = 200
+
+
+# Global variables to store which sections belong to which chapter (and thus to which file)
+@dataclass
+class ReferenceChapter:
```
gilrrei, line 64: We could add a property called `description`, which would be added in the file. In this case:
image
between `Boundary and constraint conditions` and `BEAM INTERACTION/BEAM TO BEAM CONTACT CONDITIONS`.  A bit more explanation might be useful. In the future, this could also come from the metadata file directly.

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+
+    if name in section_in_tests:
+        content = "\n - " + "\n - ".join(
+            [f"[{n.split('/')[-1]}]({n})" for n in sorted(section_in_tests[name])]
+        )
+
+        n_files = len(section_in_tests[name])
+
+        if n_files > TOO_MANY_TESTS_TO_SHOW:
+            string += f"\n Used in more than {TOO_MANY_TESTS_TO_SHOW} tests (in total {n_files}), not showing them here."
```
gilrrei: This is a good approach, but I think we could display the first 10 files or something similar. For me, as a user, it is really helpful to look at some files.

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+def get_sections_in_files(directory):
+    sections = {}
+    for file in directory.glob("*.4C.yaml"):
+        sections_names = [k.strip() for k in load_yaml(file).keys()]
+        for s in sections_names:
+            file_name = str(file).split("tests/")[-1]
+            file_name = (
+                f"https://github.com/4C-multiphysics/4C/blob/main/tests/{file_name}"
+            )
+            if s in sections:
```
gilrrei, line 549: The function sections need to be treated differently; this is really annoying, but due to the fact that the ID is in the section name `FUNCT27`. So the section name has to be `FUNCT<n>`

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,631 @@
...
+            ["^DESIGN ", ".*CONDITIONS$"],
+        ),
+        ReferenceChapter(
+            "General parameters", "general_reference", "general_reference", [".*"]
+        ),
+    ]
+
+    # -- Additional scripts to generate more fancy output -------------------------------------------------
+
+    obj = All_Of.from_4C_metadata(fourc_metadata["sections"])
```
gilrrei, line 629: We should add some description for the FUNCT section. Ideally, this should be done in 4C, but as a quick solution, this can be done here:
```python
for i, section in enumerate(obj):
    if section.name == "FUNCT<n>":
        obj.specs[i].description = "Add some description here"
```

Also, the `FUNCT<n>` section is not correctly displayed:
image

utilities/four_c_python/src/four_c_documentation/make_documentation.py
ischeider: Sorry, I had global variables before, then changed them to a dataclass

doc/documentation/src/analysis_guide_templates/boundaryconditions.rst.j2
ischeider: yes, they are.

utilities/four_c_python/src/four_c_documentation/make_documentation.py
ischeider: Okay, added variable: `TESTS_TO_SHOW_IF_TOO_MANY`

utilities/four_c_python/src/four_c_documentation/make_documentation.py
ischeider: Maybe we have to do some formatting together in a future PR

utilities/four_c_python/src/four_c_documentation/make_documentation.py
ischeider: No problem to enter it with the dataclass Reference_Documents, but currently we don't have such descriptions. Maybe we shift this to a future PR.

ischeider:
This commit contains all the reviewer comments (thanks @gilrrei @sebproell !) except a little formatting and the FUNCT stuff (don't know yet how to include this less hacky)

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,646 @@
...
+    for o in iterate(section):
+        if o.name == "ENTITY_TYPE":
+            return True
+
+    return False
+
+
+def get_sections_in_files(directory):
+    sections = {}
+    for file in directory.glob("**/*.4C.yaml"):
```
gilrrei, line 542:
```suggestion
def get_sections_in_files(directory, pattern="*/*.4C.yaml"):
    sections = {}
    for file in directory.glob(pattern):
```

Such that for the tutorials, you could use the pattern `*/*/*.4C.yaml`. (I hope)

utilities/four_c_python/src/four_c_documentation/make_documentation.py
gilrrei: I think the functions are important. Currently, the parameters of the FUNCT section have no description, so the only source of information is the tests/tutorials.

utilities/four_c_python/src/four_c_documentation/make_documentation.py
gilrrei: You have another solution which also works great :)

utilities/four_c_python/src/four_c_documentation/make_documentation.py
ischeider: I changed the FUNC<n> to FUNC[n], since<,> are treated differently in sphynx. I hope this is okay.

utilities/four_c_python/src/four_c_documentation/make_documentation.py
ischeider: I also gave a description, since FUNCT is treated in a special manner anyway ;-)

ischeider:
I also added the description for the reference chapters. Please have a look.

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -0,0 +1,660 @@
...
+            "discretization_reference",
+            [".* DOMAIN$", ".* ELEMENTS$", ".* GEOMETRY$", ".*KNOTVECTORS$"],
+            "This is the reference containing the sections related to the spatial discretization of the problem, namely finite elements. "
+            + "Note that the definition of elements or particles one by one through ``* ELEMENTS`` or ``PARTICLES`` is not yet contained due to an old input spec layout",
+        ),
+        ReferenceChapter(
+            "Material information",
+            "materials_reference",
+            "materials_reference",
+            ["MATERIALS", "CLONING MATERIAL MAP"],
```
sebproell:
```suggestion
            ["^MATERIALS$", "^CLONING MATERIAL MAP$"],
```
 to only match exactly these?

ischeider:
Resolved the merge conflict, and added the comment about MATERIALS by @sebproell above.

Issue #61, opened 2024-11-29
Replace Teuchos::ParameterList with better suited construct at gauss point level evaluations


amgebauer:
## Expectations

There should be an easy/efficient way to pass parameters and quantities from the element to the material.

Possible quantities are
* Gauß point, number of Gauß points, ...
* Timestep size, total simulation time, ...
* Temperature
* Scalars (@c-p-schmidt)
* Gauß-point fiber definition
* ... possibly more

The same "Container" should be used during setup, evaluate and update.

PR #1371, opened 2025-10-09
Add evaluation context to material interface

amgebauer:
This PR adds a struct with the evaluation context for providing the material with very basic information (time, timestep, xi, reference coordinates) without relying on the `ParameterList`.

We don't have basic information always readily available, so that the structs holds only pointers.

Part of #61

### DIFF
--- a/src/mat/4C_mat_muscle_combo.cpp
+++ b/src/mat/4C_mat_muscle_combo.cpp
@@ -20,6 +20,7 @@
 #include "4C_mat_muscle_utils.hpp"
 #include "4C_mat_par_bundle.hpp"
 #include "4C_mat_service.hpp"
+#include "4C_mat_so3_material.hpp"
 #include "4C_utils_enum.hpp"
 #include "4C_utils_exceptions.hpp"
 
@@ -217,13 +218,14 @@ void Mat::MuscleCombo::setup(int numgp, const Discret::Elements::Fibers& fibers,
 }
 
 void Mat::MuscleCombo::update(Core::LinAlg::Tensor<double, 3, 3> const& defgrd, int const gp,
-    const Teuchos::ParameterList& params, int const eleGID)
+    const Teuchos::ParameterList& params, const EvaluationContext& context, int const eleGID)
 {
 }
 
 void Mat::MuscleCombo::evaluate(const Core::LinAlg::Tensor<double, 3, 3>* defgrad,
     const Core::LinAlg::SymmetricTensor<double, 3, 3>& glstrain,
-    const Teuchos::ParameterList& params, Core::LinAlg::SymmetricTensor<double, 3, 3>& stress,
+    const Teuchos::ParameterList& params, const EvaluationContext& context,
+    Core::LinAlg::SymmetricTensor<double, 3, 3>& stress,
     Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3>& cmat, int gp, int eleGID)
 {
   const Core::LinAlg::Matrix<3, 3> defgrd_mat = Core::LinAlg::make_matrix_view(*defgrad);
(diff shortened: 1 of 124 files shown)


src/mat/4C_mat_so3_material.hpp
```diff
@@ -25,6 +25,25 @@ FOUR_C_NAMESPACE_OPEN
...
+  /**
+   * @brief Context information for material evaluation
+   */
+  struct EvaluationContext
+  {
+    /// Total time (nullptr if not available)
+    const double* total_time;
+
+    /// Timestep size (nullptr if not available)
+    const double* delta_time;
```
mayrmt: `delta_time` refers to a math notation on paper, but requires additional explanation. Why not be more descriptive in the name itself?
```suggestion
    const double* time_step_size;
```

src/mat/4C_mat_muscle_combo.cpp
```diff
@@ -384,13 +386,16 @@ void Mat::MuscleCombo::evaluate(const Core::LinAlg::Tensor<double, 3, 3>* defgra
...
 void Mat::MuscleCombo::evaluate_active_nominal_stress(const Teuchos::ParameterList& params,
-    const int eleGID, const double lambdaM, double& intPa, double& Pa, double& derivPa)
+    const EvaluationContext& context, const int eleGID, const double lambdaM, double& intPa,
+    double& Pa, double& derivPa)
 {
   // save current simulation time
-  double t_tot = get_or<double>(params, "total time", -1);
+  FOUR_C_ASSERT(context.total_time, "Time not given in evaluation context.");
+  double t_tot = *context.total_time;
   if (abs(t_tot + 1.0) < 1e-14) FOUR_C_THROW("No total time given for muscle Combo material!");
```
c-p-schmidt: This can't happen anymore, right? Instead, you would run into the assertion (if assertions are activated).

Same some lines below.

src/mat/4C_mat_muscle_giantesio.cpp
```diff
@@ -402,12 +403,14 @@ void Mat::MuscleGiantesio::evaluate(const Core::LinAlg::Tensor<double, 3, 3>* de
   const double omega0 = params_->omega0_;
 
   // save current simulation time
-  double currentTime = get_or<double>(params, "total time", -1);
+  FOUR_C_ASSERT(context.total_time, "Time not given in evaluation context.");
+  double currentTime = *context.total_time;
   if (std::abs(currentTime + 1.0) < 1e-14)
```
c-p-schmidt: Also here.

src/mixture/src/4C_mixture_constituent_remodelfiber_expl.cpp
```diff
@@ -137,14 +137,15 @@ void Mixture::MixtureConstituentRemodelFiberExpl::setup(
     const Teuchos::ParameterList& params, int eleGID)
 {
   Mixture::MixtureConstituent::setup(params, eleGID);
-  update_homeostatic_values(params, eleGID);
+  update_homeostatic_values(params, 0.0, eleGID);
```
c-p-schmidt, line 140: What does this magic 0.0 mean?

src/mixture/src/4C_mixture_constituent_remodelfiber_impl.cpp
```diff
@@ -127,19 +128,20 @@ void Mixture::MixtureConstituentRemodelFiberImpl::setup(
     const Teuchos::ParameterList& params, int eleGID)
 {
   Mixture::MixtureConstituent::setup(params, eleGID);
-  update_homeostatic_values(params, eleGID);
+  update_homeostatic_values(params, 0.0, eleGID);
```
c-p-schmidt, line 131: Same here.

src/mixture/src/4C_mixture_constituent_remodelfiber_expl.cpp
amgebauer: It is the setup function called at the beginning of the simulation (at t=0.0)

src/mat/4C_mat_so3_material.hpp
amgebauer: I renamed it to `time_step_size`

Issue #1121, opened 2025-08-18
[LinAlg] Prepare vector class for Tpetra


knarfnitram:
## Description

Sub issue from https://github.com/4C-multiphysics/4C/issues/875 to prepare the class `Core::LinAlg::Vector` for the switch. 
Here notes, steps and open problems can be found. 

Data access - Tpetra has no `[]` Operator and it core's concept relies on views: 
Therefore all occurrences of the `[] operator` should be substituted with the function called `get_values()` or `get_local_values` or `local_values_as_span` , since this functions can be substituted with an unmanaged view in first stage.
- [x] `Vector<int>` (see https://github.com/4C-multiphysics/4C/pull/1372)
- [ ] `Vector<double>` (see #1456)

Catch `Epetra` error in the wrapper itself #1190:
- [x] Catch errors with `CHECK_EPETRA_CALL` (see https://github.com/4C-multiphysics/4C/pull/1390/)

Bulk local/global setters/adders - Tpetra offers only scalar versions
- [ ] remove `replace_local_values` with `replace_local_value`
- [ ] remove `replace_global_values` with `replace_global_value` 
- [ ] remove `sum_into_global_values` with `sum_into_global_value` 

Simple vector functions (with probably a 1:1 substitution)
- Norms / dot / scale / update / abs / reciprocal / reciprocalMultiply
- Multiplication
- lengths and & maps
- Print

Obvious Functions related to Epetra objects:
- [ ] Epetra_BlockMap, Epetra_FEVector, Importer/ Exporter

maxfirmbach:
@knarfnitram Can we already mark some of the things as "done"?

knarfnitram:
> [@knarfnitram](https://github.com/knarfnitram) Can we already mark some of the things as "done"?

@maxfirmbach: once #1456  and #1464 are merged, I think we can mark the vector interface as temporarily ready.

maxfirmbach:
@knarfnitram @dharinib98 Can this be closed?

knarfnitram:
Nearly, during https://github.com/4C-multiphysics/4C/pull/1465 the implicit conversions popped up, so I think it would be good to remove them here as well just to make sure that there are no epetra calls hidden.

PR #1372, opened 2025-10-09
Remove element access operators for integer-based vector interface 

knarfnitram:
## Description and Context

Removes the element access operators from the class `class Vector<int>` and substitutes it either by `get_values()` or  `local_values_as_span()`.

Related to https://github.com/4C-multiphysics/4C/issues/1121

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_vector.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_vector.hpp
@@ -315,16 +315,13 @@ namespace Core::LinAlg
     //! returns the values (data) as span
     std::span<int> get_local_values()
     {
-      return {get_values(), static_cast<size_t>(local_length())};
+      return {vector_->Values(), static_cast<size_t>(local_length())};
     };
 
     std::span<const int> get_local_values() const
     {
       return {vector_->Values(), static_cast<size_t>(local_length())};
     };
-
-
-    int* get_values() { return vector_->Values(); };
 
     int local_length() const { return vector_->MyLength(); };
 
--- a/src/coupling/src/adapter/4C_coupling_adapter.cpp
+++ b/src/coupling/src/adapter/4C_coupling_adapter.cpp
@@ -372,7 +372,7 @@ void Coupling::Adapter::Coupling::finish_coupling(const Core::FE::Discretization
 
   std::shared_ptr<const Core::LinAlg::Map> permmasternodemap =
       std::make_shared<Core::LinAlg::Map>(-1, permmasternodevec->local_length(),
-          permmasternodevec->get_values(), 0, masterdis.get_comm());
+          permmasternodevec->get_local_values().data(), 0, masterdis.get_comm());
 
   if (not slavenodemap->point_same_as(*permmasternodemap))
     FOUR_C_THROW("slave and permuted master node maps do not match");
@@ -641,7 +641,7 @@ void Coupling::Adapter::Coupling::master_to_slave(
     const Core::LinAlg::Vector<int>& mv, Core::LinAlg::Vector<int>& sv) const
 {
   Core::LinAlg::Vector<int> perm(*permslavedofmap_);
-  std::ranges::copy(mv.get_local_values(), perm.get_values());
+  std::ranges::copy(mv.get_local_values(), perm.get_local_values().begin());
 
   const int err = sv.export_to(perm, *slaveexport_, Insert);
   if (err) FOUR_C_THROW("Export to slave distribution returned err={}", err);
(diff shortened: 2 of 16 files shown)


src/core/linalg/src/sparse/4C_linalg_vector.hpp
```diff
@@ -312,11 +312,19 @@ namespace Core::LinAlg
...
+    //! returns the values (data) as span
+    std::span<int> local_values_as_span()
+    {
+      return {get_values(), static_cast<size_t>(local_length())};
+    };
+
+    std::span<const int> local_values_as_span() const
+    {
+      return {vector_->Values(), static_cast<size_t>(local_length())};
+    };
```
mayrmt, line 324: Why do you want to encode the return type in the function name? Why not just `local_values()` or `get_local_values()`?

I see, that `get_values` does not return `std::span`, but `int*`. So there is a difference. Yet, we usually do not include the return type into the name.

src/core/linalg/src/sparse/4C_linalg_vector.hpp
```diff
@@ -312,11 +312,19 @@ namespace Core::LinAlg
...
 
-    const int& operator[](int index) const { return (*vector_)[index]; }
 
-    int* get_values() const { return vector_->Values(); };
+    int* get_values() { return vector_->Values(); };
```
sebproell:
```suggestion
```

Let's remove this. The span is superior and safer. If you really need the raw pointer, you can ask the span for `.data()`.

src/coupling/src/adapter/4C_coupling_adapter.cpp
```diff
@@ -641,7 +641,7 @@ void Coupling::Adapter::Coupling::master_to_slave(
     const Core::LinAlg::Vector<int>& mv, Core::LinAlg::Vector<int>& sv) const
 {
   Core::LinAlg::Vector<int> perm(*permslavedofmap_);
-  std::copy(mv.get_values(), mv.get_values() + (mv.local_length()), perm.get_values());
+  std::copy(mv.local_values_as_span().begin(), mv.local_values_as_span().end(), perm.get_values());
```
sebproell:
```suggestion
  std::ranges::copy(mv.local_values_as_span(), perm.get_values());
```

src/core/linalg/src/sparse/4C_linalg_vector.hpp
```diff
@@ -312,11 +312,19 @@ namespace Core::LinAlg
 
     int min_value();
 
-    int& operator[](int index) { return (*vector_)[index]; }
+    //! returns the values (data) as span
+    std::span<int> local_values_as_span()
```
sebproell:
```suggestion
    std::span<int> local_values()
```

I'm not sure about the naming, but it seems we could simply drop the `as_span` suffix.

src/core/linalg/src/sparse/4C_linalg_vector.hpp
knarfnitram: Since this class will be merged with `class Vector` for Tpetra, I have taken the same name as in `class Vector`.

src/core/linalg/src/sparse/4C_linalg_vector.hpp
knarfnitram: I removed `get_values`, but I had to add an additional constructor for the maps based on the span object.  Any objections regarding the additional constructor?

src/core/linalg/src/sparse/4C_linalg_vector.hpp
knarfnitram: I renamed it..

mayrmt: @knarfnitram While writing my first comment, I inserted `local` to distinguish from "global", since I somehow thought that the two methods are different. They are not. So, `local` would not be necessary. Sorry for the confusion!

src/core/linalg/src/sparse/4C_linalg_map.hpp
```diff
@@ -36,6 +36,9 @@ namespace Core::LinAlg
     Map(int NumGlobalElements, int NumMyElements, const int* MyGlobalElements, int IndexBase,
         const MPI_Comm& Comm);
 
+    Map(int NumGlobalElements, int NumMyElements, std::span<const int> MyGlobalElements,
+        int IndexBase, const MPI_Comm& Comm);
```
sebproell:
```suggestion
    Map(int NumGlobalElements, std::span<const int> MyGlobalElements,
        int IndexBase, const MPI_Comm& Comm);
```

src/coupling/src/adapter/4C_coupling_adapter.cpp
```diff
@@ -372,7 +372,7 @@ void Coupling::Adapter::Coupling::finish_coupling(const Core::FE::Discretization
 
   std::shared_ptr<const Core::LinAlg::Map> permmasternodemap =
       std::make_shared<Core::LinAlg::Map>(-1, permmasternodevec->local_length(),
-          permmasternodevec->get_values(), 0, masterdis.get_comm());
+          permmasternodevec->get_local_values().data(), 0, masterdis.get_comm());
```
knarfnitram, line 375: @sebproell 
When I added the new Constructor, I thought that we instantiate more `Map` objects from the span. 
However I've checked the code and this is not the case, so I think creating a new constructor for a single instantiation is not necessary and removed the constructor again.

PR #1374, opened 2025-10-09
Fix bug in beam-to-surface pair creation

isteinbrecher:
## Description

Up until now there was a but in the beam interaction framework that resulted in wrong beam-to-surface pairs being created for shell formulations. This bug only materialised when the coupled shell elements in the coupling set are not starting from GID 0. An existing test case was adapted to account for this.

For some reason the old PR #1364 did recognise additional commits. Thus closing the old one and opening this PR.

### DIFF
--- a/unittests/geometry_pair/4C_geometry_pair_line_to_surface_patch_averaged_normals_test.cpp
+++ b/unittests/geometry_pair/4C_geometry_pair_line_to_surface_patch_averaged_normals_test.cpp
@@ -75,8 +75,9 @@ namespace
     const double eps = 1e-12;
 
     // Fill the discretization object with the geometry.
+    std::vector<std::shared_ptr<Core::Elements::Element>> core_elements;
     std::unordered_map<int, std::shared_ptr<GeometryPair::FaceElement>> face_elements_map;
-    xtest_surface_patch_quad4<face_element_type>(*discret_, face_elements_map);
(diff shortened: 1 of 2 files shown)


tests/input_files/beam3r_herm2line3_static_beam_to_solid_surface_contact_mortar_line2_gap_variation_reference_segmentation_penalty_lin_quad_kl_shell.4C.yaml
```diff
@@ -242,10 +242,6 @@ STRUCTURE KNOTVECTORS:
             - 1.0
       ID: 1
 DSURF-NODE TOPOLOGY:
-  - NODE 1 DSURFACE 1
-  - NODE 2 DSURFACE 1
-  - NODE 3 DSURFACE 1
-  - NODE 4 DSURFACE 1
```
davidrudlstorfer, line 248: Same question as in the other PR; can these just be deleted without changing any simulation results?

davidrudlstorfer: Whoops - just saw you answered in the other PR

PR #1376, opened 2025-10-10
Add small geometry pair improvements

isteinbrecher:
## Description and Context

Add some small improvements to the geometry pair.

### DIFF
--- a/src/fbi/4C_fbi_input.cpp
+++ b/src/fbi/4C_fbi_input.cpp
@@ -71,7 +71,7 @@ std::vector<Core::IO::InputSpec> FBI::valid_parameters()
               .default_value = FBI::BeamToFluidMeshtingMortarShapefunctions::none}),
   };
   // Add the geometry pair input parameters.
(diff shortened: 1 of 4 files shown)


src/geometry_pair/4C_geometry_pair_element_faces.hpp
```diff
@@ -340,6 +316,11 @@ namespace GeometryPair
           connected_faces_(),
           evaluate_current_normals_(evaluate_current_normals)
     {
+      static_assert(Surface::n_val_ == 1,
+          "Patch face elements only work for if the nodal values are the direct displacements, not "
+          "if the nodal values contain additional tangent information (e.g. Hermitian elements).");
```
knarfnitram: How about:
"Patch face elements only work when nodal values are direct displacements, not when nodal values include tangent information (e.g., Hermitian elements)."

isteinbrecher: Thanks, that makes more sense. I uploaded the changes.

PR #1377, opened 2025-10-10
Beam-to-solid data structures

isteinbrecher:
## Description and Context

Small improvements to beam-to-solid data structures.

### DIFF
--- a/src/geometry_pair/4C_geometry_pair_evaluation_data_base.hpp
+++ b/src/geometry_pair/4C_geometry_pair_evaluation_data_base.hpp
@@ -11,8 +11,6 @@
 
 #include "4C_config.hpp"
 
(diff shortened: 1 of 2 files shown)


src/geometry_pair/4C_geometry_pair_evaluation_data_base.hpp
```diff
@@ -25,10 +23,8 @@ namespace GeometryPair
    public:
     /**
      * \brief Constructor.
-     *
-     * \param input_parameter_list (in) Parameter list with the geometry evaluation parameters.
      */
-    GeometryEvaluationDataBase(const Teuchos::ParameterList& input_parameter_list) {};
```
maxfirmbach, line 31: I guess it is worth adding another member variable to get rid of this parameter list input argument.

PR #1381, opened 2025-10-10
Add beam-to-solid edge contact

isteinbrecher:
## Description and Context

Add beam-to-solid edge contact. As the existing beam-to-solid contact is purely based on line-to-surface interaction, this can lead to issues if a beam is moved over the edge of a solid surface. This PR adds a formulation to account for this.

For now the beam-to-edge contact does not average the tangents on the solid edge, this will be added in a follow up PR.


src/beaminteraction/src/4C_beaminteraction_beam_to_solid_conditions.cpp
```diff
@@ -859,4 +865,55 @@ BeamInteraction::BeamToSolidConditionSurface::create_contact_pair_internal(
...
+  const bool beam_is_hermite = beam_element->hermite_centerline_interpolation();
+  const auto& core_element = other_line_map_[ele_ptrs[1]->id()];
+  const auto shape = core_element->shape();
+
+  if (beam_is_hermite and shape == Core::FE::CellType::line2)
+  {
+    return std::make_shared<BeamToSolidEdgeContactPair<t_hermite, t_line2>>(
+        beam_to_edge_parameters_, core_element);
+  }
+  FOUR_C_THROW("Got unexpected element input, could not create BeamToSolidEdgeContactPair");
```
davidrudlstorfer: Maybe extend the error here on why the input was unexpected?

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_pair.cpp
```diff
@@ -0,0 +1,226 @@
...
+#include "4C_linalg_sparsematrix.hpp"
+#include "4C_utils_exceptions.hpp"
+#include "4C_utils_fad.hpp"
+
+FOUR_C_NAMESPACE_OPEN
+
+
+/**
+ *
+ */
```
davidrudlstorfer, line 32: Can these be removed if they are empty? Or add the description?

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_pair.cpp
```diff
@@ -0,0 +1,226 @@
...
+  scalar_type eta_beam = 0.0;
+  scalar_type eta_edge = 0.0;
+  const auto projection_result =
+      GeometryPair::line_to_line_closest_point_projection(beam_pos, edge_pos, eta_beam, eta_edge);
+
+  if (projection_result != GeometryPair::ProjectionResult::projection_found_valid)
+  {
+    is_active_ = false;
+    return;
+  }
```
davidrudlstorfer, line 92: Based on the stuff I did over the last few months I am interested in this section :D 

So this determines the bilateral closest point projection? What happens if the "real" closest point projection does not lie on the two evaluated elements?

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_pair.hpp
```diff
@@ -0,0 +1,176 @@
...
+    };
+
+    /**
+     * \brief Get coordinates of all active contact points on element1. Not yet implemented.
+     */
+    void get_all_active_contact_point_coords_element1(
+        std::vector<Core::LinAlg::Matrix<3, 1, double>>& coords) const override
+    {
+      FOUR_C_THROW("get_all_active_contact_point_coords_element1 not yet implemented!");
+    }
```
davidrudlstorfer, line 117: Is this necessary for this PR or a future feature?

davidrudlstorfer: Same for the methods down below

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_params.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+FOUR_C_NAMESPACE_OPEN
+
+
+namespace BeamInteraction
+{
+  /// Parameters for beam-to-solid edge contact
+  struct BeamToSolidEdgeContactParameters
+  {
+    Inpar::BeamToSolid::BeamToSolidConstraintEnforcement constraint_enforcement;
```
davidrudlstorfer, line 25: This is one of my (hopefully) next projects. The Beam Inpar Stuff should finally move to the physics but we need to restructure the beaminteraction folder firstly. Otherwise the name clashes get out of hand

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_params.cpp
davidrudlstorfer, line 1: I think it would make sense to somehow reflect that this file contains the input parameters.

With the other files across the repo I added `input` in the file somewhere

tests/input_files/beam3r_herm2line3_static_beam_to_solid_edge_contact_penalty_minimal.4C.yaml
```diff
@@ -0,0 +1,166 @@
...
+  TRIAD_VISUALIZATIONPOINT: true
+  STRAINS_GAUSSPOINT: true
+  ELEMENT_GID: true
+IO/RUNTIME VTK OUTPUT/STRUCTURE:
+  OUTPUT_STRUCTURE: true
+  DISPLACEMENT: true
+  STRESS_STRAIN: true
+  ELEMENT_OWNER: true
+  ELEMENT_GID: true
+  ELEMENT_MAT_ID: true
```
davidrudlstorfer, line 62: This should probably be disabled if the vtk output is not tested, otherwise we create so much testing data again.

tests/input_files/beam3r_herm2line3_static_beam_to_solid_edge_contact_penalty_minimal.nox.xml
davidrudlstorfer, line 1: Could we reuse some other settings file for this case?

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_conditions.cpp
isteinbrecher: Good point, I added some more information to the error.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_pair.cpp
isteinbrecher: That is just my personal style to add spacing between functions in the `cpp` file. I only add the description for functions purely defined in the `cpp` file, everything that is related to the header has the documentation there.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_pair.cpp
isteinbrecher: Nice to hear that this also has some possible other applications.

> So this determines the bilateral closest point projection? What happens if the "real" closest point projection does not lie on the two evaluated elements?

Yes, this is a closest point projection between two curves. If the real cpp does not lie on the elements this pair will do nothing, since there will be another pair that has the real cpp and this one will evaluate the interaction terms.

There are two "but"s here:
- If the cpp is exactly at an element node: In this case two (or 4) pairs will have a valid cpp. For now this would lead to a double evaluation of the interaction terms. In beam-to-beam contact this is "solved" by throwing a warning. I think this case can also be solved by applying some logic to the GIDs of the interacting elements. Also there is always the option to add a sort of global tacker object to the `GeometryPairEvaluationData` that helps with this. For now I would leave this to a follow up PR (see also the next point)
- The more critical case is if there are kinks in the edges between elements (e.g., for edges of Lagrangian solid elements, or Lagrangian beam elements). In this case it is possible that there is no unique cpp. To avoid something like that, I will add an averaged tangent implementation, that averages tangents on nodes and then interpolates between them, in a follow up PR. This is commonly used for problems like these, we even have a fully working implementation for averaged normals on surfaces, that should be more or less directly reusable for tangents. I didn't add this with this PR to not overload this one.

Long story short, if you have Hermitian beam elements, you should already be able to fully use this function.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_pair.hpp
isteinbrecher: Unfortunately these are needed because these are abstract methods in the base class.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_params.hpp
isteinbrecher: I would very much appreciate that! For now I would leave this file name as it is, as it fits in with the naming convention in beam interaction, basically all files ending with `params` are input related files.

tests/input_files/beam3r_herm2line3_static_beam_to_solid_edge_contact_penalty_minimal.4C.yaml
isteinbrecher: Do we have issues with the amount of test data?

In general, I like to have these sections activated in test files, even if they are not tested, because then one can simply run the input file and look at the results. Otherwise, one always has to fiddle around to get the output working.

tests/input_files/beam3r_herm2line3_static_beam_to_solid_edge_contact_penalty_minimal.nox.xml
isteinbrecher: Good point, I added this directly into the input file.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_pair.cpp
davidrudlstorfer: Thanks for the nice explanation and outlook :)

tests/input_files/beam3r_herm2line3_static_beam_to_solid_edge_contact_penalty_minimal.4C.yaml
davidrudlstorfer: We had some problems some time ago.

You can also just comment it out?

isteinbrecher: Commenting out has the downside that when we modify the input file with some tool such as FourCIPP, the comments will likely be gone.

For now I would leave this as is, if we run into issues with the space again, I think the better way would be to automatically remove files after a test that are not needed for successive tests.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_edge_contact_pair.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+  is_active_ = true;
+
+  // Evaluate the positions on the curves (todo: maybe directly take them from the CPP)
+  Core::LinAlg::Matrix<3, 1, scalar_type> r_beam;
```
knarfnitram, line 97: do you want to merge it with that open todo?

isteinbrecher: Yes, I would do so at the moment. As suggested in this comment https://github.com/4C-multiphysics/4C/pull/1381#discussion_r2428503539 there will be quite some changes to the cpp function soon, and I will see if it makes sense to add these "return" values. That mainly depends on how the handling of FAD types will be done in the projection function going further.

PR #1382, opened 2025-10-10
Enable Tensor and Symmetric Tensor input from yaml/input spec

amgebauer:
I added a new helper struct `ProxyType<T>`, where you describe a type based on already existing types, and the conversion between them.

I use it to be able to read Tensors with nested arrays as underlying type.

Note, the first commit only renames a file (which I need to be able to generalize the contents). The second commits contains all changes.

 @sebproell fyi

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
@@ -4002,10 +4002,8 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
     known_materials[Core::Materials::mix_prestress_strategy_constant] =
         group("MIX_Prestress_Strategy_Constant",
             {
-                parameter<std::vector<double>>("PRESTRETCH",
-                    {.description =
-                            "Definition of the prestretch as a 6x1 vector [xx, yy, zz, xy, yz, xz]",
-                        .size = 6}),
+                parameter<Core::LinAlg::SymmetricTensor<double, 3, 3>>("PRESTRETCH",
+                    {.description = "Definition of the (symmetric) prestretch tensor"}),
             },
             {.description = "Simple predefined prestress"});
   }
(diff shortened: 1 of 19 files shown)


src/core/io/src/4C_io_yaml.hpp
```diff
@@ -10,6 +10,7 @@
 
 #include "4C_config.hpp"
 
+#include "4C_io_input_spec_proxy_types.hpp"
```
sebproell: Can we put the proxy types into `io_types.hpp`? The dependency looks wrong right now. (edit: outdated)

src/core/io/tests/4C_io_input_spec_test.cpp
```diff
@@ -1311,17 +1311,82 @@ c: [[1.0, 2.0], [[1.0, 2.0, 8.0], {a: true, b: false, c: true, d: false}]])",
     }
   }
 
-  TEST(InputSpecTest, MatchYamlVectorOfArraysOfVectorsOfArrays)
```
sebproell, line 1314: Please keep the old test.

src/core/linalg/src/dense/4C_linalg_tensor.hpp
```diff
@@ -372,6 +372,12 @@ namespace Core::LinAlg
     requires(is_tensor<decltype(tensor)> && sizeof...(new_order) == decltype(tensor)::rank() &&
              !is_compressed_tensor<decltype(tensor)>);
 
+  /*!
+   * @brief Print a pretty short name of the tensor
```
sebproell:
```suggestion
   * @brief Print a pretty and short name of the tensor
```

src/core/io/src/4C_io_input_types.hpp
```diff
@@ -10,6 +10,7 @@
 
 #include "4C_config.hpp"
 
+#include "4C_io_proxy_types.hpp"
```
sebproell, line 13: Maybe move the whole content into this header?

src/core/io/tests/4C_io_input_spec_test.cpp
amgebauer: Already fixed in the latest version

src/core/io/src/4C_io_input_types.hpp
amgebauer: I also have to specialize some internals (like `Internal::GeneralizedTypeT<T>`), so I cannot centralize all in the proxy-header.

PR #1383, opened 2025-10-14
Publish clang18 compile time report

amgebauer:
Generate a static html report of your clang18 nightly job and publish it on our Github pages.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,7 +2,7 @@ name: deploy
 
 on:
   workflow_run:
-    workflows: [documentation, coverage]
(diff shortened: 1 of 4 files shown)


amgebauer:
Generation works: https://github.com/amgebauer/4C/actions/runs/18496292463

I added a short section to our docs. I think we should not add it more prominent since the value for the normal user is limited.

amgebauer:
Note: The deploy job will fail after merge until the next nightly pipeline run. I will merge today in the evening.

amgebauer:
Now available here: https://4c-multiphysics.github.io/4C/clang18_compile_time_report.html

PR #1384, opened 2025-10-14
Remove dead code after update to documentation generation

sebproell:
Following #1368, this PR cleans the `create_rtd` target and removes more legacy dat-style printing from InputSpec.


sebproell:
> I am desperately sorry that I missed so much code to delete...

@ischeider It is not your fault! This is mostly internal stuff in `InputSpec`.

PR #1385, opened 2025-10-14
Fix typo in README.md

reginabuehler:
## Description and Context

As the title states.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ See our [documentation](https://4c-multiphysics.github.io/4C/documentation/insta
 how to build 4C from source on your machine.
 Consult our [Tutorials](https://4c-multiphysics.github.io/4C/documentation/tutorials/tutorials.html) to get an overview of the
 general workflow in 4C.
-Also have a look at related [projects/tools](https://4c-multiphysics.github.io/4C/documentation/tools/tools.html) that work with 4C, e.g., tools that help with input generatation or visualization.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,7 +63,7 @@ See our [documentation](https://4c-multiphysics.github.io/4C/documentation/insta
 how to build 4C from source on your machine.
 Consult our [Tutorials](https://4c-multiphysics.github.io/4C/documentation/tutorials/tutorials.html) to get an overview of the
 general workflow in 4C.
-Also have a look at related [projects/tools](https://4c-multiphysics.github.io/4C/documentation/tools/tools.html) that work with 4C, e.g., tools that help with input generatation or visualization.
+Also have a look at related [projects/tools](https://4c-multiphysics.github.io/4C/documentation/tools/tools.html) that work with 4C, e.g., tools that help with input generation or visualization.
```
amgebauer, line 66: @sebproell Does our typo-check in the commit hooks also work for md and rst files? ~~Currently, we have only enabled c++, python, cmake and shell.~~
EDIT: We have it also enabled for rst and markdown. Why aren't these typos found earlier?

sebproell: My understanding is:

`typos` tries to avoid false positives at all costs since they can  be very annoying. Thus, it uses a handwritten list of classical typos. Just because a word is not in a dictionary does not trigger it. We can always extend the list of common misspellings ourselves.

PR #1390, opened 2025-10-15
Utilize CHECK_EPETRA_CALL in LinAlg::Vector

knarfnitram:
## Description and Context

So far the `LinAlg::Vector` and `LinAlg::Vector<int>` interfaces forwarded error codes of operations directly. 
Since these error codes will not be forwarded in Tpetra the signature of functions should be refactored.

This MR adapts the interfaces to catch potential errors by changing the signature and using `CHECK_EPETRA_CALL`. Remaining error messages with corresponding `FOUR_C_THROW` are removed and all functions with `CHECK_EPETRA_CALL` are moved to the source file. 

See #1190 and #1121

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_vector.cpp
+++ b/src/core/linalg/src/sparse/4C_linalg_vector.cpp
@@ -9,10 +9,10 @@
 
 #include "4C_comm_mpi_utils.hpp"
 #include "4C_linalg_multi_vector.hpp"
+#include "4C_utils_epetra_exceptions.hpp"
 #include "4C_utils_exceptions.hpp"
 
 #include <Epetra_Vector.h>
-
 
 FOUR_C_NAMESPACE_OPEN
 
@@ -149,16 +149,16 @@ void Core::LinAlg::Vector<T>::scale(double ScalarA, const Epetra_MultiVector& A)
 }
 
 template <typename T>
-int Core::LinAlg::Vector<T>::update(double ScalarA, const Epetra_MultiVector& A, double ScalarThis)
+void Core::LinAlg::Vector<T>::update(double ScalarA, const Epetra_MultiVector& A, double ScalarThis)
 {
-  return vector_->Update(ScalarA, A, ScalarThis);
+  CHECK_EPETRA_CALL(vector_->Update(ScalarA, A, ScalarThis));
 }
 
 template <typename T>
-int Core::LinAlg::Vector<T>::update(double ScalarA, const Epetra_MultiVector& A, double ScalarB,
+void Core::LinAlg::Vector<T>::update(double ScalarA, const Epetra_MultiVector& A, double ScalarB,
     const Epetra_MultiVector& B, double ScalarThis)
 {
-  return vector_->Update(ScalarA, A, ScalarB, B, ScalarThis);
+  CHECK_EPETRA_CALL(vector_->Update(ScalarA, A, ScalarB, B, ScalarThis));
 }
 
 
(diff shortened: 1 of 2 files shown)


maxfirmbach (dismissed):
Great cleanup job! 👍  Not only helps this to make the transition to `Tpetra` smoother, but it also helps a lot to make the overall code cleaner. No more `int err` and respective `if()` calls all over the codebase.

src/core/linalg/src/sparse/4C_linalg_vector.hpp
```diff
@@ -177,64 +179,64 @@ namespace Core::LinAlg
      */
     int replace_map(const Map& map);
 
-    int replace_local_value(int MyRow, double ScalarValue)
+    void replace_local_value(int MyRow, double ScalarValue)
```
sebproell: These calls should probably move to the cpp file.  Does not need to be tackled here, but something to keep in mind.

sebproell (dismissed):

mayrmt (dismissed):

maxfirmbach (approved):

sebproell (approved):

maxfirmbach:
Seems like `Epetra` is throwing in quite a few places that where not covered before 😉.

src/core/linalg/src/sparse/4C_linalg_vector.cpp
```diff
@@ -183,7 +190,39 @@ void Core::LinAlg::Vector<T>::scale(double ScalarA, const Vector& A)
 template <typename T>
 void Core::LinAlg::Vector<T>::update(double ScalarA, const Vector& A, double ScalarThis)
 {
-  CHECK_EPETRA_CALL(vector_->Update(ScalarA, A, ScalarThis));
+  const auto& dstMap = vector_->Map();
+  const auto& srcMap = A.vector_->Map();
```
knarfnitram: I started to fix some maps, but there are still some failing test cases.  
Here I added a map check, which ensures that updates only happens on matching maps - not quite sure if we really should merge this, but it visualizes good what is going wrong.

mayrmt: @knarfnitram thanks! I think this is a good tool to identify and debug current problems. In the long run, I am not a fan of automatically switching maps internally without the knowledge of the user. I'd rather force the user to supply valid maps for both vectors.

maxfirmbach:
@knarfnitram Let's merge some of the fixes separately and independently before this PR and rebase onto them.

src/core/linalg/src/sparse/4C_linalg_vector.cpp
maxfirmbach: I think for debugging it is ok, but I think we should not bloat the wrapper methods and finally merge this. The user has to make sure the correct thing goes into this method, otherwise it throws, plain and simple.

src/adapter/4C_adapter_fld_fluid_fsi.cpp
```diff
@@ -607,11 +607,27 @@ void Adapter::FluidFSI::adams_bashforth2(const Core::LinAlg::Vector<double>& vel
...
+    Core::LinAlg::Vector<double> accnm_on_vel(velnp.get_map());
+    Core::LinAlg::export_to(accn, accn_on_vel);
+    Core::LinAlg::export_to(accnm, accnm_on_vel);
+
+    velnp.update(a, accn_on_vel, b, accnm_on_vel, c);
+  }
+  else
+  {
+    velnp.update(a, accn, b, accnm, c);
+  }
```
mayrmt: @knarfnitram do you know, if we ever hit the `else` case?

src/core/fem/src/condition/4C_fem_condition_locsys.cpp
```diff
@@ -100,6 +100,19 @@ void Core::Conditions::LocsysManager::update(const double time,
...
+
+  if (!locsystoggle_ || !locsystoggle_->get_map().same_as(node_row_map))
+  {
+    // (Re)create on the correct map
+    locsystoggle_ = Core::LinAlg::create_vector(node_row_map, /*zero*/ true);
+  }
+
+  // Optional: sanity print once
+  FOUR_C_ASSERT(
+      locsystoggle_->get_map().same_as(node_row_map), "locsystoggle_ must use node row map");
```
mayrmt: Is this still relevant? Doesn't the content of the `if` make sure, that this error will never be triggered?

src/core/linalg/src/sparse/4C_linalg_equilibrate.cpp
```diff
@@ -336,6 +333,9 @@ void Core::LinAlg::EquilibrationBlock::equilibrate_matrix(
       // compute inverse column sums of current column block of global system matrix
       else
       {
+        Core::LinAlg::Vector<double> invcolsums_overlap(blocksparsematrix->matrix(j, j).col_map());
+        invcolsums_overlap.put_scalar(0.0);
```
mayrmt:
```suggestion
        Core::LinAlg::Vector<double> invcolsums_overlap(blocksparsematrix->matrix(j, j).col_map(), true);
```

src/core/linalg/src/sparse/4C_linalg_vector.cpp
```diff
@@ -203,6 +210,115 @@ int Core::LinAlg::Vector<T>::replace_map(const Map& map)
...
+}
+
+template <typename T>
+void Core::LinAlg::Vector<T>::sum_into_global_values(
+    int NumEntries, const double* Values, const int* Indices)
+{
+  CHECK_EPETRA_CALL(vector_->SumIntoGlobalValues(NumEntries, Values, Indices));
+}
+
+//! Matrix-Matrix multiplication, \e this = ScalarThis*\e this + ScalarAB*A*B.
```
mayrmt: Looks like a Doxygen comment, that should be placed into the header file at the declaration.

(See also other methods below.)

src/scatra/4C_scatra_timint_genalpha.cpp
```diff
@@ -312,12 +312,45 @@ void ScaTra::TimIntGenAlpha::compute_time_derivative()
...
+    phinp_owned->import(*phinp_, *importer_phinp_, Insert);
+    phinp_ref = phinp_owned.get();
+  }
+  if (!phin_->get_map().same_as(tgtMap))
+  {
+    importer_phin_ = std::make_unique<Core::LinAlg::Import>(tgtMap, phin_->get_map());
+    phin_owned->put_scalar(0.0);
+    phin_owned->import(*phin_, *importer_phin_, Insert);
+    phin_ref = phin_owned.get();
+  }
```
mayrmt, line 343: Are there cases, where these `if`s are not executed?

src/core/linalg/src/sparse/4C_linalg_equilibrate.cpp
```diff
@@ -356,12 +360,17 @@ void Core::LinAlg::EquilibrationBlock::equilibrate_matrix(
 
               // add entries of current matrix row to column sums
               for (int ientry = 0; ientry < numentries; ++ientry)
-                invcolsums->sum_into_global_value(
-                    matrix.col_map().gid(indices[ientry]), std::abs(values[ientry]));
```
maxfirmbach: The vector `invcolsums` is constructed by using `blocksparsematrix->matrix(j, j).domain_map()`, but here we check for the `col_map().gid(...)`. Might it be sufficient to just swap this by `domain_map().gid(...)`?

knarfnitram:
thanks for the suggestions @maxfirmbach  and @mayrmt . 
I'll start to clean up the commits and merge them in different requests to make sure everything works as intended.
These are the remaining failing tests:
- [x] restart for fails xfsi_comp_struct_fsi_2D_mono_solidsided
- [x] fsi_dc_mono_fs_ada_ab2_ab2 -> `tolerances will be adapted`
- [x] fsi_dc_mono_mtrss_ada_ab2_none -> `tolerances will  be adapted`
- [x] elch_3D_tet4_s2i_butlervolmer_mortar_standard_BGS-AMG_3x3   -> `tolerances will be adapted`

knarfnitram:
I subdivided the changes into #1416 ,#1417,#1418 and #1419 - once these are merged and I have dropped the commits we can proceed here.
I'll port/address the review comments once the pipelines have passed.

src/core/fem/src/condition/4C_fem_condition_locsys.cpp
knarfnitram: should be resolved in https://github.com/4C-multiphysics/4C/pull/1417/files .

src/scatra/4C_scatra_timint_genalpha.cpp
knarfnitram: I think so. see https://github.com/4C-multiphysics/4C/pull/1419/ for updated version.

src/core/linalg/src/sparse/4C_linalg_equilibrate.cpp
knarfnitram: see https://github.com/4C-multiphysics/4C/pull/1419/ for updated version.

mayrmt (approved):

PR #1393, opened 2025-10-15
ArborX: Add a wrapper for ArborX bvh

isteinbrecher:
## Description and Context

Add a wrapper to the ArborX bounding volume hierarchy class, so we reuse exiting BVHs for multiple queries. Up until now, we always created the BHV and only performed a single query.


isteinbrecher:
> Great 👍 Is there a place in the code base where we actually could make use of this feature already?

At the moment not. But for the RVE search for @m-frey this will likely be useful and I have a problem with the creation of 2D-3D beam-to-solid pairs where this will also help a lot (hopefully).

mayrmt (approved):

PR #1395, opened 2025-10-15
Port beam-to-solid test to new framework

isteinbrecher:
## Description and Context

Port one test case that uses solid-to-solid, beam-to-beam and beam-to-solid contact from the old to the new beam interaction framework.

Since the underlying beam-to-solid contact formulation is different, the test results had to be adapted.

#1175


maxfirmbach:
@isteinbrecher What about the changes to `BEAMS_NEWGAP: false`? Wasn't that the feature we wanted to port?

isteinbrecher:
> @isteinbrecher What about the changes to `BEAMS_NEWGAP: false`? Wasn't that the feature we wanted to port?

Yes, we need to port this. For this example, the new gap feature was activated but not used, therefore, we can deactivate it here.

tests/input_files/beam3eb_static_contact_beamtosolidcontact_solidcontact.4C.yaml
```diff
@@ -28,17 +28,77 @@ CONTACT DYNAMIC:
...
+  STRESS_STRAIN: false
+  ELEMENT_OWNER: true
+  ELEMENT_GID: true
+  ELEMENT_MAT_ID: true
 BEAM CONTACT:
   BEAMS_STRATEGY: "Penalty"
-  BEAMS_NEWGAP: true
-  BEAMS_BTSOL: true
+  MODELEVALUATOR: "Standard"
+  BEAMS_NEWGAP: false
```
davidrudlstorfer: Should we maybe directly remove this variable if it is deactivated? I guess this is deactivated out of the box?

mayrmt:
Is the "new gap" feature still tested in any other test?

tests/input_files/beam3eb_static_contact_beamtosolidcontact_solidcontact.4C.yaml
isteinbrecher: Good point, I removed it.

isteinbrecher:
> Is the "new gap" feature still tested in any other test?

I would assume so, although I have never actually checked that. In any case, the "new gap" did not have an impact on this original test, so we can for sure remove it here.

PR #1396, opened 2025-10-15
Allow setting fiber orientation with mesh data in the mixture module

amgebauer:
This PR allows to set the fiber orientation in the G&R-framework from an input field (instead of the `FIBERx`-input). This allows us to directly use the fiber definition in vtu-files.

@tuchpaul You can have a look at the added vtu files how we defined the fibers.

Draft for now since the framework also has a legacy test that we might remove before we can proceed. And we might need some internal optimization so that we can already pre-compute the fibers at the Gauss-points during the setup-phase.


amgebauer:
> Does this mean we can get rid of the FNODEs (and the FiberNode)? If so, we should delete these parts from rtd and the code.

Yes, this will replace the FNODEs in the future. Currently it requires VTU since we cannot read point-based input fields using the dat-style mesh input. We have some cardiovascular tests (outside of the mixture module) left that still rely on the FNODEs, but once we transformed these to vtu input, we can remove FNODEs from the code and the docs.

> I assume we could do this for element-wise defined fibers as well, right?

Yes, element-wise defined fibers can also be used. They can even be used with the dat-style input with the json-style input input.

PR #1398, opened 2025-10-16
Add interface to write control files

sebproell:
Construction of the control file (containing info about 4C's results) is essentially impossible to understand. This PR introduces an interface for writing to this file. I believe we can and should replace the control file format with YAML or JSON to make our lives even easier (we have a hand-written parser in C for this file type :confused: ), but this is not yet part of this PR. 

@c-p-schmidt @isteinbrecher Correct me if I am wrong, but I think the control file will survive the deletion of the old output, since it contains general data and also info about restarts written to HDF5.

### DIFF
--- a/src/core/io/src/4C_io.cpp
+++ b/src/core/io/src/4C_io.cpp
@@ -664,52 +664,6 @@ void Core::IO::DiscretizationWriter::create_result_file(const int step)
 
 /*----------------------------------------------------------------------*/
 /*----------------------------------------------------------------------*/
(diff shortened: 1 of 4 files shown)


isteinbrecher:
> @c-p-schmidt @isteinbrecher Correct me if I am wrong, but I think the control file will survive the deletion of the old output, since it contains general data and also info about restarts written to HDF5.

Exactly, the binary output (control file and hdf5 file) will continue to contain restart information.

sebproell:
Ready for review.

PR #1399, opened 2025-10-16
Introduce `NOX::Nln::Vector` to replace `::NOX::Epetra::Vector`

vovannikov:
## Description and Context
This is a preliminary PR on the way to completely remove `::NOX:Epetra::Vector`. Here I have created a new class `NOX::Nln::Vector` by inheriting it from the latter. In the next PRs this will be changed and `NOX::Nln::Vector` is planned to implement directly `NOX::Abstract::Vector` interface.

#1292, #1077, #1098, #359

### DIFF
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.cpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.cpp
@@ -1,0 +1,66 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#include "4C_solver_nonlin_nox_vector.hpp"
+
+#include "4C_utils_exceptions.hpp"
+
+#include <Epetra_Vector.h>
+
+FOUR_C_NAMESPACE_OPEN
+
+NOX::Nln::Vector::Vector(
+    const Teuchos::RCP<Epetra_Vector>& source, MemoryType memoryType, ::NOX::CopyType type)
+    : ::NOX::Epetra::Vector(source,
+          memoryType == MemoryType::View ? ::NOX::Epetra::Vector::CreateView
+                                         : ::NOX::Epetra::Vector::CreateCopy,
+          type)
+{
+}
+
+NOX::Nln::Vector::Vector(const Epetra_Vector& source, ::NOX::CopyType type)
+    : ::NOX::Epetra::Vector(source, type)
+{
+}
+
+NOX::Nln::Vector::Vector(const NOX::Nln::Vector& source, ::NOX::CopyType type)
+    : ::NOX::Epetra::Vector(source, type)
+{
+}
+
+::NOX::Abstract::Vector& NOX::Nln::Vector::operator=(const Epetra_Vector& source)
+{
+  epetraVec->Scale(1.0, source);
+  return *this;
+}
+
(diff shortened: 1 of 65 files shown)


src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
```diff
@@ -0,0 +1,58 @@
...
+      // Overloaded copy assignment operators
+      ::NOX::Abstract::Vector& operator=(const Epetra_Vector& y) override;
+      ::NOX::Abstract::Vector& operator=(const ::NOX::Abstract::Vector& y) override;
+
+      // This operator is ingerited and should not be ever called
+      ::NOX::Abstract::Vector& operator=(const ::NOX::Epetra::Vector& y) override;
+
+      // Clone method
+      Teuchos::RCP<::NOX::Abstract::Vector> clone(
+          ::NOX::CopyType type = ::NOX::DeepCopy) const override;
```
mayrmt:
```suggestion
      //! Reproduce ctors of ::NOX::Epetra::Vector
      Vector(const Teuchos::RCP<Epetra_Vector>& source, MemoryType memoryType = MemoryType::Copy,
          ::NOX::CopyType type = ::NOX::DeepCopy);

      Vector(const Epetra_Vector& source, ::NOX::CopyType type = ::NOX::DeepCopy);

      Vector(const NOX::Nln::Vector& source, ::NOX::CopyType type = ::NOX::DeepCopy);

      //! Main copy assignment operator
      ::NOX::Abstract::Vector& operator=(const NOX::Nln::Vector& y);

      //! Overloaded copy assignment operators
      ::NOX::Abstract::Vector& operator=(const Epetra_Vector& y) override;
      ::NOX::Abstract::Vector& operator=(const ::NOX::Abstract::Vector& y) override;

      //! This operator is ingerited and should not be ever called
      ::NOX::Abstract::Vector& operator=(const ::NOX::Epetra::Vector& y) override;

      //! Clone method
      Teuchos::RCP<::NOX::Abstract::Vector> clone(
          ::NOX::CopyType type = ::NOX::DeepCopy) const override;
```

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+::NOX::Abstract::Vector& NOX::Nln::Vector::operator=(const ::NOX::Abstract::Vector& source)
+{
+  return operator=(dynamic_cast<const NOX::Nln::Vector&>(source));
+}
+
+::NOX::Abstract::Vector& NOX::Nln::Vector::operator=(const ::NOX::Epetra::Vector& source)
+{
+  (void)source;  // To avoid unused parameter warning
```
mayrmt: What about using `[[maybe_unused]]` instead?

sebproell (approved):

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.cpp
vovannikov: I would rather prefer to use an unnamed parameter (as I have made in the updated version of the commit) since the semantics of `[[maybe_unused]]` I assume implies its usage for situations when something indeed _maybe_ unused whereas here the parameter is _definitely_ unused. Moreover, this ctor with get removed in an upcoming PR: this is just a temporary thing to ensure that it is indeed not called everywhere in the code (just to confirm that I understood its usage correctly).

mayrmt (approved):

PR #1403, opened 2025-10-17
Fix dependency of beam test that requires arborx

c-p-schmidt:
## Description and Context
The nightly pipeline is failing since one beam test seems to require ArborX now. See https://github.com/4C-multiphysics/4C/actions/runs/18576258948/job/52966972382#step:7:172.

I updated the test config accordingly.


isteinbrecher (approved):
Sorry, that was my mistake, thanks @c-p-schmidt for fixing this!

PR #1407, opened 2025-10-17
Write control files in YAML format

sebproell:
Continuing from #1398, this PR replaces the control file with the standard YAML format. 

- Switch the writer (#1398) to emit YAML instead of the old format
- Write metadata (creation time, location, etc) into an actual node rather than  a comment
- Switch the control file parser to parse the file via ryml

Looks like this now:

```
- metadata:
    created_by: "proell"
    host: "<snip>"
    time: "17-10-2025 13-39-50"
    sha: "09925d50d91ff6334cfbb9079db2b31a70527185"
    version: "2025.4.0-dev"

- general:
    input_file: "<snip>/4C/tests/input_files/constr3D_MPC_direct.4C.yaml"
    problem_type: "Structure"
    spatial_approximation: "Polynomial"
    ndim: 3

- field:
    field: "structure"
    time: 0
    step: 0
    num_nd: 243
    max_nodeid: 242
    num_ele: 128
    num_dof: 729
    num_dim: 3
    num_output_proc: 2
    mesh_file: "xxx.mesh.structure.s0"
```

Note that we need the top-level sequence (indicated by the dashes), since there are multiple entries of the same name. This could be changed but would require a change in the data layout. I just ported what we had in the old file as a first step.

### DIFF
--- a/src/core/io/src/4C_io_control.hpp
+++ b/src/core/io/src/4C_io_control.hpp
@@ -186,6 +186,11 @@ namespace Core::IO
     InputControl(const std::string& filename, MPI_Comm comm);
     ~InputControl();
 
+    InputControl(const InputControl&) = delete;
+    InputControl& operator=(const InputControl&) = delete;
+    InputControl(InputControl&&) = delete;
+    InputControl& operator=(InputControl&&) = delete;
+
     MAP* control_file() { return &table_; }
 
     std::string file_name() const { return filename_; }
@@ -200,9 +205,6 @@ namespace Core::IO
         const char* filestring, MAP*& result_info, MAP*& file_info);
 
    private:
-    InputControl(const InputControl&);
-    InputControl& operator=(const InputControl&);
-
     std::string filename_;
     MAP table_;
   };
--- a/src/core/io/src/4C_io_control.cpp
+++ b/src/core/io/src/4C_io_control.cpp
@@ -335,7 +335,7 @@ Core::IO::InputControl::InputControl(const std::string& filename, const bool ser
   if (!serial)
     parse_control_file(&table_, name.str().c_str(), MPI_COMM_WORLD);
   else
-    parse_control_file_serial(&table_, name.str().c_str());
+    parse_control_file(&table_, name.str().c_str(), MPI_COMM_NULL);
 }
 
 /*----------------------------------------------------------------------*
--- a/src/core/io/src/legacy/4C_io_legacy_table.cpp
+++ b/src/core/io/src/legacy/4C_io_legacy_table.cpp
@@ -7,6 +7,7 @@
 
 #include "4C_io_legacy_table.hpp"
 
+#include "4C_comm_mpi_utils.hpp"
 #include "4C_io_legacy_types.hpp"
 #include "4C_io_yaml.hpp"
 #include "4C_utils_enum.hpp"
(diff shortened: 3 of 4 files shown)


sebproell:
This is a breaking change. Do we know any tool that depends on the control file? It is now easy to read this file with a YAML parser, making it straightforward to replace.

davidrudlstorfer:
Just out of curiosity, is there a way to restart from old simulation results, i.e., an old control file? Or some conversion from an old to a new control file?

sebproell:
@davidrudlstorfer I think what you are asking is whether it is possible to restart a simulation run from an older 4C build with a 4C build after this PR gets merged. The answer is no. Note that this is, in general, not possible since we do not provide any kind of guarantees in this direction. Nothing changes with this PR. 4C writes out internal data and reads it in again, and this ABI is not stable. For the same reason, it is also not generally possible or supported to post-process (with the old post-processor!) results that were written with an older version. We could specify guarantees for this problem, but I don't think we should. ABI-stability is hard and we should have a compelling use case to support something like this.

isteinbrecher:
> Do we know any tool that depends on the control file?

I am not aware of any tool that do.

@sebproell thanks for the nice explanation, and I agree that we should not aim for supporting restarts across different 4C versions.

sebproell:
Newest commit is probably not the final solution but I want to see if it at least let's all tests pass. Our excessive and uncalled-for use of smart pointers once more leads to cyclic references (see #413). This makes the idiomatic RAII cleanup of the control file impossible. Some results might not be flushed to the file.

sebproell:
This now works _but_ I need a pretty ugly hack at the end of the main function. I will try to fix a few reference cycles so that our output writer gets properly cleaned up.

sebproell:
Now ready. I can't break it down further since I need to switch out the reading and writing of the control file all at once. I hope the commits help to better grasp the changes I made.

src/core/io/tests/4C_io_control_test.cpp
```diff
@@ -0,0 +1,147 @@
...
+  {
+    std::stringstream stream;
+    {
+      Core::IO::ControlFileWriter writer(true, stream);
+      writer.start_group("group1")
+          .write("string", "value")
+          .start_group("group2")
+          .write("int", 1)
+          .start_group("group3")
+          .write("double", 2.0);
```
c-p-schmidt, line 107: You might want to repeat the comment from the previous method also here.
```suggestion
          .write("double", 2.0);
          
      // Writer goes out of scope here and should flush the unclosed groups.
```

src/core/io/src/4C_io_control.hpp
```diff
@@ -24,61 +24,70 @@ FOUR_C_NAMESPACE_OPEN
...
-    //! flag indicating if we write to the control file
-    bool do_write_;
-
-    //! current indent (in spaces)
-    size_t indent_ = 0;
-
-    //! amount of indent (in spaces) added per level of indentation
-    constexpr static int indent_increment_ = 4;
+    //! Pointer-to-implementation to hide implementation details.
+    std::unique_ptr<Internal::ControlFileWriterImpl> pimpl_;
```
georghammerl, line 90: the variable name is very short and not really understandable

sebproell: It is a standard idiom and thus I use its name: https://en.cppreference.com/w/cpp/language/pimpl.html

georghammerl: Interesting, this was unknown to me so far.

PR #878, opened 2025-06-10
Add node-based interface aggregation capability for contact/meshtying problems preconditioned with MueLu

PhilipOesterlePekrun:
@mayrmt
@maxfirmbach
## Description and Context

This change adds the option to use the node-based interface aggregation algorithm for contact/meshtying problems preconditioned with MueLu. A test is also added which shows the necessary .xml layout.

With this change, both the `"contact slaveDofMap"` and the new `"Interface DualNodeID to PrimalNodeID"` are constructed even though only one is used, which is not a huge performance detriment but should be looked at in the future. As we do not currently read the MueLu .xml in 4C, the only way I see is to introduce an optional parameter in the .yaml input file to control which mapping is built (though it is redundant with the .xml).

## Related Issues and Pull Requests

Related to https://github.com/trilinos/Trilinos/pull/13859

PR #1414, opened 2025-10-19
CMake: allow specifying test dependency versions

sebproell:
Versions can be specified with the typical >,<,>=,<=,== comparison operators and an internal version number.

Required for #878 

@maxfirmbach @mayrmt

### DIFF
--- a/cmake/functions/four_c_configure_dependency.cmake
+++ b/cmake/functions/four_c_configure_dependency.cmake
@@ -14,6 +14,29 @@ function(four_c_sanitize_package_name _package_name _output_var)
       )
 endfunction()
 
+# Helper macro to propagate internal version variables to parent scope.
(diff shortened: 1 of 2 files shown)


davidrudlstorfer (dismissed):

sebproell:
I tested this with a bunch of examples I could come up with, and they all seem to work. It's always difficult to consider all the branches in CMake code.

sebproell:
@maxfirmbach With this PR you should be able to add something like `Trilinos>=2025.6` as a `REQUIRED_DEPENDENCY`. This composes much better with the test infrastructure. The test is correctly marked as skipped, it prints a message why it is skipped if using verbose output, etc.

maxfirmbach:
@sebproell Great feature, thanks for the fast implementation.

PR #1417, opened 2025-10-19
Fix maps in structure and constraint conditions

knarfnitram:
## Description and Context

Before `CHECK_EPETRA_CALL` can be used within the wrapper, some errors should be removed first. 
To keep this reviewable, this is a sub part of https://github.com/4C-multiphysics/4C/pull/1390, which aims to provide solutions for wrong vector updates/manipulations due to false maps within the subparts `Structure` (LocSys) and Multipoint constraint.

Part of https://github.com/4C-multiphysics/4C/pull/1390
 See https://github.com/4C-multiphysics/4C/issues/1190 and https://github.com/4C-multiphysics/4C/issues/1121

### DIFF
--- a/src/constraint/4C_constraint_multipointconstraint3.cpp
+++ b/src/constraint/4C_constraint_multipointconstraint3.cpp
@@ -12,7 +12,6 @@
 #include "4C_fem_discretization.hpp"
 #include "4C_fem_dofset_transparent.hpp"
 #include "4C_global_data.hpp"
-#include "4C_linalg_sparsematrix.hpp"
 #include "4C_linalg_utils_sparse_algebra_assemble.hpp"
 #include "4C_rebalance_binning_based.hpp"
 #include "4C_utils_function_of_time.hpp"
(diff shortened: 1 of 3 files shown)


src/constraint/4C_constraint_multipointconstraint3.cpp
```diff
@@ -94,36 +95,36 @@ void Constraints::MPConstraint3::initialize(
...
-  // allocate vectors for amplitudes and IDs
+  const int mid = params.get("OffsetID", 0);
 
-  std::vector<double> amplit(constrcond_.size());
-  std::vector<int> IDs(constrcond_.size());
-  // read data of the input files
+  // Collect only valid abs entries
+  std::vector<double> vals;
+  std::vector<int> gids;
+  vals.reserve(constrcond_.size());
```
maxfirmbach, line 101: Can't you say `std::vector<double> vals(constrcond_.size())`? Same question also for `gids`. Or is it like that because you want to `push_back` in the loop below? Because the loop with the correct size and index is already in place.

src/constraint/4C_constraint_multipointconstraint3.cpp
```diff
@@ -15,6 +15,7 @@
 #include "4C_linalg_sparsematrix.hpp"
 #include "4C_linalg_utils_sparse_algebra_assemble.hpp"
 #include "4C_rebalance_binning_based.hpp"
+#include "4C_structure_new_timint_base.hpp"
```
maxfirmbach: Is this include really necessary?

m-frey (approved):

src/constraint/4C_constraint_multipointconstraint3.cpp
```diff
@@ -550,4 +558,4 @@ void Constraints::MPConstraint3::initialize_constraint(Core::FE::Discretization&
   }
 }  // end of initialize_constraint
 
-FOUR_C_NAMESPACE_CLOSE
+FOUR_C_NAMESPACE_CLOSE
```
maxfirmbach: Missing whitespace after `FOUR_C_NAMESPACE_CLOSE`?

src/core/fem/src/condition/4C_fem_condition_locsys.cpp
```diff
@@ -100,6 +100,18 @@ void Core::Conditions::LocsysManager::update(const double time,
   geo_hierarchy.emplace_back(Core::Conditions::LineLocsys);
   geo_hierarchy.emplace_back(Core::Conditions::PointLocsys);
 
+  // Ensure locsystoggle_ is on the node row map
+  const Core::LinAlg::Map& node_row_map = *discret_.node_row_map();
+
+  // Check if maps match, if not recreate the correct map
+  if (!locsystoggle_ || !locsystoggle_->get_map().same_as(node_row_map))
```
maxfirmbach: Is `!locsystoggle_` the correct way to do things?

src/core/fem/src/condition/4C_fem_condition_locsys.cpp
```diff
@@ -100,6 +100,18 @@ void Core::Conditions::LocsysManager::update(const double time,
...
+  {
+    locsystoggle_ = Core::LinAlg::create_vector(node_row_map, true);
+  }
+
+  FOUR_C_ASSERT(
+      locsystoggle_->get_map().same_as(node_row_map), "locsystoggle_ must use node row map");
```
maxfirmbach: I think the error message should be a full and proper sentence with a description of the situation. Maybe we should not reference the member variable, but add information on what goes wrong.

maxfirmbach (changes requested):

maxfirmbach:
@isteinbrecher @m-frey Are we currently implementing two versions of multi-point constraints again 🤔? I see there's `constraint` and `constraint_framework` 😬.

src/core/fem/src/condition/4C_fem_condition_locsys.cpp
knarfnitram: I removed this error message, since it will never be hit anyway, but it was good for debugging.

src/constraint/4C_constraint_multipointconstraint3.cpp
knarfnitram: yes, i am using `push_back` because we check if the conditions are activated after a certain time and therefore the size of `vals` can change. By using `std::vector<double> vals(constrcond_.size())` we would always have a default-initialized array and therefore we would additionally sum some `0` into the global system vector (at gid 0).

src/core/fem/src/condition/4C_fem_condition_locsys.cpp
knarfnitram: I guess the answer is no. thanks!

src/constraint/4C_constraint_multipointconstraint3.cpp
knarfnitram: I think it should not be needed, is removed.

presets/imcs/cluster/CMakePresets.json
```diff
@@ -34,7 +34,7 @@
         ".imcs_cluster"
       ],
       "cacheVariables": {
-        "CMAKE_BUILD_TYPE": "RELWITHDEBINFO",
+        "CMAKE_BUILD_TYPE": "DEBUG",
```
maxfirmbach: @knarfnitram I think this doesn't belong to this PR.

knarfnitram: true. thanks.

m-frey:
@maxfirmbach I'm only using `constrain_framework`. It seems like the pure multipoint constraint functionalities are duplicated. I don't know how general they are implemented in `constraint`.  
In the `constrain_framework`, they are primarily there to enable the periodic boundary conditions.

m-frey (approved):

PR #1418, opened 2025-10-19
Fix Vector Operations due to wrong maps in Poro

knarfnitram:
## Description and Context

Before `CHECK_EPETRA_CALL` can be used within the wrapper, some errors should be removed first. 
To keep this reviewable, this is a sub part of https://github.com/4C-multiphysics/4C/pull/1390, which aims to provide solutions for wrong vector updates/manipulations due to false maps within the subsection `Poro`. 

Part of https://github.com/4C-multiphysics/4C/pull/1390
 See https://github.com/4C-multiphysics/4C/issues/1190 and https://github.com/4C-multiphysics/4C/issues/1121

### DIFF
--- a/src/poroelast/4C_poroelast_monolithicstructuresplit.cpp
+++ b/src/poroelast/4C_poroelast_monolithicstructuresplit.cpp
@@ -221,8 +221,22 @@ void PoroElast::MonolithicStructureSplit::setup_vector(Core::LinAlg::Vector<doub
 
     // add contribution of Lagrange multiplier from previous time step
     if (lambda_ != nullptr)
-      modfv->update(-ftiparam + stiparam * (1.0 - ftiparam) / (1.0 - stiparam),
-          *structure_to_fluid_at_interface(*lambda_), 1.0);
+    {
+      const double alpha = -ftiparam + stiparam * (1.0 - ftiparam) / (1.0 - stiparam);
+
+      // lambda contribution from "slave" map
+      auto lambda_structure_slave = structure_to_fluid_at_interface(*lambda_);
+
+      // ensure lambda_temp matches modfvs map before update
+      const auto& modfv_map = modfv->get_map();
+      const auto& lambda_map = lambda_structure_slave->get_map();
+
+      if (modfv_map.point_same_as(lambda_map))
+      {
+        modfv->update(alpha, *lambda_structure_slave, 1.0);
+      }
+    }
+
 
     extractor()->insert_vector(*modfv, 1, f);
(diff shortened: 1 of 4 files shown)


c-p-schmidt (dismissed):

src/poroelast/4C_poroelast_monolithicstructuresplit.cpp
```diff
@@ -221,8 +221,33 @@ void PoroElast::MonolithicStructureSplit::setup_vector(Core::LinAlg::Vector<doub
 
     // add contribution of Lagrange multiplier from previous time step
     if (lambda_ != nullptr)
-      modfv->update(-ftiparam + stiparam * (1.0 - ftiparam) / (1.0 - stiparam),
-          *structure_to_fluid_at_interface(*lambda_), 1.0);
+    {
+      const double alpha =
+          -ftiparam + stiparam * (1.0 - ftiparam) / 1.0 - stiparam;  // same as before
```
sebproell: Did you forget the parentheses?


```suggestion
      const double alpha =
          -ftiparam + stiparam * (1.0 - ftiparam) / (1.0 - stiparam);  // same as before
```

knarfnitram: Yes. thanks for catching that!

src/poroelast/4C_poroelast_monolithicstructuresplit.cpp
```diff
@@ -221,8 +221,22 @@ void PoroElast::MonolithicStructureSplit::setup_vector(Core::LinAlg::Vector<doub
...
+      auto lambda_structure_slave = structure_to_fluid_at_interface(*lambda_);
+
+      // ensure lambda_temp matches modfvs map before update
+      const auto& modfv_map = modfv->get_map();
+      const auto& lambda_map = lambda_structure_slave->get_map();
+
+      if (modfv_map.point_same_as(lambda_map))
+      {
+        modfv->update(alpha, *lambda_structure_slave, 1.0);
+      }
```
c-p-schmidt, line 237: When the maps are not the same, then no update is performed. Shouldn't we rather do an export or throw an error in this case?

knarfnitram: You are asking the right questions, I also think that this is a little bit suspicious. 
However within the scope of this MR I do not want investigate this further, since the current status is that this path is not executed and this is preserved it. However I documented it within an the issue #1433 .

PR #1420, opened 2025-10-20
Add beam-to-beam point couplings via CPP

isteinbrecher:
## Description and Context

Up until now beam point couplings were only possible on between nodes. This PR adds the functionality to have point couplings between the closest points between two beams (found via closest point projections in 4C and not the input).

We now also have the option to directly create beam interaction pairs from the beam interaction conditions which will allow for a more fine tuned pair creation for other problems, e.g., beam-to-solid volume coupling.


tests/input_files/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_elbow_offset_indirect.4C.yaml
```diff
@@ -0,0 +1,146 @@
+PROBLEM TYPE:
+  PROBLEMTYPE: "Structure"
+BEAM INTERACTION/BEAM TO BEAM POINT COUPLING CONDITIONS:
+  - E: 1
```
knarfnitram: can you please provide here a short description explaining a little bit about the test case?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair_condition.cpp
```diff
@@ -8,91 +8,161 @@
...
+    for (const auto& adj_elem : adj_elements)
+    {
+      const Core::Elements::Element* element = adj_elem.user_element();
+      if (element && element->id() < lowest_gid)
+      {
+        lowest_gid = element->id();
+        lowest = element;
+      }
+    }
+    return lowest;
```
sebproell:
```suggestion
    return std::ranges::min_element(node->adjacent_elements(), {}, [](const auto& ele) { return ele.id();} );
```
Should work. You will get a `ElementRef` as return value, so just call `user_element` if you want the special pointer.

src/geometry_pair/4C_geometry_pair_line_to_line_evaluation_data.cpp
```diff
@@ -0,0 +1,113 @@
...
+  // Build the node to minimum element GID maps for both conditions.
+  for (size_t i_condition = 0; i_condition < 2; i_condition++)
+  {
+    const auto element_id_map =
+        condition_to_element_id_map((i_condition == 0) ? *condition_1 : *condition_2);
+    for (const auto& [_, element] : element_id_map)
+    {
+      // We can use this GID here, as we will also have access to these IDs inside the pair.
+      const int element_gid = element->id();
+      for (const auto node : element->node_range())
```
sebproell, line 39: Pretty sure this is also a simple range algorithm.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair_condition.cpp
isteinbrecher: Thanks for the suggestion, a slight modification was required, but now it works.

src/geometry_pair/4C_geometry_pair_line_to_line_evaluation_data.cpp
isteinbrecher: I am not exactly sure what you mean, but by using `try_emplace` I was able to simply this loop.

davidrudlstorfer (dismissed):

tests/input_files/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_elbow_offset_indirect_projection_on_nodes.4C.yaml
```diff
@@ -0,0 +1,156 @@
+PROBLEM TYPE:
+  PROBLEMTYPE: "Structure"
+BEAM CONTACT:
+  MODELEVALUATOR: "Standard"
```
knarfnitram, line 7: It would be nice to have here a small description what the test case is supposed to test.

knarfnitram (approved):
Thanks for providing the feature @isteinbrecher , I only have a small remark regarding documentation.

tests/input_files/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_elbow_offset_indirect_projection_on_nodes.4C.yaml
isteinbrecher: Good point, I added descriptions to the new tests.

PR #1421, opened 2025-10-21
Avoid shared_ptr on DiscretReader/Writer

sebproell:
Reader and writer should obviously not _own_ a Discretization but just use it/refer to it.

Related to #413

Helps #1407

### DIFF
--- a/src/core/io/src/4C_io.cpp
+++ b/src/core/io/src/4C_io.cpp
@@ -23,18 +23,11 @@
 FOUR_C_NAMESPACE_OPEN
 
 
-/*----------------------------------------------------------------------*/
-/*----------------------------------------------------------------------*/
-Core::IO::DiscretizationReader::DiscretizationReader() /* [PROTECTED] */
-    : dis_(nullptr), input_(nullptr), restart_step_(nullptr), reader_(nullptr), meshreader_(nullptr)
-{
-  // intentionally left blank
-}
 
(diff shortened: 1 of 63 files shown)


src/core/io/src/4C_io.cpp
```diff
@@ -470,32 +463,10 @@ std::shared_ptr<Core::IO::HDFReader> Core::IO::DiscretizationReader::open_files(
...
-}
-
-/*----------------------------------------------------------------------*/
-/*----------------------------------------------------------------------*/
-Core::IO::DiscretizationWriter::DiscretizationWriter(std::shared_ptr<Core::FE::Discretization> dis,
+Core::IO::DiscretizationWriter::DiscretizationWriter(Core::FE::Discretization& dis,
     std::shared_ptr<OutputControl> output_control,
     const Core::FE::ShapeFunctionType shape_function_type)
-    : dis_(Core::Utils::shared_ptr_from_ref(
-          *dis.get())),  // no ownership to break circle discretization<>writer
```
sebproell, line 498: The major change is this line. This comment was misleading. The only thing that `shared_ptr_from_ref` does is that `shared_ptr` does not perform a deletion but it is still incrementing the use count and thus locked Discretization and its corresponding DiscretizationWriter into a cycle where each refers to the other forever.

c-p-schmidt (approved):

PR #1422, opened 2025-10-21
Avoid shared_ptr on OutputControl

sebproell:
Follows #1421. Turns out there are more reference cycles that prevent me from finishing #1407 

~~Does not quite work yet due to micro-macro/nested parallelism hacking the infrastructure.~~

### DIFF
--- a/src/core/io/src/4C_io.hpp
+++ b/src/core/io/src/4C_io.hpp
@@ -221,8 +221,7 @@ namespace Core::IO
      * @param[in] output_control        output control file
      * @param[in] shape_function_type   shape function type of the underlying fe discretization
      */
-    DiscretizationWriter(Core::FE::Discretization& dis,
-        std::shared_ptr<OutputControl> output_control,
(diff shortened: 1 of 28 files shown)


lauraengelhardt (approved):

src/core/io/src/4C_io.hpp
```diff
@@ -354,10 +353,7 @@ namespace Core::IO
     //@}
 
     /// get output control
-    [[nodiscard]] std::shared_ptr<OutputControl> output() const { return output_; }
-
-    /// set output control
-    void set_output(std::shared_ptr<OutputControl> output);
+    [[nodiscard]] OutputControl& output() const { return output_; }
```
ppraegla, line 356: Does it make sense to return a const reference or remove the const from the function?
The shared_ptr had the same problem. But isn't that against the const-correctness?

sebproell: Depends on how we define the act of writing to disk compared to other mutations of the object. I can see where you are coming from here. I think this needs more cleanup passes anyway. I cannot just make it `const` at the moment.

PR #1431, opened 2025-10-23
Fix assembly order for scatra-x-interactions

dragos-ana:
Fixes a bug regarding the assembly order for scatra interactions with other fields, identified by @c-p-schmidt and me.

- Referenced the correct dofset index for the respective fields on the scatra integrator. This
replaces the hardcoded, assumed order and fixes
hereby an ordering bug.
- Fixed the determination of the maximum dofset within the scatra integrator, to be at least 0.


c-p-schmidt:
Thank you for the fix. Looks good to me. But could you also address the occurrences in the `ssti`-framework?

PR #1432, opened 2025-10-23
[constraint_framework] Add partial nullspace constraint

maxfirmbach:
## Description and Context

This PR adds the functionality to define the nullspace constraint used for e.g. pure Neumann problems only on parts of the discretization. For example consider three bodies with two being Dirichlet constrained while the third one is purely Neumann constraint. This PR enables to only set the nullspace constraint on the third body, while the other two are omitted.

From a sparsity pattern perspective this would look something like this:
sparsity

This PR also adds a test for this functionality considering three beams, two of those are fixed by Dirichlet constrains while the third one is deforming under a load.

A leftover block matrix definition from an earlier PR is also removed.


src/constraint_framework/4C_constraint_framework_submodelevaluator_nullspace.cpp
```diff
@@ -136,11 +150,12 @@ void Constraints::SubmodelEvaluator::NullspaceConstraintManager::evaluate_coupli
     }
   }
 
-  Q_dL_ = std::make_shared<Core::LinAlg::SparseMatrix>(*dof_map, constraint_space.num_vectors());
-
+  auto constraint_matrix = std::make_shared<Core::LinAlg::SparseMatrix>(
+      *dof_condition_map_, constraint_space.num_vectors());
   Core::LinAlg::multi_vector_to_linalg_sparse_matrix(
-      constraint_space, *dof_map, *constraint_map_, *Q_dL_);
+      constraint_space, *dof_condition_map_, *constraint_map_, *constraint_matrix);
```
mayrmt, line 156: The constraint matrix is not a sparse matrix, rather a tall-and-skinny dense matrix, right? Why convert it to a sparse matrix?

maxfirmbach: The global data structure for matrices that we use and work with is a sparse matrix, there's no way around it to use something different. Yes the constraint matrix can at some point be completely dense, but in this case there's still some sparsity left.

Sparse matrices -> global and distributed
Dense matrices -> local and serial

maxfirmbach: In addition, there's nothing like a globally distributed dense matrix. Something like this doesn't exist to my knowledge.

mayrmt: Sure! Trilinos sometimes uses MultiVectors for tall and skinny matrices, though I know that they do not support all linear algebra operations that one might need.

maxfirmbach: @mayrmt The problem also is that I can't assemble a multi vector into a block matrix.

PR #1434, opened 2025-10-24
[solver] Remove monolithic part of `amgnxn`

maxfirmbach:
## Description and Context

This PR removes the monolithic part of `amgnxn`. Coverage reports tells that this is not used anymore as all tests are already ported to use `MueLu`.

This PR also removes a few unused classes and methods.

Part of #754


maxfirmbach:
This also removes another inheritance from `Epetra_Operator`.

PR #1435, opened 2025-10-24
ControlFileEntry to traverse control files

sebproell:
First step to remove the io/legacy_table data structures. After this PR, this data structure is only used in the `post` directory.

### DIFF
--- a/src/core/io/src/4C_io_control.hpp
+++ b/src/core/io/src/4C_io_control.hpp
@@ -11,12 +11,12 @@
 #include "4C_config.hpp"
 
 #include "4C_fem_general_shape_function_type.hpp"
-#include "4C_io_legacy_types.hpp"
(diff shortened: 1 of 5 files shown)


src/core/io/src/4C_io_control.hpp
```diff
@@ -177,36 +175,127 @@ namespace Core::IO
     const bool write_binary_output_;
   };
 
+  template <typename T>
+  concept ControlFileSupportedType =
```
lauraengelhardt: I would call this `ControlFileEntrySupportedType` to make sure not to confuse it with the file type of the control file, but I'll leave that up to you!

lauraengelhardt (approved):
Thanks @sebproell, I just have a minor suggestion 😃

PR #1360, opened 2025-10-08
Add FSI tutorial with monolithic approach

reginabuehler:
## Description and Context

This PR adds the monolithic FSI tutorial, presented during the UKACM GACM Autumn School 2025 (https://github.com/mayrmt/UKACM_GACM_Tutorial_4C_QUEENS) to the 4C tutorials. This includes:
- some minor changes to generalise the material of the hands-on workshop session to a general tutorial (only the formulations of instructions and removing mentions of the virtual machine/docker where not necessary)
- a test using the provided input file with the iterative solver

Some minor (visual) points I would like to handle in a separate PR: (cf. issue #1280)
- display of references
- display of notes/hints
- robust links to ~~other parts of the documentation (if possible)~~ material references

## Related Issues and Pull Requests

Closes #1339 

## Interested parties
@mayrmt @ischeider @georghammerl @c-p-schmidt

Issue #1280, opened 2025-09-17
[Tutorials] Enhancements for the (Preconditioning) Tutorials


reginabuehler:
## Description
This issue serves as a central place to gather and track ideas for improving the Linear Solver [Preconditioner Tutorial](https://4c-multiphysics.github.io/4C/documentation/tutorials/tutorial_preconditioning.html). The goal is to collect suggestions for new content, clarifications, re-structuring, or any other enhancements that can make the tutorial more useful and comprehensive.

## Enhancements (in connection to the [preconditioning tutorial](https://4c-multiphysics.github.io/4C/documentation/tutorials/tutorial_preconditioning.html))
- [ ] FSI: align block matrix figure with implementation/preconditioner ordering
- [ ] `section_dump` for solver xml files analogously to yaml files (if possible)
- [ ] refer to extensive preconditioning material in MueLu: https://github.com/trilinos/Trilinos/pull/14461 and https://muelu-tutorial.readthedocs.io/en/latest/
- [ ] add more complicated block preconditioner example than `fsi`
- [ ] refer to the tutorial [in the preconditioner section of the documentation](https://4c-multiphysics.github.io/4C/documentation/analysis_guide/linear_solvers.html#preconditioners)

## (Minor visual) Enhancements (in connection to the monolithic FSI tutorial https://github.com/4C-multiphysics/4C/pull/1360)
- [ ] display of references
- [ ] display of notes and hints 
- [ ] display of (especially longer paths) in text
- [x] robust links within the documentation (solved in #1437)
- [x] especially: material references (see down below and #1368) (solved in #1437)
- [ ] discussion of `COUPALGO` input parameter (see down below)
- [ ] remove `make_unique_anchor` function, ensure all anchors are unique (follow up of discussion in #1437)

## Issue with material references within documentation
Here, we'll have a problem in future, since the materials are members of the list of `MATERIALS`, which won't have a link anchor when the reference documentation changes. Let's see how we will get rid of this flaw.

_Originally posted by @ischeider in https://github.com/4C-multiphysics/4C/pull/1360#discussion_r2416379311_

## Discussion of input parameter `COUPALGO`
Very nice and extensive! We have to discuss how we can make the material models linkable, they lead to an 404 page. One comment: you did not discuss the `COUPALGO`, which I find an important switch. Anyway, it's great!

_Originally posted by @ischeider in https://github.com/4C-multiphysics/4C/pull/1360#pullrequestreview-3335477628_
            
[...]

PR #1437, opened 2025-10-26
[documentation] Enhance anchors in the reference guide

reginabuehler:
## Description and Context

This PR fixes the material reference links in the monolithic FSI tutorial and introduces an enhancement to the anchors in the reference guide of the documentation.
It introduces dedicated anchors for each option and a small JavaScript utility that automatically opens the corresponding dropdown (`<details>`) when a link with a hash is followed. The anchor uses the name of the `One of` entry.
Previously, the material references were more nested than other references and could not be referenced by a link (e.g. in the documentation).

### Key changes
- Adds globally unique anchors for each `One_of` option across all automatically generated reference files (materials, general, condition, discretization), e.g. `(MATERIALS_MAT_LinElast1D)=` for materials or `(ALEDOMAIN_FLUID_QUAD4)=` for discretization, within the `<details>` environment.
- Includes a small JS script that ensures all ancestor dropdowns open when visiting a link to a specific material.

### Example
A material can be linked like this in `md` now:
```
For details, see the 4C documentation: {ref}`St.-Venant-Kirchhoff<MATERIALS_MAT_Struct_StVenantKirchhoff>`.
```
When clicked, the target page scrolls to the material and automatically expands the relevant dropdown section.
Similarly, element types, conditions, and general parameters can now be referenced in the same way using the generated anchors.

### EDIT: Reason for previously failing pipeline and fix
Initially, this PR only addressed material references. For other reference files (general, condition, discretization), the documentation build failed because duplicate labels were generated for repeated `One of` options. For example, both [ALE GEOMETRY](https://4c-multiphysics.github.io/4C/documentation/reference_guide/discretization_reference.html#ale-geometry) and [ARTERY GEOMETRY](https://4c-multiphysics.github.io/4C/documentation/reference_guide/discretization_reference.html#artery-geometry)) contain `One of` environments within `ELEMENT_BLOCKS` with identical option names, and there is even another nested `One of` layer.

This has been fixed by:
- Prefixing anchors with the surrounding section and option path (e.g. `ALEDOMAIN_FLUID_QUAD4 `) to make all labels unique within a file.
- Propagating this prefix through nested `One of` structures so also nested options get unique, readable labels.

Part of issue #1280 
Follow up of PR #1360

### DIFF
--- a/utilities/four_c_python/src/four_c_documentation/make_documentation.py
+++ b/utilities/four_c_python/src/four_c_documentation/make_documentation.py
@@ -58,6 +58,7 @@ from four_c_metadata.not_set import check_if_set
 DESCRIPTION_MISSING = '<span style="color:grey">*no description yet*</span>'
 TOO_MANY_TESTS_TO_SHOW = 200
 TESTS_TO_SHOW_IF_TOO_MANY = 20
+USED_ANCHORS = set()
 
 
 # Data class to store which sections belong to which chapter (and thus to which file)
@@ -72,6 +73,48 @@ class ReferenceChapter:
     description: str = ""  # Description of the chapter (filled later)
     content: str = ""  # Content of the chapter (filled later)
     has_content: bool = False  # Whether the chapter has any content (filled later)
+
+
+def sanitize_anchor(text: str):
+    """
+    Sanitize a string so it is safe and consistent to use as an explicit MyST anchor.
+
+    - Strip trailing HTML (e.g. ", <span ...>").
+    - Remove '<' and '>'.
+    - Remove whitespace.
+    - Replace '/' with '_'.
+    """
+    # cut off HTML decoration like ", <span style=...>"
+    text = re.sub(r"\s*, <span.*", "", text).strip()
+    text = text.replace("<", "").replace(">", "")
+    text = text.replace(" ", "")
+    text = text.replace("/", "_")
+    return text
+
+
+def make_anchor_base(option_name: str, path_prefix: str | None):
+    """
+    Build the base anchor from an option name and an optional hierarchical prefix.
+    """
+    base = sanitize_anchor(option_name)
+    if path_prefix:
+        base = f"{path_prefix}_{base}"
+    return base
+
+
+def make_unique_anchor(base: str):
+    if base not in USED_ANCHORS:
+        USED_ANCHORS.add(base)
+        return base
+
+    n = 2
+    candidate = f"{base}-{n}"
+    while candidate in USED_ANCHORS:
+        n += 1
+        candidate = f"{base}-{n}"
+
+    USED_ANCHORS.add(candidate)
+    return candidate
 
 
 def make_collapsible(content, summary=None, open_content=None):
@@ -247,7 +290,9 @@ def primitive_to_md(primitive: Primitive, make_description=make_spec_description
     )
 
 
-def vector_or_map_to_md(entry: Vector | Map, make_description=make_spec_description):
+def vector_or_map_to_md(
+    entry: Vector | Map, make_description=make_spec_description, path_prefix: str = ""
+):
     type_description = type_or_none(value_type_flatter(entry), entry)
 
     header_properties = {"size": flatten_size_vector(entry), "required": entry.required}
@@ -260,7 +305,9 @@ def vector_or_map_to_md(entry: Vector | Map, make_description=make_spec_descript
     collapsible_properties = None
     if not isinstance(entry.value_type, Primitive):
         collapsible_properties = {
-            "Value type": all_of_to_md(All_Of([entry.value_type]))
+            "Value type": all_of_to_md(
+                All_Of([entry.value_type]), path_prefix=path_prefix
+            )
         }
 
     return make_description(
@@ -272,7 +319,9 @@ def vector_or_map_to_md(entry: Vector | Map, make_description=make_spec_descript
     )
 
 
-def tuple_to_md(tuple_entry: Tuple, make_description=make_spec_description):
+def tuple_to_md(
+    tuple_entry: Tuple, make_description=make_spec_description, path_prefix: str = ""
+):
     type_description = type_or_none(value_type_flatter(tuple_entry), tuple_entry)
 
     header_properties = {"required": tuple_entry.required}
@@ -282,7 +331,9 @@ def tuple_to_md(tuple_entry: Tuple, make_description=make_spec_description):
         header_properties["default"] = none_to_null(tuple_entry.default)
 
     collapsible_properties = {
-        "Entries": all_of_to_md(All_Of(tuple_entry.value_types), indent=1)
+        "Entries": all_of_to_md(
+            All_Of(tuple_entry.value_types), indent=1, path_prefix=path_prefix
+        )
     }
 
     return make_description(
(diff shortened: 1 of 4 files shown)


reginabuehler:
Back to draft due to failing pipeline (documentation build).

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -400,12 +400,23 @@ def one_of_to_md(one_of: One_Of):
...
+    used_anchors = set()
 
     for i, spec in enumerate(one_of.specs):
+        # remove html formatting for anchor
+        opt_anchor = re.sub(r"\s*, <span.*", "", names[i]).strip()
+        # ensure uniqueness of anchor
+        n = 2
+        while opt_anchor in used_anchors:
+            opt_anchor = f"{opt_anchor}-{n}"
+            n += 1
```
ischeider: Not sure, but to me it seems that this loop produces
```
anchorname
anchorname-1
anchorname-1-2
etc.
```
```suggestion
        opt_anchor_base = opt_anchor
        while opt_anchor in used_anchors:
            opt_anchor = f"{opt_anchorbase}-{n}"
            n += 1
```

reginabuehler: Yes, that is possible, should be fixed along with the issue previously failing the pipeline.

utilities/four_c_python/src/four_c_documentation/make_documentation.py
```diff
@@ -74,6 +75,48 @@ class ReferenceChapter:
...
+        return base
+
+    n = 2
+    candidate = f"{base}-{n}"
+    while candidate in USED_ANCHORS:
+        n += 1
+        candidate = f"{base}-{n}"
+
+    USED_ANCHORS.add(candidate)
+    return candidate
```
amgebauer, line 117: Why is this needed? Is the anchor the name that is used for referencing a specific item? Why is it not unique? It is problematic to append a number if the anchor is already used. References then might silently point to the wrong location. Can we throw an error instead?

amgebauer (approved):
Thank you for adding this nice feature!

utilities/four_c_python/src/four_c_documentation/make_documentation.py
reginabuehler: This is mainly an artefact from my attempts to avoid duplicate anchors to fix the pipeline, but has been actually improved by using and propagating the prefixes in this PR.
Yes, the anchor is used for referencing a specific item. Thinking about it, it might be better to not (artificially) enforce uniqueness like this and cause a "duplicate" error during the documentation build. This would provide a warning on which one can act.
As of right now, I assume each item in the documentation (and therefore each anchor) is unique, even without this function. Let me check and see if it is possible to remove this.

reginabuehler: When removing this "artificial" uniqueness enforcement, the following duplicates listed below occur. 
These are edge cases and most of the documentation works fine. However, they have to be dealt with.
It can be seen easiest for the parameter `FUNCT[n]` [here](https://4c-multiphysics.github.io/4C/documentation/reference_guide/general_reference.html#funct-n):
The `Option (POROMULTIPHASESCATRA_FUNCTION, PARAMS)` is listed 7 times in the documentation. Each listing represents different `Choices` (e.g. `LUNG_CARBONDIOXIDE_EXCHANGE_LAW`) which each require different parameters. For the other duplicates a similar situation occurs.

```
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/condition_reference.md:11117: WARNING: Duplicate explicit target name: "designs2ikineticslineconditions_kinetic_model,numscal,stoichiometries,e-,k_r,alpha_a,alpha_c,is_pseudo_contact,resistance,convtol_implbutlervolmer,itemax_implbutlervolmer". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/condition_reference.md:12089: WARNING: Duplicate explicit target name: "designs2ikineticspointconditions_kinetic_model,numscal,stoichiometries,e-,k_r,alpha_a,alpha_c,is_pseudo_contact,resistance,convtol_implbutlervolmer,itemax_implbutlervolmer". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/condition_reference.md:13163: WARNING: Duplicate explicit target name: "designs2ikineticssurfconditions_kinetic_model,numscal,stoichiometries,e-,k_r,alpha_a,alpha_c,is_pseudo_contact,resistance,convtol_implbutlervolmer,itemax_implbutlervolmer". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/condition_reference.md:14591: WARNING: Duplicate explicit target name: "designscatramulti-scalecouplingpointconditions_numscal,stoichiometries,e-,k_r,alpha_a,alpha_c,is_pseudo_contact,resistance,convtol_implbutlervolmer,itemax_implbutlervolmer". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/condition_reference.md:27228: WARNING: Duplicate explicit target name: "electrodeboundarykineticslineconditions_alpha,i0,gamma,refcon". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/condition_reference.md:27883: WARNING: Duplicate explicit target name: "electrodeboundarykineticspointconditions_alpha,i0,gamma,refcon". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/condition_reference.md:28541: WARNING: Duplicate explicit target name: "electrodeboundarykineticssurfconditions_alpha,i0,gamma,refcon". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/general_reference.md:12788: WARNING: Duplicate explicit target name: "functn_poromultiphasescatra_function,params". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/general_reference.md:12852: WARNING: Duplicate explicit target name: "functn_poromultiphasescatra_function,params". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/general_reference.md:12916: WARNING: Duplicate explicit target name: "functn_poromultiphasescatra_function,params". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/general_reference.md:12996: WARNING: Duplicate explicit target name: "functn_poromultiphasescatra_function,params". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/general_reference.md:13080: WARNING: Duplicate explicit target name: "functn_poromultiphasescatra_function,params". [docutils]
/home/a11brebu/4C_build_docs_dec/doc/tmp/reference_guide/general_reference.md:13164: WARNING: Duplicate explicit target name: "functn_poromultiphasescatra_function,params". [docutils]
```

I see a couple of options:

1. Merge as is and enforce uniqueness of the anchors by appending the number.
2. Remove the `make_unique_anchor` function. Adapt the anchor generation to also include any nested choices which might differ. This might give some unnecessary overhead in the anchors since a lot of items have `Choices`, but do not necessarily produce duplicate anchors (i.e. all their choices have the same parameters). I could look into a more detailed solution that only uses the `Choices` in the anchor when necessary to avoid duplicates.

If there are any other solution ideas, I'd be happy to discuss.

amgebauer: ok, I understand the problem. I think about the special cases until tomorrow. If we cannot fix them quickly, we can merge as is. 

(at least the one that I checked is using some deprecated things, one might can just modernize it so that the ids are unique)

amgebauer: I think one could quite easily get rid of these duplication by merging the all-ofs in 4C, that have the same options (but different accepted values) into one. This is however quite inhomogeneous compared to what we have currently. We might wait with that once we restructure the respective condition inputs.

Alternative would be to not just use the names of the options but also the available values for selections, which is not super straight-forward.

I suggest to merge as is for now and revisit it later.

reginabuehler: Thanks for the review. I will merge this as is then and note it for possible further improvements in issue #1280.

PR #1445, opened 2025-10-27
[NOX] Implement `NOX::Nln::Vector`

vovannikov:
## Description and Context
Class `NOX::Nln::Vector` introduced earlier in #1399 is no longer inherited from `::NOX::Epetra::Vector` rather from `::NOX::Abstract::Vector`.

The current implementation contains 2 temporary methods which are implemented in order to reduce the amount of work to be performed within a single PR:
- method `getEpetraVector()` should be removed. This is relatively simple to do: the wrappings and object conversions around these calls should be simplified.
- implicit conversion from `NOX::Nln::Vector` to `::NOX::Epetra::Vector`. This is the most ugly thing so far, but it requires a bit more work to completely git rid of it. For instance, one of the problematic spots is the use of `::NOX::Epetra::MatrixFree` and the other related classes from `::NOX::Epetra` namespace inside of `FSI::Partitioned`. It looks like all of these classes have to be replaced with our own versions:
https://github.com/4C-multiphysics/4C/blob/e9cbd1b8923fa74b6bd9233af7947fc7d79be199/src/fsi/src/partitioned/4C_fsi_partitioned.cpp#L504-L509

The next step is to get rid of `getEpetraVector()` member method.

#1399

### DIFF
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
@@ -31,17 +31,17 @@ namespace NOX
         Copy
       };
 
-      //! Ctor that get a shared_ptr with the state vector
+      //! Ctor that gets a shared_ptr with the state vector
       Vector(const std::shared_ptr<Core::LinAlg::Vector<double>>& source,
           MemoryType memory = MemoryType::Copy, ::NOX::CopyType type = ::NOX::DeepCopy);
 
       //! Ctor that captures the ownership of the vector
       Vector(Core::LinAlg::Vector<double>&& source);
 
-      //! Ctor that copies that provided state vector
+      //! Ctor that copies the provided state vector or its shape
       Vector(const Core::LinAlg::Vector<double>& source, ::NOX::CopyType type = ::NOX::DeepCopy);
 
-      //! Ctor that copies from another NOX::Nln::Vector
+      //! Ctor that copies data or just shape from another NOX::Nln::Vector
       Vector(const NOX::Nln::Vector& source, ::NOX::CopyType type = ::NOX::DeepCopy);
 
       //! Main copy assignment operator
@@ -101,10 +101,10 @@ namespace NOX
       ::NOX::size_type length() const override;
 
       //! Get reference to underlying Epetra vector - temporary and is to be removed
-      Epetra_Vector& getEpetraVector();
+      Epetra_Vector& getEpetraVector();  // NOLINT(readability-identifier-naming)
 
       //! Get const reference to underlying Epetra vector - temporary and is to be removed
-      const Epetra_Vector& getEpetraVector() const;
+      const Epetra_Vector& getEpetraVector() const;  // NOLINT(readability-identifier-naming)
 
       // Create a view-like NOX::Epetra::Vector wrapper - temporary and is to be removed
       operator ::NOX::Epetra::Vector();
@@ -112,9 +112,9 @@ namespace NOX
       // Create a view-like NOX::Epetra::Vector wrapper - temporary and is to be removed
       operator ::NOX::Epetra::Vector() const;
 
-     protected:
+     private:
       //! Pointer to a storage vector owned by this object
-      std::shared_ptr<Core::LinAlg::Vector<double>> linalg_vec;
+      std::shared_ptr<Core::LinAlg::Vector<double>> linalg_vec_;
     };
   }  // namespace Nln
 }  // namespace NOX
--- a/src/fsi/src/utils/4C_fsi_statustest.cpp
+++ b/src/fsi/src/utils/4C_fsi_statustest.cpp
@@ -41,25 +41,21 @@ double NOX::FSI::GenericNormF::compute_norm(const Core::LinAlg::Vector<double>& 
 {
   int n = v.global_length();
   double norm;
-  int err;
 
   switch (norm_type_)
   {
     case ::NOX::Abstract::Vector::TwoNorm:
-      err = v.norm_2(&norm);
-      if (err != 0) FOUR_C_THROW("norm failed");
+      v.norm_2(&norm);
       if (scale_type_ == Scaled) norm /= sqrt(1.0 * n);
       break;
 
     case ::NOX::Abstract::Vector::OneNorm:
-      err = v.norm_1(&norm);
-      if (err != 0) FOUR_C_THROW("norm failed");
+      v.norm_1(&norm);
       if (scale_type_ == Scaled) norm /= n;
       break;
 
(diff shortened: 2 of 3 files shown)


vovannikov:
There was a bug with scaling initially, that has been fixed.

I also muted the clang-tidy check for `getEpetraVector()` to keep the PR compact as it will be removed anyway.

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
```diff
@@ -19,36 +22,99 @@ namespace NOX
...
+      //! Get const reference to underlying Epetra vector - temporary and is to be removed
+      const Epetra_Vector& getEpetraVector() const;  // NOLINT(readability-identifier-naming)
+
+      // Create a view-like NOX::Epetra::Vector wrapper - temporary and is to be removed
+      operator ::NOX::Epetra::Vector();
+
+      // Create a view-like NOX::Epetra::Vector wrapper - temporary and is to be removed
+      operator ::NOX::Epetra::Vector() const;
+
+     protected:
```
sebproell:
```suggestion
private:
```

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
```diff
@@ -19,36 +22,99 @@ namespace NOX
...
+      //! Pointer to a storage vector owned by this object
+      std::shared_ptr<Core::LinAlg::Vector<double>> linalg_vec;
```
sebproell: We have `OwnerOrView` which we use a member for linalg wrappers. This enables most use cases of converting to/from `LinAlg::Vector`.

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
```diff
@@ -19,36 +22,99 @@ namespace NOX
...
+      //! Ctor that get a shared_ptr with the state vector
+      Vector(const std::shared_ptr<Core::LinAlg::Vector<double>>& source,
+          MemoryType memory = MemoryType::Copy, ::NOX::CopyType type = ::NOX::DeepCopy);
+
+      //! Ctor that captures the ownership of the vector
+      Vector(Core::LinAlg::Vector<double>&& source);
 
-      Vector(const Epetra_Vector& source, ::NOX::CopyType type = ::NOX::DeepCopy);
+      //! Ctor that copies that provided state vector
+      Vector(const Core::LinAlg::Vector<double>& source, ::NOX::CopyType type = ::NOX::DeepCopy);
```
sebproell, line 42: What does the second parameter say for this ctor? Isn't this always a copy?

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
```diff
@@ -19,36 +22,99 @@ namespace NOX
...
+        View,
+        Copy
       };
 
-      //! Reproduce ctors of ::NOX::Epetra::Vector
-      Vector(const Teuchos::RCP<Epetra_Vector>& source, MemoryType memoryType = MemoryType::Copy,
-          ::NOX::CopyType type = ::NOX::DeepCopy);
+      //! Ctor that get a shared_ptr with the state vector
+      Vector(const std::shared_ptr<Core::LinAlg::Vector<double>>& source,
+          MemoryType memory = MemoryType::Copy, ::NOX::CopyType type = ::NOX::DeepCopy);
```
sebproell, line 36: I think this ctor is mostly lying in the sense that it shares ownership, but you seem to mostly pass a non-owning shared_ptr. I think you should get rid of this one next.

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
```diff
@@ -19,36 +22,99 @@ namespace NOX
...
+      Core::LinAlg::Vector<double>& get_linalg_vector();
+
+      //! Get const reference to underlying Epetra vector.
+      const Core::LinAlg::Vector<double>& get_linalg_vector() const;
+
+      //! Initialize every element of this vector with @p gamma .
+      ::NOX::Abstract::Vector& init(double gamma) override;
+
+      //! Initialize each element of this vector with a random value.
+      ::NOX::Abstract::Vector& random(bool, int) override;
```
sebproell, line 68: That's a pretty random function to have in a vector interface that NOX forces on us ;) What do I pass for `bool` and `int` here?

sebproell: To answer this myself: `(bool useSeed=false, int seed=1)` What a ridiculous function call.

vovannikov: Yep, I agree. This is a possible implementation from Trilinos. Basically, these options are just seed related:

https://github.com/trilinos/Trilinos/blob/7c8afbd28691f64cff071e579701bfe4ec65b837/packages/nox/src-epetra/NOX_Epetra_Vector.C#L128-L134

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
vovannikov: Actually, this ctor is one is the mostly used I guess. It is just those calls to `Core::Utils::shared_ptr_from_ref()` are super visible :) Probably you are right, this can get removed provided refactoring around the usage of `NOX::Nln::Vector` is perfromed.

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
vovannikov: Not always, only shape can be copied without the data itself. The same applies to all other ctors, except from the one receiving an r-value reference. I have updated the comment accordingly.

src/solver_nonlin_nox/4C_solver_nonlin_nox_vector.hpp
vovannikov: Yes, indeed. But I do not see significant benefits from using it in this context. Actually here I may see just the opposite (again from my experience when debugging these changes): the ownership scheme in 4C gets out of control so easy. Semantically, `linalg_vec` should not be a view: it has to be a shared object indeed, ensuring its lifetime is essential for functionality of `NOX::Nln::Vector`. Maybe I do not fully understand the use cases of `OwnerOrView`.

sebproell: I guess you are right that this class will become more of a non-owning adapter to make "our" vector (which happens to be a Trilinos vector under the hood) compatible with NOX :+1: Similar to how you did it in deal.II

PR #1447, opened 2025-10-28
Add spring dashpot condition for shell7p elements

mairehenke:
## Description and Context

This PR adds Robin spring–dashpot boundary condition support for `Shell7p` elements, extending functionality previously limited to (solid) elements with three degrees of freedom per node. It supports both `"xyz"` (for line and surface geometries) and `"refsurfnormal"` (for surface geometries) directions. It also enables spring–dashpot line conditions for non-shell elements.

The previous implementation of the spring–dashpot boundary condition assumed `num_dof_per_node == dim`, which caused incorrect instantiation and indexing for `Shell7p` elements, as they have six DOFs per node rather than three. This PR allows different values to be set for both `num_dof_per_node` and `dim`, ensuring correct behavior for both element types.


src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -163,16 +175,31 @@ namespace
...
 
-          Core::LinAlg::Matrix<dim, 1> velocity_at_gp;
+          Core::LinAlg::Matrix<num_dof_per_node, 1> velocity_at_gp;
           velocity_at_gp.multiply_nn(velocity, shape_functions.values);
 
-          Core::LinAlg::Matrix<dim, 1> normal_at_gp = get_normal<celltype>(jacobian_mapping);
+          Core::LinAlg::Matrix<dim, 1> normal = get_normal<celltype>(jacobian_mapping);
+
+          Core::LinAlg::Matrix<num_dof_per_node, 1> normal_at_gp;
+          normal_at_gp.clear();
```
lauraengelhardt:
```suggestion
          Core::LinAlg::Matrix<num_dof_per_node, 1> normal_at_gp(true);
```
This directly initializes with zeros.

src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -419,13 +447,302 @@ void Constraints::SpringDashpot::evaluate_robin(std::shared_ptr<Core::LinAlg::Sp
...
+                Core::FE::GaussIntegration gauss_integration =
+                    Core::FE::create_gauss_integration<celltype>(
+                        Discret::Elements::DisTypeToOptGaussRule<celltype>::rule);
+
+                if (direction == RobinSpringDashpotType::xyz)
+                {
+                  FOUR_C_ASSERT_ALWAYS(onoff.size() == 3,
+                      "Number of given functions must be 3 when using the xyz-direction of the "
+                      "Robin "
+                      "Spring-Dashpot BC!");
```
lauraengelhardt: Formating seems off here.

src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -419,13 +447,302 @@ void Constraints::SpringDashpot::evaluate_robin(std::shared_ptr<Core::LinAlg::Sp
...
+          const int numdf = 3;
+          std::vector<double> stress(numdim, 0.0);
+
+          for (int node = 0; node < curr.second->num_node(); ++node)
+          {
+            for (int dim = 0; dim < numdim; dim++) stress[dim] = res[node * numdf + dim];
+            springstress_.insert(
+                std::pair<int, std::vector<double>>(curr.second->node_ids()[node], stress));
+          }
+        }
```
amgebauer: If I see it correctly, this is a more or less exact copy of the shell-case above. I think you can generalize it if you extract it into a templated function where `num_dof_per_node` is a template parameter.

src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -419,13 +447,302 @@ void Constraints::SpringDashpot::evaluate_robin(std::shared_ptr<Core::LinAlg::Sp
...
+
+      for (auto& curr : geom)
+      {
+        auto* shell7p_line_element =
+            dynamic_cast<Discret::Elements::Shell7pLine*>(curr.second.get());
+        if (shell7p_line_element == nullptr)
+        {
+          FOUR_C_THROW(
+              "Line elements can only be of type Shell7pLine for spring-dashpot BCs. Cast to "
+              "Shell7pLine failed.");
```
amgebauer: Same as above, so that you can enable it directly also for non-shell elements.

mairehenke:
Thanks for your feedback. I have now moved the duplicate code into templated functions and also enabled support for spring-dashpot line conditions for non-shell elements, as you suggested. Additionally, I extended a test case to cover this functionality.

src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -163,16 +175,31 @@ namespace
...
 
-          Core::LinAlg::Matrix<dim, 1> normal_at_gp = get_normal<celltype>(jacobian_mapping);
+          Core::LinAlg::Matrix<dim, 1> normal = get_normal<celltype>(jacobian_mapping);
+
+          Core::LinAlg::Matrix<num_dof_per_node, 1> normal_at_gp;
+          normal_at_gp.clear();
+
+          // For shell7p elements, copy the normal vector twice (for translations of
+          // displacements and director vector)
+          if constexpr (num_dof_per_node == 6)
```
lauraengelhardt: Based on your comment, wouldn't it be cleaner to check for the element type here (since you get the element anyhow, you can access its type)?

src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -218,6 +246,206 @@ namespace
...
+          }
+
+          const Core::LinAlg::Matrix<num_dof_per_node, Core::FE::num_nodes(celltype)>
+              ele_displacements =
+                  Core::FE::get_element_dof_matrix<celltype, num_dof_per_node>(displacements);
+
+          const Core::LinAlg::Matrix<num_dof_per_node, Core::FE::num_nodes(celltype)> nodal_offset =
+              Core::FE::get_element_dof_matrix<celltype, num_dof_per_node>(displacement_offset);
+
+          Core::LinAlg::Matrix<num_dof_per_node, Core::FE::num_nodes(celltype)> velocity_matrix =
```
lauraengelhardt: const?

src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -218,6 +246,206 @@ namespace
           }
         });
   }
+
+  std::vector<std::function<std::tuple<double, double, double>(double, double)>>
+  create_spring_dashpot_evaluators(const std::vector<int>& onoff,
```
lauraengelhardt: I don't quite understand why onoff is needed as a bool and an int? Maybe I'm missing something, but can't you just pass the boolean vector?

src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -355,162 +587,162 @@ void Constraints::SpringDashpot::evaluate_robin(std::shared_ptr<Core::LinAlg::Sp
...
+          scaled_constant_offset[axis_direction] = scale_by_optional_function(
+              constant_offset[axis_direction], numfuncdisploffset[axis_direction], total_time);
+        }
 
-          const double dashpot_viscosity =
-              scale_by_optional_function(dashpotvisc[i], numfuncvisco[i], total_time);
+        // check if element is either Shell7p or Shell7pScatra
+        bool is_shell7p_element =
+            (dynamic_cast<Discret::Elements::Shell7p*>(curr.second.get()) != nullptr) ||
+            (dynamic_cast<Discret::Elements::Shell7pScatra*>(curr.second.get()) != nullptr);
```
lauraengelhardt: Just check for the `element_type`?

src/constraint/4C_constraint_springdashpot.cpp
mairehenke: You're absolutely right, we don't need both. I removed the int vector :thumbsup:

mairehenke:
Thank you for your review @lauraengelhardt! I included all your suggestions.

mairehenke:
Could you take another look and approve, @amgebauer @lauraengelhardt ?

lauraengelhardt (approved):

PR #1448, opened 2025-10-28
CMake: modular test definitions

sebproell:
This PR implements a more modular approach to defining input file tests. As already discussed in https://github.com/4C-multiphysics/4C/issues/56#issuecomment-2507884342, the old "single command" version does not scale with the increasing number of features we add to the testing framework. Instead, we now define a test and can add further features to it via `RETURN_AS` and `BASED_ON`. I already extracted the restart option and the csv/yaml comparison option. More can be done in a next step.

Draft, as this does not work for all cases yet, and I need to update the docs.

Big part of #56 

@davidrudlstorfer


tests/list_of_tests.cmake
```diff
@@ -9,26 +9,35 @@
 # Switch off cmake-format so that each test specification can be one a single line.
 
 # four_c_test: regular .4C.yaml file based tests
-four_c_test(TEST_FILE ale2d_laplace_material.4C.yaml NP 2 RESTART_STEP 1)
+four_c_test(TEST_FILE ale2d_laplace_material.4C.yaml NP 2 RETURN_AS current)
+four_c_test_restart(BASED_ON ${current} SAME_FILE NP 2 RESTART_STEP 1)
```
davidrudlstorfer, line 13: What I really liked about one of your first versions was the underscore `_` in front of the "added" tests. This not only gave a visual differentiation but also a functional differentiation because an "added" test cannot work alone.

So is the removal of the `_` purely cosmetic or does it also have a functional change?

tests/list_of_tests.cmake
```diff
@@ -9,26 +9,35 @@
...
 four_c_test(TEST_FILE ale2d_laplace_spatial.4C.yaml NP 2)
-four_c_test(TEST_FILE ale2d_solid.4C.yaml NP 2 RESTART_STEP 1)
+four_c_test(TEST_FILE ale2d_solid.4C.yaml NP 2 RETURN_AS current)
+four_c_test_restart(BASED_ON ${current} SAME_FILE NP 2 RESTART_STEP 1)
```
davidrudlstorfer, line 16:
```suggestion
four_c_test_restart(SAME_FILE NP 2 RESTART_STEP 1)
```

do we actually need this `BASED_ON ${current}` approach? If nothing is provided we could just assume current? I think this would even simplify it for the end user

sebproell: Not providing a variable feels like operating on global state. I regex-replaced the current occurrences and thus always used the same name; however, you are free to choose whichever name you prefer. I prefer the explicit mention of the base test over a minor convenience. It is not like you type hundreds of these lines.

tests/list_of_tests.cmake
sebproell: A prefix `_` is conventionally used in CMake for variables and functions that should not be used by client code. Yes, it would give a nice visual differentiation here. The module that loads the `four_c_test` functions also loads a couple of internal functions that start with `_` and I don't want to blur the line between these and the `_four_c_test` functions.

I don't know what you mean by 

> functional differentiation because an "added" test cannot work alone.

Isn't this true regardless of the prefix? Could you clarify?

sebproell:
@davidrudlstorfer

  > What's the reason for the _add_test_with_options helper function? Unfortunately, I do not fully get that

This function existed before. I think it is useful to have our own interface specifying how we define a test. The different functions need to perform different logical steps but in the end they all define a test and we expect certain options to be set (time out, test directory, processes, ...). Basically, _add_test_with_options is the central low-level function that is too cumbersome to call for users. The various `four_c_test` functions are just front-ends for it.


>  If one defines two consecutive four_c_test_restart functions, are they executed in exactly that order or could it be switched up and lead to some undefined testing behavior?

The order is unspecified. Usually, they will run sequentially, but note that there is `--schedule-random` to shuffle the order and detect implicit ordering dependencies.

sebproell:
I extended the timeout of an infamous xfluid test that was already close to the timeout before. Tests are now executed in a slightly different order, so this is somewhat expected to happen.

tests/list_of_tests.cmake
davidrudlstorfer: Good point regarding the `_` to differentiate between internal and regular cmake functions.

Do we have some sort of other differentiation? I think it still would be easier for users to visually differentiate. `-` probably dont work?

>     functional differentiation because an "added" test cannot work alone.

What I meant with that was that a user won't try to only call `four_c_test_restart` to run a test + restart. One could think that this provides the necessary infrastructure to run tests + restart rather than using the base `four_c_test` beforehand.

tests/list_of_tests.cmake
davidrudlstorfer: Sounds good

davidrudlstorfer (approved):
@sebproell thanks for the nice explanations and discussions. I still think that it would make sense to `visually` differentiate between our new base testing functions and add-on options.

Otherwise this would be good to go. Maybe someone else could also take a look. (@c-p-schmidt maybe you could review if you find the time because we discussed this today)

tests/list_of_tests.cmake
sebproell: @davidrudlstorfer I would argue that the `BASED_ON` should make it clear that this test depends on something else. It will not work without providing something here. And no: a leading `-` is not allowed.

> One could think that this provides the necessary infrastructure to run tests + restart rather than using the base four_c_test beforehand.

Yes, one could think that but I believe we are biased because it used to be like this a while ago. In a vacuum, I don't see why you wouldn't expect this to only test a restart. I agree that we can go for some distinction. This is only a simple rename, so if you ever come up with something, feel free to apply it.

davidrudlstorfer: Sounds good - let's get this merged soon to prevent any merge conflicts again.

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -173,205 +283,169 @@ function(four_c_test)
...
     set(_parsed_NP 1)
   endif()
-
-  if(NOT DEFINED _parsed_RESTART_STEP)
-    set(_parsed_RESTART_STEP "")
-  endif()
-
-  if(NOT DEFINED _parsed_TIMEOUT)
-    set(_parsed_TIMEOUT "")
+  if(_parsed_NP GREATER 3)
```
c-p-schmidt, line 291: Just out of curiosity, why exactly 3?

sebproell: I did not change that; this check existed before. I think we limited this mostly because the GitHub runners only have 4 hardware threads.

PR #1449, opened 2025-10-29
Sort the reference sections case insensitive

ischeider:
## Description and Context

According to the discussion #930, the new standard for the input parameters seems to be `snake_case`, opposite to the `SCREAMING CASE WITH SPACES` we have so far.

A side effect is that in the reference documentation, the section with capital letters come first, before the lower case sections, which is probably not wanted (at least I find that odd).
Therefore this PR first makes a case insensitive sort before writing the sections to file.

Discussion #930

### DIFF
--- a/utilities/four_c_python/src/four_c_documentation/make_documentation.py
+++ b/utilities/four_c_python/src/four_c_documentation/make_documentation.py
@@ -626,13 +626,14 @@ def create_markdown_documentation(
         ),
     ]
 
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (approved):

mayrmt:
@isteinbrecher thanks for the reminder!

ischeider:
This is why I also put this on the agenda for today's community meeting. Anyway, since people have already defined a number of section in this style, the reference guide becomes ugly (in my opinion), and the PR just polishes its structure.

PR #1451, opened 2025-10-29
CMake: error on _ROOT vars that aren't directories

sebproell:
We can help users by telling them that the `FOUR _C_<package>_ROOT` variables need to refer to directories. This is especially useful for `FOUR_C_PYTHON_ROOT` where one might pass the `python` executable.

FYI @shervas @lkoeglmeier @c-p-schmidt


lkoeglmeier:
Nice! Thank you!

c-p-schmidt (approved):

PR #1457, opened 2025-10-29
[LinAlg] Add `Epetra` error check to sparse matrix computation methods

maxfirmbach:
## Description and Context

As the title states.

Part of #1190

### DIFF
--- a/src/contact/src/4C_contact_lagrange_strategy.cpp
+++ b/src/contact/src/4C_contact_lagrange_strategy.cpp
@@ -2214,7 +2214,8 @@ void CONTACT::LagrangeStrategy::evaluate_contact(
           Core::LinAlg::matrix_multiply(*kteffmatrix, false, systrafo, false, false, false, true);
       kteffmatrix =
           Core::LinAlg::matrix_multiply(systrafo, true, *kteffmatrix, false, false, false, true);
-      systrafo.multiply(true, *feff, *feff);
+      Core::LinAlg::Vector<double> feffnew(*feff);
+      systrafo.multiply(true, feffnew, *feff);
     }
 
     // transform if necessary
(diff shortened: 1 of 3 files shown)


src/core/linalg/src/sparse/4C_linalg_sparsematrix.hpp
```diff
@@ -518,24 +518,24 @@ namespace Core::LinAlg
     //@{
 
     /// Returns the result of a matrix multiplied by a Core::LinAlg::Vector<double> x in y.
-    int multiply(
+    void multiply(
         bool TransA, const Core::LinAlg::Vector<double>& x, Core::LinAlg::Vector<double>& y) const;
 
     /// Returns the result of a Epetra_CrsMatrix multiplied by a Epetra_MultiVector X in Y.
     int multiply(bool TransA, const Core::LinAlg::MultiVector<double>& X,
         Core::LinAlg::MultiVector<double>& Y) const override;
```
mayrmt, line 526: What about this one? It is still returning `int`.

maxfirmbach: @mayrmt It's overwritten from the base class, which inherits the method from `Epetra_Operator`, thus I'm not allowed to change the return type.

mayrmt (approved):

src/core/linalg/src/sparse/4C_linalg_sparsematrix.hpp
mayrmt: Ok, then this needs to be deferred until changing the internals to Tpetra.

maxfirmbach: That's one of the reasons why we want to get rid of the `Epetra_Operator` inheritance.

Issue #1426, opened 2025-10-22
Port BEAMS_NEWGAP Functionalty to Beam Interaction


knarfnitram:
## Description

The `BEAMS_NEWGAP` feature ensures that during beam to beam contact simulation in case of penetrating or crossing beams the gap function is enhanced to account for the change within the sign of the normal.
To include this information in the calculation and determine if a switch in the sign occurs the gap function and normal needs of the previous state needs to be stored.
However in the current beam interaction the contact pairs are created new during every time step. This (probably) explains why this feature was so far not ported.
Currently blocking: https://github.com/4C-multiphysics/4C/pull/1175 
## Possible Solution and Definition of Done

- [x] Port the history variables ( `old_normal_` , `sign` and `gap_original_`) from `Beam3contactnew` to `4C_beaminteraction_beam_to_beam_contact_pair` 
- [x] Adapt calculations in `4C_beaminteraction_beam_to_beam_contact_pair` based from `4C_beamcontact_beam3contactnew`
- [x] Adapt the testcases `beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_twocrossedbeams.4C` and `beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_smoothed_twocrossedbeams.4C.yaml`
- [x] Initialize the contact pair with the previous values in  `4C_beaminteraction_beam_to_beam_contact_condition` and adding a shared pointer to the vector with the pair element ID to normals/gap map.
- [ ] ~~Find out a place and data structure, where to extract and store the information about previous pairs and `normals`.~~ Store a map between the pair element GIDs and the normals/gaps in `4C_beaminteraction_beam_to_beam_contact_condition`.
- [x] After all pairs are created, ensure Information of the pari element id map to normals is imported exported in parallel, as the pairs might not be on the same processor as in the previous tilmestep
- [ ] Rename `BEAMS_NEWGAP` to `PREVIOUS_NORMAL_FOR_GAP` , `ENHANCED_GAP_FUNCTION` ,`BEAMS_HISTORY_GAP_FUNCTION`, `BEAMS_HISTORY_GAP` or `BEAMS_ENHANCED_GAP_FUNCTION`
## Interested Parties

@isteinbrecher @maxfirmbach @davidrudlstorfer

maxfirmbach:
Thanks for opening this issue and providing this detailed description! We should definitely tackle this soon.

maxfirmbach:
[...]

PR #1459, opened 2025-10-30
Port BEAMS_NEWGAP Functionality to Beam-Interaction

knarfnitram:
## Description and Context

- Ports the `BEAMS_NEWGAP` feature to the `Beam-Interaction-Framework`. This feature ensures that the `normal` calculation between two intersecting beams is enhanced that during the event of crossing, the gap calculation is enhanced by taking the previous normal into account. This leads to a more robust evaluation of the beam to beam contact forces. 
- The normals calculated within the `beam_contact_pair` are stored at the end of the evaluation. During the set up of the beam interaction condition the previous values are redistributed. (@isteinbrecher so far I did not manage to redistribute them based on an `Epetra_Map`, what do you think about the current redistribution strategy?). 
- To ensure a small parallel test case based on multiple crossing fibers is added. 
_Note:_ Complete a complete redistribution is unfortunately not triggered with this test case. 

The test case _beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_smoothed_twocrossedbeams_ can not be ported within the scope of this MR since the option `BEAMS_SMOOTHING:` `"Cpp"` is currently not supported in the `Beam-Interaction-Framework`.  

Related to https://github.com/4C-multiphysics/4C/issues/1426 , https://github.com/4C-multiphysics/4C/pull/1175 , https://github.com/4C-multiphysics/4C/discussions/1022

### DIFF
--- a/src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_condition.cpp
+++ b/src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_condition.cpp
@@ -75,6 +75,10 @@ void BeamInteraction::BeamToBeamContactCondition::clear()
 void BeamInteraction::BeamToBeamContactCondition::setup(
     const std::shared_ptr<const Core::FE::Discretization>& discret)
 {
+  // Exchange here the previously used old_cp_normals_ to ensure that each
+  // beam to beam contact pairs, which can now be on a different proc
+  // has the normal needed.
+
   // hard coded number of pair size and length of normal
   const int pair_size = 2;
   const int normal_size = 3;
@@ -187,16 +191,18 @@ void BeamInteraction::BeamToBeamContactCondition::setup(
   // Print warning on debug if some normals are not found
   if (!local_missing_keys.empty())
   {
-    Core::IO::cout(Core::IO::debug)
-        << "[Rank " << Core::Communication::my_mpi_rank(discret->get_comm())
+    const auto& ex = *local_missing_keys.begin();
+    std::ostringstream msg;
+    msg << "[Rank " << Core::Communication::my_mpi_rank(discret->get_comm())
         << "] is still missing " << local_missing_keys.size()
-        << " contact pair normals after exchange in BeamToBeamContactCondition. Example key: "
-        << Core::IO::endl;
-    auto ex = *local_missing_keys.begin();
-    Core::IO::cout(Core::IO::debug) << "(" << ex.ele1 << "," << ex.ele2 << ")" << Core::IO::endl;
+        << " contact pair normals after exchange in BeamToBeamContactCondition. "
+        << "Example key: (" << ex.ele1 << "," << ex.ele2 << ")";
+    Core::IO::cout(Core::IO::debug) << msg.str() << Core::IO::endl;
   }
 
   Core::Communication::barrier(discret->get_comm());
+
+  // done with exchanging old_cp_normals_
 }
 
 /**
--- a/src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_condition.hpp
+++ b/src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_condition.hpp
@@ -64,6 +64,11 @@ namespace BeamInteraction
     std::shared_ptr<BeamInteraction::BeamContactPair> create_contact_pair(
         const std::vector<Core::Elements::Element const*>& ele_ptrs) override;
 
+
+    /**
+     * \brief Setup geometry data and previous calculated quantities.
+     * @param discret (in) discretization.
+     */
     void setup(const std::shared_ptr<const Core::FE::Discretization>& discret) override;
 
    protected:
--- a/src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
+++ b/src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
@@ -464,14 +464,14 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::get_activ
         }
 
         auto& v = *cpvariables_.back();
-        ElementIDKey key{static_cast<int>(element1()->id()), static_cast<int>(element2()->id())};
+        ElementIDKey key{(element1()->id()), (element2()->id())};
 
 
         auto it = old_cp_normals_->find(key);
         if (it != old_cp_normals_->end())
         {
           Core::LinAlg::Matrix<3, 1, TYPE> n_ref(Core::LinAlg::Initialization::zero);
-          for (int i = 0; i < 3; ++i) n_ref(i) = static_cast<TYPE>(it->second(i));
+          for (int i = 0; i < 3; ++i) n_ref(i) = it->second(i);
           v.set_old_normal(n_ref);
         }
       }
@@ -2289,6 +2289,9 @@ bool BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::closest_p
       // double since this factor is needed for a pure scaling of the nonlinear CCP and has not to
       // be linearized!
       double norm_delta_r = Core::FADUtils::vector_norm<3>(delta_r);
+
+      if (norm_delta_r < std::numeric_limits<double>::min())
+        FOUR_C_THROW("Norm of delta_r is too small. Aborting to avoid division by 0.");
 
       // calculate unit normal
       Core::LinAlg::Matrix<3, 1, TYPE> normal(Core::LinAlg::Initialization::zero);
@@ -4914,29 +4917,19 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::compute_n
   TYPE gap = norm_delta_r - r1_ - r2_;
   double sign = 1.0;
 
+
+
   if (params()->beam_to_beam_contact_params()->use_new_gap_function())
   {
-    if (Core::FADUtils::cast_to_double(Core::FADUtils::norm(Core::FADUtils::scalar_product(
-            normal, old_normal))) < std::numeric_limits<double>::epsilon())
-      FOUR_C_THROW("ERROR: Rotation too large! --> Choose smaller Time step!");
-
-    gap =
-        Core::FADUtils::signum(Core::FADUtils::scalar_product(normal, old_normal)) * norm_delta_r -
-        r1_ - r2_;
-
     const double scalar_product_of_normals = Core::FADUtils::scalar_product(normal, old_normal);
-    const double norm_of_normal = Core::FADUtils::cast_to_double(Core::FADUtils::norm(normal));
-    const double norm_of_old_normal =
-        Core::FADUtils::cast_to_double(Core::FADUtils::norm(old_normal));
 
-    if (norm_of_normal > std::numeric_limits<double>::epsilon() &&
-        norm_of_old_normal > std::numeric_limits<double>::epsilon() &&
-        std::abs(Core::FADUtils::cast_to_double(scalar_product_of_normals)) /
-                (norm_of_old_normal * norm_of_normal) >
-            std::cos(M_PI / 4.0))
-    {
-      sign = Core::FADUtils::cast_to_double(Core::FADUtils::signum(scalar_product_of_normals));
-    }
+    if (Core::FADUtils::cast_to_double(Core::FADUtils::norm(scalar_product_of_normals)) <
+        std::numeric_limits<double>::epsilon())
+      FOUR_C_THROW("ERROR: The old normal and the current normal are orthogonal on each other!");
+
+    gap = Core::FADUtils::signum(scalar_product_of_normals) * norm_delta_r - r1_ - r2_;
+
+    sign = Core::FADUtils::cast_to_double(Core::FADUtils::signum(scalar_product_of_normals));
   }
 
   variables->set_gap(gap);


src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_variables.hpp
```diff
@@ -91,15 +91,36 @@ namespace BeamInteraction
...
 
     /*!
-    \brief Set gap
+    \brief Set normal
     */
     void set_normal(Core::LinAlg::Matrix<3, 1, TYPE> normal) { normal_ = normal; };
 
     /*!
-    \brief Get gap
+    \brief Set old normal
```
maxfirmbach, line 99: The documentation of this and the following methods is just a replication of what the member function is called. I think in this case we can skip the description at all.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_condition.cpp
```diff
@@ -67,6 +69,136 @@ void BeamInteraction::BeamToBeamContactCondition::clear()
...
+  // Remove every local key/normals which were distributed to other procs
+  for (auto const& key : local_keys_to_be_removed) old_cp_normals_->erase(key);
+
+  // Print warning on debug if some normals are not found
+  if (!local_missing_keys.empty())
+  {
+    Core::IO::cout(Core::IO::debug)
+        << "[Rank " << Core::Communication::my_mpi_rank(discret->get_comm())
+        << "] is still missing " << local_missing_keys.size()
+        << " CP normals after exchange in BeamToBeamContactCondition. Example key: "
```
maxfirmbach: I think it makes sense to write out "CP" as "contact pair" in debug output.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -39,8 +39,10 @@ FOUR_C_NAMESPACE_OPEN
 
 /*----------------------------------------------------------------------------*
  *----------------------------------------------------------------------------*/
+
```
maxfirmbach: This should not be necessary.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -4814,24 +4862,56 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::compute_n
...
 
+  auto normal_old_ = variables->get_old_normal();
+  // Initialize normal_old_ in the first step with valid closest point projection (in this case the
+  // vector normal_old_ is zero, since no valid normal vector was available in the last time step).
+  // In case of "sliding contact", i.e. when normal_old_ has already been calculated out of the
+  // neighbor element (get_neighbor_normal_old), this is not allowed. Otherwise we would overwrite
+  // the vector normal_old_ which has already been calculated via the neighbor element. (For this
+  // reason we check with the norm of normal_old_ and not with the variable firstcall_).
+  if (Core::FADUtils::cast_to_double(
+          Core::FADUtils::norm(Core::FADUtils::scalar_product(normal_old_, normal_old_))) < 0.00001)
```
maxfirmbach: I'm not so sure if we should compare to this magic number.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -4814,24 +4862,56 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::compute_n
...
   //  if (Core::FADUtils::cast_to_double(gap)<-MAXPENETRATIONSAFETYFAC*(R1_+R2_) and numstep_>0)
-  if (Core::FADUtils::cast_to_double(gap) < -MAXPENETRATIONSAFETYFAC * (r1_ + r2_))
+  if (Core::FADUtils::cast_to_double(gap) < -MAXPENETRATIONSAFETYFAC * (r1_ + r2_) and
+      not params()->beam_to_beam_contact_params()->use_new_gap_function())
   {
     this->print(std::cout);
     FOUR_C_THROW(
-        "Gap too small, danger of penetration. Choose smaller time step or higher penalty!");
+        "Gap too small, danger of penetration. Choose smaller time step or higher penalty!"
+        "The BEAMS_NEWGAP Feature may help to suppress this error message.");
```
maxfirmbach: I think we should not formulate it in this way. Our goal should not be to suppress error messages.

maxfirmbach (changes requested):
@knarfnitram Thanks for porting this nice feature! I left you some initial comments.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -2723,6 +2768,7 @@ bool BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::point_to_
       {
         jacobi = get_jacobi(element2());
       }
+      // TODO: figure out why we overwrite here the jacobi
```
davidrudlstorfer, line 2810: is still a missing TODO for this PR?

davidrudlstorfer (approved):
@knarfnitram thanks a lot for porting this feature - this really brings us closer to simplifying the whole beam stuff and finally removing the old framework!

Is the beams smoothing feature also necessary before removing the old framework? Or can the removal be done after this PR is merged?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
knarfnitram: This is a todo for a different different issue , however I'd like to merge it asap, because this part of the code looks strange to me.
As you can see we calculate the jacobi based on either `element1()` or `element2()` just to overwrite it the next line.
So either we can skip the previous case selection or we should remove the overwriting.

tests/input_files/beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_twocrossedbeams.4C.yaml
```diff
@@ -15,8 +25,8 @@ DISCRETISATION:
 STRUCTURAL DYNAMIC:
   INT_STRATEGY: "Standard"
   DYNAMICTYPE: "GenAlphaLieGroup"
-  TIMESTEP: 0.5
-  NUMSTEP: 2
+  TIMESTEP: 0.1
+  NUMSTEP: 10
```
isteinbrecher: We have to figure out why we now need 5 times as many time steps. Ideally we can pass the test with the exact same number of time steps.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_params.cpp
```diff
@@ -191,10 +192,6 @@ void BeamInteraction::BeamToBeamContactParams::init()
       beam_contact_params_list.get<double>("BEAMS_DAMPREGPARAM2") != -1000.0)
     FOUR_C_THROW("BEAMS_DAMPING currently not supported!");
 
-  /****************************************************************************/
-  if (beam_contact_params_list.get<double>("BEAMS_MAXDISISCALEFAC") != -1.0 or
-      beam_contact_params_list.get<double>("BEAMS_MAXDELTADISSCALEFAC") != -1.0)
-    FOUR_C_THROW("BEAMS_MAXDISISCALEFAC and BEAMS_MAXDELTADISSCALEFAC currently not supported!");
```
isteinbrecher, line 197: Why is this removed?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_params.cpp
```diff
@@ -157,7 +158,7 @@ void BeamInteraction::BeamToBeamContactParams::init()
   // safety checks for currently unsupported parameter settings
   /****************************************************************************/
   if (beam_contact_params_list.get<bool>("BEAMS_NEWGAP"))
-    FOUR_C_THROW("BEAMS_NEWGAP currently not supported!");
+    btb_use_new_gap_function_ = beam_contact_params_list.get<bool>("BEAMS_NEWGAP");
```
isteinbrecher:
```suggestion
btb_use_new_gap_function_ = beam_contact_params_list.get<bool>("BEAMS_NEWGAP");
```
If I read the code correctly, the flag `btb_use_new_gap_function_` should always be set.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.hpp
```diff
@@ -54,7 +66,8 @@ namespace BeamInteraction
     /*!
     \brief Standard Constructor
     */
-    BeamToBeamContactPair();
+    BeamToBeamContactPair(
+        std::shared_ptr<std::map<ElementIDKey, Core::LinAlg::Matrix<3, 1, double>>>);
```
isteinbrecher, line 70: Would it make sense to create a structure that contains `std::map<ElementIDKey, Core::LinAlg::Matrix<3, 1, double>>` and then only reference the structure here? Then it will be a lot easier to extend the feature.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -349,6 +351,22 @@ bool BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::evaluate(
...
+    for (auto& cpvariable : cpvariables_)
+    {
+      // create key and normal
+      ElementIDKey key{(element1()->id()), (element2()->id())};
+      Core::LinAlg::Matrix<3, 1, double> normal;
+      for (int i = 0; i < 3; ++i)
+        normal(i) = Core::FADUtils::cast_to_double(cpvariable->get_normal()(i));
+
+      // save normal for next iteration
+      (*old_cp_normals_)[key] = normal;
```
isteinbrecher, line 369: Shouldn't we save normal*sign here? Otherwise there can be cases with a relatively small penalty parameter where the beam snaps through in the next timestep.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -4814,24 +4861,56 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::compute_n
...
+    if (Core::FADUtils::cast_to_double(Core::FADUtils::norm(Core::FADUtils::scalar_product(
+            normal, normal_old_))) < std::numeric_limits<double>::epsilon())
+      FOUR_C_THROW("ERROR: Rotation too large! --> Choose smaller Time step!");
+
+    gap =
+        Core::FADUtils::signum(Core::FADUtils::scalar_product(normal, normal_old_)) * norm_delta_r -
+        r1_ - r2_;
+
+    sign = Core::FADUtils::cast_to_double(
+        Core::FADUtils::signum(Core::FADUtils::scalar_product(normal, normal_old_)));
```
isteinbrecher: Should we add an error check if the "closest" angle is larger than some threshold value, e.g., $\pi/4$?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_params.cpp
knarfnitram: `BEAMS_MAXDISISCALEFAC` is only needed in the old structure time integration and `BEAM INTERACTION` is in the new one.
`BEAMS_MAXDELTADISSCALEFAC` has no influence at all. 
Therefore I think it's ok to remove them from the `BEAM INTERACTION` section.

tests/input_files/beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_twocrossedbeams.4C.yaml
knarfnitram: With the recent changes, this is done.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.hpp
knarfnitram: I think if we have to add more data, that would make sense yes, but for this pair i think it's alright to stick with that.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -427,6 +448,33 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::get_activ
...
+      if (params()->beam_to_beam_contact_params()->use_new_gap_function() and
+          old_cp_normals_ != nullptr)
+      {
+        if (cpvariables_.size() > 1)
+        {
+          FOUR_C_THROW(
+              "Currently the New Gap function is not supported if the number "
+              "of segments of the current beam contact "
+              "and previous contact change since only 1 normal per element pair is stored.");
+        }
```
knarfnitram, line 464: For the test case `beam3eb_static_contact_penalty_linposquadpen_beamslidingoverarc` the new gap may currently not be used due to this restriction.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_variables.hpp
knarfnitram: I adjusted to the existing documentation. so this could be cleaned up in a different mr.

src/beamcontact/4C_beamcontact_input.cpp
```diff
@@ -43,7 +43,7 @@ std::vector<Core::IO::InputSpec> BeamContact::valid_parameters()
 
           parameter<bool>(
               "BEAMS_NEWGAP", {.description = "choose between original or enhanced gapfunction",
-                                  .default_value = false}),
+                                  .default_value = true}),
```
davidrudlstorfer, line 46: is it anticipated to set the default to true?

knarfnitram: yes(see comments above). So it should always be set to true, once crossing of beams occur, you need to activate it anyway. 

Regarding the names, i collected them already(see #1426) , but i would suggest to do that in a follow up.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_condition.cpp
```diff
@@ -67,6 +69,136 @@ void BeamInteraction::BeamToBeamContactCondition::clear()
...
+  // Print warning on debug if some normals are not found
+  if (!local_missing_keys.empty())
+  {
+    Core::IO::cout(Core::IO::debug)
+        << "[Rank " << Core::Communication::my_mpi_rank(discret->get_comm())
+        << "] is still missing " << local_missing_keys.size()
+        << " contact pair normals after exchange in BeamToBeamContactCondition. Example key: "
+        << Core::IO::endl;
+    auto ex = *local_missing_keys.begin();
+    Core::IO::cout(Core::IO::debug) << "(" << ex.ele1 << "," << ex.ele2 << ")" << Core::IO::endl;
```
isteinbrecher: I would put this in a single `Core::IO::cout` call, especially in parallel, that helps to avoid procs printing over each other (not all the time).

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_condition.hpp
```diff
@@ -63,6 +64,8 @@ namespace BeamInteraction
     std::shared_ptr<BeamInteraction::BeamContactPair> create_contact_pair(
         const std::vector<Core::Elements::Element const*>& ele_ptrs) override;
 
+    void setup(const std::shared_ptr<const Core::FE::Discretization>& discret) override;
```
isteinbrecher, line 72: Missing documentation.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_condition.cpp
```diff
@@ -67,6 +69,136 @@ void BeamInteraction::BeamToBeamContactCondition::clear()
   condition_contact_pairs_.clear();
 }
 
+/**
+ *
+ */
+void BeamInteraction::BeamToBeamContactCondition::setup(
+    const std::shared_ptr<const Core::FE::Discretization>& discret)
+{
+  // hard coded number of pair size and length of normal
```
isteinbrecher, line 82: Can we add a comment here, or in the function definition, what is actually happening, i.e., that we distribute the normal definitions of the old time step to the other procs here.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -2240,8 +2288,41 @@ bool BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::closest_p
...
       // double since this factor is needed for a pure scaling of the nonlinear CCP and has not to
       // be linearized!
-      double norm_delta_r = Core::FADUtils::cast_to_double(Core::FADUtils::vector_norm<3>(delta_r));
-      gap = norm_delta_r - r1_ - r2_;
+      double norm_delta_r = Core::FADUtils::vector_norm<3>(delta_r);
+
+      // calculate unit normal
+      Core::LinAlg::Matrix<3, 1, TYPE> normal(Core::LinAlg::Initialization::zero);
+      Core::LinAlg::Matrix<3, 1, TYPE> normal_old(Core::LinAlg::Initialization::zero);
+      normal.update(1.0 / norm_delta_r, delta_r, 0.0);
```
isteinbrecher, line 2299: Should we add a check if the norm is 0 or close to 0?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -427,6 +448,33 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::get_activ
...
+
+        auto& v = *cpvariables_.back();
+        ElementIDKey key{static_cast<int>(element1()->id()), static_cast<int>(element2()->id())};
+
+
+        auto it = old_cp_normals_->find(key);
+        if (it != old_cp_normals_->end())
+        {
+          Core::LinAlg::Matrix<3, 1, TYPE> n_ref(Core::LinAlg::Initialization::zero);
+          for (int i = 0; i < 3; ++i) n_ref(i) = static_cast<TYPE>(it->second(i));
```
isteinbrecher: Why do we need a static cast here?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -4814,24 +4897,66 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::compute_n
...
+
+
   TYPE gap = norm_delta_r - r1_ - r2_;
+  double sign = 1.0;
+
+  if (params()->beam_to_beam_contact_params()->use_new_gap_function())
+  {
+    if (Core::FADUtils::cast_to_double(Core::FADUtils::norm(Core::FADUtils::scalar_product(
+            normal, old_normal))) < std::numeric_limits<double>::epsilon())
+      FOUR_C_THROW("ERROR: Rotation too large! --> Choose smaller Time step!");
```
isteinbrecher:
```suggestion
      FOUR_C_THROW("ERROR: The old normal and the current normal are orthogonal on each other");
```

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -4814,24 +4897,66 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::compute_n
...
+        Core::FADUtils::signum(Core::FADUtils::scalar_product(normal, old_normal)) * norm_delta_r -
+        r1_ - r2_;
+
+    const double scalar_product_of_normals = Core::FADUtils::scalar_product(normal, old_normal);
+    const double norm_of_normal = Core::FADUtils::cast_to_double(Core::FADUtils::norm(normal));
+    const double norm_of_old_normal =
+        Core::FADUtils::cast_to_double(Core::FADUtils::norm(old_normal));
+
+    if (norm_of_normal > std::numeric_limits<double>::epsilon() &&
+        norm_of_old_normal > std::numeric_limits<double>::epsilon() &&
```
isteinbrecher: Either of these cases should already be covered by the scalar product above.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
```diff
@@ -4814,24 +4897,66 @@ void BeamInteraction::BeamToBeamContactPair<numnodes, numnodalvalues>::compute_n
...
+        std::abs(Core::FADUtils::cast_to_double(scalar_product_of_normals)) /
+                (norm_of_old_normal * norm_of_normal) >
+            std::cos(M_PI / 4.0))
```
isteinbrecher: What happens in the `else` case? This check only makes sense if we error out if the old and new normal are not within a certain angle of each other.

tests/input_files/beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_twocrossedbeams.4C.yaml
isteinbrecher, line 1: So we didn't have to adapt the tilmestep here?

mayrmt:
Given @davidrudlstorfer  proposal to rename the input parameter: yes, we probably should. "New" does not convey a great deal of information. As far as I know, the essential feature of the "new" gap is to detect / prevent crossing of beams. Shouldn't this information be part of the input parameter name? `BEAMS_GAP_TO_PREVENT_CROSSING` or something similar? Does the parameter name really have to start with `BEAMS_`?

We could get rid of the input parameter by making this the only gap function (basically removing the non-new variant). Not sure, if this is desirable from an application point of view.

knarfnitram:
> Given @davidrudlstorfer proposal to rename the input parameter: yes, we probably should. "New" does not convey a great deal of information. As far as I know, the essential feature of the "new" gap is to detect / prevent crossing of beams. Shouldn't this information be part of the input parameter name? `BEAMS_GAP_TO_PREVENT_CROSSING` or something similar? Does the parameter name really have to start with `BEAMS_`?
> 
> We could get rid of the input parameter by making this the only gap function (basically removing the non-new variant). Not sure, if this is desirable from an application point of view.

I think it's best to first port the existing functionality, remove then the old beam framework and finally rename it. Otherwise we spent again time fixing code or adjusting code comments in parts which are hopefully deleted soon.
If you want to express your opinion on that single parameter feel free to do it in the respective issue -> https://github.com/4C-multiphysics/4C/issues/1426, you can also find there the current suggestions. 
Also if we start to remove the `BEAMS_` suffix from parameters, I think it would be good to remove it consistently for 
the whole `BEAM CONTACT` parameters section (fex: `BEAMS_STRATEGY`, `BEAMS_SEGANGLE`...) at once.

`BEAMS_NEWGAP` is not yet working for all cases  (see `beam3eb_static_contact_penalty_linposquadpen_beamslidingoverarc.4C.yaml`). So I think removing the input parameter completely is currently out of scope.

isteinbrecher:
@knarfnitram why is `beam3eb_static_contact_penalty_linposquadpen_beamslidingoverarc.4C.yaml` not yet working with the new new gap?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
knarfnitram: @isteinbrecher  see here.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
knarfnitram: I removed the check entirely, since the `signum` will take care of it anyway.

tests/input_files/beam3r_line2_genalpha_liegroup_contact_penalty_linpen_newgap_twocrossedbeams.4C.yaml
knarfnitram: not any more, since I also use the enhanced gap calculations within the closest point projection.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
knarfnitram: You are right, can be removed. Thanks!

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
isteinbrecher: Is new gap required for this test case, or does it also pass without it?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
isteinbrecher: So we don’t have the angle check any more? (Fine with me)

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_pair.cpp
knarfnitram: No, it's not required.

PR #1460, opened 2025-10-30
[LinAlg] Guard `Epetra` checks with `FOUR_C_ENABLE_ASSERTIONS`

maxfirmbach:
## Description and Context

For some people speed matters, especially when executing a lot of simulations. Therefore, this PR proposes to move all `Epetra` error check calls into `FOUR_C_ENABLE_ASSERTIONS`. In the end the user decides: Do I just want to execute the code and be "fast" or do I want to develop and have assertions at a ton of places.

In addition a lot of method definitions are moved into the respective `.cpp` file.

Part of #1190

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_fevector.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_fevector.hpp
@@ -14,8 +14,8 @@
 #include "4C_linalg_map.hpp"
 #include "4C_linalg_multi_vector.hpp"
 #include "4C_linalg_transfer.hpp"
+#include "4C_linalg_utils_exceptions.hpp"
 #include "4C_linalg_view.hpp"
-#include "4C_utils_epetra_exceptions.hpp"
 
 #include <Epetra_FEVector.h>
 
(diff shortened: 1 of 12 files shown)


mayrmt (commented):
@maxfirmbach I like the proposal. The current implementation introduces a lot of `#ifdef` statements. Do you know, if one can incorporate the decision of debug mode inside `CHECK_EPETRA_CALL()` instead of having to do `#ifdef` on the outside?

maxfirmbach:
@mayrmt I thought of something similar but decided against it. Yes, this PR adds a lot of boiler plate code, but it makes everything explicit in writing it out. This is how we currently differ between the two builds. I'm not sure if it is a good idea to hide it away, because then we should do it for all occurrences (also the ones not related to `Epetra`).

mayrmt:
@maxfirmbach Sure! I do not have a strong opinion on that. I just thought, that if integrated into `CHECK_EPETRA_CALL()` t then one does not have to remember to add `#ifdef`s in the future. Since I do not expect to many additions of Epetra features in our code, that's probably not a great argument.

I am fine with the current approach.

src/core/linalg/src/sparse/4C_linalg_fevector.hpp
knarfnitram, line 1: Is there a specific reason why we do not enable or disable the FOUR_C_ENABLE_ASSERTIONS for `replace_global_value` and `multiply` and `reciprocal` ?

knarfnitram (commented):
I'm also in favor of reducing the boiler plate and move the assertion into `CHECK_EPETRA_CALL`. How about renaming (or introducing)  `CHECK_EPETRA_CALL_ON_ENABLED_ASSERTION`. I think, with this it would be clear, when the check is done.

src/core/linalg/src/sparse/4C_linalg_fevector.hpp
maxfirmbach: @knarfnitram I most likely have overseen them :smile:.

maxfirmbach:
@sebproell Do you have an opinion how we should implement the proposal stated by this PR?

sebproell:
@maxfirmbach I can take a look next week

sebproell:
@maxfirmbach I would not introduce the #ifdefs explicitly. I would change this https://github.com/4C-multiphysics/4C/blob/3962668aebabbb236fd8d8c338295cce5291dc5b/src/core/utils/src/exceptions/4C_utils_epetra_exceptions.hpp#L27-L28 definition to use `FOUR_C_ASSERT` and possibly rename the macro to `ASSERT_EPETRA_CALL`.


```c++
    [[maybe_unused]] int err = (expr);                                               \
    FOUR_C_ASSERT(err == 0, "Epetra error (code {}).", err); \
```

I also noticed that this header is in the wrong place. Can you move it to core/linalg?

vovannikov:
I also vote for removing the explicit #ifdefs and wrapping them inside `CHECK_EPETRA_CALL` or another similar macro. The main reason is that the same call effectively has to be written twice in the current proposal and it is easy for some of them to get out-of-sync, especially for those having many parameters, so it gets more demanding from maintenance point of view.

mayrmt (approved):

Issue #1462, opened 2025-10-31
[LinAlg] Prepare MultiVector class for Tpetra


mayrmt:
## Description

Sub issue from https://github.com/4C-multiphysics/4C/issues/875 to prepare the class `Core::LinAlg::MultiVector` for the switch to Tpetra.

## Possible Solution and Definition of Done

- [ ] Do all necessary steps as outlined for `Core::LinAlg::Vector` in #1121 
- [ ] Replace `operator(int i)` to access the `i`th vector in the MultiVector by the Tpetra-conforming `get_vector(const int i)`. (Note: Tpetra internally uses `size_t` instead of `int`, but that is not the core issue here. Replacing the access operator with an access methods is at the core of this task). Reference to Teptra: https://docs.trilinos.org/dev/packages/tpetra/doc/html/classTpetra_1_1MultiVector.html#a7951dcc07f3da968cb2b9b9a4d5fdcdc

mayrmt:
@maxfirmbach, I assume that PR #1519 completed the open item "Refactor Epetra_CombineMode and use Core::LinAlg::CombineMode instead", right? Can we mark its checkbox as completed?

maxfirmbach:
@mayrmt I forgot to adapt it in the multi vector. Maybe @dharinib98 can take care of it. It is a rather quick and simple cleanup task.

maxfirmbach:
@dharinib98 Please make sure to keep this issue up-to-date if something has merged or new issues arise. Thanks!

PR #1465, opened 2025-11-01
[LinAlg] Clean up of the MultiVector Interface

knarfnitram:
## Description and Context

Clean up the interface by removing the implicit conversion from `Epetra_MultiVector` to the interface. 
Separates the functions into the source file and remove some functionalities which will not be available in Tpetra. 

See https://github.com/4C-multiphysics/4C/issues/1462

### DIFF
--- a/src/scatra/4C_scatra_timint_implicit_service.cpp
+++ b/src/scatra/4C_scatra_timint_implicit_service.cpp
@@ -119,7 +119,7 @@ std::shared_ptr<Core::LinAlg::MultiVector<double>> ScaTra::ScaTraTimIntImpl::cal
   if (calcflux_domain_lumped_)
   {
     // vector for integrated shape functions
-    std::shared_ptr<Core::LinAlg::Vector<double>> integratedshapefcts =
+    std::shared_ptr<Core::LinAlg::Vector<double>> integrated_shape_functions =
         Core::LinAlg::create_vector(dofrowmap);
 
     // overwrite action for elements
@@ -130,16 +130,43 @@ std::shared_ptr<Core::LinAlg::MultiVector<double>> ScaTra::ScaTraTimIntImpl::cal
     auto dofids = std::make_shared<Core::LinAlg::IntSerialDenseVector>(num_dof_per_node());
     dofids->putScalar(1);  // integrate shape functions for all dofs
     params.set("dofids", dofids);
-    discret_->evaluate(params, nullptr, nullptr, integratedshapefcts, nullptr, nullptr);
+    discret_->evaluate(params, nullptr, nullptr, integrated_shape_functions, nullptr, nullptr);
 
     // insert values into final flux vector for visualization
     // We do not solve a global, linear system of equations (exact L2 projection with good
     // consistency), but perform mass matrix lumping, i.e., we divide by the values of the
     // integrated shape functions
-    // allocate a temp with same map/shape as *flux
-    Core::LinAlg::MultiVector<double> inv_flux(flux->get_map(), flux->num_vectors());
-    inv_flux.reciprocal(flux->get_epetra_multi_vector());
-    flux_projected->multiply(1.0, integratedshapefcts->as_multi_vector(), inv_flux, 0.0);
+
+    // the mass matrix lumping was previously conducted via the
+    // ReciprocalMultiply. This realized a Hadamard product of two MultiVectors:
+    // this = ScalarThis * this + ScalarAB * B / A (Important: element-wise)
+
+    // start by creating the inverse
+    Core::LinAlg::MultiVector<double> inverse_of_integrated_shape_functions(
+        integrated_shape_functions->get_map(), 1, true);
+    {
+      const double* values = integrated_shape_functions->as_multi_vector().get_values();
+
+      // perform a zero-safe reciprocal of the local vector
+      for (int i = 0; i < integrated_shape_functions->local_length(); ++i)
+        inverse_of_integrated_shape_functions.get_values()[i] =
+            (std::abs(values[i]) > std::numeric_limits<double>::epsilon())
+                ? (1.0 / values[i])
+                : std::numeric_limits<double>::max();
+    }
+
+    // Broadcast inverse_of_integrated_shape_functions across column to match the flux
+    Core::LinAlg::MultiVector<double> inverse_of_integrated_shape_functions_bcast(
+        inverse_of_integrated_shape_functions.get_map(), flux->num_vectors(), false);
+    for (int j = 0; j < flux->num_vectors(); ++j)
+    {
+      // copy column 0 into j-th column
+      inverse_of_integrated_shape_functions_bcast(j).update(
+          1.0, inverse_of_integrated_shape_functions(0), 0.0);
+    }
+
+    // now apply the elementwise-multiplication
+    flux_projected->multiply(1.0, *flux, inverse_of_integrated_shape_functions_bcast, 0.0);
   }
 
   else
@@ -1974,7 +2001,9 @@ void ScaTra::ScaTraTimIntImpl::evaluate_error_compared_to_analytical_sol()
       {
         const int errorfunctnumber = params_->get<int>("CALCERRORNO");
         if (errorfunctnumber < 1)
-          FOUR_C_THROW("invalid value of parameter CALCERRORNO for error function evaluation!");
+          FOUR_C_THROW(
+              "Invalid value of parameter CALCERRORNO for error "
+              "function evaluation!");
 
         eleparams.set<int>("error function number", errorfunctnumber);
       }


src/core/linalg/src/sparse/4C_linalg_multi_vector.cpp
```diff
@@ -97,32 +86,32 @@ void Core::LinAlg::MultiVector<T>::scale(double ScalarValue)
...
 {
-  CHECK_EPETRA_CALL(vector_->Dot(A, Result));
+  CHECK_EPETRA_CALL(vector_->Dot(A.get_epetra_multi_vector(), Result));
 }
 
 template <typename T>
 void Core::LinAlg::MultiVector<T>::abs(const MultiVector& A)
 {
-  CHECK_EPETRA_CALL(vector_->Abs(A));
+  CHECK_EPETRA_CALL(vector_->Abs(*A.vector_));
```
maxfirmbach: Why are we using the member variable here directly? Instead of e.g. `A.get_epetra_XXX()`.

src/scatra/4C_scatra_timint_implicit_service.cpp
```diff
@@ -51,7 +51,8 @@ void ScaTra::ScaTraTimIntImpl::calc_flux(const bool writetofile)
 
     default:
     {
-      FOUR_C_THROW("Invalid option for flux calculation inside domain!");
+      FOUR_C_THROW(
+          "inverse_of_integrated_shape_functionslid option for flux calculation inside domain!");
```
c-p-schmidt: Was the change of the error messages in this file intentional?

src/scatra/4C_scatra_timint_implicit_service.cpp
```diff
@@ -130,13 +132,43 @@ std::shared_ptr<Core::LinAlg::MultiVector<double>> ScaTra::ScaTraTimIntImpl::cal
...
+    Core::LinAlg::MultiVector<double> inverse_of_integrated_shape_functions(
+        integrated_shape_functions->get_map(), 1, false);
+    {
+      const double* values = integrated_shape_functions->as_multi_vector().get_values();
+
+      // perform a zero-safe reciprocal of the local vector
+      for (int i = 0; i < integrated_shape_functions->local_length(); ++i)
+        inverse_of_integrated_shape_functions.get_values()[i] =
+            (std::abs(values[i]) > std::numeric_limits<double>::epsilon()) ? (1.0 / values[i])
+                                                                           : 0.0;
```
c-p-schmidt: Why did you put a 0? It should prevent a division by 0, right? So wouldn't std::numeric_limits<double>::max() or similar be better?

src/core/linalg/src/sparse/4C_linalg_multi_vector.hpp
```diff
@@ -70,12 +65,6 @@ namespace Core::LinAlg
     //! Compute Inf-norm of each vector
     void norm_inf(double* Result) const;
 
-    //! Compute minimum value of each vector
-    void min_value(double* Result) const;
-
-    //! Compute maximum value of each vector
-    void max_value(double* Result) const;
-
```
vovannikov, line 78: Why are these methods removed?

src/core/linalg/src/sparse/4C_linalg_multi_vector.hpp
```diff
@@ -52,11 +52,6 @@ namespace Core::LinAlg
 
     ~MultiVector() = default;
 
-    // (Implicit) conversions: they all return references, never copies
-    operator Epetra_MultiVector&() { return *vector_; }
-
-    operator const Epetra_MultiVector&() const { return *vector_; }
-
```
vovannikov, line 59: What is the general point of removing these implicit conversions? Similar conversions exist also in `Core::LinAlg::Vector` and how I understood them, they should automatically go away once we get rid of Epetra. I also prefer explicit interfaces, but all the `get_epetra_multi_vector()` calls should also ultimately get removed.

src/scatra/4C_scatra_timint_implicit_service.cpp
```diff
@@ -130,13 +132,43 @@ std::shared_ptr<Core::LinAlg::MultiVector<double>> ScaTra::ScaTraTimIntImpl::cal
...
     params.set("dofids", dofids);
-    discret_->evaluate(params, nullptr, nullptr, integratedshapefcts, nullptr, nullptr);
+    discret_->evaluate(params, nullptr, nullptr, integrated_shape_functions, nullptr, nullptr);
 
     // insert values into final flux vector for visualization
     // We do not solve a global, linear system of equations (exact L2 projection with good
     // consistency), but perform mass matrix lumping, i.e., we divide by the values of the
     // integrated shape functions
-    flux_projected->reciprocal_multiply(1., *integratedshapefcts, *flux, 0.);
+    // allocate a temp with same map/shape as *flux
```
vovannikov: I think these two lines of comments do not agree with each other well.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linearsystem.cpp
```diff
@@ -429,7 +429,7 @@ void NOX::Nln::LinearSystem::adjust_pseudo_time_step(double& delta, const double
   vec_2.scale(stepSizeInv);
   vec_1->update(1.0, vec_2, 1.0);
   /* evaluate the second vector:              d^{T} V                   */
-  vec_2.multiply(1.0, scalingDiagOp, dir.getEpetraVector(), 0.0);
+  vec_2.multiply(1.0, scalingDiagOp, dir.get_linalg_vector(), 0.0);
```
vovannikov, line 418: This has been taken care of in #1466.

src/solver_nonlin_nox/4C_solver_nonlin_nox_solver_ptc.cpp
```diff
@@ -951,7 +951,7 @@ void NOX::Nln::GROUP::PrePostOp::PseudoTransient::run_post_compute_f(
   Teuchos::RCP<NOX::Nln::Vector> fUpdate = eval_pseudo_transient_f_update(grp);
 
   // add the transient part
-  F.update(1.0, fUpdate->getEpetraVector(), 1.0);
+  F.update(1.0, fUpdate->get_linalg_vector(), 1.0);
```
vovannikov: This has been taken care of in #1466.

src/solver_nonlin_nox/4C_solver_nonlin_nox_solver_ptc.cpp
```diff
@@ -980,7 +980,7 @@ void NOX::Nln::GROUP::PrePostOp::PseudoTransient::run_pre_compute_f(
   Teuchos::RCP<NOX::Nln::Vector> fUpdate = eval_pseudo_transient_f_update(grp);
 
   // subtract the transient part
-  F.update(-1.0, fUpdate->getEpetraVector(), 1.0);
+  F.update(-1.0, fUpdate->get_linalg_vector(), 1.0);
```
vovannikov: This has been taken care of in #1466.

src/structure_new/src/4C_structure_new_integrator.cpp
```diff
@@ -288,7 +288,7 @@ void Solid::Integrator::compute_mass_matrix_and_init_acc()
   nox_soln.scale(-1.0);
 
   // get the solution vector and add it into the acceleration vector
-  accnp_ptr->update(1.0, nox_soln.getEpetraVector(), 1.0);
+  accnp_ptr->update(1.0, nox_soln.get_linalg_vector(), 1.0);
```
vovannikov, line 291: This has been taken care of in #1466.

src/core/linalg/src/sparse/4C_linalg_multi_vector.hpp
vovannikov: I found the answer in #1462 :)

src/core/linalg/src/sparse/4C_linalg_multi_vector.hpp
knarfnitram: Before the switch can happen we need to be able to easily identify all occurrences to epetra interface to be sure not to miss one. The implicit conversions potentially hides occurrences away, since you can create code passages, where you do not know, if you have actually the underlying object or the wrapper object. 
`get_epetra_multi_vector`-> will probably become `get_tpetra_multi_vector` for that cases where we actually need direct access to the object (this depends on the interface and how you use it, if we do not have to fully explore it any more even better, than we'll can remove it entirely). 
And yes `Core::LinAlg::Vector` is using them too, so I'll probably try to remove (or at least reduce) the usages of them too.

src/scatra/4C_scatra_timint_implicit_service.cpp
knarfnitram: that was not on purpose, thanks for catching that.

src/scatra/4C_scatra_timint_implicit_service.cpp
knarfnitram: Yes, that's better!

maxfirmbach:
I think now it is looking really good!

PR #1467, opened 2025-11-03
Make tutorial directories self-contained

c-p-schmidt:
## Description and Context
Today, @lkoeglmeier, while preparing introductory material to 4C for a student, brought up the issue that our tutorial input files in the tutorial folders are not runnable out of the box. The tests are only runnable in the pipeline, as the testing functionality copies relevant files to the respective testing directories.

However, this is challenging for newcomers to 4C (e.g., students or individuals who get in touch with 4C for the first time and explore the tutorials) who are unfamiliar with the intricate testing infrastructure.

Since we also do not want to double up the files, which are often default solver settings files, introducing a symbolic link is an elegant solution to achieve this, IMO. This way, the respective files are in the repository only once, but the tests are still directly executable without needing to determine where to copy them from.

That the symbolic links are working is even tested, since the paths where to copy the files from during the ctest run are adapted accordingly.


lkoeglmeier:
Thank you for this change!

PR #1468, opened 2025-11-03
Remove write_vector call of old output mechanism in Solid::ModelEvaluator::Structure::output_step_state

lkoeglmeier:
## Description and Context

This pr removes the old output call in` Solid::ModelEvaluator::Structure`. New vtk-based output is already written here.

part of #211

### DIFF
--- a/src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
+++ b/src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
@@ -1703,16 +1703,6 @@ void Solid::ModelEvaluator::Structure::determine_energy(const Core::LinAlg::Vect
 
 /*----------------------------------------------------------------------------*
  *----------------------------------------------------------------------------*/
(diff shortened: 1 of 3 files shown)


src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
```diff
@@ -1378,10 +1378,6 @@ void Solid::ModelEvaluator::Structure::write_restart(
   // write forces
   iowriter.write_vector("fstructure_old", global_state().get_fstructure_old());
   iowriter.write_vector("fint", global_state().get_fint_n());
-
-  if (forced_writerestart) return;
-
-  iowriter.write_vector("displacement", global_state().get_dis_n());
```
c-p-schmidt, line 1384: Maybe this line needs to stay and replace the other call in `src/structure_new/src/4C_structure_new_timint_base.cpp`, even though I would prefer it in the other call.

PR #1469, opened 2025-11-03
Remove  function that adds restart data to old output

lkoeglmeier:
## Description and Context

In the old output mechanism, the output and restart were written using HDF5 output. When writing the output and restart at the same time step, we only had to add the additional data required for the restart to the old output. With the new vtk-based output, the old output is no longer written and only the restart uses HDF5 output. Therefore, we have to write all restart data every time we restart. So, this function (adding restart data to the old output) is no longer useful.

part of #211


lkoeglmeier:
When looking into this in more detail it turned out (at least as far as I understand it) that it is not straight forward to delete this function as we still have to call the old output mechanism for the the contact which is not yet migrated to the new vtk-based output. Deleting this function leads to problems when writing restart output at the end of a simulation used in tests with adaptive time stepping. As we want to migrate contact anyway in the very near future and therefore can delete this function very easily we decided to close this pr for now and wait until the contact is migrated to not waste time into an interim solution that we delete again soon anyway.
Maybe interesting for you @rjoussen.

PR #1472, opened 2025-11-04
Condense metadata lists

sebproell:
Following #1103, this PR also condenses duplicated lists and better detects when groups/lists would appear often.

### DIFF
--- a/src/core/io/src/4C_io_input_spec_builders.cpp
+++ b/src/core/io/src/4C_io_input_spec_builders.cpp
@@ -175,6 +175,22 @@ namespace
     ryml::id_type num_children_before;
   };
 
+  /**
+   * Only allow referencing for specs that contain other specs. There is little point in referencing
+   * a single parameter spec.
+   */
+  [[nodiscard]] bool can_be_referenced(const Core::IO::InputSpec& spec)
+  {
+    switch (spec.impl().data.type)
+    {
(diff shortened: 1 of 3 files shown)


src/core/io/src/4C_io_input_spec_builders.cpp
```diff
@@ -183,9 +199,14 @@ namespace
 void Core::IO::Internal::emit_metadata_helper(const Core::IO::InputSpec& spec,
     Core::IO::YamlNodeRef node, Core::IO::Internal::EmitMetadataContext& context)
 {
-  if (spec.use_count() > context.options.condense_duplicated_specs_threshold &&
-      !context.currently_emitting_to_references &&
-      spec.impl().data.type == Core::IO::Internal::InputSpecType::group)
+  // We do not know how often the spec is used, but this is the minimum: the use count of the
+  // parent plus the uses of the other references to this spec minus the current one.
+  const unsigned current_min_use_count = context.current_use_count + spec.use_count() - 1;
```
sebproell, line 204: This new minimum use_count estimate is useful to use references also for `all_of` specs that are reused. I would like to only reference something that has a name, i.e. groups and lists, and not all_ofs. Instead, we now track when an all_of/one_of is used often and then reference all its nested groups. The nested groups will just have a use_count of 1 in the shared_ptr sense, but they will appear as often as the parent is duplicated.

PR #1474, opened 2025-11-04
Add sym. HA-product

tuchpaul:
Implemented the symmetric Holzapfel product using tensors and the Einstein Summation Convention.
An exemplary change was made in src/solid_3D_ele/4C_solid_3D_ele_surface_traceestimate.cpp to demonstrate the usage.

Note: The product is also used in plasticity-related code, @dragos-ana, but I did not modify those parts to avoid interfering with your material model.

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.hpp
@@ -216,6 +216,25 @@ namespace Core::LinAlg::FourTensorOperations
   void add_symmetric_holzapfel_product(Core::LinAlg::Matrix<6, 6>& X,
       const Core::LinAlg::Matrix<3, 3>& A, const Core::LinAlg::Matrix<3, 3>& B, const double fac);
 
+  /*!
+   * @brief Calculates symmetric Holzapfel product
+   *
+   * In tensor index notation this does:
+   * \f[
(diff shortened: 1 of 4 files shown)


src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.hpp
```diff
@@ -216,6 +216,26 @@ namespace Core::LinAlg::FourTensorOperations
   void add_symmetric_holzapfel_product(Core::LinAlg::Matrix<6, 6>& X,
       const Core::LinAlg::Matrix<3, 3>& A, const Core::LinAlg::Matrix<3, 3>& B, const double fac);
 
+  /*!
+   * @brief Calculates symmetric Holzapfel product
+   *
+   * In tensor index notation this does:
+   * \f[
+   *    X_{abcd} += \text{fac} \cdot \left( A_{ca} \cdot B_{db} + A_{da} \cdot B_{cb} + A_{db}
```
c-p-schmidt: You are not adding the contributions, but returning the result. If it is to be added, it has to be done by the user outside.
```suggestion
   *    X_{abcd} = \text{fac} \cdot \left( A_{ca} \cdot B_{db} + A_{da} \cdot B_{cb} + A_{db}
```

tuchpaul: Thanks for the hint! I adjusted the comment.

c-p-schmidt (approved):

PR #1477, opened 2025-11-04
Clean up singletons on uncaught exception

### LINKED ISSUE #1476: [BUG] controlfile does not write last result step

I noticed that the control file seems to show restart data only once the next time step was successful: This leads to the standard output showing, for example:
`====== Restart for field 'Structure' written in step 8`
then the next step is attempted and fails. However, the control file shows as its last entry:
```
- field:
    field: "structure"
    time: 8.799999999999999
    step: 8
    num_nd: 3375
    max_nodeid: 3374
    num_ele: 330
    num_dof: 10125
    num_dim: 3
    num_output_proc: 2
    mesh_file: "output.mesh.structure.s0"
```
but no result entry for step 8. 

## Expectations

I would expect that in the controlfile there is another result entry at the end. I checked and indeed I can add the result entry manually and the restart from step 8 (in this example) works fine. So restart info is written in binary format. 

## Your Environment

- **Relevant repo SHA1s:**
- **Relevant configure flags or configure script:**
- **Operating system and version:**
- **Compiler and TPL versions:**

- Blocks
- Follows
- Related to
- Part of
- Composed of

sebproell:
Might fix #1476 

@rjoussen

### DIFF
--- a/apps/global_full/4C_global_full_main.cpp
+++ b/apps/global_full/4C_global_full_main.cpp
@@ -428,6 +428,10 @@ int main(int argc, char* argv[])
         Core::Communication::barrier(arguments.comms.global_comm());
       }
 
+      // Calling MPI_Abort will terminate all processes in the communicator without cleaning
+      // up singletons etc. So we finalize the singleton registry here to ensure proper cleanup.
+      Core::Utils::SingletonOwnerRegistry::finalize();
+
       MPI_Abort(MPI_COMM_WORLD, EXIT_FAILURE);
     }
 #endif


sebproell:
@rjoussen Could you give this a try and see if it fixes the problem in #1476?

rjoussen:
That did not fix it unfortunately

sebproell:
This is not helpful. In general, it seems impossible to notify other MPI ranks of a failure on one rank at an arbitrary point in time.

PR #1481, opened 2025-11-05
Fix the pvd file writer for restarts from parent directory

rjoussen:
## Description and Context

The current `create_restarted_initial_collection_file_mid_section` had the following problem:
Suppose I have the folder structure:
```
parent_folder
|- folder1
|- folder2
```
`folder1` contains the restart files and the pvd collection. If I navigate to `folder2` and do a restart there with a relative part to `folder1`, i.e. `4C input output_folder2 restartfrom=../folder1/output_folder1` everything worked fine so far (thanks to @knarfnitram), the pvd collection would show entries like
```xml
file="../folder1/output_folder1/output-vtk-files/structure-00000.pvtu"
```

However, doing the same thing in the parent directory via `4C input folder2/output_folder2 restartfrom=folder1/output_folder1`, which leads to exactly the same results, leads to pvd file entries like
```xml
file="folder1/output_folder1/output-vtk-files/structure-00000.pvtu"
```
which gets continuously worse the more often I restart.

This PR fixes that by computing the relative path from the `path_existing_working_directory_` to `p_restart.parent_path()`.

#1138

### DIFF
--- a/src/core/io/src/4C_io_vtk_writer_base.cpp
+++ b/src/core/io/src/4C_io_vtk_writer_base.cpp
@@ -184,6 +184,7 @@ VtkWriterBase::VtkWriterBase(unsigned int myrank, unsigned int num_processors,
     const std::string& restart_name, double restart_time, bool write_binary_output,
     LibB64::CompressionLevel compression_level)
     : currentPhase_(VAGUE),
+      path_existing_working_directory_(path_existing_working_directory),
       num_timestep_digits_(LibB64::ndigits(max_number_timesteps_to_be_written)),
       num_processor_digits_(LibB64::ndigits(num_processors)),
       geometry_name_(geometry_name),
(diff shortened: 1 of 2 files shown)


knarfnitram (approved):
@rjoussen , thanks for the explanation and fixing the path!
The changes look good to me!
I think it would be good to cover this restart within a test case (check out the function `four_c_test_restarted_vtk`). So I think all we need there is to conduct a third run  based on your suggested relative restart (just adapt the paths and compare it to the same vtk files). 
However, most likely we should refactor `four_c_test_restarted_vtk` for that a little bit. So I'd propose to merge these changes now, but can you please open an issue stating how the `four_c_test_restarted_vtk` should be adapted to cover the restart?

rjoussen:
@knarfnitram I will look into it. I am very much not proficient at cmake though, so it might take a bit.

PR #1482, opened 2025-11-06
Add tests for relative restarts with vtk output

rjoussen:
## Description and Context

This refactors the `four_c_test_restarted_vtk` to accept the additional `RESTARTFROM_PATHTYPE` argument, which is either
- `absolute:` as before,
- `relative:` simulation is started inside the new folder where the restart results are written to, `restartfrom` is relative to the folder with the restart data (so one folder up, one folder down to the previous run). This was already working before #1481 but not tested so far,
- or `relative_from_parent` (simulation is started from the parent directory (= the test directory), `restartfrom` is relative from the parent directory into the folder with the restart data)

All three options are tested, validation happens indirectly via vtk compare, which only works for valid pvd files.

This tests the changes from #1481
@knarfnitram


tests/list_of_tests.cmake
```diff
@@ -1565,7 +1565,7 @@ four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 500)
 four_c_test(TEST_FILE red_airway_2_acinus_of_3Exponential_alveolar_ducts.4C.yaml NP 2 RETURN_AS current)
 four_c_test_restart(BASED_ON ${current} SAME_FILE NP 2 RESTART_STEP 599)
 four_c_test(TEST_FILE red_airway_3airway_2acinus_awacinter.4C.yaml NP 2)
-four_c_test(TEST_FILE red_airway_5000_airway_acinus_collapsible.4C.yaml NP 2 TIMEOUT 180)
+four_c_test(TEST_FILE red_airway_5000_airway_acinus_collapsible.4C.yaml NP 2 TIMEOUT 240)
 four_c_test(TEST_FILE red_airway_acinus_inter_acinar.4C.yaml RETURN_AS current)
```
knarfnitram, line 1569: @bgoderbauer any objections against increasing the timeout?

knarfnitram (approved):
Changes looks good to me, thanks for providing the test cases!

tests/list_of_tests.cmake
bgoderbauer: No, that's totally fine. We will delete all `red_airway` tests once their functionality is fully ported to `reduced_lung`.

PR #1484, opened 2025-11-07
Flush control_file correctly for pure structure problems

rjoussen:
## Description and Context

For pure structure problems, we can call `control_file().end_group()` at the end of `output_restart()`, since all results should have been written at that point and registered in the control file's result group.

I think it would make sense to call this also in `add_restart_to_output_state()`, but since #1469 tries to remove that anyway, I did not do that. 

For coupled problem types, this is not the case, and hence will not work. In these cases, `end_group()` needs to be called somewhere higher up.

part of #1476

### DIFF
--- a/src/structure_new/src/4C_structure_new_timint_base.cpp
+++ b/src/structure_new/src/4C_structure_new_timint_base.cpp
@@ -17,6 +17,7 @@
 #include "4C_io_control.hpp"
 #include "4C_io_gmsh.hpp"
 #include "4C_io_pstream.hpp"
+#include "4C_legacy_enum_definitions_problem_type.hpp"
 #include "4C_linalg_map.hpp"
 #include "4C_linalg_vector.hpp"
 #include "4C_structure_new_dbc.hpp"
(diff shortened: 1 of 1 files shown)


mayrmt:
@rjoussen This PR just addresses "structure" problems. What about other problem types? Does the bug described in #1476 occur there as well?

rjoussen:
@mayrmt yes, #1476 should, as far as I know, apply to all problem types. The problem is that the only good solution to #1476 that we came up with so far is - unfortunately - problem-type dependent. This PR attempts to solve that for pure structure problems only.

src/structure_new/src/4C_structure_new_timint_base.cpp
```diff
@@ -662,7 +663,13 @@ void Solid::TimeInt::Base::output_restart(bool& datawritten)
...
    * evaluators. */
   int_ptr_->write_restart(*output_ptr);
-
+  // For pure structure problems, all results should be written now,
+  // hence close the control file group to force flushing.
+  // only try closing group, as no group was opened for non-binary output
+  if (Global::Problem::instance()->get_problem_type() == Core::ProblemType::structure)
+  {
+    output_ptr->output().control_file().try_end_group();
+  }
```
sebproell, line 672: This implementation uses a ton of internal knowledge as can be seen from the chained function calls. As an alternative, how about creating a function like `end_step()` in the `DiscretizationWriter` and call that?

rjoussen: That's a good idea, actually. I will try that. I do think, though, that even this version will still fail, as there are some explicit integrators that want to append to the same result group across multiple time-step, since the restart also needs info across multiple time-steps.

rjoussen:
Unfortunately, this approach does not fix #1476 and instead causes numerous other problems (as can be seen by the failed tests.

PR #1485, opened 2025-11-07
Refactor mortar manager parameters

isteinbrecher:
## Description and Context

Improve the handling of the mortar manager parameters in beaminteraction. With this PR a lot of functionality not directly related to the mortar manager can be moved out of the class and into the function that creates the mortar manager.
This will enable a more versatile use of the mortar manager in the future.

### DIFF
--- a/src/beaminteraction/src/4C_beaminteraction_beam_to_solid_conditions.cpp
+++ b/src/beaminteraction/src/4C_beaminteraction_beam_to_solid_conditions.cpp
@@ -10,6 +10,7 @@
 #include "4C_beam3_euler_bernoulli.hpp"
 #include "4C_beam3_reissner.hpp"
 #include "4C_beaminteraction_beam_to_solid_edge_contact_pair.hpp"
+#include "4C_beaminteraction_beam_to_solid_mortar_manager.hpp"
 #include "4C_beaminteraction_beam_to_solid_mortar_manager_contact.hpp"
 #include "4C_beaminteraction_beam_to_solid_surface_contact_pair.hpp"
 #include "4C_beaminteraction_beam_to_solid_surface_contact_pair_mortar.hpp"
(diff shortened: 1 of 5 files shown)


src/beaminteraction/src/4C_beaminteraction_beam_to_solid_conditions.cpp
```diff
@@ -130,20 +131,102 @@ BeamInteraction::BeamToSolidCondition::create_indirect_assembly_manager(
...
+
+        FOUR_C_ASSERT_ALWAYS(beam_to_volume_params || beam_to_surface_params,
+            "Rotational coupling is only implemented for beam to solid volume meshtying and beam "
+            "to solid surface meshtying.");
+        if (beam_to_volume_params != nullptr)
+        {
+          mortar_shape_function_rotation =
+              beam_to_volume_params->get_mortar_shape_function_rotation_type();
+        }
+        else
```
dharinib98:
```suggestion
        else if (beam_to_surface_params != nullptr)
```

maybe this is better ?

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
dharinib98, line 1: #include "4C_beaminteraction_beam_to_solid_surface_contact_params.hpp"
#include "4C_beaminteraction_beam_to_solid_surface_meshtying_params.hpp"
#include "4C_beaminteraction_beam_to_solid_volume_meshtying_params.hpp"

We do not need these anymore here.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_conditions.cpp
```diff
@@ -130,20 +131,102 @@ BeamInteraction::BeamToSolidCondition::create_indirect_assembly_manager(
...
+            beam_to_solid_params_);
+    const auto beam_to_surface_contact_params =
+        std::dynamic_pointer_cast<const BeamInteraction::BeamToSolidSurfaceContactParams>(
+            beam_to_solid_params_);
+
+    // Get the parameters for the mortar manager.
+    MortarManagerParameters mortar_manager_params{};
+
+    {
+      //
```
maxfirmbach, line 148: Are these additional `//` really necessary? This applies to all occurrences.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_conditions.cpp
maxfirmbach: I agree, it would make things a bit clearer, despite the fact that it is most likely not necessary. Below it is actually done like proposed by @dharinib98.

maxfirmbach (commented):
Looks good so far! I just have some very minor comments.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_conditions.cpp
isteinbrecher: They are just for improved spacing between non related code blocks.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
isteinbrecher: Good catch, I removed them.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_conditions.cpp
isteinbrecher: Yes, that makes sense, I adapted it.

PR #1487, opened 2025-11-10
Remove determine_stress_strain() in preparation of output and adapt flag WRITE_FINAL_STATE

lkoeglmeier:
see commit message


rjoussen (approved):
Looks good to me, I was confused by that also when looking at how they write to the control file. Isn't there a parameter `WRITE_FINAL_STATE` also? If there is and that's what is turned on by default for the adaptive integrator it might make sense to state that explicitly in the input file

c-p-schmidt:
> Looks good to me, I was confused by that also when looking at how they write to the control file. Isn't there a parameter `WRITE_FINAL_STATE` also? If there is and that's what is turned on by default for the adaptive integrator it might make sense to state that explicitly in the input file

You are right, that parameter is available in general. However, it does not use this parameter; instead, it always writes the restart output after the maximum number of steps or when the maximum time is reached.

Maybe a (rather) clean solution would be to remove this and instead set the `WRITE_FINAL_STATE` parameter to `true`. If it seems relevant for such problems, one could also force the user of the adaptive integration to set this flag. However, I would be fine either way.

c-p-schmidt (approved):

lkoeglmeier:
Yes, we also have the parameter `WRITE_FINAL_STATE` to do the same (write output and restart data at the last time step of the simulation). This is, however, only possible in pure solid problems (only implemented in `dyn_nlnstructural_drt(`) so far). When using adaptive time stepping it is "hard" coded to always write the output and restart data at the end of the time loop in `Adapter::StructureTimeAda::integrate()`. But I am not an expert on which problem types we write the restart at the end. I only came across these tests when removing the old output so I am not super sure what comment we could add here a`t RESTARTEVERY` , or do you have a suggestion or more insights ?

lkoeglmeier:
> 
> Maybe a (rather) clean solution would be to remove this and instead set the `WRITE_FINAL_STATE` parameter to `true`. If it seems relevant for such problems, one could also force the user of the adaptive integration to set this flag. However, I would be fine either way.


I like the idea of removing the special capability in adaptive time stepping and instead set the `WRITE_FINAL_STATE` parameter to `true` in these tests. I will change it I think this will be clearer and cleaner.

src/adapter/4C_adapter_str_timeada.cpp
```diff
@@ -309,8 +309,7 @@ int Adapter::StructureTimeAda::integrate()
     }
   }
 
-  // force write output
-  output(true);
+
```
davidrudlstorfer: very minor - but I guess the white space changes are not necessary

davidrudlstorfer (dismissed):

src/structure/4C_structure_dyn_nln_drt.cpp
```diff
@@ -140,13 +140,10 @@ void dyn_nlnstructural_drt()
   // run time integration
   structadapter->integrate();
 
-  if (write_final_state && !structadapter->has_final_state_been_written())
+  if (write_final_state)
   {
     constexpr bool forceWriteRestart = true;
-    constexpr bool force_prepare = true;
-    structadapter->prepare_output(force_prepare);
```
c-p-schmidt, line 147: For documentation: This prepare_output needs to be deleted, as update has already been called at the end of `structadapter->integrate()`, and the time is already after the MAXTIME of the simulation. Additionally, potentially relevant history data may have been destroyed by the update call.

Thus, in general, we are not able to reconstruct a reasonable stress or strain state from this.

c-p-schmidt (approved):

src/structure/4C_structure_dyn_nln_drt.cpp
lkoeglmeier: I included your explanation in the commit message.

c-p-schmidt (approved):

rjoussen (approved):

PR #1495, opened 2025-11-12
Simplify handling of Lagrange multiplier flag

isteinbrecher:
## Description and Context

Improve handling of Lagrange multipliers in beaminteraction by not relying on the beam-to-solid volume meshtying parameters but taking this value out of the mortar manager directly.

### DIFF
--- a/src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_beamcontact.cpp
+++ b/src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_beamcontact.cpp
@@ -246,6 +246,28 @@ void BeamInteraction::SubmodelEvaluator::BeamContact::setup()
 
 /*----------------------------------------------------------------------*
  *----------------------------------------------------------------------*/
+bool BeamInteraction::SubmodelEvaluator::BeamContact::have_lagrange_dofs() const
+{
+  if (assembly_managers_.size() == 0) return false;
+
+  const auto indirect_assembly_manager =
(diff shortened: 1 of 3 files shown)


src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_beamcontact.hpp
```diff
@@ -148,6 +148,11 @@ namespace BeamInteraction
         return assembly_managers_;
       }
 
+      /**
+       * \brief Return flag if this submodel evaluatior caries lagrange multipliers.
```
sebproell:
```suggestion
       * \brief Return flag if this submodel evaluator carries Lagrange multipliers.
```

src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_beamcontact.cpp
```diff
@@ -244,6 +244,29 @@ void BeamInteraction::SubmodelEvaluator::BeamContact::setup()
   issetup_ = true;
 }
 
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+bool BeamInteraction::SubmodelEvaluator::BeamContact::have_lagrange_dofs() const
+{
+  if (assembly_managers_.size() > 0)
```
sebproell:
```suggestion
  if (assembly_managers_.empty()) return false;
```

Minor: use early return.

dharinib98 (approved):

c-p-schmidt (approved):

PR #1499, opened 2025-11-13
Cleanup particle surface energy

ppraegla:
Clean up of the particle surface energy calculation

Breaking change:
- Rename `ADHESION_SURFACE_ENERGY_DISTRIBUTION_VAR` to `ADHESION_SURFACE_ENERGY_DISTRIBUTION_STDDEV`
- Remove option `LogNormalSurfaceEnergyDistribution` from `ADHESION_SURFACE_ENERGY_DISTRIBUTION`
- Rename options for `ADHESION_SURFACE_ENERGY_DISTRIBUTION`:
  - `ConstantSurfaceEnergy` to `Constant`
  - `NormalSurfaceEnergyDistribution` to `Normal`

Also, the old surface energy handler is removed and replaced with a function that calculates the surface energy based on a switch-case. This makes the code easier to follow and saves some virtual function calls.


src/particle/src/interaction/4C_particle_interaction_dem_adhesion_surface_energy.hpp
```diff
@@ -8,111 +8,37 @@
...
-    //! get adhesion surface energy
-    void adhesion_surface_energy(
-        const double& mean_surface_energy, double& surface_energy) const override
-    {
-      surface_energy = mean_surface_energy;
-    };
+    SurfaceEnergyDistribution type;
+    double surface_energy;
+    double distribution_standard_deviation;
+    double distribution_cutoff_factor;
```
sebproell: I would document these. I assume the distribution parameters are only relevant if the `type` is chosen a certain way. Maybe they should also be` optional<double>` , see my other comment.

src/particle/src/interaction/4C_particle_interaction_dem_adhesion_surface_energy.hpp
```diff
@@ -8,111 +8,37 @@
...
-    //! get adhesion surface energy
-    void adhesion_surface_energy(
-        const double& mean_surface_energy, double& surface_energy) const override;
-  };
+  /*!
+   * Calculates the surface energy based on the given SurfaceEnergyDistribution type in params and
+   * the given mean_surface_energy value.
+   *
+   * \note the surface energy from DEMAdhesionSurfaceEnergyParams is not used because the function
+   * is also used for the ParticleWall
```
sebproell: Is there something missing? Why is the ParticleWall related to this function? Not clear to non-experts ;)

src/particle/src/4C_particle_input.cpp
```diff
@@ -538,13 +538,13 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
           parameter<double>("ADHESION_SURFACE_ENERGY",
               {.description = "adhesion surface energy for the calculation of the pull-out force",
                   .default_value = -1.0}),
-          parameter<double>("ADHESION_SURFACE_ENERGY_DISTRIBUTION_VAR",
-              {.description = "variance of adhesion surface energy distribution",
+          parameter<double>("ADHESION_SURFACE_ENERGY_DISTRIBUTION_STDDEV",
+              {.description = "standard deviation of adhesion surface energy distribution",
                   .default_value = -1.0}),
```
sebproell, line 543: Instead of having a dummy default, you could make this an `optional<double>`.

src/particle/src/4C_particle_input.hpp
```diff
@@ -235,8 +235,7 @@ namespace Particle
   enum SurfaceEnergyDistribution
```
sebproell:
```suggestion
  enum class SurfaceEnergyDistribution
```

and drop the suffix on the values.

src/particle/src/interaction/4C_particle_interaction_dem_adhesion_surface_energy.hpp
ppraegla: It is used here as well  https://github.com/4C-multiphysics/4C/blob/main/src/particle/src/interaction/4C_particle_interaction_dem_adhesion.cpp#L417.

The surface energy value from the particle wall can be different from the particles. But the calculation is the same: return a constant value or a value drawn from a normal distribution. 

Maybe it does not make sense to pass the params struct. I could split the struct into a struct that only contains the parameters needed for the distribution `DEMAdhesionSurfaceEnergyDistributionParams` and a struct that contains the general adhesion parameters.

src/particle/src/4C_particle_input.cpp
ppraegla: Yes. I will do that in another PR. I need to restructure the input anyway to use the new features.

src/particle/src/interaction/4C_particle_interaction_dem_adhesion_surface_energy.hpp
ppraegla: I will do that together with the input restructuring

sebproell (approved):

davidrudlstorfer (approved):

src/particle/src/interaction/4C_particle_interaction_dem_adhesion_surface_energy.hpp
ppraegla: I split the params and introduced a new adhesion_params struct. With further restructuring I will add more parameters to this struct in a future PR.

davidrudlstorfer (approved):

PR #1502, opened 2025-11-17
Point coupling with Kirchhoff

isteinbrecher:
## Description and Context
Allow point couplings to work with Kirchhoff beams.


src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.cpp
```diff
@@ -29,366 +30,323 @@ FOUR_C_NAMESPACE_OPEN
...
  *
  */
-template <typename Beam>
-void BeamInteraction::BeamToBeamPointCouplingPair<Beam>::setup()
+template <typename Beam1, unsigned int n_dof_beam_1, typename Beam2, unsigned int n_dof_beam_2>
+void BeamInteraction::BeamToBeamPointCouplingPair<Beam1, n_dof_beam_1, Beam2, n_dof_beam_2>::setup()
 {
-  // This pair only works for Simo Reissner beam elements.
-  const auto check_simo_reissner_beam = [](auto element)
+  beam_elements_ = {this->element1(), this->element2()};
```
davidrudlstorfer, line 36: Is this now also working with Euler Bernoulli elements? Should this check be added?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.cpp
```diff
@@ -29,366 +30,323 @@ FOUR_C_NAMESPACE_OPEN
...
-      {
-        // No projection was found
-        return;
-      }
+  // Print some general information: Element IDs and dofvecs.
+  out << "\n------------------------------------------------------------------------";
+  out << "\nInstance of BeamToBeamPenaltyPointCouplingPair"
+      << "\nBeam1 EleGID:  " << element1()->id() << "\nBeam2 EleGID: " << element2()->id();
+  out << "------------------------------------------------------------------------\n";
+}
```
davidrudlstorfer, line 133: Would it make sense to only print this with a certain verbosity level?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.cpp
```diff
@@ -511,55 +464,220 @@ BeamInteraction::BeamToBeamPointCouplingPair<Beam>::evaluate_rotational_coupling
...
+    const std::shared_ptr<GeometryPair::LineToLineEvaluationData>& line_to_line_evaluation_data)
 {
-  using namespace GeometryPair;
+  const auto get_element_info =
+      [&](size_t i_beam) -> std::tuple<bool, Core::FE::CellType, unsigned int>
+  {
+    const auto* reissner_beam =
+        dynamic_cast<const Discret::Elements::Beam3r*>(element_ptrs[i_beam]);
+    const auto* kirchhoff_beam =
+        dynamic_cast<const Discret::Elements::Beam3k*>(element_ptrs[i_beam]);
```
davidrudlstorfer, line 603: Same here - what happens for an Euler Bernoulli element?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.cpp
```diff
@@ -511,55 +464,220 @@ BeamInteraction::BeamToBeamPointCouplingPair<Beam>::evaluate_rotational_coupling
...
+            "with Hermite shape functions");
+      }
+      is_hermite = true;
+      n_dof = 15;
+    }
+    else
+    {
+      FOUR_C_THROW(
+          "The beam_to_beam_point_coupling_pair_factory only works for Simo Reissner or "
+          "Kirchhoff beam elements");
```
davidrudlstorfer, line 641: So this get's caught here? Would it make sense to just throw that in the beginning if its an Euler-Bernoulli element?

tests/input_files/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_elbow.xml
```diff
@@ -27,7 +27,7 @@
         <ParameterList name="Quantity 0">
           <Parameter name="Quantity Type" type="string" value="Structure"/>
           <Parameter name="Tolerance Type" type="string" value="Absolute"/>
-          <Parameter name="Tolerance" type="double" value="1e-08"/>
+          <Parameter name="Tolerance" type="double" value="1e-12"/>
```
davidrudlstorfer, line 30: Why have these tolerances decreased?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.cpp
isteinbrecher: I would leave this check up to the factory method. The pair now works in a way where it does not need to know which type of beam formulation is actually used.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.cpp
isteinbrecher: I think that is already done by the calling function, I am not sure though.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.cpp
isteinbrecher: I would keep the current way, as we only support SR and KL at the moment. If a different one is given we will error out anyways, so I would not add an additional check for non-supported formulations.

tests/input_files/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_elbow.xml
isteinbrecher: To be hones, I just saw that the tests also pass with that tolerance and I usually set the tolerance to `1e-12` so I adapted this here.

PR #1503, opened 2025-11-17
Write vtk based output in contact modelevaluator

c-p-schmidt:
## Description and Context
As part of #211, the old output of the model evaluator contact is removed and replaced by a vtk-based version.

A test is added to test the new vtk-based output and all artifacts, e.g. post-processing tests have been removed.

#part of #211


tests/list_of_tests.cmake
```diff
@@ -2780,6 +2777,7 @@ four_c_test_vtk(beam3r_herm2line3_static_LJ_singlelengthspec_smallsepapprox_simp
 four_c_test_vtk(beam3r_herm2line3_static_LJ_singlelengthspec_smallsepapprox_simple_regularization_linextpol_twocrossedbeams_pulloff_vtp_potential.4C.yaml beam3r_herm2line3_static_LJ_singlelengthspec_smallsepapprox_simple_regularization_linextpol_twocrossedbeams_pulloff.4C.yaml 2 xxx-beam-potential.pvd ref/beam3r_herm2line3_static_LJ_singlelengthspec_smallsepapprox_simple_regularization_linextpol_twocrossedbeams_pulloff-beam-potential.pvd 1e-08)
 four_c_test_vtk(beam3r_line2_singlefil_periodic_uniaxial_shear_vtk_restart_binning.4C.yaml beam3r_line2_singlefil_periodic_uniaxial_shear.4C.yaml 3 xxx-binning.pvd ref/beam3r_line2_singlefil_periodic_uniaxial_shear-vtk/binning.pvd 1e-08 0.0)
 four_c_test_vtk(beam3r_line2_singlefil_periodic_uniaxial_shear_vtk_restart_structure_beams.4C.yaml beam3r_line2_singlefil_periodic_uniaxial_shear.4C.yaml 3 xxx-structure-beams.pvd ref/beam3r_line2_singlefil_periodic_uniaxial_shear-vtk/structure-beams.pvd 1e-08 0.0)
+four_c_test_vtk(contact3D_slidingblock_coulomb_new_struct_vtk_struc_contact.4C.yaml contact3D_slidingblock_coulomb_new_struct.4C.yaml 2 xxx-structure-contact.pvd ref/contact3D_slidingblock_coulomb_new_struct.pvd 1e-08 2.0)
```
isteinbrecher:
```suggestion
four_c_test_vtk(contact3D_slidingblock_coulomb_new_struct_vtk_struc_contact.4C.yaml contact3D_slidingblock_coulomb_new_struct.4C.yaml 2 xxx-structure-contact.pvd ref/contact3D_slidingblock_coulomb_new_struct_vtk_struc_contact.pvd 1e-08 2.0)
```
Should we name the pvd file the same as the test file?

tests/list_of_tests.cmake
```diff
@@ -422,7 +420,6 @@ four_c_test(TEST_FILE contact3D_patch_linstatic.4C.yaml NP 2)
 four_c_test(TEST_FILE contact3D_patch_linstatic_new_struct.4C.yaml NP 2)
 four_c_test(TEST_FILE contact3D_patch_unbiased_self_nitsche_skew.4C.yaml)
 four_c_test(TEST_FILE contact3D_pwlin_penalty_hex20.4C.yaml NP 2)
-four_c_test_post_processing(contact3D_pwlin_penalty_hex20.4C.yaml 2 2 ndxyz ndxyz 10)
 four_c_test(TEST_FILE contact3D_pwlin_penalty_hex27.4C.yaml NP 2)
 four_c_test_post_processing(contact3D_pwlin_penalty_hex27.4C.yaml 2 2 ndxyz ndxyz 10)
```
isteinbrecher, line 424: Is there a reason this old test is still here?

isteinbrecher: Also, out of curiosity, do you know what `ndxyz` is? Ist this relevant for the "new" output?

tests/input_files/contact3D_slidingblock_coulomb_new_struct.4C.yaml
```diff
@@ -13,6 +13,14 @@ DISCRETISATION:
...
 IO:
   STRUCT_STRESS: "2PK"
+  STDOUTEVERY: 0
+IO/RUNTIME VTK OUTPUT:
+  INTERVAL_STEPS: 1
+  OUTPUT_DATA_FORMAT: ascii
+IO/RUNTIME VTK OUTPUT/STRUCTURE:
+  OUTPUT_STRUCTURE: true
+  DISPLACEMENT: true
+  STRESS_STRAIN: true
```
isteinbrecher, line 23: Can we control if/how/what contact output is written in the input file?

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_contact.cpp
```diff
@@ -89,6 +86,18 @@ void Solid::ModelEvaluator::Contact::setup()
 
   post_setup(cparams);
 
+  if (global_in_output().get_runtime_output_params() != nullptr)
```
isteinbrecher: Does this mean we write contact output if we write structure output? I think this should be controllable via the input file.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_contact.cpp
```diff
@@ -460,109 +469,134 @@ void Solid::ModelEvaluator::Contact::determine_energy() {}
...
     activeset.update(1.0, slipsetexp, 1.0);
   }
 
   // export to problem node row map
-  std::shared_ptr<const Core::LinAlg::Map> problemnodes = strategy().problem_nodes();
-  std::shared_ptr<Core::LinAlg::Vector<double>> activesetexp =
-      std::make_shared<Core::LinAlg::Vector<double>>(*problemnodes);
-  Core::LinAlg::export_to(activeset, *activesetexp);
+  auto problemnodes = strategy().problem_nodes();
+  auto activesetexp = Core::LinAlg::Vector<double>(*problemnodes, true);
```
isteinbrecher, line 507: I think the active set can be output as `int`. Also for some other quantities, e.g., slip state, master active nodes, ...

tests/list_of_tests.cmake
c-p-schmidt: In contrast to the removed test, this test is from the old structure time integration. Since this has not been migrated to the vtk-based output, this post-processing test remains. So the naming of the removed test was inconsistent with the usual naming scheme, where tests of the new structure time integration usually include a `_new_struct`.

As far as I know, this is a post-processing node-based approach, whereas cxyz is a post-processing element-based approach. Please don't ask why these names were chosen so intuitively. ;-)

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_contact.cpp
c-p-schmidt: This means that we instantiate the runtime writer object if an `IO/RUNTIME VTK OUTPUT` is defined.
So you would prefer it to be fine-grained, i.e., using a sublist: e.g., `IO/RUNTIME VTK OUTPUT/CONTACT`?

c-p-schmidt: What about the individual outputs like `activeset` etc.. Should they have a toggle to switch the output on and off, like `DISPLACEMENT` for the structure vtk-based output?

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_contact.cpp
c-p-schmidt: I agree, but the problem is that the EpetraInt Vector does not support e.g. update calls.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_contact.cpp
```diff
@@ -460,109 +469,116 @@ void Solid::ModelEvaluator::Contact::determine_energy() {}
 
 /*----------------------------------------------------------------------*
  *----------------------------------------------------------------------*/
-void Solid::ModelEvaluator::Contact::output_step_state(
-    Core::IO::DiscretizationWriter& iowriter) const
+void Solid::ModelEvaluator::Contact::runtime_pre_output_step_state() { determine_stress_strain(); }
```
eulovi, line 475: Do you need this function? I am not sure if this is being used anywhere else.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_contact.cpp
isteinbrecher: Thats true. Maybe we can leave a comment with a todo here once we support true `int` vectors.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_contact.cpp
isteinbrecher: I am not that deep into contact output, for the sake of this PR, I think it would be sufficient to simply have the ability to run it on and off.

tests/list_of_tests.cmake
isteinbrecher: Thanks for the explanation!

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_contact.cpp
c-p-schmidt: It is called from the time integrator and also implemented in other model evaluators.

However, we can rename the method as soon as the old output version is gone, as we no longer need to distinguish between old and vtk-based output then.

c-p-schmidt:
> Thanks for the contribution @c-p-schmidt! I have a general question: after this PR, is it necessary to still adapt the other contact outputs listed in #211? Namely:
> 
>     * CONTACT::Interface::postprocess_quantities
> 
>     * CONTACT::Manager::postprocess_quantities
> 
>     * CONTACT::MtManager::postprocess_quantities

If I have not overseen something, these are only called from the old-time integration. Thus, we won't migrate them.

tests/input_files/contact3D_slidingblock_coulomb_new_struct.4C.yaml
c-p-schmidt: I introduced a flag `OUTPUT_CONTACT` similar to how it is done for optional quantities.

If people who actually work with it want it more fine-grained, it can easily be introduced in a follow-up.

lkoeglmeier:
> > Thanks for the contribution @c-p-schmidt! I have a general question: after this PR, is it necessary to still adapt the other contact outputs listed in #211? Namely:
> > ```
> > * CONTACT::Interface::postprocess_quantities
> > 
> > * CONTACT::Manager::postprocess_quantities
> > 
> > * CONTACT::MtManager::postprocess_quantities
> > ```
> 
> If I have not overseen something, these are only called from the old-time integration. Thus, we won't migrate them.

Nice! Then we can update the list in #211, right?

c-p-schmidt:
> > > Thanks for the contribution @c-p-schmidt! I have a general question: after this PR, is it necessary to still adapt the other contact outputs listed in #211? Namely:
> > > ```
> > > * CONTACT::Interface::postprocess_quantities
> > > 
> > > * CONTACT::Manager::postprocess_quantities
> > > 
> > > * CONTACT::MtManager::postprocess_quantities
> > > ```
> > 
> > 
> > If I have not overseen something, these are only called from the old-time integration. Thus, we won't migrate them.
> 
> Nice! Then we can update the list in #211, right?

You are right. I just updated the description accordingly.

PR #1506, opened 2025-11-18
Add option to include bodyforce contributions in the fluid field in porofluid_pressure_based* problem types

lkoeglmeier:
This pr adds the option to include body force contributions, for example gravity, in the fluid equations of the porofluid_pressure_based framework. 
The diffusive term in the governing equation (mass balance equation) is now extended by the body force contribution (in red) $- \nabla \cdot (\frac{\mathbf{k}}{\mu} (\nabla p   \textcolor{red}{   - \rho \mathbf{b}}  ))$


src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm.cpp
```diff
@@ -337,6 +348,10 @@ void PoroPressureBased::PorofluidAlgorithm::set_element_general_parameters() con
   eleparams.set<bool>("using stationary formulation", false);
   eleparams.set<double>("alpha_F", 1.0);
 
+  eleparams.set<bool>("has_bodyforce_contribution", bodyforce_contribution_);
```
c-p-schmidt: If you would add the `has_` similar to what you do in the parameter list, it might be a bit more expressive in the sense that a boolean is to be expected from that variable?
```suggestion
  eleparams.set<bool>("has_bodyforce_contribution", has_bodyforce_contribution_);
```

src/porofluid_pressure_based/4C_porofluid_pressure_based_input.cpp
```diff
@@ -151,6 +151,19 @@ std::vector<Core::IO::InputSpec> PoroPressureBased::valid_parameters_porofluid()
               },
               {.required = false}),
 
+          // include bodyforce
+          group("include_bodyforce",
+              {
+                  parameter<bool>("active",
+                      {.description = "Activate a body force contribution to all fluid fields.",
+                          .default_value = false}),
+                  parameter<std::vector<double>>("value",
```
c-p-schmidt: Why not directly?
```suggestion
                  parameter<std::vector<double>>("vector",
```

src/porofluid_pressure_based/4C_porofluid_pressure_based_input.cpp
```diff
@@ -151,6 +151,19 @@ std::vector<Core::IO::InputSpec> PoroPressureBased::valid_parameters_porofluid()
...
+                      {.description = "External body force contribution vector. The vector should "
```
c-p-schmidt:
```suggestion
                      {.description = "External body force contribution vector. The vector must "
```

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
```diff
@@ -3409,6 +3446,184 @@ void Discret::Elements::PoroFluidEvaluator::EvaluatorDomainIntegrals<nsd,
...
+    const PoroFluidManager::PhaseManagerInterface& phasemanager,
+    const PoroFluidManager::VariableManagerInterface<nsd, nen>& variablemanager, double rhsfac,
+    double fac, bool inittimederiv)
+{
+  // get vector to fill
+  Core::LinAlg::SerialDenseVector& myvec = *elevec[0];
+
+  // body force acceleration
+  Core::LinAlg::Matrix<nsd, 1> bodyforce(Core::LinAlg::Initialization::zero);
+  bodyforce.clear();
```
c-p-schmidt: This call does not have an effect, as you always initialize with zeros the line above. But it doesn't harm either.
```suggestion
```

lkoeglmeier:
> I did not find the "-" in the formulation you wrote in the description. But probably I just missed it.
> 
> Besides this, only minor comments.

Thanks for your review @c-p-schmidt! I included all your suggestions. 
Regarding the "-": you are right, you do not see it here because the `vector` is the RHS vector (nonlinear residual) and is always scaled with `-`1 in the evaluator.

src/porofluid_pressure_based/4C_porofluid_pressure_based_input.cpp
```diff
@@ -151,6 +151,19 @@ std::vector<Core::IO::InputSpec> PoroPressureBased::valid_parameters_porofluid()
...
+              {
+                  parameter<bool>("active",
+                      {.description = "Activate a body force contribution to all fluid fields.",
+                          .default_value = false}),
+                  parameter<std::vector<double>>("vector",
+                      {.description = "External body force contribution vector. The vector must "
+                                      "have the same dimension as the problem (2D/3D).",
+                          .default_value = std::vector<double>{}}),
+              },
+              {.required = false}),
```
amgebauer: You could also just call the parameter `body_force` and make it an `optional<vector<double>>.` Then, you don't have to specify an invalid default value for vector.

I think this should express the intent quite well:

```yaml
body_force: none
```
and
```yaml
body_force: [1.0, 0.0, 0.0]
```

vs 

```yaml
include_bodyforce:
  active: false
```
and
```yaml
include_bodyforce:
  active: false
  vector: [1.0, 0.0, 0.0]
```

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
```diff
@@ -3409,6 +3446,182 @@ void Discret::Elements::PoroFluidEvaluator::EvaluatorDomainIntegrals<nsd,
   // nothing to do
 }
 
+/*----------------------------------------------------------------------*
+ * **********************************************************************
+ *----------------------------------------------------------------------*/
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
```
amgebauer: remove? At least a few of them 🤔  Also at the other places.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
```diff
@@ -3409,6 +3446,182 @@ void Discret::Elements::PoroFluidEvaluator::EvaluatorDomainIntegrals<nsd,
...
+    const Core::LinAlg::Matrix<nsd, nen>& xyze, int curphase, int phasetoadd, int numdofpernode,
+    const PoroFluidManager::PhaseManagerInterface& phasemanager,
+    const PoroFluidManager::VariableManagerInterface<nsd, nen>& variablemanager, double rhsfac,
+    double fac, bool inittimederiv)
+{
+  // get vector to fill
+  Core::LinAlg::SerialDenseVector& myvec = *elevec[0];
+
+  // body force acceleration
+  Core::LinAlg::Matrix<nsd, 1> bodyforce(Core::LinAlg::Initialization::zero);
```
amgebauer: I recommend to use a tensor here. This should be a lot easier than using the matrix. To me it looks like that it can be easily be done in this separated case. Of course, you can also do it later for the whole framework at some point.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
```diff
@@ -3409,6 +3446,182 @@ void Discret::Elements::PoroFluidEvaluator::EvaluatorDomainIntegrals<nsd,
...
+
+  // current pressure gradient
+  Core::LinAlg::Matrix<nsd, 1> gradpres(Core::LinAlg::Initialization::zero);
+  gradpres.clear();
+  // compute the pressure gradient from the phi gradients
+  for (int idof = 0; idof < numfluidphases; ++idof)
+    gradpres.update(phasemanager.pressure_deriv(curphase, idof), gradphi[idof], 1.0);
+  double abspressgrad = 0.0;
+  for (int i = 0; i < nsd; i++) abspressgrad += gradpres(i) * gradpres(i);
+  abspressgrad = sqrt(abspressgrad);
```
amgebauer:
```suggestion
  double abspressgrad = gradpres.norm2();
```
There is a convenience function for that (it might not be the exact name that I have in mind, but I think such a function is already implemented)

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.hpp
```diff
@@ -2354,6 +2354,270 @@ namespace Discret
...
+            int curphase,       //!< index of current phase
+            int phasetoadd,     //!< index of current phase
+            int numdofpernode,  //!< total number of DOFs/phases
+            const PoroFluidManager::PhaseManagerInterface& phasemanager,  //!< phase manager
+            const PoroFluidManager::VariableManagerInterface<nsd, nen>&
+                variablemanager,  //!< variable manager
+            double timefacfac,    //!< domain-integration factor
+            double fac,           //!< domain-integration factor times time-integration factor
+            bool inittimederiv    //!< calculate only parts for initial time derivative
+            ) override {};
```
amgebauer: You don't have to repeat the documentation every time (doxygen automatically picks them up from the base class)

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
```diff
@@ -3409,6 +3446,182 @@ void Discret::Elements::PoroFluidEvaluator::EvaluatorDomainIntegrals<nsd,
...
+    const PoroFluidManager::VariableManagerInterface<nsd, nen>& variablemanager, double rhsfac,
+    double fac, bool inittimederiv)
+{
+  // get vector to fill
+  Core::LinAlg::SerialDenseVector& myvec = *elevec[0];
+
+  // body force acceleration
+  Core::LinAlg::Matrix<nsd, 1> bodyforce(Core::LinAlg::Initialization::zero);
+  for (int idim = 0; idim < nsd; idim++)
+    bodyforce(idim) = phasemanager.bodyforce_contribution_values()[idim];
```
amgebauer: You might also make a free function out of it and reuse it below. There are maybe some more parts that you can reuse between the three implementations.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
lkoeglmeier: Thanks for your help with the tensors!

lkoeglmeier:
> Nice! I have a few comments.
> 
> You added 2 tests, but three cases (EvaluatorBodyforce, EvaluatorVolFracHomogenizedVasculatureTumorBodyforce, EvaluatorVolFracBloodLungBodyforce). Are still all new implementations covered? Does it make sense to add a 3D test case as well?
> 
> I'm not familiar with the underlying physics, so excuse any nonsense comments

Thank you for your review @amgebauer!
Yes, all three new implementations are tested. There is more than one fluid phase in each test. However, it is not possible to have additional porous networks with both closure relations in the same problem. So I needed two test cases to have everything covered.
I also added a 3D test now. Thank you for your helpful suggestions.

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
```diff
@@ -1540,7 +1573,7 @@ void Discret::Elements::PoroFluidEvaluator::EvaluatorDiff<nsd,
 
   // gradient of pressure w.r.t. reference coordinates
   Core::LinAlg::Matrix<nsd, 1> refgradpres(Core::LinAlg::Initialization::zero);
-  refgradpres.clear();
+  ;
```
amgebauer:
```suggestion
```

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_parameter.hpp
```diff
@@ -54,6 +54,11 @@ namespace Discret
       int nds_solid_pressure() const { return nds_solidpressure_; };
       int nds_scalar() const { return nds_scalar_; };
       bool has_scalar() const { return nds_scalar_ > -1; };
+      bool has_bodyforce_contribution() const { return has_bodyforce_contribution_; };
+      std::vector<double> bodyforce_contribution_values() const
```
amgebauer:
```suggestion
      const std::vector<double>& bodyforce_contribution_values() const
```
I think a const ref makes more sense if I understand it correctly (otherwise you always make a copy)

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_phasemanager.hpp
```diff
@@ -327,6 +327,9 @@ namespace Discret
         virtual void permeability_tensor_vol_frac_pressure(int volfracpressnum,
             Core::LinAlg::Matrix<1, 1>& permeabilitytensorvolfracpressure) const = 0;
 
+        //! get the body force contribution
+        virtual std::vector<double> bodyforce_contribution_values() const = 0;
```
amgebauer:
```suggestion
        virtual const std::vector<double>& bodyforce_contribution_values() const = 0;
```

src/porofluid_pressure_based_ele/4C_porofluid_pressure_based_ele_evaluator.cpp
```diff
@@ -16,6 +16,19 @@
...
+namespace
+{
+  template <int nsd>
+  Core::LinAlg::Tensor<double, nsd> get_bodyforce_vector(
+      const Discret::Elements::PoroFluidManager::PhaseManagerInterface& phasemanager)
+  {
+    Core::LinAlg::Matrix<nsd, 1> bodyforce(Core::LinAlg::Initialization::zero);
+    for (int idim = 0; idim < nsd; idim++)
+      bodyforce(idim) = phasemanager.bodyforce_contribution_values()[idim];
+    return Core::LinAlg::reinterpret_as_tensor<nsd>(bodyforce);
```
amgebauer:
```suggestion
    Core::LinAlg::Tensor<double, nsd> bodyforce();
    for (int idim = 0; idim < nsd; idim++)
      bodyforce(idim) = phasemanager.bodyforce_contribution_values()[idim];
    return bodyforce;
```

PR #1515, opened 2025-11-19
Allow for global assembly of beam-to-beam coupling

isteinbrecher:
## Description and Context

Allow to globally assemble the beam to beam point coupling terms. For now this is equivalent to the local assembly, because both use a penalty regularisation, but this will allow to solve the point couplings exactly with Lagrange multipliers down the road.


src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.cpp
```diff
@@ -148,6 +149,208 @@ void BeamInteraction::BeamToBeamPointCouplingPair<Beam1, n_dof_beam_1, Beam2,
...
+    // Get the GIDs that are not taken by another coupling pair.
+    std::vector<int> available_gid;
+    for (size_t i = 0; i < lambda_gid_total.size(); i++)
+    {
+      const int gid = lambda_gid_total[i];
+      const int lid = mortar_manager->get_lambda_dof_row_map()->lid(gid);
+      if (lid < 0) FOUR_C_THROW("Could not find the GID {} in the lambda dof row map!", gid);
+      if (global_lambda_active.get_values()[lid] < 1e-5)
+      {
+        if (available_gid.size() == 0 && i % 3 != 0)
```
knarfnitram, line 189: Would it be good to use a variable here for 3 ( and occurrences bellow) to easily spot why we use 3 here?

tests/input_files/beam3r_herm2line3_beam_to_beam_penalty_point_coupling_elbow_offset_indirect_global_assembly.4C.yaml
```diff
@@ -0,0 +1,126 @@
+TITLE:
+  - "Test that beam-to-beam point coupling works with closest point projections"
```
knarfnitram: Do you mind adding here a bit more description/information regarding the test case?

isteinbrecher: Good point, I extended the descriptions of all test cases.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.cpp
isteinbrecher: The 3 here is the number of dimensions. As this implementation is only intended for 3D, I would leave the 3 here. Also, the number of dimensions is hardcoded all over beam interaction.

knarfnitram (approved):

davidrudlstorfer (approved):

PR #1516, opened 2025-11-20
Replace AMGnxn with Teko for ssti framework

c-p-schmidt:
## Description and Context
As part of #754, the self-implemented AMGnxn framework should be replaced by Teko. This is done in this PR for the ssti framework.

part of #754


mayrmt:
@c-p-schmidt Thanks!

PR #1517, opened 2025-11-20
Delete some unused functions in 4C_linalg_mapextractor

eulovi:
## Description and Context

While going through the coverage report, in `src/core/linalg/src/sparse/4C_linalg_mapextractor.cpp`, the functions:

- `std::shared_ptr<Core::LinAlg::MultiVector<double>> Core::LinAlg::MultiMapExtractor::insert_vector(...)`
- `double Core::LinAlg::MultiMapExtractor::norm2(...)`
- `void scale(Core::LinAlg::Vector<double>& full,...)`
- `void scale(Core::LinAlg::MultiVector<double>& full,...)`

are not tested, and therefore unused. According to CLion, these functions are called in code that is guarded by flags (although I am not sure if someone activates these flags in their work). The flags are the following ones (in parenthesis is the location of these flags):
- CONSISTENTSPINSK (beam3)
- ENDPOINTSEGMENTATION (beamcontact, beaminteraction)
- DEBUG_TIMINT_STD (xfem)

In this PR, the previous untested called is deleted. As most of it is related to beams, could you check if this code can be safely deleted @davidrudlstorfer, @isteinbrecher?

### DIFF
--- a/doc/removed_code/major_changes.txt
+++ b/doc/removed_code/major_changes.txt
@@ -1,5 +1,15 @@
 List of major features that were removed from 4C, including date and revision where code was removed
 
+Guarded code in beam3, beamcontact and beaminteraction
+- While cleaning up Core::LinAlg::MultiMapExtractor, some guarded code was found related to untested functions. The
+  code was guarded by the flags:
+    - CONSISTENTSPINSK: Applies consistent variant of first spin vector
+    - ENDPOINTSEGMENTATION: Applied in beamcontact and beaminteraction
+    - DEBUG_TIMINT_STD: Activates checks in xfem
+- reasons: unused in current research, untested code, might interfere in refactoring efforts
+- removed in commit 840885ccb712f0277e0094e7bf4150a300983107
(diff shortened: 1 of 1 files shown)


eulovi:
@mayrmt yes you're right! I did a mistake in my commit history while doing a rebase. I'll fix it :)

eulovi:
I just fixed the commit history, therefore this is ready for review.

mayrmt (approved):
I am fine with the changes to `linalg` and to `xfem`. I can't speak for the beam people though.

isteinbrecher (approved):
I am fine with deleting this unused and untested code in the beam framework. This surely are useful features according to the respective publications, but we don't even know if they compile and have no tests for them.

For me the only question is if we should add this to the removed code docs so it is easier to find the hash if someone wants to reactivate this feature. What do you think @mayrmt ?

eulovi:
@isteinbrecher I added the information to the docs of removed code, could you reapprove?

PR #1518, opened 2025-11-20
Particle remove unnecessary init and setup

ppraegla:
Currently, the particle code relies on `init` and `setup` functions that need to be called before an object can be used. However, it should be possible to set up an object that is directly usable without the need to call two functions.

In a first effort to solve this, the PR removes unnecessary `init` and `setup` functions. Especially, empty `init` and `setup` functions are removed. They made the code just harder to follow. Also, almost all `init` functions are moved to the constructor.

The setup functions are more complicated and will be done in another PR.

### DIFF
--- a/src/particle/src/algorithm/4C_particle_algorithm.cpp
+++ b/src/particle/src/algorithm/4C_particle_algorithm.cpp
@@ -35,6 +35,8 @@
 #include <Teuchos_StandardParameterEntryValidators.hpp>
 #include <Teuchos_TimeMonitor.hpp>
 
+#include <memory>
+
 FOUR_C_NAMESPACE_OPEN
 
 /*---------------------------------------------------------------------------*
@@ -375,7 +377,6 @@ void Particle::ParticleAlgorithm::init_particle_engine()
 {
   // create and init particle engine
   particleengine_ = std::make_shared<Particle::ParticleEngine>(get_comm(), params_);
-  particleengine_->init();
 }
 
 void Particle::ParticleAlgorithm::init_particle_wall()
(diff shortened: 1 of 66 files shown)


src/particle/src/algorithm/4C_particle_algorithm.cpp
```diff
@@ -535,17 +518,13 @@ void Particle::ParticleAlgorithm::init_viscous_damping()
...
-  // init viscous damping handler
-  if (viscousdamping_) viscousdamping_->init();
 }
 
 void Particle::ParticleAlgorithm::generate_initial_particles()
 {
-  // create and init particle input generator
+  // create particle input generator
   std::unique_ptr<Particle::InputGenerator> particleinputgenerator =
       std::unique_ptr<Particle::InputGenerator>(new Particle::InputGenerator(get_comm(), params_));
```
c-p-schmidt:
```suggestion
      std::make_unique<Particle::InputGenerator>(get_comm(), params_);
```

src/particle/src/algorithm/4C_particle_algorithm_gravity.hpp
```diff
@@ -36,23 +33,10 @@ namespace Particle
...
-     * \param[in] params particle simulation parameter list
-     */
-    explicit GravityHandler(const Teuchos::ParameterList& params);
-
-    /*!
-     * \brief init gravity handler
-     *
-     *
-     * \param[in] gravity gravity acceleration
+     * \param[in] params the gravity vector
```
c-p-schmidt: Also for `gravityrampfctnumber`.
```suggestion
     * \param[in] gravity the gravity vector
```

c-p-schmidt (approved):
Nice work!

src/particle/src/algorithm/4C_particle_algorithm_gravity.hpp
ppraegla: Thanks. Copy-pasta mistake

PR #1520, opened 2025-11-20
Refactor flag for partitioning of discretization

maxfirmbach:
## Description and Context

Adapts the functionality of `do_extended_ghosting` flag for partitioning of the discretization.

### DIFF
--- a/src/core/fem/src/discretization/4C_fem_discretization_partition.cpp
+++ b/src/core/fem/src/discretization/4C_fem_discretization_partition.cpp
@@ -299,7 +299,7 @@ std::shared_ptr<Core::LinAlg::Graph> Core::FE::Discretization::build_node_graph(
  *----------------------------------------------------------------------*/
 std::pair<std::shared_ptr<Core::LinAlg::Map>, std::shared_ptr<Core::LinAlg::Map>>
 Core::FE::Discretization::build_element_row_column(const Core::LinAlg::Map& noderowmap,
-    const Core::LinAlg::Map& nodecolmap, bool do_extended_ghosting) const
+    const Core::LinAlg::Map& nodecolmap, bool find_ghost_elements_with_no_owned_node) const
 {
   const int myrank = Core::Communication::my_mpi_rank(get_comm());
   const int numproc = Core::Communication::num_mpi_ranks(get_comm());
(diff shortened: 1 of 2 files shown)


src/core/fem/src/discretization/4C_fem_discretization_partition.cpp
```diff
@@ -384,11 +384,11 @@ Core::FE::Discretization::build_element_row_column(const Core::LinAlg::Map& node
       for (int j = 0; j < numnode; ++j)
         if (noderowmap.my_gid(nodeids[j])) ++nummine;
 
-      // Check if I own nodes of this element
-      if (!nummine)
+      // if I do not own any of the nodes, it is definitely not my element
+      // and I do not ghost it
```
isteinbrecher: I think that comment is misplaced. If no nodes are owned, the element is still ghosted if all nodes are in the column map (for `do_extended_ghosting==true`)

eulovi: Maybe the following comment would be sufficient:
"Do ghosting on no owned nodes of this element if required."

maxfirmbach: I updated the comments :+1:.

eulovi (dismissed):

src/core/fem/src/discretization/4C_fem_discretization_partition.cpp
```diff
@@ -384,11 +384,11 @@ Core::FE::Discretization::build_element_row_column(const Core::LinAlg::Map& node
       for (int j = 0; j < numnode; ++j)
         if (noderowmap.my_gid(nodeids[j])) ++nummine;
 
-      // Check if I own nodes of this element
-      if (!nummine)
+      // Do ghosting on no owned nodes of this element if required
```
isteinbrecher: Sorry to be pick, but this comment is very hard to read/understand.

isteinbrecher (commented):
@maxfirmbach I had a look at the logic again, and think that this PR is a lot more confusing than before. Is there a reason you switched the `if` statements?

maxfirmbach:
@eulovi @isteinbrecher Resolved this in a short offline discussion. I will just rename the flag indicating to do extended ghosting.

mayrmt (approved):

PR #1522, opened 2025-11-21
Refactor build monolithic node graph

maxfirmbach:
## Description and Context

This PR adds the possibility to add an `offset`-vector to the monolithic node graph build to adapt the coordinates of the bounding volumes build during construction.

### DIFF
--- a/src/core/rebalance/src/4C_rebalance_graph_based.cpp
+++ b/src/core/rebalance/src/4C_rebalance_graph_based.cpp
@@ -348,14 +348,13 @@ std::shared_ptr<const Core::LinAlg::Graph> Core::Rebalance::build_monolithic_nod
         "graph.");
 
   // 1. Do a global geometric search
-  Core::LinAlg::Vector<double> zero_vector =
-      Core::LinAlg::Vector<double>(*(dis.dof_col_map()), true);
-
   std::vector<std::pair<int, Core::GeometricSearch::BoundingVolume>> bounding_boxes;
   for (auto element : dis.my_row_element_range())
   {
     if (displacement == nullptr)
     {
+      Core::LinAlg::Vector<double> zero_vector(*dis.dof_col_map(), true);
+
       bounding_boxes.emplace_back(std::make_pair(element.global_id(),
           element.user_element()->get_bounding_volume(dis, zero_vector, params)));
     }


m-frey (dismissed):

src/core/rebalance/src/4C_rebalance_graph_based.hpp
```diff
@@ -137,12 +137,14 @@ namespace Core::Rebalance
   \pre The discretization has to be filled()==true.
 
   @param[in] dis discretization whose monolithic node graph will be build
+  @param[in] offset a vector containing e.g. displacement values that shifts coordinate values
```
isteinbrecher: I don't like the name `offset`. I would simply name it `displacement_vector`, that makes the meaning much clearer in my opinion.

isteinbrecher (commented):
@maxfirmbach is this needed for replacing the binning?

src/core/rebalance/src/4C_rebalance_graph_based.hpp
maxfirmbach: But then this is fixed to a displacement vector and rather specific. I'm fine with this name, I just want to mention it.

maxfirmbach:
@isteinbrecher Yep.

src/core/rebalance/src/4C_rebalance_graph_based.hpp
isteinbrecher: That's how we call this vector everywhere else in the code, so I see no issues with displacement vector.

isteinbrecher:
> @isteinbrecher Yep.

I knew I have seen that code before 😅

src/core/rebalance/src/4C_rebalance_graph_based.hpp
maxfirmbach: Changed it to `displacement` :+1:.

src/core/rebalance/src/4C_rebalance_graph_based.cpp
```diff
@@ -353,8 +354,16 @@ std::shared_ptr<const Core::LinAlg::Graph> Core::Rebalance::build_monolithic_nod
   std::vector<std::pair<int, Core::GeometricSearch::BoundingVolume>> bounding_boxes;
   for (auto element : dis.my_row_element_range())
   {
-    bounding_boxes.emplace_back(std::make_pair(element.global_id(),
-        element.user_element()->get_bounding_volume(dis, zero_vector, params)));
+    if (displacement == nullptr)
+    {
+      bounding_boxes.emplace_back(std::make_pair(element.global_id(),
+          element.user_element()->get_bounding_volume(dis, zero_vector, params)));
```
mayrmt, line 359: The `zero_vector` is only needed here, so its creation should be moved inside this `if` branch.

src/core/rebalance/src/4C_rebalance_graph_based.hpp
```diff
@@ -137,12 +137,15 @@ namespace Core::Rebalance
...
+  @param[in] displacement vector containing displacement values that shifts coordinate values for
+                          the bounding volume construction
 
   @return Completed monolithic node graph of input discretization
   */
   std::shared_ptr<const Core::LinAlg::Graph> build_monolithic_node_graph(
       const Core::FE::Discretization& dis,
-      const Core::GeometricSearch::GeometricSearchParams& params);
+      const Core::GeometricSearch::GeometricSearchParams& params,
+      const std::shared_ptr<const Core::LinAlg::Vector<double>>& displacement = nullptr);
```
mayrmt, line 148: Does this really need to be a reference to a pointer? Wouldn't a `const Core::LinAlg::Vector<double>&` suffice? The displacement vector is just forwarded to another function as a const reference.

mayrmt (changes requested):
@maxfirmbach I left some comments for possible changes. I am sure about moving the instantiation of the `zero_vector`. The other comment might be up for discussion, but it looks pretty clear to me at first glance.

src/core/rebalance/src/4C_rebalance_graph_based.hpp
maxfirmbach: The `nullptr` trick wouldn't work then :smile:.

src/core/rebalance/src/4C_rebalance_graph_based.cpp
maxfirmbach: I moved the creation of the zero vector into the if branch.

PR #1524, opened 2025-11-21
[LinAlg] Remove the inheritance of `Epetra_Operator` from `SparseOperator`

### LINKED ISSUE #1508: What to do with `::NOX::Epetra::MatrixFree` and `::NOX::Epetra::FiniteDifference`
## Description
Before completely removing `::NOX::Epetra::` classes, the following question has to be resolved. What do we do with `::NOX::Epetra::MatrixFree` and `::NOX::Epetra::FiniteDifference`? For the former, `Trilinos` provides `NOX::Thyra::MatrixFreeJacobianOperator` as a more modern replacement. Regarding the finite differences, there is `Thyra::DefaultFiniteDifferenceModelEvaluator` but it does not look like a direct replacement.

I see 2 possible strategies:
- remove these classes completely,
- use `NOX::Thyra::MatrixFreeJacobianOperator` - this replaces only the first class and may require a lot of preparatory boilerplate code on our side,
- write our own replacements of these 2 classes. Maybe we do not need a finite difference functionality at all, as it is rather a debug functionality, but for debugging itself, it can be quite helpful, I assume. These replacemens can be later abandoned if we decide to use `Thyra`.

## Interested Parties
@maxfirmbach @mayrmt

knarfnitram:
## Description and Context

The `SparseOperator` which is the the general virtual Interface for the 4c matrices, inherits directly from the `Epetra_Operator`, offering to access functions from Epetra directly within our codebase. 
I tried to remove the inheritance in a first step, but came up with some issues related to nox.

Since the inheritance is removed, the still necessary `Epetra_Operator` could be reused from the stored matrix as done with the `SparseMatrix` or for our _custom_ implementations (`BlockSparseMatrixBase`), where we overwrite methods of the `Epetra_Operator` we would still need to directly inherit from the `Epetra_Operator` (according to @maxfirmbach we could probably also remove this inheritance by switching from the belos interface to thyra for the iterative solver interface).

This PR is meant as discussion with some code example. 
Open questions:
- How can this be aligned with our new structure time integration (eg. nox)  since the `Epetra_Operator` is used there. 
How is it planed to be used? 
- What do you think about the general design of this interface? 
- I'd be happy to hear your opinions @maxfirmbach @vovannikov @sebproell 

In the current state, this PR:
- Removes the inheritance of `Epetra_Operator` from `SparseOperator` and therefore also removes it from `SparseOperatorWrapper` and `FSIMatrixFree` and propagates it down to `BlockSparseMatrix`.
- Refactors the `SparseOperator` interface to adhere to our coding guidelines and removes a lot `Epetra`-related code.
- Makes the `SparseMatrix` wrapper independent of `Epetra_Operator` and deletes and adjust several methods.

Closes #1508

### DIFF
--- a/src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_jacobian.cpp
+++ b/src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_jacobian.cpp
@@ -52,7 +52,7 @@ NOX::FSI::FSIMatrixFree::FSIMatrixFree(Teuchos::ParameterList& printParams,
   }
 }
 
-Epetra_Operator& NOX::FSI::FSIMatrixFree::epetra_operator() { return *this; }
+Epetra_Operator& NOX::FSI::FSIMatrixFree::epetra_operator() { FOUR_C_THROW("Not implemented"); }
 
 void NOX::FSI::FSIMatrixFree::zero() { FOUR_C_THROW("Not implemented"); }
 
@@ -133,7 +133,6 @@ void NOX::FSI::FSIMatrixFree::multiply(bool TransA, const Core::LinAlg::MultiVec
         << "ERROR: FSIMatrixFree::multiply() - Transpose is unavailable in Matrix-Free mode!"
         << std::endl;
     throw "NOX Error";
-    return;
   }
 
   // Calculate the matrix-vector product:
@@ -193,77 +192,6 @@ void NOX::FSI::FSIMatrixFree::multiply(bool TransA, const Core::LinAlg::MultiVec
   // scale back
   // nevY.update(xscale, perturbY, 0.0);
   nevY.update(1., perturbY, 0.0);
-
-  return;
-}
-
-int NOX::FSI::FSIMatrixFree::SetUseTranspose(bool UseTranspose)
-{
-  FOUR_C_THROW("Not implemented");
-  return -1;
-}
-
-
-int NOX::FSI::FSIMatrixFree::Apply(const Epetra_MultiVector& X, Epetra_MultiVector& Y) const
-{
-  FOUR_C_THROW("Not implemented");
-  return -1;
-}
-
-
-int NOX::FSI::FSIMatrixFree::ApplyInverse(const Epetra_MultiVector& X, Epetra_MultiVector& Y) const
-{
-  FOUR_C_THROW("Not implemented");
-  return -1;
-}
-
-
-double NOX::FSI::FSIMatrixFree::NormInf() const
-{
-  FOUR_C_THROW("Not implemented");
-  return 1.0;
-}
-
-
-const char* NOX::FSI::FSIMatrixFree::Label() const
-{
-  FOUR_C_THROW("Not implemented");
-  return label.c_str();
-}
-
-
-bool NOX::FSI::FSIMatrixFree::UseTranspose() const
-{
-  FOUR_C_THROW("Not implemented");
-  return false;
-}
-
-
-bool NOX::FSI::FSIMatrixFree::HasNormInf() const
-{
-  FOUR_C_THROW("Not implemented");
-  return false;
-}
-
-
-const Epetra_Comm& NOX::FSI::FSIMatrixFree::Comm() const
-{
-  FOUR_C_THROW("Not implemented");
-  return Core::Communication::as_epetra_comm(currentX.get_linalg_vector().get_map().get_comm());
-}
-
-
-const Epetra_Map& NOX::FSI::FSIMatrixFree::OperatorDomainMap() const
-{
-  FOUR_C_THROW("Not implemented");
-  return *epetraMap;
-}
-
-
-const Epetra_Map& NOX::FSI::FSIMatrixFree::OperatorRangeMap() const
-{
-  FOUR_C_THROW("Not implemented");
-  return *epetraMap;
 }
 
 
--- a/src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_jacobian.hpp
+++ b/src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_jacobian.hpp
@@ -21,9 +21,6 @@
 
 #include <memory>
 
-// Forward Declarations
-class Map;
-
 FOUR_C_NAMESPACE_OPEN
 
 namespace NOX
@@ -37,8 +34,7 @@ namespace NOX
   {
     /// Matrix Free Newton Krylov based on an approximation of the residuum derivatives
     class FSIMatrixFree : public Core::LinAlg::SparseOperator,
-                          public virtual NOX::Nln::Interface::JacobianBase,
-                          public Epetra_Operator
+                          public virtual NOX::Nln::Interface::JacobianBase
     {
      public:
       /*! \brief Constructor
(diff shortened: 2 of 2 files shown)


maxfirmbach:
I think this is a good first approach due to several reasons:
1. It allows us to define and steer our own abstract matrix interface.
2. It isolates the inheritance to a place, where it is actually used. Our `SparseMatrix` holds an `Epetra_CrsMatrix` inside its own wrapper, so no need to inherit from `Epetra_Operator`.

src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
```diff
@@ -237,4 +243,4 @@ namespace Core::LinAlg
 FOUR_C_NAMESPACE_CLOSE
 
 #endif
-/*LINALG_SPARSEOPERATOR_H_*/
+/*LINALG_SPARSEOPERATOR_H_*/
```
sebproell:
```suggestion
```

src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
```diff
@@ -81,14 +81,17 @@ namespace Core::LinAlg
       \warning Only low level solver routines are interested in the internal
       Epetra_Operator.
      */
-    virtual Epetra_Operator& epetra_operator() { return *this; }
+    virtual ~SparseOperator();  // declaration only
```
sebproell:
```suggestion
    virtual ~SparseOperator() = default;
```


and remove the corresponding cpp file.

src/fsi/src/monolithic/nonlinear_solver/4C_fsi_nox_linearsystem.cpp
```diff
@@ -33,15 +33,16 @@ NOX::FSI::LinearSystem::LinearSystem(Teuchos::ParameterList& printParams,
...
       operator_(J),
       scaling_(s),
       callcount_(0),
       solver_(solver),
       timer_("", true)
 {
   tmp_vector_ptr_ = std::make_shared<NOX::Nln::Vector>(cloneVector);
-
+  // TODO: Potential error here?
+  jac_ptr_ = std::shared_ptr<Epetra_Operator>(J, &J->epetra_operator());
```
sebproell: This is correct AFAICT, however, aliasing shared_ptrs is complicated enough that it might warrant its own helper function.

maxfirmbach:
@vovannikov What's your take on the nonlinear solver? From the linear solver perspective I'm pretty confident, that we can get rid of the inheritance by using the `Thyra` based interface and conversion utilities.

vovannikov:
@knarfnitram Thank you for this effort!

In my opinion, `Epetra_Operator` should get wiped out completely from 4C. I think creating our own wrappers is the way to go. It is fine that some of these wrappers may be inherited at the moment, but eventually this inheritance gets broken, once the code is ready for that.

Similarly here, this inheritance from `Epetra_Operator` should be broken, but I have a feeling that it is a bit too early to do this, as it requires a lot of work within a single PR. One may be prone to adhering to fast but suboptimal solutions here and there in the code, since at the current state, breaking this inheritance is indeed challenging: the code is not yet ready for this.

It is possible to make NOX independent from `Epetra` since, NOX, by design, contains generalized abstract base classes. The classes from `NOX::Epetra::` should not be used eventually. This can not be achieved immediately, but I have been steadily working on this. I have recently also added our own implementation of the abstract NOX vector (#1399, #1445). Similar idea should be used for other NOX related components.

@maxfirmbach I think you are right. `Thyra` + the conversion utils is the way to go, I have seen how you made it for preconditioners, nice. I have a similar plan for the NOX refactoring, however, I would also like to wrap `Thyra` operators inside our own wrapper, or something like that.

knarfnitram:
Thanks for the update @vovannikov. Good to hear, that we are on the same page. Just let us know if you need any help regarding the matrix interfaces or once the NOX interface is ready such that we can continue here.

maxfirmbach:
@knarfnitram Maybe it is worth rebasing this branch as there has been some work done by e.g. @vovannikov to resolve some issues related to this PR.

maxfirmbach:
@vovannikov Only problematic code line is the necessity to convert from `Epetra_Operator` to `SparseOperator` in the `structure_new` nox-interface. If this is somehow resolved, this PR could go into `main`.

vovannikov:
@maxfirmbach, I will try to fix it swiftly. Perfectly, today.

src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
maxfirmbach: Done, makes total sense.

src/fsi/src/monolithic/nonlinear_solver/4C_fsi_nox_linearsystem.cpp
maxfirmbach: After cleaning up some things, we don't need this piece of code at all.

vovannikov:
The aforementioned issue is addressed in #1630. However, even after #1630 there are still may be a couple of obstacles with this PR, let's see after rebasing it.

maxfirmbach:
After rebasing, the only failing code line is the cast in `computeJacobian()` itself.

maxfirmbach:
@mayrmt @vovannikov @knarfnitram @sebproell This PR is ready for review.

src/core/linalg/src/sparse/4C_linalg_blocksparsematrix.hpp
```diff
@@ -28,7 +28,7 @@ namespace Core::LinAlg
     implemented in terms of the matrix blocks.
 
    */
-  class BlockSparseMatrixBase : public SparseOperator
+  class BlockSparseMatrixBase : public SparseOperator, public Epetra_Operator
```
vovannikov: Why is inheritance from `Epetra_Operator` still required for `BlockSparseMatrixBase`?

src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
vovannikov: This comment does not seem to be true after this PR :)

src/solver_nonlin_nox/4C_solver_nonlin_nox_matrixfree.hpp
```diff
@@ -63,7 +63,7 @@ namespace NOX
         Teuchos::RCP<const ::Thyra::VectorSpaceBase<double>> vector_space_;
       };
 
-      class SparseOperatorWrapper : public Core::LinAlg::SparseOperator
+      class SparseOperatorWrapper : public Core::LinAlg::SparseOperator, public Epetra_Operator
```
vovannikov: Do you need to add the inheritance from `Epetra_Operator` here? Again, the dummy implementations of `Epetra_Operator` for this class have to be removed (which is even better).

maxfirmbach: No, not necessary. Actually, I have the commit ready for removing the `Epetra_Operator` from this again.

src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
maxfirmbach: That's true, I removed the comment.

src/solver_nonlin_nox/4C_solver_nonlin_nox_matrixfree.hpp
maxfirmbach: I removed the `Epetra_Operator` related things from the wrapper.

src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_jacobian.cpp
```diff
@@ -131,7 +133,7 @@ int NOX::FSI::FSIMatrixFree::multiply(bool TransA, const Core::LinAlg::MultiVect
         << "ERROR: FSIMatrixFree::multiply() - Transpose is unavailable in Matrix-Free mode!"
         << std::endl;
     throw "NOX Error";
-    return -1;
+    return;
```
vovannikov:
```suggestion
```

src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_jacobian.cpp
```diff
@@ -167,7 +169,7 @@ int NOX::FSI::FSIMatrixFree::multiply(bool TransA, const Core::LinAlg::MultiVect
     // In the first call is x=0. No need to calculate the
     // residuum. y=0 in that case.
     nevY.init(0.);
-    return 0;
+    return;
```
vovannikov, line 171:
```suggestion
```

src/fsi/src/partitioned/nonlinear_solver/4C_fsi_nox_jacobian.cpp
```diff
@@ -192,76 +194,7 @@ int NOX::FSI::FSIMatrixFree::multiply(bool TransA, const Core::LinAlg::MultiVect
...
-  FOUR_C_THROW("Not implemented");
-  return *epetraMap;
-}
-
-
-const Epetra_Map& NOX::FSI::FSIMatrixFree::OperatorRangeMap() const
-{
-  FOUR_C_THROW("Not implemented");
-  return *epetraMap;
+  return;
```
vovannikov:
```suggestion
```

Missed these obsolete returns.

maxfirmbach:
In the current state, this PR should close #1508 for now.

vovannikov (approved):

maxfirmbach (approved):

PR #1525, opened 2025-11-22
[LinAlg] Cleanup of sparse matrix implementations

maxfirmbach:
## Description and Context

This PR:
- Removes the method `Core::LinAlg::block_matrix2x2()` and refactors its use in `xfluid`.
- Removes the method `Core::LinAlg::BlockSparseMatrixBase::get_partial_extractor()` and everything related to it, as it is not used.
- Makes `epetra_operator()` a pure virtual method and let's the classes inheriting it provide a specific implementation.
- Removes some mentions of `Epetra_Operator` in contact.
- Adds methods to the matrix implementations returning `Thyra` based objects for everything `Trilinos` solver related in the future.

Related to #863

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_sparsematrix.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_sparsematrix.hpp
@@ -18,7 +18,6 @@
 
 #include <Epetra_CrsMatrix.h>
 #include <Epetra_FECrsMatrix.h>
-#include <Thyra_EpetraLinearOp.hpp>
 
 
 FOUR_C_NAMESPACE_OPEN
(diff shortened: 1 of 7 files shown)


dharinib98 (dismissed):

dharinib98:
Thanks @maxfirmbach looks good to me!

src/core/linalg/src/sparse/4C_linalg_sparsematrix.hpp
```diff
@@ -18,6 +18,7 @@
 
 #include <Epetra_CrsMatrix.h>
 #include <Epetra_FECrsMatrix.h>
+#include <Thyra_EpetraLinearOp.hpp>
```
sebproell: This will expose Thyra in 4C. I would highly recommend to keep this an internal detail only in the places where we interface with the solvers. Exposing it will lead to similar problems that we are currently facing with Epetra. Essentially, we should go for a high-level abstraction of linear algebra without referring to any implementation details.

maxfirmbach: Alright, I can move it to some utils converter-style function in the linear solver, this should work too.

maxfirmbach: @sebproell I removed the `Thyra` related functionality from the sparse linear algebra and added converter utils functions to the linear solver :+1:.

PR #1531, opened 2025-11-26
Fix failing no optional dependencies pipeline

c-p-schmidt:
## Description and Context
The no optional dependency pipeline fails with the error that ArborX is not available (see https://github.com/4C-multiphysics/4C/actions/runs/19685821006/job/56413944554). This PR adds the required dependency to the test, which should make the no-optional-dependencies pipeline pass again.

Most likely following #1459


c-p-schmidt:
One further question. Currently, there is a specific section in the file `tests/list_of_tests.cmake`  for special tests like those that require, e.g., ArborX. Do we want to keep it, i.e., move this test down?

maxfirmbach (approved):
Might make sense to move it down, on the other hand it shouldn't matter in general.

knarfnitram (approved):

PR #1533, opened 2025-11-26
Move CI python tools into own 4C Python subpackage

### LINKED ISSUE #1325: [BUG] Coverage is failing

amgebauer:
The current coverage report fails during chunking of the test suite. The script that chunks the test suite installs the 4C-python package with all build requirements (which are quite a lot including vtk with a half a GB).

I now added a own subpackage `ci`, so that we only need to install the required tools (which is a few KB). I now expect that the coverage build should pass again. Here is the test on my fork: https://github.com/amgebauer/4C/actions/runs/19700230260

Closes #1325


c-p-schmidt (approved):

utilities/four_c_python/README.md
```diff
@@ -1,8 +1,10 @@
 A Python package that bundles functionality used with/in 4C.
 
-It contains 4 Python packages
+It contains the following Python packages:
```
c-p-schmidt, line 3: Directly future proove without a number ;-)

davidrudlstorfer (commented):
Nice addition @amgebauer!

I know I'm late to the party, but I think that it would probably make sense to document this change somewhere in the readme. It's really specific that we need to split the requirements because vtk is too large for the environment on Github runners 😅

PR #1534, opened 2025-11-26
Change defaults for RESULTSEVERY and RESTARTEVERY of structure module

c-p-schmidt:
## Description and Context
While working on the structure time integration output, mainly switching to the vtk-based output and preparing it, @lkoeglmeier and I encountered some problems due to the set default values for `RESULTSEVERY` and `RESTARTEVERY`, which are both set to `1` for the structure time integration.

Thus, we propose the more reasonable default of `0`, i.e., no restart, and the result (which is still the old binary output that needs to be post-processed) is written. Through this, I have already identified an additional bug in the microstructure model evaluator, which is also addressed in this PR (see the first commit).

The change in default behavior is implemented in the second commit, which does not simply replicate the `old` settings by copying a `1` to the input files, but rather the values that are actually required, which enabled the discovery of the aforementioned bug.

Furthermore, I encountered another bug in the old structural time integration, which only works if `RESULTSEVERY` is set to `1`; otherwise, one ends up in a segmentation fault. However, as we want to eliminate the old time integration anyway, I did not investigate it further.

Follows #1503 #1498 #1487 #1496 #1468


src/stru_multi/4C_stru_multi_microstatic.cpp
```diff
@@ -105,7 +105,10 @@ MultiScale::MicroStatic::MicroStatic(const int microdisnum, const double V0)
   restart_every_ = sdyn_macro.get<int>("RESTARTEVERY");
 
   // i/o options should be read from the corresponding micro-file
-  results_every_ = sdyn_micro.get<int>("RESULTSEVERY");
+  results_every_ = Global::Problem::instance(microdisnum_)
+                       ->io_params()
+                       .sublist("RUNTIME VTK OUTPUT")
+                       .get<int>("INTERVAL_STEPS");
```
ppraegla, line 111: I'm not familiar with the microstructure model evaluator. Why is `RESULTSEVERY` not the correct input parameter to write the results?

c-p-schmidt: The problem with the structure time integration is that both output versions (binary output that requires postprocessing and vtk-based output) are still possible. The `RESULTSEVERY` is the setting consumed by the old output mechanism, whereas the other is the setting to be used by the new vtk-based output mechanism.

However, we are almost there, with only the vtk-based output, and then this ambiguity can and will go away.

ppraegla: Thanks :+1:

PR #1532, opened 2025-11-26
[NOX] Wrap `::NOX::Epetra::MatrixFree`

vovannikov:
## Description and Context
Prior to removing `::NOX::Epetra::Interface::Jacobian`, I decide to wrap `::NOX::Epetra::MatrixFree` (such that it will be then easier to switch to the corresponding `Thyra` infrastructure) and remove `::NOX::Epetra::FiniteDifference` as it is not in use, see #1508.

I have also added another version of test `tests/input_files/fsi_dc_part_mfnk_ost_ga.4C.yaml` to test the use of `::NOX::Epetra::MatrixFree`.

## Related Issues and Pull Requests
#1508 - the relevant discussion
#1108 - the problematic test was identified
#1486 - the planned removal of `::NOX::Epetra::Interface::Jacobian`,
#1445 - the concern on `::NOX::Epetra::MatrixFree`  was raised
#1524 - the current PR is required to proceed with that breakage of inheritance

PR #1535, opened 2025-11-27
[NOX] Introduce `JacobianBase` class

vovannikov:
## Description and Context
This PR introduces `NOX::Nln::Interface::JacobianBase` instead of `::NOX::Epetra::Interface::Jacobian`. This is similar to #1486. Later these both inheritances will be broken.

#1486, #1532


mayrmt (approved):
@vovannikov, thanks! This is ready to be merged.

Issue #1162, opened 2025-08-25
Refactor Convergence tests for Beaminteraction with Lagrange Multipliers


dharinib98:
## Description

This is another post implementation issue of #637. The tests that use Lagrange Multipliers, right now use tolerances relaxed atleast by 6 orders. This could be improved by fixing how the residual is evaluated. Along similar lines, it would be helpful looking into field-wise convergence tests to be independent of scaling effects.  See #1004 
## Possible Solution and Definition of Done

- [ ] Fix residual evaluation 
- [ ] Field-wise convergence tests ? 
## Interested Parties

@isteinbrecher @mayrmt @maxfirmbach

mayrmt:
@dharinib98 As this issue came up again in #1501, we need to tackle it soon. Please look into it. @isteinbrecher and I are happy to discuss any open questions.

isteinbrecher:
@dharinib98 can this issue be closed?

dharinib98:
closed

PR #1540, opened 2025-11-29
Switch to incremental formulation for Lagrange Multipliers in Beaminteraction

dharinib98:
## Description and Context

This PR converts the coupling in beam interaction using Lagrange Multipliers from a full-λ formulation to an incremental Lagrange multiplier formulation. 

#1162


src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
```diff
@@ -722,7 +722,40 @@ void BeamInteraction::BeamToSolidMortarManager::assemble_force(
...
 
+  auto lambda = get_global_lambda();
+  Core::LinAlg::Vector<double> force_solid_lin_lambda_times_lambda(
+      force_solid_lin_lambda_->range_map());
+  Core::LinAlg::Vector<double> force_beam_lin_lambda_times_lambda(
+      force_beam_lin_lambda_->range_map());
+
+  // obtain the matrix blocks of the rhs of the lambda incremental scheme
+  force_solid_lin_lambda_->Apply(*lambda, force_solid_lin_lambda_times_lambda);
+  force_beam_lin_lambda_->Apply(*lambda, force_beam_lin_lambda_times_lambda);
```
mayrmt: The comment here is misleading. This code does not compute matrix blocks, but vectors to be added to the rhs.
```suggestion
  // Compute contribution of Lagrange multiplier from previous iteration to right-hand side
  force_solid_lin_lambda_->Apply(*lambda, force_solid_lin_lambda_times_lambda);
  force_beam_lin_lambda_->Apply(*lambda, force_beam_lin_lambda_times_lambda);
```

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
```diff
@@ -722,7 +722,40 @@ void BeamInteraction::BeamToSolidMortarManager::assemble_force(
...
+
+  // move to global f-map
+  Core::LinAlg::Vector<double> force_solid_lin_lambda_times_lambda_on_f(f.get_map());
+  Core::LinAlg::export_to(force_solid_lin_lambda_times_lambda,
+      force_solid_lin_lambda_times_lambda_on_f);  // Export/Add as your helper does
```
mayrmt:
```suggestion
  // Export to full map of the residual / res vector
  Core::LinAlg::Vector<double> force_solid_lin_lambda_times_lambda_on_f(f.get_map());
  Core::LinAlg::export_to(force_solid_lin_lambda_times_lambda,
      force_solid_lin_lambda_times_lambda_on_f);
```

mayrmt (changes requested):
@dharinib98 I have left some comments regarding comments in the source code to make them hopefully clearer and more accurate.

So far, this PR only adds contributions to the residual vector (as expected). How is the update of the solution vector handled, i.e. adding the increment after each Newton step to the current solution? Is this already done by the time integration framework without the need for any beam/solid-specific changes?

dharinib98:
> @dharinib98 I have left some comments regarding comments in the source code to make them hopefully clearer and more accurate.
> 
> So far, this PR only adds contributions to the residual vector (as expected). How is the update of the solution vector handled, i.e. adding the increment after each Newton step to the current solution? Is this already done by the time integration framework without the need for any beam/solid-specific changes?

@mayrmt Yes, this addition of increment is a default behavior of the time integration implemented through the `pre_compute_x`. Earlier we had one line of code in `pre_compute_x` that extracted and stored the full lambda value at the end of each iteration. We do not need that anymore.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
```diff
@@ -722,7 +722,40 @@ void BeamInteraction::BeamToSolidMortarManager::assemble_force(
...
+  Core::LinAlg::Vector<double> force_solid_lin_lambda_times_lambda(
+      force_solid_lin_lambda_->range_map());
+  Core::LinAlg::Vector<double> force_beam_lin_lambda_times_lambda(
+      force_beam_lin_lambda_->range_map());
+
+  // Compute contribution of Lagrange multiplier from previous iteration to right-hand side
+  force_solid_lin_lambda_->Apply(*lambda, force_solid_lin_lambda_times_lambda);
+  force_beam_lin_lambda_->Apply(*lambda, force_beam_lin_lambda_times_lambda);
+
+  // Export to full map of the residual / res vector
```
mayrmt:
```suggestion
  // Export to full map of the residual / rhs vector
```

mayrmt (dismissed):
@dharinib98, looks good to me!

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
```diff
@@ -891,7 +893,6 @@ void Solid::ModelEvaluator::BeamInteraction::run_pre_compute_x(
     const Core::LinAlg::Vector<double>& xold, Core::LinAlg::Vector<double>& dir_mutable,
     const NOX::Nln::Group& curr_grp)
 {
-  if (have_lagrange_dofs()) Core::LinAlg::export_to(dir_mutable, *ia_state_ptr_->get_lambda());
 }
```
isteinbrecher, line 896: I assumed this is where we set lambda in beam interaction. Can you please check that `auto lambda = get_global_lambda();` in `assemble_force` actually contains the Lagrange Multipliers of the current iteration and is not zero?

dharinib98: Yeah you are right, we need to set lambda here. lambda values are now overwritten to zero.

tests/input_files/ref/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_line3-vtk-files/beam-to-solid-volume-mortar-continuous-00002-0.vtu
```diff
@@ -10,38 +10,518 @@
...
+-2.229758063704534e-01 -2.489151457588146e-02 1.753119093211241e+00
+-2.229758063704526e-01 -2.489151457588146e-02 1.753119093211241e+00
+-2.273396158117520e-01 -2.537994673889804e-02 1.782341474169036e+00
+-2.317111366593703e-01 -2.586944412237102e-02 1.811587231307498e+00
+-2.360933115388391e-01 -2.636015839185735e-02 1.840865979856301e+00
+-2.404890830756941e-01 -2.685224121291380e-02 1.870187335045120e+00
+-2.449013938954692e-01 -2.734584425109743e-02 1.899560912103631e+00
         </DataArray>
         <DataArray type="Float64" Name="lambda" NumberOfComponents="3" format="ascii">
+0.000000000000000e+00 0.000000000000000e+00 0.000000000000000e+00
```
isteinbrecher: The zeros here, indicate that we are not setting lambda correctly in beam interaction.

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
dharinib98: @isteinbrecher @mayrmt The current state of the branch seems to work again and lambdas are not zeros anymore. Looks like the update after every iteration is taken care by the default behavior (`xnew` = `xold` + `dir_mutable` is by default carried out in `pre_compute_x` and `post_compute_x`). However, what I still do not understand is, we get the global lambda values from this variable called `dir_mutable`, which I assume is the current solution vector and even after adding the additional contributions, we still get the lambda values from this same variable. So I am not very sure if this current implementation is correct and if I am still missing the update.

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
```diff
@@ -900,6 +901,7 @@ void Solid::ModelEvaluator::BeamInteraction::run_post_compute_x(
     const Core::LinAlg::Vector<double>& xold, const Core::LinAlg::Vector<double>& dir,
     const Core::LinAlg::Vector<double>& xnew)
 {
+  if (have_lagrange_dofs()) Core::LinAlg::export_to(xnew, *ia_state_ptr_->get_lambda());
```
isteinbrecher, line 904: For now this works here, but please have a look on how to move this update to `setup` and `reset` (similar to how it is done for `ia_state_ptr_->get_dis_np()`).

dharinib98:
> Thanks for adding this @dharinib98. Since you are now using the incremental formulation, we should not need the trick of relaxing the tolerances for the Lagrange multiplier tests any more. Please adjust the tolerances accordingly to a sensible value. Should be the residual tolerances for all tests that use the Lagrange multiplier system (saddle point and regularised).

@isteinbrecher I adapted the tolerances in the LM tests. Are these acceptable tolerance values? I just used the same tolerance value that is being used for displacements

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
mayrmt: Should we move this within this PR?

dharinib98: I think we can merge this. I would open up an issue for this and look into it during my next code clean up afternoon. What do you think @isteinbrecher ? From what I already checked it was not quite straight forward since we would need to add corresponding getter functions for lambdas in the structure time integration.

isteinbrecher: I would do that as a next step side by side with the convergence check. I think we can proceed with this PR.

PR #1543, opened 2025-11-30
Remove print "processor n finished normally"

ppraegla:
Does the printing of "processor n finished normally" hold any value? Doesn't MPI throw an error if something goes wrong?

This print statement clutters the log (especially for many procs). I always need to skip this to get to the TimeMonitor.

### DIFF
--- a/apps/global_full/4C_global_full_main.cpp
+++ b/apps/global_full/4C_global_full_main.cpp
@@ -380,19 +380,6 @@ int main(int argc, char* argv[])
     get_memory_high_water_mark(arguments.comms.global_comm());
 
     Core::Communication::barrier(arguments.comms.local_comm());
-    if (arguments.comms.num_groups() > 1)
-    {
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (approved):

sebproell:
I might be wrong here, but I believe that QUEENS is looking for this output? @gilrrei @sbrandstaeter ?

sbrandstaeter:
> I might be wrong here, but I believe that QUEENS is looking for this output? @gilrrei @sbrandstaeter ?

Thank you for thinking of QUEENS.
I have just verified that there is only one reference to these print statements on the main branch of QUEENS
It is in a test (see [here](https://github.com/queens-py/queens/blob/main/tests/unit_tests/data_processors/test_txt_file.py#L84)) and should be easy to fix.
In fact, the pipeline would not even fail, since we test against a commited log-file.

TL;DR:
From what I have seen, the print statements are not essential for QUEENS.

Issue #1544, opened 2025-11-30
[LinAlg] Static sparsity layout for sparse matrices


maxfirmbach:
## Description

Currently the sparsity pattern used for our sparse matrix implementation with `Epetra_CrsMatrix` can be changed dynamically by setting the option `StaticProfile = false`. The effects are the following: we specify an approximate upper bound as number of non-zeros per row. This number is not a "hard" bound it can be considered as a hint. Afterwards we are still able to add more values in a row if necessary. Usually this will still throw a warning, if overused (e.g. one can specify zero nnz per row and we would still be able to insert values).

There are some drawbacks to this however:
- I would like to have a hard assert for warnings and errors from `Epetra`, which currently fails for `insert_into_global_values()` as we on a regular basis insert values without having the allocation ready, resulting in a lot of re-allocation during value insertion.
- The mechanism of `un_complete` called on a matrix does not help. We pre-allocate the memory and complete, but afterwards modify everything again and usually insert more values than allocated.

=> Violating allocation limits and dynamically changing memory slows down operations.

It makes sense to be cautious about when and where to allocate memory for a sparse matrix. Now to the main topic though. In `Tpetra` dynamic memory allocation is deprecated for matrices (mainly due to GPU computing). Therefore we need to switch to a static memory profile `StaticProfile = true` for our matrix implementation to make a later transition to `Tpetra` as smooth as possible.

I assume this will be a lot of work and code will break in several places.

## Possible Solution and Definition of Done

- [ ] Turn the `StaticProfile` option in the `Epetra_CrsMatrix` constructor to `true` in all places.
- [ ] Run a complete `ctest` routine and document the failing tests.
- [ ] Get all tests running again by either reworking how many entries are pre-allocated for a specific matrix and/or search for a upper bound.

maxfirmbach:
After a first initial test it seems like **ALL** tests do fail. We might need to rethink how we want to tackle this issue properly.

mayrmt:
@maxfirmbach if all tests are failing, that could also be just one problem in the matrix class -- at least I hope so.

Do you know, if the error can be traced back to a single spot in our matrix implementation or if we really need to work on every construction of a matrix throughout the entire code base?

maxfirmbach:
I'm 90% convinced that we have to touch each sparse matrix constructor throughout the code base. Our matrix class (from a constructor perspective) only hands over values (in most cases).

maxfirmbach:
[...]

PR #1546, opened 2025-12-01
[LinAlg] Add new sum_or_insert method to sparse matrix

### LINKED ISSUE #1190: [LinAlg] Move FOUR_C_THROW into respective wrapper classes

maxfirmbach:
## Description and Context

This PR adds an intermediate method `sum_or_insert` to the sparse matrix interface, which centralizes a code pattern appearing here and there in `4C`. This method should be removed at some point after #1544 is resolved. In addition the last missing asserts are implemented.

Related to #863
Closes #1190

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_sparsematrix.cpp
+++ b/src/core/linalg/src/sparse/4C_linalg_sparsematrix.cpp
@@ -1568,14 +1568,12 @@ void Core::LinAlg::SparseMatrix::replace_global_values(
 
 /*----------------------------------------------------------------------*
  *----------------------------------------------------------------------*/
-int Core::LinAlg::SparseMatrix::insert_global_values(
+void Core::LinAlg::SparseMatrix::insert_global_values(
     int global_row, int num_entries, const double* values, const int* indices)
 {
   // For now, we cant add ASSERT_EPETRA_CALL() here, as we need to allow warnings.
-  const int err = sysmat_->InsertGlobalValues(global_row, num_entries, values, indices);
+  [[maybe_unused]] int err = sysmat_->InsertGlobalValues(global_row, num_entries, values, indices);
   FOUR_C_ASSERT(err >= 0, "Epetra error (code {}).", err);
-
-  return err;
 }
 
 /*----------------------------------------------------------------------*
@@ -1594,7 +1592,8 @@ void Core::LinAlg::SparseMatrix::sum_or_insert_global_values(
   const int errone = sysmat_->SumIntoGlobalValues(global_row, num_entries, values, indices);
   if (errone > 0)
   {
-    const int errtwo = sysmat_->InsertGlobalValues(global_row, num_entries, values, indices);
+    [[maybe_unused]] int errtwo =
+        sysmat_->InsertGlobalValues(global_row, num_entries, values, indices);
     FOUR_C_ASSERT(errtwo >= 0, "Epetra error (code {}).", errtwo);
   }
   else if (errone < 0)
--- a/src/core/linalg/src/sparse/4C_linalg_sparsematrix.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_sparsematrix.hpp
@@ -562,7 +562,7 @@ namespace Core::LinAlg
         int global_row, int num_entries, const double* values, const int* indices);
 
     /// Inserts values into a global row.
-    int insert_global_values(
+    void insert_global_values(
         int global_row, int num_entries, const double* values, const int* indices);
 
     /// Sum values into a global row.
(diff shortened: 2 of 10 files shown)


src/core/linalg/src/sparse/4C_linalg_sparsematrix.cpp
```diff
@@ -1568,18 +1517,56 @@ void Core::LinAlg::SparseMatrix::replace_global_values(
 
 /*----------------------------------------------------------------------*
  *----------------------------------------------------------------------*/
-int Core::LinAlg::SparseMatrix::insert_global_values(
+void Core::LinAlg::SparseMatrix::insert_global_values(
+    int global_row, int num_entries, const double* values, const int* indices)
+{
+  // For now, we cant add ASSERT_EPETRA_CALL() here, as we need to allow warnings.
+  [[maybe_unused]] int err = sysmat_->InsertGlobalValues(global_row, num_entries, values, indices);
+  FOUR_C_ASSERT(err >= 0, "Epetra error (code {}).", err);
```
mayrmt, line 1576: Shouldn't this also use `ASSERT_EPETRA_CALL()`?

src/core/linalg/src/sparse/4C_linalg_sparsematrix.cpp
```diff
@@ -1568,18 +1517,56 @@ void Core::LinAlg::SparseMatrix::replace_global_values(
...
+ *----------------------------------------------------------------------*/
+void Core::LinAlg::SparseMatrix::sum_or_insert_global_values(
+    int global_row, int num_entries, const double* values, const int* indices)
+{
+  const int errone = sysmat_->SumIntoGlobalValues(global_row, num_entries, values, indices);
+  if (errone > 0)
+  {
+    [[maybe_unused]] int errtwo =
+        sysmat_->InsertGlobalValues(global_row, num_entries, values, indices);
+    FOUR_C_ASSERT(errtwo >= 0, "Epetra error (code {}).", errtwo);
```
mayrmt: Shouldn't this also use ASSERT_EPETRA_CALL()?

src/core/linalg/src/sparse/4C_linalg_sparsematrix.cpp
```diff
@@ -1568,18 +1517,56 @@ void Core::LinAlg::SparseMatrix::replace_global_values(
...
+void Core::LinAlg::SparseMatrix::replace_or_insert_global_values(
     int global_row, int num_entries, const double* values, const int* indices)
 {
-  return sysmat_->SumIntoGlobalValues(global_row, num_entries, values, indices);
+  const int errone = sysmat_->ReplaceGlobalValues(global_row, num_entries, values, indices);
+  if (errone > 0)
+  {
+    [[maybe_unused]] int errtwo =
+        sysmat_->InsertGlobalValues(global_row, num_entries, values, indices);
+    FOUR_C_ASSERT(errtwo >= 0, "Epetra error (code {}).", errtwo);
```
mayrmt: Shouldn't this also use ASSERT_EPETRA_CALL()?

mayrmt (dismissed):

isteinbrecher:
Does this also work for FE assembly?

src/core/linalg/src/sparse/4C_linalg_sparsematrix.cpp
maxfirmbach: We are running into #1544 here. If I use `ASSERT_EPETRA_CALL()` we get a ton of warnings related to missing allocations. Thus for now, we allow warnings, but error out for errors.

src/core/linalg/src/sparse/4C_linalg_sparsematrix.cpp
maxfirmbach: Same as above.

src/core/linalg/src/sparse/4C_linalg_sparsematrix.cpp
maxfirmbach: Same as above.

maxfirmbach:
@isteinbrecher FE assembly is its own story, will be tackled in a separate PR. The current changes are solely for `Epetra_CrsMatrix`.

maxfirmbach:
@isteinbrecher Actually one of the `fe_assembly` methods uses this mechanism under the hood ... there are a lot of weird things going on in the `SparseMatrix` interface ...

PR #1548, opened 2025-12-01
Rewrite muscle combo material using tensors

lauraengelhardt:
## Description and Context
In this PR I rewrote the stress and material tangent computation of the MuscleCombo material using the `SymmetricTensors`. 

Note that `Core::LinAlg::FourTensorOperations::add_derivative_of_inva_b_inva_product` will be deleted once I port the other muscle materials.  `Core::LinAlg::FourTensorOperations::derivative_of_inva_b_inva_product` will be the function that stays. 

#862

### DIFF
--- a/src/mat/4C_mat_muscle_weickenmeier.cpp
+++ b/src/mat/4C_mat_muscle_weickenmeier.cpp
@@ -11,10 +11,9 @@
 #include "4C_global_data.hpp"
 #include "4C_linalg_fixedsizematrix_tensor_derivatives.hpp"
 #include "4C_linalg_fixedsizematrix_tensor_products.hpp"
-#include "4C_linalg_fixedsizematrix_voigt_notation.hpp"
 #include "4C_linalg_symmetric_tensor.hpp"
 #include "4C_linalg_tensor.hpp"
-#include "4C_linalg_tensor_conversion.hpp"
+#include "4C_linalg_tensor_generators.hpp"
 #include "4C_mat_elast_aniso_structuraltensor_strategy.hpp"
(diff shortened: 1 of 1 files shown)


src/mat/4C_mat_muscle_combo.cpp
```diff
@@ -321,68 +297,55 @@ void Mat::MuscleCombo::evaluate(const Core::LinAlg::Tensor<double, 3, 3>* defgra
...
-  // adds scalar * (invC boeppel invC) to cmat, see Holzapfel2000, p. 254
-  Core::LinAlg::FourTensorOperations::add_holzapfel_product(
-      ccmat, invCv, -(J * expbeta - std::pow(detC, -kappa)));
-  // adds -expbeta*detC * dinvCLinvCdCv to cmats
-  Core::LinAlg::FourTensorOperations::add_derivative_of_inva_b_inva_product(
-      -expbeta * detC, invCv, invCLinvCv, ccmat);
-  // additional term for corrected derivative of strain energy function
-  ccmat.multiply_nt(dEta / (8 * lambdaM), Mv, Mv, 1.0);
-  ccmat.scale(gamma);
+  Core::LinAlg::SymmetricTensor<double, 3, 3> _stress =
```
c-p-schmidt: Couldn't you just assign it directly? Besides, the variables starting with _ (also below for the tangent) are rather uncommon, at least in 4C. So far, I only know this from Python projects.

```suggestion
  stress =
```

lauraengelhardt: You're completely right. Changed in the last version.

c-p-schmidt (approved):

PR #1549, opened 2025-12-01
Remove old beamcontact input parameters

davidrudlstorfer:
## Description and Context

This PR removes outdated beam contact parameters which are unused. After the removal, I plan to move the input files to the new framework

Follows #1175


davidrudlstorfer:
The following parameters are also up for discussion:

Do we really need `BEAMS_STRATEGY`? Only penalty is set, the other option is none and it's a deprecated selection.

The following parameters are not used in any input file:

`BEAMS_ENDPOINTPENALTY`
`BEAMS_GAPSHIFTPARAM`
`BEAMS_BASICSTIFFGAP`

Can these be removed as well? @knarfnitram @isteinbrecher

davidrudlstorfer:
(Locally all beam tests succeeded)

davidrudlstorfer:
As far as I can see in https://github.com/4C-multiphysics/4C/blob/c0fd0080c4d9d7053bc5e4db85f8971a00e8b68e/src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_params.cpp#L59

I can remove the BEAMS_STRATEGY as well, it is nowhere used and just throws an error if it's not penalty

I'll push it in a dedicated commit

knarfnitram:
`BEAMS_GAPSHIFTPARAM` is used within the penalty law, so i think even this is currently untested, it can stay. 
`BEAMS_BASICSTIFFGAP` just avoids adding some linearizations terms, if the contact is not yet activated, from my point of view this can be removed and simply add always the respective terms. 
 `BEAMS_ENDPOINTPENALTY` I'd also keep for the moment.

davidrudlstorfer:
@knarfnitram thanks for your input! Then this would be ready from my side :)

tests/input_files/beam3eb_static_contact_beamtosolidcontact_solidcontact.4C.yaml
```diff
@@ -51,8 +51,6 @@ IO/RUNTIME VTK OUTPUT/STRUCTURE:
   ELEMENT_GID: true
   ELEMENT_MAT_ID: true
 BEAM CONTACT:
-  BEAMS_STRATEGY: "Penalty"
-  MODELEVALUATOR: "Standard"
   BEAMS_BTBPENALTYPARAM: 0.1
   BEAMS_BTBLINEPENALTYPARAM: 0.1
   BEAMS_MAXDELTADISSCALEFAC: -1
```
knarfnitram: I think `BEAMS_MAXDELTADISSCALEFAC` can also be removed.

davidrudlstorfer: Good catch - true!

knarfnitram:
Thanks for cleaning it up @davidrudlstorfer - the input files look now much cleaner! :heart_eyes:

Issue #1400, opened 2025-10-16
[solver] Move internals to `Thyra` objects


maxfirmbach:
## Description

Currently the iterative solver specifically relies on the fact that we inherit our `SparseOperator` from an `Epetra_Operator` implementing `Apply()` and `ApplyInverse()` for the block matrix implementation.
By switching the solver internals to `Thyra` and wrap them into the correct data structure should avoid this issue and should make it possible to cut the inheritance to `Epetra_Operator`. In addition it would simplify the construction of the linear solver with `Stratimikos`. If implemented correctly we should be able to switch to `Tpetra` without touching the linear solver internals at all.

## Possible Solution and Definition of Done

- [ ] Replace the `epetra_operator()` method by `get_operator()` for `SparseMatrix` and `BlockSparseMatrix` and wrap the matrix into a `Thyra::LinearOp`
- [ ] Refactor the preconditioner classes to hold a `Thyra::LinearOp` as internal member instead of an `Epetra_Operator`
  - [x] Use `Thyra` for `Teko` based preconditioner
  - [ ] Use `Thyra` for `Ifpack` based preconditioner (see #766)
  - [ ] Use `Thyra` for `MueLu` based preconditioner
- [ ] Refactor the iterative solver to use `Thyra` based objects instead of `Epetra` ones

PR #970, opened 2025-07-04
[docker] Update Trilinos to 16.2.0 Release

maxfirmbach:
## Description and Context

This PR updates the internal Trilinos version used by Docker and our testing to release 16.2.0.

With this PR the **only** Trilinos version 4C supports is **16.2.0**. All other releases or intermediate
updates given by the hashes below are invalid from now on:
```terminal
ef517e44826ce3a9e0189325aec4eee6e1d4dafe 2025.5
06db4c850654feacabdaed61ee8308219266b6a5 2025.4
8cdf0375d278598c12edc3fd7059cb133c6f4426 2025.3
11a16481fa1af48e3cfdb0aa84c29617ce94479f 2025.2
1eab15637f2998d1e86fe127b78200f2c9687cb5 2025.1
```

This comes with the fact that Trilinos version 17.0 will not support anything related to `Epetra` anymore. We cut backwards compatibility to be able to close a few issues and make progress on some PRs related to `Tpetra`.

## Related Issues and Pull Requests

Closes #968 #675 #661 #192

PR #1550, opened 2025-12-01
[solver] Rework `setup` and `solve` input arguments

maxfirmbach:
## Description and Context

This PR moves the solution vector directly to the solve method as input argument. This allows to directly modify the solution vector on spot and therefore the respective member variable is removed. This also makes modifications to the solution vector easier and more explicit instead of holding a pointer to a pointer to a pointer that enters some function at some point somewhere before.

Related to  #1400


maxfirmbach:
Should be merged after #970.

Issue #1536, opened 2025-11-27
Clean up btb input params


knarfnitram:
## Description

For  `BEAMS_NEWGAP`  in https://github.com/4C-multiphysics/4C/issues/1426 and [#1459](https://github.com/4C-multiphysics/4C/pull/1459) a discussion came out, indicating that it does not resemble a good variable name. This seems to be the case for more of the beam contact parameter from the file `4C_beaminteraction_beam_to_beam_contact_params`. 

## Possible Solution and Definition of Done

- [ ] Identify and remove unused parameters from `4C_beaminteraction_beam_to_beam_contact_params` after the deletion of the old beam contact framework
- [ ] Discuss which renaming of parameter list (at least remove `BEAMS_` suffix)
- [ ] Create regex based python script to rename all variables and adapt all test cases
- [ ] Update `beamme` (https://github.com/beamme-py/beamme/blob/main/src/beamme/four_c/header_functions.py) 

## List of beam contact parameters
- `BEAMS_ADDITEXT`
- `BEAMS_BASICSTIFFGAP`
- `BEAMS_BOXESINOCT`
- `BEAMS_BTBLINEPENALTYPARAM`
- `BEAMS_BTBPENALTYPARAM`
- `BEAMS_BTSOL`
- `BEAMS_BTSPENALTYPARAM`
- `BEAMS_DAMPING`
- `BEAMS_DAMPINGPARAM`
- `BEAMS_DAMPREGPARAM1`
- `BEAMS_DAMPREGPARAM2`
- `BEAMS_DEBUG`
- `BEAMS_ENDPOINTPENALTY`
- `BEAMS_EXTVAL`
- `BEAMS_GAPSHIFTPARAM`
- `BEAMS_INACTIVESTIFF`
- `BEAMS_MAXDELTADISSCALEFAC`
- `BEAMS_MAXDISISCALEFAC`
- `BEAMS_NEWGAP` -> 
- [ ]  `PREVIOUS_NORMAL_FOR_GAP` 
- [ ]  `ENHANCED_GAP_FUNCTION `
- [ ] `BEAMS_HISTORY_GAP_FUNCTION`
- [ ] `BEAMS_HISTORY_GAP `
- [ ] `BEAMS_ENHANCED_GAP_FUNCTION`
- `BEAMS_NUMINTEGRATIONINTERVAL`
- `BEAMS_OCTREE`
- `BEAMS_PARSHIFTANGLE1`
[...]

PR #1551, opened 2025-12-01
Move beam to beam contact to beaminteraction

davidrudlstorfer:
## Description and Context

This PR moves the beam contact input from the old framework directly to the new one.

Follows #1175 and #1549


davidrudlstorfer:
Currently I am experiencing some linker errors but this is my basic plan FYI @isteinbrecher @knarfnitram

davidrudlstorfer:
So everything compiled but i have a bug somewhere...

Maybe @isteinbrecher @knarfnitram you could have a very quick look at the changes. I am kinda blind right now

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
```diff
@@ -317,9 +317,10 @@ void Solid::ModelEvaluator::BeamInteraction::set_sub_model_types()
   // ---------------------------------------------------------------------------
   // check for beam contact
   // ---------------------------------------------------------------------------
-  if (Teuchos::getIntegralValue<Inpar::BeamInteraction::Strategy>(
-          Global::Problem::instance()->beam_interaction_params().sublist("BEAM TO BEAM CONTACT"),
-          "STRATEGY") != Inpar::BeamInteraction::bstr_none or
+  if ((Global::Problem::instance()
+              ->beam_interaction_params()
+              .sublist("BEAM TO BEAM CONTACT")
+              .numParams() > 0) or
```
davidrudlstorfer: @knarfnitram probably this is the problem i am currently experiencing

we no longer have the parameter strategy which will be set

is there another way to check if beam contact is active? The Parameters are always > 0 because the defaults are set...

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_params.cpp
```diff
@@ -155,10 +157,13 @@ void BeamInteraction::BeamToBeamContactParams::init()
   /****************************************************************************/
   // for the time being only allow all-angle-beam contact formulation ...
   if (not beam_contact_params_list.get<bool>("BEAMS_SEGCON"))
+  {
+    beam_contact_params_list.print(std::cout, 2, true, true);
```
davidrudlstorfer: personal TODO => remove print tomorrow

davidrudlstorfer:
@isteinbrecher @knarfnitram this PR would now be ready for review

- This PR now moves the cleaned-up beam to beam contact parameters to the beam interaction framework
- Renames the `BEAM CONTACT` section to `BEAM INTERACTION/BEAM TO BEAM CONTACT`
- Cleanes-up imports

Based on the renaming of the section this is a breaking change.

The next steps would be to now also finally move the other beam interaction input from inpar to beam interaction. But these are only internal changes without any outward facing breaking changes.

Therefore, #1536 can now be easily fixed.

Nevertheless, I would suggest when tackling #1536 to think about also updating all beam interaction parameters. For example the "fake" subsections, i.e., `BEAM INTERACTION/BEAM TO BEAM CONTACT/RUNTIME VTK OUTPUT`, should just be a real subsection within the beam contact section.

Also it would be pretty straight forward to directly change the parameters to snake case if all need to be touched anyways.

PR #1561, opened 2025-12-02
Setup peridynamic interaction

georghammerl:
This is the initial draft for implementing peridynamic interaction within the particle-based SPH framework. The methodological work is by @armanshojaei

Peridynamic in a nutshell:
Peridynamics is a nonlocal continuum mechanics theory that replaces classical partial differential equations with integral equations, making it well-suited for modeling discontinuities like cracks. Unlike traditional continuum methods, it treats material points as interacting through bonds over a finite distance, similar to particles. This approach conceptually relates to particle methods like DEM (Discrete Element Method), but while DEM is purely particle-based for granular materials, peridynamics retains continuum properties while enabling fracture and damage simulation without special crack-tracking algorithms.

The main idea of the current implementation is as follows:
- A peridynamic body consists of several rigid bodies (assumption: each rigid body consists of a single rigidphase particle) which are tied together via peridynamic interaction forces
- A peridynamic body can fracture and even break into smaller pieces
- Breakage between interacting bonds occurs if a certain stretch is exceeded
- Peridynamic bodies interact with each other (and also with walls and other rigid bodies) in a DEM-like fashion

The code changes can be summarized as follows:
- Insert a new handler for peridynamic in the SPH framework in `4C_particle_interaction_sph.cpp`
- Computation of peridynamic interaction force and care of the peridynamic bonded pairs is taken in `4C_particle_interaction_sph_peridynamic.hpp`
- To distinguish between "normal" DEM-like particle contact scenario (=short range interaction) and the peridynamic interaction scenario (=long range interaction), an additional pair is introduced: `4C_particle_interaction_pd_neighbor_pairs.hpp` which is capable of excluding the peridynamic pairs from the colliding pairs.
- In `4C_particle_interaction_sph_rigid_particle_contact.cpp`, the condition for evaluation of collisions relies on the `initialparticlespacing` (in the standard SPH case) which needs further clarification in the peridynamic contact scenario (WIP)

Update 2026-01-09: first step to encapsulate peridynamic implementation better
- Peridynamic bodies are no longer based on rigid body implementation
- `4C_particle_interaction_sph_rigid_particle_contact.cpp` is reset to original state, no more changes are necessary

Update 2026-01-11: second step to encapsulate peridynamic implementation better
- `PDNeighborPairs` derive no longer from `DEMNeighborPairs` 
- Reduced changes in `4C_particle_algorithm.cpp` and `4C_particle_algorithm_timint.cpp` to a minimum

Update 2026-01-12: third step to encapsulate peridynamic implementation better
- Introduce `Particle::PDPhase` to be independent of the `Particle::RigidPhase` 
- No more changes in `4C_particle_algorithm.cpp` and `4C_particle_algorithm_timint.cpp` necessary

Update 2026-01-15: fourth step to encapsulate peridynamic implementation better
- Use separate input parameters for the contact stiffness and damping to be independent of rigid body input parameters

Update 2026-01-19: fifth step to encapsulate peridynamic implementation better
- Use separate `PDParticlePairData` and `PDParticleWallPairData` to be more independant of the DEM code

## Interested groups
@ppraegla @slfuchs @armanshojaei


src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
```diff
@@ -0,0 +1,102 @@
...
+    //! communicator
+    const MPI_Comm& comm_;
+
+    //! unpack peridynamic bondlist data
+    void unpack_peridynamic_bond_list_data(const std::vector<char>& buffer);
+
+    //! reference to bond list
+    std::shared_ptr<
+        std::list<std::pair<Particle::LocalGlobalIndexTuple, Particle::LocalGlobalIndexTuple>>>
+        bondlist_;
```
ppraegla, line 107: Why do you use a `std::list`? If I saw it correctly, you only insert elements at the back. So, a `std::vector` ~~is~~ should be more efficient. But you can also check the performance later.

The complexity for  `std::vector::push_back` is amortized constant. The vector grows exponentially, meaning that it is not resized for every new element added. It is contiguous in memory, making it faster to iterate over. Therefore, a vector would be more suitable in this case. But I would be interested in a performance comparison.

src/particle/src/algorithm/4C_particle_algorithm_result_test.cpp
```diff
@@ -177,6 +177,15 @@ void Particle::ParticleResultTest::test_special(
         else if (quantity == "tempgradz")
           dim = 2;
       }
+      // phi
+      else if (quantity == "phi")
```
ppraegla: How about 
```suggestion
      else if (quantity == "pd_damage_phi")
```
to be more descriptive?

src/particle/src/interaction/4C_particle_interaction_dem_neighbor_pairs.cpp
```diff
@@ -82,6 +82,10 @@ void Particle::DEMNeighborPairs::evaluate_particle_pairs()
     int particle_j;
     std::tie(type_j, status_j, particle_j) = potentialneighbors.second;
 
+    if ((type_i != Particle::RigidPhase and type_i != Particle::BoundaryPhase) or
+        (type_j != Particle::RigidPhase and type_j != Particle::BoundaryPhase))
+      continue;
+
```
ppraegla: This check should not be in here. DEM only knows `phase1` and `phase2`.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
```diff
@@ -0,0 +1,102 @@
...
+{
+  class DiscretizationReader;
+}
+
+/*---------------------------------------------------------------------------*
+ | class declarations                                                        |
+ *---------------------------------------------------------------------------*/
+namespace Particle
+{
+  class PDNeighborPairs final : public DEMNeighborPairs
```
ppraegla: It is not a good idea to inherit from DEMNeighborPairs. This creates a coupling between PD and DEM. It is better to have some duplicate code instead, like it is done for SPHNeighborPairs and DEMNeighborPairs. In the future one could check if the common functionality could be extracted into a free function that can be reused.

src/particle/src/interaction/4C_particle_interaction_sph_rigid_particle_contact.cpp
ppraegla, line 1: As discussed, you want a different behavior for the "rigid" bodies in peridynamics. Your bodies are able to deform and the SPH bodies are rigid. So it is best to create a new class that handles the PD bodies.

slfuchs:
Same as @ppraegla wrote above, I would propose to strictly keep DEM, SPH and PD separated. This means one has to introduce a new case for peridynamics `case Particle::interaction_pd`. In addition, the particle algorithm should not call peridynamic specific functions. Instead, the particle algorithm calls methods from the particle interaction base class and everything that is pd specific is defined within the derived PD interaction class.

src/particle/src/algorithm/4C_particle_algorithm.cpp
```diff
@@ -133,8 +133,12 @@ void Particle::ParticleAlgorithm::setup()
   // build global id to local index map
   particleengine_->build_global_id_to_local_index_map();
 
-  // build potential neighbor relation
-  if (particleinteraction_) build_potential_neighbor_relation();
+  // build potential neighbor relation & build the peridynamic bond list
+  if (particleinteraction_)
+  {
+    build_potential_neighbor_relation();
+    if (not isrestarted_) particleinteraction_->build_peridynamic_relation();
```
slfuchs: At this level only potential neighbor pairs are evaluated, that is particle to particle and particle to wall neighbors. If the bin size is larger than the interaction radius, this method must not be called at every time step resulting in shorter runtime. The actual neighbor pairs are build within the particle interaction itself. I assume that the peridynamic relations could be evaluated within a peridynamic interaction type at a lower level?

src/particle/src/rigidbody/4C_particle_rigidbody.cpp
```diff
@@ -59,7 +59,7 @@ void Particle::RigidBodyHandler::setup(
     Particle::ParticleContainerBundleShrdPtr particlecontainerbundle =
         particleengineinterface_->get_particle_container_bundle();
 
-    if (not particlecontainerbundle->get_particle_types().count(Particle::RigidPhase))
+    if (not particlecontainerbundle->get_particle_types().contains(Particle::RigidPhase))
```
slfuchs, line 62: Good catch, one could replace `count` with `contains` also throughout the remainder of the particle code for std::set.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
georghammerl: Good point, this has even enabled further simplifications of the code.

georghammerl: Good point, this has even enabled further simplifications of the code.

src/particle/src/algorithm/4C_particle_algorithm_result_test.cpp
georghammerl: changed it

src/particle/src/interaction/4C_particle_interaction_dem_neighbor_pairs.cpp
georghammerl: True, removed it.

src/particle/src/interaction/4C_particle_interaction_sph_rigid_particle_contact.cpp
georghammerl: Peridynamic body is no longer comprised of rigid bodies. It is now based on rigidphase particles. Rigid body motion cannot be combined with peridynamic interaction for now.

src/particle/src/rigidbody/4C_particle_rigidbody.cpp
georghammerl: Thanks. I will postpone this to a separate PR. This PR is already very large. I do not want to touch unnecessary files.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
georghammerl: Copied code to resolve undesired code coupling (git tells me about a similarity index of 58% of the files). Let us see how we can unify better.

src/particle/src/algorithm/4C_particle_algorithm.cpp
georghammerl: Removed this line and shifted it to the proper place in `set_initial_states`.

georghammerl:
> Same as @ppraegla wrote above, I would propose to strictly keep DEM, SPH and PD separated. This means one has to introduce a new case for peridynamics `case Particle::interaction_pd`.

Yes and no. As the planned application targets PD bodies submerged to SPH fluid, it is essential to enable the coupling (this PR draft is the proof of concept for this). However, for standalone peridynamic problems, a `case Particle::interaction_pd` is the better choice to have a clearer path through the code. 

> In addition, the particle algorithm should not call peridynamic specific functions. Instead, the particle algorithm calls methods from the particle interaction base class and everything that is pd specific is defined within the derived PD interaction class.

@slfuchs: Thank you for this comment, this made the intendend structure of the code clearer to me. I changed the implementation accordingly.

src/particle/src/algorithm/4C_particle_algorithm.cpp
```diff
@@ -953,8 +953,11 @@ void Particle::ParticleAlgorithm::set_gravity_acceleration()
   // iterate over particle types
   for (auto& typeEnum : particlecontainerbundle->get_particle_types())
   {
-    // gravity is not set for boundary or rigid particles
-    if (typeEnum == Particle::BoundaryPhase or typeEnum == Particle::RigidPhase) continue;
+    // gravity is not set for boundary particles
+    if (typeEnum == Particle::BoundaryPhase) continue;
+
+    // rigid body handler takes care of gravity itself
+    if (typeEnum == Particle::RigidPhase and particlerigidbody_) continue;
```
slfuchs: @georghammerl Is it possible to add a new phase called `Particle::PDPhase` for the particles that belong to a Peridynamics Body/Continuum? Or are there other reasons why it is important that the new PD particles are also of the phase `Particle::RigidPhase`? As far as I can see it, this would probably easen some parts of this PR ... but it may be that I am overlooking something.

georghammerl: Yes, it is. I changed in the second commit to a newly introduced `Particle::PDphase`. I would like to include interaction between SPH fluid and the new pdhpase. I am unsure about the changes in the surface tension files whether they are correct? Might you please give me a hint @slfuchs @ppraegla

slfuchs: Thanks @georghammerl! I think the changes for the surface tension files should be fine. From what I see, the idea is that a peridynamics body behaves (as seen from the fluid phase) similar as a boundary phase or a rigid phase in terms of momentum, heat, and surface tension. However, within a peridynamics body different physics is applied, thus it is definitely good to introduce the new phase.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
georghammerl: Let us leave this for the moment and postpone to a separate PR.

slfuchs:
Can we set the team::particle label for this PR and also future pull requests?

slfuchs (approved):

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
```diff
@@ -0,0 +1,586 @@
...
+    if (not is_valid_peridynamic_bond_entry(particle_i, globalid_i, container_i) or
+        not is_valid_peridynamic_bond_entry(particle_j, globalid_j, container_j))
+    {
+      (*bondlist_)[iter] = std::move(bondlist_->back());
+      bondlist_->pop_back();
+      continue;
+    }
+
+    ++iter;
+  }
```
ppraegla, line 219: The while loop could be simplified using [`std::erase_if`](https://en.cppreference.com/w/cpp/container/vector/erase2), e.g., something like



```suggestion
  std::erase_if(*bondlist_,
      [&](auto& particlepair)
      {
        auto [type_i, status_i, particle_i, globalid_i] = particlepair.first;
        auto [type_j, status_j, particle_j, globalid_j] = particlepair.second;

        // one of the bond list partners may no longer reside on this proc
        if (particle_i == -1 or particle_j == -1) return true;

        // get corresponding particle containers
        Particle::ParticleContainer* container_i =
            particlecontainerbundle_->get_specific_container(type_i, status_i);

        Particle::ParticleContainer* container_j =
            particlecontainerbundle_->get_specific_container(type_j, status_j);

        // check for valid bond if one of its particles has left the processor
        return not is_valid_peridynamic_bond_entry(particle_i, globalid_i, container_i) or
            not is_valid_peridynamic_bond_entry(particle_j, globalid_j, container_j);
      });
```

This makes it clearer what you intent. Though, I'm unsure if the structured bindings make the code less readable concerning the types.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
```diff
@@ -0,0 +1,586 @@
...
+#endif
+
+    // gap between particle and wall contact point
+    const double gap = absdist - rad_i[0];
+
+    // neighboring particle and wall element within interaction distance
+    if (gap < 0.0)
+    {
+      // initialize particle-wall pair
+      particlewallpairdata_.push_back(DEMParticleWallPair());
```
ppraegla: Can you introduce a new `PDParticleWallPair` (and `PDParticlePair`) struct?
They will have almost the same data. But seeing DEMParticlePair made me wonder where DEM is used.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
```diff
@@ -0,0 +1,127 @@
...
+    //! unpack peridynamic bond list data
+    void unpack_peridynamic_bond_list_data(const std::vector<char>& buffer);
+
+    //! reference to bond list
+    std::shared_ptr<
+        std::vector<std::pair<Particle::LocalGlobalIndexTuple, Particle::LocalGlobalIndexTuple>>>
+        bondlist_;
+
+    //! map the hashkey to its corresponding entry in the bond list
+    std::unique_ptr<std::map<long, size_t>> map_hashkey_to_bondlist_entry_;
```
ppraegla: This does not need to be a pointer, correct? You directly initialize the variable in the constructor.

src/particle/src/interaction/4C_particle_interaction_sph.hpp
```diff
@@ -166,6 +171,9 @@ namespace Particle
     //! neighbor pair handler
     std::shared_ptr<Particle::SPHNeighborPairs> neighborpairs_;
 
+    //! neighbor pair handler for peridynamic phase particles
+    std::shared_ptr<Particle::PDNeighborPairs> neighborpairs_solid_type_;
```
ppraegla: What do you mean by solid_type? Why not 
```suggestion
    std::shared_ptr<Particle::PDNeighborPairs> neighborpairs_pd_;
```

src/particle/src/interaction/4C_particle_interaction_sph.cpp
```diff
@@ -320,6 +343,31 @@ void Particle::ParticleInteractionSPH::set_initial_states()
...
+      container->update_state(0.0, Particle::ReferencePosition, 1.0, Particle::Position);
+
+      // get material for current particle type
+      const Mat::PAR::ParticleMaterialPD* material =
+          dynamic_cast<const Mat::PAR::ParticleMaterialPD*>(
+              particlematerial_->get_ptr_to_particle_mat_parameter(type_i));
+
+      // set Young's modulus for all peridynamic phase particles
+      std::vector<double> young(1);
+      young[0] = material->young_;
```
ppraegla, line 359: Why do you set the Young's modulus for all particles if it is the same? Same for the critical stretch below

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+  auto normalcontacttype = Teuchos::getIntegralValue<Particle::NormalContact>(
+      particle_params.sublist("PD"), "NORMALCONTACTLAW");
+
+  if (normalcontacttype == Particle::NormalLinSpring and damp_ != 0.0)
+    FOUR_C_THROW(
+        "NORMAL_DAMP in PARTICLE DYNAMIC/PD section needs to be 0.0 if NORMALCONTACTLAW is set "
+        "to NormalLinearSpring!");
+
+  if (horizon_pd_ == 0.0)
```
ppraegla:
```suggestion
  if (horizon_pd_ <= 0.0)
```
You might also use `FOUR_C_ASSERT_ALWAYS`. But I plan to clean this up in the particle code anyway.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+void Particle::SPHPeridynamic::init(
+    const std::shared_ptr<Particle::PDNeighborPairs> neighborpairs_solid_type)
+{
+  // init with particle phase relevant in peridynamic interaction
+  pdtypes_ = {Particle::PDPhase};
+
+  // set bond list to pd neighbor pairs here because read_restart requires already bond list
+  std::static_pointer_cast<Particle::PDNeighborPairs>(neighborpairs_solid_type)
+      ->set_bond_list(bondlist_);
```
ppraegla: The static_pointer_cast is not necessary anymore.

```suggestion
  neighborpairs_solid_type->set_bond_list(bondlist_);
```

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -0,0 +1,454 @@
...
+  const double minimum_bin_size = binning_params.get<double>("BIN_SIZE_LOWER_BOUND");
+  if (horizon_pd_ > minimum_bin_size)
+    FOUR_C_THROW("Peridynamic INTERACTION_HORIZON must be smaller than BIN_SIZE_LOWER_BOUND!");
+}
+
+void Particle::SPHPeridynamic::init(
+    const std::shared_ptr<Particle::PDNeighborPairs> neighborpairs_solid_type)
+{
+  // init with particle phase relevant in peridynamic interaction
+  pdtypes_ = {Particle::PDPhase};
```
ppraegla: Can you set this in the constructor? In the long run, we want to get rid of the init and setup methods

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -0,0 +1,454 @@
...
+    if (stretch < critical_stretch)
+    {
+      double m[3];
+      ParticleUtils::vec_set_scale(m, 1.0 / xi_eta_norm, xi_eta);
+
+      // calculate the bond force of the pair
+      double const fac = (12.00 * (young_i[0] + young_j[0]) * 0.5) /
+                         (M_PI * horizon_pd_ * horizon_pd_ * horizon_pd_ * horizon_pd_) * stretch *
+                         calculate_volume_correction_factor(xi_norm) * dx_pd_ * dx_pd_ * dx_pd_ *
+                         dx_pd_ * dx_pd_ * dx_pd_;
```
ppraegla:
```suggestion
      double const fac = (12.00 * (young_i[0] + young_j[0]) * 0.5) /
                         (std::numbers::pi * horizon_pd_ * horizon_pd_ * horizon_pd_ * horizon_pd_) * stretch *
                         calculate_volume_correction_factor(xi_norm) * dx_pd_ * dx_pd_ * dx_pd_ *
                         dx_pd_ * dx_pd_ * dx_pd_;
```
std::numbers::pi is from the `<numbers>` header

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
georghammerl: The structured binding hides away the types which is indeed for new people to the code more complex to read.

As order does not matter in the `bondlist_`, I prefer to stick to the faster variant with swap-and-pop instead of the order-preserving (but slower) version with `std::erase_if`.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
georghammerl: True

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
georghammerl: Removed it, ist not necessary, copy-paste remnant.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
georghammerl: Good idea!

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
georghammerl: Done.

src/particle/src/interaction/4C_particle_interaction_sph.cpp
georghammerl: True, however the standard case (in future ) will be a inhomogeneous field of those quantities. Hence, this is already prepared here.

georghammerl:
@ppraegla @slfuchs Thank you for all your comments! I squashed everything.

slfuchs (approved):

slfuchs:
Seems like my approval is not sufficient 😆

slfuchs:
I am wondering why this review / approval settings are tied to a employment contract at one of the partnering institutes. Wouldn't it make sense to also include other "experienced" developers?

georghammerl:
> I am wondering why this review / approval settings are tied to a employment contract at one of the partnering institutes. Wouldn't it make sense to also include other "experienced" developers?

It is not related to some contract, this just follows our community roles,
see https://4c-multiphysics.github.io/4C/documentation/community/community_roles.html#c-developer .
As you are quite active, developer status is appropriate.

slfuchs:
Thanks for the clarification 👍

PR #1562, opened 2025-12-02
Rewrite Giantesio muscle material using tensors

lauraengelhardt:
## Description and Context
In this PR I rewrote the stress and material tangent computation of the muscle material by Giantesio et al. using the `SymmetricTensors`.

#862


src/mat/4C_mat_muscle_giantesio.cpp
```diff
@@ -26,190 +27,79 @@ FOUR_C_NAMESPACE_OPEN
...
+    const double Je = detCe * Core::LinAlg::trace(invCe * L);
+
     // exponential prefactors
-    double expalpha = std::exp(alpha * (Ie - 1.0));
-    double expbeta = std::exp(beta * (Je - 1.0));
+    const double expalpha = std::exp(alpha * (Ie - 1.0));
+    const double expbeta = std::exp(beta * (Je - 1.0));
 
     // elastic second Piola Kirchhoff stress tensor Se
     // Se = (0.5 * gamma * (expalpha * L + expbeta * (Je * invCe - invCeLinvCe)))
```
c-p-schmidt: Is the comment wrong? I think the implementation correctly replaces the original implementation (I'm referring to the `detCe`.

By the way, since the code so nicely expresses the equation, I don't think that the comment is necessary anymore.

src/mat/4C_mat_muscle_giantesio.cpp
```diff
@@ -26,190 +27,79 @@ FOUR_C_NAMESPACE_OPEN
...
-    Core::LinAlg::FourTensorOperations::add_holzapfel_product(dSedCev, invCev, -Je * expbeta);
-    // adds -expbeta * detCe * dinvCLinvCdCv to cmat
-    Core::LinAlg::FourTensorOperations::add_derivative_of_inva_b_inva_product(
-        -expbeta * detCe, invCev, invCeLinvCev, dSedCev);
-    dSedCev.scale(gamma / 2);
-
-    Core::LinAlg::FourTensor<3> dSedCe(true);
-    Core::LinAlg::Voigt::setup_four_tensor_from_6x6_voigt_matrix(dSedCe, dSedCev);
+    auto dSedCe = alpha * expalpha * Core::LinAlg::dyadic(L, L);
+    dSedCe += beta * expbeta * std::pow(detCe, 2) * Core::LinAlg::dyadic(invCeLinvCe, invCeLinvCe);
```
c-p-schmidt: You could think about replacing the std::pow
```suggestion
    dSedCe += beta * expbeta * detCe * detCe * Core::LinAlg::dyadic(invCeLinvCe, invCeLinvCe);
```

src/mat/4C_mat_muscle_giantesio.cpp
lauraengelhardt: The comment is wrong, thanks! I'll remove it.

PR #1563, opened 2025-12-02
[fluid] Remove two tests

maxfirmbach:
## Description and Context

This PR removes two fluid based test cases from our 4C testing.
- `f3_forced_hit_geogen_hdg.4C.yaml` is actually executed in serial because it does not work in parallel. Even in the serial case I'm convinced it does not do the correct things, e.g. finding the correct faces for the discontinuous Galerkin method etc. -> thus removal.
- `f3_channel_mixedhybrid.4C.yaml` has some weird/random behavior to it as (almost) all turbulence test. It would make sense to make this test smaller and use a direct solver, tighten tolerances etc. I don't know how to do it and I am defenitly not an expert on this topic -> thus removal.

Im convinced that there are actually a few more test cases that do not work properly.


maxfirmbach:
@ischeider Is there any specific reason why we use `f3_forced_hit_geogen_hdg.4C.yaml` in the documentation?

maxfirmbach:
I updated the documentation with a better fluid test case that also uses the "on the fly cube discretization generation" feature of `4C`, which is the main reason why the removed test case was referenced in the documentation in the first place.

Issue #503, opened 2025-03-13
Remove LNM presets


davidrudlstorfer:
## Description

Based on #499 we decided that we want to remove the institute specific cmake presets from the main repo. We decided to remove the LNM presets once the upgrade to Ubuntu 24.04 is finished (as this breaks the current setup).

In the future each user provides her/his own presets. @ppraegla also wrote in #499 that the documentation will be updated to ease the setup for each user

## Possible Solution and Definition of Done

- [ ] remove the LNM cmake presets
(- [ ] also improve the documentation)

sbrandstaeter:
Making @gilrrei aware of this as we rely on the presets being available in the cluster pipelines of QUEENS.

PR #1564, opened 2025-12-02
feat: update Trilinos in LNM presets to 2025-6

davidrudlstorfer:
## Description and Context

Updates the LNM presets with internal Trilinos version 2025-6

**ONLY MERGE AFTER #970**

#970


amgebauer (dismissed):
4C also builds without #970 , so we could merge before (at least on my workstation)

maxfirmbach (approved):

maxfirmbach:
I think this can be merged.

amgebauer:
@davidrudlstorfer Are you reverting the lnm workstation presets, or should I do it?

davidrudlstorfer:
To document it here and as discussed internally

=> the workstation presets in 4C for the LNM are for older workstations on Ubuntu 20.04. This does not work with Trilinos 2025.6

=> we only update the cluster preset and, as before, each user must update her/his preset independently

davidrudlstorfer:
And this should be tackled soonish with #503

Issue #136, opened 2024-12-09
[EpetraExt] Evaluate the sparse matrix addition methods


eulovi:
> Originating from https://gitlab.lrz.de/baci/baci/-/issues/1316

## Description

We need to remove most of the things related to `Epetra` from our matrix addition implementation e.g. function input etc.

Currently we implement the method to add two matrices together by ourselves. Maybe it would make sense to use `EpetraExt::MatrixMatrix::Add`, similar to the matrix multiplication methods.

## Possible Solution and Definition of Done

Refactor:
- [x] Remove `Epetra_CrsMatrix` from function calls and replace by e.g. `Core::LinAlg::SparseMatrix` (see #1152)
- [ ] Remove weird member function structure of `add()` method of the matrix implementations and make it a free function if possible. We already provide a `Core::LinAlg::add()` free function. It would make sense to rename it to `Core::LinAlg::matrix_add()` in a first step, and later use it troughout the code instead of the member function `add()` from `SparseMatrix` etc.

## Related Issues and Merge Requests

* Related to #189 

## Additional Information

In the long term it would make sense to reduce self-implementations for core functionality to a minimum, especially if it's already available.

maxfirmbach:
We should think about evaluating this, but it can wait till we are on `Tpetra`.

maxfirmbach:
@eulovi Are you still working on this? Or do you want to un-assign you?

maxfirmbach:
I can have a look at moving the `add()` methods to a free function and call that instead of a entangled member function pattern, if necessary.

eulovi:
@maxfirmbach I can take a look at this. Some time ago I tried to replace all the calls of add() with the free function version and some things didn't work. I will revise this and move the `add()` methods to free functions.

eulovi:
In an offline discussion with @maxfirmbach, the deletion of the add() function is not straightforward, as through the code this function is called from `SparseOperator` instances and not directly from the classes where its properly implemented and used (namely, `SparseMatrix` and `BlockSparseMatrixBase`). 
Therefore, we can close this issue, as the code is cleaner now.

PR #1568, opened 2025-12-03
[NOX] Inherit matrix-free operators from `SparseOperator`

vovannikov:
## Description and Context
In order to proceed with removal of the base inheritances of `NOX::Nln::Interface::RequiredBase` and `NOX::Nln::Interface::JacobianBase` quite substantial modifications are required. This is just one of them and it effectively removes `Epetra_Operator` yet from one more place. It is also required for #1524.

A temporary wrapper `NOX::Nln::MatrixFree::SparseOperatorWrapper` is introduced here. Ultimately, it will be either simplified or completely removed later.

While working on this PR, I noticed that the default interface of `Core::LinAlg::SparseOperator` is excessive: most of the methods should be removed. For instance, 
- `is_dbc_applied()` - it is very specific and is used in a single place of the code.
- `add_other()` - it is odd and not used at all.
This is just what I identified by a quick look.

#1524, #1535, #1486

### DIFF
--- a/src/fsi/src/monolithic/nonlinear_solver/4C_fsi_nox_linearsystem_gcr.cpp
+++ b/src/fsi/src/monolithic/nonlinear_solver/4C_fsi_nox_linearsystem_gcr.cpp
@@ -14,7 +14,6 @@
 #include "4C_utils_shared_ptr_from_ref.hpp"
 
 #include <Epetra_LinearProblem.h>
-#include <Epetra_Operator.h>
 #include <Teuchos_ParameterList.hpp>
 
 #include <vector>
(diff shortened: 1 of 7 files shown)


maxfirmbach:
@vovannikov I agree with the excessiveness of the `SparseOperator` interface. There are already a few issues to clean those things up e.g. #136 concerning the matrix addition methods and #863 as general effort to clean up our matrix implementations and interfaces.

maxfirmbach:
I think we should aim to get rid of `SparseOperatorWrapper` again at a later point in time if possible.

vovannikov:
> I think we should aim to get rid of `SparseOperatorWrapper` again at a later point in time if possible.

@maxfirmbach, I fully agree. To prevent its unintended spread around the code, I made it a nested private class, such that it is impossible to use it in any other instance.

PR #1569, opened 2025-12-03
Refactor is_dbc_applied into own function

knarfnitram:
## Description and Context

Moves the `is_dbc_applied` from the interface to an utils function. 

Related to https://github.com/4C-multiphysics/4C/pull/1568


src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_assemble.hpp
```diff
@@ -265,6 +265,32 @@ namespace Core::LinAlg
...
+
+  \param A (in) : SparseOperator A
+   \param dbcmap (in) : DBC map holding all dbc dofs
+   \param bool (in) diagonalblock: Is this matrix a diagonalblock of a blocksparsematrix?
+                              If it is only one block/matrix, this boolean should be TRUE.
+   \param trafo (in) : pointer to an optional trafo matrix (see LocSys).
+
+   */
+  bool is_dirichlet_boundary_condition_already_applied(const Core::LinAlg::SparseOperator& A,
+      const Core::LinAlg::Map& dbcmap, bool diagonalblock, const Core::LinAlg::SparseMatrix* trafo);
```
vovannikov: I would use here `const std::shared_ptr<const Core::LinAlg::SparseMatrix>&` instead of a raw pointer, especially considering that `get_loc_sys_trafo()` returns a shared pointer. The use of reference here is fine, since this function does not require ownership.

The same applies to another version of `is_dirichlet_boundary_condition_already_applied()`

knarfnitram: Thanks for the suggestion, that interface looks now better.

Issue #1566, opened 2025-12-02
[tutorial] FSI results in test timeout


maxfirmbach:
The testing of `fsi/tutorial_fsi_monolithic.4C.yaml-p2-fw (Timeout)` often results in timeouts. We should fix this.

Might be interesting for @mayrmt @reginabuehler

mayrmt:
I quickly looked into this. Running on more processors does not speed-up the simulation. The issue is that we run 5 time steps, where each step takes a lot of newton iterations. To cut simulation time, we either have to reduce the number of time steps or loosen Newton tolerances.

mayrmt:
We can loosen the tolerances for the pressure increment without affecting the results, yet saving 2-3 Newton iterations per time steps. This reduces the run time on my machine by 25%, thus, it should fix the problem. I'll open a PR in a minute.

PR #1571, opened 2025-12-03
Add option to increase timeout in tutorials

knarfnitram:
## Description and Context

Provide option to set a timeout in the cmake function `four_c_test_tutorial`.

https://github.com/4C-multiphysics/4C/issues/1566


knarfnitram:
I added a timeout value to the fsi tutorial test case.

davidrudlstorfer:
I am not completely sold on adding this feature because it fails on local workstations? Why then add it to the pipeline if it's not necessary? Or is the timeout also happening in the testing pipeline? Unfortunately, I do not really get that from #1566

knarfnitram:
So to me it seemed, that the fsi test case runs into a simple timeout within the pipeline and I noticed that for the cmake function `four_c_test_tutorial`, its currently not possible to increase the timeout of a single test case as it is for the standard `four_c_test` cmake-function. This MR adds the option to specify it. (As we can see from the failing pipeline by 75sec, the specification works, I just did not proper calculate the amount of time need for this test case).

reginabuehler (approved):

PR #1574, opened 2025-12-03
Evaluate BTB penalty parameters by a function

knarfnitram:
## Description and Context

Selecting a single value for the beam to beam penalty parameter can be cumbersome, this MR modifies the beam contact to specify the penalty parameters by a function instead of single value from the inputfile. 
Hence two new parameters, `BEAMS_BTBPENALTYPARAM_BY_FUNCT` and  `BEAMS_BTBLINEPENALTYPARAM_BY_FUNCT` are introduced, which evaluate the parameters within the contact pair at the specified time. 

Two test cases, one for the line and the other for the point penalty value are provided.


src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_beamcontact.cpp
```diff
@@ -1103,7 +1103,7 @@ void BeamInteraction::SubmodelEvaluator::BeamContact::create_beam_contact_elemen
 
       for (auto& pair : newbeaminteractionpairs)
       {
-        pair->init(beam_contact_params_ptr_, ele_ptrs);
+        pair->init(beam_contact_params_ptr_, ele_ptrs, g_state().get_time_n());
```
knarfnitram, line 1106: Here I am not so sure, if we really want to pass the time into the pair, and especially the previous or next time.

src/beaminteraction/src/contact/4C_beaminteraction_contact_beam_to_beam_input.cpp
```diff
@@ -42,6 +42,12 @@ std::vector<Core::IO::InputSpec> BeamInteraction::Contact::BeamToBeam::valid_par
           parameter<double>("BEAMS_BTBLINEPENALTYPARAM",
               {.description = "Penalty parameter per unit length for beam-to-beam line contact",
                   .default_value = -1.0}),
+          parameter<int>("BEAMS_BTBPENALTYPARAM_BY_FUNCT",
+              {.description = "Penalty parameter for beam-to-beam point contact by function id",
+                  .default_value = -1}),
```
davidrudlstorfer, line 47: Let's not add a default of -1 anymore. You can also make it a noneable default.

davidrudlstorfer: Also - would it make more sense to add a new "subsection" and then you chose if you either have a constant value or a function id?

Like it's done with the input fields. Because the parameter is somehow duplicated as far as I understand?

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_contact_params.cpp
```diff
@@ -83,20 +85,41 @@ void BeamInteraction::BeamToBeamContactParams::init()
...
+
+  if (btb_point_penalty_param_ < 0.0 and btb_function_id_for_point_penalty < 0.0)
+    FOUR_C_THROW(
+        "beam-to-beam point penalty parameter or beam-to-beam function id for point penalty must "
+        "be specified!");
+
+  if (btb_point_penalty_param_ > 0.0 and btb_function_id_for_point_penalty > 0.0)
+    FOUR_C_THROW(
+        "Please specify the beam-to-beam point penalty parameter either with BEAMS_BTBPENALTYPARAM "
+        "or BEAMS_BTBPENALTYPARAM_BY_FUNCT. Not both!");
```
davidrudlstorfer: If you already verify this during input it would probably make more sense

PR #1580, opened 2025-12-05
Remove object files after build in GitHub Actions

amgebauer:
Remove object files after building 4C in the pipeline. This should reduce the size of the build directory after the build. We copy the build directory from one stage to the other, so we should keep the amount of data that we copy small.


c-p-schmidt (approved):
Let's give it a try!

davidrudlstorfer (approved):

PR #1581, opened 2025-12-05
Enable vtu input by default on lnm cluster

amgebauer:
Enable vtu mesh input by default on LNM clusters.


davidrudlstorfer (approved):
Thanks a lot :)

PR #596, opened 2025-04-02
[Trilinos] Move from Isorropia to Zoltan2

maxfirmbach:
## Description and Context

This PR mainly consists of the first changes moving from `Isorropia` to `Zoltan2` and for now mostly acts to view code changes and observe failing tests.

Current problems:
- ~~It seems like the underlying hypergraph partitioning is somehow slightly different (https://github.com/trilinos/Trilinos/issues/14010)~~
- ~~Something related to periodic boundary conditions doesn't seem to work correctly~~
- ~~Nested parallelism seems to make some issues~~
- ~~Micro/macro tests fail~~

## Related Issues and Pull Requests

- Part of #189 
- Related to #594

PR #1583, opened 2025-12-05
[structure_new] Fix pure Neumann test

maxfirmbach:
## Description and Context

As observed by @amgebauer the constraint-based pure Neumann problem test case makes some problems. The rotational component of the nullspace is actually not constraint leading to this issue. In addition the projection based method is executed with two procs to make sure everything works properly in parallel.

Related to #596


amgebauer (approved):
Nice, thank you for fixing it 👍

eulovi (approved):

PR #1587, opened 2025-12-08
Use (Symmetric)Tensor in shell7p element evaluation

mairehenke:
## Description and Context
 This PR migrates the shell7p element evaluation to use (Symmetric)Tensor for stresses and strains and adapts the dependent computations.

In the second commit, a redundant stiffness matrix symmetrization was removed, as the computation only involves symmetric tensors/matrices and is therefore symmetric by default.

 #862 , #934

### DIFF
--- a/src/shell7p/4C_shell7p_ele_calc.cpp
+++ b/src/shell7p/4C_shell7p_ele_calc.cpp
@@ -457,16 +457,6 @@ void Discret::Elements::Shell7pEleCalc<distype>::evaluate_nonlinear_force_stiffn
           Shell::add_geometric_stiffness_matrix<distype>(shapefunctions_collocation,
               shape_functions_ans, shape_functions, stress_enh.stress_, shell_data_.num_ans,
               integration_factor, *stiffness_matrix);
-          // make stiffness matrix absolute symmetric
-          for (int i = 0; i < Shell::Internal::numdofperelement<distype>; ++i)
-          {
-            for (int j = i + 1; j < Shell::Internal::numdofperelement<distype>; ++j)
-            {
-              const double average = 0.5 * ((*stiffness_matrix)(i, j) + (*stiffness_matrix)(j, i));
-              (*stiffness_matrix)(i, j) = average;
-              (*stiffness_matrix)(j, i) = average;
-            }
-          }
         }
         // add internal force vector
         if (force_vector != nullptr)
--- a/src/shell7p/4C_shell7p_ele_calc_eas.cpp
+++ b/src/shell7p/4C_shell7p_ele_calc_eas.cpp
@@ -699,20 +699,6 @@ void Discret::Elements::Shell7pEleCalcEas<distype>::evaluate_nonlinear_force_sti
   {
     Shell::EAS::add_eas_internal_force(LinvDTilde, eas_iteration_data_.RTilde_, *force_vector);
   }
-
-  if (stiffness_matrix != nullptr)
-  {
-    // make stiffness matrix absolute symmetric
-    for (int i = 0; i < Shell::Internal::numdofperelement<distype>; ++i)
-    {
-      for (int j = i + 1; j < Shell::Internal::numdofperelement<distype>; ++j)
-      {
-        const double average = 0.5 * ((*stiffness_matrix)(i, j) + (*stiffness_matrix)(j, i));
-        (*stiffness_matrix)(i, j) = average;
-        (*stiffness_matrix)(j, i) = average;
-      }
-    }
-  }
 }
 
 


src/shell7p/4C_shell7p_ele_calc_lib.hpp
```diff
@@ -985,42 +955,20 @@ namespace Discret::Elements::Shell
...
-    Core::LinAlg::FourTensorOperations::inverse_fourth_tensor_rotation(
-        g_metrics_trans, stress.cmat_, Cmat);
-
-    // re-arrange indices for shell element formulation
-    static constexpr std::array voigt_inconsistent_ = {0, 3, 5, 1, 4, 2};
-    for (unsigned int i = 0; i < 6; ++i)
-      for (unsigned int j = 0; j < 6; ++j)
-        stress.cmat_(i, j) = Cmat(voigt_inconsistent_[i], voigt_inconsistent_[j]);
+    Core::LinAlg::Tensor<double, 3, 3> kontravariant_tensor =
+        Core::LinAlg::make_tensor(g_reference.kontravariant_);
```
amgebauer: fyi, you could also view the data of the other tensor to not copy them:


```suggestion
    Core::LinAlg::TensorView<const double, 3, 3> kontravariant_tensor =
        Core::LinAlg::make_tensor_view(g_reference.kontravariant_);
```

src/shell7p/4C_shell7p_ele_calc_lib.hpp
```diff
@@ -1359,18 +1282,26 @@ namespace Discret::Elements::Shell
    */
   template <Core::FE::CellType distype>
   void thickness_integration(Discret::Elements::Shell::StressEnhanced& stress_enh,
-      const Discret::Elements::Shell::Stress<Mat::NUM_STRESS_3D>& stress,
-      const double& integration_factor, const double& zeta)
+      const Discret::Elements::Shell::Stress& stress, const double& integration_factor,
+      const double& zeta)
   {
+    // shell-specific voigt ordering: {S11, S12, S13, S22, S23, S33}
+    static constexpr std::array<int, 6> voigt_inconsistent = {0, 3, 5, 1, 4, 2};
```
amgebauer, line 1288: Why do shell elements have a different voigt ordering? Can we unify it? Or better, replace all voigt notations with symmetric tensors 🤔

amgebauer (dismissed):
Nice @mairehenke! Thank you for migrating the shell elements to tensors 👍

src/shell7p/4C_shell7p_ele_calc_lib.hpp
mairehenke: The original shell formulation uses this "inconsistent Voigt" notation, so I stuck with it initially. We could unify it, but it’s a bit more complex and requires rewriting the EAS and ANS parts, so I plan to do it in a separate PR. I would definitely like to get rid of this.

mairehenke:
Thanks for your review, @amgebauer! I’ve just replaced the copies with views, as you suggested.

amgebauer (dismissed):

src/shell7p/4C_shell7p_ele_calc_lib.hpp
```diff
@@ -213,55 +205,34 @@ namespace Discret::Elements::Shell
    * @param defgrd_enh [in/out]: enhanced (consistent) deformation gradient
    */
   template <unsigned dim>
-  void calc_consistent_defgrd(const Core::LinAlg::Matrix<dim, dim>& defgrd_disp,
-      const Core::LinAlg::Matrix<dim*(dim + 1) / 2, 1>& glstrain_enh,
-      Core::LinAlg::Matrix<dim, dim>& defgrd_enh)
+  Core::LinAlg::Tensor<double, dim, dim> calc_consistent_defgrd(
+      const Core::LinAlg::Tensor<double, dim, dim>& defgrd_disp,
+      const Core::LinAlg::SymmetricTensor<double, dim, dim>& enhanced_gl_strain)
```
eulovi: I think the documentation should be updated here, as the enhanced deformation gradient is not passed into the function but is its output.

src/shell7p/4C_shell7p_ele_calc.cpp
```diff
@@ -295,12 +285,9 @@ void Discret::Elements::Shell7pEleCalc<distype>::calculate_stresses_strains(
...
           auto strains = Shell::evaluate_strains(g_reference, g_current);
 
-          // update the deformation gradient
-          Core::LinAlg::Matrix<Shell::Internal::num_dim, Shell::Internal::num_dim> defgrd_enh(
-              Core::LinAlg::Initialization::uninitialized);
-          Shell::calc_consistent_defgrd<Shell::Internal::num_dim>(
-              strains.defgrd_, strains.gl_strain_, defgrd_enh);
-          strains.defgrd_ = defgrd_enh;
+          // compute the consistent deformation gradient based on enhanced gl strains
+          strains.defgrd_ = Shell::calc_consistent_defgrd<Shell::Internal::num_dim>(
```
eulovi: I am not an expert on shells here and it's out of the scope of this PR, but isn't it a bit confusing to update the strains.defgrd_ with the enhanced deformation gradient? Wouldn't it be cleaner to have two variables? one for the deformation gradient and one based on the enhanced GL strains?

mairehenke: Good point! I also just noticed that we use this function in computations without EAS, where we don’t actually need it. I’ll open another PR to remove it there and introduce the enhanced strains and deformation gradients separately. Thanks for catching this!

mairehenke:
Thanks to both of you for the reviews, @amgebauer @eulovi ! I’ve updated the documentation and will open a new PR to address the inconsistent part with the enhanced deformation gradient and strains.

eulovi (approved):

PR #1588, opened 2025-12-09
Implement symmetric positive definite tensor interpolation with weights

amgebauer:
This PR implements a tensor interpolation of symmetric positive definite tensors. The interpolation is (in contrast to existing one from @dragos-ana) purely based on normalized weights.

I also rewrote some parts of the tensor interpolation algorithms based on Tensors (instead of Matrix).

I use it to interpolate a pure stretch tensor input field.


src/core/linalg/src/dense/4C_linalg_utils_tensor_interpolation.hpp
```diff
@@ -303,8 +376,147 @@ namespace Core::LinAlg
...
+          {
+            // flip the sign
+            for (std::size_t col = 0; col < dim; ++col) t(i, col) = -t(i, col);
+          }
+        }
+
+        // set the last eigenvector to ensure det(...) = 1 (cross-product of the first two)
+        t(2, 0) = t(0, 1) * t(1, 2) - t(0, 2) * t(1, 1);
+        t(2, 1) = t(0, 2) * t(1, 0) - t(0, 0) * t(1, 2);
+        t(2, 2) = t(0, 0) * t(1, 1) - t(0, 1) * t(1, 0);
```
tuchpaul, line 421: Maybe one could add an explicit normalization of the third eigenvector to ensure unit length, since prior operations, such as sign flips and the cross product, may introduce small numerical inaccuracies.

```suggestion
        double e2_0 = t(0, 1) * t(1, 2) - t(0, 2) * t(1, 1);
        double e2_1 = t(0, 2) * t(1, 0) - t(0, 0) * t(1, 2);
        double e2_2 = t(0, 0) * t(1, 1) - t(0, 1) * t(1, 0);
        
        const double norm_e2 = std::sqrt(e2_0 * e2_0 + e2_1 * e2_1 + e2_2 * e2_2);
        
        t(2, 0) = e2_0 / norm_e2;
        t(2, 1) = e2_1 / norm_e2;
        t(2, 2) = e2_2 / norm_e2;
```

src/core/linalg/src/dense/4C_linalg_utils_tensor_interpolation.hpp
```diff
@@ -65,6 +68,61 @@ namespace Core::LinAlg
     }
   }
 
+  /*!
+   * @brief Interpolate a symmetric positive-definite second-order tensor using given weights and
+   * tensors according to the R-LOG method.
+   *
+   * The eigenvalues of the stretch tensor are interpolated using the logarithmic weighted average.
+   * The rotation tensors are interpolated using rotation vector interpolation.
```
dragos-ana: Maybe we should call them eigenvector rotation tensors to avoid confusion.

src/core/linalg/src/dense/4C_linalg_utils_tensor_interpolation.hpp
```diff
@@ -65,6 +68,61 @@ namespace Core::LinAlg
...
+      return interpolate_spd(weights, tensors);
+    }
+  };
+
+  /*!
+   * @brief Interpolate a rotation tensors using normalized weights using rotation vector
+   * interpolation.
+   *
+   * @param weights normalized weights
+   * @param rotation_tensors reference rotation tensors to be interpolated
```
dragos-ana:
```suggestion
   * @param rotation_tensors reference rotation tensors to be used for interpolation
```

src/core/linalg/src/dense/4C_linalg_utils_tensor_interpolation.hpp
amgebauer: I don't think this is necessary since sign flips don't cause any floating point inaccuracies and the error from the cross-product should not be worse than the renormalization (which is also not exact)

src/core/linalg/src/dense/4C_linalg_utils_tensor_interpolation.hpp
```diff
@@ -303,8 +376,156 @@ namespace Core::LinAlg
...
+  std::vector<Core::LinAlg::Tensor<double, dim, dim>> relative_rotation_tensors(
+      rotation_tensors.size());
+
+  std::transform(rotation_tensors.begin(), rotation_tensors.end(),
+      relative_rotation_tensors.begin(), [&rotation_tensors, closest_index](const auto& R_i)
+      { return Core::LinAlg::transpose(rotation_tensors[closest_index]) * R_i; });
+
+  // Create rotation vector from relative rotation tensor
+  std::vector<Core::LinAlg::Tensor<double, dim>> rotation_vectors;
+  rotation_vectors.reserve(relative_rotation_tensors.size());
```
dragos-ana: We could call them `relative_rotation_vectors` for consistency, same for the interpolated one.

src/core/linalg/src/dense/4C_linalg_utils_tensor_interpolation.hpp
```diff
@@ -303,8 +376,156 @@ namespace Core::LinAlg
...
+    const auto [eigenvalues, eigenvectors] = Core::LinAlg::eig(t);
+
+    // store eigenvectors and eigenvalues (The rows should be the eigenvectors, hence transpose it
+    // here)
+    Q_i.push_back(Core::LinAlg::transpose(eigenvectors));
+    eigenvalues_i.push_back(eigenvalues);
+  }
+
+  // Note: At this stage, eigenvectors are not uniquely defined (sign ambiguity).
+  Internal::make_eigenvectors_unique(weights, Q_i);
```
dragos-ana, line 460: While `make_eigenvectors_unique` is required at this stage to flip eigenvector signs, I think we would also need special treatment in case of multiple eigenvalues prior to it. The main problem is that the eigenvalue decomposition does not also order the eigenvectors of multiple eigenvalues in any special manner (as far as I know). This can lead to a scenario where these eigenvectors are misaligned. For example, the following scenario with "random" multiple eigenvalue ordering will lead to an unnecessary 90-degree rotation in the first and second eigenpairs:

Matrix A: eigenvalues [1, 1, 1]; eigenvectors [[1,0,0],[0,1,0],[0,0,1]];
Matrix B: eigenvalues [1.5, 1.5, 1.5]; eigenvectors [[0,1,0],[1,0,0],[0,0,1]];

which will have an influence on the interpolated tensor. Specifically, it will experience eigenvector rotation even though there is actually none.

This is exactly what I tried to mitigate via `Core::LinAlg::align_eigenpairs_of_base_matrix` and `Core::LinAlg::order_eigenpairs_wrt_reference`. 
As far as I understand, @abhi-satheesh is currently reworking the implemented tensor interpolation framework, and he may have an even better approach to address this problem than my current one. Hence, addressing it may not be  part of this PR, but we should account for the underlying problem once the tensor interpolation framework has been reworked.

tuchpaul (dismissed):

dragos-ana (dismissed):
Looks good, thanks for adding this! 
I added some comments --- the main one addresses eigenvector alignment for multiple eigenvalues. I would wait for the rework from @abhi-satheesh before properly addressing it, but we should be aware that the underlying problem exists (at least it existed at the time when I first implemented the interpolation framework).

src/core/linalg/src/dense/4C_linalg_utils_tensor_interpolation.hpp
amgebauer: Ah, I see! Simple reordering does not suffice since any orthogonal unit vectors are eigenvectors in that case. For now, I don't allow any degenerate eigenvalues. Reliably avoiding this seems non-trivial :confused:

dragos-ana (approved):

tuchpaul (approved):

PR #1595, opened 2025-12-11
Add eigenvalue utilities for non-symmetric tensors (compute eigenvalues, obtain eigenvector to known eigenvalue)

amgebauer:
This PR adds two new utilities for eigenvalues of 2-tensors:

 * `eig(t)` -> `tuple[array, Tensor]`:
 Computes the (complex) eigenvalues and (complex) eigenvectors of a non-symmetric 2-tensors. Throws if it is not diagonalizable
* `compute_eigenvector(t, lambda)` -> `Tensor`
Computes the (real) eigenvector of a (real) eigenvalue.

Add unit-tests to test both methods.

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_tensor_eigen.hpp
+++ b/src/core/linalg/src/dense/4C_linalg_tensor_eigen.hpp
@@ -12,10 +12,13 @@
 
 #include "4C_linalg_symmetric_tensor.hpp"
 #include "4C_linalg_tensor.hpp"
+#include "4C_linalg_tensor_generators.hpp"
+#include "4C_linalg_tensor_svd.hpp"
 #include "4C_utils_exceptions.hpp"
 
 #include <Teuchos_LAPACK.hpp>
 
+#include <complex>
 #include <type_traits>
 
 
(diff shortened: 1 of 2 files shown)


src/core/linalg/src/dense/4C_linalg_tensor_eigen.hpp
```diff
@@ -0,0 +1,152 @@
...
+    auto [_, S, Vh] = Core::LinAlg::svd(
+        t - Core::LinAlg::get_full(
+                eigenvalue * Core::LinAlg::TensorGenerators::identity<double, size, size>));
+
+    // The smallest singular value should be zero
+    constexpr double singular_value_tolerance = 1e-16;
+    FOUR_C_ASSERT_ALWAYS(S[size - 1] < singular_value_tolerance,
+        "Could not compute eigenvector for eigenvalue {}. The provided eigenvalue is probably not "
+        "an eigenvalue of the given matrix!",
+        eigenvalue);
```
mairehenke: I’m not fully sure here, but since singular values are scale-dependent, would a relative criterion such as σ_min < eps * σ_max be more robust than an absolute tolerance?

amgebauer: You are right. I adapted the tolerance check. Similarly, I adapted it for the `eig()` case, where I removed the determinant check and just return a singular eigenvalue-tensor (which is consistent to other math libraries)

mairehenke (approved):

PR #1599, opened 2025-12-11
Refactor EAS implementation in shell7p

mairehenke:
## Description and Context

Following #1587, this PR refactors the EAS implementation in the Shell7p element.
The changes are split into three commits:

**1) Remove `calc_consistent_defgrd()` from non-EAS shells**
The consistent deformation gradient is only required when enhanced strains are present to maintain compatibility between the enhanced strain field and the deformation gradient.
Non-EAS elements now pass the displacement-based deformation gradient directly to the material model, avoiding unnecessary computations.

**2) Introduce EAS kinematics struct + direct enhancement of GL strain**
I added a new struct to collect all EAS-related kinematic quantities and switched to tensor types (e.g., SymmetricTensor) for clearer separation between displacement-based and enhanced quantities.
The enhanced Green–Lagrange strain is now computed directly from $M * \alpha$, rather than modifying the shell metrics first.
I also renamed the old `EnhancedStress` struct to `StressResultants`, since it stores integrated resultants rather than enhanced stresses.

**3) Standardize Voigt ordering**
The previous implementation used a shell-specific Voigt order. To avoid repeated reordering, I updated all affected mappings, matrices, and tensors to use the standard Voigt convention.

**4) Fix spelling of covariant and contravariant and rename associated variables**

 #1587

### DIFF
--- a/src/shell7p/4C_shell7p_ele_calc_eas.cpp
+++ b/src/shell7p/4C_shell7p_ele_calc_eas.cpp
@@ -222,9 +222,9 @@ double Discret::Elements::Shell7pEleCalcEas<distype>::calculate_internal_energy(
 
           Shell::evaluate_metrics(shape_functions, g_reference, g_current, nodal_coordinates, zeta);
 
-          // modify the current kovariant metric tensor to neglect the quadratic terms in
+          // modify the current covariant metric tensor to neglect the quadratic terms in
           // thickness directions
-          Shell::modify_kovariant_metrics(g_reference, g_current, a_reference, a_current, zeta,
+          Shell::modify_covariant_metrics(g_reference, g_current, a_reference, a_current, zeta,
               shape_functions_ans, metrics_collocation_reference, metrics_collocation_current,
               shell_data_.num_ans);
 
@@ -371,9 +371,9 @@ void Discret::Elements::Shell7pEleCalcEas<distype>::calculate_stresses_strains(
           zeta = intpoints_thickness_.qxg[gpt][0] / condfac;
           Shell::evaluate_metrics(shape_functions, g_reference, g_current, nodal_coordinates, zeta);
 
-          // modify the current kovariant metric tensor to neglect the quadratic terms in thickness
+          // modify the current covariant metric tensor to neglect the quadratic terms in thickness
           // directions
-          Shell::modify_kovariant_metrics(g_reference, g_current, a_reference, a_current, zeta,
+          Shell::modify_covariant_metrics(g_reference, g_current, a_reference, a_current, zeta,
               shape_functions_ans, metrics_collocation_reference, metrics_collocation_current,
               shell_data_.num_ans);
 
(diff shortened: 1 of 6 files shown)


amgebauer (approved):
Looks good!

src/shell7p/4C_shell7p_ele_calc_eas.cpp
```diff
@@ -295,25 +222,36 @@ double Discret::Elements::Shell7pEleCalcEas<distype>::calculate_internal_energy(
 
           Shell::evaluate_metrics(shape_functions, g_reference, g_current, nodal_coordinates, zeta);
 
-          // modify the current kovariant metric tensor to neglect the quadratic terms in thickness
-          // directions
+          // modify the current kovariant metric tensor to neglect the quadratic terms in
```
eulovi: Could you fix the small typo of kovariant to covariant? ;)

src/shell7p/4C_shell7p_ele_calc_eas_utils.hpp
```diff
@@ -60,88 +93,88 @@ namespace Discret::Elements::Shell::EAS
     }
 
     T0inv(0, 0) = t11 * t11;
-    T0inv(1, 0) = 2 * t11 * t21;
-    T0inv(2, 0) = 2 * t11 * t31;
-    T0inv(3, 0) = t21 * t21;
+    T0inv(3, 0) = 2 * t11 * t21;
+    T0inv(5, 0) = 2 * t11 * t31;
```
eulovi: Out of curiosity, why is it necessary to change the indices of `T0inv`? Could you also order the indices, so that: 
T0inv(0, 0) = ...
T0inv(1, 0) = ...
T0inv(2, 0)  = ...
...

src/shell7p/4C_shell7p_ele_calc_eas.cpp
mairehenke: This typo was actually present throughout the shell7p code. I’ve fixed it and also renamed several related variables for consistency.

src/shell7p/4C_shell7p_ele_calc_eas_utils.hpp
mairehenke: I need to change the indices of T0inv because it operates in the same Voigt space as the strain vector.
After switching from the shell-specific ordering [11,12,13,22,23,33] to the standard Voigt ordering [11,22,33,12,23,13] every matrix that maps, rotates, or otherwise transforms strain components must use the same ordering.
T0inv transforms strain components from the curvilinear basis to the Cartesian basis, and its row/column indices refer directly to the positions of the Voigt components. If we changed only the strain vector ordering (or the M matrix), but left T0inv in the old ordering, the components would be mixed incorrectly (e.g., a shear component could be interpreted as a normal strain).

mairehenke: I reordered all T0inv assignments so that rows and the column indices appear in increasing order.

eulovi: Thank you very much for the explanation and for fixing the ordering!

mairehenke:
@amgebauer could you approve again?

PR #1600, opened 2025-12-12
Enable Trilinos debug assertions in pipeline

amgebauer:
Draft for now since I don't expect that everything works.


davidrudlstorfer (commented):
Hoping for the Brownian Dyn tests to still succeed 🫣

maxfirmbach:
@amgebauer Thanks for looking into this! Totally makes sense to active the `Trilinos` debug build.

PR #1601, opened 2025-12-12
Point coupling pairs objective variation

isteinbrecher:
## Description and Context

The beam point coupling formulation had a mistake in the equations, which caused some issues with the conservation of angular momentum. This is fixed, some test cases have to be adapted slightly.


davidrudlstorfer (dismissed):
Is the tightened tolerance everywhere anticipated? Would 1e-10 be enough?

isteinbrecher:
> Is the tightened tolerance everywhere anticipated? Would 1e-10 be enough?

The results from the "old" variant and the new one were so close ~1e-10 that I decided to tighten the tolerance to avoid wrong reference results not being caught due to a too high tolerance.

PR #1602, opened 2025-12-12
Enable `restart=last_possible` when previous simulation failed.

rjoussen:
As described in #1476, the controlfile is not flushed correctly when the next step throws an error. This leads to the last entry in the control file being a `field` entry, without an accompanying result description.

So far, when using `restart=last_possible` on the command line, the restart is attempted from the last field entry. However, since there is no result description in such cases, this obviously won't work.

This fixes that by (only in such cases) searching the controlfile for the most recent complete restart information, i.e. for the most recent field entry with a step value larger than the very last one (since there are multiple field entries for the same step for coupled problems).

Along the way, I also updated the documentation on restarts slightly, as it was incomplete in my opinion (especially by restricting `last_possible` to pure structure problems.

necessary due to #1476

### DIFF
--- a/src/core/io/src/4C_io_control.cpp
+++ b/src/core/io/src/4C_io_control.cpp
@@ -13,6 +13,7 @@
 #include "4C_comm_mpi_utils.hpp"
 #include "4C_io_pstream.hpp"
 #include "4C_io_yaml.hpp"
+#include "4C_utils_exceptions.hpp"
 
 #include <pwd.h>
 #include <unistd.h>
@@ -434,12 +435,13 @@ Core::IO::InputControl::~InputControl() = default;
 
 const std::string& Core::IO::InputControl::file_name() const { return pimpl_->filename; }
 
-
-Core::IO::ControlFileEntry Core::IO::InputControl::last_entry(const std::string& key) const
+Core::IO::ControlFileEntry Core::IO::InputControl::nth_last_entry(
+    std::optional<std::string> key, size_t n) const
 {
   auto root = pimpl_->tree.rootref();
   FOUR_C_ASSERT(root.is_seq(), "Internal error: expected top-level sequence in control file");
 
+  size_t count = 0;
   for (int i = root.num_children() - 1; i >= 0; --i)
   {
     ryml::NodeRef entry = root[i];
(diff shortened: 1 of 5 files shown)


src/core/io/src/4C_io_control.cpp
```diff
@@ -520,15 +551,59 @@ Core::IO::InputControl::find_group(int step, const std::string& discretization_n
...
+    int counter = 2;
+    while (true)
+    {
+      auto second_last_possible = inputcontrol.nth_last_entry("field", counter)["step"].as<int>();
+      // throw if no more field entries exist
+      if (!second_last_possible.has_value())
+      {
+        FOUR_C_THROW(
+            "No restart entry in symbol table. "
+            "Control file corrupt?\n\nLooking for control file at: {}",
```
amgebauer: Isn't it much more likely that there is no restart file? The user might think that his file system is defect if he reads "Control file corrupt?". Maybe


```suggestion
            "Cannot find any restart entry in the control file at: {}",
```

src/core/io/src/4C_io_control.hpp
```diff
@@ -277,10 +277,23 @@ namespace Core::IO
...
+    /**
+     * A control file contains a list of entries. This function returns the last key in the control
+     * file. If no last key exists, an empty optional is returned.
+     */
+    [[nodiscard]] std::optional<std::string> last_key() const;
+
+    /**
+     * Return the last entry in the control file.
+     */
+    // [[nodiscard]] ControlFileEntry last_entry() const;
```
amgebauer:
```suggestion
```

rjoussen: Thanks

src/core/io/src/4C_io_control.cpp
rjoussen: As far as I know, almost all time integrators write a mesh at step 0, hence there always should be a field entry at step 0, which would mean that restart step is set to 0 and a complete new run is done. This is probably not intended behaviour, but it is at least visible in the standard output. So I think if there is no field entry at all, that actually means that there the control file is corrupt.

amgebauer (dismissed):

src/core/io/src/4C_io_control.cpp
amgebauer: ok, if 4C always writes the field at step 0, it is ok

tuchpaul (dismissed):

src/core/io/src/4C_io_control.cpp
rjoussen: I did a quick review, and it seems like particles, elch 1D, multiscale and reduced lung don't write field at step 0. So I will change the error message to something more meaningful.

rjoussen: Now, the first throw, meaning that there was no field entry whatsoever, gives:
```
Cannot find restart entry in the control file at: <controlfile>
Control file corrupt?
```
while the second throw means there is a field entry, but with incomplete result description. It prints:
```
Cannot find complete restart entry in the control file at: <controlfile>
```

PR #1610, opened 2025-12-16
Switch to `::NOX::Thyra::MatrixFreeJacobianOperator` from `::NOX::Epetra::MatrixFree`

vovannikov:
## Description and Context
This PR switches to `::NOX::Thyra::MatrixFreeJacobianOperator` from `NOX::Epetra::MatrixFree` inside the partitioned FSI. All `Thyra` based algorithms require an evaluation model. While working on this, I created a simple `ThyraModelWrapper` that basically delegates the corresponding call to `NOX::Nln::Interface::RequiredBase::computeF()` (which in the moment is basically `::NOX::Epetra::Interface::Required::computeF()`). Sounds simple, but one needs to write some non-obvious boilerplate code. Currently, this wrapper stays inside class `NOX::Nln::MatrixFree` but it can become a separate class if the same delegation will be required again in some other parts of the code.

As a side note, the implementation of `MatrixFreeJacobianOperator::setBaseEvaluationToRawThyra()` is broken in Trilinos, I will make a PR to Trilinos to fix it, but still at the moment we can not use since our Trilinos version is frozen.

#1591, #1524

### DIFF
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_matrixfree.cpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_matrixfree.cpp
@@ -78,9 +78,9 @@ void NOX::Nln::MatrixFree::ThyraModelWrapper::evalModelImpl(
   const Teuchos::RCP<::Thyra::VectorBase<double>> f_out = outArgs.get_f();
 
   const Teuchos::RCP<const Epetra_Vector> x_in_epetra =
-      ::Thyra::get_Epetra_Vector(map_->get_epetra_map(), x_in);
+      Core::LinearSolver::Utils::get_epetra_vector_from_thyra(*map_, x_in);
   const Teuchos::RCP<Epetra_Vector> f_out_epetra =
-      ::Thyra::get_Epetra_Vector(map_->get_epetra_map(), f_out);
+      Core::LinearSolver::Utils::get_epetra_vector_from_thyra(*map_, f_out);
 
   model_->computeF(*x_in_epetra, *f_out_epetra, ::NOX::Epetra::Interface::Required::Residual);
 }
--- a/src/core/linear_solver/src/utils/4C_linear_solver_thyra_utils.cpp
+++ b/src/core/linear_solver/src/utils/4C_linear_solver_thyra_utils.cpp
@@ -85,4 +85,23 @@ Teuchos::RCP<const Thyra::LinearOpBase<double>> Core::LinearSolver::Utils::creat
   return block_matrix;
 }
 
+
+//------------------------------------------------------------------------------
+//------------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


src/solver_nonlin_nox/4C_solver_nonlin_nox_matrixfree.cpp
```diff
@@ -7,16 +7,94 @@
...
+
+void NOX::Nln::MatrixFree::ThyraModelWrapper::evalModelImpl(
+    const ::Thyra::ModelEvaluatorBase::InArgs<double>& inArgs,
+    const ::Thyra::ModelEvaluatorBase::OutArgs<double>& outArgs) const
+{
+  const Teuchos::RCP<const ::Thyra::VectorBase<double>> x_in = inArgs.get_x();
+  const Teuchos::RCP<::Thyra::VectorBase<double>> f_out = outArgs.get_f();
+
+  const Teuchos::RCP<const Epetra_Vector> x_in_epetra =
+      ::Thyra::get_Epetra_Vector(map_->get_epetra_map(), x_in);
```
maxfirmbach: Maybe it's worth adding this to the other utils-conversion functions.

src/solver_nonlin_nox/4C_solver_nonlin_nox_matrixfree.hpp
```diff
@@ -27,10 +30,44 @@ namespace NOX
   {
     class MatrixFree : public NOX::Nln::Interface::JacobianBase
     {
+      class ThyraModelWrapper : public ::Thyra::StateFuncModelEvaluatorBase<double>
+      {
+       public:
+        ThyraModelWrapper(const std::shared_ptr<NOX::Nln::Interface::RequiredBase> model,
```
maxfirmbach, line 36: Should we consider this for other implementations too at some point? Or is this solely used for the matrix-free implementation?

vovannikov: We might, if there is a need for that. To me this is not yet absolutely clear. For this reason I decided to keep this class as private. Prior to that, I would like to break the inheritance for `NOX::Nln::Interface::RequiredBase` and `NOX::Nln::Interface::JacobianBase` from their Epetra base classes, then we'll see how it goes.

vovannikov:
> @vovannikov Thanks for tackling this. Starts to get more `Thyra` based, to which I have some small questions 😉.

@maxfirmbach This seems to be unavoidable if we want to use the modern Trilinos components, since many of them (like the current matrix-free operator) are dependent on `Thyra`. But I can not say that I was impressed with `Thyra` interfaces, so I would prefer, at least while it is possible, to limit and encapsulate the use `Thyra` as much as possible.

src/solver_nonlin_nox/4C_solver_nonlin_nox_matrixfree.cpp
vovannikov: Yep, done that.

PR #1613, opened 2025-12-18
Draft: Implement Nitsche method to couple embedded meshes

eulovi:
## Description and Context

In this PR, a non-symmetric Nitsche method is implemented for coupling embedded meshes. This is introduced in the constraint_framework library. For now, I set this as a draft, as there is redundancy in the code and I might have broke some tests from the mortar coupling. 

The general strategy for introducing this functionality is the following:
1. a Nitsche coupling pair is introduced to do the evaluation and assembly of the Nitsche contributions, similar to evaluate_and_assemble_mortar_contributions. 
2. a Nitsche manager is introduced for holding the stiffness matrices resulting from the Nitsche contributions. Many of its functions are similar to the mortar manager, like its construction and setup. The main differences are in the functions `evaluate_global_coupling_contributions` and `add_global_force_stiffness_contributions`. Maybe I could do a base class, called coupling_manager that hold the functionality of defining the pairs and doing the setup and define the subclasses of `mortar_manager` and `nitsche_manager` that have their own `evaluate_global_coupling_contributions` and `add_global_force_stiffness_contributions`. Although I am open to ideas. 

As there is a lot going on here, some clean up should be done in other PRs before merging this one:
- [ ] There are some functions in the mortar coupling pairs that could be free, as they are not related to the main functionality of the coupling pairs and are used by the nitsche coupling pairs.
- [ ] For evaluating the Cauchy stresses in NURBS solid elements, it is necessary to pass the discretization to the function `get_normal_cauchy_stress_at_xi` and to the functions that call it.


src/solid_3D_ele/4C_solid_3D_ele.hpp
```diff
@@ -190,11 +190,14 @@ namespace Discret::Elements
...
      * are null)
+     * @param discret (in) : Discretization of this element
      * @return double (out) : The scalar Cauchy stress value
      */
     double get_normal_cauchy_stress_at_xi(const std::vector<double>& disp,
         const Core::LinAlg::Tensor<double, 3>& xi, const Core::LinAlg::Tensor<double, 3>& n,
-        const Core::LinAlg::Tensor<double, 3>& dir, CauchyNDirLinearizations<3>& linearizations);
+        const Core::LinAlg::Tensor<double, 3>& dir, CauchyNDirLinearizations<3>& linearizations,
+        std::optional<std::reference_wrapper<const Core::FE::Discretization>> discret =
+            std::nullopt);
```
amgebauer: It's a lot easier if you just use a const pointer. It is not less safe in practice.

```suggestion
        const Core::FE::Discretization* discret = nullptr);
```

Optional makes sense if you want to have value semantics, but here, you just want to have an optional reference, what is basically a pointer.

The same applies in the other places too.

src/constraint_framework/4C_constraint_framework_embeddedmesh_interaction_pair_nitsche.cpp
```diff
@@ -0,0 +1,656 @@
...
+                                                                 discret,
+    const Constraints::EmbeddedMesh::SolidToSolidNitscheManager* nitsche_manager,
+    Core::LinAlg::SparseMatrix& global_penalty_interface,
+    Core::LinAlg::SparseMatrix& global_penalty_background,
+    Core::LinAlg::SparseMatrix& global_penalty_interface_background,
+    Core::LinAlg::SparseMatrix& global_disp_interface_stress_interface,
+    Core::LinAlg::SparseMatrix& global_disp_interface_stress_background,
+    Core::LinAlg::SparseMatrix& global_disp_background_stress_interface,
+    Core::LinAlg::SparseMatrix& global_disp_background_stress_background,
+    Core::LinAlg::FEVector<double>& global_constraint, double& nitsche_stabilization_param,
```
isteinbrecher: I am not so sure about this interface. Why do we need so many matrices? Nitsches method should be like the penalty method, so I would assume that we only need to assemble in the global stiffness matrix and force vector.

src/constraint_framework/4C_constraint_framework_submodelevaluator_embeddedmesh.hpp
```diff
@@ -54,15 +56,25 @@ namespace Constraints::SubmodelEvaluator
...
 
    private:
+    //! Couple strategy between overlapping and underlying meshes
+    Constraints::EmbeddedMesh::EmbeddedMeshParams embedded_mesh_coupling_params_;
+
     //! Pointer to the mortar manager. This object stores the relevant mortar matrices.
     std::shared_ptr<Constraints::EmbeddedMesh::SolidToSolidMortarManager> mortar_manager_;
+
+    //! Pointer to the Nitsche manager.
+    std::shared_ptr<Constraints::EmbeddedMesh::SolidToSolidNitscheManager> nitsche_manager_;
```
isteinbrecher, line 77: Instead of #1614 you can move the common functionality of the mortar and nitsche manager into this submode evaluator and not require the two to be based on the same base class.

isteinbrecher (commented):
Thanks for the new functionality @eulovi .

I did not have time to go through everything in detail, but I left some general comments.

PR #1622, opened 2025-12-23
Add utils to read MultiVector from matrix-market

amgebauer:
This PR adds a utility to read a multi-vector from a matrix-market format.


maxfirmbach:
Should we rename `4C_linalg_utils_sparse_algebra_print` to `4C_linalg_utils_sparse_algebra_io` and put it there? We already have several output methods to write to matrix-market files (it says print to matlab, but it is a matrix-market file in essence).

maxfirmbach:
I like the feature :+1:. Thanks for adding this.

amgebauer:
Interesting, the unit-test that now fails is not related to my changes.

amgebauer:
Thanks for the hint @maxfirmbach. I moved it to `4C_linalg_utils_sparse_algebra_io`.

Now the test should pass. I called the `MatrixMarketFileToMultiVector` only on rank 0, however, it is a collective call. This caused undefined behavior and the "unrelated" tests failed. Now, I'm calling it on all ranks using an artificial dummy map and distribute the values onto the desired map later.

maxfirmbach (dismissed):
LGTM!

amgebauer:
I had to fix merge conflicts. Now again ready for review

PR #1623, opened 2025-12-23
Adapt sparse matrix-matrix multiply

amgebauer:
@maxfirmbach I think you changed this: Why do you explicitly compute the transpose instead of using the transpose in `EpetraExt::MatrixMatrix::Multiply`?

I have some non-square matrices that end up in an epetra error during the epetra transposer, so the existing code seems broken. When using the build-in method, it works fine.


maxfirmbach:
@amgebauer Interesting, I experienced exactly the way round ;). I think some matrix maps in the code are not correct so the transpose inside the multiply would break. But I agree, this is not optimal yet.

maxfirmbach (dismissed):
Looks good and simplifies things! :+1:

maxfirmbach (approved):

PR #1624, opened 2025-12-23
[EpetraExt] Move CrsMatrix input to utils

maxfirmbach:
## Description and Context

Encapsulate the matrix read-in from matrixmarket file and put it into the linear algebra io utils.

This also simplifies some code.

Related to #1622

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_print.cpp
+++ b/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_print.cpp
@@ -9,6 +9,8 @@
 
 #include "4C_linalg_blocksparsematrix.hpp"
 
+#include <EpetraExt_CrsMatrixIn.h>
+
 #include <fstream>
 
 FOUR_C_NAMESPACE_OPEN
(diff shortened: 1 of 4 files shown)


src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_print.cpp
```diff
@@ -228,4 +230,18 @@ void Core::LinAlg::print_map_in_matlab_format(
...
+/*----------------------------------------------------------------------*/
+Core::LinAlg::SparseMatrix Core::LinAlg::read_matrix_market_file_as_sparse_matrix(
+    const std::string& filename, MPI_Comm comm)
+{
+  Epetra_CrsMatrix* crs_marix;
+
+  ASSERT_EPETRA_CALL(EpetraExt::MatrixMarketFileToCrsMatrix(
+      filename.c_str(), Core::Communication::as_epetra_comm(comm), crs_marix));
+
+  return SparseMatrix(Core::Utils::shared_ptr_from_ref(*crs_marix), DataAccess::Copy);
```
amgebauer: We need to make sure that the crs_matrix is deleted once it gets out of scope. Here, we can directly create a shared-pointer and share the data with the returned object.


```suggestion
  return SparseMatrix(std::shared_ptr<Epetra_CrsMatrix>(crs_marix), DataAccess::Share);
```

maxfirmbach: Changed it :+1:.

c-p-schmidt (approved):

PR #1627, opened 2025-12-23
Move MOR with POD to new structure framework (for solid-cardiovascular0d systems)

amgebauer:
Moving the last remaining step of the cardiovascular0d framework from the old time-integration framework to the new one. POD projection is implemented as a simple projection operator for the linear system. I implemented it as a block-pod-projector so that it can easily be extended to other block-type systems by providing the POD-basis vectors for each block separately.

Requires #1622 #1626

@mrp089 Making your research ready for the upcoming years :tada:


src/core/linear_solver/src/method/4C_linear_solver_method_direct.cpp
```diff
@@ -101,7 +104,20 @@ void Core::LinearSolver::DirectSolver::setup(std::shared_ptr<Core::LinAlg::Spars
 //----------------------------------------------------------------------------------
 int Core::LinearSolver::DirectSolver::solve(Core::LinAlg::MultiVector<double>& x)
 {
-  solver_->setX(Teuchos::rcpFromRef(x.get_epetra_multi_vector()));
+  std::unique_ptr<LinAlg::Vector<double>> projected_x = nullptr;
```
maxfirmbach, line 107: Is this really necessary? Vector `x` shouldn't have any influence on the solution, so a projection should not be necessary or does the size/length of `x` change due to the projection?

maxfirmbach (commented):
Changes look good. Only one small theoretical question.

src/core/linear_solver/src/method/4C_linear_solver_method_direct.cpp
amgebauer: @maxfirmbach thanks for the hint. Yes, the size of `x` changes during POD projection. I now create an empty vector instead of mapping the input vector.

PR #1628, opened 2025-12-23
Let ParameterList survive lin-solver builder

amgebauer:
Cherry-pick from #1600

Avoids a segmentation fault when compiling with Trilinos in debug mode. The `ParameterList` needs to outlive the `LinearSolverBuilder`.


maxfirmbach:
Does this also hold true for the `MueLu` and `Teko` adapters? I think there it is done in a similar fashion.

amgebauer:
In MueLu and Teko, we do it differently: We create an actual owning `RCP` that goes into Statimikos. In that case it is fine since the lifetime of the `ParameterList` is then managed by the `RCP`.

mayrmt (approved):

PR #1632, opened 2025-12-28
[NOX] Refactor `computeF()` and `RequiredBase`

maxfirmbach:
## Description and Context

This PR cuts the `RequiredBase` class free from `NOX::Epetra::Interface::Required` and provides our own signature of `compute_f()`. The other changes in this PR are just refactors to use the newly provided signature.

### DIFF
--- a/src/fsi/src/monolithic/4C_fsi_monolithic.cpp
+++ b/src/fsi/src/monolithic/4C_fsi_monolithic.cpp
@@ -974,15 +974,14 @@ Teuchos::RCP<::NOX::Direction::Generic> FSI::Monolithic::buildDirection(
 
 /*----------------------------------------------------------------------------*/
 /*----------------------------------------------------------------------------*/
-bool FSI::Monolithic::compute_f(
-    const Core::LinAlg::Vector<double>& x, Core::LinAlg::Vector<double>& f, FillType fill_flag)
+bool FSI::Monolithic::compute_f(const Core::LinAlg::Vector<double>& x,
+    Core::LinAlg::Vector<double>& f, NOX::Nln::FillType fill_flag)
 {
   TEUCHOS_FUNC_TIME_MONITOR("FSI::Monolithic::computeF");
-  Core::LinAlg::Vector<double> x_new = Core::LinAlg::Vector<double>(x);
-  evaluate(Core::Utils::shared_ptr_from_ref(x_new));
-  Core::LinAlg::Vector<double> F_new = Core::LinAlg::Vector<double>(f);
-  setup_rhs(F_new);
-  f = F_new;
+
+  evaluate(Core::Utils::shared_ptr_from_ref(x));
+  setup_rhs(f);
+
   return true;
 }
 
--- a/src/fsi/src/partitioned/4C_fsi_partitioned.cpp
+++ b/src/fsi/src/partitioned/4C_fsi_partitioned.cpp
@@ -49,12 +49,12 @@ FSI::Partitioned::Partitioned(MPI_Comm comm)
       idispn_(nullptr),
       iveln_(nullptr),
       raw_graph_(nullptr),
-      counter_(7),
       mfresitemax_(0),
       coupsfm_(nullptr),
       matchingnodes_(false)
 {
-  // empty constructor
+  counter_ = {{NOX::Nln::FillType::Residual, 0.0}, {NOX::Nln::FillType::MF_Res, 0.0},
+      {NOX::Nln::FillType::User, 0.0}};
 }
 
 
@@ -391,7 +391,10 @@ void FSI::Partitioned::timeloop(const std::shared_ptr<NOX::Nln::Interface::Requi
     prepare_time_step();
 
     // reset all counters
-    std::fill(counter_.begin(), counter_.end(), 0);
+    for (auto& [key, value] : counter_)
+    {
+      value = 0.0;
+    }
     lsParams.sublist("Output").set("Total Number of Linear Iterations", 0);
     linsolvcount_.resize(0);
 
@@ -455,7 +458,13 @@ void FSI::Partitioned::timeloop(const std::shared_ptr<NOX::Nln::Interface::Requi
              << lsParams.sublist("Output").get("Total Number of Linear Iterations", 0);
       for (std::vector<int>::size_type i = 0; i < counter_.size(); ++i)
       {
-        (*log) << " " << counter_[i];
+        for (const auto& [key, value] : counter_)
+        {
+          if (value != 0)
+          {
+            (*log) << " " << value;
+          }
+        }
       }
       (*log) << std::endl;
       log->flush();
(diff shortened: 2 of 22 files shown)


src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required_base.hpp
```diff
@@ -20,8 +20,36 @@ namespace NOX
...
+          //! The Jacobian matrix is being estimated.
+          Jac,
+          //! The preconditioner matrix is being estimated.
+          Prec,
+          //! The fill context is from a FD approximation (includes FDC)
+          FD_Res,
+          //! The fill context is from a MF approximation
+          MF_Res,
+          //! The fill context is from a MF computeJacobian() approximation
+          MF_Jac,
```
mayrmt: I am not sure, if everybody is familiar with the abbreviations "FD" and "MF". You could spell them out.

mayrmt (approved):

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required_base.hpp
maxfirmbach: @mayrmt I can adjust the documentation. I would keep the enum itself for now as this is a direct copy from `NOX::Epetra`. I can't say if things still work properly if we change them.

mayrmt: I just meant to change the documentation. The enum itself is fine for me (to keep it short)

sebproell (dismissed):

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required_base.hpp
```diff
@@ -20,8 +20,36 @@ namespace NOX
   {
     namespace Interface
     {
-      class RequiredBase : public ::NOX::Epetra::Interface::Required
+      class RequiredBase
       {
+       public:
+        enum FillType
```
amgebauer: It would be good if we make this an enum class. Then it is a strong type and independent of the same enum in Trilinos. Currently, we implicitly reliy on the same ordering of these enum entries here and in Trilinos. Of course, we would need to define the mapping from the 4C FillType to the Trilinos FillType.

vovannikov: I agree with this comment. But I think we do not even need the mappings to Trilinos entities, since this flag is not passed to any internal Trilinos function after this PR. I would even pull `FillType` 2 levels higher, such that it does not sit inside `RequiredBase` and a user will access it as `NOX::Nln::FillType`.

Furthermore, we do not need all the values `FillType`, we use just a few of them, like `User`, `MF_Res` and probably that's it, but it is better to check it. But we can keep also `Residual` and `Jac` for generality. I guess the use of these flags should be later also refactored, not for this PR for sure.

src/fsi/src/partitioned/4C_fsi_partitioned.cpp
```diff
@@ -678,20 +679,20 @@ bool FSI::Partitioned::computeF(const Epetra_Vector& x, Epetra_Vector& F, const
   if (Core::Communication::my_mpi_rank(get_comm()) == 0)
   {
     utils_->out() << "\n " << "FSI residual calculation" << ".\n";
-    if (fillFlag != Residual) utils_->out() << " fillFlag = " << flags[fillFlag] << "\n";
+    if (fill_flag != Residual) utils_->out() << " fillFlag = " << flags[fill_flag] << "\n";
   }
 
   // we count the number of times the residuum is build
-  counter_[fillFlag] += 1;
+  counter_[fill_flag] += 1;
```
vovannikov: The suggestion of @amgebauer will require to redeclare `counter_` as `std::map`. An alternative would be to define a helper for explicit conversion of the enum, but I it is quite ugly (smth like `return static_cast<typename std::underlying_type<E>::type>(e)`) and does not reflect the purpose of `counter_`.

src/fsi/src/partitioned/4C_fsi_partitioned.cpp
```diff
@@ -668,7 +668,8 @@ std::shared_ptr<Core::LinAlg::Vector<double>> FSI::Partitioned::interface_force(
 
 /*----------------------------------------------------------------------*/
 /*----------------------------------------------------------------------*/
-bool FSI::Partitioned::computeF(const Epetra_Vector& x, Epetra_Vector& F, const FillType fillFlag)
+bool FSI::Partitioned::compute_f(
+    const Core::LinAlg::Vector<double>& x, Core::LinAlg::Vector<double>& f, FillType fill_flag)
```
vovannikov: Similarly to `counter_`, `flags` also need to become a `std::map`.

src/fsi/src/monolithic/4C_fsi_monolithic.cpp
```diff
@@ -974,14 +974,15 @@ Teuchos::RCP<::NOX::Direction::Generic> FSI::Monolithic::buildDirection(
...
 /*----------------------------------------------------------------------------*/
-bool FSI::Monolithic::computeF(const Epetra_Vector& x, Epetra_Vector& F, const FillType fillFlag)
+bool FSI::Monolithic::compute_f(
+    const Core::LinAlg::Vector<double>& x, Core::LinAlg::Vector<double>& f, FillType fill_flag)
 {
   TEUCHOS_FUNC_TIME_MONITOR("FSI::Monolithic::computeF");
   Core::LinAlg::Vector<double> x_new = Core::LinAlg::Vector<double>(x);
   evaluate(Core::Utils::shared_ptr_from_ref(x_new));
-  Core::LinAlg::Vector<double> F_new = Core::LinAlg::Vector<double>(F);
+  Core::LinAlg::Vector<double> F_new = Core::LinAlg::Vector<double>(f);
```
vovannikov:
```suggestion
  Core::LinAlg::Vector<double> f_new = Core::LinAlg::Vector<double>(f);
```

src/fsi/src/monolithic/4C_fsi_monolithic.cpp
```diff
@@ -974,14 +974,15 @@ Teuchos::RCP<::NOX::Direction::Generic> FSI::Monolithic::buildDirection(
...
   setup_rhs(F_new);
-  F = F_new;
+  f = F_new;
```
vovannikov:
```suggestion
  f = f_new;
```

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required_base.hpp
maxfirmbach: I moved `FillType` to be an `enum class` and adapted the respective objects using to be a `std::map<>`.

amgebauer (approved):

PR #1638, opened 2025-12-30
Pass only retained nodes to MueLu

### LINKED ISSUE #1634: [BUG] MueLu does not work properly with periodic boundary conditions

The system matrix is reduced with the redundant (periodic) dofs, but the coordinates that we pass to muelu is still the full coordinates vector. With trilinos-debug, we run in an internal assertion of Trilinos.

In `tests/input_files/tsi_heatflux_monolithic_pbc.4C.yaml`, the operator has a size of 81 (27*3), but we pass 42 coordinates (instead of `27` or none). The test passes if I skip passing coordinates to muelu.

A potential fix could be to not pass the coordinates to muelu for periodic boundary condition cases. https://github.com/4C-multiphysics/4C/blob/main/src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_teko.cpp#L121

## Interested Parties

@maxfirmbach

amgebauer:
Filtering eliminated nodes before passing them to MueLu. Eliminated nodes occur in periodic boundary conditions.

On the way, I had to fix some `consts` to make it possible.

Closes #1634 

Part of #1600

### DIFF
--- a/src/core/fem/src/discretization/4C_fem_discretization_utils.hpp
+++ b/src/core/fem/src/discretization/4C_fem_discretization_utils.hpp
@@ -60,6 +60,17 @@ namespace Core::FE
   std::shared_ptr<Core::LinAlg::MultiVector<double>> extract_node_coordinates(
       const Core::FE::Discretization& discretization, const Core::LinAlg::Map& node_row_map);
 
+  /**
+   * @brief Extract all retained node coordinates of a discretization.
+   *
+   * @note Allows to specify a node row map of the non-eliminated nodes that is used to obtain the
+   * filtered node row map. The map has to be a submap of the overall full node rowmap of this
+   * discretization.
+   * @note Eliminated nodes (e.g., when using periodic boundary conditions) are skipped.
+   */
+  std::unique_ptr<Core::LinAlg::MultiVector<double>> extract_retained_node_coordinates(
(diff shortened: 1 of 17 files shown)


src/core/fem/src/discretization/4C_fem_discretization_utils.hpp
```diff
@@ -60,6 +60,17 @@ namespace Core::FE
   std::shared_ptr<Core::LinAlg::MultiVector<double>> extract_node_coordinates(
       const Core::FE::Discretization& discretization, const Core::LinAlg::Map& node_row_map);
 
+  /**
+   * Similar to the other extract_node_coordinates function, but allows to specify a node row map
+   * and filters all eliminated nodes (e.g., in periodic boundary conditions).
```
mayrmt:
```suggestion
  /**
   * @brief Compute node map with all retained nodes in case of periodic boundary conditions 
   *
   * Similar to the other extract_node_coordinates function, but allows to specify a node row map
   * and filters all eliminated nodes (e.g., in periodic boundary conditions).
```

mayrmt (approved):

src/core/fem/src/discretization/4C_fem_discretization_utils.hpp
amgebauer: Thanks for the suggestion. I reworked the documentation of the method.

maxfirmbach (approved):

PR #1639, opened 2025-12-30
Fix browndynamic tests for Trilinos assertions

amgebauer:
This fixes the failing browndyn tests in #1600.

### DIFF
--- a/src/beaminteraction/src/4C_beaminteraction_link_beam3_reissner_line2_pinjointed.cpp
+++ b/src/beaminteraction/src/4C_beaminteraction_link_beam3_reissner_line2_pinjointed.cpp
@@ -43,7 +43,7 @@ BeamInteraction::BeamLinkBeam3rLine2PinJointed::BeamLinkBeam3rLine2PinJointed()
     : BeamLinkPinJointed(),
       triad_(Core::LinAlg::Initialization::zero),
       linkele_(nullptr),
-      bspotforces_(2, Core::LinAlg::SerialDenseVector(true))
+      bspotforces_(2, Core::LinAlg::SerialDenseVector(3, true))
 {
 }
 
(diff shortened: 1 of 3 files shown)


davidrudlstorfer (approved):
Thanks for investigating this and also the fix :)

Out of curiosity - what was wrong here or rather what was the problem so I can understand the fix :D

amgebauer:
the binding spot forces were only created with a size of 1 row and never resized. [Here](https://github.com/4C-multiphysics/4C/blob/main/src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_spherebeamlinking.cpp#L689) and [here](https://github.com/4C-multiphysics/4C/blob/main/src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_crosslinking.cpp#L1313), they were accessed with the actual spatial dimensions, which is 3.

You have to decide whether 3 is correct for the binding spot forces. It might be 6 to include moments? At least accessed are only the first 3 entries.

davidrudlstorfer:
> the binding spot forces were only created with a size of 1 row and never resized. [Here](https://github.com/4C-multiphysics/4C/blob/main/src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_spherebeamlinking.cpp#L689) and [here](https://github.com/4C-multiphysics/4C/blob/main/src/beaminteraction/src/4C_beaminteraction_submodel_evaluator_crosslinking.cpp#L1313), they were accessed with the actual spatial dimensions, which is 3.
> 
> You have to decide whether 3 is correct for the binding spot forces. It might be 6 to include moments? At least accessed are only the first 3 entries.

Thanks for the explanation :) Makes total sense. 3 should be correct. Based on the file names the code occurrences only happen in trusses and jointed pins, these should only be able to transmit forces.

Thanks for the fix and explanation 🚀

PR #1641, opened 2025-12-31
Resolve redundancy of add_other functions

eulovi:
## Description and Context

In this PR, the `add_other()` methods are deleted from `SparseOperator` and its sub-classes. In the current implementation, these methods where used to call the overloaded add() functions from the `SparseOperator` for the sub-classes `SparseMatrix` and `BlockSparseMatrixBase`, which results in confusing code. I think this is done to be sure that only a  `SparseMatrix` is added to a `SparseMatrix` instance and a `BlockSparseMatrixBase` added to a `BlockSparseMatrixBase` instance. 

In this PR, a `dynamic_cast` is added to be sure that the correct matrix types are added. 

Related to #136

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_blocksparsematrix.cpp
+++ b/src/core/linalg/src/sparse/4C_linalg_blocksparsematrix.cpp
@@ -253,41 +253,26 @@ int Core::LinAlg::BlockSparseMatrixBase::ApplyInverse(
 void Core::LinAlg::BlockSparseMatrixBase::add(const Core::LinAlg::SparseOperator& A,
     const bool transposeA, const double scalarA, const double scalarB)
 {
-  A.add_other(*this, transposeA, scalarA, scalarB);
-}
+  auto* blocksparse_matrix = dynamic_cast<const Core::LinAlg::BlockSparseMatrixBase*>(&A);
+  FOUR_C_ASSERT(blocksparse_matrix != nullptr,
+      "Matrix A cannot be added to this block sparse matrix as it is not a block sparse matrix!");
+  FOUR_C_ASSERT(blocksparse_matrix->rows() == rows(),
+      "The number of rows of the block matrix does not match: {} vs {}.",
+      blocksparse_matrix->rows(), rows());
+  FOUR_C_ASSERT(blocksparse_matrix->cols() == cols(),
+      "The number of columns of the block matrix does not match: {} vs {}.",
+      blocksparse_matrix->cols(), cols());
 
-
-/*----------------------------------------------------------------------*
- *----------------------------------------------------------------------*/
-void Core::LinAlg::BlockSparseMatrixBase::add_other(Core::LinAlg::BlockSparseMatrixBase& A,
-    const bool transposeA, const double scalarA, const double scalarB) const
-{
-  A.add(*this, transposeA, scalarA, scalarB);
-}
-
-
-/*----------------------------------------------------------------------*
- *----------------------------------------------------------------------*/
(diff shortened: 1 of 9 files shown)


maxfirmbach (approved):
Looks good for me! Thanks for removing the `add_other()` construct.

I just a have follow up question. How do we want to handle the following?
```cpp
void Core::LinAlg::BlockSparseMatrixBase::add(const Core::LinAlg::BlockSparseMatrixBase& A,
const bool transposeA, const double scalarA, const double scalarB)`
```
To me it seems like this is just a copy of the other `add()` method using `Core::LinAlg::BlockSparseMatrixBase` instead of `Core::LinAlg::SparseOperator`. Could this be further simplified?

The final remark for a further PR would be, if we want to keep the `add()` method in general or try to put it into a free function, or modify the `matrix_add()` free function to be able to handle all cases.

src/core/linalg/src/sparse/4C_linalg_blocksparsematrix.cpp
```diff
@@ -253,25 +253,22 @@ int Core::LinAlg::BlockSparseMatrixBase::ApplyInverse(
...
- *----------------------------------------------------------------------*/
-void Core::LinAlg::BlockSparseMatrixBase::add_other(Core::LinAlg::BlockSparseMatrixBase& A,
-    const bool transposeA, const double scalarA, const double scalarB) const
-{
-  A.add(*this, transposeA, scalarA, scalarB);
-}
-
+  auto* blocksparse_matrix = dynamic_cast<const Core::LinAlg::BlockSparseMatrixBase*>(&A);
+  FOUR_C_ASSERT(blocksparse_matrix != nullptr,
+      "Matrix A cannot be added to this block sparse matrix as it is not a block sparse matrix!");
```
amgebauer, line 265: We should check whether both matrices have the same size:

```suggestion
FOUR_C_ASSERT(blocksparse_matrix.rows() == rows(), "The number of rows of the block matrix does not match: {} vs {}.", blocksparse_matrix.rows(), rows());
FOUR_C_ASSERT(blocksparse_matrix.cols() == cols(), "The number of colums of the block matrix does not match: {} vs {}.", blocksparse_matrix.cols(), cols());
```

eulovi: Thanks for the suggestion! Done.

eulovi:
> Looks good for me! Thanks for removing the `add_other()` construct.
> 
> I just a have follow up question. How do we want to handle the following?
> 
> ```c++
> void Core::LinAlg::BlockSparseMatrixBase::add(const Core::LinAlg::BlockSparseMatrixBase& A,
> const bool transposeA, const double scalarA, const double scalarB)`
> ```
> 
> To me it seems like this is just a copy of the other `add()` method using `Core::LinAlg::BlockSparseMatrixBase` instead of `Core::LinAlg::SparseOperator`. Could this be further simplified?

You're right @maxfirmbach. I think this function can be deleted.

eulovi:
> The final remark for a further PR would be, if we want to keep the `add()` method in general or try to put it into a free function, or modify the `matrix_add()` free function to be able to handle all cases.

@maxfirmbach I think the best option is to have the `add()` method as a free function. As `add()` is present in the `SparceOperator` class, it is has to be implemented by its subclasses. The only classes that really need it are `SparceMatrix` and `BlockSparseMatrixBase` and it is redundant for the classes `FSIMatrixFree` and `SparceOperatorWrapper`, as they only throw a "Not implemented" error message if called.

maxfirmbach:
@eulovi Then let's get this merged and do the next steps in another PR.

PR #1649, opened 2026-01-09
[Qhull] Rework configure

maxfirmbach:
## Description and Context

Lets see if this works.


maxfirmbach:
@sebproell What do you think about the `cmake` related changes?

cmake/configure/configure_Qhull.cmake
```diff
@@ -5,13 +5,23 @@
...
+message(STATUS "Fetch content for Qhull")
 
-if(QHULL_FOUND)
-  message(STATUS "QHULL include directory: ${QHULL_INCLUDE_DIR}")
-  message(STATUS "QHULL library directory: ${QHULL_LIBRARY}")
+set(QHULL_ENABLE_TESTING "OFF")
+set(BUILD_APPLICATIONS "OFF")
+set(BUILD_STATIC_LIBS "OFF")
+set(LINK_APPS_SHARED "OFF")
+set(BUILD_SHARED_LIBS "ON")
```
amgebauer, line 16: Here we set these options globally that affect all other dependencies and our code. I think that is not what we want. Can we se them so that they are only used to build qhull?

maxfirmbach: True, do you know if there is a way to give them directly to Qhull?

maxfirmbach: @amgebauer I remember a discussion with @sebproell about this from some time ago ... he had the same remark especially related to `BUILD_SHARED_LIBS`, which we  use in different places too.

amgebauer: Yes, I think that's why we introduced `FOUR_C_BUILD_SHARED_LIBS` so that the variable only has the scope of 4C. I actually don't know how to make it clean so that these options are only passed to qhull and don't affect the rest of our build.

sebproell: @amgebauer @maxfirmbach These variables will not affect the rest of the build.

We specifically ensure that nothing can leak from these configure_<package> scripts by including them in the `four_c_configure_dependency` function. A function introduces a scope for variables in CMake. Thus, these variables will not be available after the function returns.

cmake/configure/configure_Qhull.cmake
```diff
@@ -5,13 +5,23 @@
...
+  GIT_REPOSITORY https://github.com/qhull/qhull.git
+  GIT_TAG d1c2fc0caa5f644f3a0f220290d4a868c68ed4f6
+  )
 
-  four_c_remember_variable_for_install(QHULL_INCLUDE_DIR QHULL_LIBRARY)
-endif()
+fetchcontent_makeavailable(libqhull)
+set(FOUR_C_QHULL_ROOT "${CMAKE_INSTALL_PREFIX}")
+
+four_c_add_external_dependency(four_c_all_enabled_external_dependencies qhull_r)
```
sebproell, line 27:
```suggestion
four_c_add_external_dependency(four_c_all_enabled_external_dependencies QHull::qhull_r)
```

Does this work? I am looking at https://github.com/qhull/qhull/blob/master/CMakeLists.txt#L744, which gives an export namespace but unfortunately they do not alias the `qhull_r` target as `QHull::qhull_r`. Maybe we need to do this ourselves here. 

Also, QHull is still installed in the image, right? I think the wrong one is picked up during the install test: https://github.com/4C-multiphysics/4C/actions/runs/20850200696/job/59903586127?pr=1649#step:9:46
which would give you a pretty random result. I would drop the QHull from the image. Afterwards, you probably need to fix the install process here.

cmake/configure/configure_Qhull.cmake
```diff
@@ -5,13 +5,23 @@
...
+set(BUILD_APPLICATIONS "OFF")
+set(BUILD_STATIC_LIBS "OFF")
+set(LINK_APPS_SHARED "OFF")
+set(BUILD_SHARED_LIBS "ON")
 
-  target_link_libraries(four_c_all_enabled_external_dependencies INTERFACE qhull::qhull)
+fetchcontent_declare(
+  libqhull
+  GIT_REPOSITORY https://github.com/qhull/qhull.git
+  GIT_TAG d1c2fc0caa5f644f3a0f220290d4a868c68ed4f6
```
sebproell: This seems to be the current development state. Why not take a release? In any case, please add a comment indicating the version.

maxfirmbach: Yep, this was mainly for initial testing, will revert to the current release and add versioning information.

eulovi: @maxfirmbach I can fix the embeddedmesh tests after you change the qhull to its release version, as I think the triangulation of the integration cells changes depending on the qhull version.

maxfirmbach: I now use the latest official Qhull release + added version information as comment.

PR #1650, opened 2026-01-09
Fix yaml dbc writer

### LINKED ISSUE #1647: [BUG] Monitor Structure Dbc fails for yaml without condition information

yaml output seems to work only when `WRITE_CONDITION_INFORMATION` is true. The following gives a vector out of bounds error in the yaml writers:

```
IO/MONITOR STRUCTURE DBC:
  INTERVAL_STEPS: 1
  WRITE_CONDITION_INFORMATION: false
  FILE_TYPE: yaml
```

## Possible Solution

I think the problem is that in `MonitorDbc::setup`, the `dbc_monitor_yaml_file_trees_` vector is only populated when `WRITE_CONDITION_INFORMATION` information is true. That logic would need to be changed slightly.

## Your Environment

- **Relevant repo SHA1s:**
- **Relevant configure flags or configure script:**
- **Operating system and version:**
- **Compiler and TPL versions:**

- Blocks
- Follows
- Related to
- Part of
- Composed of

## Interested Parties

@c-p-schmidt

rjoussen:
## Description and Context
This fixes the bug in #1647.

The check if `WRITE_CONDITION_INFORMATION`is true is moved inside the case yaml->no restart because it is only necessary then. Especially, a YAML tree is initiated and put in `dbc_monitor_yaml_file_trees_`, and a data node is appended to it now, regardless of `WRITE_CONDITION_INFORMATION`.

I modified a beam test to test this, whose monitor_dbc output has not been checked at all so far.

Closes #1647 
Related to #1290 and #1211


tests/input_files/ref/beam3r_monitor_dbc_yaml_no_condition_information-1-101_monitor_dbc.yaml
isteinbrecher, line 1: This tests the new feature, however, I suggest to create a new input file that only tests the expected behavior (preferably with an expressive name) instead of adding it to an otherwise completely unrelated test. Otherwise the desired "test" might be removed by someone who ist not aware that this feature is tested there.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
```diff
@@ -169,98 +168,91 @@ void Solid::MonitorDbc::setup()
...
-            Core::Communication::my_mpi_rank(get_comm()) == 0)
+        // only write yaml on rank 0
+        if (Core::Communication::my_mpi_rank(get_comm()) == 0)
         {
-          const std::string full_restart_dirpath(
-              Global::Problem::instance()->output_control_file()->restart_name());
+          // handle restart, read from existing file and remove steps beyond restart step
+          if (Global::Problem::instance()->restart())
+          {
+            const std::string full_restart_dirpath(
```
sebproell: A rather general comment: could you use early returns and separate functions to decrease the indentation level? It is quite hard to follow what is going on in this deeply nested code.

tests/input_files/ref/beam3r_monitor_dbc_yaml_no_condition_information-1-101_monitor_dbc.yaml
rjoussen: I copied the file, now renamed it, and do the test only on the new test. I also tested with and without condition information, as the other case wasn't explicitly tested as well.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
rjoussen: I gathered the handling functionality now in a separate function, called `read_restart_yaml_file`.

tests/input_files/ref/beam3r_monitor_dbc_yaml_no_condition_information-1-101_monitor_dbc.yaml
isteinbrecher: Thanks for the additional effort.

If I were to nitpick, you have now copied a beam-to-solid test case which is quite a lot of additional "noise" in the test, where you want to test just your feature. For what you are trying to test, a simple block with a very coarse discretisation, where a nodeset containing a few nodes is moved (**point** Neumann condition). This will reduce runtime, and more importantly, should lead to a very simple input file.

Out of experience, once you have setup such an example, one can then reuse it for other test cases in the future.

I will leave this up to you, otherwise, everything looks good.

isteinbrecher (dismissed):
Thanks again for the contribution @rjoussen , feel free to address my previous comment.

tests/input_files/ref/beam3r_monitor_dbc_yaml_no_condition_information-1-101_monitor_dbc.yaml
rjoussen: @isteinbrecher you are right, that would be better. I was just being lazy. I will change it :)
Addressed in the latest force-push

davidrudlstorfer (approved):

isteinbrecher (approved):
@rjoussen thanks for going the extra mile.

PR #1652, opened 2026-01-09
Add direct gmsh input

rjoussen:
## Description and Context

This is a first step to use gmsh to read in .msh files, as outlined in #825.

Summary:
- All gmsh physical groups are converted to point sets and element blocks. 
- Physical groups are not added to element blocks if they consist of multiple element types.
- Physical group tags must be unique across dimensions. The tag can then directly be used to identify blocks/node sets in BCs or Element block definitions.

I added a simple test case for with one tet and one hex block and a couple of node sets of different dimensions. Additionally, I wrote unittests for all element types supported by both gmsh and 4C, to ensure positive Jacobians.

Note: I noticed that the line6 element is currently not tested at all. 

Part of #825, #60


src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,451 @@
...
+      {
+        std::cout << "Warning: Multiple element types detected "
+                  << "in physical group " << (group_name.empty() ? "(unnamed)" : group_name)
+                  << " (dim=" << dim << ", tag=" << tag << "). "
+                  << "Detected at least " << Core::FE::cell_type_to_string(cell_type) << " and "
+                  << Core::FE::cell_type_to_string(pair_in_map->second.cell_type) << ".\n"
+                  << "This physical group is removed from element blocks, but can still be used as "
+                     "a point set.\n";
+        mesh.cell_blocks.erase(pair_in_map);
+      }
```
rjoussen, line 336: The reason this is only asserted here is that it could be helpful in the future to sort different element types of the same physical group into different internal element blocks, and the later refer to the element blocks by their name rather their internal id

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+#ifdef FOUR_C_WITH_GMSH
+#include <gmsh.h>
+#endif
+
+FOUR_C_NAMESPACE_OPEN
+
+#ifdef FOUR_C_WITH_GMSH
+
+namespace GmshUtils
```
sebproell:
```suggestion
namespace GmshImplementation
```

This better indicates that you only need this code internally.

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,438 @@
...
+   * @brief Get the elements for a gmsh physical group
+   *
+   * @param dim The dimension of the physical group.
+   * @param tag The tag of the physical group.
+   * @param elementType_to_element_data_map An empty map from gmsh element type (int) to ElementData
+   * struct to be filled. The resulting map collects for each element type in that group the
+   * corresponding element data.
+   */
+  void get_elements_for_physical_group(
+      const int dim, const int tag, std::map<int, ElementData>& elementType_to_element_data_map)
```
sebproell:
```suggestion
      const int dim, const int tag, std::map<int, ElementData>& element_type_to_element_data_map)
```

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,451 @@
...
+      case 12:
+        return Core::FE::CellType::hex27;
+      case 13:
+        FOUR_C_THROW(
+            "Detected a 18-node prism (wedge) element in the mesh.\n"
+            "4C does not support this element type.\n"
+            "You may try to use the gmsh option 'Mesh.SecondOrderIncomplete = 1' to use wedge15 "
+            "instead.\n");
+      case 14:
+        // pyramid14, not implemented
```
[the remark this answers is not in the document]
sebproell: This Copilot comment is still relevant.

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,438 @@
...
+      case 27:
+        return Core::FE::CellType::line5;
+      case 28:
+        return Core::FE::CellType::line6;
+      default:
+        // all other element types are not implemented
+        FOUR_C_THROW("Unsupported Gmsh element type: {}", gmsh_element_type);
+    }
+  }
+  // clang-format off
```
sebproell: I guess you want a specific array format, so just add trailing `//` after the relevant lines


```suggestion
```

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,438 @@
...
+    assert(node_count == gmsh_connectivity.size());
+    assert(gmsh_node_order_in_four_c != nullptr);
+
+    std::vector<int> four_c_connectivity(node_count);
+    // reorder connectivity
+    for (std::size_t i = 0; i < node_count; ++i)
+    {
+      four_c_connectivity[i] = gmsh_connectivity[gmsh_node_order_in_four_c[i]];
+    }
+    return four_c_connectivity;
```
sebproell: Extract this as a lambda at the top of this function and just call the function in the four specific cases, directly returning the result. Gives a much simpler control flow.

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,438 @@
...
+        node_count = wedge15_gmsh_node_order_in_four_c.size();
+        break;
+      default:
+        // in all other cases, no reordering is needed. Directly return the input connectivity.
+        return {gmsh_connectivity.begin(), gmsh_connectivity.end()};
+    }
+
+    // sanity check
+    assert(node_count == gmsh_connectivity.size());
+    assert(gmsh_node_order_in_four_c != nullptr);
```
sebproell: Use `FOUR_C_ASSERT(...)`. A few more times in this file.

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,438 @@
...
+  Core::IO::MeshInput::RawMesh<3> mesh{};
+
+  gmsh::initialize();
+
+  try
+  {
+    gmsh::open(msh_file.string());
+  }
+  catch (...)
+  {  // catch all exceptions from Gmsh
```
sebproell: Do we know the type of exception?

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,438 @@
...
+#include "4C_io_gmsh_reader.hpp"
+
+#include "4C_fem_general_cell_type.hpp"
+#include "4C_fem_general_cell_type_traits.hpp"
+#include "4C_io_mesh.hpp"
+#include "4C_utils_exceptions.hpp"
+
+#include <array>
+#include <cassert>
+#include <cstddef>
```
sebproell:
```suggestion
```

src/core/io/src/4C_io_gmsh_reader.cpp
rjoussen: After looking into it briefly in the gmsh source code, I cannot really find a specific error that gmsh would throw in these cases.

src/core/io/src/4C_io_gmsh_reader.cpp
rjoussen: It did indeed, thank you!

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,440 @@
...
+}  // namespace GmshImplementation
+
+Core::IO::MeshInput::RawMesh<3> Core::IO::Gmsh::read_msh_file(const std::filesystem::path& msh_file)
+{
+  FOUR_C_ASSERT_ALWAYS(
+      std::filesystem::exists(msh_file), "File {} does not exist.", msh_file.string());
+
+  Core::IO::MeshInput::RawMesh<3> mesh{};
+
+  gmsh::initialize();
```
amgebauer: I think it is worth to introduce a RAII guard here. Something like

```c++
class GmshSession
{
public:
    explicit GmshSession(const std::filesystem::path& msh_file)
    {
        gmsh::initialize();
        try {
            gmsh::open(msh_file.string());
        }
        catch (...) {
            gmsh::finalize();
            FOUR_C_THROW("Failed to open Gmsh file {}\nIs the file corrupted?\nGmsh version: {}", msh_file.string(), GMSH_API_VERSION);
        }
    }

    // Non-copyable (Gmsh is a global singleton)
    GmshSession(const GmshSession&) = delete;
    GmshSession& operator=(const GmshSession&) = delete;

    // We don't need move operations
    GmshSession(GmshSession&&) = delete;
    GmshSession& operator=(GmshSession&&) = delete;

    ~GmshSession() noexcept
    {
        gmsh::finalize();
    }
};
```

and use it as

```c++
GmshSession gmsh(msh_file);
```

This guarantees that `gmsh::finalize()` is always called even in case of an exception.

rjoussen: Thanks for the hint, I implemented that. I'm now also checking if another gmsh session is already active, such that there are no concurrent gmsh sessions.

rjoussen:
In the last push, I refactored the `Gmsh::Implentation` namespace into an anonymous namespace, such that it definitely cant be found and used from anywhere outside. 

The `GmshSession` RAII guard class is deliberately not in the anonymous namespace, since it actually should be used if we ever want to use gmsh anywhere else.

> In a follow-up, you might also add a small tutorial.

Yes, I will do that as soon as I think the gmsh input will stay unchanged for a while. As mentioned [in this comment](https://github.com/4C-multiphysics/4C/issues/825#issuecomment-3729613058), I would like to use gmsh physical names more, since dealing with IDs all the time is a little annoying. At least for the condition definition, this is little extra effort I think.

maxfirmbach:
@PhilipOesterlePekrun @bennoschoenstein Might be interesting for you too.

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,474 @@
...
+
+/**
+ * @brief RAII guard to manage a gmsh session.
+ *
+ * Initializes and opens the given msh file on construction and finalizes gmsh on destruction.
+ * Explicitly non-copyable and non-movable to avoid multiple finalization calls.
+ * @throws if a GmshSession is already active (nested sessions are not supported).
+ *
+ */
+class GmshSession
```
sebproell: How do you plan on using this class elsewhere if it is only appearing in the cpp file? It should live in a proper namespace and in the header, I think.

src/core/io/src/4C_io_gmsh_reader.cpp
```diff
@@ -0,0 +1,474 @@
...
+{
+ public:
+  explicit GmshSession(const std::filesystem::path& msh_file)
```
sebproell:
```suggestion
  [[nodiscard]] explicit GmshSession(const std::filesystem::path& msh_file)
```

Important to disallow something like `GmshSession();`

sebproell (commented):
Only had a look at the new `GmshSession` class.

src/core/io/src/4C_io_gmsh_reader.cpp
rjoussen: I  am not planning on using it elsewhere, but as far as I understood, the whole point of this guard class is to make gmsh usage safe if it ever is or was used anywhere else. Hence I put it in the header now, next to `read_msh_file()`.

src/core/io/src/4C_io_gmsh_reader.cpp
rjoussen: That makes sense, thanks!

amgebauer (approved):

PR #1653, opened 2026-01-09
Cleanup inpar dependency

maxfirmbach:
## Description and Context

As the title states.

### DIFF
--- a/src/ale/CMakeLists.txt
+++ b/src/ale/CMakeLists.txt
@@ -15,7 +15,6 @@ set(_dependencies
     core
     fluid_ele
     global_data
-    inpar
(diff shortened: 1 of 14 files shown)


mayrmt (approved):

maxfirmbach:
I thought this is rather easy, but it seems like we include a lot of `inpar` stuff that is still left ...

maxfirmbach:
I think I need to close this for now. Somehow it still compiles despite deleting the `inpar` dependency with some `#include <inpar...>` left. I thought the code would break ... but it doesn't ... maybe due to cyclic things happening ...

PR #1660, opened 2026-01-15
Add additional NOX status test for beaminteraction using Lagrange multipliers 

dharinib98:
## Description and Context

Following #1540, this PR adds two more additional convergence checks, by checking the update norm and the rhs norm specifically for the Lagrange multipliers  in beaminteraction. This results in improved convergence behavior of the NL solver when dealing with beaminteraction problems using Lagrange multipliers. 

#1162

### DIFF
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_enum_lists.hpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_enum_lists.hpp
@@ -29,7 +29,8 @@ namespace NOX
       sol_meshtying,           ///< meshtying problem
       sol_cardiovascular0d,    ///< 0D cardiovascular problem
       sol_lag_pen_constraint,  ///< Lagrange or/and penalty enforced constraint problem
-      sol_scatra               ///< scalar transport problem
+      sol_scatra,              ///< scalar transport problem
+      sol_beaminteraction_lm   ///< beaminteraction with Lagrange multipliers problem
     };
 
     //! Map quantity enum to std::string
@@ -268,6 +269,7 @@ namespace NOX
         quantity_eas,                 ///< check eas dofs
         quantity_levelset_reinit,     ///< check levelset reinitialization
         quantity_constraints,         ///< check constraint framework quantities
+        quantity_beaminteraction_lm,  ///< check lagrange multiplier quantities for beaminteraction
       };
 
       /// Map quantity name to std::string
@@ -295,6 +297,8 @@ namespace NOX
             return "LevelSet-Reinit";
           case quantity_constraints:
             return "Constraints";
+          case quantity_beaminteraction_lm:
+            return "Beaminteraction-LM";
           case quantity_unknown:
           default:
             return "unknown quantity type";
@@ -326,6 +330,8 @@ namespace NOX
           type = quantity_levelset_reinit;
         else if (name == "Constraints")
           type = quantity_constraints;
+        else if (name == "Beaminteraction-LM")
+          type = quantity_beaminteraction_lm;
 
         return type;
       };
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
@@ -240,7 +240,14 @@ Teuchos::RCP<const std::vector<double>> NOX::Nln::Group::get_rhs_norms(
     if (rval >= 0.0)
     {
       norms->push_back(rval);
+      continue;
     }
+
+    rval = get_nln_req_interface_ptr()->get_primary_rhs_norms(RHSVector.get_linalg_vector(), chQ[i],
+        type[i], (*scale)[i] == ::NOX::StatusTest::NormF::Scaled);
+
+    if (rval >= 0.0)
+      norms->push_back(rval);
     else
     {
       std::ostringstream msg;
@@ -343,7 +350,15 @@ Teuchos::RCP<std::vector<double>> NOX::Nln::Group::get_solution_update_norms(
     if (rval >= 0.0)
     {
       norms->push_back(rval);
+      continue;
     }
+
+    rval = get_nln_req_interface_ptr()->get_primary_solution_update_norms(
+        xVector.get_linalg_vector(), xOldNox.get_linalg_vector(), chQ[i], type[i],
+        (*scale)[i] == StatusTest::NormUpdate::Scaled);
+
+    if (rval >= 0.0)
+      norms->push_back(rval);
     else
     {
       std::ostringstream msg;
@@ -381,7 +396,14 @@ Teuchos::RCP<std::vector<double>> NOX::Nln::Group::get_previous_solution_norms(
     if (rval >= 0.0)
     {
       norms->push_back(rval);
+      continue;
     }
+    rval = get_nln_req_interface_ptr()->get_previous_primary_solution_norms(
+        xOldNox.get_linalg_vector(), chQ[i], type[i],
+        (*scale)[i] == StatusTest::NormUpdate::Scaled);
+
+    if (rval >= 0.0)
+      norms->push_back(rval);
     else
     {
       std::ostringstream msg;
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required.hpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required.hpp
@@ -41,8 +41,7 @@ namespace NOX
        public:
         //! Constructor
         Required() {};
-
-        //! returns the right-hand-side norms of the primary DoF fields
+        //! Return the desired norm of the residual of the primary DoF fields
         virtual double get_primary_rhs_norms(const Core::LinAlg::Vector<double>& F,
             const NOX::Nln::StatusTest::QuantityType& checkQuantity,
             const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
@@ -54,14 +53,14 @@ namespace NOX
             const NOX::Nln::StatusTest::QuantityType& checkQuantity,
             const bool& disable_implicit_weighting = false) const = 0;
 
-        //! Returns the increment norm of the primary DoF fields
+        //! Return the desired norm of the increment of the primary DoF fields
         virtual double get_primary_solution_update_norms(const Core::LinAlg::Vector<double>& xNew,
             const Core::LinAlg::Vector<double>& xOld,
             const NOX::Nln::StatusTest::QuantityType& checkQuantity,
             const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
             const bool& isScaled = false) const = 0;
 
-        //! Returns the previous solution norm of primary DoF fields
+        //! Return the desired norm of the solution of primary DoF fields from previous iteration
         virtual double get_previous_primary_solution_norms(const Core::LinAlg::Vector<double>& xOld,
             const NOX::Nln::StatusTest::QuantityType& checkQuantity,
             const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
--- a/src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
+++ b/src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
@@ -162,6 +162,7 @@ double Solid::TimeInt::NoxInterface::get_primary_rhs_norms(const Core::LinAlg::V
       break;
     }
     case NOX::Nln::StatusTest::quantity_pressure:
+    case NOX::Nln::StatusTest::quantity_beaminteraction_lm:
     {
       // export the model specific solution if necessary
       auto rhs_ptr = gstate_ptr_->extract_model_entries(mt, F);
@@ -244,6 +245,7 @@ double Solid::TimeInt::NoxInterface::get_primary_solution_update_norms(
     case NOX::Nln::StatusTest::quantity_structure:
     case NOX::Nln::StatusTest::quantity_cardiovascular0d:
     case NOX::Nln::StatusTest::quantity_pressure:
+    case NOX::Nln::StatusTest::quantity_beaminteraction_lm:
     {
       // export the displacement solution if necessary
       auto model_incr_ptr = gstate_ptr_->extract_model_entries(mt, xold);
@@ -295,6 +297,7 @@ double Solid::TimeInt::NoxInterface::get_previous_primary_solution_norms(
     case NOX::Nln::StatusTest::quantity_structure:
     case NOX::Nln::StatusTest::quantity_cardiovascular0d:
     case NOX::Nln::StatusTest::quantity_pressure:
+    case NOX::Nln::StatusTest::quantity_beaminteraction_lm:
     {
       // export the displacement solution if necessary
       auto model_xold_ptr = gstate_ptr_->extract_model_entries(mt, xold);
--- a/src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.hpp
+++ b/src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.hpp
@@ -75,7 +75,7 @@ namespace Solid
       bool compute_f_and_jacobian(const Core::LinAlg::Vector<double>& x,
           Core::LinAlg::Vector<double>& rhs, Core::LinAlg::SparseOperator& jac) override;
 
-      /*! Get the norm of right hand side rows/entries related to
+      /*! Get the norm of residual rows/entries related to
        *  primary DoFs (derived from NOX::Nln::Interface::Required) */
       double get_primary_rhs_norms(const Core::LinAlg::Vector<double>& F,
           const NOX::Nln::StatusTest::QuantityType& checkquantity,
@@ -97,7 +97,7 @@ namespace Solid
           const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
           const bool& isscaled = false) const override;
 
-      /*! Returns the previous solution norm of primary DoF fields
+      /*! Return the desired norm of solution of the primary DoF fields from previous iteration
        *  (derived from NOX::Nln::Interface::Required) */
       double get_previous_primary_solution_norms(const Core::LinAlg::Vector<double>& xold,
           const NOX::Nln::StatusTest::QuantityType& checkquantity,
--- a/tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange.xml
+++ b/tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange.xml
@@ -1,0 +1,69 @@
+<ParameterList name="Status Test">
+  <!-- Outer Status Test: This test is an OR combination of the structural convergence and the maximum number of iterations -->
+  <ParameterList name="Outer Status Test">
+    <Parameter name="Test Type" type="string" value="Combo"/>
+    <Parameter name="Combo Type" type="string" value="OR"/>
+    <!-- Structural convergence is an AND combination of the residuum and step update -->
+    <ParameterList name="Test 0">
+      <Parameter name="Test Type" type="string" value="Combo"/>
+      <Parameter name="Combo Type" type="string" value="AND"/>
+      <!-- BEGIN: Combo AND - Test 0: "NormF" -->
+      <ParameterList name="Test 0">
+        <Parameter name="Test Type" type="string" value="Combo"/>
+        <Parameter name="Combo Type" type="string" value="AND"/>
+        <ParameterList name="Test 0">
+          <Parameter name="Test Type" type="string" value="NormF"/>
+          <!-- NormF - Quantity 0: Check the residual norm of the structural quantities -->
+          <ParameterList name="Quantity 0">
+            <Parameter name="Quantity Type" type="string" value="Structure"/>
+            <Parameter name="Tolerance Type" type="string" value="Absolute"/>
+            <Parameter name="Tolerance" type="double" value="1e-12"/>
+            <Parameter name="Norm Type" type="string" value="Two Norm"/>
+            <Parameter name="Scale Type" type="string" value="Scaled"/>
+          </ParameterList>
+          <!-- NormF - Quantity 1: Check the residual norm of the Lagrange Multiplier quantities -->
+          <ParameterList name="Quantity 1">
+            <Parameter name="Quantity Type" type="string" value="Beaminteraction-LM"/>
+            <Parameter name="Tolerance Type" type="string" value="Absolute"/>
+            <Parameter name="Tolerance" type="double" value="1.0e-12"/>
+            <Parameter name="Norm Type" type="string" value="Two Norm"/>
+            <Parameter name="Scale Type" type="string" value="Scaled"/>
+          </ParameterList>
+        </ParameterList>
+      </ParameterList>
+      <!-- END: Combo AND - Test 0: "NormF" -->
+      <!-- BEGIN: Combo AND - Test 1: "UpdateNorm" -->
+      <ParameterList name="Test 1">
+        <Parameter name="Test Type" type="string" value="Combo"/>
+        <Parameter name="Combo Type" type="string" value="AND"/>
+        <ParameterList name="Test 0">
+          <Parameter name="Test Type" type="string" value="NormUpdate"/>
+          <!-- UpdateNorm - Quantity 0: Check the increment of the structural displacements -->
+          <ParameterList name="Quantity 0">
+            <Parameter name="Quantity Type" type="string" value="Structure"/>
+            <Parameter name="Tolerance Type" type="string" value="Absolute"/>
+            <Parameter name="Tolerance" type="double" value="1e-12"/>
+            <Parameter name="Norm Type" type="string" value="Two Norm"/>
+            <Parameter name="Scale Type" type="string" value="Scaled"/>
+          </ParameterList>
+          <!-- UpdateNorm - Quantity 1: Check the increment of the Lagrange Multiplier quantities -->
+          <ParameterList name="Quantity 1">
+            <Parameter name="Quantity Type" type="string" value="Beaminteraction-LM"/>
+            <Parameter name="Tolerance Type" type="string" value="Absolute"/>
+            <Parameter name="Tolerance" type="double" value="1.0e-12"/>
+            <Parameter name="Norm Type" type="string" value="Two Norm"/>
+            <Parameter name="Scale Type" type="string" value="Scaled"/>
+          </ParameterList>
+        </ParameterList>
+      </ParameterList>
+      <!-- END: Combo AND - Test 1: "UpdateNorm" -->
+    </ParameterList>
+    <!-- END: Combo 0 - Test 0: "Combo" -->
+    <!-- BEGIN: Combo OR - Test 1: "MaxIters" -->
+    <ParameterList name="Test 1">
+      <Parameter name="Test Type" type="string" value="MaxIters"/>
+      <Parameter name="Maximum Iterations" type="int" value="10"/>
+    </ParameterList>
+    <!--END: "MaxIters" -->
+  </ParameterList>
+</ParameterList>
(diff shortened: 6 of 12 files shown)


src/solver_nonlin_nox/4C_solver_nonlin_nox_enum_lists.hpp
```diff
@@ -24,12 +24,13 @@ namespace NOX
...
-      sol_cardiovascular0d,    ///< 0D cardiovascular problem
-      sol_lag_pen_constraint,  ///< Lagrange or/and penalty enforced constraint problem
-      sol_scatra               ///< scalar transport problem
+      sol_structure,            ///< structural problem
+      sol_contact,              ///< contact problem
+      sol_meshtying,            ///< meshtying problem
+      sol_cardiovascular0d,     ///< 0D cardiovascular problem
+      sol_lag_pen_constraint,   ///< Lagrange or/and penalty enforced constraint problem
+      sol_scatra,               ///< scalar transport problem
+      sol_lagrange_multipliers  ///< beaminteraction with Lagrange multipliers problem
```
mayrmt: I am not convinced by this name. We use Lagrange multipliers for all sorts of things, so it might be good to include the use case (beam interaction) in the name of the enum value.

src/solver_nonlin_nox/4C_solver_nonlin_nox_enum_lists.hpp
```diff
@@ -260,14 +261,15 @@ namespace NOX
...
-        quantity_constraints,         ///< check constraint framework quantities
+        quantity_contact_friction,      ///< check (semi-smooth) contact quantities (frictionless)
+        quantity_meshtying,             ///< check meshtying quantities
+        quantity_cardiovascular0d,      ///< check 0d cardiovascular quantities
+        quantity_lag_pen_constraint,    ///< check Lagrange/penalty enforced constraint quantities
+        quantity_pressure,              ///< check pressure dofs
+        quantity_eas,                   ///< check eas dofs
+        quantity_levelset_reinit,       ///< check levelset reinitialization
+        quantity_constraints,           ///< check constraint framework quantities
+        quantity_lagrange_multipliers,  ///< check lagrange multiplier quantities
```
mayrmt: I am not convinced by this name. We use Lagrange multipliers for all sorts of things, so it might be good to include the use case (beam interaction) in the name of the enum value.

src/solver_nonlin_nox/4C_solver_nonlin_nox_enum_lists.hpp
```diff
@@ -295,6 +297,8 @@ namespace NOX
             return "LevelSet-Reinit";
           case quantity_constraints:
             return "Constraints";
+          case quantity_lagrange_multipliers:
+            return "Lagrange-Multipliers";
```
mayrmt: I am not convinced by this name. We use Lagrange multipliers for all sorts of things, so it might be good to include the use case (beam interaction) in the name of the returned string.

src/solver_nonlin_nox/4C_solver_nonlin_nox_enum_lists.hpp
```diff
@@ -308,7 +312,7 @@ namespace NOX
 
         if (name == "Structure")
           type = quantity_structure;
-        else if (name == "Contact-Normal")
+        else if (name == "Contact-Normal")  //
```
mayrmt:
```suggestion
        else if (name == "Contact-Normal")
```

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
```diff
@@ -343,7 +352,15 @@ Teuchos::RCP<std::vector<double>> NOX::Nln::Group::get_solution_update_norms(
...
+      continue;
     }
+
+    rval = get_nln_req_interface_ptr()->get_lagrange_multiplier_solution_update_norms(
+        xVector.get_linalg_vector(), xOldNox.get_linalg_vector(), chQ[i], type[i],
+        (*scale)[i] == StatusTest::NormUpdate::Scaled);
+
+    if (rval >= 0.0)
+      norms->push_back(rval);
     else
```
mayrmt, line 362: 1. This changes the behavior of the code: for the solution update norms, there is no `else` with an error message any more.
2. The `continue` looks weirds and points to a suboptimal design. I think one should organize these cases differently.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
```diff
@@ -381,7 +398,14 @@ Teuchos::RCP<std::vector<double>> NOX::Nln::Group::get_previous_solution_norms(
...
       norms->push_back(rval);
+      continue;
     }
+    rval = get_nln_req_interface_ptr()->get_previous_lagrange_multiplier_solution_norms(
+        xOldNox.get_linalg_vector(), chQ[i], type[i],
+        (*scale)[i] == StatusTest::NormUpdate::Scaled);
+
+    if (rval >= 0.0)
+      norms->push_back(rval);
     else
```
mayrmt, line 407: This changes the behavior of the code: for the previous solution update norms, there is no else with an error message any more.
The continue looks weirds and points to a suboptimal design. I think one should organize these cases differently.

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required.hpp
```diff
@@ -48,6 +48,12 @@ namespace NOX
             const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
             const bool& isScaled = false) const = 0;
 
+        //! returns the right-hand-side norms of the Lagrange Multiplier DoF fields
+        virtual double get_lagrange_multiplier_rhs_norms(const Epetra_Vector& F,
```
mayrmt: right-hand side != residual: right-hand side is a term from linear systems of equations, while residual is used for nonlinear systems. Although both are related in case of Newton schemes, I personally prefer to use the accurate terms to avoid confusion.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
dharinib98: I thought about this aspect. I think we still have the error but just that it only throws the error if neither of the norms (primary and LM norms) is available. Also the `continue` logic is what is being used in contact that I tried to replicate: https://github.com/4C-multiphysics/4C/blob/c6766090fa8ca1751ef53e1716a8249e8549d35c/src/solver_nonlin_nox/4C_solver_nonlin_nox_constraint_group.cpp#L179

Maybe @isteinbrecher has some suggestions on how we deal with this

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -200,6 +200,42 @@ double Solid::TimeInt::NoxInterface::get_primary_rhs_norms(const Core::LinAlg::V
...
+double Solid::TimeInt::NoxInterface::get_lagrange_multiplier_rhs_norms(const Epetra_Vector& F,
+    const NOX::Nln::StatusTest::QuantityType& checkquantity,
+    const ::NOX::Abstract::Vector::NormType& type, const bool& isscaled) const
+{
+  check_init_setup();
+  double rhsnorm = -1.0;
+
+  // convert the given quantity type to a model type
+  const Inpar::Solid::ModelType mt = Solid::Nln::convert_quantity_type2_model_type(checkquantity);
+  auto mt_temp = Inpar::Solid::model_beaminteraction;
```
isteinbrecher: Is there a use for this `mt_temp` variable?

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -200,6 +200,42 @@ double Solid::TimeInt::NoxInterface::get_primary_rhs_norms(const Core::LinAlg::V
...
+  switch (checkquantity)
+  {
+    case NOX::Nln::StatusTest::quantity_beaminteraction_lm:
+    {
+      // export the model specific solution if necessary
+      auto rhs_ptr = gstate_ptr_->extract_model_entries(mt, Core::LinAlg::Vector<double>(F));
```
isteinbrecher: I am not very familiar with this part of the code, but I am not seeing where we extract the actual Lagrange multipliers from F here.

Did you check that the vector `lm_rhs_ptr` has exactly the same size as we have Lagrange multipliers? (Also applies to the other functions in this file)

Also, did you check that in the standard structure convergence test, we only check the structural DOFs?

dharinib98: From what I understood about how this works, the `extract_model_entries` extracts the respective DOFs from `F` depending on what `mt` we pass. For `mt = Inpar::Solid::structure` , we extract only the structural DOFs and for `mt = beaminteraction` we extract only the LM DOFs. Using `convert_quantity_type_to_solution_type` and `convert_sol_type2_model_type`, we set the model type to `beaminteraction` for the case `quantity_beaminteraction_lm` for it to extract the LMs from `F`. 

I just checked that `lm_rhs_ptr` has only the LM DOFs. However I am not very sure how to calculate the number of Lagrange multipliers per node, for a two solid element and two beam problem without rotations, it should be 9 LMs ? 

So far for the convergence tests, only `quantity_structure` was called, which meant that the model type was set to `structure` and hence we used to check convergence of only the structural DOFs (yes, I did check this before).

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
dharinib98: yeah, you are right. This has to be removed

tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange.xml
```diff
@@ -0,0 +1,69 @@
...
+          </ParameterList>
+        </ParameterList>
+      </ParameterList>
+      <!-- END: Combo AND - Test 1: "NormWRMS" -->
+    </ParameterList>
+    <!-- END: Combo 0 - Test 0: "Combo" -->
+    <!-- BEGIN: Combo OR - Test 1: "MaxIters" -->
+    <ParameterList name="Test 1">
+      <Parameter name="Test Type" type="string" value="MaxIters"/>
+      <Parameter name="Maximum Iterations" type="int" value="100"/>
```
isteinbrecher: Can we make this smaller? I would be surprised if the tests need more than 10 iterations to converge.

dharinib98: Yeah, that is true. I will change this to 10.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
isteinbrecher: I am not really sure what the `rval` is, but the changes look ok to me. Although, I agree with @mayrmt that the code flow here might indicate suboptimal design.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -200,6 +200,38 @@ double Solid::TimeInt::NoxInterface::get_primary_rhs_norms(const Core::LinAlg::V
...
+
+  // convert the given quantity type to a model type
+  const Inpar::Solid::ModelType mt = Solid::Nln::convert_quantity_type2_model_type(checkquantity);
+  switch (checkquantity)
+  {
+    case NOX::Nln::StatusTest::quantity_beaminteraction_lm:
+    {
+      // export the model specific solution if necessary
+      auto lm_rhs_ptr = gstate_ptr_->extract_model_entries(mt, Core::LinAlg::Vector<double>(F));
+      int_ptr_->remove_condensed_contributions_from_rhs(*lm_rhs_ptr);
```
isteinbrecher: Why do you have this `remove_condensed_contributions_from_rhs` here, but not in `get_lagrange_multiplier_solution_update_norms` or `get_previous_lagrange_multiplier_solution_norms`?

dharinib98: Yeah, this is a contact specific operation and we would not need it.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
dharinib98: I assume `rval` stands for return value, which is used to store the returned norms in each case. I just used this naming scheme from the other get norm functions.

dharinib98:
Should be ready again. I made the necessary changes.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
```diff
@@ -240,7 +240,15 @@ Teuchos::RCP<const std::vector<double>> NOX::Nln::Group::get_rhs_norms(
     if (rval >= 0.0)
     {
       norms->push_back(rval);
+      continue;
     }
+
+    rval = get_nln_req_interface_ptr()->get_lagrange_multiplier_rhs_norms(
+        RHSVector.get_linalg_vector(), chQ[i], type[i],
+        (*scale)[i] == ::NOX::StatusTest::NormF::Scaled);
```
mayrmt: Just to double-check: this hard-codes to use a scaled norm, right? (I am fine with that, just want to make sure, that we all agree on this.)

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required.hpp
```diff
@@ -48,6 +48,12 @@ namespace NOX
             const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
             const bool& isScaled = false) const = 0;
 
+        //! Returns the norms of the Lagrange multiplier Newton RHS.
```
mayrmt:
```suggestion
        //! Return the scaled two-norm of the Lagrange multiplier residual
```

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required.hpp
```diff
@@ -61,12 +67,25 @@ namespace NOX
             const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
             const bool& isScaled = false) const = 0;
 
+        //! Returns the increment norm of the Lagrange Multiplier DoF fields
```
mayrmt:
```suggestion
        //! Return the scaled two-norm of the Lagrange multiplier increment
```

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required.hpp
```diff
@@ -61,12 +67,25 @@ namespace NOX
...
+            const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
+            const bool& isScaled = false) const = 0;
+
         //! Returns the previous solution norm of primary DoF fields
         virtual double get_previous_primary_solution_norms(const Core::LinAlg::Vector<double>& xOld,
             const NOX::Nln::StatusTest::QuantityType& checkQuantity,
             const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
             const bool& isScaled = false) const = 0;
 
+        //! Returns the previous solution norm of Lagrange Multiply DoF fields
```
mayrmt:
```suggestion
        //! Return the scaled two-norm of the Lagrange multiplier solution from previous iteration
```
What exactly does this function do? I do not understand it from the documentation.

Norm of the solution or solution increment?

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required.hpp
```diff
@@ -48,6 +48,12 @@ namespace NOX
             const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
             const bool& isScaled = false) const = 0;
 
+        //! Returns the norms of the Lagrange multiplier Newton RHS.
+        virtual double get_lagrange_multiplier_rhs_norms(const Epetra_Vector& F,
```
mayrmt: "rhs" is wrong. This is the residual, not the right-hand side. There is a difference!

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -344,6 +409,42 @@ double Solid::TimeInt::NoxInterface::get_previous_primary_solution_norms(
   return xoldnorm;
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+double Solid::TimeInt::NoxInterface::get_previous_lagrange_multiplier_solution_norms(
```
mayrmt: Why do we need this implementation? The only difference to `Solid::TimeInt::NoxInterface::get_lagrange_multiplier_rhs_norms()` above are variable names. The computation is exactly the same: extract the norm from an input vector.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.hpp
```diff
@@ -86,6 +86,12 @@ namespace Solid
           const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
           const bool& isscaled = false) const override;
 
+      /*! Get the norm of right hand side rows/entries related to
+       *  Lagrange Multipliers DoFs (derived from NOX::Nln::Interface::Required) */
+      double get_lagrange_multiplier_rhs_norms(const Epetra_Vector& F,
```
mayrmt: rhs != residual

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.hpp
```diff
@@ -101,13 +107,28 @@ namespace Solid
...
+
       /*! Returns the previous solution norm of primary DoF fields
        *  (derived from NOX::Nln::Interface::Required) */
       double get_previous_primary_solution_norms(const Core::LinAlg::Vector<double>& xold,
           const NOX::Nln::StatusTest::QuantityType& checkquantity,
           const ::NOX::Abstract::Vector::NormType& type = ::NOX::Abstract::Vector::TwoNorm,
           const bool& isscaled = false) const override;
 
+      /*! Returns the previous solution norm of lagrange multiplier DoF fields
+       *  (derived from NOX::Nln::Interface::Required) */
```
mayrmt: Please fix comment as outlined above.

tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange.xml
```diff
@@ -0,0 +1,69 @@
...
+            <Parameter name="Quantity Type" type="string" value="Beaminteraction-LM"/>
+            <Parameter name="Tolerance Type" type="string" value="Absolute"/>
+            <Parameter name="Tolerance" type="double" value="1.0e-12"/>
+            <Parameter name="Norm Type" type="string" value="Two Norm"/>
+            <Parameter name="Scale Type" type="string" value="Scaled"/>
+          </ParameterList>
+        </ParameterList>
+      </ParameterList>
+      <!-- END: Combo AND - Test 0: "NormF" -->
+      <!-- BEGIN: Combo AND - Test 1: "NormRMS" -->
```
mayrmt: What does "NormRMS" refer to?

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
dharinib98: We would still be able to control the type of the norm  and whether it is scaled or unscaled from the input files. This would be used as the default settings. However like you said we need to agree on what suits best for our problem because currently the xml also uses the same settings.

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required.hpp
dharinib98: it returns the norm of the previous (last accepted) solution. I will change the documentation accordingly.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
dharinib98: One reason is the two functions operate on different data types, the  get_lagrange_multiplier_rhs_norms operates on the residual vector F (Epetra_Vector) and this on the last solution vector XOld ( LinAlg::Vector).  Another reason is that, that is how the  structural solution update norms are computed due to the current structure of the code. So I    tried to follow the same.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -344,6 +409,42 @@ double Solid::TimeInt::NoxInterface::get_previous_primary_solution_norms(
...
+  // convert the given quantity type to a model type
+  const Inpar::Solid::ModelType mt = Solid::Nln::convert_quantity_type2_model_type(checkquantity);
+
+  switch (checkquantity)
+  {
+    case NOX::Nln::StatusTest::quantity_beaminteraction_lm:
+    {
+      // export the displacement solution if necessary
+      auto model_xold_ptr =
+          gstate_ptr_->extract_model_entries(mt, Core::LinAlg::Vector<double>(xold));
```
mayrmt: This copies the vector. Why? To extract from a vector, one should not need to copy it first.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -296,6 +327,40 @@ double Solid::TimeInt::NoxInterface::get_primary_solution_update_norms(
...
+  // convert the given quantity type to a model type
+  const Inpar::Solid::ModelType mt = Solid::Nln::convert_quantity_type2_model_type(checkquantity);
+  switch (checkquantity)
+  {
+    case NOX::Nln::StatusTest::quantity_beaminteraction_lm:
+    {
+      auto model_incr_ptr =
+          gstate_ptr_->extract_model_entries(mt, Core::LinAlg::Vector<double>(xold));
+      auto model_xnew_ptr =
+          gstate_ptr_->extract_model_entries(mt, Core::LinAlg::Vector<double>(xnew));
```
mayrmt: These lines create copies of `xold` and `xnew`. Extraction from constant vectors should be possible without copying them first.

src/solver_nonlin_nox/4C_solver_nonlin_nox_interface_required.hpp
dharinib98: I would change it to "Return the desired norm of the Lagrange multiplier residual" since the default is overwritten by the input file parameters.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
mayrmt: Regarding vector types, the difference is just the input argument. Internally, everything is converted to `LinAlg::Vector`. This could be unified, e.g. have a function accepting an `Epetra_Vector`, which then does not compute the norm itself, but calls the norm computation function accepting a `LinAlg::Vector`.

Code duplication just makes it look complicated and harder to maintain, although everything performs the same operation.

dharinib98:
@mayrmt @isteinbrecher  This should be ready for review again.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
mayrmt: The function name appears to be very specific, however the function is of general nature: It computes a possible `isscaled` norm of type `type` for a model contribution. The name implies, that it computes Lagrange multiplier norms, though the implementation is not restricted to this. Please rename appropriately.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -200,6 +200,16 @@ double Solid::TimeInt::NoxInterface::get_primary_rhs_norms(const Core::LinAlg::V
   return rhsnorm;
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+double Solid::TimeInt::NoxInterface::get_lagrange_multiplier_rhs_norms(const Epetra_Vector& F,
+    const NOX::Nln::StatusTest::QuantityType& checkquantity,
+    const ::NOX::Abstract::Vector::NormType& type, const bool& isscaled) const
+{
+  return get_previous_lagrange_multiplier_solution_norms(
```
mayrmt: Again, function names are misleading: the function `get_lagrange_multiplier_rhs_norms()` now calls `get_previous_lagrange_multiplier_solution_norms()`, though both functions are not limited to specific Lagrange multiplier quantities, but rather compute norms of a model block in the global model. The only difference is the type of input vector. 

Please use appropriate function names!

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -296,6 +306,38 @@ double Solid::TimeInt::NoxInterface::get_primary_solution_update_norms(
   return updatenorm;
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+double Solid::TimeInt::NoxInterface::get_lagrange_multiplier_solution_update_norms(
```
mayrmt: Also here, the function name implies restriction to Lagrange multipliers, though the function is much more general, since it can compute the norm of _any_ increment defined by `xold` and `xnew`.

dharinib98:
@mayrmt @isteinbrecher  I found a better solution. I just realized we can reuse the existing methods by including another case for quantity `quantity_beaminteraction_lm`. So we do not add any new methods at all for the implementation. I believe this should work.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -266,6 +274,14 @@ double Solid::TimeInt::NoxInterface::get_primary_solution_update_norms(
       }
       break;
     }
+    case NOX::Nln::StatusTest::quantity_beaminteraction_lm:
+    {
+      auto model_incr_ptr = gstate_ptr_->extract_model_entries(mt, xold);
+      auto model_xnew_ptr = gstate_ptr_->extract_model_entries(mt, xnew);
+      model_incr_ptr->update(1.0, *model_xnew_ptr, -1.0);
+      updatenorm = NOX::Nln::Aux::calc_vector_norm(*model_incr_ptr, type, isscaled);
+      break;
```
mayrmt: This is exactly the same code as in the first `case` of this `switch`. Given that we always use the `mt` to specify the block to be extracted, this is not surprising.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
```diff
@@ -314,6 +330,15 @@ double Solid::TimeInt::NoxInterface::get_previous_primary_solution_norms(
...
     }
+    case NOX::Nln::StatusTest::quantity_beaminteraction_lm:
+    {
+      // export the displacement solution if necessary
+      auto model_xold_ptr = gstate_ptr_->extract_model_entries(mt, xold);
+
+      xoldnorm = NOX::Nln::Aux::calc_vector_norm(*model_xold_ptr, type, isscaled);
+
+      break;
+    }
```
mayrmt: This is exactly the same code as in the first `case` of this `switch`. Given that we always use the mt to specify the block to be extracted, this is not surprising.

src/structure_new/src/nonlinear_solver/4C_structure_new_timint_noxinterface.cpp
dharinib98: Oh yeah! Did not notice that. I changed it now.

mayrmt:
@dharinib98 please make sure to squash the commits down to one commit.

mayrmt (approved):
Looks good!

PR #1661, opened 2026-01-16
Migrate monolithic FSI to new solid time integration

georghammerl:
## Description and Context
This draft outlines the changes required to migrate the monolithic FSI approach to the new solid time-integration scheme. At this stage, the adaptive time‑stepping framework is not included; this will follow in a second step. To simplify comparison of results, a temporary boolean switch is introduced in the FSI files to toggle between the old and new time integration. This flag will, of course, be removed once the transition is complete.

The following code modifications are required to enable the new solid time integration for monolithic FSI:
- `StructureNOXCorrectionWrapper` must be applied.
- The call to `initial_guess()` is no longer needed with the new solid time integration.
- For the structure split schemes: In the old time integration, a `BlockSparseMatrix` replaces the `Core::LinAlg::SparseMatrix` to account for the necessary block structure. This approach is highly intrusive and not (easily) feasible in the new solid time integration. Current workaround: whenever the block matrix is needed in the FSI routines, the matrix is copied into a  `BlockSparseMatrix`. This is clearly not a performant solution, but no detailed performance analysis has been conducted yet.
- NOX handles the RHS differently in the new solid time integration, requiring a sign inversion.
- The "FSI specific check of Dirichlet boundary conditions" in the constructor of `MonolithStructureSplit` needs to be moved to `setup_system()` because `dbc_map_extractor` is not available at that point in the new solid time integration.
- The call to `structure_field()->get_constraint_manager()->have_monitor()` has not yet been addressed. Is this required?

Possible remedies to improve the current implementation:
- Move the sign inversion of the solid RHS into `FSIStructureWrapper` to ensure a unified location for solid → FSI RHS handling.
- Revisit the block-matrix handling in `MonolithicStructureSplit` to find a cleaner or more performant solution.

## Interested parties
@mayrmt @maxfirmbach


mayrmt:
@georghammerl

PR #1662, opened 2026-01-16
Refactor Airways in ReducedLung

maxiludwig:
# Description and Context
This PR refactors the Airway logic in `ReducedLung`. The layout is oriented according to the one of the `TerminalUnits`.
 The most significant changes are the addition compliant airway evaluation routines and the refactoring of several helper functions to operate on the new model containers.

### Airway model evaluation routines

Added `4C_reduced_lung_airways.cpp`, which implements functions for evaluating residuals, Jacobians, and derivatives for different model combinations. The `Airway consists of a flow model (linear, nonlinear) and a wall model (rigid, Kelvin-Voigt). Inertia terms can be switched on and off.

### DIFF
--- a/src/reduced_lung/src/4C_reduced_lung_airways.hpp
+++ b/src/reduced_lung/src/4C_reduced_lung_airways.hpp
@@ -1,0 +1,1141 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+
+
+#ifndef FOUR_C_REDUCED_LUNG_AIRWAYS_HPP
+#define FOUR_C_REDUCED_LUNG_AIRWAYS_HPP
+
+#include "4C_config.hpp"
+
+#include "4C_fem_discretization.hpp"
+#include "4C_linalg_sparsematrix.hpp"
+#include "4C_linalg_vector.hpp"
+#include "4C_reduced_lung_input.hpp"
+#include "4C_utils_exceptions.hpp"
+
+#include <cstddef>
+#include <cstdlib>
+#include <functional>
+#include <type_traits>
+#include <variant>
+#include <vector>
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace ReducedLung
+{
+  namespace Airways
+  {
+    /**
+     * @brief Shared data container for all airway elements.
+     *
+     * Stores identifiers and physical parameters (length, area, etc.) for each
+     * airway, which are used across all flow and wall models.
+     *
+     * @note This data is independent of specific models and is shared across evaluators.
+     */
+    struct AirwayData
+    {
+      // Global element IDs.
+      std::vector<int> global_element_id;
+      // Local element IDs.
+      std::vector<int> local_element_id;
+      // IDs of the local rows in the row map. This defines the place for these airways in the
+      // system of equations.
+      std::vector<int> local_row_id;
+      // Global dof IDs
+      std::vector<int> gid_p1, gid_p2, gid_q1, gid_q2;
+      // Dof IDs from locally relevant / column map. Used for lookup in dof-vector during assembly.
+      std::vector<int> lid_p1, lid_p2, lid_q1, lid_q2;
+
+      // Physical quantities of each airway.
+      std::vector<double> ref_length;
+      std::vector<double> ref_area;
+      struct AirProperties
+      {
+        double dynamic_viscosity;
+        double density;
+      } air_properties;
+
+      // Dof values at previous time step
+      std::vector<double> q1_n;
+      std::vector<double> q2_n;
+      std::vector<double> p1_n;
+      std::vector<double> p2_n;
+
+      int n_state_equations;  // number of state equations per airway
+      // convenience
+      [[nodiscard]] size_t number_of_elements() const { return global_element_id.size(); }
+    };
+
+    // ----- Flow resistance models -----
+
+    /**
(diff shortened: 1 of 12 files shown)


src/reduced_lung/src/4C_reduced_lung_airways.hpp
```diff
@@ -0,0 +1,1143 @@
...
+    {
+      std::vector<bool> has_inertia;
+    };
+
+    /**
+     * @brief Nonlinear resistive model considering turbulence effects and convective resistance
+     * from momentum flux.
+     *
+     * Airway resistance is defined as \f$R = R_p \, k_{turb} + \frac{2 \alpha
+     * \rho}{A^2}(Q_2-Q_1)\f$, where \f$_p\f$ is the Poiseuille resistance and \f$k_{turb}\f$ is a
```
bgoderbauer: Thanks for tackling this! 😄 Looks very nice already
AFAIK, we do not render LaTeX in the docs?

src/reduced_lung/src/4C_reduced_lung_main.cpp
```diff
@@ -166,13 +142,26 @@ namespace ReducedLung
...
+        model.data.gid_p1.push_back(first_dof_gid);
+        model.data.gid_p2.push_back(first_dof_gid + 1);
+        model.data.gid_q1.push_back(first_dof_gid + 2);
+        if (model.data.n_state_equations == 2)
+        {
+          model.data.gid_q2.push_back(first_dof_gid + 3);
+        }
+        else if (model.data.n_state_equations == 1)
+        {
+          // do nothing, remains empty
```
bgoderbauer:
```suggestion
          // rigid airways -> only 3 unknowns
```

src/reduced_lung/src/4C_reduced_lung_airways.hpp
```diff
@@ -0,0 +1,1143 @@
...
+    {
+      /**
+       * Creates the model-specific derivative evaluator of the flow resistance term.
+       * The evaluators return the derivatives of the flow resistance term w.r.t. the flow DOFs.
+       * \f$ \frac{\partial f_{1,QR}}{\partial Q_1} \f$ for rigid walls
+       * and \f$ \frac{\partial f_{1,QR}}{\partial Q_1} \f$, \f$ \frac{\partial f_{1,QR}}{\partial
+       * Q_2}
+       * \f$ for Kelvin-Voigt walls.
+       */
+      struct MakeFlowResistanceDerivativeEvaluator
```
bgoderbauer, line 677: I guess that's the price we have to pay to keep things flexible. I think we need to be careful not to overengineer things. Here, we allow differentiation of wall models, but we also impose hard constraints so that each model stores only dR/dQ as a pair. This doesn't feel like an easy-to-understand solution to me. I was stumbling over the details when I was wondering what `resistance.first` and `resistance.second` should mean. I would suggest we leave it like this for now and have a look maybe when we restructure the files.

src/reduced_lung/tests/4C_reduced_lung_airways_test.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+
+namespace
+{
+  using namespace FourC::ReducedLung;
+  using namespace FourC::Core::LinAlg;
+  using namespace FourC::ReducedLung::Airways;
+
+
+  TEST(AirwayTests, JacobianVsFiniteDifferenceRigidWall)
```
bgoderbauer, line 29: Thanks for the extensive test suite! 😄

src/reduced_lung/src/4C_reduced_lung_airways.hpp
maxiludwig: You're right. These `FlowResistanceDerivativeEvaluator` are somehow arbitrarily the derivatives w.r.t to the flow DOFs (Q1 (and Q2)). The differnt signatures with two different return types result from the flexible setup of rigid walls (1 flow dof) and compliant walls (2 flow dofs). I am fine with leaving it for now, and trying to find a better solution with the upcoming restucturing.

src/reduced_lung/src/4C_reduced_lung_airways.hpp
```diff
@@ -0,0 +1,1143 @@
...
+    struct AirwayContainer
+    {
+      std::vector<AirwayModel> models;
+    };
+
+    // compile-time traits for number of state equations contributed by each model type
+    template <typename F>
+    struct FlowModelStateCount
+    {
+      static constexpr int value = 0;
```
amgebauer:
```suggestion
```

Do you need the default here? If you don't define the default here, you get a compile time error if you add a new resistive model

src/reduced_lung/src/4C_reduced_lung_airways.hpp
```diff
@@ -0,0 +1,1143 @@
...
+    template <>
+    struct FlowModelStateCount<NonLinearResistive>
+    {
+      static constexpr int value = 1;
+    };
+
+    template <typename W>
+    struct WallModelStateCount
+    {
+      static constexpr int value = 0;
```
amgebauer:
```suggestion
```
Same as above. I would not define the default value here to force to explicitly set these values

amgebauer (dismissed):

bgoderbauer (approved):

src/reduced_lung/src/4C_reduced_lung_airways.hpp
amgebauer: @maxiludwig Did you forget to commit the changes?

maxiludwig: Oups, now I've added them 😅

bgoderbauer (approved):

PR #1663, opened 2026-01-16
Remove redundant thermo vtk input parameters

rjoussen:
## Description and Context

This removes the `THERMAL DYNAMIC/RUNTIME VTK OUTPUT`'s `HEATFLUX` and `TEMPGRAD` options. Neither of them can be used anyway, and postrpocessing them is rather easy.


c-p-schmidt (approved):

davidrudlstorfer (approved):

rjoussen:
I added breaking change, as input files with these parameters set to false could now fail due to incorrect input.

maxfirmbach:
I just want to highlight that these variables were in place to highlight that both of them are not ported to runtime output yet, as they are still using the old output mechanism based on `Thermo::TimInt::output_heatflux_tempgrad` and post-processing!

We need to decide if it is worth porting them or not, but the changes made kind of imply it is not worth porting them. I don't use them, but we should still think about it though.

PR #1664, opened 2026-01-16
[thermo] Fix pvd file for restarts

rjoussen:
## Description and Context

This allows for complete pvd collection files in restarted thermo simulations.

The issue was that the vtk writer for thermo problems is initiated with `time_[0]`. However, at this point the time was not updated to the restart time yet (only happens later in `read_restart`). As an easy fix, we hand over the restart time to the vtk writer initialization directly. 

I used the `four_c_test_restarted_vtk` testing function to test this behaviour.


tests/input_files/thermo3D_FBC_ost.4C.yaml
```diff
@@ -13,12 +13,21 @@ DISCRETISATION:
 IO:
   THERM_HEATFLUX: "Initial"
   THERM_TEMPGRAD: "Initial"
+  WRITE_INITIAL_STATE: false
```
rjoussen, line 16: this is just because I have another fix incoming which will lead to `WRITE_INITIAL_STATE` being actually used in the thermo time integration. At the moment, `WRITE_INITIAL_STATE` is completely irrelevant, but I don't want to change this test again then.

maxfirmbach:
@rjoussen Would you mind providing the `.vtu` files in `ascii`? Will make changes to these files in the future more clear, e.g. updating values etc.

rjoussen:
@maxfirmbach I converted them.

maxfirmbach:
Thanks!

src/thermo/src/implicit/4C_thermo_timint.cpp
```diff
@@ -130,21 +130,26 @@ Thermo::TimInt::TimInt(const Teuchos::ParameterList& ioparams,
...
           .output_element_owner = output_element_owner,
           .output_element_gid = output_element_gid,
           .output_node_gid = output_node_gid};
 
       runtime_vtk_writer_ = Core::IO::DiscretizationVisualizationWriterMesh(
           discret_, Core::IO::visualization_parameters_factory(
                         Global::Problem::instance()->io_params().sublist("RUNTIME VTK OUTPUT"),
-                        *Global::Problem::instance()->output_control_file(), time_[0]));
+                        *Global::Problem::instance()->output_control_file(),
+                        Global::Problem::instance()->restart()));
```
c-p-schmidt: The last parameter should be a time, but `Global::Problem::instance()->restart()` returns the step from which it is restarted.

Does this only work because the time step size in the test case is 1, i.e., time and time step are the same?

rjoussen: That's a very good catch! Yes, the dt=1 was exactly what happened

rjoussen:
I updated it now such that the correct time state is set in the thermo time integrator constructor, before initializing the vtk writer. The read_restart method now only reads in the actual data, as the correct time state is already set. 
Otherwise the vtk writer initialization would need to move somewhere else, potentially in a setup method, which would introduce a lot more changes.
> You could consider reducing the output frequency so that only fewer reference files are needed.

I increased `RESULTSEVERY` to 2, which is the maximum to test that the restart pvd file is correctly read and appended to.

c-p-schmidt (approved):

maxfirmbach:
I think this is ready to be merged.

PR #1665, opened 2026-01-17
[thermo] Enable `WRITE_INITIAL_STATE`

rjoussen:
## Description and Context

`WRITE_INITIAL_STATE` now has the desired effect of writing output at `t=0` if it was set to true in the input. Otherwise, it has no effect. This mainly required an optional `output` call in `thermo_dyn_drt()`, as well as a guard in thermo's `output_step()` to prevent restart output at the first (restarted) step.


src/thermo/src/implicit/4C_thermo_timint.cpp
```diff
@@ -438,35 +438,37 @@ void Thermo::TimInt::write_runtime_output()
  *----------------------------------------------------------------------*/
 void Thermo::TimInt::output_step(bool forced_writerestart)
 {
+  bool is_initial_step = (step_ == Global::Problem::instance()->restart());
```
c-p-schmidt: Shouldn't this be the check for the initial state instead? Or do you also consider the first step to be restarted from as an initial step?

```suggestion
  bool is_initial_step = (step_ == 0);
```

src/thermo/src/4C_thermo_dyn.cpp
```diff
@@ -26,10 +26,16 @@ void thermo_dyn_drt()
 
   // do restart if demanded from input file
   const int restart = Global::Problem::instance()->restart();
+  const int write_initial_state =
+      Global::Problem::instance()->io_params().get<bool>("WRITE_INITIAL_STATE");
   if (restart)
   {
     algorithm->thermo_field()->read_restart(restart);
   }
+  else if (write_initial_state)
```
c-p-schmidt, line 35: Isn't that a contradiction to what you had above? Because of the `else if`, the initial state is written only if it is not a restart simulation.

However, the logic in the file `src/thermo/src/implicit/4C_thermo_timint.cpp` was that is_initial_state is true if the current step equals the restart_step.

tests/input_files/thermo3D_annulus_elementwise_anisotropic_statics.4C.yaml
```diff
@@ -2,6 +2,7 @@ PROBLEM TYPE:
   PROBLEMTYPE: "Thermo"
 IO:
   VERBOSITY: "Standard"
+  WRITE_INITIAL_STATE: false
```
sebproell, line 5: It might be better to introduce a new test for this feature. Hiding the test of a certain feature within another test is a sure way to delete it at some point. The test can be super small, which would also decrease the diff.

src/thermo/src/implicit/4C_thermo_timint.cpp
rjoussen: Yes, no restart output should be written if `step_` is the first step of this simulation (restarted or not). Previously, this was only checked for forced restart output. The check in case of a regular restart step was not needed previously, since `ouptut_step` was never called at the start of the simulation before (except for with forced restart). You're right that `output_step` is currently never called without forced restart at an initial step that is not 0, but I think its not an issue to protect against that anyway.

src/thermo/src/4C_thermo_dyn.cpp
rjoussen: No, the initial state should only be written if this is not the start of the simulation (restarted or not). The `is_initial_state` in `output_step` prevents that restart information is written at step 0 (which is necessary) **and** that no restart information is written at the beginning of a restarted simulation (which is not strictly necessary since this never happens anyway).

tests/input_files/thermo3D_annulus_elementwise_anisotropic_statics.4C.yaml
rjoussen: Setting `WRITE_INITIAL_STATE: false` is necessary here in order to not to modify the reference data of this test. I also did that now for the other test (thermo3D_annulus_anisotropic...)

Instead, this feature is now tested alongside the corrected restart vtk output for thermo problems introduced in #1664, which makes the diff much smaller. I changed the name of that test to clearly reflect what it tests. I think it makes sense to test both in a single test, since both features ensure that the vtk output works as expected for thermo problems.

Issue #1666, opened 2026-01-21
[BUG] Failing nightly pipeline


maxiludwig:
## Bug Description: What's going wrong?

After https://github.com/4C-multiphysics/4C/pull/1490 has been merged, the following benchmarktest failed:

`benchmarktests_solid_3D_ele`

## Expectations

Test should not fail 

## Possible Solution

Initialize the communicator config correctly.

maxiludwig:
#1667 fixed the pipeline. Closing the issue

PR #1667, opened 2026-01-21
Fix uninitialized CommConfig in solid 3D benchmark

maxiludwig:
## Description and Context
The benchmark test was creating a default-constructed CommConfig without initializing the np_type field, leading to undefined behavior and a segmentation fault in create_comm().

Initialize CommConfig by default with: 
- empty group_layout
- no_nested_parallelism for np_type
- default diffgroup value (-1)

Fixes the benchmarktests_solid_3D_ele test failure.

#1666 #1490


maxiludwig:
You're right. Would be the better solution :)
I'll change this

PR #1668, opened 2026-01-21
[particle] Extract binning related variables into struct

ppraegla:
All variables related to binning are moved to a struct. This is a preparatory step to share the binning with two ParticleEngines.


slfuchs:
Looks good!

PR #1676, opened 2026-01-26
[TSI] Enable vtk output for thermo field

rjoussen:
## Description and Context

The temperature (rate) in the thermo field shouldn't be written node-based, but dof-based. For monolithic tsi, this made it impossible to write thermo vtk output, since then dof and node ids differ for the thermo field (after calling `TSI::Monolithic::setup_system()`). 

The added tests ensure that writing thermo **and** structure vtk output works for both monolithic and partioned tsi (since they have a different `setup_system` function). The existing thermo vtk output tests still pass, i.e. writing the temperature as dof-based even in purely thermal simulations is not an issue.


rjoussen:
I also think this makes it possible to remove the tsi-specific `RESULTSEVERY`, since all it does can now be done with vtk output in the fields individually, for which `IO/RUNTIME VTK OUTPUT/INTERVAL_STEPS` should be responsible. Right now, the tsi `RESULTSEVERY` overwrites the thermal one, which manages the frequency of vtk output of the thermofield, while the structure vtk output is managed by `IO/RUNTIME VTK OUTPUT/INTERVAL_STEPS`.

rjoussen:
@maxfirmbach yes, that's why I assume this wasn't an issue until now.

PR #1678, opened 2026-01-26
[TSI] Use `WRITE_INITIAL_STATE`

rjoussen:
## Description and Context

This enables using `WRITE_INITIAL_STATE`  for tsi simulations. If true, this now writes the structure and thermo vtk output at step 0.

This is tested by amending the reference data of `tsi_heatflux_monolithic` vtk compare test. 

#1665
#1676


src/tsi/4C_tsi_dyn.cpp
```diff
@@ -76,19 +76,28 @@ void tsi_dyn_drt()
   }  // end switch
 
   const int restart = Global::Problem::instance()->restart();
+  const int write_initial_state =
```
c-p-schmidt: Why not `const bool`, or `const auto`?

src/tsi/4C_tsi_dyn.cpp
```diff
@@ -76,19 +76,28 @@ void tsi_dyn_drt()
...
+    // now do the coupling setup and create the combined dofmap
+    tsi->setup_system();
   }
   else
   {
     // run post_setup to update the structure field
     tsi->post_setup();
-  }
+    // now do the coupling setup and create the combined dofmap
+    tsi->setup_system();
```
c-p-schmidt: It looks a bit strange that `setup_system` is performed after `post_setup`. But since it used to be like that before, this probably isn't an issue.

src/tsi/4C_tsi_dyn.cpp
maxfirmbach: I agree, this looks strange. I think `const bool` does the job.

rjoussen: Absolutely! I changed it to `const bool`

c-p-schmidt (dismissed):

src/tsi/4C_tsi_dyn.cpp
rjoussen: Yes, it seems to work for now, but it's probably good to keep that in mind in case that leads to problems in the future.

maxfirmbach (dismissed):

amgebauer:
@tuchpaul I think SSI also needs to take care of this flag

rjoussen:
Following #1975, this should work now without any problems.

I realized that initial structure `output()` requires calling `prepare_output()` with the `force_prepare` set to `true` before. I found this quite hard to understand and hence added a doxygen comment to the respective `prepare_output()` function that of the structure adapter (this flag to eventually propagated (more or less convolutedly) to `Solid::TimeInt::Base::prepare_output()` where its actually being used. This is done in the second commit now.

src/adapter/4C_adapter_str_structure.hpp
```diff
@@ -365,7 +365,14 @@ namespace Adapter
     /// Access to output object
     virtual std::shared_ptr<Core::IO::DiscretizationWriter> disc_writer() = 0;
 
-    /// prepare output (i.e. calculate stresses, strains, energies)
+    /**
```
c-p-schmidt, line 368: You do not need to replicate the docs if nothing has changed compared to the parent. Methods that override automatically use the docs of the parent.

src/adapter/4C_adapter_str_structure_new.hpp
```diff
@@ -286,7 +286,14 @@ namespace Adapter
     /// Access to output object
     std::shared_ptr<Core::IO::DiscretizationWriter> disc_writer() override = 0;
 
-    /// prepare output (i.e. calculate stresses, strains, energies)
+    /**
```
c-p-schmidt: Same for all other methods.

c-p-schmidt (approved):

src/adapter/4C_adapter_str_structure.hpp
rjoussen: The issue is that the force_prepare_flag is specific to the solid field, yet it is already present in the generic field adapter. I don't want to document it higher up, where a lot of implementations don't even use the flag. In this case, having the documentation here would have helped me understand what this flag actually does and when it is required. Also, the parent documentation might show up in doxygen, but clangd doesn't pick it up, unfortunately.

maxfirmbach (approved):
LGTM!

PR #1685, opened 2026-01-27
Improve ArborX interface

isteinbrecher:
## Description and Context
Improve our ArborX interface. The query function to `BoundingVolumeHierarchy` now returns the same data structures that ArborX returns, thus allowing for a more flexible usage of our implementation of `BoundingVolumeHierarchy`.

### DIFF
--- a/src/core/geometric_search/src/4C_geometric_search_bvh.hpp
+++ b/src/core/geometric_search/src/4C_geometric_search_bvh.hpp
@@ -18,6 +18,9 @@
 
 #include "4C_geometric_search_access_traits.hpp"
 
+#include <Kokkos_Core.hpp>
+
+#include <tuple>
 #include <vector>
 
 FOUR_C_NAMESPACE_OPEN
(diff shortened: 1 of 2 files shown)


src/core/geometric_search/src/4C_geometric_search_bvh.hpp
```diff
@@ -76,44 +77,49 @@ namespace Core::GeometricSearch
     {
     }
 
-    /*
-     * @brief Query the bounding volume hierarchy with a given vector of predicates.
+    /*! \brief Query the bounding volume hierarchy with a given vector of predicates and record
+     * results in {indices, offsets}.
+     *
+     * (From ArborX documentation)
```
isteinbrecher, line 89: @m-frey this seems quite complicated, but is actually very simple. With this data structure, your search should be very easy to evaluate and also verify.

src/core/geometric_search/src/4C_geometric_search_bvh.hpp
```diff
@@ -54,8 +57,10 @@ namespace Core::GeometricSearch
           "The class 'Core::GeometricSearch::BoundingVolumeHierarchy' can only be used with ArborX."
           "To use it, enable ArborX during the configure process.");
     }
+    using MemorySpace = typename Kokkos::DefaultExecutionSpace::memory_space;
```
maxfirmbach, line 61: This is a little bit cramped in here. Would it make sense to put it on top to the other alias?

maxfirmbach: Maybe also leave a blank space in between the methods.

isteinbrecher: This has to stay here, since it is compiled in the non-arborx case. I added a new line.

PR #1688, opened 2026-01-28
Use external node set names in condition definitions

rjoussen:
## Description and Context

This proposes to use the (string) names specified by some external mesh files (currently exodus and gmsh) as optional identifiers in the input specification of boundary conditions. I think it is a lot nicer to be able to identify node sets by their names rather than remembering their IDs all the time. This would look as follows in the input:
```yaml
DESIGN POINT DIRICH CONDITIONS:
  - NODE_SET_NAME: left_faces
    NUMDOF: 3
    ONOFF: [1, 1, 1]
    VAL: [0, 0, 0]
    FUNCT: [0, 0, 0]
    TAG: monitor_reaction
```
This requires making the previous `E: ` entry optional. `E:` cannot be used for both as it cannot take string and int values at the same time. At runtime, it is ensured that `E` and `NODE_SET_NAME` are not given at the same time. 

`ENTITY_TYPE` is optional as well, and assumed to be `node_set_name` if only a `NODE_SET_NAME` is provided, optherwise it falls back to `legacy_id`.

Only once a boundary condition is (successfully) identified by its `node_set_name`, it carries a value for that parameter, which is later used for the filename in the dbc writer (So for example if a boundary condition is identified as "left_support", the filename will be "basename-left_support_monitor_dbc.csv". If a BC is identified via `E: ID` the naming remains as before).

This is mainly tested using gmsh input for solids with a yaml and dbc writer. In the last commit, I modified the contact tutorial to show that this also works for exodus files. ~However, this would also require changing some things in the actual tutorial text, so I will remove that again before merging.~ 

Note: I moved the `get_node_sets_from_mesh` and `get_element_block_nodes_from_mesh` functions from `global_data_read` into the meshreader and modified `get_node_sets_from_mesh` there.

#825

### DIFF
--- a/src/core/fem/src/condition/4C_fem_condition.cpp
+++ b/src/core/fem/src/condition/4C_fem_condition.cpp
@@ -14,13 +14,15 @@ FOUR_C_NAMESPACE_OPEN
 
 Core::Conditions::Condition::Condition(const int id, const Core::Conditions::ConditionType type,
     const bool buildgeometry, const Core::Conditions::GeometryType gtype,
-    const EntityType entity_type, const std::optional<std::string>& node_set_name)
+    const EntityType entity_type, const std::vector<int>& nodes,
+    IO::InputParameterContainer parameters, const std::optional<std::string>& node_set_name)
     : id_(id),
       buildgeometry_(buildgeometry),
       type_(type),
       gtype_(gtype),
       entity_type_(entity_type),
-      node_set_name_(node_set_name)
+      node_set_name_(node_set_name),
+      container_(std::move(parameters))
 {
   if (entity_type_ == EntityType::node_set_name)
   {
@@ -28,6 +30,8 @@ Core::Conditions::Condition::Condition(const int id, const Core::Conditions::Con
         "Condition {} of ENTITY_TYPE 'node_set_name' requires a NODE_SET_NAME to be provided.",
         id_);
   }
+
+  set_nodes(nodes);
 }
 
 std::ostream& operator<<(std::ostream& os, const Core::Conditions::Condition& cond)
@@ -76,9 +80,7 @@ std::unique_ptr<Core::Conditions::Condition> Core::Conditions::make_condition(
     const Core::Conditions::ConditionSpec& condition_spec,
     const Core::Conditions::InputNodeSets& input_node_sets)
 {
-  std::unique_ptr<Core::Conditions::Condition> condition;
-
-  switch (condition_spec.type)
+  switch (condition_spec.entity_type)
   {
     case Core::Conditions::EntityType::legacy_id:
     {
@@ -95,7 +97,7 @@ std::unique_ptr<Core::Conditions::Condition> Core::Conditions::make_condition(
             "This is probably because the geometry is handled in an external file.\n"
             "If this is the case, you must specify a specific entity type (node_set_id or "
             "element_block_id) or identify the node set via its name.\n",
-            condition_spec.type, id);
+            condition_spec.entity_type, id);
       }
       switch (condition_spec.geometry_type)
       {
@@ -108,11 +110,10 @@ std::unique_ptr<Core::Conditions::Condition> Core::Conditions::make_condition(
                 "Could not read set from entity type.",
                 id, input_node_sets.dnode_fenode.size());
           }
-          condition =
-              std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
-                  condition_spec.build_geometry, condition_spec.geometry_type, condition_spec.type);
-          condition->set_nodes(input_node_sets.dnode_fenode[id]);
-          break;
+          return std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
+              condition_spec.build_geometry, condition_spec.geometry_type,
+              condition_spec.entity_type, input_node_sets.dnode_fenode[id],
+              condition_spec.condition_data);
         case Core::Conditions::geometry_type_line:
           if (id < 0 or static_cast<unsigned>(id) >= input_node_sets.dline_fenode.size())
           {
@@ -122,11 +123,10 @@ std::unique_ptr<Core::Conditions::Condition> Core::Conditions::make_condition(
                 "Could not read set from entity type.",
                 id, input_node_sets.dline_fenode.size());
           }
-          condition =
-              std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
-                  condition_spec.build_geometry, condition_spec.geometry_type, condition_spec.type);
-          condition->set_nodes(input_node_sets.dline_fenode[id]);
-          break;
+          return std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
+              condition_spec.build_geometry, condition_spec.geometry_type,
+              condition_spec.entity_type, input_node_sets.dline_fenode[id],
+              condition_spec.condition_data);
         case Core::Conditions::geometry_type_surface:
           if (id < 0 or static_cast<unsigned>(id) >= input_node_sets.dsurf_fenode.size())
           {
@@ -136,11 +136,10 @@ std::unique_ptr<Core::Conditions::Condition> Core::Conditions::make_condition(
                 "Could not read set from entity type.",
                 id, input_node_sets.dsurf_fenode.size());
           }
-          condition =
-              std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
-                  condition_spec.build_geometry, condition_spec.geometry_type, condition_spec.type);
-          condition->set_nodes(input_node_sets.dsurf_fenode[id]);
-          break;
+          return std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
+              condition_spec.build_geometry, condition_spec.geometry_type,
+              condition_spec.entity_type, input_node_sets.dsurf_fenode[id],
+              condition_spec.condition_data);
         case Core::Conditions::geometry_type_volume:
           if (id < 0 or static_cast<unsigned>(id) >= input_node_sets.dvol_fenode.size())
           {
@@ -149,17 +148,14 @@ std::unique_ptr<Core::Conditions::Condition> Core::Conditions::make_condition(
                 "DVolume condition on non existent DVolume?",
                 id, input_node_sets.dvol_fenode.size());
           }
-          condition =
-              std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
-                  condition_spec.build_geometry, condition_spec.geometry_type, condition_spec.type);
-          condition->set_nodes(input_node_sets.dvol_fenode[id]);
+          return std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
+              condition_spec.build_geometry, condition_spec.geometry_type,
+              condition_spec.entity_type, input_node_sets.dvol_fenode[id],
+              condition_spec.condition_data);
           break;
         default:
           FOUR_C_THROW("geometry type unspecified");
-          break;
       }
-
-      break;
     }
     case Core::Conditions::EntityType::node_set_id:
     {
@@ -168,10 +164,9 @@ std::unique_ptr<Core::Conditions::Condition> Core::Conditions::make_condition(
       FOUR_C_ASSERT_ALWAYS(input_node_sets.node_sets.contains(id),
           "Cannot apply condition '{}' to node set {} which is not specified in the mesh file.",
           condition_spec.condition_type, id);
-      condition = std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
-          condition_spec.build_geometry, condition_spec.geometry_type, condition_spec.type);
-      condition->set_nodes(input_node_sets.node_sets.at(id));
-      break;
+      return std::make_unique<Core::Conditions::Condition>(id, condition_spec.condition_type,
+          condition_spec.build_geometry, condition_spec.geometry_type, condition_spec.entity_type,
+          input_node_sets.node_sets.at(id), condition_spec.condition_data);
     }
     case Core::Conditions::EntityType::node_set_name:
     {
@@ -186,11 +181,9 @@ std::unique_ptr<Core::Conditions::Condition> Core::Conditions::make_condition(
           "NODE_SET_NAME '{}' is not unique in the meshfile ({} occurrences).", node_set_name,
           ids.size());
 
-      condition = std::make_unique<Core::Conditions::Condition>(ids[0],
-          condition_spec.condition_type, condition_spec.build_geometry,
-          condition_spec.geometry_type, condition_spec.type, node_set_name);
-      condition->set_nodes(input_node_sets.node_sets.at(ids[0]));
-      break;
+      return std::make_unique<Core::Conditions::Condition>(ids[0], condition_spec.condition_type,
+          condition_spec.build_geometry, condition_spec.geometry_type, condition_spec.entity_type,
+          input_node_sets.node_sets.at(ids[0]), condition_spec.condition_data, node_set_name);
     }
     case Core::Conditions::EntityType::element_block_id:
     {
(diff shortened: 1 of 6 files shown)


src/core/fem/src/condition/4C_fem_condition_definition.cpp
```diff
@@ -38,9 +39,13 @@ Core::Conditions::ConditionDefinition::ConditionDefinition(std::string sectionna
...
-  add_component(
-      parameter<int>("E", {.description = "ID of the condition. This ID refers to the respective "
-                                          "topological entity of the condition."}));
+  add_component(parameter<std::optional<int>>(
+      "E", {.description = "ID of the condition. This ID refers to the respective "
+                           "topological entity of the condition. Not allowed if ENTITY_TYPE is "
+                           "NODE_SET_NAME."}));
+  add_component(parameter<std::optional<std::string>>("NODE_SET_NAME",
+      {.description = "This refers to the respective node set name in the external mesh file. Only "
+                      "allowed if ENTITY_TYPE is NODE_SET_NAME and no E: ID is given."}));
```
amgebauer: What you could also do is either you specify `E` and `EntityType`, or you specify just the name.

In pseudo-input spec code:

```
one_of({
   group({component<int>("E", ...), component<EntityType>("ENTITY_TYPE", ...)}),
   component<std::string>("NODE_SET_NAME",...)
})
```

src/core/fem/src/condition/4C_fem_condition_definition.cpp
```diff
@@ -81,22 +87,69 @@ void Core::Conditions::ConditionDefinition::read(Core::IO::InputFile& input,
...
+        "Cannot apply {} condition with external name '{}' which is not unique in the mesh file "
+        "({} occurrences found).",
+        condtype_, node_set_name.value(), ids.size());
+    return ids[0];
+  }
+
+  // Other entity types
+  FOUR_C_ASSERT_ALWAYS(id.has_value() && id.value() >= 0,
+      "{} condition of entity type {} requires a non-negative ID specified via E: ID.", condtype_,
+      entity_type);
```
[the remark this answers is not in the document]
rjoussen: Addressed in L145

davidrudlstorfer:
Just a small note from my side: With the last changes to the metadata file (#1669) we've seen a problem with FourCIPP (https://github.com/4C-multiphysics/fourcipp/pull/115)

I have not checked the code of this PR further but it may also be a breaking change for other forms of input. @rjoussen maybe you could be so kind and quickly run the fourcipp testsuite with the metadata file created from this PR and check if everything is still passing.

rjoussen:
@davidrudlstorfer Thanks for the hint, I was not aware that this could happen. I will look into it :)

src/core/fem/src/condition/4C_fem_condition_definition.cpp
```diff
@@ -38,12 +39,17 @@ Core::Conditions::ConditionDefinition::ConditionDefinition(std::string sectionna
...
-  add_component(parameter<Core::Conditions::EntityType>(
-      "ENTITY_TYPE", {.description = "The type of entity that E refers to.",
-                         .default_value = Core::Conditions::EntityType::legacy_id}));
+  add_component(parameter<std::optional<int>>(
+      "E", {.description = "ID of the condition. This ID refers to the respective "
+                           "topological entity of the condition. Not allowed if ENTITY_TYPE is "
+                           "NODE_SET_NAME."}));
+  add_component(parameter<std::optional<std::string>>("NODE_SET_NAME",
+      {.description = "This refers to the respective node set name in the external mesh file. Only "
+                      "allowed if ENTITY_TYPE is NODE_SET_NAME and no E: ID is given."}));
```
gilrrei: Can this be wrapped in a `one_of`? I think this would fit this scenario perfectly.

src/core/fem/src/condition/4C_fem_condition_definition.cpp
```diff
@@ -81,22 +87,69 @@ void Core::Conditions::ConditionDefinition::read(Core::IO::InputFile& input,
...
 
-    cmap.emplace(id, condition);
+    if (entity_type == EntityType::node_set_name)
+      condition->set_node_set_name(node_set_name.value());
+
+    cmap.emplace(resolved_id, condition);
+  }
+}
+
+int Core::Conditions::ConditionDefinition::resolve_entity_id(EntityType entity_type,
```
amgebauer: Do we need to resolve an internal id? In a vtu-mesh for example, we would not have an id if we make a name-based input of point-sets. Of course, we could just generate one, but still, I would find it cleaner if we just pass the set of points to the condition instead of an internal id of the point set.

src/core/fem/src/condition/4C_fem_condition_definition.cpp
amgebauer: Duplicate of https://github.com/4C-multiphysics/4C/pull/1688#discussion_r2737128031

src/core/fem/src/condition/4C_fem_condition_definition.cpp
rjoussen: @gilrrei @amgebauer I initially wanted to use a `one_of`, because this would be a really nice use case for that, indeed. However, there are some conditions (namely `S2IKinetics`, `ScatraMultiScaleCoupling`, `ElchBoundaryKinetics` and `SSISurfaceManifoldKinetics`) which already use `one_of`s in their definitions and multiple parallel `one_of` in the same all_of spec are not allowed (which I think makes sense, it gets out of hand very quickly otherwise). I don't think its easy to refactor them and could not be done without any breaking changes. That's why its done like this, though I don't think its optimal either.

rjoussen: What I could do is to make entity type optional as well such that it does not need to be specified when the condition is identified via node set name. This would make the input a bit lighter.

src/core/fem/src/condition/4C_fem_condition_definition.cpp
rjoussen: I am not sure if I completely get what you mean. Even if conditions _would_ not need an internal ID, point sets definitely do since they must work for all input formats. Otherwise, we would need to check everywhere whether a `PointSet` is a point set with or without an ID.  I assume that's also why we store them in  `int, PointSet` maps. 
Besides, Conditions actually do need an ID, because it's used at some places to identify them (even though the occurrences of `Condition.id()` look a little suspicious). So I think I would keep the IDs. 

Nevertheless, it could be nice to move the `Condition.set_nodes()` part into `ConditionDefinition::read()`, which would make things a little easier, maybe. Especially, this could mean that the `EntityType` switch is not needed at multiple places. But I think that's a separate issue from this PR.

amgebauer: The source of my comment is that I don't like the idea of this method and tried to avoid it. It needs a bunch of data that we need to collect first so that we can convert the name-based input to an id-based input. But I agree, refactoring this and passing the point-sets to the constructor is not straight forward and cannot be fixed in this PR.

src/core/fem/src/condition/4C_fem_condition_definition.cpp
amgebauer: ok. I think the condition input needs some overhaul. So for now, both ways are fine for me.

src/core/fem/src/condition/4C_fem_condition.cpp
```diff
@@ -54,5 +54,13 @@ std::shared_ptr<Core::Conditions::Condition> Core::Conditions::Condition::copy_w
   return copy;
 }
 
+std::string Core::Conditions::Condition::node_set_name() const
+{
+  if (node_set_name_)
+    return node_set_name_.value();
+  else
+    FOUR_C_THROW("Condition {} has no node_set_name assigned!", id_);
```
amgebauer: I personally find this more readable, but is only minor.
```suggestion
  FOUR_C_ASSERT_ALWAYS(node_set_name_, "Condition {} has no node_set_name assigned!", id_);
  return *node_set_name_;
```

src/core/fem/src/condition/4C_fem_condition.hpp
amgebauer: Can we pass the (optional) name directly here? I think it is clearer instead of adding an additional setter. Then we can also directly throw if we don't set a name but the entity type requires it.

src/core/fem/src/condition/4C_fem_condition_definition.cpp
rjoussen: I left it as is for now, since enum input specs cannot be optional at the moment (the validator does not work).

src/core/fem/src/condition/4C_fem_condition_definition.cpp
rjoussen: I also think that in any case, there should be a primary identifier for node sets, either int IDs or string names.

rjoussen:
@davidrudlstorfer I ran the fourcipp testsuite with the new schema and metadata file, and it seems to run fine, except for some skipped tests:
```bash
485 passed, 7 skipped, 1 deselected, 3 warnings in 3.52s
```

davidrudlstorfer:
@rjoussen thanks a lot for checking that 🙏🏻 the skipped tests are only executed within the 4C docker container, so that is perfectly fine :)

amgebauer (approved):
Thank you @rjoussen

rjoussen:
The CMake test fixtures were slightly wrong.

rjoussen:
> However, this would also require changing some things in the actual tutorial text, so I will remove that again before merging. 

I would actually keep these changes now, since the tutorial does not address this anyway and this way this is also tested for exodus files. I will include an explanation of this in a future gmsh input tutorial.

src/core/fem/src/condition/4C_fem_condition_definition.cpp
sebproell: > I left it as is for now, since enum input specs cannot be optional at the moment (the validator does not work).


@rjoussen, could you please clarify? I just tested it, and it is definitely possible to have a `parameter<std::optional<EnumType>>` with an `in_set` validator. If there is a bug, please open an issue and assign it to me.

rjoussen: @sebproell I created a separate issue for this. As I said, I might not be aware of how to use the input spec correctly for this use case. E.g., I was not aware of the `in_set` validator. If optional enum types work already at the moment or its easy to make them work, I am happy to include that in this PR still.

rjoussen:
Following #1720, the `ENTITY_TYPE` is now optional and does not need to be specified if identification via `node_set_name` is chosen (it must not be specified to be precise).

This means that the entity type needs to be determined at runtime in addition to the internal ID. Especially, just providing E: ID with no `ENTITY_TYPE` specified must still result in `legacy_id` for backwards compatibility, while just `NODE_SET_NAME` and no `ENTITY_TYPE` is parsed as `EntityType::node_set_name`. 

While this certainly makes the input file look nicer and less redundant (see PR description), the new `parse_entity()` is quite ugly.
I did not really see how to avoid that, as we need to resolve the internal ID since it is expected by global_data_read. I think in the future, it would make sense to move the whole geometry resolution (adding nodes to the conditions, [here](https://github.com/4C-multiphysics/4C/blob/main/src/global_data/4C_global_data_read.cpp#L1594-L1680)) into the conditions themselves, since they know the entity type they are and so on. This way, conditions would not need to carry an internal ID anymore (it is barely used anyway), which is a prerequisite for `point_sets` not to have an internal id necessarily anymore (as pointed out by @amgebauer).

It is just an additional commit for now, I will squash them before merging, but this way its easier to revert :)

src/core/io/src/4C_io_input_spec_validators.hpp
```diff
@@ -540,10 +540,9 @@ template <typename T>
 
     void describe(std::ostream& os) const
     {
-      using namespace EnumTools;
       std::ostringstream set_description;
       set_description << "in_set{";
-      for (const auto& val : set) set_description << val << ",";
+      for (const auto& val : set) set_description << EnumTools::enum_name(val) << ",";
```
rjoussen, line 545: @sebproell I changed this to make clang-tidy happy (it was failing, see [here](https://github.com/4C-multiphysics/4C/actions/runs/21760083659/job/62781079394)), but I don't really know why, to be honest. I could imagine that it's due to #1703 though.

ischeider:
Regarding the idea for the future to put the node set into the boundary condition, I must admit that I don't like this idea too much. I usually create a structure and generate many node sets which I intend to use only partly in boundary conditions for several load cases. so I would like to have the node sets separate from the conditions and choose the use of the sets from case to case.

rjoussen:
@ischeider That's not what I meant, I was thinking of a change purely in the implementation. Currently, the resolution of which nodes belong to a certain nodeset happens in read_global_data, whereas it would make more sense if this happened in the code for the conditions (since they know which kind of node set they are referring to). Your workflow would not be affected by this at all.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
```diff
@@ -112,15 +112,20 @@ int Solid::MonitorDbc::get_unique_id(int tagged_id, Core::Conditions::GeometryTy
 void Solid::MonitorDbc::create_reaction_force_condition(
     const Core::Conditions::Condition& tagged_cond, Core::FE::Discretization& discret) const
 {
-  const int new_id = get_unique_id(tagged_cond.id(), tagged_cond.g_type());
+  const std::string rcond_name =
+      (tagged_cond.entity_type() == Core::Conditions::EntityType::node_set_name)
```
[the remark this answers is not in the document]
rjoussen: The reaction conditions use to write dbc reaction data are now designed to always have a node set name. This is easier to handle later in the code with the file names. This wil break only if the first simulation was done with an older version than introduced by this commit.

rjoussen:
I squashed some commits and rebased. 

In order to make the condition definition less convoluted due to the added possibility of specifiy a node set by its name, I introduced the following changes:

- `ConditionDefinition::read()` is now only **reading** the input file and returns a vector of `ConditionSpec`, a struct containing all information to actually create a condition. While reading/parsing, it is ensured that the right combination of `EntityType` / `ID` / `NODE_SET_NAME` is given, and that the IDs are valid. The reason why this function is not creating conditions yet (as it was before) is that a) `NODE_SET_NAME` conditions require knowledge about the mesh to resolve the internal id (which we still need unfortunately), and b) in my opinion, it doesnt really makes sense to have Condition creation and adding nodes that far apart from each other (as it was done before)
- These condition specifications are then turned into actual `Conditions` using the new factory method `condition_from_input` which now only takes the input for a single condition via `ConditionSpec` and all collected node sets from the input (both legacy and external). This method hence combines `Condition` creation **and** setting the nodes, which I think makes more sense. This also enables resolving the internal id of a node set /condition identified by its name more smoothly, since all information is available in one place (at condition creation).

Now (again) ready for review I think.

src/core/fem/src/condition/4C_fem_condition.cpp
```diff
@@ -54,5 +70,150 @@ std::shared_ptr<Core::Conditions::Condition> Core::Conditions::Condition::copy_w
   return copy;
 }
 
+std::string Core::Conditions::Condition::node_set_name() const
```
amgebauer:
```suggestion
const std::string& Core::Conditions::Condition::node_set_name() const
```

src/core/fem/src/condition/4C_fem_condition.cpp
```diff
@@ -54,5 +70,150 @@ std::shared_ptr<Core::Conditions::Condition> Core::Conditions::Condition::copy_w
...
+{
+  FOUR_C_ASSERT(entity_type_ == EntityType::node_set_name && node_set_name_.has_value(),
+      "Condition of ENTITY_TYPE 'node_set_name' has no node_set_name assigned!");
+  return node_set_name_.value();
+}
+
+std::shared_ptr<Core::Conditions::Condition> Core::Conditions::condition_from_input(
```
amgebauer:
```suggestion
std::unique_ptr<Core::Conditions::Condition> Core::Conditions::make_condition(
```

amgebauer (commented):
I like this new approach! In a second step, we could set the nodes (and maybe even surface elements) directly in the condition constructor.

rjoussen:
> In a second step, we could set the nodes (and maybe even surface elements) directly in the condition constructor.

I agree, my plan was to have construction, `set_nodes` and setting `parameters()` all in the same place, such that all of this can be moved to the constructor in the future. Turns out this was easier than I thought, so I added it to this PR. Setting the surface elements directly is more complicated.

src/core/fem/src/condition/4C_fem_condition.hpp
```diff
@@ -107,7 +113,11 @@ namespace Core::Conditions
     /*!
     \brief Return condition id
     */
-    [[nodiscard]] inline int id() const { return id_; }
+    [[nodiscard]] int id() const { return id_; }
+
+    [[nodiscard]] const std::string& node_set_name() const;
+
+    [[nodiscard]] bool has_node_set_name() const { return node_set_name_.has_value(); }
```
ischeider: You define the function `has_node_set_name()`, but as far as I can see you only use node_set_name.has_value() in the code. But may be I missed some usage.

rjoussen: You're right, I removed it.

davidrudlstorfer:
@rjoussen unfortunately this PR broke FourCIPP as far as I can see (=> https://github.com/4C-multiphysics/fourcipp/actions/runs/22704792990)

Could you please take a look into that so we get up and running soon again :)

PR #1501, opened 2025-11-14
Add rotational coupling test for beaminteraction with Lagrange Multipliers

dharinib98:
## Description and Context

This PR adds a test checking the implementation of Lagrange Multipliers for beaminteraction problems (https://github.com/4C-multiphysics/4C/pull/637 ) with rotational coupling enabled. 

## Related Issues and Pull Requests

#1206

PR #1691, opened 2026-01-29
Beam to beam point couplings with Lagrange multipliers

isteinbrecher:
## Description and Context

Allow to enforce beam point couplings with Lagrange multipliers.


src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
```diff
@@ -431,14 +431,23 @@ void BeamInteraction::BeamToSolidMortarManager::evaluate_force_stiff_penalty_reg
...
   // For now we need to set a link to the global Lagrange multiplier vector in the beam interaction
   // data state.
   global_lambda_ = data_state->get_lambda();
+
+  // Add the force and stiffness contributions that are assembled directly by the pairs.
+  Core::LinAlg::Vector<double> lambda_col(*lambda_dof_colmap_);
+  Core::LinAlg::export_to(*global_lambda_, lambda_col);
+  for (const auto& elepairptr : contact_pairs_)
+    elepairptr->evaluate_and_assemble(
+        *discret_, this, force, stiff, lambda_col, *data_state->get_dis_col_np());
```
isteinbrecher, line 450: @dharinib98 this should fix the slow convergence in #1501

dharinib98: Thanks! yes, it does improve the convergence. We require only 5 iterations now as opposed to 15. However I am not sure if I fully understand how this helps. Is it not more like an extra feature ( the beam to beam coupling)? 

Also general question regarding  this PR. Does this mean that until now some contact pairs were being evaluated with the wrong lambda values?

isteinbrecher: > However I am not sure if I fully understand how this helps.
> Does this mean that until now some contact pairs were being evaluated with the wrong lambda values?

This is simply an assembly of coupling terms that was missing previously - only for the Lagrange formulation, they were always there for penalty. So for certain problems, we had a slightly wrong stiffness matrix in the Lagrange multiplier case.

davidrudlstorfer:
@isteinbrecher out of curiosity, when do you plan on merging this? I would like to move the contact stuff this afternoon. But this would require a lot of rebasing for either of these two PR's so it would make sense that I wait till this is merged.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
```diff
@@ -780,6 +789,23 @@ void BeamInteraction::BeamToSolidMortarManager::assemble_stiff(
   auto block_lm_displ_row_map = jac_block_sparse_matrix_base->matrix(1, 0).row_map();
   auto block_displ_lm_row_map = jac_block_sparse_matrix_base->matrix(0, 1).row_map();
 
+  Core::LinAlg::SparseMatrix lm_lm =
+      Core::LinAlg::SparseMatrix(block_lm_displ_row_map, 81, true, true);
+
+  auto lambda_non_active = Core::LinAlg::Vector<double>(lambda_active_->get_map());
+  for (int lid = 0; lid < lambda_active_->get_map().num_my_elements(); lid++)
+  {
+    if (lambda_active_->get_values()[lid] < 0.1)
```
mayrmt, line 798: This `0.1` sounds like an arbitrary constant. Where is it coming from?

src/inpar/4C_inpar_beaminteraction.cpp
```diff
@@ -240,10 +240,7 @@ void Inpar::BeamInteraction::set_valid_conditions(
                     .default_value = 0}),
             parameter<::FourC::BeamInteraction::BeamToBeamPointCouplingPairParameters::
                     ConstraintEnforcement>("CONSTRAINT_ENFORCEMENT",
-                {.description = "How the constraints for this condition shall be enforced",
-                    .default_value =
-                        ::FourC::BeamInteraction::BeamToBeamPointCouplingPairParameters::
-                            ConstraintEnforcement::penalty_direct})},
+                {.description = "How the constraints for this condition shall be enforced"})},
```
mayrmt:
```suggestion
                {.description = "Specifies the constraint enforcement technique for beam-to-beam point couplings"})},
```

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.hpp
```diff
@@ -56,7 +56,8 @@ namespace BeamInteraction
     enum class ConstraintEnforcement
     {
       penalty_direct,
-      penalty_indirect
+      penalty_indirect,
+      lagrange_multiplier
     };
```
mayrmt, line 62: Can you document the differences between those options, esp. for `penalty_direct` vs. `penalty_indirect`?

isteinbrecher:
> @isteinbrecher out of curiosity, when do you plan on merging this? I would like to move the contact stuff this afternoon. But this would require a lot of rebasing for either of these two PR's so it would make sense that I wait till this is merged.

I think it makes sense to wait, I will try to merge this soon.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
isteinbrecher: Since we can't store boolean values in vectors, we use a little hack here, we add `1.0` to this vector if an entry is active. The `0.1` here are simply a save choice of checking that a value is set.

maxfirmbach: We should use `Core::LinAlg::Vector<int> toggle` here.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.hpp
dharinib98: Just another small detail here. Can we call the `penalty_indirect` as `lagrange_regularized` and `lagrange_multiplier` as `lagrange_saddlepoint` ? Since in the input file parameters , we would be specifying it like this.

src/beaminteraction/src/4C_beaminteraction_beam_to_solid_mortar_manager.cpp
mayrmt: I see.

Does the `lambda_active_` need to be an `FEVector`? If it can be a regular `Vector`, then you could template it on `int`, which at least is somehow closer to a `bool` than a `double`. Templating on `bool` is not possible, since there are no `bool` backends in Epetra.

isteinbrecher:
> We should use Core::LinAlg::Vector<int> toggle here.

@maxfirmbach does this also work for FE vectors? I remember when this construct was initially introduced, there was no FE vector for `int` data types. Anyway, this is beyond the scope of the current PR.

maxfirmbach:
@isteinbrecher No, there can only be `FEVector<double>` with `Epetra` right now.

src/beaminteraction/src/4C_beaminteraction_beam_to_beam_point_coupling_pair.hpp
isteinbrecher: No, that would not work. First, the input file parameters are not the same as for beam-to-solid coupling. Second, `penalty_indirect` is not the same as `lagrange_regularized`. In beam-to-solid words `penalty_indirect` would be the plain `penalty` case, but that naming would not work for the point couplings.

mayrmt:
@isteinbrecher Does `lambda_active_` need to be an FEVector? If it can be a regular Vector, then you could template it on `int`, which at least is somehow closer to a `bool` than a `double`.

isteinbrecher:
@maxfirmbach @mayrmt Yes, it needs to be an FEVector, so I think until we really support Tpetra, this will have to stay as it is.

isteinbrecher:
@davidrudlstorfer There are no planned PRs in the near future in beam contact, so feel free to restructure the folder.

davidrudlstorfer:
@isteinbrecher thanks for the ping, I will start to restructure that today:)

PR #1694, opened 2026-01-29
refactor: move crosslinking into own folder

davidrudlstorfer:
## Description and Context

Directly follows #1693 and moves all crosslinking files into its own dedicated subfolder. Also removed an unnecessary dependence in potential and spherebeamlinking by adding the correct binning header.

- [x] This PR is based on #1693, once merged I will rebase.

### DIFF
--- a/src/beaminteraction/CMakeLists.txt
+++ b/src/beaminteraction/CMakeLists.txt
@@ -10,13 +10,15 @@ four_c_auto_define_module()
 set(_dependencies
     # cmake-format: sortable
     beam3
-    config
+    contact
     core
+    coupling
     geometry_pair
     global_data
     inpar
     mat
     rigidsphere
+    shell_kl_nurbs
     solid_3D_ele
     structure
     structure_new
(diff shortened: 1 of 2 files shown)


davidrudlstorfer:
@isteinbrecher the next and last step in the rough reorganization of the beaminteraction folder would be to move all beam contact related stuff (via the submodel evaluator contact) into its own `contact` subfolder.

Then I would be finally able to split up the inpar files into the "submodules" of beaminteraction. Is this fine for you or is still aligned with the plan we made some time ago?

davidrudlstorfer:
@isteinbrecher then we could also split up a few files that would remain in the top level beaminteraction folder such as `4C_beaminteraction_data.hpp/cpp`

isteinbrecher:
> @isteinbrecher the next and last step in the rough reorganization of the beaminteraction folder would be to move all beam contact related stuff (via the submodel evaluator contact) into its own contact subfolder.
>Then I would be finally able to split up the inpar files into the "submodules" of beaminteraction. Is this fine for you or is still aligned with the plan we made some time ago?

Yes, that sounds good. So we would move beam-to-beam contact and beam-to-solid-xxx into the contact folder?

davidrudlstorfer:
> Yes, that sounds good. So we would move beam-to-beam contact and beam-to-solid-xxx into the contact folder?

Yes, exactly. The input from the old contact framework is already in there.

PR #1695, opened 2026-01-30
Move next inpar files to submodules

davidrudlstorfer:
## Description and Context

After some time finally a few inpar moves again :partying_face: 

One question @amgebauer: Do you use the `cardiovascular0D` module? I renamed the Inpar::Cardiovascular0D namespace to `Cardiovascular0DInput`. Like in the other modules I would have just preferred `Cardiovascular0D` but this clashes with the same class name and this would require a lot of restructuring and rework to get it done. Let me know if you use it or if this is a good solution if no one uses it anyways.

Part of #59


davidrudlstorfer:
(for review: the commits are split into the 4 inpar files I moved, I think this should make it easier to review)

Issue #1677, opened 2026-01-26
[BUG] How does `BinningStrategy::extend_element_col_map()` work?


vovannikov:
## Bug Description: What's going wrong?
I have encountered an issue when working with the benchmarks of @slfuchs with the recent version of 4C. I am still collecting the data to properly describe the issue, but what happens is that 4C crashes in MPI case with rank > 1 with the following error:

```
PROC 1 ERROR in /home/vladimir/development/4C/src/core/fem/src/discretization/4C_fem_discretization_partition.cpp, line 247:
Proc 1: Element gid=93 from oldmap is not in newmap
------------------
 0# void FourC::Core::Internal::format_and_throw_error<int const&, int&>(std::source_location const&, std::basic_format_string<char, std::type_identity<int const&>::type, std::type_identity<int&>::type>, int const&, int&) in /home/vladimir/development/4C-debug/lib4C.so
 1# FourC::Core::FE::Discretization::export_column_elements(FourC::Core::LinAlg::Map const&, bool, bool) in /home/vladimir/development/4C-debug/lib4C.so
 2# FourC::Core::Binstrategy::Utils::extend_discretization_ghosting(FourC::Core::FE::Discretization&, FourC::Core::LinAlg::Map&, bool, bool, bool) in /home/vladimir/development/4C-debug/lib4C.so
 3# FourC::Particle::WallHandlerDiscretCondition::extend_wall_element_ghosting(std::map<int, std::set<int, std::less<int>, std::allocator<int> >, std::less<int>, std::allocator<std::pair<int const, std::set<int, std::less<int>, std::allocator<int> > > > >&) in /home/vladimir/development/4C-debug/lib4C.so
 4# FourC::Particle::WallHandlerDiscretCondition::distribute_wall_elements_and_nodes() in /home/vladimir/development/4C-debug/lib4C.so
 5# FourC::Particle::ParticleAlgorithm::distribute_load_among_procs() in /home/vladimir/development/4C-debug/lib4C.so
 6# FourC::Particle::ParticleAlgorithm::setup() in /home/vladimir/development/4C-debug/lib4C.so
 7# FourC::PaSI::PartitionedAlgo::setup() in /home/vladimir/development/4C-debug/lib4C.so
 8# FourC::PaSI::PasiPartTwoWayCoup::setup() in /home/vladimir/development/4C-debug/lib4C.so
 9# FourC::pasi_dyn() in /home/vladimir/development/4C-debug/lib4C.so
[...]

PR #1696, opened 2026-01-30
Provide standard column map to `extend_wall_element_ghosting()`

vovannikov:
## Description and Context
It seems the lack of a standard column map caused the problem had been reported in #1677. It is though difficult to say why previously this error was not triggered. The issue was identified and fixed together with @georghammerl.

@ppraegla, @slfuchs 

#1677

### DIFF
--- a/src/core/binstrategy/4C_binstrategy.hpp
+++ b/src/core/binstrategy/4C_binstrategy.hpp
@@ -605,11 +605,11 @@ namespace Core::Binstrategy
      * \return extended element column map
      */
     std::shared_ptr<Core::LinAlg::Map> extend_element_col_map(
-        std::map<int, std::set<int>> const& bin_to_row_ele_map,
-        std::map<int, std::set<int>>& bin_to_row_ele_map_to_lookup_requests,
+        const std::map<int, std::set<int>>& bin_to_row_ele_map,
+        const std::map<int, std::set<int>>& bin_to_row_ele_map_to_lookup_requests,
         std::map<int, std::set<int>>& ext_bin_to_ele_map,
-        std::shared_ptr<Core::LinAlg::Map> bin_colmap = nullptr,
-        std::shared_ptr<Core::LinAlg::Map> bin_rowmap = nullptr,
+        std::shared_ptr<const Core::LinAlg::Map> bin_colmap = nullptr,
+        std::shared_ptr<const Core::LinAlg::Map> bin_rowmap = nullptr,
         const Core::LinAlg::Map* ele_colmap_from_standardghosting = nullptr) const;
 
     /*!
--- a/src/particle/src/wall/4C_particle_wall.cpp
+++ b/src/particle/src/wall/4C_particle_wall.cpp
@@ -486,7 +486,7 @@ void Particle::WallHandlerDiscretCondition::distribute_wall_elements_and_nodes()
       *walldiscretization_, walldiscretization_->my_row_element_range(), bintorowelemap, disn_col);
 
   // extend wall element ghosting
-  extend_wall_element_ghosting(bintorowelemap);
+  extend_wall_element_ghosting(bintorowelemap, stdelecolmap);
 
   // update maps of state vectors
   walldatastate_->update_maps_of_state_vectors();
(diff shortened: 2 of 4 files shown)


src/particle/src/wall/4C_particle_wall.hpp
```diff
@@ -322,8 +322,10 @@ namespace Particle
      *
      *
      * \param[in] bintorowelemap bin to row wall element distribution
+     * \param[in] stdelecolmap standard element column map
      */
-    void extend_wall_element_ghosting(std::map<int, std::set<int>>& bintorowelemap);
+    void extend_wall_element_ghosting(const std::map<int, std::set<int>>& bintorowelemap,
+        std::shared_ptr<const Core::LinAlg::Map> stdelecolmap);
```
vovannikov, line 328: This is the most relevant part of the PR. The other changes are related to the fact that I added some more `const`s.

georghammerl (approved):

src/particle/src/wall/4C_particle_wall.cpp
```diff
@@ -510,18 +510,19 @@ void Particle::WallHandlerDiscretCondition::transfer_wall_elements_and_nodes()
...
 
 void Particle::WallHandlerDiscretCondition::extend_wall_element_ghosting(
-    std::map<int, std::set<int>>& bintorowelemap)
+    const std::map<int, std::set<int>>& bintorowelemap,
+    std::shared_ptr<const Core::LinAlg::Map> stdelecolmap)
 {
   std::map<int, std::set<int>> colbintoelemap;
   std::shared_ptr<Core::LinAlg::Map> extendedelecolmap = binstrategy_->extend_element_col_map(
-      bintorowelemap, bintorowelemap, colbintoelemap, bincolmap_);
+      bintorowelemap, bintorowelemap, colbintoelemap, bincolmap_, nullptr, stdelecolmap.get());
```
ppraegla, line 525: Why do you pass the stdelecolmap as the rowmap argument and not as the colmap argument?

slfuchs: No that should be fine, the last argument of the method is now:
```const Core::LinAlg::Map* ele_colmap_from_standardghosting = nullptr```

File: src/core/binstrategy/4C_binstrategy.hpp [hunk 3]
```diff
@@ -605,11 +605,11 @@ namespace Core::Binstrategy
...
     std::shared_ptr<Core::LinAlg::Map> extend_element_col_map(
-        std::map<int, std::set<int>> const& bin_to_row_ele_map,
-        std::map<int, std::set<int>>& bin_to_row_ele_map_to_lookup_requests,
+        const std::map<int, std::set<int>>& bin_to_row_ele_map,
+        const std::map<int, std::set<int>>& bin_to_row_ele_map_to_lookup_requests,
         std::map<int, std::set<int>>& ext_bin_to_ele_map,
-        std::shared_ptr<Core::LinAlg::Map> bin_colmap = nullptr,
-        std::shared_ptr<Core::LinAlg::Map> bin_rowmap = nullptr,
+        std::shared_ptr<const Core::LinAlg::Map> bin_colmap = nullptr,
+        std::shared_ptr<const Core::LinAlg::Map> bin_rowmap = nullptr,
```
slfuchs, line 612: @ppraegla question below made me look closer at this function again. I think one could change the order of `bin_colmap` and `bin_rowmap` as function arguments here to follow the 4C standards where mostly row comes for col. However, I haven't checked if this requires additional files to be touched since I don't know where `extend_element_col_map` is also called.

src/particle/src/wall/4C_particle_wall.cpp
ppraegla: You are right. I overlooked the last argument.

src/core/binstrategy/4C_binstrategy.hpp
vovannikov: That is an interesting observation. I think it indeed makes sense to swap these two agruments. However, I would prefer to do that in a follow up PR, since this change requires modification of a few more lines of code, not related to the issue targeted by this PR. Just to keep this PR focused and clear.

slfuchs: I would also do this in a follow up PR.

PR #1698, opened 2026-01-30
Move beam contact

davidrudlstorfer:
## Description and Context

This PR now restructures the major part of beaminteraction (specifically contact)

The commits are separated as follows:

- The first commit moves all contact related files into the contact subfolder
- The second commit moves the beaminteraction unit tests to the beaminteraction subfolder (according to #81)
- The third commit further restructures `beaminteraction/contact` into `beam_to_beam`, `beam_to_solid` => `edge` & `surface` & `volume` and `beam_to_sphere`. @isteinbrecher this commit is more or less optional for me but I think the added structure and improved overview is still worth it.

This would be the necessary step to move (and more importantly split up) the inpar beaminteraction files.

Follow up from #1693 and #1694 
Indirect part of #59 
Part of #81


davidrudlstorfer:
Note: it would now be possible to further clean-up/separate files in `beaminteraction/` and `beaminteraction/contact` into the corresponding "sub"modules. But this is not really in my interest right now and not super important.

maxfirmbach:
@davidrudlstorfer Thanks for cleaning this up so nicely :+1:.

PR #1701, opened 2026-01-30
Fix initial output of Gauss point data

amgebauer:
We evaluated the Gauss point data for the column map, but we only need the row map. This caused an error since in the initial setup, the element is only evaluated once for the column map (for the stress output), so the internal state is not set for ghosted elements.

fyi @tuchpaul


tuchpaul (approved):
Nice! Thanks for finding this "bug".

maxfirmbach (approved):

PR #1702, opened 2026-01-30
Correct comments on poroelastic preconditioner

shervas:
Correct names of the field preconditioners for the poroelastic approach.

## Description and Context

The block preconditioners in porofluid_pressure_based can be categorised according to the number of fields being solved for in the problem. In elast_scatra_BSG-AMG_3x3, three fields are solved in the following order: structure, fluid, and scalar transport (scatra).
In this case, artery pressure is not solved. There was a lapse in the comments, but the settings are correct.


mayrmt:
@shervas This PR adds an xml file, though I do not see, where it is used in the tests. Is the something still missing?

shervas:
@mayrmt sorry, my mistake. I was working on preconditioners for a problem without arteries but in the test files those are not solved with a preconditioner. I'm closing this PR then.
I'll work on a test file for than an open the PR again in the future.

Thanks!

PR #1703, opened 2026-01-30
Implement proper type erasure for Validators

sebproell:
`std::function` (also a type-erasure implementation) is awesome for prototyping but quite a heavy object. This PR replaces the triple `std::function` usage within the `InputSpec::Validators` with a manual type-erasure implementation.


davidrudlstorfer (approved):
🚀

PR #1713, opened 2026-02-02
Fix and test last_possible restart in scatra

maxiludwig:
# Description and Context

- in scatra problems the CLI argument `--restart=last_step` did not work properly. The simulation still started from step 1
- to use the last restart step by the CLI argument `--restart=last_step` the `field` needs to be written in the output control file
- this was not done in scatra problems so far
- writing mesh before new step fixes it
- I also added the possibility to test if the correct restart step is used by reading the control file after a test is finished
- The `last_possible` option is so far only tested for scatra and structure

Thanks to @lkoeglmeier @c-p-schmidt


c-p-schmidt (dismissed):

maxiludwig:
I thought of how to test whether the simulation actually starts from the last step. Simply adding a new input file and changing the number of steps is not sufficient, as a simulation starting from step 1 (the behavior before the fix) would have the same result tests as a simulation starting from the last step. 
A possible solution would be to change the (time-dependent) functions to alter the results for tests that use the output from a restart. It would be cumbersome to do this for all problem types though. Do you have a better idea? @c-p-schmidt @lkoeglmeier

tests/input_files/scatra_3D_tet4_hex8_s2i_constperm_mortar_standard_restart.4C.yaml
```diff
@@ -0,0 +1,4988 @@
+TITLE:
+  - "Academic test case for scatra-scatra interface coupling:"
+  - "- three-dimensional problem with two-dimensional scatra-scatra interface"
+  - "- inner sphere and outer hollow sphere with constant initial concentrations"
+  - "- membrane with constant permeability in between both spheres"
+  - "- symmetry allows only a quarter of the entire domain to be considered"
+  - "- no fluxes through outer surface of outer hollow sphere"
+  - "- 2735 tet4 and hex8 elements with non-matching interface nodes"
```
c-p-schmidt, line 8: Maybe you could add in the description that this is a test modified from `tests/input_files/scatra_3D_tet4_hex8_s2i_constperm_mortar_standard.4C.yaml` to test the capability to restart from the last possible step?

tests/list_of_tests.cmake
```diff
@@ -496,7 +495,7 @@ four_c_test(TEST_FILE elch_1D_10ele_3ions_stab_fdcheck.4C.yaml NP 2)
 four_c_test(TEST_FILE elch_1D_5ele_line3_3ions_stab_genalpha_fdcheck.4C.yaml NP 2 LABELS minimal)
 four_c_test(TEST_FILE elch_1D_binaryNewman_twoEqu_divi_2iter.4C.yaml NP 3)
 four_c_test(TEST_FILE elch_1D_DLcap_linearKinetics_galvanostat.4C.yaml NP 3 RETURN_AS current)
-four_c_test_restart(BASED_ON ${current} SAME_FILE NP 3 RESTART_STEP 10)
+four_c_test_restart(BASED_ON ${current} SAME_FILE NP 3 RESTART_STEP 10 RETURN_AS current_restart)
```
c-p-schmidt: Why do you need this?

maxiludwig: I don't need it :rofl:

c-p-schmidt (approved):

Issue #210, opened 2025-01-08
Unify structure time integration


eulovi:
> This issue was originally created by @isteinbrecher in https://gitlab.lrz.de/baci/baci/-/issues/889 on 2022-11-14.

## Description

Currently, we have two implementations for the structure time integration.
In the past, this has proven to be an obstacle when developing new methods and combining them with existing ones, therefore, the two frameworks **have** to be unified.
To decide which one of the frameworks should be the "new" and only structural time integration, it is important to know what features of the old and new time integration are currently being used by the BACI developers.
Please add your use cases to the lists below.
If you are unsure which features you use, please ask a colleague for advice. It is very important to have an accurate picture of the current needs of all users.

**Features used in the old structural time integration:**
- Monolithic Fluid-Structure Interaction @NHagmeyer @mayrmt @maxfirmbach 
- Mortar Meshtying & Contact with penalty @mayrmt @eulovi @lengelhardt 
- Mortar Meshtying & Contact with Lagrange multipliers @mayrmt @eulovi @ChrSte @SchaefP1 @lengelhardt 
- Poroelasticity @frankma @SchaefP1
- FPSI @frankma 

**Features used in the new structural time integration:**
- Standard solid implementation @isteinbrecher @NHagmeyer @lkoeglmeier @agebauer @cschmidt @ssinzig @maxfirmbach @eulovi
- Beam framework (including Gen-alpha Lie group time integration) @isteinbrecher @NHagmeyer 
- Beam interaction framework @isteinbrecher @NHagmeyer @maxfirmbach 
- Partitioned Fluid-Structure interaction @NHagmeyer 
- (Partitioned) Fluid-Beam Interaction @NHagmeyer 
- Multi-Scale contact(only tested in new time integration) @RShaw 
- Runtime Output / Gauss point data output @agebauer @NHagmeyer @isteinbrecher @maxfirmbach @eulovi 
- nox @agebauer @NHagmeyer @maxfirmbach 
- material iterative prestressing @agebauer @cschmidt @ssinzig 
- mulf @agebauer @NHagmeyer 
- cardiovascular0d @agebauer 
- Mortar Contact with penalty @NHagmeyer 
- Truss implementation @cschmidt @ssinzig 
- (Partitioned) ScaTra-Structure(-Temperature) Interaction (SS(T)I) @cschmidt @ssinzig 
- Nitsche contact (for SSI) @cschmidt @ssinzig 

## Related Issues and Merge Requests

* Blocks
[...]

PR #1714, opened 2026-02-02
Fix spring dashpot cursurfnormal condition in standard integration strategy

lauraengelhardt:
## Description and Context
With this PR, the test `spring_dashpot_struct_bc_curnormal` now works in the new structural time integration. 
I also removed some unused parameters and function from the spring dashpot implementation.

@c-p-schmidt Let me know if you want to change the default for the integration strategy first. I can merge after if so. 

Part of https://github.com/4C-multiphysics/4C/issues/210


src/constraint/4C_constraint_springdashpot.cpp
```diff
@@ -477,7 +472,7 @@ Constraints::SpringDashpot::SpringDashpot(
   offset_prestr_new_->put_scalar(0.0);
 
   // set type of this spring
-  set_spring_type();
+  springtype_ = spring_->parameters().get<RobinSpringDashpotType>("DIRECTION");
```
c-p-schmidt, line 475: The variable's name looks a bit weird (springtype <-> DIRECTION), but it's been like that before. So I would be fine with keeping it if you do not have a better idea.

PR #1706, opened 2026-02-02
Rework solid post-processing tests

c-p-schmidt:
## Description and Context
In preparation of changing the default setting for the solid time integration to the new one as discussed [here](https://github.com/4C-multiphysics/4C/issues/210#issuecomment-3817610708) onwards, some post-processing tests are adapted. This includes
- simply remove them if functionality is already covered by a vtk test
- extent available or add vtk tests if quantities have been missing so far (add output of GL to solid_runtime_vtk_2d, implement output of the contact gap for the new time integration and add a test contact3D_pwlin_penalty_hex27) and remove the corresponding pp test

All changes, so also the upcoming to change the default behavior can already be seen on my fork if you‘re interested: https://github.com/c-p-schmidt/4C/pull/9

## Related Issues and Pull Requests
Part of #210

PR #1715, opened 2026-02-03
Change default solid time integration setting

c-p-schmidt:
## Description and Context
As discussed [here](https://github.com/4C-multiphysics/4C/issues/210#issuecomment-3817610708) onwards, the default of the solid time integration shall be set to the new solid time integration this is performed in two steps:
1. Change the default and explicitly add "INT_STRATEGY Old" where necessary
2. Remove all redundant "INT_STRATEGY Standard" as it is the default now.

part of #210 
follows #1706


maxfirmbach:
I highly support these changes 🙂.

PR #1716, opened 2026-02-03
Restructure and modularize beaminteraction input

davidrudlstorfer:
## Description and Context

This PR now finally starts to restructure and modularize the beam interaction input.

The initial big goal was to move the beam interaction input from the inpar folder to the physics submodule. This was not easily possible due to substantial clashes within the name space (enums and different classes had the same names within different namespaces)

The first commit resolves the largest name clash by renaming one central enum class. Afterwards, I removed the Inpar namespace which again showed substantial name clashes (I will add a comment down below).

Afterwards I refactored the whole input parameter setup within beam interaction. I now propose to centrally collect all beam interaction parameters within the top level of beaminteraction (rather than calling weird functions from the global call).

This is now a first draft of everything because I still need to check a few things (all tests pass locally). After this PR the same logic should be applied to the conditions of beam interaction.

#59


src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
davidrudlstorfer, line 1: In this file we have the problem that we now always need the `FourC::` in front because the name space resolution does not work due to the same name of the class ...

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.hpp
```diff
@@ -55,10 +55,12 @@ namespace Solid
     // forward declaration
     class BeamInteractionDataState;
 
+    // ! TODO this class has a nameclash / problem with the namespace BeamInteraction, how to solve
+    // this nicely? Rename this to something more senseful?
     class BeamInteraction : public Generic
```
davidrudlstorfer: ... with this class name. More or less it is not optimal in my opinion to name this class BeamInteraction if the namespace is also `BeamInteraction`.

Ideally, I would wish to rename this class. @isteinbrecher do you have ideas for a meaningful name?

src/beaminteraction/src/4C_beaminteraction_input.cpp
```diff
@@ -0,0 +1,143 @@
...
+  Core::IO::InputSpec beam_potential_specs = BeamInteraction::Potential::valid_parameters();
+  specs.push_back(beam_potential_specs);
+
+  // get sphere beam linking parameters
+  std::vector<Core::IO::InputSpec> spherebeamlinking_specs =
+      BeamInteraction::valid_parameters_spherebeamlinking();
+  specs.insert(specs.end(), spherebeamlinking_specs.begin(), spherebeamlinking_specs.end());
+
+  return specs;
+}
```
davidrudlstorfer, line 65: The main change and improvement is now in this file / function here: we now collect all the parameters from the "submodules" centrally in once place ...

src/global_legacy_module/4C_global_legacy_module_validparameters.cpp
```diff
@@ -312,10 +312,7 @@ std::vector<Core::IO::InputSpec> Global::valid_parameters()
   push_specs(specs, Inpar::IORuntimeOutput::FLUID::valid_parameters());
   push_specs(specs, Inpar::IORuntimeOutput::Solid::valid_parameters());
   push_specs(specs, Beam::IORuntimeOutput::valid_parameters());
-  push_specs(specs, BeamInteraction::Contact::BeamToBeam::valid_parameters());
-  push_specs(specs, BeamInteraction::Potential::valid_parameters());
-  push_specs(specs, BeamInteraction::valid_beam_to_edge_contact_parameters());
-  push_specs(specs, Inpar::BeamInteraction::valid_parameters());
+  push_specs(specs, BeamInteraction::valid_parameters());
```
davidrudlstorfer, line 312: ... and here we now only get the beam interaction parameters from one central space rather than all submodules in a mixed way. The same logic should be applied to the beam interaction conditions in another PR.

src/beaminteraction/src/4C_beaminteraction_input.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+FOUR_C_NAMESPACE_OPEN
+
+// forward declaration
+namespace Core::Conditions
+{
+  class ConditionDefinition;
+}
+
+namespace BeamInteraction
```
davidrudlstorfer, line 28: The second ToDo (besides the `BeamInteraction` name clash) would be to move a lot of these enums to the files where they are actually needed. But I would also prefer to do this in a separate PR

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.hpp
isteinbrecher: We could go for `BeamInteractionModelEvaluator`.

davidrudlstorfer: I renamed that and cleaned up all the unnecessary FourC:: namespace lookups in beaminteraction. We now have 0 of them left! :partying_face:

src/beaminteraction/src/4C_beaminteraction_input.hpp
davidrudlstorfer: And I've also already done this in the current PR :)

davidrudlstorfer:
This would be ready for review :)

davidrudlstorfer:
(and I am a little bit undecided if I should squash all commits after the review; the executable is buildable with every commit and all tests pass with every commit; if we want to revert some change in the future because it's to spread out we can easily do that on a single commit basis)

PR #1683, opened 2026-01-27
Implement 1D pipeflow for pulmonary system

engstlerL:
## Description and Context
This PR adds a new module to model fluid flow in the pulmonary circulation using a 1D formulation following Formaggia et al. ([Cardiovascular Mathematics](https://doi.org/10.1007/978-88-470-1152-6)). 
The 1D formulation enables spatially resolved blood flow simulations in the pulmonary circulation and allows direct comparison with existing 0D airway models, which cannot capture the characteristic wave propagation effects.

The PR includes the creation of an input file for specifying the relevant parameters, a file implementing the basic model for fluid flow in a continuous tube, and an end-to-end test. The implementation has been verified against results in literature regarding the wave propagation.  If this is considered too many changes for a single contribution, I could split it into smaller steps.

The current implementation supports a single continuous vessel; bifurcations, discontinuities, and capillary beds will be added in future PRs.

PR #1717, opened 2026-02-03
Add junction implementation in 1D pipeflow

engstlerL:
## Description and Context
This PR introduces the handling of discontinuities and junctions in the 1D pipe flow model.
It includes the identification of junctions in the geometry and the computation of the corresponding internal boundary conditions.

The implementation has been verified against reference results from the literature concerning wave reflection and transmission at junctions.

Additionally, two test cases are introduced:

- A continuous tube defined via a discontinuous geometry
- A simple three-node junction example

This PR follows #1683 and generalizes the previously implemented continuous tube approach by introducing junctions and discontinuities.


src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_main.cpp
```diff
@@ -147,6 +148,275 @@ namespace ReducedLung1dPipeFlow
...
+        junction_normal[i] = normals_for_junction.get_values()[local_node_id];
+
+        FOUR_C_ASSERT(junction_normal[i] == -1.0 || junction_normal[i] == 1.0,
+            "Junction normals need to be in/ outlet");
+
+        junction_ref_area_A0[i] = area0_for_junction.get_values()[local_node_id];
+        junction_beta[i] = beta_for_junction.get_values()[local_node_id];
+      }
+
+      /**************************************************************/
```
maxfirmbach, line 285: I recommend using the local Newton solver in `4C_utils_local_newton.hpp` and adapting the input/output interface to match the existing functionality. @engstlerL If there is functionality that is still missing in the Newton method feel free to reach out so we can add it.

For context: We should try to avoid adding `Teuchos` headers and functionality as much as possible.

sebproell: While I think I know the answer to this comment, could you please comment on this @engstlerL? It can be useful to have these comments available later on.

src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_main.cpp
```diff
@@ -147,6 +148,275 @@ namespace ReducedLung1dPipeFlow
...
+        constexpr double RTOL = 1e-6;
+        w(i) = 1.0 / (RTOL * std::abs(x(i)) + ATOL(i));
+      }
+
+      // derive residual vector
+      Teuchos::SerialDenseVector<int, double> f(2 * N_connected_nodes);
+      compute_residual(f, N_connected_nodes, x, junction_normal, junction_ref_area_A0,
+          junction_characteristic_out, junction_beta, density_rho);
+
+      auto compute_wrms = [&](const Teuchos::SerialDenseVector<int, double>& vec) -> double
```
sebproell:
```suggestion
const auto compute_wrms = [&](const Teuchos::SerialDenseVector<int, double>& vec)
```

minor

src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_main.cpp
```diff
@@ -261,10 +649,29 @@ namespace ReducedLung1dPipeFlow
...
+        // check if junction or outer boundary
         if (normal_in_out[inode] != 0)
         {
-          is_boundary[inode] = true;
+          bool found_in_junction = false;
+          if (auto normal_junction_id = normals_for_junctions.get_map().lid(global_node_id);
+              normal_junction_id != -1)
+          {
+            found_in_junction = true;
+          }
```
sebproell:
```suggestion
          const bool found_in_junction = normals_for_junctions.get_map().lid(global_node_id) != -1;
```

src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_main.cpp
engstlerL: Since the vectors are dynamic, `Core::LinAlg::SerialDenseMatrix` would need to be used instead of `Teuchos::SerialDenseVector`. However, there is currently no solver implemented for that case.
Therefore, `Teuchos` is kept for the moment, with the intention to switch to local_newton in the long term.

maxfirmbach: We had a short discussion offline, fine for me :smiley:.

sebproell: @maxfirmbach I thought so, but as I said, it does not hurt to write down the result.

Issue #1708, opened 2026-02-02
Edge cases for enums in InputSpec


rjoussen:
## TODOs

- [x] Fix broken assignment: e.g. may not assign `Validator<int>` to `Validator<optional<int>>` (#1719)
- [x] Fix enum choices and descriptions not being fully compatible
- [ ] Fix enum inside vector (write out choices to metadata in this case)
- [x] Fix optional enum metadata emit (#1720)

sebproell:
With #1720, I was able to make `ENTITY_TYPE` optional by modifying this:

https://github.com/4C-multiphysics/4C/blob/78d81cf0e067aa2d3a389f726e901b98309b0414/utilities/four_c_python/src/four_c_metadata/metadata.py#L409-L416

with an additional

```python
        if noneable:
            choices = list(choices) + [None] 
```

before line 410.

@rjoussen @gilrrei I leave it to you whether this is the correct fix on the Python side, or if you want to handle it differently.

rjoussen:
>         if noneable:
>             choices = list(choices) + [None] 

Looks good to me, but shouldn't that be part of #1720 maybe? Otherwise optional enums as introduced by #1720 can only be used if the default value is not null_opt/None.

gilrrei:
@sebproell @rjoussen sounds good :) Just as a reminder, we should also change the typehinting

https://github.com/4C-multiphysics/4C/blob/78d81cf0e067aa2d3a389f726e901b98309b0414/utilities/four_c_python/src/four_c_metadata/metadata.py#L389

to
```python
        choices: Sequence[str | None],
```

PR #1720, opened 2026-02-03
Emit optional enum choices to metadata

sebproell:
Part of #1708 

Not the prettiest fix, as it further circumvents inconsistencies rather than fixing them, but it should at least help with the immediate problem regarding optional enums. I made a checklist of possible improvements in #1708 so that enums integrate more smoothly with the rest of the types.


rjoussen (approved):
Thanks for working on this @sebproell!

PR #1725, opened 2026-02-04
Append `None` to choices of optional enums

rjoussen:
## Description and Context

As discussed in #1708. This will allow `std::optional<EnumType>` input specs to have a null default value.

#1708


rjoussen:
Yes, it shows an additional `null` option for optional enums (in addition to all choices or whatever is specified with `in_set`). I also have an updated version of #1688 which uses this feature for `ENTITY_TYPE`.

rjoussen:
This probably won't work if there are also descriptions for the enum values, which is why it's not yet closing (at least that part of) #1708

PR #1728, opened 2026-02-05
[Particles] Simplifications/nullifications for 2D problems

vovannikov:
Some of the 2D benchmarks, especially those operating with virtual walls, require some simplifications in the code. Eventually, this PR introduces 3 operations:

1. Nullify z-coordinate of the closest point returned by `Core::Geo::nearest_3d_object_on_element()` for the 2D case. The implementation additionally nullifies y-component for the 1D scenario.
2. The virtual wall particles are not generated along the z-axis for 2D and z- and y-axes for 1D.
3. A new parameter `ELEMENT_THICKNESS_FACTOR` is introduced to account for that the interaction force from the SPH field in the case of the 2D scenario needs to be scaled to 3D space with the element thickness in order to get the correct forces on the 3D structure.

There is one more component required, but it will be in a separate PR since it requires a bit more work.

Of course, the best way would be to write things in more rigorous way, e.g. templatizing certain parts of the particle code from the dimension (like deal.II does) such that unnecessary dummy degrees of freedom would not appear in the first place. But these modifications is the bare minimum required to handle 2D problems with the existing infrastructure.

@slfuchs, @ppraegla

### DIFF
--- a/src/particle/src/4C_particle_input.cpp
+++ b/src/particle/src/4C_particle_input.cpp
@@ -243,6 +243,13 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
               {.description = "type of transport velocity formulation",
                   .default_value = Particle::NoTransportVelocity}),
 
+          // reduced dimension scale factor for computing SPH forces
+          parameter<double>("REDUCED_DIMENSION_SCALE_FACTOR",
+              {.description =
+                      "scaling factor for the wall contact force to consider reduced "
+                      "dimensional modeling from 1D or 2D particle field to 3D structural field",
+                  .default_value = 1.0}),
+
           // type of temperature evaluation scheme
           parameter<TemperatureEvaluationScheme>(
               "TEMPERATUREEVALUATION", {.description = "type of temperature evaluation scheme",
--- a/src/particle/src/interaction/4C_particle_interaction_sph_virtual_wall_particle.cpp
+++ b/src/particle/src/interaction/4C_particle_interaction_sph_virtual_wall_particle.cpp
@@ -88,12 +88,22 @@ void Particle::SPHVirtualWallParticle::init_relative_positions_of_virtual_partic
   // number of particles per direction
   const int numparticleperdir = std::round(maxinteractiondistance / initialparticlespacing);
 
+  // get the kernel dimensionality
+  int kernel_dim = 0;
+  kernel_->kernel_space_dimension(kernel_dim);
+
+  const int s_start = (kernel_dim > 1) ? (-numparticleperdir + 1) : 0;
+  const int s_end = (kernel_dim > 1) ? numparticleperdir : 1;
+
+  const int t_start = (kernel_dim > 2) ? (-numparticleperdir + 1) : 0;
+  const int t_end = (kernel_dim > 2) ? numparticleperdir : 1;
+
   // iterate over virtual particles
   for (int r = 0; r < numparticleperdir; ++r)
   {
-    for (int s = (-numparticleperdir + 1); s < numparticleperdir; ++s)
+    for (int s = s_start; s < s_end; ++s)
     {
-      for (int t = (-numparticleperdir + 1); t < numparticleperdir; ++t)
+      for (int t = t_start; t < t_end; ++t)
       {
         // relative position of current virtual particle
         std::vector<double> currvirtualparticle(3);
--- a/src/particle/src/interaction/4C_particle_interaction_sph_momentum.cpp
+++ b/src/particle/src/interaction/4C_particle_interaction_sph_momentum.cpp
@@ -46,6 +46,7 @@ Particle::SPHMomentum::SPHMomentum(const Teuchos::ParameterList& params)
           Teuchos::getIntegralValue<Particle::TransportVelocityFormulation>(
               params_sph_, "TRANSPORTVELOCITYFORMULATION")),
       writeparticlewallinteraction_(params_sph_.get<bool>("WRITE_PARTICLE_WALL_INTERACTION")),
+      reduced_dimension_scale_factor_(params_sph_.get<double>("REDUCED_DIMENSION_SCALE_FACTOR")),
       allfluidtypes_(
           {Particle::Phase1, Particle::Phase2, Particle::DirichletPhase, Particle::NeumannPhase}),
       intfluidtypes_({Particle::Phase1, Particle::Phase2, Particle::NeumannPhase}),
@@ -852,7 +853,8 @@ void Particle::SPHMomentum::momentum_equation_particle_wall_contribution() const
       std::vector<double> nodal_force(numnodes * 3);
       for (int node = 0; node < numnodes; ++node)
         for (int dim = 0; dim < 3; ++dim)
-          nodal_force[node * 3 + dim] = funct[node] * wallcontactforce[dim];
+          nodal_force[node * 3 + dim] =
+              funct[node] * wallcontactforce[dim] * reduced_dimension_scale_factor_;
 
       // assemble nodal forces
       walldatastate->get_force_col()->sum_into_global_values(
--- a/src/particle/src/interaction/4C_particle_interaction_sph_momentum.hpp
+++ b/src/particle/src/interaction/4C_particle_interaction_sph_momentum.hpp
@@ -147,6 +147,9 @@ namespace Particle
     //! write particle-wall interaction output
     const bool writeparticlewallinteraction_;
 
+    //! reduced dimension scale factor
+    const double reduced_dimension_scale_factor_;
+
     //! set of all fluid particle types
     std::set<Particle::TypeEnum> allfluidtypes_;
 


src/particle/src/interaction/4C_particle_interaction_sph_neighbor_pairs.cpp
```diff
@@ -244,6 +244,10 @@ void Particle::SPHNeighborPairs::evaluate_particle_wall_pairs()
   // index of particle-wall pairs
   int particlewallpairindex = 0;
 
+  // kernel dimensionality
+  int kernel_dim = 0;
+  kernel_->kernel_space_dimension(kernel_dim);
```
georghammerl: What about refactoring this getter to:
`int kernel_dim = kernel_->get_kernel_space_dimension();`
before adding another call of the current variant.

src/particle/src/interaction/4C_particle_interaction_sph_virtual_wall_particle.cpp
```diff
@@ -88,12 +88,24 @@ void Particle::SPHVirtualWallParticle::init_relative_positions_of_virtual_partic
   // number of particles per direction
   const int numparticleperdir = std::round(maxinteractiondistance / initialparticlespacing);
 
+  // get the kernel dimensionality
+  int kernel_dim = 0;
+  kernel_->kernel_space_dimension(kernel_dim);
+  const bool has_2d = kernel_dim > 1;
```
georghammerl:
```suggestion
  const bool is_2d = kernel_dim > 1;
```

src/particle/src/4C_particle_input.cpp
```diff
@@ -243,6 +243,13 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
               {.description = "type of transport velocity formulation",
                   .default_value = Particle::NoTransportVelocity}),
 
+          // element thickness factor for computing SPH forces
+          parameter<double>("ELEMENT_THICKNESS_FACTOR",
+              {.description =
+                      "element thickness factor for computing wall nodal forces, mainly relevant "
+                      "for 2D simulations, whereas for 3D cases should be set to 1.0",
```
georghammerl: also relevant for 1D? Then, it is element area factor?

src/particle/src/interaction/4C_particle_interaction_sph_neighbor_pairs.cpp
vovannikov: I got the exact same idea when encountered this, but decided to proceed with these modifications and only then change the declaration of `get_kernel_space_dimension()` since it is used not only here. Just to keep this PR clean.

PS: it looks like we need a backlog, so many things appear while working on smth

src/particle/src/4C_particle_input.cpp
vovannikov: Yep, you right. For 1D, it should kind of square of it, since then we need to scale not just with the respect of the element thickness but to its surface.

src/particle/src/interaction/4C_particle_interaction_sph_virtual_wall_particle.cpp
vovannikov: I would not say that if `kernel_dim > 1`, then the kernel _is_ 2D it rather _may_ be 2D or higher.

georghammerl: Got it. I did not pay enough attention to the detailed logic.

src/particle/src/4C_particle_input.cpp
ppraegla: Maybe name it something like `WALL_NODAL_FORCE_DIMENSION_FACTOR` if it isn't too long 🤔 . And could you include what value the factor should be for 2d simulations? Just factor is a bit general.

slfuchs: I would also propose to change the name. What do you think about the following for input name, description and variable name?

`REDUCED_DIMENSION_SCALE_FACTOR`

"scaling factor for the wall contact force to consider reduced dimensional modeling from 1D particle field to 3D structural field"

`reduced_dimension_scale_factor`

src/particle/src/interaction/4C_particle_interaction_sph_virtual_wall_particle.cpp
```diff
@@ -88,12 +88,24 @@ void Particle::SPHVirtualWallParticle::init_relative_positions_of_virtual_partic
...
+  const bool has_3d = kernel_dim > 2;
+
+  const int s_start = has_2d ? (-numparticleperdir + 1) : 0;
+  const int s_end = has_2d ? numparticleperdir : 1;
+
+  const int t_start = has_3d ? (-numparticleperdir + 1) : 0;
+  const int t_end = has_3d ? numparticleperdir : 1;
```
slfuchs:
```suggestion
  const int s_start = (kernel_dim > 1) ? (-numparticleperdir + 1) : 0;
  const int s_end = (kernel_dim > 1) ? numparticleperdir : 1;

  const int t_start = (kernel_dim > 2) ? (-numparticleperdir + 1) : 0;
  const int t_end = (kernel_dim > 2) ? numparticleperdir : 1;
```

slfuchs: What about directly checking for the condition here? This is only performed once during initialization.

src/particle/src/interaction/4C_particle_interaction_sph_neighbor_pairs.cpp
slfuchs: To be honest I can't remember why I implemented it that way 😆

src/particle/src/interaction/4C_particle_interaction_sph_neighbor_pairs.cpp
```diff
@@ -279,6 +283,9 @@ void Particle::SPHNeighborPairs::evaluate_particle_wall_pairs()
     Core::Geo::ObjectType objecttype =
         Core::Geo::nearest_3d_object_on_element(ele, colelenodalpos, pos_i, closestpos);
 
+    // ensure that the identified point is in agreement with the kernel dimensionality
+    for (int dim = kernel_dim; dim < 3; ++dim) closestpos(dim) = 0.0;
```
slfuchs: This line is necessary to avoid close to zero values (due to round-off errors).

I think, we should add a safety check above, that is only performed in a debug version, that checks if one of the entries of closestpos contain positions that are larger than any value one would expect stemming from round-off errors. To give an example, it might be, that a wall element is not perpendicular to a 1D particle field, something like that (ASCII generated particle field and wall element):
/ooooooooooo\\
where it should actually be:
|ooooooooooo|
With the current implementation this would lead to a wrong distance.

src/particle/src/interaction/4C_particle_interaction_sph_virtual_wall_particle.cpp
vovannikov: Yep, makes sense.

vovannikov:
> This looks good and is a nice implementation of my hacks from the patch files!
> 
> One question that arises, also considering #1729, do we ensure that in 1D particles only have non-positive x-coordinates, and in 2D respectively non-positive x- and y-coordinates?

@slfuchs, I guess you meant non-zero coordinates. No, there is not such a check. I thought of this but decided not to add at least so far and within this PR. But we can consider it. 

> What happens if a user inputs a particle field with zero x-coordinates and non-zero y- and z-coordinates? Is there a safety check implemented?

Good question. We could try. But as far as I remember, in the current particles implementation, there is nothing that would restrain this explicitly. I guess it should remain working. Among the modifications, proposed in this PR, only this one is restrictive in this sense:

```c++
for (int dim = kernel_dim; dim < 3; ++dim) closestpos(dim) = 0.0;
```

And I have to admit, this is the one I had the most doubts with. After multiple numerical experiments, I personally observed that it has no effect, provided #1729 is enabled. So, probably, for consistency and simplicity of the code, this fix should be discarded.

src/particle/src/interaction/4C_particle_interaction_sph_neighbor_pairs.cpp
vovannikov: I see your concern, you are right. I would not like to introduce some tolerances here, this is always a bit fuzzy. I would probably remove this modification completely (see my reasoning below).

vovannikov:
> Maybe name it something like WALL_NODAL_FORCE_DIMENSION_FACTOR if it isn't too long 🤔 . And could you include what value the factor should be for 2d simulations? Just factor is a bit general.

I decided to pick the notation proposed by @slfuchs after all. It is a bit shorter and reflects nicely the meaning for both 2D and 3D.

vovannikov:
The recent commit accounts for all the comments and removes also the sanitation for `closestpos` components to avoid the situation described by @slfuchs, since the potential round-off errors do not seem to harm once the constraints from #1729 are active.

PS: ready for review

slfuchs (approved):
Looks good from my side!

PR #1729, opened 2026-02-05
[Particles] Introduce constraints feature to the particles framework

vovannikov:
To proceed with the 2D simulations in the SPH setting, we need to ensure that no action is produced in the z-direction, that may appear when the particles are coupled to 3D structures, e.g. particle walls. It is then required to explicitly nullify the corresponding velocity and acceleration components, since the 2D SPH kernel is not capable to handle these contributions properly.

As originally proposed by @slfuchs, such sanitation is easy to perform in the `pre_interaction_routine()` step of the time integrator. I just decided to introduce the concept of constraints, that would encapsulate this functionality and that could be extended to other tasks if needed. One could use the Dirichlet boundary conditions handler, but that handler has a bit more functionality than needed, making it a bit too expensive to apply at each timestep to each particle of the simulation domain.

Together with #1728, the 2D SPH cases requiring coupling to structural elements can not be simulated (including those from the paper of @slfuchs).


src/particle/src/algorithm/4C_particle_algorithm_constraints.hpp
```diff
@@ -0,0 +1,55 @@
...
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_PARTICLE_ALGORITHM_CONSTRAINTS_HPP
+#define FOUR_C_PARTICLE_ALGORITHM_CONSTRAINTS_HPP
+
+/*---------------------------------------------------------------------------*
+ | headers                                                                   |
+ *---------------------------------------------------------------------------*/
```
ppraegla: If you don't mind, could you remove these comments (same for the other `/*----*/` comments)? I plan to remove them in the particle code as I don't see their value.
```suggestion
```

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
```diff
@@ -0,0 +1,77 @@
...
+  }
+}
+
+std::unique_ptr<Particle::ConstraintsHandler> Particle::create_constraints(
+    const Teuchos::ParameterList& params)
+{
+  std::unique_ptr<Particle::ConstraintsHandler> constraints = nullptr;
+
+  const auto restrain_to_2d =
+      params.sublist("INITIAL AND BOUNDARY CONDITIONS").get<bool>("RESTRAIN_TO_2D");
```
ppraegla: Does it make sense to always restrain the simulation to 2d if the particle simulation is 2d? So, set the constraints based on the supplied kernel dimension and not an additional parameter?

src/particle/src/algorithm/4C_particle_algorithm_constraints.hpp
slfuchs: Thumbs up from my side. We did this in the old days, but these header lines have no real value for the code.

src/particle/src/algorithm/4C_particle_algorithm_constraints.hpp
```diff
@@ -0,0 +1,55 @@
...
+  class ConstraintsHandler
+  {
+   public:
+    virtual ~ConstraintsHandler() {};
+
+    virtual void apply(Particle::ParticleContainerBundleShrdPtr particle_container_bundle,
+        const std::set<Particle::TypeEnum>& types_to_integrate, const double time) const = 0;
+  };
+
+  class ConstraintsProjection2D : public ConstraintsHandler
```
slfuchs: Does it make sense to also add a `ConstraintsProjection1D` Handler along with this PR?

src/particle/src/4C_particle_input.cpp
```diff
@@ -177,7 +177,12 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
               {.description =
                       "Refer to the function ID describing the temperature boundary condition of "
                       "particle phase",
-                  .default_value = "none"})},
+                  .default_value = "none"}),
+
+          // restrain simulation to 2D
+          parameter<bool>("RESTRAIN_TO_2D",
```
slfuchs: Can we change the input to a tuple with option restrain to 1D, 2D or none?

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
vovannikov: I thought of that, and at first this is how I actually implemented it. But then I observed a few drawbacks of this approach:
1. Making this option dependent on the kernel type means that this constraint applies to SPH only, but the same need may arise in 2D DEM problems.
2. Some recent PD tests get broken. I do not know how it works currently on the PD side, but it looks that Kernel2D is used somewhere inside to assemble an actually a 3D problem. @georghammerl, please correct me if I got it wrong.
3. Some 2D SPH cases work just fine without these constraints (e.g., if no virtual walls are used), which is still an extra effort at each timestep.

src/particle/src/algorithm/4C_particle_algorithm_constraints.hpp
vovannikov: Indeed, good idea.

src/particle/src/4C_particle_input.cpp
vovannikov: Sure, aligns well with the comment above.

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
slfuchs: I would also do not mix any SPH specific stuff here in and keep it general.

vovannikov:
After thinking further and also looking at #1728, I could extend the flexibility of this feature and detect to which plane (2D) or axis (1D) projection should take place. This will be also merged with the coordinates check @slfuchs mentioned.

vovannikov:
> Thanks for the implementation. Could you add an input file test that uses the constraints? Right now, `create_constraints` is unused (unless I missed something).

Done! Now there is a test that crashes without this feature.

vovannikov:
Here, for the test to pass, #1728 needs to be merged first. Once #1728 is merged, I will rebase this one and proceed.

Ready for review.

As I have commented above, I would like to remove the requirement for the particles to be located at the plane z=0. I have already programmed an algo that detects the plane automatically and checks whether all the particles are coplanar, but I would prefer to add it in a follow up PR due to its relative complexity. Again to keep things for review simple.

tests/input_files/particle_sph_2d_openboundary_straight_channel_virtual_wall.4C.yaml
```diff
@@ -0,0 +1,417 @@
...
+      EXPONENT: 1
+      BACKGROUNDPRESSURE: 1
+      BULK_MODULUS: 1
+      DYNAMIC_VISCOSITY: 0.001
+      BULK_VISCOSITY: 0
+      ARTIFICIAL_VISCOSITY: 0
+  - MAT: 2
+    MAT_ParticleSPHBoundary:
+      INITRADIUS: 0.03
+      INITDENSITY: 1
```
ppraegla: Is material 2 unused?

tests/input_files/particle_sph_2d_openboundary_straight_channel_virtual_wall.4C.yaml
```diff
@@ -0,0 +1,417 @@
...
+  - "TYPE neumannphase POS 0.125 0.025 0.0"
+  - "TYPE neumannphase POS 0.105 0.035 0.0"
+  - "TYPE neumannphase POS 0.115 0.035 0.0"
+  - "TYPE neumannphase POS 0.125 0.035 0.0"
+  - "TYPE neumannphase POS 0.105 0.045 0.0"
+  - "TYPE neumannphase POS 0.115 0.045 0.0"
+  - "TYPE neumannphase POS 0.125 0.045 0.0"
+DESIGN SURFACE PARTICLE WALL:
+  - E: 1
+    MAT: -1
```
ppraegla, line 389: I think this is not used for pure particle problems.

vovannikov: This seems to be also required for virtual walls too.

tests/input_files/particle_sph_2d_openboundary_straight_channel_virtual_wall.4C.yaml
vovannikov: Indeed, I have removed it.

tests/input_files/particle_sph_2d_openboundary_straight_channel_virtual_wall.4C.yaml
ppraegla: You are right.

ppraegla (approved):

src/particle/src/4C_particle_input.hpp
```diff
@@ -112,6 +112,14 @@ namespace Particle
     RandlesCorrection
   };
 
+  //! type of constraint projection
+  enum Constraint
+  {
+    NoConstraint,
+    Projection1D,
```
georghammerl, line 119: It is not just 1D/2D, the projected directions are hard coded. Maybe reflect this also in the Projection name?

src/particle/src/4C_particle_input.cpp
```diff
@@ -177,7 +177,13 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
               {.description =
                       "Refer to the function ID describing the temperature boundary condition of "
                       "particle phase",
-                  .default_value = "none"})},
+                  .default_value = "none"}),
+
+          // kinematic constrain
+          parameter<Constraint>(
```
georghammerl, line 183: constrain**t**

src/particle/src/4C_particle_input.hpp
vovannikov: They are hard coded _so far_. This will be lifted in an upcoming PR.

PR #1730, opened 2026-02-05
Remove unnecessary member of enum class MatrixType

c-p-schmidt:
## Description and Context
Minor cleanup, as the title states. Alongside, I changed:
- switch from deprecated_selection to parameter
- make one class member optional, as an undefined enum class value is removed

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
@@ -31,9 +31,8 @@ namespace Core::LinAlg
   //! type of global system matrix in global system of equations
   enum class MatrixType
   {
-    undefined,           /*!< Type of system matrix is undefined. */
     sparse,              /*!< System matrix is a sparse matrix. */
-    block_field,         /*!< System matrix is a block matrix that consists of NxN matrices.
+    block,               /*!< System matrix is a block matrix that consists of NxN matrices.
                             In the simplest case, where each (physical) field is represented by just one
                             sparse matrix, N equals the number of (physical) fields of your problem.
                             However, it is also possible that the matrix of each (physical) field itself is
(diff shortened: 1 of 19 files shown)


src/core/linalg/src/sparse/4C_linalg_equilibrate.hpp
```diff
@@ -23,7 +23,7 @@ FOUR_C_NAMESPACE_OPEN
 
 namespace Core::LinAlg
 {
-  enum class MatrixType;
+  enum class MatrixType : std::uint8_t;
```
mayrmt: Is it really necessary to concretize the type? What's the benefit?

A drawback is that it had to be dinner at every forward declaration now.

sebproell: Depends heavily on the context. In most  cases, this is irrelevant, although in certain scenarios shrinking this from 4 byte (the default)  to 1 byte can drastically reduce the size of objects that have an enum member. If you have a ton of these objects in an array layout this will be relevant. I doubt that is the case here, and e.g. in the Matrix classes, probably the reduction in size is completely negated by padding.

TL;DR I would remove the `: std::uint8_t`.

src/core/linalg/src/sparse/4C_linalg_sparseoperator.hpp
```diff
@@ -29,9 +29,8 @@ namespace Core::LinAlg
   class SparseMatrix;
 
   //! type of global system matrix in global system of equations
-  enum class MatrixType
+  enum class MatrixType : std::uint8_t
   {
-    undefined,           /*!< Type of system matrix is undefined. */
     sparse,              /*!< System matrix is a sparse matrix. */
     block_field,         /*!< System matrix is a block matrix that consists of NxN matrices.
```
sebproell:
```suggestion
    block,         /*!< System matrix is a block matrix that consists of NxN matrices.
```

If you change the enum value, the old input files should remain valid. Otherwise, this is breaking change.

maxfirmbach (approved):

PR #1731, opened 2026-02-05
Cleanup TSI Algorithms

rjoussen:
## Description and Context

This PR attempts to cleanup the TSI Algorithms a little:

Remove some methods:
- `Monolithic::print_newton_conv (used, but was empty)`
- `Monolithic::combined_dbc_map` (unused)
- `Monolithic::apply_struct_coupling_state` (implemented identically in the Base Algorithm already)

Move between Base Algorithm and Monolithic/Partitioned:
- `prepare_contact_strategy` is moved to base, as it was implemented identically in both algorithms (and they just called the base method internally anyway)
- `update` is removed from base. It was implemented there before, but then overwritten only for Partitioned. Now its purely virtual in base and gets overwritten for Monolithic as well.
- `timeloop` was identical in both algorithms, hence moved to base.

Both specialised TSI Algorithms are declared `final`.

The last commit is just to make clangd a little happier.

### DIFF
--- a/src/tsi/4C_tsi_algorithm.cpp
+++ b/src/tsi/4C_tsi_algorithm.cpp
@@ -10,7 +10,6 @@
 #include "4C_adapter_str_factory.hpp"
 #include "4C_adapter_str_structure_new.hpp"
 #include "4C_adapter_str_wrapper.hpp"
-#include "4C_contact_lagrange_strategy.hpp"
 #include "4C_contact_lagrange_strategy_tsi.hpp"
 #include "4C_contact_meshtying_contact_bridge.hpp"
 #include "4C_contact_strategy_factory.hpp"
@@ -22,8 +21,6 @@
 #include "4C_global_data.hpp"
 #include "4C_io.hpp"
 #include "4C_mortar_multifield_coupling.hpp"
-#include "4C_solid_3D_ele.hpp"
-#include "4C_structure_new_model_evaluator_structure.hpp"
 #include "4C_thermo_adapter.hpp"
 #include "4C_tsi_input.hpp"
 #include "4C_tsi_utils.hpp"
@@ -169,25 +166,7 @@ TSI::Algorithm::Algorithm(MPI_Comm comm)
   // reset states
   structure_field()->discretization()->clear_state(true);
   thermo_field()->discretization()->clear_state(true);
-
-  return;
 }
-
-
-
-/*----------------------------------------------------------------------*
- | update (protected)                                        dano 12/09 |
- *----------------------------------------------------------------------*/
-void TSI::Algorithm::update()
-{
-  apply_thermo_coupling_state(thermo_field()->tempnp());
-  structure_field()->update();
-  thermo_field()->update();
-  if (contact_strategy_lagrange_ != nullptr)
-    contact_strategy_lagrange_->update((structure_field()->dispnp()));
-  return;
-}
-
 
 /*----------------------------------------------------------------------*
  | output (protected)                                        dano 12/09 |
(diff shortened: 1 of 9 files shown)


dragos-ana:
LGTM! It would make sense to squash the commits though

dragos-ana (approved):

src/tsi/4C_tsi_input.hpp
```diff
@@ -11,7 +11,8 @@
 #include "4C_config.hpp"
 
 #include "4C_io_input_spec.hpp"
-#include "4C_utils_exceptions.hpp"
+
+#include <sys/types.h>
```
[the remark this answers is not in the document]
maxfirmbach: @rjoussen I think Copilot is right here. Do you really need this header?

src/tsi/4C_tsi_algorithm.cpp
```diff
@@ -10,7 +10,6 @@
 #include "4C_adapter_str_factory.hpp"
 #include "4C_adapter_str_structure_new.hpp"
 #include "4C_adapter_str_wrapper.hpp"
-#include "4C_contact_lagrange_strategy.hpp"
```
maxfirmbach, line 13: Getting rid / reducing those headers is also good for modularization 👍.

src/tsi/4C_tsi_input.hpp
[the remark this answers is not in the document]
rjoussen: No, it was auto-included when I changed the enum definitions to `uint8_t` which I then reverted. But now I will include it again, so the header is used then :)

src/tsi/4C_tsi_algorithm.cpp
rjoussen: Yes, unfortunately, it's sometimes a little hard to see which headers can be deleted safely

amgebauer: If you are using VS Code with clangd, you can set these options to strict so that clang tells you which headers are unused or transitive includes: https://clangd.llvm.org/config#unusedincludes

rjoussen: I have that, but it often marks headers as unnecessary when they later turn out to be required. I think this happens whenever a method is called on a pointer that's typed as a base class, but the method exists only in a derived class. In that case, the required header is the one for the derived class, but clangd would like/only sees the necessity to include the base class header instead. At least thats my understanding.

rjoussen:
I squashed the commit and resolved the merge conflict due to #1732, so now ready for review :)

maxfirmbach (approved):
Thank your for cleaning this up! Always nice to see a lot of code deleted, while not loosing functionality.

src/tsi/4C_tsi_input.hpp
```diff
@@ -21,7 +22,7 @@ FOUR_C_NAMESPACE_OPEN
 namespace TSI
 {
   //! Type of coupling strategy for TSI problems
-  enum class SolutionSchemeOverFields
+  enum class SolutionSchemeOverFields : uint8_t
```
davidrudlstorfer: Just a question, doesn't the same apply to all of these?

https://github.com/4C-multiphysics/4C/pull/1730#discussion_r2770713933

src/tsi/4C_tsi_input.hpp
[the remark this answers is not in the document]
davidrudlstorfer: Correct me if I am wrong here, but normally we do not include sys headers in 4C, there's only one other in the particle module and otherwise we do not include them

src/tsi/4C_tsi_input.hpp
rjoussen: > [#1730 (comment)](https://github.com/4C-multiphysics/4C/pull/1730#discussion_r2770713933)

I did not read this. I don't really care if its 1 or 4 byte, I just changed it because clangd told me to and clangd is normally better at things like that than I am

src/tsi/4C_tsi_algorithm.cpp
amgebauer: In the case you described, you don't need the derived class to be included (you don't even know all derived classes normally). Typically what you describe is the problem of transitive includes, i.e., you rely that a header you include, includes a header that you actually need. We have a lot of these transitive includes in 4C. They are problematic: I you once be able to remove the header, some other (maybe unrelated) module fails to compile...

rjoussen:
@davidrudlstorfer The system header is no longer included, and the enums are changed back to as they were

maxfirmbach (approved):

davidrudlstorfer (approved):
Thanks for fixing this :)

src/tsi/4C_tsi_algorithm.cpp
rjoussen: @amgebauer Thanks for the explanation! Maybe in the long run it would make sense to have a pre-commit hook for that, which at least triggers for any newly added unused includes...

PR #1738, opened 2026-02-09
[discretization] Add Clement Interpolant

maxfirmbach:
## Description and Context

This PR adds a Clement quasi-interpolant to 4C with respective unit-testing and usage in thermo (with adaption of one output test as values change slightly).

The implementation also works for NURBS, I don't really use that one though ... The implementation was necessary such that all code paths triggered by the `thermo`-based tests run through.

### DIFF
--- a/src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
+++ b/src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
@@ -16,52 +16,74 @@ std::shared_ptr<Core::LinAlg::MultiVector<double>> Core::FE::compute_nodal_cleme
     const Core::FE::Discretization& dis, const Core::LinAlg::MultiVector<double>& values)
 {
   const size_t dimensions = dis.n_dim();
+  const size_t spatial_dimensions = 3;
+
+  FOUR_C_ASSERT(spatial_dimensions >= dimensions,
+      "The spatial dimension must be greater equal than the dimension of the discretization.");
+
   auto nodal_values = std::make_shared<Core::LinAlg::MultiVector<double>>(
       *dis.node_row_map(), values.num_vectors(), true);
 
+  const auto nurbsdis = dynamic_cast<const Core::FE::Nurbs::NurbsDiscretization*>(&dis);
+
   for (const auto& node : dis.my_row_node_range())
   {
-    auto element_patch = node.adjacent_elements();
+    const auto element_patch = node.adjacent_elements();
+
+    struct ElementData
+    {
+      int local_element_id;
+      double volume;
+    };
+
+    std::vector<ElementData> patch_data;
+    patch_data.reserve(element_patch.size());
+
+    double total_weight = 0.0;
+
+    for (const auto& element : element_patch)
+    {
+      const auto* user_element = element.user_element();
+      const auto* element_nodes = user_element->nodes();
+      const int number_of_nodes = user_element->num_node();
+
+      Core::LinAlg::SerialDenseMatrix node_coordinates(spatial_dimensions, number_of_nodes, true);
+      for (int node_number = 0; node_number < number_of_nodes; ++node_number)
+        for (size_t dim = 0; dim < dimensions; ++dim)
+          node_coordinates(dim, node_number) = element_nodes[node_number]->x()[dim];
+
+      std::vector<Core::LinAlg::SerialDenseVector> myknots(dimensions);
+      Core::LinAlg::SerialDenseVector myweights(number_of_nodes);
+
+      if (nurbsdis != nullptr)
+      {
+        Core::FE::Nurbs::get_my_nurbs_knots_and_weights(
+            *nurbsdis, user_element, myknots, myweights);
+      }
+
+      const double volume =
+          Core::Geo::element_volume(user_element->shape(), node_coordinates, myknots, myweights);
+
+      const int global_element_id = element.global_id();
+      const int local_element_id = values.get_vector(0).get_map().lid(global_element_id);
+
+      FOUR_C_ASSERT(local_element_id >= 0, "Global element id {} not owned by this process",
+          global_element_id);
+
+      patch_data.push_back({local_element_id, volume});
+      total_weight += volume;
+    }
 
     for (int column = 0; column < values.num_vectors(); ++column)
     {
-      double weight = 0.0;
+      const auto& element_value_vector = values.get_vector(column);
+      const auto values_span = element_value_vector.local_values_as_span();
+
       double node_value = 0.0;
+      for (const auto& data : patch_data)
+        node_value += values_span[data.local_element_id] * data.volume;
 
-      Core::LinAlg::Vector<double> element_value_vector(values.get_vector(column));
-
-      for (const auto& element : element_patch)
-      {
-        const auto* element_nodes = element.user_element()->nodes();
-        const int number_of_nodes = element.user_element()->num_node();
-
-        Core::LinAlg::SerialDenseMatrix node_coordinates(dimensions, number_of_nodes);
-        for (int node_number = 0; node_number < number_of_nodes; node_number++)
-          for (size_t dim = 0; dim < dimensions; ++dim)
-            node_coordinates(dim, node_number) = element_nodes[node_number]->x()[dim];
-
-        std::vector<Core::LinAlg::SerialDenseVector> myknots(dis.n_dim());
-        Core::LinAlg::SerialDenseVector myweights(number_of_nodes);
-
-        const auto nurbsdis = dynamic_cast<const Core::FE::Nurbs::NurbsDiscretization*>(&dis);
-        if (nurbsdis != nullptr)
-        {
-          Core::FE::Nurbs::get_my_nurbs_knots_and_weights(
-              *nurbsdis, element.user_element(), myknots, myweights);
-        }
-
-        const double volume = Core::Geo::element_volume(
-            element.user_element()->shape(), node_coordinates, myknots, myweights);
-
-        const int global_element_id = element.global_id();
-        const int local_element_id = element_value_vector.get_map().lid(global_element_id);
-
-        node_value += element_value_vector.local_values_as_span()[local_element_id] * volume;
-        weight += volume;
-      }
-
-      double interpolated_nodal_value = node_value / weight;
-
+      const double interpolated_nodal_value = node_value / total_weight;
       nodal_values->replace_global_value(node.global_id(), column, interpolated_nodal_value);
     }
   }


maxfirmbach:
Ready for review.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+        Core::LinAlg::SerialDenseMatrix node_coordinates(dimensions, number_of_nodes);
+        for (int node_number = 0; node_number < number_of_nodes; node_number++)
+          for (size_t dim = 0; dim < dimensions; ++dim)
+            node_coordinates(dim, node_number) = element_nodes[node_number]->x()[dim];
+
+        std::vector<Core::LinAlg::SerialDenseVector> myknots(dis.n_dim());
+        Core::LinAlg::SerialDenseVector myweights(number_of_nodes);
+
+        const auto nurbsdis = dynamic_cast<const Core::FE::Nurbs::NurbsDiscretization*>(&dis);
```
lauraengelhardt: You could move this cast outside of the loop.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+      for (const auto& element : element_patch)
+      {
+        const auto* element_nodes = element.user_element()->nodes();
+        const int number_of_nodes = element.user_element()->num_node();
+
+        Core::LinAlg::SerialDenseMatrix node_coordinates(dimensions, number_of_nodes);
+        for (int node_number = 0; node_number < number_of_nodes; node_number++)
+          for (size_t dim = 0; dim < dimensions; ++dim)
+            node_coordinates(dim, node_number) = element_nodes[node_number]->x()[dim];
```
lauraengelhardt: As far as I get it the geometry information should be redundant across all columns, right? Maybe this could also be moved out of the loop (if you consider worth the effort/performance gain)?

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
```diff
@@ -0,0 +1,72 @@
...
+          Core::FE::Nurbs::get_my_nurbs_knots_and_weights(
+              *nurbsdis, element.user_element(), myknots, myweights);
+        }
+
+        const double volume = Core::Geo::element_volume(
+            element.user_element()->shape(), node_coordinates, myknots, myweights);
+
+        const int global_element_id = element.global_id();
+        const int local_element_id = element_value_vector.get_map().lid(global_element_id);
+
```
lauraengelhardt, line 66:
```suggestion
FOUR_C_ASSERT(local_element_id >= 0, "Global element id {} not found in local map", global_element_id);
```

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: I moved the cast outside of the whole loop :+1:.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: I added the assert.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: Good point :+1:. I split the two inner loops, which are not nested anymore. First the element information is calculated and in a second separate loop we update the columns of the vector.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
```diff
@@ -0,0 +1,89 @@
...
+      Core::LinAlg::SerialDenseVector myweights(number_of_nodes);
+
+      if (nurbsdis != nullptr)
+      {
+        Core::FE::Nurbs::get_my_nurbs_knots_and_weights(
+            *nurbsdis, user_element, myknots, myweights);
+      }
+
+      const double volume =
+          Core::Geo::element_volume(user_element->shape(), node_coordinates, myknots, myweights);
```
mayrmt, line 65: Why can't this be done inside `Core::Geo::element_volume()`, where I just hand in a pointer to the current element and everything else is done inside?

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+      const int global_element_id = element.global_id();
+      const int local_element_id = values.get_vector(0).get_map().lid(global_element_id);
+
+      FOUR_C_ASSERT(
+          local_element_id >= 0, "Global element id {} not found in local map", global_element_id);
```
mayrmt: What is a "local map"?

```suggestion
      FOUR_C_ASSERT(
          local_element_id >= 0, "Global element id {} not owned by this process.", global_element_id);
```

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
```diff
@@ -0,0 +1,89 @@
...
+    {
+      const auto& element_value_vector = values.get_vector(column);
+      const auto values_span = element_value_vector.local_values_as_span();
+
+      double node_value = 0.0;
+      for (const auto& data : patch_data)
+        node_value += values_span[data.local_element_id] * data.volume;
+
+      const double interpolated_nodal_value = node_value / total_weight;
+      nodal_values->replace_global_value(node.global_id(), column, interpolated_nodal_value);
```
mayrmt, line 87: Why not replace a local value to avoid global lookup?

mayrmt (approved):
@maxfirmbach Looks good. I left some optional remarks.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: I don't want to pollute the `element_volume()` calculation routine too much.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: Somehow if I use the local id, replace_local_values complains that the id is not owned etc ... somehow the local indexing is not working properly(?)

mayrmt: Not great!

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
mayrmt: I was imagining a routine `element_volume()`, that just takes a pointer to an element and then returns the volume, so that the user does not have to care about pre-computing or pre-filling other data structures. Internally, a new clean `element_volume()` routine could call the existing one.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: Changed the description.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: I can do that, but I need to make a separate PR for this before merging this one, otherwise we clutter things together unnecessarily.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "4C_fem_geometry_element_volume.hpp"
+#include "4C_fem_nurbs_discretization_utils.hpp"
+
+FOUR_C_NAMESPACE_OPEN
+
+std::shared_ptr<Core::LinAlg::MultiVector<double>> Core::FE::compute_nodal_clement_interpolation(
+    const Core::FE::Discretization& dis, const Core::LinAlg::MultiVector<double>& values)
+{
+  const size_t dimensions = dis.n_dim();
+  const size_t spatial_dimensions = 3;
```
c-p-schmidt, line 19: Shouldn't there be an assertion that `dimensions` and `spatial_dimensions` match? At least you seem to assume that in lines 46-49.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
```diff
@@ -0,0 +1,90 @@
...
+      const auto* user_element = element.user_element();
+      const auto* element_nodes = user_element->nodes();
+      const int number_of_nodes = user_element->num_node();
+
+      Core::LinAlg::SerialDenseMatrix node_coordinates(spatial_dimensions, number_of_nodes, true);
+      for (int node_number = 0; node_number < number_of_nodes; ++node_number)
+        for (size_t dim = 0; dim < dimensions; ++dim)
+          node_coordinates(dim, node_number) = element_nodes[node_number]->x()[dim];
+
+      std::vector<Core::LinAlg::SerialDenseVector> myknots(dis.n_dim());
```
c-p-schmidt:
```suggestion
      std::vector<Core::LinAlg::SerialDenseVector> myknots(dimensions);
```

src/core/fem/tests/utils/4C_fem_general_utils_clement_interpolation_test.np2.cpp
```diff
@@ -0,0 +1,517 @@
...
+      nodal_values->get_vector(1).max_value(&max_value);
+      EXPECT_NEAR(max_value, 100.0, 1e-14);
+      nodal_values->get_vector(2).max_value(&max_value);
+      EXPECT_NEAR(max_value, 1000.0, 1e-14);
+
+      // The nodes in the middle have to hold the arithmetic average of all values, due to a uniform
+      // mesh size
+      if (my_rank == 0)
+      {
+        EXPECT_NEAR(nodal_values->get_vector(0).get_values()[1], 15.0, 1e-14);
```
c-p-schmidt, line 335: What you check in the if and else branches is arbitrary, or did I miss something?

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: @mayrmt I had another look and this is not easily possible due to the `Cut::Element` implementation. Or we have another method that calls another method to get the actual thing we want.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: That's the thing here, they don't have to match.

c-p-schmidt: But then spatial_dimensions needs to be >= dimensions.

src/core/fem/tests/utils/4C_fem_general_utils_clement_interpolation_test.np2.cpp
maxfirmbach: No its actually the average of the values on the left 10.0 and the right 20.0 :smile:.

maxfirmbach: The if-else is there to test for a specific point at the interface. The point itself is kind off arbitrary yes, it just has to be positioned at the interface.

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: Fixed!

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: That is correct, yes!

mayrmt (approved):

src/core/fem/src/general/utils/4C_fem_general_clement_interpolation.cpp
maxfirmbach: @c-p-schmidt I added a respective assert :+1:.

PR #1739, opened 2026-02-09
Rewrite MAT_Struct_[Thermo]PlasticLinElast to work with tensors

ischeider:
## Description and Context

I changed the variables in these two material subroutines to tensors in order to make it more consistent with other routines. Doing so I found a small typo in one of the equations, which has been fixed as well. Nevertheless, the hope that I could speed up the simulations was not fulfilled. Times don#t change significantly.

The thermoPlasticLinElast contains the option of defining a multi-linear hardening curve. However, up to now this is not tested. I will refactor this option and add a test in a second PR.

Related to #583

### DIFF
--- a/src/mat/4C_mat_thermoplasticlinelast.cpp
+++ b/src/mat/4C_mat_thermoplasticlinelast.cpp
@@ -102,38 +102,32 @@ void Mat::ThermoPlasticLinElast::pack(Core::Communication::PackBuffer& data) con
   add_to_pack(data, matid);
 
   // pack history data
-  int histsize;
   // if material is not initialised, i.e. start simulation, nothing to pack
-  if (!initialized())
-  {
-    histsize = 0;
-  }
-  else
-  {
-    // if material is initialised (restart): size equates number of gausspoints
-    histsize = strainpllast_->size();
-  }
+  int histsize = (initialized()) ? strainpllast_.size() : 0;
+
   add_to_pack(data, histsize);  // Length of history vector(s)
   for (int var = 0; var < histsize; ++var)
   {
     // insert last converged states
-    add_to_pack(data, strainpllast_->at(var));
-    add_to_pack(data, backstresslast_->at(var));
-    add_to_pack(data, strainbarpllast_->at(var));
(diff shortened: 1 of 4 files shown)


amgebauer (approved):
> Nevertheless, the hope that I could speed up the simulations was not fulfilled. Times don#t change significantly.

Oh, you cannot expect a speedup when switching from matrix to tensors. Both are using the same math kernel under the hood. Many cases should compile to the same or similar assembly. What you can expect, though, is a more readable source code.

ischeider:
> Oh, you cannot expect a speedup when switching from matrix to tensors. Both are using the same math kernel under the hood. 

I agree, I rather expected a speed up because of the fixed flaw, since it affects the calculation of cmat.

rjoussen:
@ischeider: As a fellow (occasional) user of this material, could you point out where the flaw was? Mainly out of curiosity:)

src/mat/4C_mat_thermoplasticlinelast.cpp
```diff
@@ -1179,41 +1013,25 @@ void Mat::ThermoPlasticLinElast::dissipation_coupl_cond(
...
+  double fac_kinlin2 = sqrt(3.0 / 2.0) * fac_kinlin1;
   // contribution: Id4^#
-  Dmech_kin_d.update(fac_kinlin1, id4sharp);
-  // contribution: Id \otimes Id
-  Dmech_kin_d.multiply_nt(fac_kinlin2, id2, id2, 1.0);
+  Dmech_kin_d = fac_kinlin2 * Id;
 
   double fac_lin_3 = 0.0;
-  double fac_lin_4 = 0.0;
-  fac_lin_4 = 3.0 * G * Hkin * Hiso;
```
ischeider, line 1202: The flaw was the multiplication in this line, it should be `3 G + Hkin + Hiso`

rjoussen: Makes sense, thanks for fixing this!

amgebauer (approved):

PR #1740, opened 2026-02-09
Add default status and type

georghammerl:
## Description and Context
During communication in Peridynamics: `ParticleType` and `ParticleStatus` are unknown while sending data and will be filled on the receiving processor properly. Hence, a default/not yet initialized value is necessary.

## Interested parties
@ppraegla @slfuchs

### DIFF
--- a/src/particle/src/engine/4C_particle_engine_enums.hpp
+++ b/src/particle/src/engine/4C_particle_engine_enums.hpp
@@ -157,11 +157,12 @@ namespace Particle
     Phase2,  //!< particle type for particles of second phase
     /*---------------------------------------------------------------------------*/
     // particle types for SPH interaction
-    BoundaryPhase,   //!< particle type for boundary phase particles
-    RigidPhase,      //!< particle type for rigid phase particles
-    DirichletPhase,  //!< particle type for dirichlet phase particles (open boundary)
-    NeumannPhase,    //!< particle type for neumann phase particles (open boundary)
-    PDPhase          //!< particle tpe for peridynamic phase particles
+    BoundaryPhase,     //!< particle type for boundary phase particles
+    RigidPhase,        //!< particle type for rigid phase particles
+    DirichletPhase,    //!< particle type for dirichlet phase particles (open boundary)
+    NeumannPhase,      //!< particle type for neumann phase particles (open boundary)
+    PDPhase,           //!< particle type for peridynamic phase particles
+    UninitializedType  //!< not yet defined particle type
     /*---------------------------------------------------------------------------*/
   };
 
@@ -204,8 +205,9 @@ namespace Particle
    */
   enum ParticleStatus
   {
-    Owned,   //!< particle status for particles being owned on processors
-    Ghosted  //!< particle status for particles being ghosted on processors
+    Owned,               //!< particle status for particles being owned on processors
+    Ghosted,             //!< particle status for particles being ghosted on processors
+    UninitializedStatus  //!< not yet defined particle status
   };
 
   /*!
--- a/src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
+++ b/src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
@@ -504,20 +504,16 @@ void Particle::PDNeighborPairs::unpack_peridynamic_bond_list_data(const std::vec
   Core::Communication::UnpackBuffer data(buffer);
   while (!data.at_end())
   {
-    Particle::TypeEnum type_i;
-    Particle::StatusEnum status_i;
-
-    Particle::TypeEnum type_j;
-    Particle::StatusEnum status_j;
-
     int globalid_i;
     extract_from_pack(data, globalid_i);
     int globalid_j;
     extract_from_pack(data, globalid_j);
 
     // setup default tuples with proper global ids
-    Particle::LocalGlobalIndexTuple tuple_i = std::make_tuple(type_i, status_i, -1, globalid_i);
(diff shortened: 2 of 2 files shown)


src/particle/src/engine/4C_particle_engine_enums.hpp
```diff
@@ -161,7 +161,8 @@ namespace Particle
     RigidPhase,      //!< particle type for rigid phase particles
     DirichletPhase,  //!< particle type for dirichlet phase particles (open boundary)
     NeumannPhase,    //!< particle type for neumann phase particles (open boundary)
-    PDPhase          //!< particle tpe for peridynamic phase particles
+    PDPhase,         //!< particle tpe for peridynamic phase particles
```
slfuchs:
```suggestion
    PDPhase,         //!< particle type for peridynamic phase particles
```

slfuchs:
I am fine with introducing enums for undefined type and status. However i think i didn't completely understand why this is necessary here. With DEM interaction we implemented history pairs that are also communicated and there was no need for an undefined type or status.

src/particle/src/engine/4C_particle_engine_enums.hpp
```diff
@@ -204,8 +205,9 @@ namespace Particle
    */
   enum ParticleStatus
   {
-    Owned,   //!< particle status for particles being owned on processors
-    Ghosted  //!< particle status for particles being ghosted on processors
+    Owned,           //!< particle status for particles being owned on processors
+    Ghosted,         //!< particle status for particles being ghosted on processors
+    UndefinedStatus  //!< not yet defined particle status
```
sebproell: It's unusual to put the `undefined` enum value last instead of first. If there is no specific reason, I would put it first.

slfuchs: I prefer to have it at the end. The enum is also used to index vectors of the particle container bundle. Putting it first would mean that we have additional entries for an undefined status that actually never gets filled with a particle container. I know this is not performance critical at all, but it just feels not correct to do it. Right now, the undefined enums are only used in the particle object for communication.

sebproell: Well, that is a reason. Nvm my comment then ;)

slfuchs (approved):

ppraegla (approved):
> I am fine with introducing enums for undefined type and status. However i think i didn't completely understand why this is necessary here. With DEM interaction we implemented history pairs that are also communicated and there was no need for an undefined type or status.

I'm also interested in why this state is necessary ;)

georghammerl:
Thanks to @ppraegla for discussion : It is more a "not yet defined" status/type, such that renaming to `ParticleType::UninitializedType` and `ParticleStatus::UninitializedStatus` seems appropriate.

Tried to refactor the code to circumvent the not yet initialized data but it makes the code more complex with parallel data structures which need to be kept in sync. Not really desireable in my eyes.

Background/Difference to DEM implementation: Usually, loop is over close neighbor pairs with known status, type and local id of the pair's particles. From this data, the global id can be accessed which is used, e.g., in DEM to get information about history data. In peridynamics, it is necessary the other way round. The loop over peridynamic pairs would include an inverse data access from the global id of the particle to the respective close neighbor pair. This is not intended. Hence, little bit special treatment is necessary.

src/particle/src/engine/4C_particle_engine_enums.hpp
```diff
@@ -157,11 +157,12 @@ namespace Particle
...
-    RigidPhase,      //!< particle type for rigid phase particles
-    DirichletPhase,  //!< particle type for dirichlet phase particles (open boundary)
-    NeumannPhase,    //!< particle type for neumann phase particles (open boundary)
-    PDPhase          //!< particle tpe for peridynamic phase particles
+    BoundaryPhase,     //!< particle type for boundary phase particles
+    RigidPhase,        //!< particle type for rigid phase particles
+    DirichletPhase,    //!< particle type for dirichlet phase particles (open boundary)
+    NeumannPhase,      //!< particle type for neumann phase particles (open boundary)
+    PDPhase,           //!< particle type for peridynamic phase particles
+    UninitializedType  //!< not yet defined particle type
```
ppraegla:
```suggestion
    Uninitialized  //!< not yet defined particle type
```
I think we can remove the suffix because the enum is already called ParticleType.

src/particle/src/engine/4C_particle_engine_enums.hpp
```diff
@@ -204,8 +205,9 @@ namespace Particle
    */
   enum ParticleStatus
   {
-    Owned,   //!< particle status for particles being owned on processors
-    Ghosted  //!< particle status for particles being ghosted on processors
+    Owned,               //!< particle status for particles being owned on processors
+    Ghosted,             //!< particle status for particles being ghosted on processors
+    UninitializedStatus  //!< not yet defined particle status
```
ppraegla:
```suggestion
    Uninitialized  //!< not yet defined particle status
```
Same as above

src/particle/src/engine/4C_particle_engine_enums.hpp
georghammerl: Sure

src/particle/src/engine/4C_particle_engine_enums.hpp
georghammerl: Fine

src/particle/src/engine/4C_particle_engine_enums.hpp
ppraegla: Sorry. I just saw that this doesn't work because `Uninitialized` is declared twice since we use enums and not enum classes. `UninitializedType` and `UninitializedStatus` are fine to be unique.

PR #1742, opened 2026-02-10
[Particles] Detect automatically primary axis for the 1D and 2D projection constraints

vovannikov:
This PR is the continuation of #1729. It adds an algorithm to detect the projection axis for both 1D and 2D situations. It does so locally at each rank and then gathers the info from all ranks checking for consistency between them. All the particles are then checked if they are properly position with respect to the identified projection axis.

#1728, #1729

### DIFF
--- a/src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
+++ b/src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
@@ -70,13 +70,6 @@ int Particle::ConstraintsProjectionBase::calc_primary_axis(
 {
   int local_result = calc_primary_axis_local(particle_container_bundle, types_to_integrate);
 
-  int local_invalid = (local_result == axis_invalid_);
-  int global_invalid;
-
-  MPI_Allreduce(&local_invalid, &global_invalid, 1, MPI_INT, MPI_LOR, comm_);
-
-  FOUR_C_ASSERT_ALWAYS(!global_invalid, "Invalid or inconsistent primary axis detected");
-
   int global_result;
   MPI_Allreduce(&local_result, &global_result, 1, MPI_INT, MPI_MAX, comm_);
 


src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
```diff
@@ -55,13 +64,158 @@ void Particle::ConstraintsProjectionBase::apply(
...
+    // no owned particles of current particle type or not enough particles at all
+    if (n_particle_stored < 3) continue;
+
+    // get pointer to particle position
+    double* pos = container->get_ptr_to_state(Particle::Position, 0);
+
+    // get particle state dimension
+    const int pos_state_dim = container->get_state_dim(Particle::Position);
+
+    int i0 = 0, i1 = 1, i2 = i1 + 1;
```
georghammerl:
```suggestion
    int i0 = 0, i1 = 1, i2 = 2;
```

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
```diff
@@ -55,13 +64,158 @@ void Particle::ConstraintsProjectionBase::apply(
...
+
+    const auto p0 = pos + pos_state_dim * i0;
```
georghammerl:
```suggestion
    const auto p0 = pos;
```

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
georghammerl: or even do not introduce `i0` and `i1` ?

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
```diff
@@ -55,13 +64,158 @@ void Particle::ConstraintsProjectionBase::apply(
...
+    for (; i2 < n_particle_stored; ++i2)
+    {
+      const auto p2 = pos + pos_state_dim * i2;
+
+      ParticleUtils::vec_set(b, p2);
+      ParticleUtils::vec_add_scale(b, -1., p0);
+
+      ParticleUtils::vec_set_cross(c, a, b);
+
+      if (ParticleUtils::vec_norm_two(c) > eps_) break;
```
georghammerl: Not sure whether such a hard coded (quite strict) tolerance is reasonable for all setups (tiny particles vs large ones)?

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
```diff
@@ -55,13 +64,158 @@ void Particle::ConstraintsProjectionBase::apply(
...
+
+    // Means we have found the normal
+    if (i2 < n_particle_stored)
+    {
+      // Identify the direction
+      int dim = 0;
+      for (; dim < 3; ++dim)
+      {
+        double axis[3] = {0., 0., 0.};
+        axis[dim] = 1;
```
georghammerl:
```suggestion
        axis[dim] = 1.0;
```

georghammerl:
@vovannikov Nice automatic detection 👍

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
vovannikov: I can set it to `1e-16`. There is a way to customize it, but I do not want to create yet another input parameter.

ppraegla: You could specify the tolerance with respect to the smallest particle radius. Something like

```c++
min_radius = container->get_min_value_of_state(Radius);
if (ParticleUtils::vec_norm_two(c) > eps_ * min_radius) ...
```
Since you call it only during setup, the performance loss would be fine.

I think this can be helpful also at other places where we need to set some tolerance. Maybe a free function `double get_tolerance_based_on_feature_size(Particle::ParticleContainerBundleShrdPtr particle_container_bundle, ParticleState state)` would be helpful.

But this can be done later as well. I'm fine with the current hard coded solution.

georghammerl: `1e-16` is even more strict. Maybe leave it to `1e-12` for now and let us test with tiny dimension and large scale examples and see whether this works for all configurations equally good. Maybe, a problem size (radius/distance) tolerance is benefitial.

slfuchs: In SPH checks also the radius is used in such scenarios. I guess this should work here, since all particle interaction types use the radius so far.

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
```diff
@@ -12,16 +12,25 @@
...
+Particle::ConstraintsProjectionBase::ConstraintsProjectionBase(MPI_Comm comm, double tol)
+    : comm_(comm), eps_(tol)
+{
+}
+
 void Particle::ConstraintsProjectionBase::apply(
     Particle::ParticleContainerBundleShrdPtr particle_container_bundle,
     const std::set<Particle::TypeEnum>& types_to_integrate, const double time) const
 {
+  // perform lazy setup
```
slfuchs: I think we can leave the comment here, or not call it lazy 😆

slfuchs:
Looks good, thanks 👍

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
vovannikov: I have changed the tolerances to account for the particle radii. Also the dot products are compared to 1 when this is relevant.

georghammerl (approved):

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
```diff
@@ -55,13 +64,166 @@ void Particle::ConstraintsProjectionBase::apply(
...
+
+  FOUR_C_ASSERT_ALWAYS(!global_invalid, "Invalid or inconsistent primary axis detected");
+
+  int global_result;
+  MPI_Allreduce(&local_result, &global_result, 1, MPI_INT, MPI_MAX, comm_);
+
+  FOUR_C_ASSERT_ALWAYS(global_result != axis_invalid_, "No rank could determine primary axis");
+
+  FOUR_C_ASSERT_ALWAYS(local_result == axis_invalid_ || local_result == global_result,
+      "Primary axis mismatch across ranks");
```
ppraegla, line 79: Can these asserts be reached? If an axis is invalid on any rank. `FOUR_C_ASSERT_ALWAYS(!global_invalid` should already have triggered.

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
```diff
@@ -55,13 +64,166 @@ void Particle::ConstraintsProjectionBase::apply(
...
+}
+
+int Particle::ConstraintsProjection2D::calc_primary_axis_local(
+    Particle::ParticleContainerBundleShrdPtr particle_container_bundle,
+    const std::set<Particle::TypeEnum>& types_to_integrate) const
+{
+  int axis_direction = axis_invalid_;
+
+  // working vectors
+  double a[3], b[3], c[3];
```
ppraegla, line 137: Is there a reason you use c-style arrays over `std::array`? Also at other places

vovannikov: The main reason is historic legacy of that how `ParticleUtils::` functions are organized: they work with raw pointers and `std::array` does not decay implicitly to a raw pointer.

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
vovannikov: If the first assert passes, it means that at least one rank was able to identify the primary axis and this data was communicated globally as a reduction operation. Then we go to the second one. The possible situations on a given rank then are:

- the current rank was not able previously to identify the local axis (e.g., did not have enough particles), then this is fine, no contradiction, it will just use the information provided globally by a more successful partner,
- the current rank has previously identified its local axis, then we need to check if this axis is the same as the one globally distributed, they should agree, otherwise, inconsistency is detected.

None of these questions is automatically answered on all ranks even if some global axis has been identified.

vovannikov: Ah sorry, misunderstood your question at first. You referred the assert before these two. It looks you are right. I would remove then exactly the assert `FOUR_C_ASSERT_ALWAYS(!global_invalid` for simplicity, good point.

src/particle/src/algorithm/4C_particle_algorithm_constraints.cpp
ppraegla: You could use `a.data()` when passed to the `ParticleUtils`. But I leave it up to you if you want to change it.

slfuchs: It's not legacy, it's the design. The `ParticleUtils` are meant to work with pointers to the particle states of an individual particle. Note that applying a Structure of Array approach the particle states are stored in the particle container that is indead a std:: vector. When reusing the `ParticleUtils` I mostly used c style arrays for simplicity.

vovannikov: Yes, I got kind of the way you have described it. I am fine with this solution, for this reason I decided to write my code in a similar way as you have done it, much more consistent then.

Issue #1428, opened 2025-10-23
[solver] Provide interface node map in new structure time integration


maxfirmbach:
## Description

Due to  #878 we are able to construct an interface node map, which is provided to `MueLu` to be able to perform an interface based aggregation. This feature is really nice, but only available in the deprecated old structure time integration of 4C. We should provide this information in the new structure time integration too.

## Possible Solution and Definition of Done

- [ ] Provide `"Interface DualNodeID to PrimalNodeID"` in the new structure time integration
- [ ] Add test that perform the respective linear solve using this feature

## Interested Parties

@mayrmt @PhilipOesterlePekrun

PhilipOesterlePekrun:
I will look into it soon @maxfirmbach

maxfirmbach:
@PhilipOesterlePekrun I assigned you to this issue, just for formal completeness.

PR #1744, opened 2026-02-10
Make the new structure time integration compatible with saddle-point contact/meshtying preconditioners (MueLu)

PhilipOesterlePekrun:
@mayrmt @maxfirmbach 

## Description and Context

This PR enables the use of the new structural time integration (INT_STRATEGY: Standard) alongside MueLu preconditioning for saddle point contact/meshtying problems, both with the dof-based or node-based interface aggregation strategies. The main change is the additional construction of the null space for the Lagrange multiplier coupling matrix and the dual to primal node mapping.

Additionally, Epetra_Map was replaced with Core::LinAlg in the old and new time integration.

## Remarks
This currently works for a single contact or meshtying interface, just as in the old structural time integration. If there is a desire to implement this for any number of interfaces, some restructuring might be required.

The functions `NOX::Nln::CONTACT::LinearSystem::set_solver_options()` and `NOX::Nln::MeshTying::LinearSystem::set_solver_options()` are identical (is it supposed to be that way?), so one could of course call one of these functions in the other one (as they are both conceptually const-qualified). That would require includes and maybe it somehow confuses the organization/logic, so just an idea.

## Tests
I have added three tests:
- `contact3D_sp_std_simple_new_struct.4C.yaml`
- contact3D_sp_std_uzawa_nbinterfaceagg_new_struct.4C.yaml (using the new file `contact3D_sp_uzawa_nbinterfaceagg.xml`)
- `meshtying3D_sp_dual_simple_new_struct.4C.yaml`

which I chose just to have a bit of variety. As the names suggest, they all use the new structural time integration, with one also using the node-based interface aggregation and the other two using the dof-based interface aggregation.

Closes https://github.com/4C-multiphysics/4C/issues/1428

### DIFF
--- a/src/contact/src/4C_contact_manager.cpp
+++ b/src/contact/src/4C_contact_manager.cpp
@@ -33,6 +33,7 @@
 #include "4C_io_control.hpp"
 #include "4C_io_input_parameter_container.hpp"
 #include "4C_linalg_utils_sparse_algebra_manipulation.hpp"
+#include "4C_linear_solver_method.hpp"
 #include "4C_mortar_input.hpp"
 #include "4C_mortar_utils.hpp"
 
@@ -703,6 +704,24 @@ bool CONTACT::Manager::read_and_check_input(Teuchos::ParameterList& cparams) con
       Teuchos::getIntegralValue<Mortar::ParallelRedist>(
           mortarParallelRedistParams, "PARALLEL_REDIST") != Mortar::ParallelRedist::redist_none)
     FOUR_C_THROW("Parallel redistribution not yet implemented for TSI problems");
+
+  if (const int solverNumber = contact.get<int>("LINEAR_SOLVER");
+      solverNumber != -1 &&
+      Teuchos::getIntegralValue<CONTACT::SolvingStrategy>(contact, "STRATEGY") ==
+          CONTACT::SolvingStrategy::lagmult &&
+      Teuchos::getIntegralValue<CONTACT::SystemType>(contact, "SYSTEM") !=
+          CONTACT::SystemType::condensed &&
+      Teuchos::getIntegralValue<CONTACT::SystemType>(contact, "SYSTEM") !=
+          CONTACT::SystemType::condensed_lagmult &&
+      Global::Problem::instance()
+              ->solver_params(solverNumber)
+              .get<Core::LinearSolver::PreconditionerType>("AZPREC") ==
+          Core::LinearSolver::PreconditionerType::multigrid_muelu &&
+      Teuchos::getIntegralValue<Mortar::ParallelRedist>(
+          mortarParallelRedistParams, "PARALLEL_REDIST") != Mortar::ParallelRedist::redist_none)
+    FOUR_C_THROW(
+        "Parallel redistribution is currently not supported for CONTACT problems in saddle-point "
+        "formulation preconditioned with MueLu");
 
   // *********************************************************************
   // adhesive contact
--- a/src/contact/src/4C_contact_meshtying_manager.cpp
+++ b/src/contact/src/4C_contact_meshtying_manager.cpp
@@ -16,6 +16,7 @@
 #include "4C_global_data.hpp"
 #include "4C_io.hpp"
 #include "4C_linalg_utils_sparse_algebra_manipulation.hpp"
+#include "4C_linear_solver_method.hpp"
 #include "4C_mortar_defines.hpp"
 #include "4C_mortar_element.hpp"
 #include "4C_mortar_input.hpp"
@@ -441,6 +442,24 @@ bool CONTACT::MtManager::read_and_check_input(
           Teuchos::getIntegralValue<CONTACT::SystemType>(meshtying, "SYSTEM") ==
               CONTACT::SystemType::condensed_lagmult))
     FOUR_C_THROW("Condensation of linear system only possible for dual Lagrange multipliers");
+
+  if (const int solverNumber = meshtying.get<int>("LINEAR_SOLVER");
+      solverNumber != -1 && (onlymeshtying || meshtyingandcontact) &&
+      Teuchos::getIntegralValue<CONTACT::SolvingStrategy>(meshtying, "STRATEGY") ==
+          CONTACT::SolvingStrategy::lagmult &&
+      Teuchos::getIntegralValue<CONTACT::SystemType>(meshtying, "SYSTEM") !=
+          CONTACT::SystemType::condensed &&
+      Teuchos::getIntegralValue<CONTACT::SystemType>(meshtying, "SYSTEM") !=
+          CONTACT::SystemType::condensed_lagmult &&
+      Global::Problem::instance()
+              ->solver_params(solverNumber)
+              .get<Core::LinearSolver::PreconditionerType>("AZPREC") ==
+          Core::LinearSolver::PreconditionerType::multigrid_muelu &&
+      Teuchos::getIntegralValue<Mortar::ParallelRedist>(
+          mortarParallelRedistParams, "PARALLEL_REDIST") != Mortar::ParallelRedist::redist_none)
+    FOUR_C_THROW(
+        "Parallel redistribution is currently not supported for MESHTYING problems in saddle-point "
+        "formulation preconditioned with MueLu");
 
   if (Teuchos::getIntegralValue<Mortar::ParallelRedist>(mortarParallelRedistParams,
           "PARALLEL_REDIST") == Mortar::ParallelRedist::redist_dynamic and
--- a/src/contact/src/4C_contact_meshtying_strategy_factory.cpp
+++ b/src/contact/src/4C_contact_meshtying_strategy_factory.cpp
@@ -19,6 +19,7 @@
 #include "4C_io.hpp"
 #include "4C_io_pstream.hpp"
 #include "4C_linalg_utils_sparse_algebra_math.hpp"
+#include "4C_linear_solver_method.hpp"
 #include "4C_mortar_element.hpp"
 #include "4C_mortar_node.hpp"
 #include "4C_mortar_utils.hpp"
@@ -137,6 +138,23 @@ void Mortar::STRATEGY::FactoryMT::read_and_check_input(Teuchos::ParameterList& p
     FOUR_C_THROW(
         "ERROR: Consistent dual shape functions in boundary elements only for Lagrange multiplier "
         "strategy.");
+  if (const int solverNumber = meshtying.get<int>("LINEAR_SOLVER");
+      solverNumber != -1 && (onlymeshtying || meshtyingandcontact) &&
+      Teuchos::getIntegralValue<CONTACT::SolvingStrategy>(meshtying, "STRATEGY") ==
+          CONTACT::SolvingStrategy::lagmult &&
+      Teuchos::getIntegralValue<CONTACT::SystemType>(meshtying, "SYSTEM") !=
+          CONTACT::SystemType::condensed &&
+      Teuchos::getIntegralValue<CONTACT::SystemType>(meshtying, "SYSTEM") !=
+          CONTACT::SystemType::condensed_lagmult &&
+      Global::Problem::instance()
+              ->solver_params(solverNumber)
+              .get<Core::LinearSolver::PreconditionerType>("AZPREC") ==
+          Core::LinearSolver::PreconditionerType::multigrid_muelu &&
+      Teuchos::getIntegralValue<Mortar::ParallelRedist>(
+          mortarParallelRedistParams, "PARALLEL_REDIST") != Mortar::ParallelRedist::redist_none)
+    FOUR_C_THROW(
+        "Parallel redistribution is currently not supported for MESHTYING problems in saddle-point "
+        "formulation preconditioned with MueLu");
 
   if (Teuchos::getIntegralValue<Mortar::ConsistentDualType>(mortar, "LM_DUAL_CONSISTENT") !=
           Mortar::consistent_none &&
--- a/src/contact/src/4C_contact_strategy_factory.cpp
+++ b/src/contact/src/4C_contact_strategy_factory.cpp
@@ -32,6 +32,7 @@
 #include "4C_io.hpp"
 #include "4C_io_pstream.hpp"
 #include "4C_linalg_utils_sparse_algebra_math.hpp"
+#include "4C_linear_solver_method.hpp"
 #include "4C_scatra_timint_meshtying_strategy_s2i.hpp"
 #include "4C_ssi_input.hpp"
 #include "4C_structure_new_timint_basedataglobalstate.hpp"
@@ -110,6 +111,24 @@ void CONTACT::STRATEGY::Factory::read_and_check_input(Teuchos::ParameterList& pa
       Teuchos::getIntegralValue<CONTACT::SolvingStrategy>(contact, "STRATEGY") !=
           CONTACT::SolvingStrategy::nitsche)
     FOUR_C_THROW("Parallel redistribution not yet implemented for TSI problems");
+
+  if (const int solverNumber = contact.get<int>("LINEAR_SOLVER");
+      solverNumber != -1 &&
+      Teuchos::getIntegralValue<CONTACT::SolvingStrategy>(contact, "STRATEGY") ==
+          CONTACT::SolvingStrategy::lagmult &&
+      Teuchos::getIntegralValue<CONTACT::SystemType>(contact, "SYSTEM") !=
+          CONTACT::SystemType::condensed &&
+      Teuchos::getIntegralValue<CONTACT::SystemType>(contact, "SYSTEM") !=
+          CONTACT::SystemType::condensed_lagmult &&
+      Global::Problem::instance()
+              ->solver_params(solverNumber)
+              .get<Core::LinearSolver::PreconditionerType>("AZPREC") ==
+          Core::LinearSolver::PreconditionerType::multigrid_muelu &&
+      Teuchos::getIntegralValue<Mortar::ParallelRedist>(
+          mortarParallelRedistParams, "PARALLEL_REDIST") != Mortar::ParallelRedist::redist_none)
+    FOUR_C_THROW(
+        "Parallel redistribution is currently not supported for CONTACT problems in saddle-point "
+        "formulation preconditioned with MueLu");
 
   // ---------------------------------------------------------------------
   // adhesive contact
--- a/tests/input_files/meshtying3D_sp_simple.xml
+++ b/tests/input_files/meshtying3D_sp_simple.xml
@@ -177,7 +177,7 @@
   <!-- Definition of the multigrid preconditioner -->
   <ParameterList name="Hierarchy">
     <Parameter name="max levels" type="int" value="3"/>
-    <Parameter name="coarse: max size" type="int" value="100"/>
+    <Parameter name="coarse: max size" type="int" value="800"/>
     <Parameter name="verbosity" type="string" value="Extreme"/>
     <Parameter name="use kokkos refactor" type="bool" value="false"/>
     <ParameterList name="AllLevel">
--- a/tests/input_files/meshtying3D_sp_simple_nbinterfaceagg.xml
+++ b/tests/input_files/meshtying3D_sp_simple_nbinterfaceagg.xml
@@ -172,7 +172,7 @@
   <!-- Definition of the multigrid preconditioner -->
   <ParameterList name="Hierarchy">
     <Parameter name="max levels" type="int" value="3"/>
-    <Parameter name="coarse: max size" type="int" value="800"/>
+    <Parameter name="coarse: max size" type="int" value="100"/>
     <Parameter name="verbosity" type="string" value="Extreme"/>
     <Parameter name="use kokkos refactor" type="bool" value="false"/>
     <ParameterList name="AllLevel">


src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
```diff
@@ -138,6 +138,67 @@ Core::LinAlg::SolverParams NOX::Nln::CONTACT::LinearSystem::set_solver_options(
       else
         FOUR_C_THROW("Currently we support only a pure meshtying OR a pure contact problem!");
 
+      // construct the mapping of the dual node IDs to primal node IDs
```
mayrmt: Just some more context, so that we know, why we need this information:
```suggestion
      // construct the mapping of the dual node IDs to primal node IDs for Lagrange multiplier aggregation
```

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
```diff
@@ -138,6 +138,67 @@ Core::LinAlg::SolverParams NOX::Nln::CONTACT::LinearSystem::set_solver_options(
...
+      std::shared_ptr<std::map<int, int>> dual2primal_map = std::make_shared<std::map<int, int>>();
+
+      const std::shared_ptr<const Mortar::StrategyBase>& strategy =
+          std::dynamic_pointer_cast<const Mortar::StrategyBase>(
+              Core::Utils::shared_ptr_from_ref(*i_constr_prec_.begin()->second));
+      const auto& gs_node_row_map = strategy->slave_row_nodes_ptr();
```
mayrmt: Why not `slave_node_row_map` instead of `gs_node_row_map`? What does `gs` stand for? (I know, we use `gs` at several places throughout the contact code, though I do not really know what it stands for.)

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
```diff
@@ -138,6 +138,67 @@ Core::LinAlg::SolverParams NOX::Nln::CONTACT::LinearSystem::set_solver_options(
...
+          // compute the nullspace vectors for the Lagrange multiplier field for MueLu
+          if (solverPtr->params().isSublist("MueLu Parameters"))
+          {
+            const int dim_nullspace = discret->n_dim();
+
+            auto block_mat_blocked_operator =
+                std::dynamic_pointer_cast<const Core::LinAlg::BlockSparseMatrixBase>(
+                    get_jacobian_operator());
+            if (!block_mat_blocked_operator)
+              FOUR_C_THROW("Failed to cast blockMat to BlockSparseMatrixBase");
```
mayrmt: This could be a check in debug only:
```suggestion
            FOUR_C_ASSERT (block_mat_blocked_operator != nullptr, "Failed to cast blockMat to BlockSparseMatrixBase");
```

src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_muelu.cpp
```diff
@@ -193,13 +193,14 @@ void Core::LinearSolver::MueLuPreconditioner::setup(
 
     if (muelulist_.sublist("Belos Parameters").isParameter("contact slaveDofMap"))
     {
-      Teuchos::RCP<Epetra_Map> ep_slave_dof_map =
+      const auto ep_slave_dof_map =
```
mayrmt: `ep_` refers to Epetra. That's not accurate anymore.

```suggestion
      const auto slave_dof_map =
```

src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_muelu.cpp
```diff
@@ -193,13 +193,14 @@ void Core::LinearSolver::MueLuPreconditioner::setup(
...
           muelulist_.sublist("Belos Parameters")
-              .get<Teuchos::RCP<Epetra_Map>>("contact slaveDofMap");
+              .get<Teuchos::RCP<Core::LinAlg::Map>>("contact slaveDofMap");
 
       if (ep_slave_dof_map.is_null()) FOUR_C_THROW("Interface contact map is not available!");
 
-      Teuchos::RCP<EpetraMap> x_slave_dof_map = Teuchos::make_rcp<EpetraMap>(ep_slave_dof_map);
+      Teuchos::RCP<EpetraMap> x_slave_dof_map =
```
mayrmt: The `x_` use to refer to Xpetra, though that is not the case here. `ep_` sounds better... Ideally, we would convert from `Core::LinAlg::Map` directly to `Xpetra::Map`, though that is not possible I guess.

tests/input_files/contact3D_sp_uzawa_nbinterfaceagg.xml
```diff
@@ -0,0 +1,191 @@
+<!-- Define MueLu preconditioner --><ParameterList name="MueLu">
```
mayrmt: Formatting looks weird, though this might actually be conforming with our style, so probably nothing we can do about it.
```suggestion
<!-- Define MueLu preconditioner -->
<ParameterList name="MueLu">
```
On could delete the first comment line.

tests/input_files/contact3D_sp_uzawa_nbinterfaceagg.xml
```diff
@@ -0,0 +1,191 @@
...
+      <Parameter name="factory" type="string" value="SubBlockAFactory"/>
+      <Parameter name="block row" type="int" value="1"/>
+      <Parameter name="block col" type="int" value="1"/>
+    </ParameterList>
+    <!-- for submatrix A_{01}) -->
+    <ParameterList name="mySubBlockAFactoryOffDiagonal">
+      <Parameter name="factory" type="string" value="SubBlockAFactory"/>
+      <Parameter name="block row" type="int" value="0"/>
+      <Parameter name="block col" type="int" value="1"/>
+    </ParameterList>
```
mayrmt: Does the node-based aggregation strategy require this off-diagonal block?

tests/input_files/contact3D_sp_std_simple_new_struct.4C.yaml
davidrudlstorfer, line 1: Would it be possible to provide a smaller testing file? This contains nearly 10000 elements and only the test files are more than 45 thousand lines. A smaller example would be really nice for future maintainability

davidrudlstorfer: I've missed that there's even a third file and in total it's ~23,000*3=69,000 lines of reference files. That's a lot for an end to end test

maxfirmbach: @davidrudlstorfer @PhilipOesterlePekrun I agree, even with a smaller file you should be able to generate e.g. 3 levels in the multigrid hierarchy.

src/core/linear_solver/src/preconditioner/4C_linear_solver_preconditioner_muelu.cpp
maxfirmbach: Yes, there is a reason why we don't provide direct conversion from `LinAlg` to `Xpetra` though, as the later will disappear at some point from `4C`. I would prefer a naming scheme that does not lean too much to naming packages explicitly, in this part of the code it is ok though.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
```diff
@@ -121,14 +125,10 @@ Core::LinAlg::SolverParams NOX::Nln::CONTACT::LinearSystem::set_solver_options(
       // (3) activeDofMap
       std::vector<Teuchos::RCP<Core::LinAlg::Map>> prec_maps(4, Teuchos::null);
```
maxfirmbach: I would prefer if you could use `std::shared_ptr` instead of `Teuchos::RCP` in the construction. This will result in removing `Teuchos::RCP` downstream.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
```diff
@@ -138,6 +138,67 @@ Core::LinAlg::SolverParams NOX::Nln::CONTACT::LinearSystem::set_solver_options(
       else
         FOUR_C_THROW("Currently we support only a pure meshtying OR a pure contact problem!");
 
+      // construct the mapping of the dual node IDs to primal node IDs
+      std::shared_ptr<std::map<int, int>> dual2primal_map = std::make_shared<std::map<int, int>>();
+
+      const std::shared_ptr<const Mortar::StrategyBase>& strategy =
```
maxfirmbach: Does this need to be so complicated?

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
```diff
@@ -138,6 +138,67 @@ Core::LinAlg::SolverParams NOX::Nln::CONTACT::LinearSystem::set_solver_options(
       else
         FOUR_C_THROW("Currently we support only a pure meshtying OR a pure contact problem!");
 
+      // construct the mapping of the dual node IDs to primal node IDs
+      std::shared_ptr<std::map<int, int>> dual2primal_map = std::make_shared<std::map<int, int>>();
```
maxfirmbach: Why shared here? You could just create the object here and make it shared later, when giving it to the parameter list.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
```diff
@@ -138,6 +138,67 @@ Core::LinAlg::SolverParams NOX::Nln::CONTACT::LinearSystem::set_solver_options(
...
+
+      for (int dual_lid = 0; dual_lid < gs_node_row_map->num_my_elements(); dual_lid++)
+      {
+        const int dual_gid = gs_node_row_map->gid(dual_lid);
+        if (discret->have_global_node(dual_gid))
+          (*dual2primal_map)[dual_lid] = solid_node_map->lid(dual_gid);
+      }
+
+      mueluParams.set<Teuchos::RCP<std::map<int, int>>>(
+          "Interface DualNodeID to PrimalNodeID", Teuchos::rcp(dual2primal_map));
```
maxfirmbach: Would prefer `std::shared_ptr` here again. `Teuchos` is fine inside the actual preconditioner interface.

src/contact/src/4C_contact_nox_nln_meshtying_linearsystem.cpp
```diff
@@ -118,14 +122,10 @@ Core::LinAlg::SolverParams NOX::Nln::MeshTying::LinearSystem::set_solver_options
       // (3) activeDofMap
       std::vector<Teuchos::RCP<Core::LinAlg::Map>> prec_maps(4, Teuchos::null);
```
maxfirmbach: My comments from above also apply to here.

maxfirmbach:
@PhilipOesterlePekrun Thanks for working on this! I only have a few minor comments.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
```diff
@@ -19,6 +21,8 @@
 #include "4C_solver_nonlin_nox_interface_required.hpp"
 #include "4C_solver_nonlin_nox_vector.hpp"
 
+#include <Teuchos_RCPStdSharedPtrConversions.hpp>
```
maxfirmbach: Can we avoid the inclusion of this header?

PhilipOesterlePekrun:
> @PhilipOesterlePekrun thanks for looking into this!
> 
> Regarding the duplication of `set_solver_options()` in mesh tying and contact, do you see options to avoid that? Do both maybe share a base class, that could take care of this? Or can we move the actual implementation to a free function, that is then called by both physics modules, so that we implement the actual computation of the solver parameters only once?

They don't share a contact-specific base class, so a free function somewhere would probably be best, I'll try that.

tests/input_files/contact3D_sp_std_simple_new_struct.4C.yaml
PhilipOesterlePekrun: No problem, I'll find a smaller discretization. I just made copies of the other test files with one or two changes. I suppose it would be easier if we had the mesh in independent files, but a smaller discretization is also no issue.

mayrmt:
@PhilipOesterlePekrun 

You could put the free function into `src/contact/src/4C_contact_utils.{hpp/cpp}` for example.

PhilipOesterlePekrun:
@mayrmt Yes, I'll probably do that. Overall, the changes are pretty easy, but I'll probably address everything after my next exam on Friday if that's ok (also hope it's ok to mention this on Github, though I know it's more personal clutter)

mayrmt:
@PhilipOesterlePekrun Sure! Feel free to work on this according to your personal time line! No pressure from our side.

tests/input_files/contact3D_sp_std_simple_new_struct.4C.yaml
PhilipOesterlePekrun: Is it not possible to use the same exodus (.e) file for different .yaml files? Or can we not have exodus files in the input_files/ ?

mayrmt: Reusing exodus files across yaml files should be possible.

PhilipOesterlePekrun: Sorry, I guess my question was more whether I am allowed to do this in the testing directory. I don't see this done for any of the other input files, even though it would save space i.e. make overall test directory smaller.

Mainly, though, I want to use a smaller discretization, but pre_exodus no longer exists. If it is preferable to use native 4C ASCII discretization in a single file, I guess I can build a version of 4C from before pre_exodus was removed.

mayrmt: I am fine with adding exodus files to the testing directories. 

The feature of reading exodus files is rather new, so I guess, nobody has added tests with it so far.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
PhilipOesterlePekrun: `CONTACT::MtAbstractStrategy::collect_maps_for_preconditioner()` in the same file just above already does this (does the same thing as `fill_maps_for_preconditioner()`, just with `shared_ptr`s), but it is labelled as deprecated, though I don't know why.

tests/input_files/contact3D_sp_uzawa_nbinterfaceagg.xml
PhilipOesterlePekrun: It gets formatted back to this by clang-format.

tests/input_files/contact3D_sp_uzawa_nbinterfaceagg.xml
PhilipOesterlePekrun: Indeed not needed. Fixed it.

tests/input_files/contact3D_sp_std_simple_new_struct.4C.yaml
PhilipOesterlePekrun: I just ended up finding some existing files with geometry that is small and works for the new tests.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
mayrmt: It is labeled as deprecated, because this relates to the old time integration. If you see a clean way to re-use it, feel free to do so.

PhilipOesterlePekrun:
Ok both the meshtying and contact class do call some member variables/functions actually, so a free function in `Utils` would not be very elegant. I think a better way to remove the code duplication would be to have the class `CONTACT::LinearSystem` inherit from `MeshTying::LinearSystem`. The constructor of `CONTACT::LinearSystem` also just adds one thing to the initializer list but is otherwise the same, and `throw_error()` is the same. `get_active_lin_solver()` could be made virtual in `MeshTying::LinearSystem` and then overridden.

I guess the counterargument would be that `MeshTying::LinearSystem` is not conceptually a base class of `CONTACT::LinearSystem` though (and also, we often do it the other way around in other areas of 4C--make contact a base class of mesh tying). What do you think @mayrmt ?

mayrmt:
@PhilipOesterlePekrun I am not convinced that adding more inheritance is the right approach here. We do have lots of code duplication already between meshtying and contact, so I'd rather proceed with this PR to move forward with the migration to the new time integration.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
PhilipOesterlePekrun: These two functions are essentially the same, and not functionally related to either new or old structural timint. They are also used nowhere else, so ~~we could just delete the one with Teuchos::RCP if we want to move toward using std::shared_ptr instead of Teuchos::RCP.~~

PhilipOesterlePekrun: Actually, I see the structure now. `fill_maps_for_preconditioner(vector<RCP...>)` comes from the base class `class NOX::Nln::CONSTRAINT::Interface::Preconditioner`, while `collect_maps_for_preconditioner(shared_ptr...)` comes from `class MORTAR::StrategyBase : public NOX::Nln::CONSTRAINT::Interface::Preconditioner`.

So I guess the question is whether there is a reason why `fill_maps_for_preconditioner(vector<RCP...>)` is a function of Teuchos::RCP and not of std::shared_ptr. Otherwise, we could just use this with shared_ptrs and delete the "old" `collect_maps_for_preconditioner(shared_ptr...)`. Neither of the functions are actually used anywhere else, so it wouldn't break anything.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
PhilipOesterlePekrun: Fied it, though the cast is necessary.

PhilipOesterlePekrun:
@mayrmt Yeah, makes sense. I tried making a portion of it into a free function but realized that a member function is called in the innermost loop, so it would not be easy/clear.

All the other comments should be addressed with the newest commit.

tests/input_files/contact3D_sp_uzawa_nbinterfaceagg.xml
maxfirmbach: Maybe due to the comment before? Actually most of these comments are also not necessary, I think its quite clear that we want to define a preconditioner in this file 😄.

maxfirmbach (dismissed):
I'm fine with the current state. We can think about doing further cleanup in a separate PR in my opinion.

@PhilipOesterlePekrun Thanks again for cleaning this up.

mayrmt (dismissed):

mayrmt:
@PhilipOesterlePekrun For the failing test, feel free to just adapt the number of expected iterations.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
```diff
@@ -114,30 +116,85 @@ Core::LinAlg::SolverParams NOX::Nln::CONTACT::LinearSystem::set_solver_options(
...
+
+      if (strategy.is_saddle_point_system())
+      {
+        const auto& sol_type = Teuchos::getIntegralValue<FourC::CONTACT::SolvingStrategy>(
+            strategy.params(), "STRATEGY");
+        const auto& systype =
+            Teuchos::getIntegralValue<FourC::CONTACT::SystemType>(strategy.params(), "SYSTEM");
+        if (sol_type == FourC::CONTACT::SolvingStrategy::lagmult &&
+            (systype != FourC::CONTACT::SystemType::condensed &&
+                systype != FourC::CONTACT::SystemType::condensed_lagmult))
```
mayrmt: Are these two conditions needed? The check for saddle-point system above should make these two obsolete, right?

PhilipOesterlePekrun:
I will change the expected iterations (which is due to a change in "coarse: max size" to allow the smaller examples to have 3 levels too), but I'm more concerned that `meshtying3D_sp_std_simple_new_struct_nbinterfaceagg.4C.yaml-p3` with `localNumDualNodes != Teuchos::as<LocalOrdinal>(mapNodesDualToPrimal->size())`. The current code works with the other examples, but somehow here the dual2Primal map is incorrect in parallel. I'm trying to figure out why.

src/contact/src/4C_contact_nox_nln_contact_linearsystem.cpp
PhilipOesterlePekrun: It was done like this in the old structure, but I can remove it.

PhilipOesterlePekrun:
Ok, I have found the root cause of the issue. All of the other MueLu+contact/meshtying test files have
```yaml
MORTAR COUPLING/PARALLEL REDISTRIBUTION:
  PARALLEL_REDIST: "None"
```
But the test that is now failing has the default of `PARALLEL_REDIST: "Static"`. I thought that reconstructing the dual2primal map in each nonlinear solver iteration would ensure that something like parallel redistribution is handled correctly.

**So the issue is that MueLu+contact/meshtying does not work with parallel redistribution.**

Here are `solid_node_map` and `slave_node_row_map` for the test with 2 processors (`PARALLEL_REDIST: "Static"`):
[mapsPrintout.txt](https://github.com/user-attachments/files/25368358/mapsPrintout.txt)

The `slave_node_row_map` shows the primal node which corresponds to the dual node (lid). So, for example, the slave node corresponding to the primal node with gid 182 is owned by proc1 (line 627 of my .txt file). But, the actual primal node with gid 182 is owned by proc0 (see line 200).

This does not only affect the dual2primal map, but also the dof-based interface aggregation, for both contact and meshtying. For example, `meshtying3D_sp_std_braess_sarazin.4C.yaml`, `meshtying3D_sp_std_uzawa.4C.yaml`, `meshtying3D_sp_dual_simple.4C.yaml`, `contact3D_sp_std_simple.4C.yaml`, and `contact3D_sp_std_uzawa.4C.yaml` throw similar errors if you set `PARALLEL_REDIST: "Static"`. Dynamic also does not work (though this is explicitly stated for meshtying at least).

Does the parallel distribution not redistribute the slave nodes such that they match the primal nodes on the surface? Is this a known issue? Should we throw an earlier error?

@mayrmt @maxfirmbach

mayrmt:
@PhilipOesterlePekrun That is not surprising! We are aware of this limitation. I can explain it in detail in our next meeting. 

For now, just disable redistribution by setting it to `None`. Getting to work both options at the same time is beyond the scope of this PR.

PhilipOesterlePekrun:
@mayrmt Yes, I kind of realized that when the dof-based tests were failing too. It wasn't immediately clear to me what the issue was though--should we maybe add a throw closer to the input? (maybe similar to the error when using Dynamic redistribution with mesh tying, `Dynamic parallel redistribution not possible for meshtying`)

mayrmt:
@PhilipOesterlePekrun Sure, an error message helps to immediately identify the problem until we resolve the underlying issue.

tests/input_files/contact3D_sp_uzawa_nbinterfaceagg.xml
mayrmt: Let's just delete the first comment:
```suggestion
<ParameterList name="MueLu">
```

mayrmt:
@PhilipOesterlePekrun I have seen that you added an error to the contact and meshtying managers. Have you checked, that this error also throws for the new time integration? Somewhere in the back of my mind, I always connect the contact and meshtying managers with the old time integration (and some copied versions of them with the new time integration.) I might be wrong though.

PhilipOesterlePekrun:
@mayrmt Ah you're right, this is only for the old timint. Will amend shortly.

PhilipOesterlePekrun:
For new structure, I found the right place to put the throw was in `contact_meshtying_strategy_factory.cpp` and `contact_strategy_factory.cpp`. I think everything should be in place now.

mayrmt (approved):
@PhilipOesterlePekrun Looks good! Thanks for going some extra rounds on this PR. 

Please squash the commits into a meaningful set of commits before merging.

PR #1748, opened 2026-02-11
Enable optional vtk output of shell7p thickness director and thickness

mairehenke:
## Description and Context

This PR enables optional vtk output of the thickness and thickness director for shell7p elements.

Previously, only the thickness was available through the `post_processor`. With this change, we can also output the thickness director, providing the 3D state of the shell directly without the need to compute normals afterwards. The thickness (twice the norm of the director vector) can also be output. To support this, the element evaluation was updated so that the thickness director is computed at the beginning of the simulation. 

I also removed the thickness output via the post-processor completely. It is now only supported through vtk output.

### DIFF
--- a/src/shell7p/4C_shell7p_ele.cpp
+++ b/src/shell7p/4C_shell7p_ele.cpp
@@ -273,8 +273,6 @@ void Discret::Elements::Shell7p::set_params_interface_ptr(const Teuchos::Paramet
 
 void Discret::Elements::Shell7p::vis_names(std::map<std::string, int>& names)
 {
-  std::string result_thickness = "thickness";
-  names[result_thickness] = 1;
   solid_material()->vis_names(names);
 }  // vis_names()
 
@@ -283,8 +281,6 @@ bool Discret::Elements::Shell7p::vis_data(const std::string& name, std::vector<d
 {
   // Put the owner of this element into the file (use base class method for this)
   if (Core::Elements::Element::vis_data(name, data)) return true;
-
-  shell_interface_->vis_data(name, data);
 
   return solid_material()->vis_data(name, data, id());
 
--- a/src/shell7p/4C_shell7p_ele_calc.cpp
+++ b/src/shell7p/4C_shell7p_ele_calc.cpp
@@ -586,22 +586,6 @@ void Discret::Elements::Shell7pEleCalc<distype>::update(Core::Elements::Element&
   solid_material.update();
 }
 
-template <Core::FE::CellType distype>
-void Discret::Elements::Shell7pEleCalc<distype>::vis_data(
-    const std::string& name, std::vector<double>& data)
-{
-  auto thickness_gp = get_cur_thickness_director();
-  if (name == "thickness")
-  {
-    if (data.size() != 1) FOUR_C_THROW("size mismatch");
-    for (auto& thickness_data : thickness_gp)
-    {
-      data[0] += 2.0 * thickness_data.norm2();
-    }
-    data[0] = data[0] / intpoints_midsurface_.num_points();
-  }
-}  // vis_data()
-
 // template classes
 template class Discret::Elements::Shell7pEleCalc<Core::FE::CellType::quad4>;
 template class Discret::Elements::Shell7pEleCalc<Core::FE::CellType::quad8>;
--- a/src/shell7p/4C_shell7p_ele_calc.hpp
+++ b/src/shell7p/4C_shell7p_ele_calc.hpp
@@ -86,8 +86,6 @@ namespace Discret
       void reset_to_last_converged(
           Core::Elements::Element& ele, Mat::So3Material& solid_material) override;
 
-      void vis_data(const std::string& name, std::vector<double>& data) override;
-
       [[nodiscard]] const std::vector<Core::LinAlg::Matrix<3, 1>>& get_cur_thickness_director()
           const override
       {
--- a/src/shell7p/4C_shell7p_ele_calc_eas.cpp
+++ b/src/shell7p/4C_shell7p_ele_calc_eas.cpp
@@ -842,22 +842,6 @@ void Discret::Elements::Shell7pEleCalcEas<distype>::update(Core::Elements::Eleme
   solid_material.update();
 }
 
-template <Core::FE::CellType distype>
-void Discret::Elements::Shell7pEleCalcEas<distype>::vis_data(
-    const std::string& name, std::vector<double>& data)
-{
-  auto thickness_gp = get_cur_thickness_director();
-  if (name == "thickness")
-  {
-    if (data.size() != 1) FOUR_C_THROW("size mismatch");
-    for (auto& thickness_data : thickness_gp)
-    {
-      data[0] += 2.0 * thickness_data.norm2();
-    }
-    data[0] = data[0] / intpoints_midsurface_.num_points();
-  }
-}  // vis_data()
-
 // template classes
 template class Discret::Elements::Shell7pEleCalcEas<Core::FE::CellType::quad4>;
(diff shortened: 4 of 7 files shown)


tests/list_of_tests.cmake
```diff
@@ -2843,6 +2843,7 @@ four_c_test_vtk(poro_3D_hex8_new_struc-vtk-structure.4C.yaml poro_3D_hex8_new_st
...
 four_c_test_vtk(porofluid_pressure_based_elast_scatra_2D_quad4_nodetopoint_artery_airway_coupling_mono-vtk-artery.4C.yaml porofluid_pressure_based_elast_scatra_2D_quad4_nodetopoint_artery_airway_coupling_mono.4C.yaml 3 xxx-artery.pvd ref/porofluid_pressure_based_elast_scatra_2D_quad4_nodetopoint_artery_airway_coupling_mono-artery.pvd 1e-08 2.4)
 four_c_test_vtk(porofluid_pressure_based_elast_scatra_2D_quad4_nodetopoint_artery_airway_coupling_volfrac_closingrelation_lung_mono-vtk-porofluid.4C.yaml porofluid_pressure_based_elast_scatra_2D_quad4_nodetopoint_artery_airway_coupling_volfrac_closingrelation_lung_mono.4C.yaml 3 xxx-porofluid.pvd ref/porofluid_pressure_based_elast_scatra_2D_quad4_nodetopoint_artery_airway_coupling_volfrac_closingrelation_lung_mono-porofluid.pvd 1e-08 0.3)
+four_c_test_vtk(shell7p_cylinder_nonlinear_orthopressure_incompressible_quad4-vtk.4C.yaml shell7p_cylinder_nonlinear_orthopressure_incompressible_quad4.4C.yaml 3 xxx-structure.pvd ref/shell7p_cylinder_nonlinear_orthopressure_incompressible_quad4-vtk/structure.pvd 1e-08)
```
c-p-schmidt: With this change, you could remove all `structure-00000*` vtk reference files as the files are only checked for the time 1.0 and thereby save quite some reference files.
```suggestion
four_c_test_vtk(shell7p_cylinder_nonlinear_orthopressure_incompressible_quad4-vtk.4C.yaml shell7p_cylinder_nonlinear_orthopressure_incompressible_quad4.4C.yaml 3 xxx-structure.pvd ref/shell7p_cylinder_nonlinear_orthopressure_incompressible_quad4-vtk/structure.pvd 1e-08 1.0)
```

tests/input_files/shell7p_cylinder_nonlinear_orthopressure_incompressible_quad4.4C.yaml
```diff
@@ -16,12 +16,13 @@ IO:
   STRUCT_STRESS: "Cauchy"
   STRUCT_STRAIN: "GL"
 IO/RUNTIME VTK OUTPUT:
-  INTERVAL_STEPS: 1
+  INTERVAL_STEPS: 10
   OUTPUT_DATA_FORMAT: ascii
 IO/RUNTIME VTK OUTPUT/STRUCTURE:
   OUTPUT_STRUCTURE: true
   DISPLACEMENT: true
   STRESS_STRAIN: true
```
c-p-schmidt: You could also think about deactivating the STRESS_STRAIN output if this is not really what you are interested in testing with this test.

src/shell7p/4C_shell7p_ele_scatra_evaluate.cpp
```diff
@@ -149,6 +155,55 @@ int Discret::Elements::Shell7pScatra::evaluate(Teuchos::ParameterList& params,
           discretization, nodal_directors_, la[0].lm_, params);
     }
     break;
+    case Core::Elements::struct_calc_thickness:
```
c-p-schmidt, line 158: At first sight, this looks identical to the implementation in `src/shell7p/4C_shell7p_ele_evaluate.cpp`. Could you maybe unify this?

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
```diff
@@ -868,6 +868,24 @@ void Solid::ModelEvaluator::Structure::write_output_runtime_structure(
...
+        eval_data().get_opt_quantity_data_node_postprocessed(), Core::IO::OutputEntity::node,
+        context);
+  }
+  else if (structure_output_params.output_optional_quantity() ==
+           Inpar::Solid::optquantity_shell7pthicknessdirector)
+  {
+    // Write nodal shell7p thickness director vector
+    std::vector<std::optional<std::string>> context(3, "shell7p_thickness_director");
+    vtu_writer_ptr_->append_result_data_vector_with_context(
+        eval_data().get_opt_quantity_data_node_postprocessed(), Core::IO::OutputEntity::node,
```
c-p-schmidt, line 889: Doesn't it always have the same size and map as the dofs?
```suggestion
        eval_data().get_opt_quantity_data_node_postprocessed(), Core::IO::OutputEntity::dof,
```

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
```diff
@@ -868,6 +868,24 @@ void Solid::ModelEvaluator::Structure::write_output_runtime_structure(
...
+    std::vector<std::optional<std::string>> context(1, "shell7p_thickness");
+    vtu_writer_ptr_->append_result_data_vector_with_context(
+        eval_data().get_opt_quantity_data_node_postprocessed(), Core::IO::OutputEntity::node,
+        context);
+  }
+  else if (structure_output_params.output_optional_quantity() ==
+           Inpar::Solid::optquantity_shell7pthicknessdirector)
+  {
+    // Write nodal shell7p thickness director vector
+    std::vector<std::optional<std::string>> context(3, "shell7p_thickness_director");
```
c-p-schmidt: Doesn't it always have the size of the dim?
```suggestion
    std::vector<std::optional<std::string>> context(global_state().get_dim(), "shell7p_thickness_director");
```

tests/list_of_tests.cmake
mairehenke: I would like to keep the initial output to verify the thickness director initialization, which is newly computed in this PR. For this purpose, I restricted the output to the time steps 0.0 and 1.0 only now.
Additionally, I am not entirely sure, but do we not usually write the initial output?

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
mairehenke: The maps are configured for node-based output. Changing the entity type to dof leads to the error
'dof_vector is neither in column nor in row map`, which further indicates that dof-based output is not supported by the current map setup.

mairehenke:
Thank you for your feedback, @c-p-schmidt! I had originally planned to remove the direct post-processor output later, but I agree that it makes sense to clean it up now. I have removed the post-processor for the thickness now

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
c-p-schmidt: You can keep it like that.

tests/list_of_tests.cmake
c-p-schmidt: Sounds reasonable to me, thank you for the explanation. This might also differ between different parts of 4C if we write the initial output, but I think it is fine as you do it here.

lkoeglmeier (approved):

mairehenke:
I have just incorporated the 'const' suggestion from Copilot. Could you approve again?

PR #1751, opened 2026-02-11
Enable coupled dof equation condition in 3d

m-frey:
## Description and Context
This PR enables PointLinearCoupledEquation constraints for 3D displacement coupling by adding support for the dispz degree of freedom. There are no additional test as the underlying functionality is already tested in 3d.


isteinbrecher (commented):
Looks good to me, can you add a small test case for this feature?

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
maxfirmbach: I think this error message also has to be adapted right?

maxfirmbach (changes requested):
@m-frey I just have a small remark. I would also like to see a test that explicitly uses the new feature e.g. ADD: "dispz" somewhere.

src/constraint_framework/4C_constraint_framework_input.cpp
```diff
@@ -110,8 +110,9 @@ void Constraints::set_valid_conditions(std::vector<Core::Conditions::ConditionDe
       Core::Conditions::PointLinearCoupledEquation, false, Core::Conditions::geometry_type_point);
 
   linear_ce.add_component(parameter<int>("EQUATION", {.description = "EQUATION"}));
```
ischeider: Could you please also change the description of EQUATION to something more useful? What kind of equation is possible, with which dependencies? See also comment to `COEFFICIENT`. 
If it refers to a function, then shouldn't we redefine the parameter to `FUNCT`, which is the name that we use in general? Even though it is a *breaking change* then.

src/constraint_framework/4C_constraint_framework_input.cpp
```diff
@@ -110,8 +110,9 @@ void Constraints::set_valid_conditions(std::vector<Core::Conditions::ConditionDe
...
-  linear_ce.add_component(deprecated_selection<std::string>("ADD", {"dispx", "dispy", "undefined"},
-      {.description = "degrees of freedom", .default_value = "undefined"}));
+  linear_ce.add_component(
+      deprecated_selection<std::string>("ADD", {"dispx", "dispy", "dispz", "undefined"},
+          {.description = "degrees of freedom", .default_value = "undefined"}));
   linear_ce.add_component(parameter<double>("COEFFICIENT"));
```
ischeider: The coefficient does not have a description at all. Since there is also EQUATION available, I don't know what an additional coefficient could do (maybe I fully misunderstand the parameters). If it is a factor added to the function/equation, I recommend to make it optional with 1 as a default value.

src/constraint_framework/4C_constraint_framework_input.cpp
ischeider:
```suggestion
      "Definition of the term of a linear couple equation coupling specified degrees of "
```

ischeider (commented):
Just asking for some more documentation ;-)

isteinbrecher (commented):
Can we add a test case for this feature?

maxfirmbach:
@m-frey This is in stale now. We should try merging it 😉.

src/constraint_framework/4C_constraint_framework_input.cpp
```diff
@@ -110,8 +110,9 @@ void Constraints::set_valid_conditions(std::vector<Core::Conditions::ConditionDe
       Core::Conditions::PointLinearCoupledEquation, false, Core::Conditions::geometry_type_point);
 
   linear_ce.add_component(parameter<int>("EQUATION", {.description = "EQUATION"}));
-  linear_ce.add_component(deprecated_selection<std::string>("ADD", {"dispx", "dispy", "undefined"},
-      {.description = "degrees of freedom", .default_value = "undefined"}));
+  linear_ce.add_component(
+      deprecated_selection<std::string>("ADD", {"dispx", "dispy", "dispz", "undefined"},
```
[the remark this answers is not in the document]
m-frey: ✔️

src/constraint_framework/4C_constraint_framework_input.cpp
m-frey: I added a better description for both.

src/constraint_framework/4C_constraint_framework_input.cpp
m-frey: Same as above. I hope the new description is clearer this time, let me know if anything is still unclear

src/constraint_framework/4C_constraint_framework_input.cpp
m-frey: ✔️

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
m-frey: ✔️

isteinbrecher (approved):

knarfnitram (approved):

PR #1523, opened 2025-11-21
[beaminteraction] Avoid binning based partitioning

maxfirmbach:
## Description and Context

This PR aims to avoid the binning based partitioning in the beaminteraction framework and proposes the option of building a monolithic graph and using graph partitioning instead.

The PR does the following:
- avoid everything related to a binning based discretization for the interaction pair evaluation for the beam-contact code path (this means we don't even create the binning discretization) and use a monolithic graph partitioning approach instead (this also holds true for writing and reading restarts)
- shows based on two test cases, that we are able to delete the `BINNING STRATEGY` section from the respective input files
- modifies a small subgroup of runtime vtk based tests as results change slightly there

PR #1752, opened 2026-02-11
[beaminteraction] Remove bruteforce search from beam contact and make `ArborX` a mandatory dependency for beam-contact-interaction

maxfirmbach:
## Description and Context

This PR adds a local interaction pair search based on `ArborX` for the potential calculations in the beaminteraction framework. In the same step the bruteforce search related code path is deleted and with it the respective input parameter.

With this PR `ArborX` is a mandatory dependency for beam-contact-interaction. All relevant tests are now requiring `ArborX` to be enabled. `ArborX` is still not mandatory in the global scope of `4C`.

Part of #1523

### DIFF
--- a/doc/documentation/src/installation/installation.rst
+++ b/doc/documentation/src/installation/installation.rst
@@ -88,6 +88,8 @@ See the `ArborX repository <https://github.com/arborx/ArborX>`_ for details and 
 
 Building |FOURC| with ArborX enabled automatically fetches the repository during the configure stage and later builds the library as dependency.
 
+Note: For contact based beaminteraction simulations ArborX is a mandatory dependency.
+
 .. _trilinos:
 
 **Trilinos**


tests/list_of_tests.cmake
isteinbrecher, line 1: Why is this file restructured so much?

src/beaminteraction/src/potential/4C_beaminteraction_potential_submodel_evaluator.cpp
```diff
@@ -659,45 +667,49 @@ void BeamInteraction::SubmodelEvaluator::BeamPotential::find_and_store_neighbori
...
+  {
+    // Check if the current element is relevant for beam-to-xxx contact.
+    Core::Elements::Element* currele = discret().l_col_element(colele_i);
 
-      // do not check on existence here -> shifted to GetBinContent
-      bin_strategy_ptr()->get_neighbor_and_own_bin_ids(*biniter, loc_neighboring_binIds);
+    other_bounding_boxes.emplace_back(std::make_pair(currele->id(),
+        currele->get_bounding_volume(discret(),
+            *beam_interaction_data_state_ptr()->get_dis_col_np(), *geometric_search_params_ptr_)));
+  }
```
isteinbrecher: I think for beam-potential interaction, we don't need he `other_bounding_boxes`, but can simply search `beam_bounding_boxes` against `beam_bounding_boxes`.

tests/list_of_tests.cmake
maxfirmbach: We usually group tests with the same dependency. That is happening here.

maxfirmbach:
@isteinbrecher @davidrudlstorfer @knarfnitram Please check if any of the failing tests is relevant to you and check why the test fails and if the new result is ok.

src/beaminteraction/src/potential/4C_beaminteraction_potential_submodel_evaluator.cpp
maxfirmbach: For me tests are failing if I use `beam_bounding_boxes` as primitives.

isteinbrecher: @davidrudlstorfer is the potential interaction framework also working for something other than beam-to-beam, e.g., beam-to-sphere?

isteinbrecher:
> @isteinbrecher @davidrudlstorfer @knarfnitram Please check if any of the failing tests is relevant to you and check why the test fails and if the new result is ok.

The one beam-to-solid test that is failing is only failing in the vtk comparison, which is to be expected if the search ordering changes.

src/beaminteraction/src/potential/4C_beaminteraction_potential_submodel_evaluator.cpp
maxfirmbach: It might be that beam-to-sphere is a thing there.

isteinbrecher: Then I think it is the best to keep this as is, but some comments in the code still have to be adapted, as they refer to the beam contact case.

davidrudlstorfer: Yes beam-potential is also available for beam-to-sphere potential interactions

davidrudlstorfer:
@maxfirmbach thanks a lot for working on this! 

I really need to check the evaluated pairs in detail compared to the old approach. The differences in the tests are really large (for example the energy is off by a factor of 2)

Also as I've seen in at least one test case the example does not even converge anymore.

@maxfirmbach @isteinbrecher I would now set up simple example and log out / print out all evaluated / found beam-beam pairs. Do you have any other ideas / suggestions I am missing right now?

src/beaminteraction/src/4C_beaminteraction_input.cpp
```diff
@@ -36,11 +36,7 @@ std::vector<Core::IO::InputSpec> BeamInteraction::valid_parameters()
                   {"everydt", repstr_everydt},
               },
               {.description = "Type of employed repartitioning strategy",
-                  .default_value = repstr_adaptive}),
-
-          parameter<SearchStrategy>("SEARCH_STRATEGY",
-              {.description = "Type of search strategy used for finding coupling pairs",
-                  .default_value = SearchStrategy::bruteforce_with_binning})},
```
davidrudlstorfer, line 43: Really nice that this top level parameter is gone :)

davidrudlstorfer:
And another question; what is actually happening when one builds 4C without ArborX and then runs beaminteraction simulation. Is there a nice error output that ArborX is not available or is it some cryptic error message? Is this checked somewhere?

maxfirmbach:
@davidrudlstorfer Yep, at runtime you will get an error "You need to configure ArborX for geometric bounding volume search operations ..." etc.

maxfirmbach:
After a short discussion with @knarfnitram offline, the reason for the failing tests is that some bounding volumes are too small by construction.
@isteinbrecher For a few test cases I had to artificially enlarge the bounding volume to make the test run through correctly. The actual values are pretty high for some cases though and make no sense to me. Anything we can do about it?

This is mostly related to beam-to-beam contact scenarios.

davidrudlstorfer:
@maxfirmbach it would make sense to write out the vtk tests in readable/ASCII form rather than binary output

maxfirmbach:
> @maxfirmbach it would make sense to write out the vtk tests in readable/ASCII form rather than binary output

Yeah, I just saw that :see_no_evil:.

maxfirmbach:
These vtk output tests are just pure pain ... :imp: The tests are now written in ASCII and only the relevant files are changed. The respective testing input files are changed to explicitly use ASCII based output.

maxfirmbach:
Ready for another review round!

tests/input_files/beam3r_herm2line3_backweuler_browndyn_vanderWaals_doublelengthspec_smallsepapprox_regularization_constextpol_twocrossedbeams.4C.yaml
```diff
@@ -39,6 +39,8 @@ STRUCT NOX/Printing:
 BINNING STRATEGY:
   BIN_SIZE_LOWER_BOUND: 2
   DOMAINBOUNDINGBOX: "-2 -2 -1 2 2 1"
+BOUNDINGVOLUME STRATEGY:
+  BEAM_RADIUS_EXTENSION_FACTOR: 50.0
```
davidrudlstorfer: What I still do not fully understand is this `BEAM_RADIUS_EXTENSION_FACTOR`

In beam potential we need to look for all interactions within a specific distance (see https://github.com/maxfirmbach/4C/blob/cc969734e2b0a9eabca2b0dd7f3b5e30f65c2b75/src/beaminteraction/src/potential/4C_beaminteraction_potential_input.cpp#L46)

If I understand everything correctly, this BEAM_RADIUS_EXTENSION_FACTOR is basically doing the same as the CUTOFF_RADIUS.

For example, in this input file there is no specified cutoff radius defined so all potential interactions are computed. Probably all pairs within this example are within 50x radius and that why it works here.

Would it make sense to use the CUTOFF_RADIUS from the beam potential input here? Can we hand over that specific variable in the beam potential submodel evaluator?

maxfirmbach: Yes, and this would be for sure the cleaner way to do it. That is also the reason why I posed the question from above. Can you check at which point the extension is called in the potential interaction search and maybe replace it there with the cut-off radius? Hopefully this provides a robust base such that all pairs are found, instead of giving a wonky number.

src/beaminteraction/src/potential/4C_beaminteraction_potential_submodel_evaluator.cpp
```diff
@@ -659,45 +666,49 @@ void BeamInteraction::SubmodelEvaluator::BeamPotential::find_and_store_neighbori
...
   for (int rowele_i = 0; rowele_i < numroweles; ++rowele_i)
   {
     int const elegid = ele_type_map_extractor_ptr()->beam_map()->gid(rowele_i);
-    Core::Elements::Element* currele = discret_ptr()->g_element(elegid);
+    Core::Elements::Element* currele = discret().g_element(elegid);
 
-    // (unique) set of neighboring bins for all col bins assigned to current element
-    std::set<int> neighboring_binIds;
+    beam_bounding_boxes.emplace_back(std::make_pair(elegid,
+        currele->get_bounding_volume(discret(),
```
davidrudlstorfer: And one further question, here we use the bounding volume of the nodes. What does this mean for the beams? Is this an AABB? What happens if it is a straight line? Unfortunately, I have a lot questions with this :D

maxfirmbach: We use k-DOPs as bounding volumes. we can discuss this later.

src/core/geometric_search/src/4C_geometric_search_bounding_volume.hpp
```diff
@@ -118,6 +118,21 @@ namespace Core::GeometricSearch
       }
     }
 
+    /**
+     * @brief Uniformly extends the bounding volume along all k-DOP directions.
+     *
+     * @param value The value by which the bounding volume is extended to in both the negative and
+     * positive direction for each k-DOP axis.
+     */
+    inline void extend_boundary_to(const double value)
```
isteinbrecher: Why do we need this new function, I think `extend_boundaries` would work as well.

davidrudlstorfer:
@isteinbrecher in the last meeting with @maxfirmbach it was discussed that we need some alternative way to find all interacting pairs with an infinite search radius, i.e., bruteforce search between all interacting partners. I do not have any knowledge in ArborX but @maxfirmbach was not aware of any possibility to do that (@maxfirmbach please correct me if I am wrong here)

src/core/geometric_search/src/4C_geometric_search_bounding_volume.hpp
davidrudlstorfer: the problem was that extend_boundaries needs the radius if i remember correctly

isteinbrecher: I had a look and `extend_boundary_to` is actually wrong, as it does not account for the scaling of the k-DOP directions. I think we can remove this and directly use `extend_boundaries` instead.

isteinbrecher:
@davidrudlstorfer thanks for the update. I don't fully understand yet why we would need to find all-vs-all pairs. Is this part of the potential formulation, that all beams interact with each other? If so, then how does that currently work, since at the moment, only the pairs between neighbouring bins will be found.

If we really need all-vs-all (globally) in the potential interaction, then we will likely also have to adapt ghosting, such that all beams are available on all processors.

davidrudlstorfer:
@isteinbrecher in the potential interaction framework one wants to evaluate all beam to beam pairs. Only if the interaction radius is set smaller the furthest pairs can be discarded.

I thought that we currently ghost all elements on all procs anyways? And that's why this worked right now?

isteinbrecher:
> I thought that we currently ghost all elements on all procs anyways? And that's why this worked right now?

No, we currently only ghost neighbouring bins on each rank (the bin size is governed by the cutoff radius). Thus you will in general not have all elements available on all ranks.

Currently, I think that potential interaction simulations search all vs. all in neighbouring bins - which is highly dependent on the parallel distribution. This PR does not change the parallel distribution, only the way that local pairs are searched against each other (thus this works with the extended boundary volumes).

Do your evaluated pairs (that actually give a force/stiffness contribution) change if you increase the binning cutoff radius?

This PR is intended as a preliminary to #1523. #1523 will change the parallel distribution, so likely also potential results.
Maybe it makes sense to first only port the beam contact submodel evaluator to ArborX and then later on the potential submodel evaluator once we exactly know what pairs are actually needed there.

tests/input_files/beam3eb_static_contact_penalty_newgap_twocrossedbeams.4C.yaml
```diff
@@ -46,7 +46,7 @@ BEAM INTERACTION/BEAM TO BEAM CONTACT CONDITIONS:
     COUPLING_ID: 1
 BOUNDINGVOLUME STRATEGY:
   WRITE_GEOMETRIC_SEARCH_VISUALIZATION: true
-  BEAM_RADIUS_EXTENSION_FACTOR: 12
+  BEAM_RADIUS_EXTENSION_FACTOR: 350
```
isteinbrecher, line 49: It is not surprising that the bounding volume has to be increased by this much here, as this test case is intended to check the new gap functionality, with very thin beams. Thus, we have to increase the bounding volume by this high amount.

In the future, we can think if we want to add a parameter that extends the beam bounding volumes by a constant value, e.g., `BEAM_RADIUS_EXTENSION_CONSTANT`, the bounding volume offset would then be calculated via
```
offset = BEAM_RADIUS_EXTENSION_CONSTANT + radius * BEAM_RADIUS_EXTENSION_FACTOR
```
This would allow us to specify an extension value that is not proportional to the beams radius.

tests/input_files/ref/beam3r_herm2line3_static_beam_to_solid_surface_coupling_mortar_reference_forced_line3-vtk/xxx-vtk-files/beam-to-solid-surface-mortar-00003.pvtu
```diff
@@ -14,7 +14,5 @@
       <PDataArray type="Int32" Name="uid_1_pair_solid_id" format="ascii"/>
     </PPointData>
     <Piece Source="beam-to-solid-surface-mortar-00003-0.vtu"/>
-    <Piece Source="beam-to-solid-surface-mortar-00003-1.vtu"/>
-    <Piece Source="beam-to-solid-surface-mortar-00003-2.vtu"/>
```
isteinbrecher, line 18: That is the only thing I don't get, why these two file were in this file and now have to be removed. Is it possible that we used to run this test with 3 cores and now run it with 1? (There are also some other occurrences in this PR).

maxfirmbach: I did not change the test setup etc. I just redid some reference results.  I think the testing might have change or the number of procs used before.

isteinbrecher: I think I checked in the parallel result back in the day, so your changes are fine - the only think, we should then remove the files `beam-to-solid-surface-mortar-00003-1.vtu` and `beam-to-solid-surface-mortar-00003-2.vtu` from the repo (and in the other cases where we remove files here).

isteinbrecher:
@maxfirmbach I had a look and found a small bug, which I fixed in an additional commit.

I added the reference results again created with 3 processors, and we compare them to the test results with 1 processors. On my machine it works as expected. Let's see if the pipeline passes. From my point of view this can be merged.

davidrudlstorfer:
> No, we currently only ghost neighbouring bins on each rank (the bin size is governed by the cutoff radius). Thus you will in general not have all elements available on all ranks.
> 
> Currently, I think that potential interaction simulations search all vs. all in neighbouring bins - which is highly dependent on the parallel distribution. This PR does not change the parallel distribution, only the way that local pairs are searched against each other (thus this works with the extended boundary volumes).
> 
> Do your evaluated pairs (that actually give a force/stiffness contribution) change if you increase the binning cutoff radius?
> 
> This PR is intended as a preliminary to https://github.com/4C-multiphysics/4C/pull/1523. https://github.com/4C-multiphysics/4C/pull/1523 will change the parallel distribution, so likely also potential results.
> Maybe it makes sense to first only port the beam contact submodel evaluator to ArborX and then later on the potential submodel evaluator once we exactly know what pairs are actually needed there.

@isteinbrecher thanks for the information, I did not know that we do not ghost everything. Then this is already questionable in its current form.

And yes, the evaluated pairs change if the binning cutoff radius is increased/decreased.

Looks good from my side 🚀

tests/list_of_tests.cmake
```diff
@@ -70,159 +45,30 @@ four_c_test(TEST_FILE beam3r_herm2line3_backweuler_browndyn_crosslinking_beam3rl
 four_c_test_restart(BASED_ON ${current} SAME_FILE NP 3 RESTART_STEP 30)
 four_c_test(TEST_FILE beam3r_herm2line3_backweuler_browndyn_randcrosslinking_beam3rline2.4C.yaml NP 2)
 four_c_test(TEST_FILE beam3r_herm2line3_backweuler_browndyn_randcrosslinking_beam3rline2_twobonds.4C.yaml NP 3)
-four_c_test(TEST_FILE beam3r_herm2line3_backweuler_browndyn_repLJ_singlelengthspec_smallsepapprox_twocrossedbeams.4C.yaml NP 2)
 four_c_test(TEST_FILE beam3r_herm2line3_backweuler_browndyn_singlefil.4C.yaml NP 3)
 four_c_test(TEST_FILE beam3r_herm2line3_backweuler_browndyn_singlefil_ptc_elementbased_everydt.4C.yaml NP 2)
 four_c_test(TEST_FILE beam3r_herm2line3_backweuler_browndyn_singlefil_ptc_elementbased_everyiter.4C.yaml NP 2)
-four_c_test(TEST_FILE beam3r_herm2line3_backweuler_browndyn_vanderWaals_doublelengthspec_smallsepapprox_regularization_constextpol_twocrossedbeams.4C.yaml NP 2)
```
davidrudlstorfer, line 77: Now also the beam potential tests are moved into the ArborX guard - I have no problem with that, just FYI

davidrudlstorfer (dismissed):

tests/input_files/beam3eb_static_contact_penalty_newgap_twocrossedbeams.4C.yaml
davidrudlstorfer: I like the idea with the constant offset not depending on the radius.

tests/list_of_tests.cmake
maxfirmbach: Ah damn, I tried to move as many back as possible ... might need to have another close look ;).

ischeider:
Just a small remark: In the documentation it says "ArborX (optional)". Please, change it as well. In addition, I noticed by adding the ArborX support myself that this TPL is different from others in the list, since it is not necessary to install it beforehand. It would be good to have a note "(installed automatically during the 4C build process)"

maxfirmbach:
> Just a small remark: In the documentation it says "ArborX (optional)". Please, change it as well. In addition, I noticed by adding the ArborX support myself that this TPL is different from others in the list, since it is not necessary to install it beforehand. It would be good to have a note "(installed automatically during the 4C build process)"

Good point! I will update the documentation.

tests/list_of_tests.cmake
isteinbrecher: Good catch - @maxfirmbach can you move them back  to avoid confusion later on?

isteinbrecher:
> Just a small remark: In the documentation it says "ArborX (optional)". Please, change it as well. In addition, I noticed by adding the ArborX support myself that this TPL is different from others in the list, since it is not necessary to install it beforehand. It would be good to have a note "(installed automatically during the 4C build process)"

@ischeider I think there is some confusion here. With this PR, ArborX is still an optional dependency in the global scope, but now required for beam-to-beam/solid interaction/contact simulations. @maxfirmbach will update this accordingly.

ischeider:
> > Just a small remark: In the documentation it says "ArborX (optional)". Please, change it as well. In addition, I noticed by adding the ArborX support myself that this TPL is different from others in the list, since it is not necessary to install it beforehand. It would be good to have a note "(installed automatically during the 4C build process)"
> 
> @ischeider I think there is some confusion here. With this PR, ArborX is still an optional dependency in the global scope, but now required for beam-to-beam/solid interaction/contact simulations. @maxfirmbach will update this accordingly.

Ah, ok (the title sounds different - but maybe just to me)

maxfirmbach:
@ischeider I updated the title :).

isteinbrecher:
> Thanks for the information, I did not know that we do not ghost everything. Then this is already questionable in its current form.
> And yes, the evaluated pairs change if the binning cutoff radius is increased/decreased.

@davidrudlstorfer I am not super familiar with the potential framework, but from what I understand, it would make sense modify the formulation, such that it is not dependent on the binning options but rather on some pair internal cut-off radius. If you want to go that way, I would be happy to discuss possible ways to do so.

Positive effect of this:
- Your formulation becomes more deterministic
- One would not have to find all vs all pairs but only those who are within that radius, wich can easily be achieved with ArborX bounding volumes. Also #1523 should then work directly with the potential interactions.

davidrudlstorfer:
@isteinbrecher thanks for the offer for the discussion - I would appreciate it a lot.

Actually your idea with the cutoff radius is already implemented, for example see the input parameter: https://github.com/4C-multiphysics/4C/blob/59681e34144e775200c81a19ab3dc5471e44f1d9/src/beaminteraction/src/potential/4C_beaminteraction_potential_input.cpp#L46-L50

Actually what @maxfirmbach and me discussed is to use this cutoff value as the "bounding volume increase factor/value" but this then would need the constant value you proposed above rather than a radius dependent value like you proposed:

```cpp
offset = BEAM_RADIUS_EXTENSION_CONSTANT + radius * BEAM_RADIUS_EXTENSION_FACTOR
```

I am also using the cutoff radius most of the times in my simulations. The only time where I want to include all potential interactions are cases where I compare results to a theoretical/analytical result. Then, I really want to take all pair interactions of the whole setup into account.

tests/list_of_tests.cmake
maxfirmbach: I think the diff is shown a bit wrong here, the potential test already reside outside of the `ArborX` guard, its just shown in a weird way.

PR #1757, opened 2026-02-12
Refactor reduced lung discretization

bgoderbauer:
## Description and Context

TLDR: This PR implements the remaining basic input for the `ReducedLung` 0D-framework (discretization and boundary conditions), removing the last dependencies on the `RedAirway` module.

### Key changes:
- The lung tree input format is now aligned with the already ported structure of the models. Elements and nodes are loaded from `InputFields` in a new `Topology` section inside the `ReducedLungParameters`. 
- A new dummy discretization with generic line elements (#1761 :pray:) is created to interface 4C-infrastructure, like rebalancing and output writing. The difference from before is that now we don't create our models based on the discretization; instead, we create the discretization in line with the model setup, so there is only one input data structure, not two that need to be merged.
- Boundary conditions are no longer conditions of the discretization nodes; they are now also `InputFields` in `ReducedLungParameters`, where the corresponding dof of the system is encoded. Currently, we can directly set specific pressure and flow values at distinct dofs. More complicated boundary conditions should now also be easy to implement.
- All test files have been ported to the new input style, and a unittest checking all the different steps from an input file to model creation has been implemented.

### Points to discuss:
- I had to write a whole new element `ReducedLungLine` in order to create our own discretization. However, many of the functionalities I had to implement due to the virtual base class are never used. Around 150 lines of code for the element are unnecessary. I don't know, is there maybe a better way to do it? If I could wish for something, there would be a generic line element implementation somewhere central that already has the basic things I need to use rebalancing and output infrastructure. This still feels all around hacky, even though now it's more clearly hidden from a user that we have a discretization-like object that we only use for infrastructure. Resolved with #1761.
- What do you think about the input file design now, @maxiludwig? I kinda like that now the fields and varying values are all potentially in a separate file and the main input file can focus on the problem setup.


src/reduced_lung/src/4C_reduced_lung_main.cpp
```diff
@@ -38,26 +38,31 @@ namespace ReducedLung
...
+        *actdis, parameters.lung_tree.topology, rebalance_parameters);
+    actdis->fill_complete();
+    Core::LinAlg::Solver solver(problem->solver_params(parameters.dynamics.linear_solver),
+        actdis->get_comm(), problem->solver_params_callback(),
+        Teuchos::getIntegralValue<Core::IO::Verbositylevel>(problem->io_params(), "VERBOSITY"));
     compute_null_space_if_necessary(*actdis, solver.params());
     // Create runtime output writer
     Core::IO::DiscretizationVisualizationWriterMesh visualization_writer(
         actdis, Core::IO::visualization_parameters_factory(
-                    Global::Problem::instance()->io_params().sublist("RUNTIME VTK OUTPUT"),
```
bgoderbauer, line 59: This is now a complete setup here inside the main. I don't really know whether this is a good or correct way to approach this. I would be happy to hear some opinions/ideas!

tests/input_files/reduced_lung_3_aw_2_tu.4C.yaml
```diff
@@ -15,6 +15,13 @@ reduced_dimensional_lung:
     results_every: 1
     time_increment: 1
   lung_tree:
+    topology:
+      num_nodes: 6
+      num_elements: 5
```
maxiludwig, line 20: Do we actually need to write the number of nodes and elements explicitly or can they be read from the json file?

tests/input_files/reduced_lung_3_aw_2_tu.4C.yaml
```diff
@@ -42,68 +49,19 @@ reduced_dimensional_lung:
...
             constant: 1.0
-
-REDUCED DIMENSIONAL AIRWAYS DYNAMIC:
-  TIMESTEP: 1
-  NUMSTEP: 1
-  LINEAR_SOLVER: 1
+  boundary_conditions:
+    num_conditions: 3
+    bc_type:
+      constant: Pressure
```
maxiludwig, line 55: We would also want to have the possibility to change the type of the boundary condition during ventilation to simulate the switch from pressure controlled ventilation to volume controlled ventilation

tests/input_files/reduced_lung_3_aw_2_tu.4C.yaml
```diff
@@ -42,68 +49,19 @@ reduced_dimensional_lung:
         linear:
           elasticity_e:
             constant: 1.0
-
-REDUCED DIMENSIONAL AIRWAYS DYNAMIC:
-  TIMESTEP: 1
-  NUMSTEP: 1
-  LINEAR_SOLVER: 1
+  boundary_conditions:
+    num_conditions: 3
```
maxiludwig, line 53: Same as above. Do we need to set the number of conditions here?

tests/input_files/reduced_lung_3_aw_2_tu.4C.yaml
bgoderbauer: Sure, but I'd say this should be a different boundary condition that we should implement independently. In our current handling of boundary conditions, a switch from flow to pressure would mean that we have a different equation we need to solve. This is not that easy to achieve; we would need to "unfill" the Jacobian, clear the respective rows, fill the new entries, and call FillComplete() again. We can think about different BC handling, but in the current state, I would say we should have a distinct case for the switch between pressure and flow.

tests/input_files/reduced_lung_3_aw_2_tu.4C.yaml
bgoderbauer: I think yes, since we don't have an iterator over `InputField` entries (I'm not sure, please correct me if I'm wrong :smile:). We have some loops over all elements and all nodes during setup, without the numbers we don't know the size in advance

bgoderbauer: I wouldn't mind having these input parameters in the input file as well, but I understand that it feels quite unnecessary. However, they can give a hint about the problem size and whether everything was generated correctly, so I would prefer having them rather than not.

bgoderbauer:
I think this state is ready to be reviewed :smile:

src/reduced_lung/src/4C_reduced_lung_discretization_helpers.hpp
```diff
@@ -0,0 +1,143 @@
...
+{
+  namespace Elements
+  {
+    /**
+     * @brief Element type wrapper for the reduced lung dummy line element.
+     *
+     * Provides the minimal registration and null-space information required by the
+     * base class so ReducedLungLine can live inside a 4C discretization.
+     */
+    class ReducedLungLineType : public Core::Elements::ElementType
```
sebproell: As discussed offline, I am pretty sure you can use `PureGeometryElement` which is currently only used for testing. I think it is correct to enhance DiscretizationBuilder such that it allocates these dummy elements in case you don't need any custom ones. Let me open a PR.

bgoderbauer:
Many thanks @sebproell for #1761, this is very nice! I removed the dummy element specification `ReducedLungLine` again in the last commit, and it works perfectly! Now we can create the discretization in just a couple of lines of code, registering nodes and elements manually without the need for additional overridden virtual methods. I think it is very easy to see the role of discretization in this module now.

PR #1758, opened 2026-02-13
Remove SumIntoMyValue in Multivector class

dharinib98:
## Description and Context

Remove SumIntoMyValue Epetra call and replace it with a local extraction and update mechanism. 

#1462


maxfirmbach:
@dharinib98 Sorry for the naive question, but why do we need to replace the internals of this method?

As far as I know `Tpetra` provides
```cpp
void MultiVector<Scalar, LocalOrdinal, GlobalOrdinal, Node>::
sumIntoLocalValue(const LocalOrdinal lclRow, const size_t col, const impl_scalar_type& value, const bool atomic)
```

dharinib98:
No longer needed as suggested by @maxfirmbach

PR #1760, opened 2026-02-13
Refactor MAT_struct_thermoplasticlinelast

ischeider:
## Description and Context

There was an untested feature in the material `MAT_struct_ThermoPlasticLinElast`, namely a multilinear isotropic hardening. When I wanted to test it, I noticed several things

1. The feature did not work at all
2. The respective output routine is missing
3. The runtime structure vtu output of TSI problems does not work anyway

This PR shall

1. Change the routine so that a multilinear iostropic hardening can be used as intended
2. a test file is added so that the runtime output and the multilinear hardening are tested
3. The runtime structure vtu output is enabled, thanks to a patch provided by @tuchpaul (thanks for that!)
4. I added the the runtime vtu output also for the MAT_struct_PlasticLinElast for pure structural problems

## Interested parties
@tuchpaul


tuchpaul (dismissed):
LGTM! Thanks for the PR.

tests/input_files/ref/tsi_plastic_heating_monolithic-vtk/tsi_plastic_heating_monolithic-vtk-files/structure-00003-0.vtu
rjoussen, line 1: I recently was made aware that it would be nicer to have non-binary VTU files, in order to potentially see a diff in them if they ever need to be changed. But not sure if thats strictly necessary.

rjoussen (dismissed):
That's a nice addition! Also, last time I checked, the heat generation due to mechanical dissipation was not tested at all for 
these materials, because there wasn't a single tsi test where the yield stress was reached. This should also be fixed with this.

rjoussen:
> The runtime structure vtu output of TSI problems does not work anyway

What do you mean by that? I was hoping that the TSI runtime output would generally work now after #1676. Or do you mean the vtk output of Gauss point data specifically?

tests/input_files/ref/tsi_plastic_heating_monolithic-vtk/tsi_plastic_heating_monolithic-vtk-files/structure-00003-0.vtu
ischeider: For the two elements one might have a look and see but, anyway one can see the values in paraview. When I think of the discussion about binary node and element input (at this time I was also against it), I would say the benefit of non-binary output is even smaller. So I hesitate changing this unless you find some supporter ;-)

ischeider:
> What do you mean by that? I was hoping that the TSI runtime output would generally work now after https://github.com/4C-multiphysics/4C/pull/1676. Or do you mean the vtk output of Gauss point data specifically?

Even more specifically, I mean the vtk output of the gauss point data **of the structure results**. The error was `The element action struct_init_gauss_point_data_output is not yet implemented for the new solid-scatra elements`. As far as I know this was not fixed yet (or I missed some very recent PR).

ischeider:
@rjoussen :
> Also, last time I checked, the heat generation due to mechanical dissipation was not tested at all for
these materials, because there wasn't a single tsi test where the yield stress was reached. This should also be fixed with this.

The temperature increases by 24°C or so in the new test, solely by plastic deformation.

tests/input_files/ref/tsi_plastic_heating_monolithic-vtk/tsi_plastic_heating_monolithic-vtk-files/structure-00003-0.vtu
rjoussen: Okay, honestly I don't care that much, so it's fine with me :)

rjoussen:
> > What do you mean by that? I was hoping that the TSI runtime output would generally work now after #1676. Or do you mean the vtk output of Gauss point data specifically?
> 
> Even more specifically, I mean the vtk output of the gauss point data **of the structure results**. The error was `The element action struct_init_gauss_point_data_output is not yet implemented for the new solid-scatra elements`. As far as I know this was not fixed yet (or I missed some very recent PR).

That makes sense, thanks for the fix!

rjoussen:
> The temperature increases by 24°C or so in the new test, solely by plastic deformation.

Nice!

ischeider:
> > > What do you mean by that? I was hoping that the TSI runtime output would generally work now after #1676. Or do you mean the vtk output of Gauss point data specifically?
> > 
> > 
> > Even more specifically, I mean the vtk output of the gauss point data **of the structure results**. The error was `The element action struct_init_gauss_point_data_output is not yet implemented for the new solid-scatra elements`. As far as I know this was not fixed yet (or I missed some very recent PR).
> 
> That makes sense, thanks for the fix!

This was the part that @tuchpaul provided.

src/mat/4C_mat_thermoplasticlinelast.hpp
```diff
@@ -210,9 +210,9 @@ namespace Mat
 
     //! calculates the derivative of get_sigma_y_at_strainbarnp() w.r.t. astrain_{n+1}
     //! and returns the isotropic hardening modulus
-    double get_iso_hard_at_strainbarnp(
-        const double strainbar_p  //!< current accumulated strain
-                                  //!< or if damage!=0: isotropic hardening internal variable
+    std::pair<double, double> get_iso_hard_at_strainbarnp(
```
amgebauer: It would be nicer if you return a struct and give the parameters a name

tests/input_files/ref/tsi_plastic_heating_monolithic-vtk/tsi_plastic_heating_monolithic-vtk-files/structure-00003-0.vtu
amgebauer, line 1: Could you use ascii instead of binary? Other than that, you can also test all quantities in the result test if you want. That the output in vtu itself is working is already checked in other tests. This would reduce the number of reference files.

ischeider: Actually the check itself is not necessary, the results are checked sufficiently by `RESULT DESCRIPTION`. The routines that provide `RUNTIME VTK OUTPUT` (`register_output_variable_names` and `evaluate_output_data` ) are not tested I don't produce the output. Actually, these routines are not tested in other materials as well except by comparing vtu data. Should one write a unittest for these routines instead, or are they unimportant enough to skip the tests?

amgebauer: You can test all quantities from the `register_output_variable_names` and `evaluate_output_data` also in the result description so that you don't have to add an extra vtu test all the time. See for example https://github.com/4C-multiphysics/4C/blob/main/tests/input_files/mixture_growth_expl_remodel_fiber_no_basal_mass_production.4C.yaml#L186-L191

amgebauer: If you keep the vtu test, then please use ascii output so that it is easier to track changes if one has to adapt some results for whatever reason

ischeider: Okay, this makes sense, thanks for the hint. I'll change it.

ischeider: With the evaluationof internal variables within `RESULT_DESCRIPTION`, it is not necessary to compare the vtu data anymore. Since `IO/RUNTIME VTK OUTPUT/STRUCTURE` is necessary to get the data, the new feature of writing structure gauss point data in a tsi simulation is automatically tested.

Issue #1492, opened 2025-11-12
Refactor NURBS input format


isteinbrecher:
## Description

Currently the NURBS input is a very strange place in 4C. With the recent refactoring this has become a lot better, but still some issues remain. Most prominent, that we currently require NURBS elements to be in a specific ordering in the input file and discretisation to work.

Very simply speaking, each NURBS element is part of a NURBS patch that we define via `STRUCTURE KNOTVECTORS`. Currently, we require the NURBS elements to be in a specific ordering such that we can identify the location of an element inside a patch. This should be changes such that we define the position of the element in the patch in the element definition, e.g.,
```yaml
  - 3 SHELL_KIRCHHOFF_LOVE_NURBS NURBS9 5 6 7 9 10 11 13 14 15 MAT 1 GP 3 3 PATCH 1 PATCH_LOCATION 2 4
```
Here, `PATCH` refers to the id of the patch and `PATCH_LOCATION` specifies the location of the element inside the patch.

Furthermore, we currently only support NURBS of degree 1 or 2, i.e., we are not able to leverage the advantage of higher order continuity for more than $C^1$. The evaluation of NURBS shape functions for higher degrees is already possible in 4C, so the main issue that remains in this respect is a suitable definition of the input. Ideally, we should be able to specify arbitrary combinations of degrees (up to a certain level) without having an individual type for every combination - we should avoid `NURBSp1p1`, `NURBSp1p2`, `NURBSp2p1`, `NURBSp2p2`, ...
A suitable solution could look like this:
```yaml
  - 3 SHELL_KIRCHHOFF_LOVE_NURBS NURBS NURBS_DIM 2 NURBS_DEGREE 2 4 CP 5 6 7 9 10 11 13 14 15 MAT 1 GP 3 3 PATCH 1 PATCH_LOCATION 2 4
```
Here, `NURBS_DIM` describes how many degrees (and patch locations) we expect. `NURBS_DEGREE` then gives the degrees and `CP` gives the control point connectivity (the number of controls points depends on the degrees.

Another nice side effect of a changed NURBS input is that we can avoid the addition of „zero“ elements. In some cases we have to add elements that are never evaluated just to satisfy the expected ordering in the NURBS discretization.

## Possible Solution and Definition of Done
- [ ] Define a suitable input format
- [ ] Implement it
- [ ] Adapt the test cases

## Interested Parties
@eulovi @sebproell

mayrmt:
[...]

PR #1763, opened 2026-02-16
Allow to specify cell data for element creation when reading mesh

### LINKED ISSUE #1690: Allow to specify cell data for element creation when reading mesh
## Description
We can currently define meshes based on external formats via something like:
```yaml
STRUCTURE GEOMETRY:
  SHOW_INFO: "summary"
  ELEMENT_BLOCKS:
    - ID: 1
      SOLID:
        HEX8:
          MAT: 1
          KINEM: nonlinear
  FILE: vtu/solid_vtu_input.vtu
```

Certain elements in 4C, e.g., NURBS and beams, can't be defined with this approach, as they require element specific data. For example the shear deformable beams require an input like this:
```yaml
  - "3 BEAM3R LINE3 7 9 8 MAT 1 TRIADS 1.4821898202742558 -0.613943125568937 -0.613943125568937 1.4821898202742558 -0.613943125568937 -0.613943125568937 1.4821898202742558 -0.613943125568937 -0.613943125568937 HERMITE_CENTERLINE true"
```
The values for `TRIADS` are different for each element.

A suitable solution would be to allow
```yaml
STRUCTURE GEOMETRY:
  SHOW_INFO: "summary"
  ELEMENT_BLOCKS:
    - ID: 1
      SOLID:
        BEAM3R:
          MAT: 1
          HERMITE_CENTERLINE: true
          TRIADS:
            from_mesh: triads_from_mesh
  FILE: vtu/beam_vtu_input.vtu
```
where we provide a cell data filed named "triads_from_mesh" in the input mesh that provide the triad definition for each element.

Does anyone have an idea if something like this is already possible?

## Interested Parties
@davidrudlstorfer @sebproell @amgebauer @maxfirmbach

isteinbrecher:
## Description and Context

This PR adds the functionality to provide cell data in an external mesh format that can be used in the setup of the element. Previously we would only define constant cell data for a cell block. This allows to define beam elements with an external mesh format, as we have to define the nodal rotation vectors there for each element. This should also help to improve NURBS input in the future #1492.

Note: Currently in 4C, we directly convert vector valued mesh data fields to tensors. Until this is generalised, we resort to the "hack" that we use the symmetric tensor (6 unknowns) for 2-noded beam elements (6 rotation vector components) and the non-symmetric tensor (9 unknowns) for 3-noded beam elements (9 rotation vector components). These are the most commonly used beam elements, so this will allow to work on the beam input generation in the meantime.

Closes #1690


src/core/io/src/4C_io_mesh.hpp
```diff
@@ -485,6 +485,65 @@ namespace Core::IO::MeshInput
...
+   */
+  struct CellDataView
+  {
+    static constexpr unsigned dim = 3;
+
+    // Pointer to the cell data map.
+    const std::unordered_map<std::string, FieldDataVariantType<dim>>* cell_data = nullptr;
+
+    // Block local cell index for which the view is created.
+    std::size_t cell_index = 0;
```
amgebauer: I think it does not make sense to give this view as default an invalid state. The view-object should only be constructible in a valid state, so if you provide the cell data and the cell index.

In the read_element method, you can there pass the CellDataView as a pointer so show that this is optional and might be the nullpointer (if the mesh is read from the legacy yaml format).

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -485,6 +485,65 @@ namespace Core::IO::MeshInput
...
+            }
+            else
+            {
+              if (cell_index >= data.size())
+              {
+                FOUR_C_THROW(
+                    "Cell index {} is out of bounds for cell data with name '{}' which has "
+                    "size {}.",
+                    cell_index, cell_data_key, data.size());
+              }
```
amgebauer: This check can also be moved to the construction.

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -319,7 +319,8 @@ namespace
         linedef.fully_parse(parser, data);
 
         ele->set_node_ids_one_based_index(cell_info.nodal_ids);
-        ele->read_element(eletype, cell_info.cell_type_str, data);
+        ele->read_element(
+            eletype, cell_info.cell_type_str, data, Core::IO::MeshInput::CellDataView{});
```
amgebauer: see my comment above: Creating an empty cell data view is bad. Better, make it a pointer and pass the nullptr here so that it is clear in the element that the cell data is not passed.

amgebauer (commented):
I will not find the time this week to generalize the input to read arbitrary array-types in the vtu input. But, I think I can do it until end next week to avoid the hack of reading it in a (symmetric) tensor. Alternatively, you could also pass two or three names and read the rotation vector for each node separately.

Also, I think we should enable to have input fields ready already during element setup so that we can make use of it here. That would me a lot more elegant here. But, this is not trivial here for now, since we might need some larger refactoring of the mesh input (easy for vtu/gmsh/ensight input, but not trivial for the legacy yaml mesh input).

src/core/io/src/4C_io_mesh.hpp
isteinbrecher: I get your point, but I prefer this approach. With the proposed approach we have to create a few empty views in parts of the code that rarely change. Otherwise, we would have to check in the `read_element` method if the given pointer is valid. With the current approach we get a useable error warning.

src/core/io/src/4C_io_mesh.hpp
isteinbrecher: No, because we don't know the name of the array there. All arrays should have the same size, so we could check all of them, but I think it makes more sense here.

src/core/io/src/4C_io_meshreader.cpp
isteinbrecher: See my comment above. In the element we don't need to be able to check if cell data is passed or not. If it is requested, and not available we get an error, that makes clear what is happening.

isteinbrecher:
Thanks for the review @amgebauer 

>I will not find the time this week to generalize the input to read arbitrary array-types in the vtu input. But, I think I can do it until end next week to avoid the hack of reading it in a (symmetric) tensor. Alternatively, you could also pass two or three names and read the rotation vector for each node separately.

Thanks for looking into this. As this "hack" only affects a couple of lines in the beam element definition, I think we can proceed with this PR for now. Once we can read in general vectors, this can be adapted in the beam element definition.

> Also, I think we should enable to have input fields ready already during element setup so that we can make use of it here. That would me a lot more elegant here. But, this is not trivial here for now, since we might need some larger refactoring of the mesh input (easy for vtu/gmsh/ensight input, but not trivial for the legacy yaml mesh input).

Initially I had the same idea, but conceptually input field are quite different that what is needed here. We need the block cell data once for the element creation, so we simply pass it here, then it is stored in the element. The block cell data is not needed later on. Input fields on the contrary, are keep around for longer and are even distributed and so on - all of that is not needed for the element creation.

src/core/io/src/4C_io_mesh.hpp
amgebauer: I disagree with this. All modern views in c++ don't have such an invalid state, so we should also not do this in our own new view types (if there are no good reasons for that). No matter whether we throw an error or not, you only see when looking at the internals whether it is in a valid state.

src/beam3/src/4C_beam3_reissner.cpp
```diff
@@ -261,9 +261,14 @@ void Discret::Elements::Beam3rType::setup_element_definition(
...
 
+  const std::string nodal_rotation_vector_definition =
+      "Cell data field name for nodal rotation vectors.";
+
   defs[Core::FE::CellType::line2] = all_of({
       parameter<int>("MAT"),
-      parameter<std::vector<double>>("TRIADS", {.size = 6}),
+      parameter<std::optional<std::vector<double>>>("TRIADS", {.size = 6}),
+      parameter<std::optional<std::string>>(
+          "NODAL_ROTATION_VECTORS", {.description = nodal_rotation_vector_definition}),
```
amgebauer: Additional point: What you actually need here is a `one_of`: You can either specify `TRIADS` or `NODAL_ROTATION_VECTORS`. If you use a `one_of`, it is baked into our input (and is also validated with the schema even before running the simulation).

src/core/io/src/4C_io_mesh.hpp
isteinbrecher: Maybe there is a confusion here. This is not a view in the classical sense, but more of a helper object that allows for an easy access to the element cell data inside the element creation method. We can rename it to `Core::IO::MeshInput::ElementCellData` to avoid such confusions.

src/beam3/src/4C_beam3_reissner.cpp
isteinbrecher: Thanks for the tip, I will try that.

src/core/io/src/4C_io_mesh.hpp
amgebauer: It is the perfect use case of a view here :thinking:

isteinbrecher: I don't think so. But the object is not really a view, it is a wrapper to the block cell data and the local cell index that allows the element to access its cell data. Otherwise we would have to pass the cell index (which is different to the element ID) along as well. I implemented this after the offline discussion with you, and I think the current way is quite nice to allow for an easy use in the element creation method. I renamed it, so it is not confused with a view.

amgebauer: It does not own the data, so it is a view onto others data. How it internally holds the data is irrelevant, especially at the setup of the element where we use it. The beam-element does not need to care. Currently, it is a weird construct since it can be valid or invalid during element construction. To determine this, the element has to look into the internals of the object. It would be super easy to make it a robust view construct for what it is the perfect use case.

isteinbrecher: > To determine this, the element has to look into the internals of the object.

No, the element never checks if the element data is empty or not. What the element does is, it checks if it needs to get cell data (independently if the data is provided or not). Then it accesses the data. If the data is empty, an error will be thrown.

amgebauer: This is your specific use case. But you pass this object to all elements (which is ok). However, now all elements get an object where they don't know whether it actually contains data. If you make it a pointer (or an optional) it is clear: If the pointer is defined, data is passed, otherwise, no data is passed.

isteinbrecher: Like I said, the elements don't need to know if the object contains data or not. There is no case where the elements might want to access element data when given and resort to a default otherwise - if such a thing occurs, this is a user error in the input.

However, if for whatever reason (and I don't see one) this would actually occur at some point, we can always add a method to check if there is data or not.

But the general case is that the element decides if it needs the data or not, and if so accesses it. This general use case is much nicer with the current way than passing a pointer.

If you have any more questions, let's discuss this in a short call.

amgebauer: ok, I agree that we disagree here.

src/beam3/src/4C_beam3_reissner_input.cpp
```diff
@@ -50,20 +51,73 @@ bool Discret::Elements::Beam3r::read_element(const std::string& eletype, const s
...
+        // the mesh.
+        const auto& triad_field_data =
+            element_cell_data.get_element_cell_data<Core::LinAlg::SymmetricTensor<double, 3, 3>>(
+                triad_field_name);
+        nodal_rotation_vectors[0] = triad_field_data(0, 0);
+        nodal_rotation_vectors[1] = triad_field_data(1, 1);
+        nodal_rotation_vectors[2] = triad_field_data(2, 2);
+        nodal_rotation_vectors[3] = triad_field_data(0, 1);
+        nodal_rotation_vectors[5] = triad_field_data(0, 2);
+        nodal_rotation_vectors[4] = triad_field_data(1, 2);
```
sebproell: I would comment on this flipped index ordering.

src/core/io/src/4C_io_mesh.hpp
sebproell: >All arrays should have the same size

I would check all array sizes in the constructor. I tend to make newly developed functionality as restrictive as possible and assert early and everywhere.

src/core/io/src/4C_io_mesh.hpp
sebproell: Regarding default-constructible vs not default-constructible: I think both approaches are fine here. @isteinbrecher If you stick to what you have right now, it is not super clear to me why I have public access to the map but also have the getter function. I agree with @amgebauer that `0` is not a good default for `cell_index` since that value would make sense if data is available but is also used now if no data is available. This should be more robust such that all ways to query the data tell you it's invalid. @amgebauer suggested passing a nullptr, but you want the value type, which is fine but then you need to make it more robust and not only consider your singular use case.

src/core/io/src/4C_io_mesh.hpp
isteinbrecher: I added a check for all arrays in the construction and converted the check here to an assertion.

src/core/io/src/4C_io_mesh.hpp
isteinbrecher: Thanks for the comments @sebproell! I made the members private and added a more sensible default value for the `cell_index`.

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -485,6 +485,93 @@ namespace Core::IO::MeshInput
    */
   void print(const PointSet& point_set, std::ostream& os, VerbosityLevel verbose);
 
+  /*!
+   * @brief Allows to pass a pointer to the block cell data and a block local cell index to the
+   * element creation routine.
+   */
+  class ElementCellData
```
amgebauer: What is `ElementCellData`? Either `ElementData` or `CellData`

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -485,6 +485,93 @@ namespace Core::IO::MeshInput
...
+
+    /*!
+     * @brief Get the value of the cell data with the given name for the cell corresponding to this
+     * view.
+     *
+     * @throws if no cell data with the given name exists or if the type of the cell data is not
+     * implicitly convertible to the expected type.
+     */
+    template <typename T>
+    T get_element_cell_data(const std::string& cell_data_key) const
```
amgebauer:
```suggestion
    T get(const std::string& cell_data_key) const
```

src/core/io/src/4C_io_mesh.hpp
maxfirmbach: I agree, `ElementCellData` sounds a bit weird.

src/beam3/src/4C_beam3_reissner.cpp
```diff
@@ -261,9 +261,16 @@ void Discret::Elements::Beam3rType::setup_element_definition(
 
   using namespace Core::IO::InputSpecBuilders;
 
+  const std::string nodal_rotation_vector_definition =
```
maxfirmbach: It is more like a description than a definition, I think.

src/core/io/src/4C_io_mesh.hpp
```diff
@@ -485,6 +485,93 @@ namespace Core::IO::MeshInput
...
+                  cell_index_, cell_data_key, data.size());
+              return static_cast<T>(data[cell_index_]);
+            }
+          },
+          data);
+    }
+
+   private:
+    // Pointer to the cell data map.
+    const CellDataMap* cell_data_ = nullptr;
```
maxfirmbach, line 571: Is a raw pointer a good thing to do here?

isteinbrecher: In this case it is ok, as this object is never stored, only used to create the element. I remove the copy and move operations, so it can not directly be assigned somewhere in element creation.

src/core/io/src/4C_io_mesh.hpp
isteinbrecher: Good point. The most descriptive description would be `ElementDataFromCellData`, I changed it to this.

amgebauer (approved):

PR #1768, opened 2026-02-17
Fix input field interpolation for 1d and 2d elements

mairehenke:
## Description and Context

Currently, the input field interpolation does not work for 1d and 2d elements because the shape functions are selected based on the number of Gauss point coordinates.. However, in `Mat::EvaluationContext`, the coordinates xi are always stored as `Tensor<double, 3>`, and the unused components are zero-padded for 1d and 2d elements.

As a result, the interpolation always selects 3d shape functions, even for 1d and 2d elements that use EvaluationContext.

This PR fixes the issue by determining the dimensionality based on the element cell type instead of the size of xi.

Additionally, I added a small test using vtu input fields with shell7p elements.

### DIFF
--- a/src/core/io/src/4C_io_input_field.hpp
+++ b/src/core/io/src/4C_io_input_field.hpp
@@ -335,6 +335,9 @@ namespace Core::IO
       [[nodiscard]] T interpolate(IndexType element_id, std::span<const double> xi,
           std::string_view field_name = "unknown field") const
       {
+        FOUR_C_ASSERT(xi.size() == 3,
+            "xi must have exactly 3 components (1d/2d elements use zero-padded coordinates)");
+
         if (const T* data = std::get_if<T>(&data_))
         {
           return *data;
@@ -391,7 +394,7 @@ namespace Core::IO
             weights_ptr = weights_on_heap.data();
           }
           std::span<double> weights(weights_ptr, ele.num_node());
-          switch (std::size(xi))
+          switch (Core::FE::get_dimension(ele.shape()))
           {
             case 1:
               Core::FE::shape_function_1d(weights, xi[0], ele.shape());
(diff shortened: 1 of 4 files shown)


src/core/io/src/4C_io_input_field.hpp
```diff
@@ -335,6 +335,10 @@ namespace Core::IO
       [[nodiscard]] T interpolate(IndexType element_id, std::span<const double> xi,
           std::string_view field_name = "unknown field") const
       {
+#ifdef FOUR_C_ENABLE_ASSERTIONS
+        FOUR_C_ASSERT(xi.size() == 3,
+            "xi must have exactly 3 components (1D/2D elements use zero-padded coordinates)");
+#endif
```
amgebauer: This is anyway only active when debug assertions are enabled.
```suggestion
        FOUR_C_ASSERT(xi.size() == 3,
            "xi must have exactly 3 components (1D/2D elements use zero-padded coordinates)");
```

src/core/io/src/4C_io_input_field.hpp
```diff
@@ -391,20 +395,33 @@ namespace Core::IO
             weights_ptr = weights_on_heap.data();
           }
           std::span<double> weights(weights_ptr, ele.num_node());
-          switch (std::size(xi))
+          switch (ele.shape())
```
amgebauer: instead of manually maintaining the list here, you could also

```suggestion
          switch (Core::FE::get_dim(ele.shape()))
```

I think this helper function does not yet exist, but you can add

```c++
  constexpr int get_dim(CellType cell_type)
  {
    return cell_type_switch<all_physical_celltypes>(cell_type,
        [&](auto cell_type_t) { return Core::FE::dim<cell_type_t()>; });
  }
```
in `src/general/element/4C_fem_general_cell_type_traits.hpp`

mairehenke: Thanks for the tip, the helper function was already implemented!

PR #1770, opened 2026-02-17
PoroElast framework: enable Gauss point output / fix case output

dragos-ana:
This PR enables / fixes output for the solid poro elements used in the Poroelasticity framework:

- Enabled Gauss point output by calling the dedicated methods from the underlying material of the StructPoro material.
- Fixed `vis_data` call for solid poro elements: now the implemented method from the StructPoro material is called, instead of the other empty `vis_data` method (see number of arguments). Now, we also obtain the porosity values after postprocessing, instead of the default zero values. 
@c-p-schmidt and I see this only as a temporary fix before completely enabling the new output for the Poroelasticity framework.

### DIFF
--- a/src/solid_poro_3D_ele/4C_solid_poro_3D_ele_pressure_velocity_based.cpp
+++ b/src/solid_poro_3D_ele/4C_solid_poro_3D_ele_pressure_velocity_based.cpp
@@ -377,7 +377,11 @@ bool Discret::Elements::SolidPoroPressureVelocityBased::vis_data(
   // Put the owner of this element into the file (use base class method for this)
   if (Core::Elements::Element::vis_data(name, data)) return true;
 
-  return solid_poro_material().vis_data(name, data, id());
+
+  const unsigned int dummy_gp = 0;
+  return solid_poro_material().vis_data(name, data, dummy_gp,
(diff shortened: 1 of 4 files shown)


c-p-schmidt (approved):

lkoeglmeier (approved):
Thanks for fixing this!

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_pressure_velocity_based.cpp
```diff
@@ -377,7 +377,11 @@ bool Discret::Elements::SolidPoroPressureVelocityBased::vis_data(
   // Put the owner of this element into the file (use base class method for this)
   if (Core::Elements::Element::vis_data(name, data)) return true;
 
-  return solid_poro_material().vis_data(name, data, id());
+
+  const unsigned int dummy_gp = 0;
+  return solid_poro_material().vis_data(name, data, dummy_gp,
+      id());  // we use 0 here for the number of Gauss points, since this is not properly tracked
+              // for the old output; works for now for the underlying struct poro material
```
amgebauer, line 384: It is weird that we need the number of Gauss points here. Materials should know how many Gauss points they have if they evaluate some Gauss point data (typically the size of the array they are writing).

PR #1771, opened 2026-02-18
Fix compilation on darwin using clang

vryy:
## Description and Context

The type `std::invoke_result_t<UnsupportedCellTypeCallable, std::integral_constant<CellType, celltype>>` can be evaluated prematurely as `void` in `4C_fem_general_cell_type_traits_details.hpp`, leading to compilation failure on Apple machine using Xcode's header. This PR proposes a workaround to fix this problem.

None

### DIFF
--- a/src/core/fem/src/general/element/4C_fem_general_cell_type_traits_details.hpp
+++ b/src/core/fem/src/general/element/4C_fem_general_cell_type_traits_details.hpp
@@ -13,6 +13,7 @@
 #include "4C_fem_general_cell_type.hpp"
 #include "4C_fem_general_cell_type_traits_def.hpp"
 #include "4C_utils_exceptions.hpp"
+#include "4C_utils_std23_unreachable.hpp"
 
 #include <array>
 #include <tuple>
@@ -102,19 +103,41 @@ namespace Core::FE::Internal
       return fct(std::integral_constant<CellType, celltype>{});
     }
 
-    constexpr bool is_convertible = std::is_convertible_v<
-        std::invoke_result_t<UnsupportedCellTypeCallable,
-            std::integral_constant<CellType, celltype>>,
-        std::invoke_result_t<Function,
-            std::integral_constant<CellType, first_celltype_in_sequence<CelltypeSequence>::value>>>;
-    constexpr bool is_void = std::is_void_v<std::invoke_result_t<UnsupportedCellTypeCallable,
-        std::integral_constant<CellType, celltype>>>;
-    static_assert(is_convertible || is_void,
-        "The unsupported celltype callable must either return a convertible type to the return "
-        "type of the regular callable or must return void (and throw an error)!");
+    // 1. Determine the actual return types
+    using ErrorReturnType = std::invoke_result_t<UnsupportedCellTypeCallable,
+        std::integral_constant<CellType, celltype>>;
 
-    if constexpr (is_convertible)
+    using SuccessReturnType = std::invoke_result_t<Function,
+        std::integral_constant<CellType, first_celltype_in_sequence<CelltypeSequence>::value>>;
+
+    // 2. Check for void safely
+    constexpr bool is_void = std::is_void_v<ErrorReturnType>;
+
+    // 3. Safely check for convertibility
+    // If ErrorReturnType is void, we pass SuccessReturnType instead (Success is always convertible
+    // to itself). This prevents the "reference to void" error.
+    constexpr bool is_convertible =
+        std::is_convertible_v<std::conditional_t<is_void, SuccessReturnType, ErrorReturnType>,
+            SuccessReturnType>;
+
+    // 4. Now the static_assert works as intended
+    static_assert(is_void || is_convertible,
+        "The unsupported celltype callable must either return void (and throw) "
+        "or return a type convertible to the success path return type!");
+
+    if constexpr (is_void)
     {
+      // Call the handler (which throws).
+      // No 'return' here because it returns void.
+      unsupported_celltype_callable(std::integral_constant<CellType, celltype>{});
+
+      // This is never reached but required to satisfy the compiler. The compiler may still assume
+      // the void type is returned although it is blocked by the previous static_assert.
+      std23::unreachable();
+    }
+    else if constexpr (is_convertible)
+    {
+      // This path is safe because we know it's not void and it's convertible.
       return unsupported_celltype_callable(std::integral_constant<CellType, celltype>{});
     }
 


maxfirmbach:
@vryy Are you able to get 4C running on a Mac machine? (Without any hacks).

src/core/fem/src/general/element/4C_fem_general_cell_type_traits_details.hpp
```diff
@@ -102,19 +102,41 @@ namespace Core::FE::Internal
...
+
+    if constexpr (is_void)
+    {
+      // Call the handler (which throws).
+      // No 'return' here because it returns void.
+      unsupported_celltype_callable(std::integral_constant<CellType, celltype>{});
+
+      // We still need to satisfy the compiler's return requirement for this path,
+      // even though this line will never be reached at runtime.
+      throw std::runtime_error("Unreachable code after unsupported celltype handler");
```
amgebauer:
```suggestion
      std23::unreachable();
```

vryy:
> @vryy Are you able to get 4C running on a Mac machine? (Without any hacks).

@maxfirmbach yeah. This time no hack :))) There're some other changes in the code but it's mainly aesthetic. I compiled Trilinos using 4C's compilation script, and it was running smoothly.

src/core/fem/src/general/element/4C_fem_general_cell_type_traits_details.hpp
vryy: @amgebauer do we need `-std=c++23` for that? AFAIK we binded to `-std=c++20`?

davidrudlstorfer:
@vryy out of curiosity, was this on an Intel Macbook or ARM M* Macbook?

vryy:
> @vryy out of curiosity, was this on an Intel Macbook or ARM M* Macbook?

@davidrudlstorfer on Mac M4 to be exact. The Intel Mac is long gone.

src/core/fem/src/general/element/4C_fem_general_cell_type_traits_details.hpp
amgebauer: no, we have forward-implemented it, that's why it is not in the `std`-namespace but in our `std23`-namspace. The implementation is in `src/core/utils/src/stl_extension/c++23/4C_utils_std23_unreachable.hpp`

maxfirmbach:
@vryy Would you mind providing a little guide how to do the installation on Mac?

vryy:
> @vryy Would you mind providing a little guide how to do the installation on Mac?

I'll think about it. Let me redo it on a clean system, then I can create a short guide to repeat/check.

amgebauer:
@vryy Do the tests work? I remember that @j-brotz also compiled 4C and all libraries on a M* MAC, however, all or many tests failed back then.

vryy:
> @vryy Do the tests work? I remember that @j-brotz also compiled 4C and all libraries on a M* MAC, however, all or many tests failed back then.

That I have not checked. It depends a lot on how Trilinos behaves on ARM and the test tolerance. I recalled that I also never got all the tests passed on my Linux laptop before.

src/core/fem/src/general/element/4C_fem_general_cell_type_traits_details.hpp
```diff
@@ -102,19 +103,40 @@ namespace Core::FE::Internal
...
+        "or return a type convertible to the success path return type!");
+
+    if constexpr (is_void)
+    {
+      // Call the handler (which throws).
+      // No 'return' here because it returns void.
+      unsupported_celltype_callable(std::integral_constant<CellType, celltype>{});
+
+      // This is never reached but required to satisfy the compiler
+      std23::unreachable();
```
sebproell, line 136: Somehow duplicate but inconsistent with the code at the end of this function.

src/core/fem/src/general/element/4C_fem_general_cell_type_traits_details.hpp
```diff
@@ -102,23 +103,39 @@ namespace Core::FE::Internal
...
+    // 4. Now the static_assert works as intended
+    static_assert(is_void || is_convertible,
+        "The unsupported celltype callable must either return void (and throw) "
+        "or return a type convertible to the success path return type!");
+
+    if constexpr (is_void)
+    {
+      // This is never reached but required to satisfy the compiler
+      std23::unreachable();
+    }
```
sebproell, line 137: Why do you need this branch at all?

vryy: Well, without this branch, the compiler will assume that returning a void is possible, meanwhile it's not. Probably the clang compiler on Mac is too greedy/stupid here. At runtime, this will never be reached because of the static_assert before.

amgebauer: But this is now wrong, isn't it? You need to call `unsupported_celltype_callable()`. If the return-type is void, it is a no-return function (throwing the error internally), so `std23::unreachable()` is actually correct.

vryy: yes, I assume `std23::unreachable()` will throw error internally so `unsupported_celltype_callable` call is removed. This is an unreached branch, so it shall be minimal as well.

amgebauer: No, this is not how it is supposed to work. The idea is that you can pass your own function into this cell-type switch that is called if you pass one of the unsupported celltypes. You are running into this function, if the function that you are passing is throwing an error. This is typically used to allow better error messages. Now, even the default error message defined a few lines below it in `ThrowUnsupportedCellTypeError` does not work anymore:

```
      FOUR_C_THROW(
          "The function you are calling is not implemented for the cell type {}. Supported "
          "celltypes are {}",
          Internal::CellTypeInformation<celltype_t()>::name, celltypes_str_acc.c_str());
```

vryy: @amgebauer could you suggest a fix? I'm a bit lost here.

amgebauer: yes, of course. The fix is to go back to the state before the last (force-)push

sebproell: Sorry for the confusion!

PR #1772, opened 2026-02-18
Add some descriptions to a few boundary conditions

ischeider:
## Description and Context

I added a few description to boundary condition options. Please read the description; I appreciate any comments and recommendations for modifications/extensions/language polishing.

I would also like to encourage everyone to add or extend a parameter description if one had to look into the code too often to find out what the parameter does.

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validconditions.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validconditions.cpp
@@ -147,7 +147,12 @@ std::vector<Core::Conditions::ConditionDefinition> Global::valid_conditions()
         "FUNCT", {.description = "function ids", .size = from_parameter<int>("NUMDOF")}));
     cond.add_component(deprecated_selection<std::string>("TYPE",
         {"Live", "Dead", "pseudo_orthopressure", "orthopressure", "PressureGrad"},
-        {.description = "type", .default_value = "Live"}));
+        {.description = "`Live/Dead`: stress vector components (`Dead` is not avail. for solids), "
+                        "`[pseudo_]orthopressure`: pressure normal to surface (only 1 component); "
+                        "for `orthopressure` in a structural analysis one has to provide "
+                        "`LOADLIN:true` in the `STRUCTURAL DYNAMIC` section, `PressureGrad`: "
+                        "prescribed pressure gradient in fluids",
+            .default_value = "Live"}));
 
     condlist.emplace_back(cond);
   };
@@ -268,8 +273,15 @@ std::vector<Core::Conditions::ConditionDefinition> Global::valid_conditions()
         "FUNCT", {.description = "", .size = from_parameter<int>("NUMDOF")}));
 
     // optional
-    cond.add_component(deprecated_selection<std::string>(
-        "TAG", {"none", "monitor_reaction"}, {.description = "", .default_value = "none"}));
+    cond.add_component(deprecated_selection<std::string>("TAG", {"none", "monitor_reaction"},
+        {.description =
+                "If set to **monitor_reaction**, reaction forces (and moments in structural "
+                "analyses) at the constrained nodes are written to data files "
+                "(currently supported: csv and yaml) in a subdirectory of the simulation folder. "
+                "Note that the output has to be switched on in the section `IO/MONITOR STRUCTURE "
+                "DBC` as well. "
+                "Results are only produced for structural analyses.",
+            .default_value = "none"}));
 
     condlist.emplace_back(cond);
   };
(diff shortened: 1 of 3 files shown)


src/global_legacy_module/4C_global_legacy_module_validconditions.cpp
```diff
@@ -268,8 +273,10 @@ std::vector<Core::Conditions::ConditionDefinition> Global::valid_conditions()
         "FUNCT", {.description = "", .size = from_parameter<int>("NUMDOF")}));
 
     // optional
-    cond.add_component(deprecated_selection<std::string>(
-        "TAG", {"none", "monitor_reaction"}, {.description = "", .default_value = "none"}));
+    cond.add_component(deprecated_selection<std::string>("TAG", {"none", "monitor_reaction"},
+        {.description =
+                "for monitor_reaction, the output is controlled via IO/MONITOR STRUCTURE DBC",
```
lauraengelhardt: You provided a nice description in rtd once, maybe we can include parts of this here as well? 

> The ``TAG`` option allows to monitor the reaction forces at the constraint
> nodes by setting it to *monitor_reaction*. With this (TAG) set, the
> reaction forces are written to .csv files in a newly created sub
> directory of the simulation directory. Note that even even the TAG
> parameter can be given for any dirichlet boundary condition, it only
> produces results for forces and moments in a structural analysis.

So how about:
`Optional flag for Dirichlet boundary conditions.
If set to monitor_reaction, reaction forces (and moments in structural analyses) at the constrained nodes are written to .csv files in a subdirectory of the simulation folder. The output is controlled via IO/MONITOR STRUCTURE DBC. Results are only produced for structural analyses.`

src/global_legacy_module/4C_global_legacy_module_validconditions.cpp
```diff
@@ -308,12 +315,14 @@ std::vector<Core::Conditions::ConditionDefinition> Global::valid_conditions()
   // Point coupling (e.g. joints - couple X out of Y nodal DoFs)
 
   Core::Conditions::ConditionDefinition pointcoupling("DESIGN POINT COUPLING CONDITIONS",
-      "PointCoupling", "Point Coupling", Core::Conditions::PointCoupling, false,
-      Core::Conditions::geometry_type_point);
+      "PointCoupling",
+      "All nodes in this Point set share the same displacement in the specified direction(s)",
```
lauraengelhardt:
```suggestion
      "All nodes in this point set share the same displacement in the specified direction(s)",
```

src/global_legacy_module/4C_global_legacy_module_validconditions.cpp
```diff
@@ -308,12 +315,14 @@ std::vector<Core::Conditions::ConditionDefinition> Global::valid_conditions()
...
+      Core::Conditions::PointCoupling, false, Core::Conditions::geometry_type_point);
 
   Core::Conditions::ConditionDefinition pointthermocoupling(
-      "DESIGN POINT THERMO COUPLING CONDITIONS", "PointThermoCoupling", "Point Coupling",
-      Core::Conditions::PointCoupling, false, Core::Conditions::geometry_type_point);
+      "DESIGN POINT THERMO COUPLING CONDITIONS", "PointThermoCoupling",
+      "All nodes in this Point set share the same temperature", Core::Conditions::PointCoupling,
```
lauraengelhardt:
```suggestion
      "All nodes in this point set share the same temperature", Core::Conditions::PointCoupling,
```

src/global_legacy_module/4C_global_legacy_module_validconditions.cpp
```diff
@@ -458,29 +467,42 @@ std::vector<Core::Conditions::ConditionDefinition> Global::valid_conditions()
...
+                "Cartesian components of the rotation vector; rotation direction is defined by the "
+                "direction of this vector, its norm defines the rotation angle",
+            .size = 3}));
+    cond.add_component(parameter<std::vector<std::optional<int>>>(
+        "FUNCT", {.description = "Rotation angle components are multiplied by this function "
+                                 "(mainly for spatial variation)",
+                     .size = 3}));
+    cond.add_component(parameter<int>("USEUPDATEDNODEPOS",
+        {.description = "1: Use node coordinates at new position within function definition, 0: "
+                        "use node coordinates at position of previous time step"}));
```
lauraengelhardt:
```suggestion
                        "Use node coordinates at position of previous time step"}));
```

lauraengelhardt (commented):
Thanks for adding this, @ischeider!
I left some remarks, feel free to change what you consider relevant.

src/global_legacy_module/4C_global_legacy_module_validconditions.cpp
ischeider: Thanks for the hint, I took part of it for the description.

rjoussen: > `are written to .csv files in a subdirectory of the`

It can also write .yaml files! So rather `are written to .csv or .yaml files` or we don't mention the file extension at all.

rjoussen: Also I think it could be clearer that you need to turn on that output is written. But I think the problem is rather that `IO/MONITOR STRUCTURE DBC` **and** the `monitor_reaction` `TAG` are both turned off by default. So you need to turn on both to get output. But that's definitely not this PRs concern :)

ischeider:
Sorry for pushing another version, but the comments by @rjoussen were indeed worth to be included.

PR #1773, opened 2026-02-19
Add required dependency to tests

c-p-schmidt:
## Description and Context
The no-optional-dependencies pipeline is failing due to missing required dependencies (see https://github.com/4C-multiphysics/4C/actions/runs/22160393079/job/64109286461). This is fixed with this PR.


c-p-schmidt:
> Thanks for fixing!
> 
> Would there be any way that we could detect this earlier, i.e., already during the PR? This happens more often in recent times.

We would also need to run the problematic pipelines (which have always been the no-optional-dependencies pipeline) for all PRs. Considering that the fixes are very easy and fast, I'm not sure we really want to have that, as it would almost double the number of jobs to run before a merge can happen.

isteinbrecher:
Thanks for the fix, that was my bad 😬

PR #1774, opened 2026-02-19
Remove untested penalty-based 3D multipoint constraint

lauraengelhardt:
## Description and Context

The condition `DESIGN SURFACE NORMALDIR MULTIPNT CONSTRAINT 3D PEN` is not tested.

We should either test it or remove it. Most multipoint constraints implemented within the `constraint` framework don't work with the `Standard` structural time integration (see https://github.com/4C-multiphysics/4C/issues/210#issuecomment-3818461297). So - even when testing it - we would need to fix some things there when moving to the `Standard` structural time integration.

I thus suggest to remove the condition, if there are no objections.

We have still have `DESIGN POINT COUPLING CONDITION` (that also works with the `Standard` structural time integration) and `DESIGN SURFACE NORMALDIR MULTIPNT CONSTRAINT 3D` (currently only working with `Old`.


mayrmt:
@m-frey @isteinbrecher @eulovi What's your take on this?

m-frey:
This functionality is duplicative here anyway, as it is also implemented in the `constraint_framework` model evaluator. At the moment, it is only used and tested there in 2D. However, there is an open pr (#1751) to extend the implementation to 3D. So I have no objections to removing it here.

isteinbrecher:
> @m-frey @isteinbrecher @eulovi What's your take on this?

I have never used this functionality, so I am fine with removing it.

eulovi:
> @m-frey @isteinbrecher @eulovi What's your take on this?

I am also fine removing this, as I don't use this functionality.

lauraengelhardt:
Nice, thanks for your feedback! I fixed the failing pipeline, so this would be ready from my side :smiley:

m-frey:
@ischeider are you using this ?

lauraengelhardt:
Can I get a second review here, please?

PR #1775, opened 2026-02-19
Switch 2D/3D peridynamics

georghammerl:
## Description and Context
So far, only a 3D peridynamic formulation was available. Now, a switch is included to choose between 3D, 2D plane stress and 2D plane strain formulation. The existing test cases have been adapted and assigned the proper flags - together with an update of the results.

Theory part provided by @armanshojaei


tests/input_files/particle_sph_3d_pdbody_colliding_fast.4C.yaml
```diff
@@ -25,6 +25,7 @@ PARTICLE DYNAMIC/SPH:
 PARTICLE DYNAMIC/PD:
   INTERACTION_HORIZON: 0.3000001
```
slfuchs, line 26: Out of curiosity, why is the value not directly set to 0.3? In most of our SPH cases we set the interaction distance exactly to 3 times the particle spacing.

tests/input_files/particle_sph_2d_pdbody_gravity.4C.yaml
```diff
@@ -45,22 +48,22 @@ RESULT DESCRIPTION:
   - PARTICLE:
       ID: 132
       QUANTITY: "posx"
-      VALUE: 7.92739040181917076e+00
+      VALUE: 7.92632290305826448e+00
```
slfuchs, line 51: I assume the results here and in the cases below need to be adjusted because so far 3D bond forces were evaluated instead of the 2D plane strain bond forces.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -287,11 +308,29 @@ void Particle::SPHPeridynamic::compute_interaction_forces() const
...
+                (std::numbers::pi * horizon_pd_ * horizon_pd_ * horizon_pd_) * stretch *
+                calculate_volume_correction_factor(xi_norm) * dx_pd_ * dx_pd_ * dx_pd_ * dx_pd_;
+          break;
+        case Particle::PD_2DPlaneStrain:
+          fac = (48.00 * (young_i[0] + young_j[0]) * 0.5) /
+                (std::numbers::pi * 5.0 * horizon_pd_ * horizon_pd_ * horizon_pd_) * stretch *
+                calculate_volume_correction_factor(xi_norm) * dx_pd_ * dx_pd_ * dx_pd_ * dx_pd_;
+          break;
+        default:
+          /*"You should not show up here"*/
```
slfuchs: do we need to implement the default case, or should we throw an error here?

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -287,11 +308,29 @@ void Particle::SPHPeridynamic::compute_interaction_forces() const
...
-      double const fac =
-          (12.00 * (young_i[0] + young_j[0]) * 0.5) /
-          (std::numbers::pi * horizon_pd_ * horizon_pd_ * horizon_pd_ * horizon_pd_) * stretch *
-          calculate_volume_correction_factor(xi_norm) * dx_pd_ * dx_pd_ * dx_pd_ * dx_pd_ * dx_pd_ *
-          dx_pd_;
+      double fac;
+      switch (peridynamic_dimension_)
+      {
+        case Particle::PD_3D:
+          fac = (12.00 * (young_i[0] + young_j[0]) * 0.5) /
```
slfuchs:
```suggestion
          fac = (12.0 * (young_i[0] + young_j[0]) * 0.5) /
```

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -287,11 +308,29 @@ void Particle::SPHPeridynamic::compute_interaction_forces() const
...
+                (std::numbers::pi * horizon_pd_ * horizon_pd_ * horizon_pd_ * horizon_pd_) *
+                stretch * calculate_volume_correction_factor(xi_norm) * dx_pd_ * dx_pd_ * dx_pd_ *
+                dx_pd_ * dx_pd_ * dx_pd_;
+          break;
+        case Particle::PD_2DPlaneStress:
+          fac = (9.00 * (young_i[0] + young_j[0]) * 0.5) /
```
slfuchs:
```suggestion
          fac = (9.0 * (young_i[0] + young_j[0]) * 0.5) /
```

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -287,11 +308,29 @@ void Particle::SPHPeridynamic::compute_interaction_forces() const
...
+                (std::numbers::pi * horizon_pd_ * horizon_pd_ * horizon_pd_) * stretch *
+                calculate_volume_correction_factor(xi_norm) * dx_pd_ * dx_pd_ * dx_pd_ * dx_pd_;
+          break;
+        case Particle::PD_2DPlaneStrain:
+          fac = (48.00 * (young_i[0] + young_j[0]) * 0.5) /
```
slfuchs:
```suggestion
          fac = (48.0 * (young_i[0] + young_j[0]) * 0.5) /
```

tests/input_files/particle_sph_2d_pdbody_gravity.4C.yaml
georghammerl: Exactly

tests/input_files/particle_sph_3d_pdbody_colliding_fast.4C.yaml
georghammerl: We wanted to make sure that not numeric precision decides on whether the further distant particles are to be bonded with peridynamic forces or not.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
georghammerl: Sure, thanks.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
georghammerl: good point, is added

tests/input_files/particle_sph_3d_pdbody_colliding_fast.4C.yaml
slfuchs: ok makes sense, thanks.

src/particle/src/4C_particle_input.hpp
```diff
@@ -255,6 +255,14 @@ namespace Particle
     LogNormalRadiusDistribution
   };
 
+  // decide on the dimension of the peridynamic problem
+  enum PeridynamicDimension
+  {
+    PD_3D,
+    PD_2DPlaneStress,
+    PD_2DPlaneStrain
+  };
```
sebproell: Make this an enum class, drop the `PD_` prefix, and use `parameter<PeridynamicDimension>` instead of `deprecated_selection` in the input, please.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
sebproell: FYI, std23::unreachable() might also be used for these cases and does not incur the exception overhead.

src/particle/src/4C_particle_input.hpp
georghammerl: I get the idea, especially to not use `deprecated_selection`, however enum names are not allowed to start with a number. Some prefix will likely be necessary?!

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.hpp
```diff
@@ -103,6 +103,9 @@ namespace Particle
 
     //! contact damping parameter
     const double damp_;
+
+    //! peridynamic dimension
+    const Particle::PeridynamicDimension peridynamic_dimension_;
```
ppraegla:
```suggestion
    const PeridynamicDimension peridynamic_dimension_;
```

src/particle/src/4C_particle_input.cpp
```diff
@@ -591,7 +591,17 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
           parameter<double>(
               "NORMAL_STIFF", {.description = "normal contact stiffness", .default_value = 0.0}),
           parameter<double>("NORMAL_DAMP",
-              {.description = "normal contact damping parameter", .default_value = 0.0})},
+              {.description = "normal contact damping parameter", .default_value = 0.0}),
+
+          // peridynamic dimension
+          deprecated_selection<PeridynamicDimension>("PD_DIMENSION",
```
ppraegla: As @sebproell noted, please use `parameter<PeridynamicDimension>`.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.hpp
georghammerl: Good point. Removed several `Particle::` namespaces inside hpp file.

src/particle/src/4C_particle_input.cpp
georghammerl: Changed it. Now prefix is `Peridynamic_` and the enum looks like: `PeridynamicDimension::Peridynamic_2DPlaneStrain`

src/particle/src/4C_particle_input.hpp
georghammerl: Prefix `Peridynamic_` is added to enums.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
georghammerl: `C++23` is not yet supported. Added `__builtin_unreachable();` instead

ppraegla: We have our own implementation of `unreachable` that will be replace once we support C++23. See https://github.com/4C-multiphysics/4C/blob/main/src/core/utils/src/stl_extension/c%2B%2B23/4C_utils_std23_unreachable.hpp

georghammerl: Nice, I was not aware of this.

ppraegla (approved):

PR #996, opened 2025-07-14
Update deal.II context with fe capabilities in FEValues fashion

marksidoro:
@sebproell 

In the code below, the DealiiWrappers namespace is extended by functionality so that a finite element discretization can be set up using the dealii::FEValues objects. The code allows for assembling global FE matrices in a typical deal.II fashion while keeping the DOF layout compatible with 4C.  The goal is to be able to assemble coupling blocks between 4C and deal.II based discretizations. 

The main classes added are FEValuesContext (and FEFaceValuesContext), which are wrappers around an FEValues object, together with some functionality that handles the reindexing of the local shape functions so that the underlying finite element objects are compatible with the internal ones from 4C. 

For this, currently already available parts of the namespace have been reworked, namely the Context class has been updated to provide some further utility, akin to what a dealii::DofHandler would provide. I have also updated the structure/namespaces in the element_conversion.hpp. 

A helper class to handle the required deal.II mapping is implemented together with a function to set it up with an isogeometric mapping (currently only working for a single cell_type in the Discretization). 

Besides that, there are some further quality of life functions, as well as tests computing an L2-projection, from a simple deal.II Vector directly onto a vector in 4C layout. The problem is solved using a deal.II  solver interface and using deal.II matrices and vectors.

Next steps/limitations
- Using this code, I have already successfully set up a small FSI toy problem which is based on the tutorial step-46 of the deal.II library using the code provided there for the Fluid part of the Domain, while using the Solid from 4C. That code will be merged next once this PR is through. 
- Currently, I have not tested this code in parallel yet. 
- The code is not tested yet for multiple different cell_types within the same discretization

PR #1776, opened 2026-02-19
Add workflow to close stale PRs

ppraegla:
This PR adds a new workflow that uses the [stale action](https://github.com/actions/stale) to automatically handle stale PRs.

After 180 days of inactivity, the action will post a comment and apply a `stale` label.
If there is no activity after another 14 days, the PR will be closed and the label `closed-due-to-inactivity` added.

Issues are not closed.

The action runs once every day. We could also think about running it only once a week.

- [x] TODO: add the new labels before merging

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,41 @@
+# This workflow flags PRs that have been inactive for more than 180 days and closes them after another 14 days.
+# Issues will not be flagged or closed.
+
+name: "Close stale PRs"
+on:
+  schedule:
+    - cron: '30 12 * * *'
+
+env:
+  PR_INACTIVITY_AFTER: 180 # days
+  PR_CLOSE_AFTER: 14 # days
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/stale@b5d41d4e1d5dceea10e7104786b73624c18a190f # v10.2.0
+        with:
+          stale-pr-message: |
+            This pull request has been inactive for **${{ env.PR_INACTIVITY_AFTER }}** days.
+
+            If this feature is still intended to be merged, please leave a brief comment outlining the
+            next intended steps and remove the `stale` label.
+            This will reset the inactivity counter.
+
+            If the pull request can be closed, feel free to close it. Otherwise do **not** change any
+            label or comment as this will wrongly reset the inactivity counter.
+            If no action is taken, this pull request will be automatically closed in
+            **${{ env.PR_CLOSE_AFTER }}** days.
+          close-pr-message: >
+            This pull request is closed due to inactivity. If this was a mistake, feel free to re-open
+            the PR.
+          days-before-pr-stale: ${{ env.PR_INACTIVITY_AFTER }}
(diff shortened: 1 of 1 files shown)


ppraegla:
This PR is ready for review now. I tested the workflow in my fork https://github.com/ppraegla/4C/pull/7.

Maybe we need to increase [operations-per-run](https://github.com/actions/stale?tab=readme-ov-file#operations-per-run). But I left it at the default. Similarly, we could change the processing order from descending to ascending. But I suggest trying the action first and adjusting later.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,41 @@
...
+
+env:
+  PR_INACTIVITY_AFTER: 180 # days
+  PR_CLOSE_AFTER: 14 # days
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
```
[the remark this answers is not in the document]
ppraegla: According to the [documentation](https://github.com/actions/stale?tab=readme-ov-file#stale-pr-message), we only need `pull-requests: write`. `issues: write` is only required for `stale-issue-message` which we do not use.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          stale-pr-message: |
+            This pull request has been inactive for **${{ env.PR_INACTIVITY_AFTER }}** days.
+
+            If this feature is still intended to be merged, please leave a brief comment outlining the
+            next intended steps and remove the `stale` label.
+            This will reset the inactivity counter.
+
+            If the pull request can be closed, feel free to close it. Otherwise do **not** change any
+            label or comment as this will wrongly reset the inactivity counter.
```
lauraengelhardt, line 29: I might be overthinking this, but the wording around “Otherwise” confused me a bit at first: "Otherwise" - _so if the PR should not be closed_ - I should **not** change any labels vs. I **should** remove the `stale` label in case I want to still work on the PR. 

A suggestion would be:
```suggestion
If the pull request can be closed, feel free to close it yourself. 
If no action is taken, this pull request will be automatically closed in  **${{ env.PR_CLOSE_AFTER }}** days. 
Please do  **not** change any label or comment unless you intent to merge this feature as this will wrongly reset the inactivity counter.
```

I'm also fine with the current version, consider this a suggestion. Thanks for implementing!

.github/workflows/stale.yml
```diff
@@ -0,0 +1,41 @@
...
+      pull-requests: write
+    steps:
+      - uses: actions/stale@b5d41d4e1d5dceea10e7104786b73624c18a190f # v10.2.0
+        with:
+          stale-pr-message: |
+            This pull request has been inactive for **${{ env.PR_INACTIVITY_AFTER }}** days.
+
+            If this feature is still intended to be merged, please leave a brief comment outlining the
+            next intended steps and remove the `stale` label.
+            This will reset the inactivity counter.
```
mayrmt, line 26: @ppraegla These line breaks show up in the final comment by the bot. See https://github.com/4C-multiphysics/4C/pull/996#issuecomment-3952029766 for example.

ppraegla: I will look into it. Thanks.

PR #1777, opened 2026-02-20
Cleanup reduced lung main

bgoderbauer:
Takes the different remaining setup stages of the 0D lung framework out of the main function into distinct helper functions and tests them.

### DIFF
--- a/src/reduced_lung/tests/4C_reduced_lung_helpers_test.cpp
+++ b/src/reduced_lung/tests/4C_reduced_lung_helpers_test.cpp
@@ -1,0 +1,351 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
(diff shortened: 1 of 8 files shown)


src/reduced_lung/tests/4C_reduced_lung_helpers_test.cpp
```diff
@@ -0,0 +1,367 @@
...
+        .assign_degrees_of_freedom = true,
+        .init_elements = true,
+        .do_boundary_conditions = false,
+    });
+    return discretization;
+  }
+
+  TEST(ReducedLungHelpersTests, CreateGlobalDofMapsBuildsOffsetsInElementOrder)
+  {
+    skip_if_parallel();
```
sebproell: AFACIT, the test framework doesn't attempt to run this test in parallel, so I don't understand the skip here. If it should run in parallel, you would need to addd a `.np2` or something simliar to the file.

bgoderbauer: Thanks for clarifying! I wasn't sure about this one, but then I just remove it again.

davidrudlstorfer (approved):

PR #1778, opened 2026-02-21
Fix material evaluation in Nitsche trace estimation for solid-surface elements

dragos-ana:
This PR fixes a bug in the Nitsche maximum eigenvalue evaluation, where the material evaluation had to occur on solid surface elements while passing nullptr for the time step size and the maximum time. This led to problems for materials which really required these values. Specifically, the following work was performed:

- Added storage of the time quantities above for the Nitsche contact strategy; these get updated based on the contact parameter interface in each contact integration (and upon restart).
- To also enable evaluations of the reference state, problem-specific timesteps (used specifically for this reference state evaluation) were added to the respective parameter list used in the construction of the contact strategy.
The contact manager was adapted accordingly to also enable evaluations of the old structure time integration + contact framework.
- The time step size and the total time are now passed down via the material evaluation context from the Nitsche strategy to the solid surface elements, and then further to the materials. Hence, the contributions to the trace inequality matrices become computable for the materials requiring time information (e.g. materials with local time integration). A new poro-viscoplasticity test is added for verification, where the viscoplastic material requires the time step size and the total time to be set values in order to proceed with local time integration.
- The evaluations of the Nitsche contributions in the XFEM framework were adapted accordingly.


src/contact/src/4C_contact_nitsche_strategy.cpp
```diff
@@ -461,4 +480,12 @@ void CONTACT::NitscheStrategy::reconnect_parent_elements()
   }
 }
 
+void CONTACT::NitscheStrategy::set_time_step_size_and_total_time(double delta_t)
+{
+  // save time step size and total time
+  time_step_size_ = delta_t;
+  total_time_ = time_step_size_ * step_;
```
c-p-schmidt: This only works if the time step size is not changing over the simulation time.

dragos-ana: Thanks for the review.
True --- however, passing the correct total time seems nontrivial at this point.
I have addressed this for now by renaming the variable to `computed_total_time` and specifying the computation in the doxygen.

c-p-schmidt:
Can someone else please have a look for the second review?

src/contact/src/4C_contact_manager.cpp
```diff
@@ -1083,20 +1083,94 @@ bool CONTACT::Manager::read_and_check_input(Teuchos::ParameterList& cparams) con
...
+      // rauch 01/16
+      if (Core::Communication::my_mpi_rank(get_comm()) == 0)
+        std::cout << "\n \n  Warning: CONTACT::Manager::read_and_check_input() reads TIMESTEP = "
+                  << stru.get<double>("TIMESTEP") << " from --STRUCTURAL DYNAMIC \n"
+                  << std::endl;
+      const double timestep = stru.get<double>("TIMESTEP");
+      cparams.set<double>("TIMESTEP", timestep);
+      cparams.set<double>("TIMESTEP_EVAL_REFERENCE_STATE", timestep);
+
+      break;
```
amgebauer: Is this now still needed? Might it make more sense to throw an error to not silently take the wrong timestep?

src/contact/src/4C_contact_nitsche_strategy.cpp
```diff
@@ -384,9 +392,19 @@ void CONTACT::NitscheStrategy::setup(bool redistributed, bool init)
...
-void CONTACT::NitscheStrategy::update_trace_ineq_etimates()
+void CONTACT::NitscheStrategy::update_trace_ineq_estimates()
 {
   auto NitWgt = Teuchos::getIntegralValue<CONTACT::NitscheWeighting>(params(), "NITSCHE_WEIGHTING");
+
+  // create material evaluation context
+  const double* total_time_ptr = &computed_total_time_;
+  const double* time_step_size_ptr = &time_step_size_;
+  Mat::EvaluationContext mat_eval_context{.total_time = total_time_ptr,
+      .time_step_size = time_step_size_ptr,
```
amgebauer:
```suggestion
  Mat::EvaluationContext mat_eval_context{.total_time = &computed_total_time_,
      .time_step_size = &time_step_size_,
```

src/contact/src/4C_contact_nitsche_strategy.cpp
dragos-ana: After looking more into it based on the review of @amgebauer, I added the total time to the contact parameter interface. Hence, both time step size and total time are now available for the Nitsche strategy.

src/contact/src/4C_contact_manager.cpp
dragos-ana: I could remove it without failing tests.
An error is now thrown if the problem type is not within the switch-case.

src/contact/src/4C_contact_nitsche_strategy.cpp
```diff
@@ -461,4 +478,13 @@ void CONTACT::NitscheStrategy::reconnect_parent_elements()
   }
 }
 
+void CONTACT::NitscheStrategy::save_time_step_size_and_total_time(
+    const double time_step_size, const double total_time)
+{
+  // save time step size and compute total time according to time step * step
```
amgebauer:
```suggestion
  // save time step size and total time
```

src/contact/src/4C_contact_nitsche_strategy.hpp
```diff
@@ -274,11 +276,41 @@ namespace CONTACT
...
+     * @brief Initialize time step size \f$ \Delta t \f$, and total time \f$ t_{n+1} \f$
+     *
+     * @param[in] params parameter list containing timestep for evaluating the reference state
+     */
+    void init_time_step_size_and_total_time(const Teuchos::ParameterList& params)
+    {
+      const double time_step_size = params.get<double>("TIMESTEP_EVAL_REFERENCE_STATE");
+      const double dummy_total_time = 2.0 * time_step_size;
+      save_time_step_size_and_total_time(time_step_size, dummy_total_time);
+    }
```
amgebauer, line 300: If I see it correctly, this is only called during construction. Isn't the correct total time `0.0` in that case? And why do you need the extra parameter `TIMESTEP_EVAL_REFERENCE_STATE`, can't you just use `TIMESTEP`? From what I see on first sight, they are set to the same value.

src/contact/src/4C_contact_nitsche_strategy.hpp
```diff
@@ -274,11 +276,41 @@ namespace CONTACT
...
     std::vector<std::shared_ptr<CONTACT::Interface>> interface_;
 
     std::shared_ptr<Core::LinAlg::Vector<double>> curr_state_;
     bool curr_state_eval_;
 
+    //! time step size
+    double time_step_size_;
+
+    //! total time \f$ t_{n+1} \f$
+    double total_time_;
```
amgebauer, line 312: Is it necessary to also pack and unpack these?

src/contact/src/4C_contact_nitsche_strategy.hpp
dragos-ana: Not all problem types set a `TIMESTEP` in `CONTACT::STRATEGY::Factory::read_and_check_input`. To "play it safe", I introduced this additional construction timestep, which should not interfere with `TIMESTEP` in the respective parameter list. 
In regards to the total time: yes, it is 0.0 at this time point. But maybe some materials will not evaluate it, if the total time t_{n+1} is 0. That's why I made sure that the dummy total time is at least larger than the time step size for the reference state evaluation.

src/contact/src/4C_contact_nitsche_strategy.hpp
dragos-ana: Good question. Nothing else gets packed or unpacked in the Nitsche strategy; since these specific values are set directly in the constructor, it works without packing / unpacking.

amgebauer (dismissed):

src/contact/src/4C_contact_nitsche_strategy.hpp
dragos-ana: I now removed `TIMESTEP_EVAL_REFERENCE_STATE` and added `TIMESTEP` to several other problem types. Also, the total time was set to 0.0. This worked for now without any failing tests.

PR #1779, opened 2026-02-21
Fixes to compile and run on Darwin

vryy:
## Description and Context

Remaining (minor) fixes to compile on Darwin using AppleClang 17.

#1771


vryy:
As requested in #1771, this is test log on Darwin. Not all passed, but looks promising to me.
[log_ctest.txt](https://github.com/user-attachments/files/25460935/log_ctest.txt)

src/core/comm/src/4C_comm_pack_helpers.hpp
```diff
@@ -117,7 +115,14 @@ namespace Core::Communication
     // If T is trivially copyable, we can just copy the bytes. Otherwise, recursively call the
     // pack function for every element. Note that vector<bool> is a special case, as it does not
     // provide the data() method.
-    if constexpr (std::is_trivially_copyable_v<T> && !std::is_same_v<T, bool>)
+#ifdef __APPLE__
+    if constexpr (std::is_same_v<T, bool>)
+    {
+      for (bool b : stuff) add_to_pack(data, b);
+    }
+    else
```
amgebauer, line 126: Why is this needed? Isn't this catched by the `else`-block already?

src/scatra/4C_scatra_functions.cpp
```diff
@@ -10,7 +10,22 @@
...
+  using boost::math::ellint_1;
+  using boost::math::ellint_2;
+  using boost::math::ellint_3;
+  inline double comp_ellint_1(double k) { return boost::math::ellint_1(k); }
+  inline double comp_ellint_2(double k) { return boost::math::ellint_2(k); }
+  inline double comp_ellint_3(double k, double nu) { return boost::math::ellint_3(k, nu); }
+}  // namespace std
+#else
 #include <cmath>
+#endif
```
amgebauer, line 43: If I'm not mistaken, this code here is UB since we are adding things to the std-namespace. If we need these functions, you need to put them into our own namespace and either use the one from std or the one from boost depending on the compiler.

src/core/comm/src/4C_comm_pack_helpers.hpp
vryy: On mac it gave the error

> ~/baci_bundle/baci_master/src/core/comm/src/4C_comm_pack_buffer.hpp:57:58: error: no viable conversion from '__bit_iterator<vector<bool, allocator<bool>>, true>' to 'const void *'
   57 |       std::memcpy(buf_.data() + buf_.size() - sizeof(T), &stuff, sizeof(T));
      |                                                          ^~~~~~
~/baci_bundle/baci_master/src/core/comm/src/4C_comm_pack_helpers.hpp:55:10: note: in instantiation of function template specialization 'FourC::Core::Communication::PackBuffer::add_to_pack<std::__bit_const_reference<std::vector<bool>>>' requested here
   55 |     data.add_to_pack(stuff);
      |          ^
~/baci_bundle/baci_master/src/core/comm/src/4C_comm_pack_helpers.hpp:131:38: note: in instantiation of function template specialization 'FourC::Core::Communication::add_to_pack<std::__bit_const_reference<std::vector<bool>>>' requested here
  131 |       for (const auto& elem : stuff) add_to_pack(data, elem);
  
  so the added lines are to specialize for `bool` type.

src/scatra/4C_scatra_functions.cpp
vryy: The error if those are not turned on is:

> ~/baci_bundle/baci_master/src/scatra/4C_scatra_functions.cpp:311:27: error: no member named 'comp_ellint_1' in namespace 'std'
  311 |     const double K = std::comp_ellint_1(std::sqrt(1.0 - k_squared));

I'm really confused about this because `cmath` is included and `-std=c++20` is on. Probably Xcode's header is so outdated, therefore, those additional functions are wrapped under `#if defined(__APPLE__) && defined(__clang__)`. If we are to put under our namespace, what could it be?

amgebauer: On Apple you are using libc++ from clang, which does not yet implement these special functions. Here is the respective LLVM issue: https://github.com/llvm/llvm-project/issues/99939

There are two options for where we can put these: either in `core/utils/stl_extension` under `Core::Utils`, or you can do it just locally in the file (e.g., `Internal` namespace).

src/core/comm/src/4C_comm_pack_helpers.hpp
amgebauer: ok, now I understand it. I think we could roll this out for all compilers? The cause is the special case that std::vector<bool> is implemented as. This also simplifies the second branch.

mayrmt:
@vryy Nice to see this running on Mac!

Do you have plans to include this into testing, so that we do not loose this capability accidentally? Are there documentation or presets to compile on Mac?

src/scatra/4C_scatra_functions.cpp
```diff
@@ -12,6 +12,36 @@
...
 #include <cmath>
 
+#if defined(__cpp_lib_math_special_functions) && __cpp_lib_math_special_functions >= 201603L
+#define HAS_STD_COMP_ELLINT_X 1
+#else
+#define HAS_STD_COMP_ELLINT_X 0
+#endif
+
+#if HAS_STD_COMP_ELLINT_X
+namespace Internals
```
amgebauer: We call all internal namespaces Internal. Alternatively, you could use an anonymous namespace.

```suggestion
namespace Internal
```

src/core/comm/src/4C_comm_pack_helpers.hpp
```diff
@@ -117,9 +115,9 @@ namespace Core::Communication
     // If T is trivially copyable, we can just copy the bytes. Otherwise, recursively call the
     // pack function for every element. Note that vector<bool> is a special case, as it does not
     // provide the data() method.
-    if constexpr (std::is_trivially_copyable_v<T> && !std::is_same_v<T, bool>)
+    if constexpr (std::is_same_v<T, bool>)
     {
-      add_to_pack(data, stuff.data(), numele * sizeof(T));
+      for (bool b : stuff) add_to_pack(data, b);
     }
```
amgebauer, line 121: Now, we are not making use of trivially copyable types.

I suggest

```suggestion
    }
    else if constexpr (std::is_trivially_copyable_v<T>) {
      add_to_pack(data, stuff.data(), numele * sizeof(T));
    }
```

vryy:
> @vryy Nice to see this running on Mac!
> 
> Do you have plans to include this into testing, so that we do not loose this capability accidentally? Are there documentation or presets to compile on Mac?

Hi @mayrmt, I have thought about it. It's possible to add a workflow for darwin on github. But first I'd like to create an installation guide on a clean system, then we have to think about problematic tests (listed above). We may need to disable them at first, then gradually turn them back on. I'm happy to put this in any taskforce for progress tracking/issue/etc.

src/scatra/4C_scatra_functions.cpp
vryy: I would prefer to use `Internal` here. Only in this class the `comp_ellint_x` is used. If it's used more frequently, we can think about putting in `Core::Utils`.

amgebauer:
> > @vryy Nice to see this running on Mac!
> > Do you have plans to include this into testing, so that we do not loose this capability accidentally? Are there documentation or presets to compile on Mac?
> 
> Hi @mayrmt, I have thought about it. It's possible to add a workflow for darwin on github. But first I'd like to create an installation guide on a clean system, then we have to think about prolematic tests (listed above). We may need to disable them at first, then gradually turn them back on. I'm happy to put this in any taskforce for progress tracking/issue/etc.

Yes, having this in the (nightly) pipeline would be nice! I don't know, however, whether we can cache the compilation of the dependencies (since we cannot use a pre-built Docker). Nevertheless, in the nightly pipeline, we could also run the entire setup in a single job.

PR #1780, opened 2026-02-23
[LinAlg] Add method to detect Dirichlet condition

maxfirmbach:
## Description and Context

This PR adds a method, which is able to detect if a matrix contains Dirichlet rows. This is useful for the linear solver at some point, as missing Dirichlet rows most likely result in a singular matrix (if no other steps have been taken).

### DIFF
--- a/src/core/linalg/tests/4C_linalg_utils_sparse_algebra_assemble_test.np2.cpp
+++ b/src/core/linalg/tests/4C_linalg_utils_sparse_algebra_assemble_test.np2.cpp
@@ -41,31 +41,86 @@ namespace
     Core::LinAlg::SparseMatrix A = Core::LinAlg::read_matrix_market_file_as_sparse_matrix(
         TESTING::get_support_file_path("test_matrices/poisson1d.mm").c_str(), comm_);
 
-    has_dirichlet = Core::LinAlg::has_dirichlet_boundary_condition(A);
-    EXPECT_EQ(has_dirichlet, false);
-
-    // Explicitly modify a row to be Dirichlet
-    if (rank == 0)
+    // Matrix has no Dirichlet rows
     {
-      int num_entries;
-      int* indices;
-      double* values;
-
-      A.extract_my_row_view(0, num_entries, values, indices);
-
-      for (int i = 0; i < num_entries; i++)
-      {
-        if (i == 0)
-          values[i] = 1.0;
-        else
-          values[i] = 0.0;
-      }
-
-      A.replace_global_values(0, num_entries, values, indices);
+      has_dirichlet = Core::LinAlg::has_dirichlet_boundary_condition(A);
+      EXPECT_EQ(has_dirichlet, false);
     }
 
-    has_dirichlet = Core::LinAlg::has_dirichlet_boundary_condition(A);
-    EXPECT_EQ(has_dirichlet, true);
+    // Explicitly modify a row to contain one value
+    {
+      if (rank == 0)
+      {
+        int num_entries;
+        int* indices;
+        double* values;
+
+        A.extract_my_row_view(0, num_entries, values, indices);
+
+        for (int i = 0; i < num_entries; i++)
+        {
+          if (i == num_entries - 1)
+            values[i] = 2.0;
+          else
+            values[i] = 0.0;
+        }
+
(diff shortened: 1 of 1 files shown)


mayrmt:
@maxfirmbach What is an actual 4C use case? If we run dynamic simulations, we do not need Dirichlet BCs. For vector valued PDEs, a single Dirichlet row is not sufficient. How would this method help in such a case?

maxfirmbach:
It can help, for example, to detect singular matrices before we do a factorization to compute an inverse. It just checks if a Dirichlet row exists, this also holds true for elasticity, even if only one degree of freedom is constrained.

EDIT: In addition it can be used as part of the functionality to enforce values of Dirichlet rows on an initial guess, as it is necessary for CG to work with explicit Dirichlet rows.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_assemble.cpp
```diff
@@ -435,4 +435,32 @@ bool Core::LinAlg::is_dirichlet_boundary_condition_already_applied(
...
+    int* indices;
+    double* values;
+
+    A.extract_my_row_view(row, num_entries, values, indices);
+    int number_of_nonzeros = 0;
+
+    for (int j = 0; j < num_entries; j++)
+      if (std::abs(values[j]) > std::numeric_limits<double>::epsilon()) number_of_nonzeros++;
+
+    if (number_of_nonzeros == 1) dirichlet_rows.push_back(row);
```
mayrmt: This only checks the number of entries per row, not the actual value. Would there be additional value in checking if this one entry is an actual 1.0?

maxfirmbach: Yes, I can add a small test that does the respective comparison.

maxfirmbach: @mayrmt I now also check that the diagonal value is 1.0 to really make sure it is Dirichlet.

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_assemble_test.np2.cpp
```diff
@@ -0,0 +1,72 @@
...
+  {
+    bool has_dirichlet;
+    int rank = Core::Communication::my_mpi_rank(comm_);
+
+    Core::LinAlg::SparseMatrix A = Core::LinAlg::read_matrix_market_file_as_sparse_matrix(
+        TESTING::get_support_file_path("test_matrices/poisson1d.mm").c_str(), comm_);
+
+    has_dirichlet = Core::LinAlg::has_dirichlet_boundary_condition(A);
+    EXPECT_EQ(has_dirichlet, false);
+
```
mayrmt, line 49: You could add more tests, e.g. to catch a single non-diagonal entry or a diagonal entry other than 1.0.

maxfirmbach: I added two more tests, reproducing what you described.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_assemble.cpp
```diff
@@ -435,4 +435,48 @@ bool Core::LinAlg::is_dirichlet_boundary_condition_already_applied(
...
+        number_of_nonzeros++;
+      }
+    }
+
+    if (number_of_nonzeros == 1)
+    {
+      const int col = indices[nonzero_index];
+      const double val = values[nonzero_index];
+
+      if (col == row && std::abs(val - 1.0) < std::numeric_limits<double>::epsilon())
```
isteinbrecher, line 469: Out of interest, what would happen, if some weird element assembles just a 1 on the diagonal? That would then also be identified as a Dirichlet condition - would that be an issue?

maxfirmbach: Yep, would also be detected as Dirichlet condition (which it actually also would be), so I assume that's not an issue.

PR #1781, opened 2026-02-23
Add terminal units in 1D pipe flow implementation

engstlerL:
## Description and Context
To enable a direct comparison between the reduced fluid flow implementations in 0D and 1D using an airflow example, terminal units have been introduced into the 1D implementation.
Due to differences in primary variables and the application of boundary conditions between the 0D and 1D formulations, the existing 0D terminal unit implementation could not be reused. Therefore, `reduced_lung_1d_pipe_flow_terminal_unit.cpp ` and `reduced_lung_1d_pipe_flow_terminal_unit.hpp` have been added for the 1D implementation. 
The implementation supports selecting:
- `LinearElasticity` or `OgdenHyperElasticity` as the `ElasticityModel`, and
- `KelvinVoigt` or `FourElementMaxwell` as the `RheologicalModel`.

The corresponding 0D terminal units impose boundary conditions on the 1D terminal nodes, which are coupled via pressure equality to the 1D boundary nodes.

The 1D simulation results have been compared against the simulation results of the 0D test case in `tests/input_files/reduced_lung_3_aw_2_tu.4C.yaml`. While identical results are not expected because of the differing model formulations, the outcomes show sufficiently close agreement.


src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_input.cpp
```diff
@@ -8,6 +8,8 @@
 #include "4C_reduced_lung_1d_pipe_flow_input.hpp"
 
 #include "4C_io_input_spec_builders.hpp"
+#include "4C_reduced_lung_1d_pipe_flow_terminal_unit.hpp"
+#include "4C_reduced_lung_terminal_unit.hpp"
```
bgoderbauer: Do you need this include?

src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_main.cpp
```diff
@@ -1028,6 +1060,50 @@ namespace ReducedLung1dPipeFlow
...
+                    characteristic_W_outgoing, dt);
+
+                int iter = 0;
+                // Newton-Raphson solving for A at 1D terminal node
+                while (abs(residual_f) > 1e-8)
+                {
+                  A_condition -= residual_f / jacobian;
+                  // recompute residual with new values
+                  auto result = terminal_unit.evaluate_residual_jacobian(A_condition,
+                      reference_area_element, beta_element, boundary_Pext, input.fluid.density_rho,
```
bgoderbauer, line 1086: Just a question to clarify the flow: you currently have this inner Newton Loop for each element. In the explicit scheme, you first calculate the outgoing characteristic at the boundary to the terminal unit, and, based on this, you then equilibrate the remaining unknowns in this Newton loop. Did I understand this correctly?

bgoderbauer (approved):
Looks very good, thank you! I think we can merge the 1D and 0D terminal unit model in the future. The "model" data structs are basically the same, we only need different evaluators. I will keep it in mind when we will do potential refactors of the file layout 😄.

src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_main.cpp
engstlerL: Yes: The outgoing characteristic is computed explicitly from the interior state. The Newton loop then solves a single nonlinear equation for the boundary area so that the terminal unit relation and the characteristic compatibility condition are satisfied, which affects the remaining unknowns.

src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_input.cpp
engstlerL: No, I removed them now

src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_main.cpp
bgoderbauer: Interesting, thanks for clarifying!

PR #1782, opened 2026-02-23
[LinAlg] Add locally replicated maps and orthonormalization for multi vector columns

maxfirmbach:
## Description and Context

This PR adds the following things:
- It enables to create maps either in a `globally_distributed` or `locally_replicated` fashion. The first one is the standard case, the second one comes in handy in certain situations e.g. the map is really small and many processors would be empty (`Tpetra` also has this option, in `Epetra` it is an additional object/class).
- A routine is implemented, which orthonormalizes the columns of a multi vector e.g. useful to make use of certain mathematical aspects of the null space of a body. Respective unit-testing is added.
- It fixes a small comment mistake in the outer product of two multi vectors.

### DIFF
--- a/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
+++ b/src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
@@ -533,7 +533,7 @@ Core::LinAlg::MultiVector<double> Core::LinAlg::orthonormalize_multi_vector(
   temp.multiply('T', 'N', 1.0, multi_vector, multi_vector, 0.0);
 
   auto Q = Core::LinAlg::SerialDenseMatrix(Teuchos::DataAccess::Copy, temp.get_values(),
-      temp.get_epetra_multi_vector().Stride(), temp.num_vectors(), temp.num_vectors());
+      temp.stride(), temp.num_vectors(), temp.num_vectors());
 
   Teuchos::LAPACK<int, double> lapack;
   int err = 0;
--- a/src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
+++ b/src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
@@ -244,7 +244,7 @@ namespace
 
     auto map = Core::LinAlg::Map(10, 0, comm_);
     auto multi_vector = Core::LinAlg::MultiVector<double>(map, num_basis_vectors, true);
-    multi_vector.get_epetra_multi_vector().Random();
+    multi_vector.random();
 
     auto local_map = Core::LinAlg::Map(multi_vector.num_vectors(), 0, multi_vector.get_comm(),
         Core::LinAlg::LocalGlobal::locally_replicated);
@@ -261,8 +261,7 @@ namespace
 
       result.multiply('T', 'N', 1.0, multi_vector, multi_vector, 0.0);
       auto gram_matrix = Core::LinAlg::SerialDenseMatrix(Teuchos::DataAccess::Copy,
-          result.get_values(), result.get_epetra_multi_vector().Stride(), result.num_vectors(),
-          result.num_vectors());
+          result.get_values(), result.stride(), result.num_vectors(), result.num_vectors());
 
       FOUR_C_EXPECT_NEAR(gram_matrix, identity, 1e-12);
     }
--- a/src/core/linalg/src/sparse/4C_linalg_multi_vector.hpp
+++ b/src/core/linalg/src/sparse/4C_linalg_multi_vector.hpp
@@ -97,6 +97,12 @@ namespace Core::LinAlg
     //! Print method
     void print(std::ostream& os) const { vector_->Print(os); }
 
+    //! Initialize with random numbers
+    void random() { vector_->Random(); }
+
+    //! Compute stride of the multi vector
+    int stride() const { return vector_->Stride(); }
+
     //! Returns the number of vectors in the multi-vector.
     int num_vectors() const { return vector_->NumVectors(); }
 


src/core/linalg/src/sparse/4C_linalg.hpp
```diff
@@ -53,6 +53,18 @@ namespace Core::LinAlg
...
 
+  /**
+   * @brief Describes how an object is distributed across MPI ranks.
+   *
+   * This enumeration specifies whether an object is stored redundantly on every process or
+   * partitioned across processes in a distributed-memory parallel environment.
+   */
+  enum class LocalGlobal
+  {
+    locally_replicated,
```
mayrmt, line 64: Does this mean, that every process has access to every entry of the map? If yes, that case occurs in 4C, when "extending the ghosting" for interface problems such as in contact.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.hpp
```diff
@@ -207,6 +207,20 @@ namespace Core::LinAlg
...
+   * \brief Orthonormalize the columns of a multi-vector.
+   *
+   * This function takes a multi-vector whose columns represent a set of vectors (e.g. nullspace
+   * or near-nullspace basis vectors) and returns a new multi-vector whose columns form an
+   * orthonormal basis for the same column space.
+   *
+   * \param multi_vector (in): Input multi-vector whose columns will be orthonormalized.
+   *
+   * \return A new multi-vector with orthonormal columns spanning the same column space as the
+   * input.
```
mayrmt, line 220: `\return` is misleading here, since the return type is `void`. I assume, that the implementation performs in -place updates of the input data `multi_vector`, right?

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
```diff
@@ -229,6 +230,43 @@ namespace
...
+
+    auto map = Core::LinAlg::Map(10, 0, comm_);
+    auto multi_vector = Core::LinAlg::MultiVector<double>(map, num_basis_vectors, true);
+    multi_vector.get_epetra_multi_vector().Random();
+
+    auto local_map = Core::LinAlg::Map(multi_vector.num_vectors(), 0, multi_vector.get_comm(),
+        Core::LinAlg::LocalGlobal::locally_replicated);
+    auto result = Core::LinAlg::MultiVector<double>(local_map, multi_vector.num_vectors());
+
+    // The inner product is equal an identity, thus the vector columns are orthonormal.
```
mayrmt:
```suggestion
    // The inner product is equal to an identity, thus the vector columns are orthonormal.
```

src/core/linalg/src/sparse/4C_linalg.hpp
maxfirmbach: Yes, in this case all processors hold the same map (which is usually very small). I can't speak for any physics implementation. The current one looks at things from a pure linear algebra perspective. It is also more of an extension, the functionality is already there in `Tpetra`/`Epetra`.

mayrmt: Sure!

Maps in contact are not small though. Nevermind. i was just wondering.

maxfirmbach: I assume from a technical point of view the same thing is happening. Maybe it would make the implementation in `contact` easier using the locally replicated map directly.

mayrmt: Yes, it probably would. Right now, we are manually computing "fully overlapping column maps". This can probably be improved.

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
maxfirmbach: Fixed.

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.hpp
maxfirmbach: Now the method actually returns the `orthonormal_multi_vector`.

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
```diff
@@ -229,6 +230,43 @@ namespace
...
+   * After orthonormalization, the Gram matrix of the resulting vectors is computed. If the columns
+   * are orthonormal, the Gram matrix must equal the identity matrix.
+   */
+  TEST_F(SparseAlgebraMathTest, MultiVectorOrthonormalization)
+  {
+    const int num_basis_vectors = 3;
+
+    auto map = Core::LinAlg::Map(10, 0, comm_);
+    auto multi_vector = Core::LinAlg::MultiVector<double>(map, num_basis_vectors, true);
+    multi_vector.get_epetra_multi_vector().Random();
```
vovannikov: Is there a way to get a random matrix without descending to the level of `Epetra`? To me it looks like we should have a wrapper around it.

src/core/linalg/tests/4C_linalg_utils_sparse_algebra_math_test.np2.cpp
```diff
@@ -229,6 +230,43 @@ namespace
...
+    {
+      auto identity = Core::LinAlg::SerialDenseMatrix(num_basis_vectors, num_basis_vectors, true);
+      identity(0, 0) = 1.0;
+      identity(1, 1) = 1.0;
+      identity(2, 2) = 1.0;
+
+      multi_vector = Core::LinAlg::orthonormalize_multi_vector(multi_vector);
+
+      result.multiply('T', 'N', 1.0, multi_vector, multi_vector, 0.0);
+      auto gram_matrix = Core::LinAlg::SerialDenseMatrix(Teuchos::DataAccess::Copy,
```
vovannikov, line 263: Here is the same as above. Can we get stride without accessing the `Epetra` entity?

src/core/linalg/src/sparse/4C_linalg_utils_sparse_algebra_math.cpp
```diff
@@ -521,4 +521,37 @@ void Core::LinAlg::multiply_multi_vectors(Core::LinAlg::MultiVector<double>& mul
...
+    const Core::LinAlg::MultiVector<double>& multi_vector)
+{
+  auto local_map = Map(multi_vector.num_vectors(), 0, multi_vector.get_comm(),
+      Core::LinAlg::LocalGlobal::locally_replicated);
+  auto temp = Core::LinAlg::MultiVector<double>(local_map, multi_vector.num_vectors());
+
+  temp.multiply('T', 'N', 1.0, multi_vector, multi_vector, 0.0);
+
+  auto Q = Core::LinAlg::SerialDenseMatrix(Teuchos::DataAccess::Copy, temp.get_values(),
+      temp.get_epetra_multi_vector().Stride(), temp.num_vectors(), temp.num_vectors());
```
vovannikov: And the same question about the stride here 😃

maxfirmbach:
@vovannikov You are absolutely right. I added the missing methods to the multi-vector wrapper.

PR #1783, opened 2026-02-23
Increase error message verbosity

georghammerl:
## Description and Context
Get more information in error case.

### DIFF
--- a/src/particle/src/algorithm/4C_particle_algorithm.cpp
+++ b/src/particle/src/algorithm/4C_particle_algorithm.cpp
@@ -746,6 +746,7 @@ void Particle::ParticleAlgorithm::get_max_particle_position_increment(
       particleengine_->get_particle_container_bundle();
 
   // iterate over particle types
+  ParticleType type_of_particle_with_max_travel_distance = UninitializedType;
   for (auto& typeEnum : particlecontainerbundle->get_particle_types())
   {
     // get container of owned particles of current particle type
@@ -777,13 +778,17 @@ void Particle::ParticleAlgorithm::get_max_particle_position_increment(
 
       // iterate over spatial dimension
       for (int dim = 0; dim < statedim; ++dim)
+      {
         maxpositionincrement = std::max(maxpositionincrement, std::abs(positionincrement[dim]));
+        type_of_particle_with_max_travel_distance = typeEnum;
+      }
     }
   }
 
   // bin size safety check
   if (maxpositionincrement > particleengine_->min_bin_size())
-    FOUR_C_THROW("a particle traveled more than one bin on this processor!");
+    FOUR_C_THROW("a particle of phase '{}' traveled more than one bin on this processor!",
+        enum_to_type_name(type_of_particle_with_max_travel_distance));
 
   // get maximum particle position increment on all processors
   allprocmaxpositionincrement = Core::Communication::max_all(maxpositionincrement, get_comm());


src/particle/src/algorithm/4C_particle_algorithm.cpp
```diff
@@ -777,13 +778,17 @@ void Particle::ParticleAlgorithm::get_max_particle_position_increment(
...
         maxpositionincrement = std::max(maxpositionincrement, std::abs(positionincrement[dim]));
+        type_of_particle_with_max_travel_distance = typeEnum;
+      }
     }
   }
 
   // bin size safety check
   if (maxpositionincrement > particleengine_->min_bin_size())
-    FOUR_C_THROW("a particle traveled more than one bin on this processor!");
+    FOUR_C_THROW("a particle of phase '{}' traveled more than one bin on this processor!",
```
maxfirmbach, line 790: Big letter at the beginning of the sentence is my only complaint 😉.

slfuchs: I fear we need to touch all safety checks in the particle framework then 😆 I don't know why but somehow this became it became a habit to just have small letter error messages

georghammerl:
> Looks good for me. Out of curiosity, did you get this error messages when running simulations? And is the information about the phase of help for debugging? I am just asking, because for SPH it was always the case, that when the error message appears, the simulation got unstable and actually more than one particle mostly of the fluid phases were involved.

Not sure whether it will help. But we actually had exactly this question, whether it is a pdphase or a fluid phase particle that gets unstable. It is not worth to further track this error message down, e.g., which particle id is the problematic one.

src/particle/src/algorithm/4C_particle_algorithm.cpp
maxfirmbach: Alright, then we might keep the current state for now 😅.

src/particle/src/algorithm/4C_particle_algorithm.cpp
```diff
@@ -777,13 +778,17 @@ void Particle::ParticleAlgorithm::get_max_particle_position_increment(
 
       // iterate over spatial dimension
       for (int dim = 0; dim < statedim; ++dim)
+      {
         maxpositionincrement = std::max(maxpositionincrement, std::abs(positionincrement[dim]));
+        type_of_particle_with_max_travel_distance = typeEnum;
```
ppraegla, line 783: Does this work? `type_of_particle_with_max_travel_distance` is assigned the type in every loop. So, at the end, it will just hold the value of the last type in the particle container bundle.

I also added this kind of check for debugging once. There, I did something like
```c++
if (std::abs(positionincrement[dim]) > maxpositionincrement)
{
  type_of_particle_with_max_travel_distance = typeEnum;
  maxpositionincrement = std::abs(positionincrement[dim])
}
```

Alternatively, one could do something like the following to save the assignment in the loop over all particles (this could potentially be all particles). Then the type is only assigned at most the number of types we have.
```c++
double previous_max_position_increment = maxpositionincrement;

// iterate over owned particles of current type
for (int i = 0; i < particlestored; ++i)
{
   ...
    maxpositionincrement = std::max(maxpositionincrement, std::abs(positionincrement[dim]));
 ...
}

if (maxpositionincrement > previous_max_position_increment)
  type_of_particle_with_max_travel_distance = typeEnum;
```

src/particle/src/algorithm/4C_particle_algorithm.cpp
ppraegla, line 790: Also, it would be nice to print the `maxpositionincrement` as well, since we have it readily available. I found this useful to judge if the time step size may be too small when the value is just slightly larger than the bin size, or something else went wrong when it is considerably larger (though this is rather for debugging).

slfuchs:
Thanks @ppraegla, i think you made a point here. The PR is already merged, but we should fix this.

Since this check runs during runtime we should try to keep it low-level / cheap. Such that in cases were everything is fine we just have the simple check that was implemented so far. But what about writing a new method, that is only called as soon as we run into a problem that then reports more details, number of particles affected, lists all affected types and so on? At this level where we fail anyway we don't care about runtime anymore.

src/particle/src/algorithm/4C_particle_algorithm.cpp
slfuchs: see my comment below, that could be part of an advanced check ...

ppraegla:
@slfuchs, I had already considered this advanced check during a debugging session. I discarded it because it would have basically duplicated the function. But we can certainly come up with an advanced check. This probably also makes sense for other error messages.

PR #1787, opened 2026-02-24
Refactor `SPHKernelBase::kernel_space_dimension()` to return a value

vovannikov:
As observed in #1728 it is worth refactoring `SPHKernelBase::kernel_space_dimension()` to return a value instead of modifying a variable passed by reference.

#1728

### DIFF
--- a/src/particle/src/interaction/4C_particle_interaction_sph_kernel.cpp
+++ b/src/particle/src/interaction/4C_particle_interaction_sph_kernel.cpp
@@ -9,6 +9,7 @@
 
 #include "4C_particle_interaction_utils.hpp"
 #include "4C_utils_exceptions.hpp"
+#include "4C_utils_std23_unreachable.hpp"
 
 #include <Teuchos_StandardParameterEntryValidators.hpp>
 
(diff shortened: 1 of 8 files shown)


src/particle/src/interaction/4C_particle_interaction_sph_kernel.hpp
```diff
@@ -35,7 +35,7 @@ namespace Particle
     virtual ~SPHKernelBase() = default;
 
     //! get kernel space dimension
-    virtual void kernel_space_dimension(int& dim) const final;
+    int kernel_space_dimension() const;
```
slfuchs: I prefer virtual final here. This ensures that the derived kernel classes do not override this method.

src/particle/src/interaction/4C_particle_interaction_sph_kernel.cpp
```diff
@@ -24,23 +24,23 @@ Particle::SPHKernelBase::SPHKernelBase(const Teuchos::ParameterList& params)
...
+int Particle::SPHKernelBase::kernel_space_dimension() const
 {
   switch (kernelspacedim_)
   {
-    case Particle::Kernel1D:
+    case Particle::KernelSpaceDimension::Kernel1D:
     {
-      dim = 1;
+      return 1;
       break;
```
georghammerl: Is the `break` line still necessary as a return is done anyway?

src/particle/src/interaction/4C_particle_interaction_sph_kernel.cpp
georghammerl, line 42: Maybe use `std23::unreachable()` here as it is not possible to reach that line.

src/particle/src/interaction/4C_particle_interaction_sph_kernel.cpp
```diff
@@ -49,6 +49,8 @@ void Particle::SPHKernelBase::kernel_space_dimension(int& dim) const
...
     }
   }
+
+  return 0;
```
georghammerl: This is also not necessary (with `std23::unreachable()`)

src/particle/src/interaction/4C_particle_interaction_sph_kernel.hpp
vovannikov: I see the point, fixed.

src/particle/src/interaction/4C_particle_interaction_sph_kernel.cpp
vovannikov: Sure, I missed that.

src/particle/src/interaction/4C_particle_interaction_sph_kernel.cpp
vovannikov: Here I would say that this default branch still can be reached if another entry has been added to the `KernelSpaceDimension` enum.

src/particle/src/interaction/4C_particle_interaction_sph_kernel.cpp
vovannikov: Sure, I fixed that.

PR #1788, opened 2026-02-24
Fix stale bot message

ppraegla:
Switch to a single string with \n escapes to format the message properly. Block style scalars do not work because due to the automatic formatting of yamlfmt.
Literal block style scalars `|` do not work because this will insert a newline for every line (Lines are automatically broken by yamlfmt). Folded block style scalars `>` do not work because we need two empty lines to get an empty line in the output. But yamlfmt only keeps one empty line.


ppraegla:
The checks failed due to `2587 - xfluid_kimmoin_stat_24x24x1_wedge15_levelset.4C.yaml-p3 (Timeout)`

PR #1789, opened 2026-02-25
Change peridynamic contact formulation

georghammerl:
## Description and Context
The current assumption when contact between peridynamic bodies occurs was motivated by DEM, meaning the radii of the involved particles were considered. The SPH framework however relies on a contact formulation that does not account for the radii but compares the current distance of the particle centers with the initial grid spacing. To have a consistent formulation, this is adapted in the peridynamic interaction accordingly.

### DIFF
--- a/src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
+++ b/src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
@@ -17,6 +17,7 @@
 #include "4C_particle_engine_enums.hpp"
 #include "4C_particle_engine_typedefs.hpp"
 #include "4C_particle_interaction_pd_neighbor_pair_struct.hpp"
+#include "4C_utils_parameter_list.fwd.hpp"
 
 FOUR_C_NAMESPACE_OPEN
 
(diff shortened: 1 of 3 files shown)


slfuchs (dismissed):
Looks good from my side!

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.hpp
```diff
@@ -18,6 +18,8 @@
 #include "4C_particle_engine_typedefs.hpp"
 #include "4C_particle_interaction_pd_neighbor_pair_struct.hpp"
 
+#include <Teuchos_ParameterList.hpp>
```
ppraegla:
```suggestion
#include "4C_utils_parameter_list.fwd.hpp"
```
Can you use this header instead to save on the expensive ParameterList include

georghammerl: Sure. Thanks for the hint.

c-p-schmidt (approved):

ppraegla (approved):

PR #1794, opened 2026-03-01
Check element maps too to decide if geometry needs an update in `DiscretizationVisualizationWriterMesh`

vovannikov:
## Description and Context
It turned out, that checking only the node maps when deciding whether the update of the internal data structures inside `DiscretizationVisualizationWriterMesh` is not sufficient. Some output implementations may produce only cell output and it may happen that the corresponding nodes remain intact whereas the elements get updated: in this situation an error is later triggered here:
https://github.com/4C-multiphysics/4C/blob/9e574908659d621cc4249582411a7e7addb7bf85/src/core/io/src/4C_io_visualization_data.cpp#L114-L117

This error was encountered by @mairehenke when working with a Taylor-Couette flow SPH benchmark of @slfuchs, I also then managed to reproduce it on a larger SPH benchmark.

### DIFF
--- a/src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
+++ b/src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
@@ -136,6 +136,10 @@ namespace Core::IO
         std::make_shared<Core::LinAlg::Map>(*discretization_->node_row_map());
     nodecolmap_last_geometry_set_ =
         std::make_shared<Core::LinAlg::Map>(*discretization_->node_col_map());
+    elemrowmap_last_geometry_set_ =
+        std::make_shared<Core::LinAlg::Map>(*discretization_->element_row_map());
+    elemcolmap_last_geometry_set_ =
+        std::make_shared<Core::LinAlg::Map>(*discretization_->element_col_map());
   }
 
   /*-----------------------------------------------------------------------------------------------*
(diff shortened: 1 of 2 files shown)


slfuchs:
I think someone else should review this PR. I have no real experience with this part of the code.

mairehenke:
Thanks for fixing this! Sounds reasonable to me, but I’m also not an expert in this part of the code.

src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
```diff
@@ -136,6 +136,10 @@ namespace Core::IO
         std::make_shared<Core::LinAlg::Map>(*discretization_->node_row_map());
     nodecolmap_last_geometry_set_ =
         std::make_shared<Core::LinAlg::Map>(*discretization_->node_col_map());
+    elemrowmap_last_geometry_set_ =
+        std::make_shared<Core::LinAlg::Map>(*discretization_->element_row_map());
+    elemcolmap_last_geometry_set_ =
```
isteinbrecher: Do we need the element column map as well? We never write column elements. We do write column nodes, thus we check for them.

vovannikov: I can remove it then, I am just not a big expert on how ghosting exactly operates in 4C.

isteinbrecher: I thought about this, and I am not completely sure if we ever write ghosted cells. Do you have an estimate, if this check (and the map comparison is expensive)? Maybe just keep the row map as well, to be sure.

slfuchs: There once was a feature to output the ghosted cells of a bin discretization as a debug feature. This was quite useful. Don't know if this is still implementation.

c-p-schmidt: I'm pretty sure that this is still not ported. See https://github.com/4C-multiphysics/4C/issues/211#issuecomment-2627297846 the entry: `core: Core::Binstrategy::BinningStrategy::write_bin_output`

isteinbrecher: I think it makes most sense to add both checks here, just to be sure. The worst thing that can happen, ist that we have unwanted communication (in very very rare cases).

vovannikov: OK, I will bring it back then.

isteinbrecher: Thanks! And sorry for causing extra work 😅

vovannikov: That's fine, not a lot of work 😄 Done!

PR #1795, opened 2026-03-02
Add cmake preset to compile 4C on Darwin

vryy:
## Description and Context

This PR adds cmake preset and necessary compilation scripts for the dependencies to compile 4C on the Mac/ARM system.

#1771, #1779


vryy:
I have tested the preset on a clean system. For the dependencies there are few remarks:

- homebrew does not provide `parmetis`, so `parmetis` has to be compile manually.
- `cmake` from homebrew is too modern (4.x.x), so cmake3 from `botantony/cmake3` is recommended. Or you can also compile yourself.
- The installation script for `trilinos` assumes other dependencies (suitesparse, superlu_dist) located in a normal folder (not system folder), so the installation script of trilinos needs additional parameter to that folder, e.g., `sh install.sh /path/to/install/dir path/to/dependency/dir`. This can be changed later for the workflow.
- The compilation guide is summarized in a script, i.e., `compile_dependencies.sh`, which is self-explanatory and compile all the dependencies at once. Then 4C can be compiled using the preset `darwin`.

Lastly, the tools required to compile on the Mac is at least: `brew install openmpi git wget gcc@14 hdf5 ninja llvm boost cln`.

sebproell:
> cmake from homebrew is too modern

Are there any issues when using CMake 4.x.x?

sebproell:
What are the future plans for adding all these Darwin-specific install scripts? Do you want to set up a docker build? These files should be tested in some way or they will stop working quickly. It would be OK to do this in a second step, I think.

I am not thrilled to see a new CMake preset, as I believe we should rather work to remove more and more of these. They are specific to your system, and the LNM and IMCS presets are here for legacy reasons.

presets/darwin/CMakePresets.json
```diff
@@ -0,0 +1,37 @@
...
+        "FOUR_C_BACKTRACE_ROOT": "$env{HOME}/opt/libbacktrace",
+        "FOUR_C_QHULL_ROOT": "$env{HOME}/opt/qhull",
+        "FOUR_C_TRILINOS_ROOT": "$env{HOME}/opt/trilinos",
+        "FOUR_C_WITH_MIRCO": "OFF",
+        "FOUR_C_WITH_ARBORX": "OFF",
+        "FOUR_C_WITH_FFTW": "OFF",
+        "FOUR_C_WITH_PYTHON": "OFF",
+        "FOUR_C_ENABLE_METADATA_GENERATION": "OFF",
+        "FOUR_C_ENABLE_FE_TRAPPING": "OFF",
+        "FOUR_C_WITH_GOOGLETEST": "OFF"   // set to off to disable unittest compilation
```
sebproell: Standard JSON does not have comments.


```suggestion
        "FOUR_C_WITH_GOOGLETEST": "OFF" 
```

vryy:
> What are the future plans for adding all these Darwin-specific install scripts? Do you want to set up a docker build? These files should be tested in some way or they will stop working quickly. It would be OK to do this in a second step, I think.
> 
> I am not thrilled to see a new CMake preset, as I believe we should rather work to remove more and more of these. They are specific to your system, and the LNM and IMCS presets are here for legacy reasons.

Hey @sebproell that makes sense, the preset is only useful for the workflow. I will prepare to create a workflow for that.

ppraegla:
>What are the future plans for adding all these Darwin-specific install scripts? Do you want to set up a docker build?

Unfortunately, it is not possible to create a Darwin Docker image. Docker on MacOS runs a Linux virtual machine which then runs Linux-based images. The alternative is to use the GitHub-hosted runners with macOS. The [free macos runner](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) comes with 7 GB of memory (16 GB for Linux runner), which could pose a problem. Also, we would need to install all the dependencies during the workflow.

There is also the possibility to create custom VM images. But this requires a paid runner and, from what I have read, is only possible for Linux- and Windows-based images.

In summary, we could try to set up a workflow that uses a macos runner, install the dependencies during that workflow and build 4C. Due to the long runtime, I think we can only run this workflow once a week. But this should be fine.

vryy:
> > What are the future plans for adding all these Darwin-specific install scripts? Do you want to set up a docker build?
> 
> Unfortunately, it is not possible to create a Darwin Docker image. Docker on MacOS runs a Linux virtual machine which then runs Linux-based images. The alternative is to use the GitHub-hosted runners with macOS. The [free macos runner](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) comes with 7 GB of memory (16 GB for Linux runner), which could pose a problem. Also, we would need to install all the dependencies during the workflow.
> 
> There is also the possibility to create custom VM images. But this requires a paid runner and, from what I have read, is only possible for Linux- and Windows-based images.
> 
> In summary, we could try to set up a workflow that uses a macos runner, install the dependencies during that workflow and build 4C. Due to the long runtime, I think we can only run this workflow once a week. But this should be fine.

Hi @ppraegla, thanks for the useful information. Would you be so kind to add some commits to setup a priori the macos workflow, since you seem to be the one to know it best? Then I will add scripts to install the dependencies and test.

In addition, the `checkcode` workflow throws the error: `The hashes differ. The computed hash is 5244e650 while the container dependencies hash is 9ffec977. You probably need to adapt the dependencies hash in the workflow.` How should I generate these dependency hash?

ppraegla:
>Hi @ppraegla, thanks for the useful information. Would you be so kind to add some commits to setup a priori the macos workflow, since you seem to be the one to know it best? Then I will add scripts to install the dependencies and test.

I can look into it some time next week.

>How should I generate these dependency hash?

This is described [here](https://github.com/4C-multiphysics/4C/tree/main/docker#how-to-update-the-docker-image-for-the-github-actions). You basically compute the hash, update the hash in the workflows (e.g. see #1813), open a PR to build the new docker image, and then rerun the actions in the PR where you need the new dependencies.

I suggest that we first set up the workflow for the macos runner. This will likely take some trial and error. Then we will see what we need to adapt in the dependency scripts.

ppraegla:
@vryy I added the stub for the macos workflow.

In order to test the changes of this PR, you need to commit the workflow file to the main branch in your fork. Otherwise, the new workflow cannot be run. You will test the changes of this PR in your fork. Once the macos.yml workflow is on the main branch in your fork. You can go to Actions -> macOS build test -> Run workflow -> select your branch. Have a look what I did in my fork https://github.com/ppraegla/4C. The output of the workflow is here https://github.com/ppraegla/4C/actions/runs/23536820844/job/68514027187

I suggest that you try to get the workflow to pass in your fork. Once the macos workflow works, we can take care of the checks in this PR, i.e. updating the Docker image.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,45 @@
...
+          g++ -v
+          cmake --version
+          uname
+          sysctl -n hw.physicalcpu
+          pwd
+      - name: Install the dependencies
+        run: |
+          echo "TODO to install the dependencies"
+          #./dependencies/darwin/compile_dependencies.sh
+          brew install openmpi git wget gcc@14 hdf5 ninja llvm boost cln metis
```
ppraegla: @vryy You need to adapt this part to install the dependencies. The build and test step should work. But may need some adapting as well.

dependencies/darwin/superlu_dist/install.sh
ppraegla, line 1: I think we should aim not to duplicate the install scripts. There are only minimal changes compared to the Linux version. This also holds for the other install scripts.

However, I would be fine to keep them separate and think later about how we can unify the installation process.

Something like [dealii candi](https://github.com/dealii/candi) looks interesting, but may be too much for us.

presets/darwin/CMakePresets.json
```diff
@@ -0,0 +1,37 @@
...
+      }
+    },
+    {
+      "name": "darwin",
+      "displayName": "Release build for Darwin/ARM",
+      "description": "Build release version for a Mac ARM-based system.",
```
ppraegla, line 39: Also add a debug configuration with assertions that we can use for the workflow. I think it makes sense to test debug instead of release.

ppraegla:
The macos runner (M processor) only has 3 processors. So, the following test may fail `preconditioner/tutorial_prec_solid_chebyshev.4C.yaml NP 4`. Maybe this can be avoided with MPI  `--oversubscribe`.

vryy:
Hi @ppraegla thanks for providing the initial setup. I have added the compilation scripts and enabled the workflow. The current result is:

      95% tests passed, 131 tests failed out of 2852

The corresponding log is here: https://github.com/vryy/4C/actions/runs/23640661288/job/68860275643

vryy:
Most of the tests failed because of unmatched tolerance, something we all knew about. Some differences are very closed, some are definitely something I will care about. Is there a way to specifically mark them as `skipped` for macos workflow?

amgebauer:
Nice work!

> Is there a way to specifically mark them as skipped for macos workflow?

No, there is not yet such a functionality. However, if it are just slight adaptions to the test tolerances, we should just adapt them instead of skipping the tests.

A further note, I would also try to activate all optional dependencies (like ArborX, VTK, GMSH) so that all tests are tested on MacOS

davidrudlstorfer:
I just had a quick look into the results of a browndyn test and the results are completely different. I guess this has something to do again with our random number generation. Could we have some way to add results for different platforms? I think dealii does something similar right?

presets/darwin/CMakePresets.json
vryy: The problem is that the debug test took a lot of time, it may exceed the allowance of 6 hours. I switch to release test for now and also enable caching.

amgebauer: I think caching will not help here since we only have 20 GB of caches available on Github. Chances are very low that a cache survives a whole day.

maxfirmbach:
@vryy Most of the third-party library packages will be optional (also including parmetis, Umfpack, Superlu_dist, etc.). Would it make sense to try this with a minimal, no-dependency build first?

maxfirmbach:
> The macos runner (M processor) only has 3 processors. So, the following test may fail `preconditioner/tutorial_prec_solid_chebyshev.4C.yaml NP 4`. Maybe this can be avoided with MPI `--oversubscribe`.

I would rather opt to reduce the test to three procs.

vryy:
> @vryy Most of the third-party library packages will be optional (also including parmetis, Umfpack, Superlu_dist, etc.). Would it make sense to try this with a minimal, no-dependency build first?

@maxfirmbach sorry I was OOF recently due to family matter. So far the tests look good except for the small tolerance ones, and browdyn. I think the minimal test should pass right away. Can you suggest a way to enable it?

vryy:
Finally, a workflow for macos is set up and passed, see https://github.com/vryy/4C/actions/runs/30552231074. There are still problematic tests but I disable that for macos by the flag `EXCLUDE_PLATFORM`. These tests can be gradually turned back on by the respective owners by inspecting the testing results on `darwin`. The unit tests are also disabled. The next step is to enable testing of this PR since the dependency hash may need to be updated, which @ppraegla here can help. I suggest to not hinder this PR further since new rebase seems to introduce new problems to the macos build.

vryy:
> > Hi @ppraegla, thanks for the useful information. Would you be so kind to add some commits to setup a priori the macos workflow, since you seem to be the one to know it best? Then I will add scripts to install the dependencies and test.
> 
> I can look into it some time next week.
> 
> > How should I generate these dependency hash?
> 
> This is described [here](https://github.com/4C-multiphysics/4C/tree/main/docker#how-to-update-the-docker-image-for-the-github-actions). You basically compute the hash, update the hash in the workflows (e.g. see #1813), open a PR to build the new docker image, and then rerun the actions in the PR where you need the new dependencies.
> 
> I suggest that we first set up the workflow for the macos runner. This will likely take some trial and error. Then we will see what we need to adapt in the dependency scripts.

Hi @ppraegla, when I updated the yml files with new hash, it automatically generated a new one. Do you know how to get past this cyclic problem?

dependencies/darwin/qhull/install.sh
ppraegla, line 1: This file can move to `dependencies/darwin`. For ubuntu qhull is installed via the apt package `libqhull-dev`. Then it should not be necessary to build the docker image because nothing changed for the docker dependencies.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,90 @@
+name: macOS built test
+
+on:
+  #schedule:
+  #  - cron: 3 4 * * 0
+  workflow_dispatch:
```
ppraegla, line 7: When should the macos workflow be run? I think the runtime will be too long to include in the normal PR. Should it be weekly (current cron setting) or nightly?
I would still keep the manual trigger to allow checking certain PRs manually.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,90 @@
...
+    name: Build and test 4C on macOS
+    runs-on: macos-26
+    steps:
+      - uses: actions/checkout@v6
+      - name: System preparation
+        run: |
+          g++ -v
+          uname
+          sysctl -n hw.physicalcpu
+          pwd
```
ppraegla:
```suggestion
          echo "System information:"
          g++ -v
          uname
          sysctl -n hw.physicalcpu
          pwd
```
I would either remove these commands or print that this is some useful  system information

.github/workflows/macos.yml
```diff
@@ -0,0 +1,90 @@
...
+          sudo ln -s /opt/homebrew/opt/jpeg-turbo/lib/libjpeg.8.dylib /opt/homebrew/Cellar/libpng/1.6.54_1/lib/libjpeg.dylib
+          sudo ln -s /opt/homebrew/opt/fltk/lib/libfltk.1.4.dylib /opt/homebrew/Cellar/fltk/1.4.4/lib/libfltk.dylib
+          sudo ln -s /opt/homebrew/opt/fltk/lib/libfltk_gl.1.4.dylib /opt/homebrew/Cellar/fltk/1.4.4/lib/libfltk_gl.dylib
+      - name: Restore dependencies cache
+        id: cache-deps
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ env.HOME }}/opt
+          key: v2-${{ runner.os }}-deps-${{ hashFiles('./dependencies/darwin/compile_dependencies.sh')
+            }}
```
ppraegla: Does the caching work? As mentioned here https://github.com/4C-multiphysics/4C/pull/1795#discussion_r3043439279, the cache will likely be outdated.

Also, does the hashing do what you want? The hash will be created only from the content of `compile_dependencies.sh`. If you change something in the install scripts called in `compile_dependencies.sh`, the hash will remain unchanged. So you either need to include the install scripts as well or remove the caching. As mentioned above it is questionable if the cache will work.

cmake/configure/configure_MIRCO.cmake
```diff
@@ -32,7 +32,7 @@ else() # Fetch MIRCO from GIT repository
   set(TRILINOS_IN_MIRCO "OFF")
 
   set(MIRCO_GIT_REPO "https://github.com/imcs-compsim/MIRCO.git")
-  set(MIRCO_GIT_TAG "b9d0c4ba27ff8463a3d2b17163fead8800b2650c") # latest hash 03.04.2026
+  set(MIRCO_GIT_TAG "de05a25cf595510b8e315d91aa56b6f7a69ad727")
```
ppraegla: Why is mirco updated?

dependencies/darwin/compile_dependencies.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/zsh
+# This file is part of 4C multiphysics licensed under the
+# GNU Lesser General Public License v3.0 or later.
+#
+# See the LICENSE.md file in the top-level for license information.
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
```
ppraegla, line 9:
```suggestion
set -e
```

I would also use this shell option to exit immediately if a command fails.

presets/docker/CMakePresets.json
```diff
@@ -11,6 +11,7 @@
         "CMAKE_C_COMPILER": "/usr/bin/mpicc",
         "CMAKE_INSTALL_PREFIX": "/home/user/4C",
         "FOUR_C_ENABLE_WARNINGS_AS_ERRORS": "ON",
+        "FOUR_C_ENABLE_FE_TRAPPING": "OFF",
```
ppraegla:
```suggestion
```

You don't need this. The macOS workflow is not running in Docker.

src/cut/4C_cut_kernel.hpp
```diff
@@ -2200,7 +2200,7 @@ namespace Cut::Kernel
         if (det < 1.0e-16) std::cout << "!!! determinant of jacobian is smaller than 1.0e-16 !!!\n";
       }
 
-      if (det < 1.0e-16 && det > 1.0e-16)
+      if (det < 1.0e-16)
```
ppraegla, line 2203: This is already tackled in #2121 . So, I would rather merge it in that PR than here.

tests/list_of_tests.cmake
ppraegla, line 1: I'm fine with excluding some tests for now. From a quick glance many excluded tests use some random number generation. The other tests can be debugged once the macos workflow is on main.

ppraegla (changes requested):
Thanks @vryy for continuously working on the macOS support. I have some comments.

The main comment is that you should not need to update the Docker image for the changes in this PR.

.github/workflows/macos.yml
vryy: I think weekly should be fine. Another problem is that it exceeds 6 hour limit. I circumvent it by running the build workflow, cache, then re-run the workflow for testing.

.github/workflows/macos.yml
vryy: yes, the caching works. And it's essential to bypass the 6-hour limit. The first time the workflow runs, the dependencies will be built and 4C is compiled, then the second time will be only for testing due to caching.

cmake/configure/configure_MIRCO.cmake
vryy: it's remnant from debugging, can be cleared though.

src/cut/4C_cut_kernel.hpp
vryy: I created #2161 to fix this since #2121 has some other problems.

.github/workflows/macos.yml
ppraegla: How does this work in practice? The workflow is triggered automatically once a week. It builds the dependencies (and saves them in the cache). Then the 4C build starts but fails at some point due to hitting the 6-hour job limit. Then you manually restart the workflow. The dependencies are restored from the cache, and 4C is built and cached. The tests start. If the build and tests are not finished in 6 hours, you manually restart the workflow again. Is this the case?

If yes, I would suggest splitting the job into multiple jobs: 1) install the dependencies, 2) build 4C, 3) run the tests. The dependencies and build files are passed via the cache, as you already do it. Maybe this is the better approach than manually restarting the workflow. But you have more practical experience with the macOS workflow.

cmake/configure/configure_MIRCO.cmake
vryy: no, actually there is bug in mirco code that breaks compilation. The new version fixes it.

.github/workflows/macos.yml
vryy: Hi @ppraegla that is the case, and your suggestion as well. But we have only one `macos.yml` for the workflow. If we split them, then the question is how these workflows interact, and how to connect the cache across workflows? I'd rather do it in a separate PR, for now I turn off the google benchmarks so that the workflow stays within the limit.

vryy: In the latest commit, I split `macos.yml` to `macos_build.yml` and `macos_test.yml`. The cache works across both of them, see

https://github.com/vryy/4C/actions/workflows/macos_build.yaml
https://github.com/vryy/4C/actions/workflows/macos_test.yaml

Also, the test workflow is set to start 3 hours later than the build workflow. I think this is sufficient for the build to complete.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,132 @@
...
```
ppraegla: The hash will be created only from the content of `compile_dependencies.sh`. If you change something in the install scripts called in compile_dependencies.sh, the hash will remain unchanged.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,132 @@
...
+        run: |
+          echo "install the dependencies"
+          cmake --version
+          sh ./dependencies/darwin/compile_dependencies.sh
+      - name: Save dependencies cache
+        if: success() && steps.cache-deps.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: ${{ env.HOME }}/opt
+          key: v2-${{ runner.os }}-deps-${{ hashFiles('./dependencies/darwin/compile_dependencies.sh')
```
ppraegla: Why the prefix `v2-` for the cache?

.github/workflows/macos.yml
```diff
@@ -0,0 +1,132 @@
...
+          mkdir -p ${{ env.HOME }}/opt
+          sudo chown -R $(whoami) ${{ env.HOME }}/opt
+          sudo mkdir -p /opt/homebrew/Cellar/fltk/1.4.4/lib
+          sudo mkdir -p /opt/homebrew/Cellar/libpng/1.6.54_1/lib
+          sudo ln -s /opt/homebrew/opt/fltk/lib/libfltk_images.1.4.dylib /opt/homebrew/Cellar/fltk/1.4.4/lib/libfltk_images.dylib
+          sudo ln -s /opt/homebrew/opt/libpng/lib/libpng16.16.dylib /opt/homebrew/Cellar/libpng/1.6.54_1/lib/libpng16.dylib
+          sudo ln -s /opt/homebrew/opt/libpng/lib/libpng.dylib /opt/homebrew/Cellar/libpng/1.6.54_1/lib/libpng.dylib
+          sudo ln -s /opt/homebrew/opt/jpeg-turbo/lib/libjpeg.8.dylib /opt/homebrew/Cellar/libpng/1.6.54_1/lib/libjpeg.dylib
+          sudo ln -s /opt/homebrew/opt/fltk/lib/libfltk.1.4.dylib /opt/homebrew/Cellar/fltk/1.4.4/lib/libfltk.dylib
+          sudo ln -s /opt/homebrew/opt/fltk/lib/libfltk_gl.1.4.dylib /opt/homebrew/Cellar/fltk/1.4.4/lib/libfltk_gl.dylib
```
ppraegla: This could be extracted into a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) so that the runner setup isn't duplicated in the test job, e.g., `.github/actions/macos_system_preparation/action.yml`. This would make it easier to keep the build and test jobs in sync.

Or, this setup can be put in a separate script `dependencies/darwin/system/install.sh` so users trying to install 4C on Darwin can find it more easily than digging through the GitHub Actions. I think this option makes more sense than the custom action.

dependencies/darwin/qhull/install.sh
```diff
@@ -0,0 +1,38 @@
...
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+# Install qhull
+# Call with
+# ./install.sh /path/to/install/dir compile_options
+
+# Exit the script at the first failure
+set -e
+
+INSTALL_DIR="$1"
```
ppraegla, line 16:
```suggestion
INSTALL_DIR="$1"
ADDITIONAL_CMAKE_OPTIONS="$2"
```
I would define another variable for the cmake options and use it below. IMO, this makes it clearer what the second argument is and where it is used.

ppraegla (dismissed):
I made some suggestions for further improvements, which can also be addressed in a follow-up PR. I'm fine with merging as-is, since the PR has been open for so long. 

I mainly looked at the workflow and installation files. Others are more into the cmake stuff.

.github/workflows/macos.yml
vryy: I'll put it in `utilities/set_up_darwin.sh` since it can be reused for a new system

.github/workflows/macos.yml
vryy: I have changed the logic to create the hash based in `install.sh` files in `dependency` folder

.github/workflows/macos.yml
```diff
@@ -0,0 +1,91 @@
+name: macOS built test
+
+on:
+  schedule:
+    - cron: 3 4 * * 0
```
c-p-schmidt, line 5:
```suggestion
    - cron: 3 4 * * 0
      timezone: "Europe/Berlin"
```

You can add this to be more specific and one does not need to search the docs to find out the default

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -397,6 +414,14 @@ function(four_c_test)
     math(EXPR _parsed_TIMEOUT "${FOUR_C_TEST_TIMEOUT_SCALE} * ${_parsed_TIMEOUT}")
   endif()
 
+  # check if the platform is matched
+  # By default we enable the test, but if EXCLUDE_PLATFORM is defined, e.g., Linux|Darwin|Windows,
```
c-p-schmidt: Wouldn't it also be valuable to have this information in the function documentation, where you mention EXCLUDE_PLATFORM, several lines above?

And why is Windows mentioned? 4C isn't runnable on Windows at the moment, right?

c-p-schmidt (commented):
Looks reasonable to me; however, I need to admit that I'm not an expert on these workflow definitions.

Thank you for your efforts to enable 4C builds on MacOS.

c-p-schmidt:
One more thing. Have you looked into the linker warnings at the end of your build? They are not critical but might be worth a look.

`       [1820/1829] Building CXX object tests/cut_test/CMakeFiles/cut_test.dir/cut_test_triangulateFacet.cpp.o

      [1821/1829] Building CXX object tests/cut_test/CMakeFiles/cut_test.dir/cut_test_units.cpp.o

      [1822/1829] Linking CXX executable post_monitor

      ld64.lld: warning: directory not found for option -L/usr/local/lib

      [1823/1829] Linking CXX executable post_processor; Create script post_ensight; Create script post_gid; Create script post_vti; Create script post_vtu

      ld64.lld: warning: directory not found for option -L/usr/local/lib

      [1824/1829] Linking CXX executable tests/benchmarktests_solid_ele

      ld64.lld: warning: directory not found for option -L/usr/local/lib

      [1825/1829] Building CXX object tests/cut_test/CMakeFiles/cut_test.dir/cut_test_volume.cpp.o

      [1826/1829] Building CXX object tests/cut_test/CMakeFiles/cut_test.dir/cut_test_utils.cpp.o

      [1827/1829] Linking CXX executable cut_test

      ld64.lld: warning: directory not found for option -L/usr/local/lib

      [1828/1829] Linking CXX executable 4C; Creating metadata and JSON schema files      

      ld64.lld: warning: directory not found for option -L/usr/local/lib `

c-p-schmidt:
@davidrudlstorfer You might want to have a look at the changes to the CMake functions related to testing.

.github/workflows/macos.yml
ppraegla: The `utilities/` folder isn't the right place for this. Everything related to dependencies should go into the `dependencies/` folder.

cmake/functions/four_c_testing_functions.cmake
vryy: `EXCLUDE_PLATFORM` effectively tells cmake to compare the input with `CMAKE_SYSTEM_NAME`, which can be `Linux|Darwin|Windows`. So these are possible values this variable can be.

vryy:
> One more thing. Have you looked into the linker warnings at the end of your build? They are not critical but might be worth a look.
> 
> ` [1820/1829] Building CXX object tests/cut_test/CMakeFiles/cut_test.dir/cut_test_triangulateFacet.cpp.o
> 
> ```
>   [1821/1829] Building CXX object tests/cut_test/CMakeFiles/cut_test.dir/cut_test_units.cpp.o
> 
>   [1822/1829] Linking CXX executable post_monitor
> 
>   ld64.lld: warning: directory not found for option -L/usr/local/lib
> 
>   [1823/1829] Linking CXX executable post_processor; Create script post_ensight; Create script post_gid; Create script post_vti; Create script post_vtu
> 
>   ld64.lld: warning: directory not found for option -L/usr/local/lib
> 
>   [1824/1829] Linking CXX executable tests/benchmarktests_solid_ele
> 
>   ld64.lld: warning: directory not found for option -L/usr/local/lib
> 
>   [1825/1829] Building CXX object tests/cut_test/CMakeFiles/cut_test.dir/cut_test_volume.cpp.o
> 
>   [1826/1829] Building CXX object tests/cut_test/CMakeFiles/cut_test.dir/cut_test_utils.cpp.o
> 
>   [1827/1829] Linking CXX executable cut_test
> 
>   ld64.lld: warning: directory not found for option -L/usr/local/lib
> 
>   [1828/1829] Linking CXX executable 4C; Creating metadata and JSON schema files      
> 
>   ld64.lld: warning: directory not found for option -L/usr/local/lib `
> ```

i'm not sure why this happens, since `/usr/local/lib` does exist. Maybe the linker is hallucinated :(

P/S: In the latest commit I added a line to create this folder in case it does not exist on github runner for any reason (Note that homebrew for Apple Silicon installed to `/opt/homebrew/lib` by default). Probably some dependencies hardcoded this folder in the linking chain.

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -397,6 +414,14 @@ function(four_c_test)
...
+  # then the test is ignored if the platform is matched.
+  set(skip_var "OFF")
+  if(CMAKE_SYSTEM_NAME IN_LIST _parsed_EXCLUDE_PLATFORM)
+    set(skip_var "ON")
+  endif()
```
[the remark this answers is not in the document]
vryy: this is good catch, checks are added to those helpers to skip the test if the "base" is skipped.

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -233,6 +240,10 @@ function(_add_test_with_options)
     set(_parsed_TOTAL_PROCS 1)
   endif()
 
+  if(NOT DEFINED _parsed_SKIP)
+    set(_parsed_SKIP "OFF")
```
davidrudlstorfer: I think a boolean like "true" and "false" would make more sense here than the arbitrary on/off values which are broadly already in 4C

SKIP True and False is nicer IMHO

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -397,6 +414,14 @@ function(four_c_test)
     math(EXPR _parsed_TIMEOUT "${FOUR_C_TEST_TIMEOUT_SCALE} * ${_parsed_TIMEOUT}")
   endif()
 
+  # check if the platform is matched
+  # By default we enable the test, but if EXCLUDE_PLATFORM is defined, e.g., Linux|Darwin|Windows,
+  # then the test is ignored if the platform is matched.
```
davidrudlstorfer: as @c-p-schmidt already mentioned, this explanation should go above in the function description, otherwise one has to check which platform it can actually be in the code rather than the description

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -397,6 +414,14 @@ function(four_c_test)
...
+  set(skip_var "OFF")
```
davidrudlstorfer: Shouldn't this variable also be initiated in the four_c_test signature? Here we randomly just create it.

cmake/functions/four_c_testing_functions.cmake
davidrudlstorfer: this and the comment below hold for all functions below

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -642,6 +688,21 @@ function(__four_c_test_add_csv_yaml_comparison)
       "diff-with-tolerance ${test_directory}/${_parsed_RESULT_FILE} ${PROJECT_SOURCE_DIR}/tests/input_files/${_parsed_REFERENCE_FILE} ${_parsed_TOL_R} ${_parsed_TOL_A}"
       )
 
+  # skip the test if the main test is skipped
+  get_test_property(${_parsed_BASED_ON} SKIP_RETURN_CODE skip_code_base_var)
+  if(skip_code_base_var STREQUAL "NOTFOUND")
+    set(skip_var "OFF")
+  else()
+    set(skip_var "ON")
+  endif()
```
davidrudlstorfer, line 695: This is automatically done right? I do not think that we need this code? If currently the base test fails the vtk test for example is not executed anyways?

davidrudlstorfer (commented):
@c-p-schmidt thanks for the tag. I checked only the changes to the cmake testing functions.

One general question with the now excluded tests, it will probably make sense to create an issue with those and somehow tackle this in the future.

All the browniandyn tests fail because the random number generation is different on ARM processors, maybe we can adjust the random number generation or as already discussed add second results somehow like dealii does it.

What about the other tests that fail? the Cardiovascular0d tests are nothing super special on a first look, are there already suspicions why these fail? Same for the contact ones (especially because most of them succeed but then some fail)

cmake/functions/four_c_testing_functions.cmake
vryy: yes, but if the test is skipped the vtk test will still be performed if there is no logic to check.

cmake/functions/four_c_testing_functions.cmake
vryy: no, this variable is turned on/off by `EXCLUDE_PLATFORM`  and will be passed down the line to skip the test, and should not be in the signature of `four_c_test` (`EXCLUDE_PLATFORM` is).

vryy:
> @c-p-schmidt thanks for the tag. I checked only the changes to the cmake testing functions.
> 
> One general question with the now excluded tests, it will probably make sense to create an issue with those and somehow tackle this in the future.
> 
> All the browniandyn tests fail because the random number generation is different on ARM processors, maybe we can adjust the random number generation or as already discussed add second results somehow like dealii does it.
> 
> What about the other tests that fail? the Cardiovascular0d tests are nothing super special on a first look, are there already suspicions why these fail? Same for the contact ones (especially because most of them succeed but then some fail)

I agree we shall open the issue to follow up with failed tests in the future. When it is done, one just remove the `EXCLUDE_PLATFORM Darwin` to enable testing on macos.

Some tests failed by just slightly off the tolerance. But again, I'd rather do it in separate PR.

c-p-schmidt (approved):

ppraegla (approved):

ppraegla:
@vryy I just saw that the macos workflow failed. From a quick look, the reason seems to be the vtk system install. 
The build job installed `Bottle Manifest vtk (9.6.2_1)`. The test job ran 1 day later (maybe due to runner availability) and  installed `Bottle Manifest vtk (9.7.0)`

Edit: I think we can just wait until the next workflow run and see if updated system dependencies are a recurring problem.

PR #1805, opened 2026-03-02
Introduce growth and remodeling as ssi problem

tuchpaul:
## Description and Context

This PR enables the solution of Growth and Remodeling (G&R) simulations as a coupled Scalar–Structure-Interaction (SSI) problem.

The evolution equations governing G&R are formulated as a reaction–(diffusion) scalar transport problem. The linear reaction coefficient is determined by the corresponding structural material, establishing the coupling between scalar transport and structural mechanics.

### 1. Scalar Transport Material: scatra_gr

A new scalar transport material `scatra_gr` is introduced. It provides:
- the linear reaction coefficient,
- a reference to the corresponding structural material ID,
- an optional diffusivity parameter.

### 2. Element Routine: scatra_ele_calc_gr

A dedicated element calculation routine `scatra_ele_calc_gr` is implemented.  
Its main task is to evaluate the reaction coefficient based on the structural material response. Apart from this extension, it follows the base class implementation.

### 3. Structural Coupling: remodelfiber_ssi

On the structural side, the constituent instance `remodelfiber_ssi` is extended with additional functionality to:
- receive and set the growth scalar, and the remodeling stretch based on the scalar solution.

## Formulation Details

Currently, the constituent ID is passed to the scalar material, and the scalar index is passed to the constituent. This approach is similar to the concentration-based growth implementation in `InelasticDefgradScalar` (see quantity `SCALAR1`). A more robust design would likely involve establishing a unique mapping during `ssi-base::init()`. However, this would require modifying the base routine, which I intentionally avoided to keep the changes localized.

## Test Case

A new test case based on a quarter cylinder geometry is introduced to verify the coupled functionality.

The scalars considered here are intensive quantities and are therefore treated in non-conservative form (see PR #1792).

In the current test setup:
- The diffusivity is set to 0.0 for all scalars.
- In general, diffusivity is intended to serve primarily as numerical stabilization rather than representing physical diffusion.


src/mixture/src/4C_mixture_constituent.hpp
```diff
@@ -203,6 +204,14 @@ namespace Mixture
      * \return double
      */
     [[nodiscard]] virtual double get_growth_scalar(int gp) const { return 1.0; }
+    [[nodiscard]] virtual double evaluate_growth_reaction_coefficient(int gp) const
+    {
+      FOUR_C_THROW("Only SSI Remodelfibers can do this");
+    }
+    [[nodiscard]] virtual double evaluate_remodeling_reaction_coefficient(int gp) const
+    {
+      FOUR_C_THROW("Only SSI Remodelfibers can do this");
```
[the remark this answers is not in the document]
tuchpaul: FOUR_C_THROW is declared as [[noreturn]]

src/scatra_ele/4C_scatra_ele_calc_growth_remodel.cpp
```diff
@@ -0,0 +1,195 @@
...
+{
+  static auto singleton_map = Core::Utils::make_singleton_map<std::pair<std::string, int>>(
+      [](const int numdofpernode, const int numscal, const std::string& disname)
+      {
+        return std::unique_ptr<ScaTraEleCalcGR<distype, probdim>>(
+            new ScaTraEleCalcGR<distype, probdim>(numdofpernode, numscal, disname));
+      });
+
+  return singleton_map[std::make_pair(disname, numdofpernode)].instance(
+      Core::Utils::SingletonAction::create, numdofpernode, numscal, disname);
```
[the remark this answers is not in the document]
tuchpaul: The key (disname, numdofpernode) is sufficient to identify a unique singleton instance,
This is consistent with other ScaTra singleton implementations (e.g. ScaTraEleCalcAdvReac).

src/scatra_ele/4C_scatra_ele_factory.cpp
```diff
@@ -299,7 +300,7 @@ Discret::Elements::ScaTraEleInterface* Discret::Elements::ScaTraFactory::define_
   {
     if (problem != Inpar::ScaTra::impltype_std and
         problem != Inpar::ScaTra::impltype_cardiac_monodomain and
-        problem != Inpar::ScaTra::impltype_advreac and
+        problem != Inpar::ScaTra::impltype_gr and problem != Inpar::ScaTra::impltype_advreac and
```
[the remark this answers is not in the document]
tuchpaul: This is the precommit-hook suggestion.

src/mat/4C_mat_scatra_gr.hpp
```diff
@@ -0,0 +1,131 @@
...
+FOUR_C_NAMESPACE_OPEN
+
+namespace Mat
+{
+  namespace PAR
+  {
+
+    /*----------------------------------------------------------------------*/
+    /// parameters for scalar transport material with growth/remodeling
+    class ScatraGRMat : public Core::Mat::PAR::Parameter
```
amgebauer: I would spell out G&R since GR might not be that obvious in the SSI framework
```suggestion
    class ScatraGrowthRemodelMat : public Core::Mat::PAR::Parameter
```

src/mixture/src/4C_mixture_constituent.hpp
```diff
@@ -203,6 +204,14 @@ namespace Mixture
...
+    }
```
amgebauer: I think we should avoid adding more stuff to the base class only because a derived class needs it. We should find another way to achieve the same.

src/mixture/src/4C_mixture_constituent.hpp
```diff
@@ -300,6 +309,15 @@ namespace Mixture
...
 
+    virtual void set_current_growth_scalar(const Teuchos::ParameterList& params, int gp)
+    {
+      // do nothing in the default case
+    }
+
+    virtual void set_current_lambda_r(const Teuchos::ParameterList& params, int gp)
+    {
+      // do nothing in the default case
+    }
```
amgebauer: Same applies here

src/mixture/src/4C_mixture_constituent_remodelfiber_ssi.cpp
```diff
@@ -0,0 +1,398 @@
...
+    {
+      data(gp, 0) = remodel_fiber_[gp].evaluate_current_lambda_r();
+    }
+    return true;
+  }
+  return MixtureConstituent::evaluate_output_data(name, data);
+}
+
+Core::LinAlg::SymmetricTensor<double, 3, 3>
+Mixture::MixtureConstituentRemodelFiberSsi::evaluate_current_p_k2(int gp, int eleGID) const
```
amgebauer:
```suggestion
Mixture::MixtureConstituentRemodelFiberSsi::evaluate_current_pk2(int gp, int eleGID) const
```

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -4123,6 +4142,44 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+            parameter<int>("DEPOSITION_STRETCH_TIMEFUNCT",
+                {.description =
+                        "Id of the time function to scale the deposition stretch (Default: 0=None)",
+                    .default_value = 0}),
+            parameter<bool>("INELASTIC_GROWTH",
+                {.description = "Mixture rule has inelastic growth (default false)",
+                    .default_value = false}),
+            parameter<int>("GROWTH_SCALAR_ID",
+                {.description =
+                        "Index of the corresponding growth scalar material in the scatra matlist"}),
```
amgebauer: You check later that the growth scalar id is larger than or equal 0. You could also directly verify it during input with a validator.

src/mixture/src/4C_mixture_constituent_remodelfiber_ssi.cpp
```diff
@@ -0,0 +1,398 @@
...
+    FOUR_C_THROW("Invalid GROWTH_SCALAR_ID {} for provided scalars (size = {}).",
+        params_->growth_scalar_id_, scalar_ptr->size());
+  }
+  double scalar = scalar_ptr->at(static_cast<std::size_t>(params_->growth_scalar_id_));
+  remodel_fiber_[gp].set_growth_scalar(scalar);
+}
+void Mixture::MixtureConstituentRemodelFiberSsi::set_current_lambda_r(
+    const Teuchos::ParameterList& params, int gp)
+{
+  auto& scalar_ptr = params.get<std::shared_ptr<std::vector<double>>>("scalars");
```
amgebauer:
```suggestion
  const auto& scalars = *params.get<std::shared_ptr<std::vector<double>>>("scalars");
```

src/mixture/src/4C_mixture_constituent_remodelfiber_ssi.cpp
```diff
@@ -0,0 +1,398 @@
...
+  if (params_->remodeling_scalar_id_ < 0 ||
+      static_cast<std::size_t>(params_->remodeling_scalar_id_) >= scalar_ptr->size())
+  {
+    FOUR_C_THROW("Invalid REMODELING_SCALAR_ID {} for provided scalars (size = {}).",
+        params_->remodeling_scalar_id_, scalar_ptr->size());
+  }
+  double scalar = scalar_ptr->at(static_cast<std::size_t>(params_->remodeling_scalar_id_));
```
amgebauer: Here you don't need the bound-check
```suggestion
  double scalar = scalars[static_cast<std::size_t>(params_->remodeling_scalar_id_)];
```

src/mixture/src/4C_mixture_constituent_remodelfiber_ssi.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+double Mixture::MixtureConstituentRemodelFiberSsi::get_growth_scalar(int gp) const
+{
+  return remodel_fiber_[gp].evaluate_current_growth_scalar();
+}
+
+void Mixture::MixtureConstituentRemodelFiberSsi::set_current_growth_scalar(
+    const Teuchos::ParameterList& params, int gp)
+{
+  auto& scalar_ptr = params.get<std::shared_ptr<std::vector<double>>>("scalars");
```
amgebauer:
```suggestion
  const auto& scalars = *params.get<std::shared_ptr<std::vector<double>>>("scalars");
```

src/mixture/src/4C_mixture_constituent_remodelfiber_ssi.cpp
```diff
@@ -0,0 +1,398 @@
...
+  if (params_->growth_scalar_id_ < 0 ||
+      static_cast<std::size_t>(params_->growth_scalar_id_) >= scalar_ptr->size())
+  {
+    FOUR_C_THROW("Invalid GROWTH_SCALAR_ID {} for provided scalars (size = {}).",
+        params_->growth_scalar_id_, scalar_ptr->size());
+  }
+  double scalar = scalar_ptr->at(static_cast<std::size_t>(params_->growth_scalar_id_));
```
amgebauer:
```suggestion
  double scalar = scalars[static_cast<std::size_t>(params_->growth_scalar_id_)];
```

tests/input_files/vtu/homogenized-constrained-mixture-qcyl.vtu
amgebauer, line 1: Could you add it in ascii format?

src/scatra_ele/4C_scatra_ele_calc_growth_remodel.cpp
```diff
@@ -0,0 +1,199 @@
...
+    case Core::Materials::m_scatra_gr:
+    {
+      mat_scatra_gr(material, k, densn, densnp, densam, visc, iquad);
+      break;
+    }
+    default:
+    {
+      my::materials(material, k, densn, densnp, densam, visc, iquad);
+      break;
+    }
```
amgebauer, line 94: I'm not an expert in SSI, but wouldn't it be wrong if one uses `ScatraEleCalcGR` not in combination with `m_scatra_gr`?

src/scatra_ele/4C_scatra_ele_calc_gr.cpp
```diff
@@ -0,0 +1,199 @@
...
+{
+  if (gp == -1)
+    FOUR_C_THROW("EVALUATION_MAT: element_center is not implemented for ScaTraEleCalcGR");
+
+  auto* constituent = structmat.constituent_by_material_id(structure_material_id);
+  if (constituent == nullptr)
+  {
+    FOUR_C_THROW(
+        "No constituent with material id {} found in mixture material!", structure_material_id);
+  }
```
amgebauer: Can get this reference once during construction/setup and the store a reference to the respective constituent so that we do not have to search in every Gauss-point loop for the constituent?

src/scatra_ele/4C_scatra_ele_calc_gr.cpp
```diff
@@ -0,0 +1,199 @@
...
+template class Discret::Elements::ScaTraEleCalcGR<Core::FE::CellType::line2, 3>;
+template class Discret::Elements::ScaTraEleCalcGR<Core::FE::CellType::line3, 1>;
+
+// 2D elements
+template class Discret::Elements::ScaTraEleCalcGR<Core::FE::CellType::tri3, 2>;
+template class Discret::Elements::ScaTraEleCalcGR<Core::FE::CellType::tri3, 3>;
+template class Discret::Elements::ScaTraEleCalcGR<Core::FE::CellType::tri6, 2>;
+template class Discret::Elements::ScaTraEleCalcGR<Core::FE::CellType::quad4, 2>;
+template class Discret::Elements::ScaTraEleCalcGR<Core::FE::CellType::quad4, 3>;
+// template class Discret::Elements::ScaTraEleCalcGR<Core::FE::CellType::quad8>;
```
amgebauer: Why are some commented?  Most of them are anyway not needed, right?

src/scatra_ele/4C_scatra_ele_calc_gr.hpp
```diff
@@ -0,0 +1,86 @@
...
+      /// base class alias
+      using my = ScaTraEleCalc<distype, probdim>;
+
+     public:
+      /// Singleton access method
+      static ScaTraEleCalcGR<distype, probdim>* instance(const int numdofpernode, const int numscal,
+          const std::string& disname  //!< creation/destruction indication
+      );
+
+     protected:
```
amgebauer:
```suggestion
     private:
```

src/scatra_ele/4C_scatra_ele_calc_growth_remodel.cpp
tuchpaul: In principle, one could introduce additional transported scalars that use `my::materials` and define their reaction coefficients in the base class via the input file. These would then appear as additional scalars in the `matlist` that are not referenced by the structural material.

However, this particular setup is currently not covered by any tests.

src/scatra_ele/4C_scatra_ele_calc_gr.cpp
tuchpaul: Only templating the 3D materials (which we currently use) did not work in SSI. The commented templates got removed.

src/scatra_ele/4C_scatra_ele_calc_growth_remodel.cpp
mairehenke: But would this make sense then? One could also use the "normal" calculation routine. How would this differ?
I would expect that `ScatraEleCalcGr` should only be used or supported together with `mat_scatra_gr`. If someone wants to use transported scalars, it might be possible to embed your calculation routine into `ScaTraEleCalcAdvReac` or derive from it.

mairehenke:
> Nice work @tuchpaul! I have a few comments. What I don't find super elegant are the many getters and setters to set the internal state of G&R. Could we move the G&R to the ScatraEleCalc and only obtain the stress/stimulus from the constituent?

I would second this. There are quite a few similarities, and we could consider either including it in or inheriting from the normal routine or the advanced reaction calculation method.

tuchpaul:
**Summary on the additional changes:**

**Cosmetic Changes:**
- Renamed `ScatraGRMat` to `ScatraGrowthRemodelMat` for improved readability.
- Added input validators for `GROWTH_SCALAR_ID` and `REMODELING_SCALAR_ID`, removing the redundant runtime checks when these IDs are accessed.
- Moved `evaluate_growth_reaction_coefficient` and `evaluate_remodeling_reaction_coefficient` from `MixtureConstituent` (base class) to `MixtureConstituentRemodelFiberSsi` (derived class only), eliminating the `FOUR_C_THROW` checks in the base class.
- Adapted VTU output format to ASCII.
- Removed untemplated element instantiations.

**Design:**
Added `constituent_` member to `ScaTraEleCalcGR` that caches the `MixtureConstituentRemodelFiberSsi` pointer for each scalar `k`. The cache is done once in `setup_calc()` before the Gauss point loop, avoiding repeated material lookups at every integration point. For each `ScatraGrowthRemodelMat` scalar, the corresponding constituent is resolved via `ssi_constituent_by_material_id()` and directly downcast to `MixtureConstituentRemodelFiberSsi` — an error is thrown if the referred constituent is of a different type. Non-`ScatraGrowthRemodelMat` scalars in the matlist are supported; a `nullptr` is stored for them since `constituent_` is never accessed for those scalars. 
An initial caching of `MixtureConstituentRemodelFiberSsi` is currently not feasible. 
The pointers could in principle be set during `set_materials()` within the element loop. 
However, after redistribution the addresses are no longer valid. As discussed with 
@amgebauer, it would be worthwhile to investigate possible strategies to reconstruct 
the correct addresses after redistribution in order to enable this functionality.

@mairehenke thanks for the comments. The issue described above currently prevents us 
from evaluating the reaction coefficient on the `ScatraGrowthRemodelMat` level and therefore 
requires the additional calculation routine. This routine is necessary because it 
provides access to the structural material through the element.

Adding the corresponding functionality by deriving from `AdvancedReaction` is in 
principle possible and planned for future work. For now, we only consider linear 
non-coupled reactions, which do not require these additional changes.

src/mat/4C_mat_scatra_gr.cpp
```diff
@@ -0,0 +1,86 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#include "4C_mat_scatra_gr.hpp"
```
c-p-schmidt: You could also consider consistently renaming the file.
```suggestion
#include "4C_mat_scatra_growth_remodel.hpp"
```

src/mat/4C_mat_scatra_growth_remodel.hpp
```diff
@@ -0,0 +1,131 @@
...
+     public:
+      /// scalar quantity type for which the reaction coefficient gets evaluated
+      enum class ScalarQuantity
+      {
+        growth,
+        remodeling
+      };
+
+      /// standard constructor
+      ScatraGrowthRemodelMat(const Core::Mat::PAR::Parameter::Data& matdata);
```
c-p-schmidt, line 38:
```suggestion
      explicit ScatraGrowthRemodelMat(const Core::Mat::PAR::Parameter::Data& matdata);
```

src/mixture/src/4C_mixture_constituent_remodelfiber_ssi.cpp
```diff
@@ -0,0 +1,386 @@
...
+#include <algorithm>
+#include <cstddef>
+#include <memory>
+
+FOUR_C_NAMESPACE_OPEN
+
+// anonymous namespace for helper classes and functions
+namespace
+{
+  [[nodiscard]] Core::LinAlg::SymmetricTensor<double, 3, 3> evaluate_c(
```
c-p-schmidt: Wouldn't that be more descriptive?
```suggestion
  [[nodiscard]] Core::LinAlg::SymmetricTensor<double, 3, 3> evaluate_cauchy_green(
```

src/mixture/src/4C_mixture_constituent_remodelfiber_ssi.cpp
```diff
@@ -0,0 +1,386 @@
...
+      const Core::LinAlg::Tensor<double, 3, 3>& F)
+  {
+    return Core::LinAlg::assume_symmetry(Core::LinAlg::transpose(F) * F);
+  }
+
+  [[nodiscard]] Core::LinAlg::SymmetricTensor<double, 3, 3> evaluatei_cext(
```
c-p-schmidt: Maybe you could keep it a bit more generic and readable? That you obtain the inverse Cauchy Green of Fext, you see, depending on what you feed to that method, right?
```suggestion
  [[nodiscard]] Core::LinAlg::SymmetricTensor<double, 3, 3> evaluate_inv_cauchy_green(
```

src/scatra_ele/4C_scatra_ele_calc_gr.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include "4C_utils_singleton_owner.hpp"
+
+#include <cmath>
+
+FOUR_C_NAMESPACE_OPEN
+
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+template <Core::FE::CellType distype, int probdim>
+Discret::Elements::ScaTraEleCalcGR<distype, probdim>::ScaTraEleCalcGR(
```
c-p-schmidt: Would it make sense to change these GRs here also to GrowthRemodel as you did in the material?

src/scatra_ele/4C_scatra_ele_calc_gr.cpp
```diff
@@ -0,0 +1,230 @@
...
+            new ScaTraEleCalcGR<distype, probdim>(numdofpernode, numscal, disname));
+      });
+
+  return singleton_map[std::make_pair(disname, numdofpernode)].instance(
+      Core::Utils::SingletonAction::create, numdofpernode, numscal, disname);
+}
+
+template <Core::FE::CellType distype, int probdim>
+void Discret::Elements::ScaTraEleCalcGR<distype, probdim>::get_material_params(
+    const Core::Elements::Element* ele,  //!< the element we are dealing with
```
c-p-schmidt: These doxygen comments are irrelevant in the cpp.
```suggestion
    const Core::Elements::Element* ele,
```

src/scatra_ele/4C_scatra_ele_calc_gr.hpp
```diff
@@ -0,0 +1,94 @@
...
+      //! setup element evaluation
+      int setup_calc(
+          Core::Elements::Element* ele, Core::FE::Discretization& discretization) override;
+
+     private:
+      using my::nen_;
+      using my::nsd_;
+
+      void get_material_params(
+          const Core::Elements::Element* ele,  //!< the element we are dealing with
```
c-p-schmidt: For methods that you `override` you do not need the doxygen docu as the parents docu is taken.

src/scatra_ele/4C_scatra_ele_calc_growth_remodel.cpp
```diff
@@ -0,0 +1,228 @@
...
+    const auto* constituent = structmat->ssi_constituent_by_material_id(struct_mat_id);
+
+    FOUR_C_ASSERT_ALWAYS(constituent != nullptr,
+        "Constituent with material id {} is not a MixtureConstituentRemodelFiberSsi. "
+        "ScaTraEleCalcGrowthRemodel requires SSI RemodelFiber constituents for m_scatra_gr "
+        "materials.",
+        struct_mat_id);
+
+    constituent_[k] = constituent;
+  };
```
amgebauer: It would be slightly more readable if you just return the constituent
```suggestion
  auto resolve = [](const std::shared_ptr<const Core::Mat::Material>& mat)
  {
    if (mat->material_type() != Core::Materials::m_scatra_gr) return nullptr;

    const auto actmat = std::dynamic_pointer_cast<const Mat::ScatraGrowthRemodelMat>(mat);
    const int struct_mat_id = actmat->structure_material_id();

    const auto* constituent = structmat->ssi_constituent_by_material_id(struct_mat_id);

    FOUR_C_ASSERT_ALWAYS(constituent != nullptr,
        "Constituent with material id {} is not a MixtureConstituentRemodelFiberSsi. "
        "ScaTraEleCalcGrowthRemodel requires SSI RemodelFiber constituents for m_scatra_gr "
        "materials.",
        struct_mat_id);

    return constituent;
  };
```

c-p-schmidt (approved):

PR #1806, opened 2026-03-03
Enable mesh fiber input for muscle materials

lauraengelhardt:
## Description and Context

As the title suggests, enable direct mesh fiber input for muscle materials using the `interpolated_input_field` (thanks to @amgebauer!)


src/mat/4C_mat_muscle_combo.cpp
```diff
@@ -242,9 +213,12 @@ void Mat::MuscleCombo::evaluate(const Core::LinAlg::Tensor<double, 3, 3>* defgra
   // inverse right Cauchy Green tensor C^-1
   const Core::LinAlg::SymmetricTensor<double, 3, 3> invC = Core::LinAlg::inv(C);
 
+  // interpolate fiber orientation at current integration point
+  Core::LinAlg::Tensor<double, 3> orientation =
+      params_->fiber_orientation_.interpolate(eleGID, context.xi->as_span());
```
amgebauer, line 218: You might think about doing this only once an store the interpolated fibers (or maybe the structural tensor) as an internal variable. I want to enable that one can already interpolate during setup, however, currently setup of the material is called very early during reading of the mesh (where we do not yet have setup the input fields).

lauraengelhardt: Good point! I'll merge the current state and do this together with some other improvements :) Thanks for reviewing!

PR #1809, opened 2026-03-04
Implement lazy conversion of mesh data to target types

amgebauer:
This allows us to not only read specific types (previously scalars, `Tensor<3>`, `SymmetricTensor<3, 3>`, `Tensor<3, 3>`) but also arbitrary `std::vector`s or `std::array`s. Each mesh format can define the mapping between the the span of the data at the nodes to the respective target type to account for different orderings of the input data.

@isteinbrecher I changed the beam input which now uses `std::vector` instead of the hacked `SymmetricTensor<3, 3>` / `Tensor<3, 3>`


davidrudlstorfer (approved):
thanks for improving the beam input :)

isteinbrecher:
Thanks @amgebauer for generalising this!

Issue #1784, opened 2026-02-23
[BUG] Superlu_dist crashes on large number of ranks


vovannikov:
## Bug Description: What's going wrong?
I am run one of the @slfuchs PASI problems - a cylinder in the shear flow. The `Superlu_dist` solver crashes somewhere inside `metis` library when solving the structural part of the problem. Maybe this issue can be reproduced even on a pure structural problem, I did not study it that deep. Or maybe the issue is related to the partitioning nuances of the PASI/particle framework.

The issue reveals itself in 3 different `Parmets`/`metis` configurations I have tested:

- Hereon cluster, my own compilation of metis + gklib + Parmetis
- Hereon cluster, @ischeider's compilation of metis + gklib + Parmetis
- my local Ubuntu 24 system, Parmetis is installed as a package via `apt` (I had to use `mpirun --oversubscribe` option to run on 32 ranks on my PC)

The issue appears when `Amesos2::numericFactorization()` is called inside `solve()` method of the direct solver:
https://github.com/4C-multiphysics/4C/blob/b636060a7482c7c7f06d88ca5c6b59d7c57cdf41/src/core/linear_solver/src/method/4C_linear_solver_method_direct.cpp#L127

Command to execute
```
mpirun --oversubscribe -np 32 4C cyl_in_shear_flow_h1_0e-4.4C.yaml test
```

The relevant part of the crash log (my Ubuntu 24 system):
```
[a0dd1900053:05453] Signal: Floating point exception (8)
[a0dd1900053:05453] Signal code: Integer divide-by-zero (1)
[a0dd1900053:05453] Failing at address: 0x7f4da420bb94
[a0dd1900053:05453] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x45330)[0x7f4dacff5330]
[a0dd1900053:05453] [ 1] /lib/x86_64-linux-gnu/libmetis.so.5(libmetis__FM_2WayCutRefine+0x174)[0x7f4da420bb94]
[a0dd1900053:05453] [ 2] /lib/x86_64-linux-gnu/libmetis.so.5(libmetis__RandomBisection+0xe7)[0x7f4da42136e7]
[a0dd1900053:05453] [ 3] /lib/x86_64-linux-gnu/libmetis.so.5(libmetis__InitSeparator+0x185)[0x7f4da4214c25]
[...]

PR #1810, opened 2026-03-04
Use UMFPACK instead of Superlu for a HCM test

amgebauer:
This test recently very often ran into the strange behavior that we see with superlu. I now changed it to UMFPACK.


ischeider (approved):
Even though this is not the test that fails sometimes without any obvious reasons in my case, the commit will probably ease the workflow for others. Since both are direct solvers, is it worth to investigate the "strange behavior"?

amgebauer:
> Since both are direct solvers, is it worth to investigate the "strange behavior"?

Yes, of course, but we have not yet understood the exact problem. And since it is not super reproducible, it's also hard to track it down. @c-p-schmidt works on enabling MUMPS in 4C so that we have a solid alternative for superlu.

Some of the issues that we have in superlu are tracked in #1784 and we have seen problems in the past (#760). My fix back then (#1217) was not yet sufficient to fully eliminate these problems.

PR #1813, opened 2026-03-05
Add mumps to the docker dependency and build Amesos2 with MUMPS dependency

c-p-schmidt:
## Description and Context
This is the first step to enable MUMPS as a direct solver, which is supposed to be an efficient alternative to SuperLU (according to discussions on last week's HPSF community summit). For more info on the solver, please refer to https://mumps-solver.org/ .

The license of MUMPS is `CeCILL-C`, which, according to Wikipedia is fully compatible with our license (LGPL) (see https://en.wikipedia.org/wiki/CeCILL#Other_CeCILL_licenses)


c-p-schmidt:
If you are interested, here you find the running docker pipeline https://github.com/4C-multiphysics/4C/actions/runs/22713807089

c-p-schmidt:
There is two main changes compared to the previous state:
- changes in the ``Dockerfile` are just changes due to warnings (as the old style was outdated)
- enabled more TPLs in Trilinos that are used by MUMPS. Without those the performance is worse (according to AI) ;-)

c-p-schmidt:
This is the new Docker run: https://github.com/4C-multiphysics/4C/actions/runs/22720749064

ischeider:
Could you add the information about the necessary TPLs also to the [documentation](https://4c-multiphysics.github.io/4C/documentation/installation/installation.html#external-dependencies)?

c-p-schmidt:
From my point of view, this PR is ready for reviews now.

davidrudlstorfer (approved):
Really nice! Looking forward to try out the new solver.

(Just to note and as discussed offline: we need to recompile Trilinos for internal and cluster use and also install the apt packages)

amgebauer (approved):
Nice, thank you for enabling Mumps in the pipeline. The only improvement that comes to my mind is to make clear in the docs that Mumps is not mandatory, but this is also true for superlu and suitesparse I think.

c-p-schmidt:
> Nice, thank you for enabling Mumps in the pipeline. The only improvement that comes to my mind is to make clear in the docs that Mumps is not mandatory, but this is also true for superlu and suitesparse I think.

Yes. You are right. From a 4C perspective, it is optional. If you need an installation is defined by your Trilinos installation script. If they are enabled there, they need to be installed on your system.

vryy:
I highly recommend to use MUMPS as parallel direct solver. Below is a scaling study I did a long time ago on a coupled problem with 2-phase. The speed-up is good until 32 processes (taking 8 as the baseline). System size is ~1e6 dofs.

mumps_speedup

PR #1814, opened 2026-03-05
Add copilot instruction for fourcipp compatibility

rjoussen:
## Description and Context

I turns out its fairly easy to break the fourcipp pipeline when changing things in `utilities/four_c_python/src/four_c_metadata` (See #1725 and #1688)

As discussed with @davidrudlstorfer, this adds a copilot instruction to remind everyone of checking the fourcipp pipeline when making any changes to these files.

fourcipp PR: [https://github.com/4C-multiphysics/fourcipp/pull/118](https://github.com/4C-multiphysics/fourcipp/pull/118)

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -3,3 +3,4 @@
 * **Documentation:** Docs are added for new code and updated for modified code.
 * **Modern C++:** No new `Teuchos::RCP` instances are introduced.
 * **CMake:** Newly added tests in `tests/list_of_tests.cmake` set `REQUIRED_DEPENDENCIES` when necessary.
+* **fourcipp:** Any changes made in `utilities/four_c_python/src/four_c_metadata` should be tested for compatibility with `fourcipp`. Some changes might need to be ported to `fourcipp`.


.github/copilot-instructions.md
```diff
@@ -2,4 +2,5 @@
 * **Testing:** New functionality is covered by either a unit or end-to-end test.
 * **Documentation:** Docs are added for new code and updated for modified code.
 * **Modern C++:** No new `Teuchos::RCP` instances are introduced.
-* **CMake:** Newly added tests in `tests/list_of_tests.cmake` set `REQUIRED_DEPENDENCIES` when necessary.
\ No newline at end of file
+* **CMake:** Newly added tests in `tests/list_of_tests.cmake` set `REQUIRED_DEPENDENCIES` when necessary.
+* **fourcipp:** Any changes made in `utilities/four_c_python/src/four_c_metadata` should be tested for compatibility with `fourcipp`. Some changes might need to be ported to `fourcipp`.
```
amgebauer, line 6: I am a bit skeptical about this rule. Without having some premium subscription on group level, copilot does not have a cross-repo scope and can not know how fourcipp looks like and works (as far as I know). The metadata changes in almost every PR. Most changes are not harmful. How can copilot judge what is an incompatible change if it cannot see fourcipp? And what is the kind of help that copilot is able to provide in such a case? Even if he spots a potential incompatibility, we would need to give some advice that copilot needs to give to the user how to test for compatibility.

So what we would need is to describe (in more detail) the types of issues that might occur that are incompatible (that should be tested manually), and what info to provide for the user to do this test.

Personally, I prefer some cross-repository testing if that is doable with not too much effort.

rjoussen: The idea was more that if someone makes changes to any files in `utilities/four_c_python/src/four_c_metadata` this changes how metadata is interpreted/read/written. Changes to these files are quite rare (3 weeks ago and before that 5 months ago). Importantly, the metadata file itself is not in this folder. 

Changes in the metadata file itself should definitely not trigger copilot. If that happens, the instruction must be removed again. But if it understands the instruction correctly I think it would be helpful.

rjoussen: In fact, any normal changes to the metadata file (e.g. if someone changes an input parameter) shouldn't even be detected by copilot since it's only created during compilation.

davidrudlstorfer: @amgebauer the cross-repo test would be super easy and also super quick but it was strongly objected in  #1185 

Maybe we can reformulate the instruction a little bit to something like:

```* **fourcipp:** Any changes made in `utilities/four_c_python/src/four_c_metadata` should be tested for compatibility with `fourcipp`. Some changes might need to be ported to `fourcipp`. Alert the user if compatibility has been ensured.```

I like it either way - thanks for adding this @rjoussen :)

amgebauer: yes, I remember. From time to time we should reevaluate decisions taken in the past. The remarked fears of deadlocks are unsubstantiated if cross-repo testing is setup properly. I don't like having a brittle solution (difficult to detect incompatibilities by copilot, not easy to actually test for incompatibility for users, etc), if we have an easy/stable solution at hand.

davidrudlstorfer: @amgebauer I would be happy to tackle this again because it's always a PITA to handle these fails....

But before tackling this once more I would appreciate if the maintainers could come to a common conception about this so the invested time is not wasted again...

rjoussen: After reading (some of) the comments in #1185 and #973, I understand the objections to make 4C in some way dependent on fourcipp (even though I could probably be convinced of the opposite too).

However, that's why I think this instruction to Copilot could be helpful. In itself, it's obviously not testing anything, but it reminds people to be aware that certain changes that they made might affect fourcipp, and then they can maybe run the fourcipp test suite themselves (which is really not that complicated!). From what I understand, that's pretty much what happens already atm, just that a human (e.g. @davidrudlstorfer) needs to come across these changes and then remind people manually. As I said, changes to the metadata code seem to be pretty rare in the past, so it shouldn't trigger that often.

amgebauer: I think the discussion #1185 ended with a "fact" that is actually wrong. Cross-repository testing is a very common approach in such situations in established codes. Sometimes, it is just less obvious to the users since many larger code-bases cannot afford such extensive PR testings that we currently do (Instead, these code-bases more aggressively revert PRs if things break, an example is LLVM).

Another common approach to test dependent code-projects are mono-repos like Trilinos. For a project with very strong coupling between the sub-projects, this makes sense.

rjoussen: I don't know enough about other codebases or how the cross-repo testing would work to assess this tbh.

Since the objections to full cross-repo testing have led to no progress on #1185 and #973, I do think this copilot instruction could be a good compromise, though, since it definitely doesn't make 4C and the 4C pipeline dependent on fourcipp.

amgebauer: Of course, we can give it a try.

PR #1815, opened 2026-03-05
Add mumps as direct solver

c-p-schmidt:
## Description and Context
This is a draft showing how MUMPS can be added as a direct solver through Amesos2. The pipeline will fail, as MUMPS will only be available after #1813 is merged.
In addition, I added convenience functions to check whether the solver is an iterative or a direct linear solver.
One test case is already adapted to run with MUMPS, which should fail, as it is not yet available. ;-)

Dependent on #1813

### DIFF
--- a/src/core/linear_solver/src/method/4C_linear_solver_method.hpp
+++ b/src/core/linear_solver/src/method/4C_linear_solver_method.hpp
@@ -45,6 +45,7 @@ namespace Core::LinearSolver
   enum class SolverType
   {
     KLU2,     ///< Amesos direct solver using KLU2
+    MUMPS,    ///< Amesos direct solver using MUMPS
     UMFPACK,  ///< Amesos direct solver using UMFPACK
     Superlu,  ///< Amesos direct solver using SuperLU_Dist
     Belos,    ///< Belos iterative solver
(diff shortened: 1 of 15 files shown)


src/core/linear_solver/src/method/4C_linear_solver_method_direct.cpp
```diff
@@ -79,6 +79,11 @@ void Core::LinearSolver::DirectSolver::setup(std::shared_ptr<Core::LinAlg::Spars
         klu_params.set("IsContiguous", false, "Are GIDs Contiguous");
         break;
       }
+      case SolverType::Mumps:
+      {
+        solver_type = "MUMPS";
```
c-p-schmidt, line 84: Here, some MUMPS settings could or would need to be added if desired. The possibilities can be found here: https://github.com/trilinos/Trilinos/blob/a323658d9cecef1f21209af56681e21f58ca12f3/packages/amesos2/src/Amesos2_MUMPS_decl.hpp#L135

maxfirmbach (dismissed):

c-p-schmidt:
FYI: @ischeider

maxfirmbach:
Looks good! Nice to have that feature.

c-p-schmidt:
@lkoeglmeier: Could you have a brief look whether the changes to poro* are reasonable?

amgebauer (dismissed):

c-p-schmidt:
@4C-multiphysics/developers: As soon as this PR is merged, we will also provide Mumps (https://mumps-solver.org) as an alternative direct solver for linear systems of equations.

To make use of Mumps, you (or an admin at your institute) need to do the following:
1. Ensure that the relevant dependencies are installed on your workstation. For Ubuntu:
```bash
sudo apt update
sudo apt install libmumps-dev libscalapack-mpi-dev mpi-default-dev
```
2. Compile Trilinos with additional TPLs updated (see dependencies/current/trilinos/install.sh from #1813 for the required changes)
3. Compile 4C with the newly compiled Trilinos from step 2
4. Adapt your input file:
```bash
SOLVER 1:
  SOLVER: Mumps
```

lkoeglmeier (dismissed):
LGTM :)

src/core/linear_solver/src/method/4C_linear_solver_method.hpp
```diff
@@ -45,6 +45,7 @@ namespace Core::LinearSolver
   enum class SolverType
   {
     KLU2,     ///< Amesos direct solver using KLU2
+    Mumps,    ///< Amesos direct solver using MUMPS
```
mayrmt: Just a side note: The official spelling is "MUMPS" in all caps. I am fine either way.

mayrmt (dismissed):

src/core/linear_solver/src/method/4C_linear_solver_method.hpp
c-p-schmidt: True, I'll anyway need to push again, so I'll also adapt it.

src/core/linear_solver/src/method/4C_linear_solver_method_direct.cpp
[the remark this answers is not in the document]
c-p-schmidt: The explanation by Copilot within the Trilinos project suggests that we should add it, not only for consistency but for functionality.
https://github.com/copilot/share/4a1c522a-48e4-8403-8900-324e607b21cc

maxfirmbach:
We should think about this at some point:
- Rename `Belos` to `iterative`
- Include a new enum `direct`
=> SOLVER: direct / iterative
- Include something like METHOD: (direct: MUMPS, UMFPACK, ...), (iterative: CG; GMRES) etc.

This would make things a bit easier in the code I guess.

PR #1818, opened 2026-03-06
Clean up remodelfibers

tuchpaul:
Title: Clean up of the remodel fibers

## Description and Context
Adapted the naming of several helper functions to match the code style discussed in PR #1805.
Removed redundant function calls in `ExplicitRemodelFiber`
Change the signature of `integrate_local_evolution_equations_implicit` to `void`, since the output wasn't used.

### DIFF
--- a/src/mixture/src/4C_mixture_remodelfiber-internal.hpp
+++ b/src/mixture/src/4C_mixture_remodelfiber-internal.hpp
@@ -174,11 +174,8 @@ namespace Mixture
        * @brief Integrate the local evolution equation with an implicit time integration scheme.
        *
        * @param dt (in) : timestep
-       *
-       * @return Derivative of the residuum of the time integration scheme w.r.t. growth scalar
-       * and lambda_r
        */
-      Core::LinAlg::Matrix<2, 2, T> integrate_local_evolution_equations_implicit(T dt);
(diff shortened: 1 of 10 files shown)


src/mixture/src/4C_mixture_remodelfiber-internal.hpp
```diff
@@ -168,11 +168,8 @@ namespace Mixture
        * @brief Integrate the local evolution equation with an implicit time integration scheme.
        *
        * @param dt (in) : timestep
-       *
-       * @return Derivative of the residuum of the time integration scheme w.r.t. growth scalar
-       * and lambda_r
        */
-      Core::LinAlg::Matrix<2, 2, T> integrate_local_evolution_equations_implicit(T dt);
+      void integrate_local_evolution_equations_implicit(T dt);
```
amgebauer, line 178: Are you sure that this is possible? Isn't it needed for the linearization?

amgebauer: I see, all tests pass. Obviously, I changed the way to evaluate the linearization and forgot to adapt the return type here :thinking:

c-p-schmidt (approved):

PR #1819, opened 2026-03-06
[structure] Add partial krylov projector

maxfirmbach:
## Description and Context

This PR enables a Krylov projector to be only defined on a part of the discretization e.g. if several bodies are in play, or embedded independent structures occur, which are not explicitly Dirichlet constraint, thus containing rigid body modes. In this case the nullspace is only build on the given condition map and inserted into the Krylov kernel vector, which is zero everywhere else, thus the projection is only taking place on a part of the domain.

A small test is added, which already works for the nullspace constraint. In addition the code for the Krylov projector construction is simplified.


src/structure_new/src/linear_solver/4C_structure_new_solver_factory.cpp
```diff
@@ -177,23 +178,43 @@ std::shared_ptr<Core::LinAlg::Solver> Solid::SOLVER::Factory::build_structure_li
...
+          const size_t my_length = constraint_space_column.local_length();
+          for (size_t j = 0; j < my_length; j++)
           {
-            ci.get_values()[j] = ni.get_values()[j];
+            constraint_space_column.get_values()[j] = nullspace_column.get_values()[j];
           }
         }
+
+        // sort vector of nullspace data into kernel vector c_
+        Core::LinAlg::export_to(constraint_space, *c);
```
mayrmt: `c_` vs. `c`?

maxfirmbach: Fixed.

PR #1824, opened 2026-03-07
Use enum classes and parameter input in TSI

sebproell:
Make a few enums into enum classes and use `parameter<Enum>` instead of `deprecated_selection`.

I was experimenting to see how hard it would be to remove the remaining `deprecated_selection` occurrences with scripts + agents. If one wants to keep backwards compatibility, there will be problems when the input string is a reserved C++ keyword (like `and`, `or`, `explicit`). We could add a path for these cases in the InputSpec, but I am not yet sure this is the best approach.

### DIFF
--- a/src/tsi/4C_tsi_input.hpp
+++ b/src/tsi/4C_tsi_input.hpp
@@ -31,18 +31,25 @@ namespace TSI
     Monolithic
   };
 
+  //! Type of coupling variable for TSI problems
+  enum class CouplingVariable
+  {
+    Displacement,
+    Temperature
+  };
+
   //! @name Solution technique and related stuff
 
   //! type of norm to check for convergence
-  enum ConvNorm
+  enum class ConvNorm
   {
-    convnorm_abs,  //!< absolute norm
-    convnorm_rel,  //!< relative norm of TSI problem with initial TSI rhs
-    convnorm_mix   //!< mixed absolute-relative norm
+    Abs,  //!< absolute norm
+    Rel,  //!< relative norm of TSI problem with initial TSI rhs
+    Mix   //!< mixed absolute-relative norm
   };
 
   //! type of norm to check for convergence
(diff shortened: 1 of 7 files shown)


maxfirmbach (approved):

src/tsi/4C_tsi_input.hpp
```diff
@@ -31,18 +31,25 @@ namespace TSI
...
+    Abs,  //!< absolute norm
+    Rel,  //!< relative norm of TSI problem with initial TSI rhs
+    Mix   //!< mixed absolute-relative norm
   };
 
   //! type of norm to check for convergence
-  enum BinaryOp
+  enum class BinaryOp
   {
     bop_and,               //!< and
```
c-p-schmidt, line 54: Why did you not remove the bop_?

src/tsi/4C_tsi_input.hpp
```diff
@@ -53,14 +60,14 @@ namespace TSI
...
-    soltech_ptc,         //!< pseudo transient continuation nonlinear iteration
+    fullnewton,  //!< full Newton-Raphson iteration
+    ptc,         //!< pseudo transient continuation nonlinear iteration
   };
 
   //! type of line-search strategy
-  enum LineSearch
+  enum class LineSearch
   {
     LS_none = 0,   //!< no line search
```
c-p-schmidt, line 72: Also here the LS_

maxfirmbach:
We should go for `NOX` in `thermo` to get rid of all these enums related to the nonlinear solution process :sweat_smile:.

src/tsi/4C_tsi_input.hpp
sebproell: I wanted to make the enum values identical to the input, but the input uses values like "and" which are not possible as enum values.

PR #1835, opened 2026-03-09
Add option to include bodyforce contributions in the solid field in porofluid_pressure_based* problem types

lkoeglmeier:
This pr adds the option to include body force contributions, for example gravity, in the solid equations (2D and 3D) of the porofluid_pressure_based framework.

### DIFF
--- a/src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_lib.hpp
+++ b/src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_lib.hpp
@@ -41,6 +41,7 @@
 #include <cstdint>
 #include <numeric>
 #include <ranges>
+#include <span>
 
 FOUR_C_NAMESPACE_OPEN
 
@@ -61,35 +62,37 @@ namespace
       const double determinant_deformation_gradient, const Mat::FluidPoroMultiPhase& porofluidmat,
       const int numfluidphases)
   {
+    // get initial volfrac
+    const double initial_volfrac =
+        PoroPressureBased::ElementUtils::
+            get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
+                porofluidmat, numfluidphases)
+                .initial_volfrac();
+
+    // get volfrac scaling parameter deformation
+    const double scaling_parameter_deformation =
+        PoroPressureBased::ElementUtils::
+            get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
+                porofluidmat, numfluidphases)
+                .scaling_parameter_deformation();
+
+    // get volfrac scaling parameter pressure
+    const double scaling_parameter_pressure =
+        PoroPressureBased::ElementUtils::
+            get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
+                porofluidmat, numfluidphases)
+                .scaling_parameter_pressure();
+
     // pressure of air (always first phase in multiphase porespace)
     if (fluidmultiphase_phi_at_gp[0] / volfracpressure <= 1.0)
     {
-      return PoroPressureBased::ElementUtils::
-                 get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                     porofluidmat, numfluidphases)
-                     .initial_volfrac() *
-             pow(determinant_deformation_gradient,
-                 PoroPressureBased::ElementUtils::
-                     get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                         porofluidmat, numfluidphases)
-                         .scaling_parameter_deformation());
+      return initial_volfrac * pow(determinant_deformation_gradient, scaling_parameter_deformation);
     }
     else
     {
-      return PoroPressureBased::ElementUtils::
-                 get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                     porofluidmat, numfluidphases)
-                     .initial_volfrac() *
-             pow(determinant_deformation_gradient,
-                 PoroPressureBased::ElementUtils::
-                     get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                         porofluidmat, numfluidphases)
-                         .scaling_parameter_deformation()) *
-             pow((fluidmultiphase_phi_at_gp[0] / volfracpressure),
-                 PoroPressureBased::ElementUtils::
-                     get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                         porofluidmat, numfluidphases)
-                         .scaling_parameter_pressure());
+      return initial_volfrac *
+             pow(determinant_deformation_gradient, scaling_parameter_deformation) *
+             pow((fluidmultiphase_phi_at_gp[0] / volfracpressure), scaling_parameter_pressure);
     }
   }
 }  // namespace
@@ -314,52 +317,12 @@ namespace Discret::Elements
 
 
         // so far only one volfrac with closing relation for blood lung is possible
-        solid_poro_fluid_additional_porous_network_variables.volfrac.at(0) = std::invoke(
-            [&]()
-            {
-              // pressure of air (always first phase in multiphase porespace)
-              if (fluid_phase_phi_at_gp[0] /
-                      solid_poro_fluid_additional_porous_network_variables.volfrac_pressure[0] <=
-                  1.0)
-              {
-                return (
-                    PoroPressureBased::ElementUtils::
-                        get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                            fluidmulti_mat, solidporo_fluid_properties
-                                                .number_of_fluid_phases_in_multiphase_porespace_)
-                            .initial_volfrac() *
-                    pow(J, PoroPressureBased::ElementUtils::
-                               get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                                   fluidmulti_mat,
-                                   solidporo_fluid_properties
-                                       .number_of_fluid_phases_in_multiphase_porespace_)
-                                   .scaling_parameter_deformation()));
-              }
-              else
-              {
-                return (
-                    PoroPressureBased::ElementUtils::
-                        get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                            fluidmulti_mat, solidporo_fluid_properties
-                                                .number_of_fluid_phases_in_multiphase_porespace_)
-                            .initial_volfrac() *
-                    pow(J, PoroPressureBased::ElementUtils::
-                               get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                                   fluidmulti_mat,
-                                   solidporo_fluid_properties
-                                       .number_of_fluid_phases_in_multiphase_porespace_)
-                                   .scaling_parameter_deformation()) *
-                    pow((fluid_phase_phi_at_gp[0] /
-                            solid_poro_fluid_additional_porous_network_variables
-                                .volfrac_pressure[0]),
-                        PoroPressureBased::ElementUtils::
-                            get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
-                                fluidmulti_mat,
-                                solidporo_fluid_properties
-                                    .number_of_fluid_phases_in_multiphase_porespace_)
-                                .scaling_parameter_pressure()));
-              }
-            });
+        solid_poro_fluid_additional_porous_network_variables.volfrac.at(0) =
+            calculate_volfrac_from_closing_relation_blood_lung(fluid_phase_phi_at_gp,
+                solid_poro_fluid_additional_porous_network_variables.volfrac_pressure[0], J,
+                fluidmulti_mat,
+                solidporo_fluid_properties.number_of_fluid_phases_in_multiphase_porespace_);
+
         solid_poro_fluid_additional_porous_network_variables.sumaddvolfrac =
             solid_poro_fluid_additional_porous_network_variables.volfrac.at(0);
       }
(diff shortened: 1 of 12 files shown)


tests/input_files/porofluid_pressure_based_elast_scatra_3D_hex8_volfrac_closing_relation_lung_bodyforce_solid_mono.4C.yaml
```diff
@@ -0,0 +1,534 @@
...
+  SKIPINITDER: true
+  LINEAR_SOLVER: 1
+SCALAR TRANSPORT DYNAMIC/STABILIZATION:
+  STABTYPE: "no_stabilization"
+  EVALUATION_TAU: "integration_point"
+  EVALUATION_MAT: "integration_point"
+SCALAR TRANSPORT DYNAMIC/NONLINEAR:
+  ITEMAX: 100
+  CONVTOL: 1e-08
+  ABSTOLRES: -1
```
c-p-schmidt: Why a negative value? That looks a bit strange to me.

tests/input_files/porofluid_pressure_based_elast_scatra_3D_hex8_volfrac_closing_relation_lung_bodyforce_solid_mono.4C.yaml
```diff
@@ -0,0 +1,534 @@
...
+  NEGLECTINERTIA: true
+STRUCTURAL DYNAMIC/ONESTEPTHETA:
+  THETA: 1
+porofluid_dynamic:
+  initial_condition:
+    type: by_condition
+porofluid_elasticity_dynamic:
+  solve_structure: true
+  coupling_scheme: twoway_monolithic
+  body_force: [0.0, 9.81e-1, 0.0]
```
c-p-schmidt, line 40: Is the unit system so different compared to the 3D test case, or did you want to choose such different values? I think there you had 9810 instead of 0.981 here.

src/w1/4C_w1_poro_evaluate.cpp
```diff
@@ -25,6 +25,37 @@
 
 FOUR_C_NAMESPACE_OPEN
 
+std::optional<Core::LinAlg::Tensor<double, 2>> get_bodyforce_contribution_from_input()
```
c-p-schmidt, line 28: You could consider doing this only once at the beginning of the simulation and storing it as a class member, since it cannot change during the simulation.

Currently, you call it several times.

src/w1/4C_w1_poro_evaluate.cpp
```diff
@@ -797,9 +844,22 @@ void Discret::Elements::Wall1Poro<distype>::gauss_point_loop_pressure_based(
...
         compute_sol_pressure_at_gp(solidporo_fluid_properties.number_of_fluid_dofs_per_node_,
             solidporo_fluid_properties.number_of_fluid_phases_in_multiphase_porespace_, phi_at_gp);
+
     // recalculate for the case of volume fractions
     if (hasvolfracs)
     {
+      // this is only possible if we have an incompressible solid because the porosity of the
+      // structure is calculated with the solidpressure which is calculated without
+      // contributions from the additional porous network
+      FOUR_C_ASSERT_ALWAYS(
```
c-p-schmidt: I think this check you could also do only once upon construction, init, or setup, so at the beginning of the simulation.

src/w1/4C_w1_poro_evaluate.cpp
```diff
@@ -1071,11 +1133,92 @@ void Discret::Elements::Wall1Poro<distype>::fill_matrix_and_vectors_pressure_bas
   const double fac1 = -detJ_w * press;
   const double fac2 = fac1 * J;
 
+
+  // get primary variables from additional porous network
+  FourC::Discret::Elements::SolidPoroFluidAdditionalPorousNetworkVariables
```
c-p-schmidt: Maybe?
```suggestion
auto
```

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_lib.hpp
```diff
@@ -244,6 +244,133 @@ namespace Discret::Elements
...
+        solid_poro_fluid_additional_porous_network_variables.volfrac.at(0) = std::invoke(
+            [&]()
+            {
+              // pressure of air (always first phase in multiphase porespace)
+              if (fluid_phase_phi_at_gp[0] /
+                      solid_poro_fluid_additional_porous_network_variables.volfrac_pressure[0] <=
+                  1.0)
+              {
+                return (
+                    PoroPressureBased::ElementUtils::
```
c-p-schmidt: I think you could precalculate this once and reuse it at several places in this lambda to increase readability.

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_lib.hpp
```diff
@@ -914,21 +1041,89 @@ namespace Discret::Elements
...
+    // relation for blood lung is possible
+    const double volfracpressure = fluid_phase_phi_at_gp[numfluidphases];
+
+    const double dvolfrac_dDetDefGrad = std::invoke(
+        [&]()
+        {
+          if ((fluid_phase_phi_at_gp[0] / volfracpressure) <= 1.0)
+          {
+            return (PoroPressureBased::ElementUtils::
+                        get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
```
c-p-schmidt: Also here.

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_pressure_based.cpp
```diff
@@ -106,19 +107,37 @@ void Discret::Elements::SolidPoroPressureBasedEleCalc<celltype>::evaluate_nonlin
...
         // derivative of press w.r.t. displacements (only in case of volfracs)
         Core::LinAlg::Matrix<1, num_dof_per_ele_> dSolidpressure_dDisp(
             Core::LinAlg::Initialization::zero);
 
         if (hasvolfracs)
         {
+          // this is only possible if we have an incompressible solid because the porosity of the
+          // structure is calculated with the solidpressure which is calculated without
+          // contributions from the additional porous network
+          FOUR_C_ASSERT_ALWAYS(
```
c-p-schmidt: Maybe this check can also be moved to the beginning of the simulation?

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_pressure_based.hpp
```diff
@@ -38,9 +39,17 @@ namespace Discret
      public:
       SolidPoroPressureBasedEleCalc();
 
-      void evaluate_nonlinear_force_stiffness(const Core::Elements::Element& ele,
-          Mat::StructPoro& porostructmat, Mat::FluidPoroMultiPhase& porofluidmat,
-          const Inpar::Solid::KinemType& kinematictype,
+      void add_solidpressure_contribution_to_nonlinear_force_stiffness(
```
c-p-schmidt, line 46: Some documentation for the new methods would be nice.

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_pressure_based.cpp
```diff
@@ -325,5 +325,34 @@ Mat::FluidPoroMultiPhase& Discret::Elements::SolidPoroPressureBased::fluid_poro_
...
+        static_cast<int>(
+            (solid_poro_params.get<std::optional<std::vector<double>>>("body_force"))->size()) ==
+            Global::Problem::instance()->n_dim(),
+        "The dimension of your bodyforce vector and the dimension of the problem must be equal!");
+
+    Core::LinAlg::Tensor<double, 3> bodyforce{};
+    for (int idim = 0; idim < 3; idim++)
+    {
+      bodyforce(idim) =
+          solid_poro_params.get<std::optional<std::vector<double>>>("body_force").value()[idim];
```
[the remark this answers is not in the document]
c-p-schmidt: As the "3" is explicitly used in several locations and it is only used in 3D, why not simply changing:
 `Global::Problem::instance()->n_dim()` with 3 in the assertion?

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_pressure_based_evaluate.cpp
```diff
@@ -59,11 +59,23 @@ int Discret::Elements::SolidPoroPressureBased::evaluate(Teuchos::ParameterList&
...
               {
-                interface->evaluate_nonlinear_force_stiffness(*this, this->struct_poro_material(),
-                    this->fluid_poro_material(), this->get_ele_kinematic_type(), discretization, la,
-                    params, &elevec1, &elemat1);
+                interface->add_solidpressure_contribution_to_nonlinear_force_stiffness(*this,
+                    this->struct_poro_material(), this->fluid_poro_material(),
+                    this->get_ele_kinematic_type(), discretization, la, params, &elevec1, &elemat1);
               },
               solidporo_press_based_calc_variant_);
+          if (this->get_bodyforce_contribution_from_input().has_value())
```
c-p-schmidt: The body force is constant throughout the simulation, as the method's name suggests.
So also here you could only calculate it once and store it instead of redoing it at every "evaluate" call.

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_lib.hpp
```diff
@@ -244,6 +244,133 @@ namespace Discret::Elements
...
+                               get_single_vol_frac_pressure_blood_lung_mat_from_multi_material(
+                                   fluidmulti_mat,
+                                   solidporo_fluid_properties
+                                       .number_of_fluid_phases_in_multiphase_porespace_)
+                                   .scaling_parameter_deformation()));
+              }
+              else
+              {
+                return (
+                    PoroPressureBased::ElementUtils::
```
c-p-schmidt: Here you probably forgot to do the pre-calculation that you nicely implemented in some other locations and IMO significantly improved the readability.

lkoeglmeier: true! Thanks!

c-p-schmidt (approved):

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_lib.hpp
```diff
@@ -2264,6 +2359,39 @@ namespace Discret::Elements
...
+  inline std::vector<double> compute_porofluid_saturation_in_multiphase_porespace(
+      const int numfluidphases, const std::vector<double>& fluidmultiphase_phi_at_gp,
+      Mat::FluidPoroMultiPhase& porofluidmat)
+  {
+    // initialize auxiliary variables
+    std::vector<double> genpress(numfluidphases, 0.0);
+    std::vector<double> sat(numfluidphases, 0.0);
+    std::vector<double> press(numfluidphases, 0.0);
+    std::vector<double> fluidphi(
+        fluidmultiphase_phi_at_gp.data(), fluidmultiphase_phi_at_gp.data() + numfluidphases);
```
amgebauer: Why do you need to do a copy here? Does `fluidmultiphase_phi_at_gp` contain more elements and you just copy the first `numfluidphases`? In that case, you could also just use a `std::span` to view only the first `numfluidphases` (in that case, you would need to adapt the functions taking the vector).

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_calc_pressure_based.cpp
```diff
@@ -211,6 +235,232 @@ void Discret::Elements::SolidPoroPressureBasedEleCalc<celltype>::evaluate_nonlin
...
+          }
+        }
+
+        Core::LinAlg::Matrix<num_dof_per_ele_, 1> bodyforce(Core::LinAlg::Initialization::zero);
+        for (int j = 0; j < num_dim_; ++j)
+        {
+          for (int i_node = 0; i_node < num_nodes_; ++i_node)
+          {
+            bodyforce(i_node * num_dim_ + j, 0) +=
+                shape_functions.shapefunctions_(i_node) * bodyforce_contribution.value()(j);
```
amgebauer: Maybe I have missed it, but `bodyforce_contribution` is optional, but if you pass a std::nullopt, you will end up in a (maybe cryptic) error here. Couldn't you just pass the Tensor directly?

src/solid_poro_3D_ele/4C_solid_poro_3D_ele_pressure_based_evaluate.cpp
```diff
@@ -116,11 +140,23 @@ int Discret::Elements::SolidPoroPressureBased::evaluate(Teuchos::ParameterList&
...
               },
               solidporo_press_based_calc_variant_);
+          if (this->get_possible_bodyforce_contribution().has_value())
+          {
+            std::visit(
+                [&](auto& interface)
+                {
+                  interface->add_bodyforce_contribution_to_nonlinear_force_stiffness(*this,
+                      this->struct_poro_material(), this->fluid_poro_material(),
+                      this->get_ele_kinematic_type(), this->get_possible_bodyforce_contribution(),
```
amgebauer: Related to my comment above. Here you know that you have a body force, so you can directly pass the tensor (and remove the optional in the method above):
```suggestion
                      this->get_ele_kinematic_type(), *this->get_possible_bodyforce_contribution(),
```

lkoeglmeier:
Thank you @c-p-schmidt and @amgebauer for your helpful comments!

PR #1842, opened 2026-03-10
Centralize Global::Problem access in porofluid_pressure_based

bgoderbauer:
This PR centralizes `Global::Problem` access in `porofluid_pressure_based` and removes direct singleton usage from the internal algorithm path. Please let me know what you think and if you have suggestions for improvement or better naming :smiley:. 

## Changes

- Added `PorofluidAlgorithmDeps` (`4C_porofluid_pressure_based_algorithm_dependencies.hpp/.cpp`) as the shared dependency wrapper for algorithm code.
- Added `make_algorithm_deps_from_problem(Global::Problem&)` as the single adapter that maps `Global::Problem` data/callbacks into `PorofluidAlgorithmDeps`.
- Kept `PorofluidPressureBasedMainContext` local in `4C_porofluid_pressure_based_dyn.cpp` for dyn/run orchestration state only.
- Updated algorithm creation call sites (including porofluid-elast path) to pass `PorofluidAlgorithmDeps`.
- Removed direct `Global::Problem` usage from internal `porofluid_pressure_based` components (algorithm/utils/meshtying/resulttest).

## Wrappers

Two wrappers, one for 'PoroFluidAlgorithm', one a step above for the dyn.
- `PorofluidAlgorithmDeps`: explicit inputs/services needed by algorithm internals.
- `PorofluidPressureBasedMainContext`: local dyn execution context (discretizations, params, comm, test execution hooks).

#1796

### DIFF
--- a/src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm_dependencies.cpp
+++ b/src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm_dependencies.cpp
@@ -1,0 +1,45 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#include "4C_porofluid_pressure_based_algorithm_dependencies.hpp"
+
+#include "4C_global_data.hpp"
+#include "4C_utils_function.hpp"
+
+#include <Teuchos_StandardParameterEntryValidators.hpp>
+
+FOUR_C_NAMESPACE_OPEN
+
+PoroPressureBased::PorofluidAlgorithmDeps PoroPressureBased::make_algorithm_deps_from_problem(
+    Global::Problem& problem)
+{
+  const bool has_artery_discretization = problem.does_exist_dis("artery");
+  const std::shared_ptr<Core::FE::Discretization> artery_discretization =
+      has_artery_discretization ? problem.get_dis("artery") : nullptr;
+
+  return PorofluidAlgorithmDeps{
+      .spatial_dimension = problem.n_dim(),
+      .restart_step = problem.restart(),
+      .verbosity =
+          Teuchos::getIntegralValue<Core::IO::Verbositylevel>(problem.io_params(), "VERBOSITY"),
+      .input_control_file = problem.input_control_file(),
+      .output_control_file = problem.output_control_file(),
+      .artery_discretization = artery_discretization,
+      .runtime_vtk_output_parameters = &problem.io_params().sublist("RUNTIME VTK OUTPUT"),
+      .artery_dynamic_parameters =
+          has_artery_discretization ? &problem.arterial_dynamic_params() : nullptr,
+      .function_manager = &problem.function_manager(),
+      .function_of_space_time_by_id =
+          [&problem](const int function_id) -> const Core::Utils::FunctionOfSpaceTime&
+      { return problem.function_by_id<Core::Utils::FunctionOfSpaceTime>(function_id); },
+      .solver_params_by_id = problem.solver_params_callback(),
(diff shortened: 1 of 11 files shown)


src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm_dependencies.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "4C_porofluid_pressure_based_algorithm_dependencies.hpp"
+
+#include "4C_global_data.hpp"
+#include "4C_utils_function.hpp"
+
+#include <Teuchos_StandardParameterEntryValidators.hpp>
+
+FOUR_C_NAMESPACE_OPEN
+
+PoroPressureBased::PorofluidAlgorithmDeps PoroPressureBased::make_algorithm_deps_from_problem(
```
lkoeglmeier, line 17:
```suggestion
PoroPressureBased::PorofluidAlgorithmDeps PoroPressureBased::evaluate_algorithm_deps_from_problem(
```

src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm_dependencies.hpp
```diff
@@ -0,0 +1,67 @@
...
+    std::shared_ptr<Core::IO::InputControl> input_control_file;
+    std::shared_ptr<Core::IO::OutputControl> output_control_file;
+    std::shared_ptr<Core::FE::Discretization> artery_discretization;
+    const Teuchos::ParameterList* runtime_vtk_output_parameters = nullptr;
+    const Teuchos::ParameterList* artery_dynamic_parameters = nullptr;
+    const Core::Utils::FunctionManager* function_manager = nullptr;
+    std::function<const Core::Utils::FunctionOfSpaceTime&(int)> function_of_space_time_by_id;
+    std::function<const Teuchos::ParameterList&(int)> solver_params_by_id;
+    std::function<void(std::shared_ptr<Core::Utils::ResultTest>)> add_field_test;
+  };
```
lkoeglmeier, line 60: maybe we could add a variable` bool is_set_` that is only used in debug mode and set to true if the struct has been evaluated to ensure that the correct values are used? But that's just a suggestion.

src/porofluid_pressure_based/4C_porofluid_pressure_based_dyn.cpp
```diff
@@ -14,138 +14,179 @@
...
+}  // namespace
 
+/*-------------------------------------------------------------------------------*
+ | Main control routine for poro fluid multiphase problems           vuong 08/16 |
+ *-------------------------------------------------------------------------------*/
+void porofluid_pressure_based_dyn(int restart)
+{
+  const PorofluidPressureBasedMainContext context =
+      make_main_context_from_problem(*Global::Problem::instance());
+  run_porofluid_pressure_based(context, restart);
```
lkoeglmeier, line 189: maybe we do not need this additional run_* function and can leave everything in the dyn to be consistent with the other folders?

src/porofluid_pressure_based/4C_porofluid_pressure_based_algorithm_dependencies.hpp
lkoeglmeier: resolved because discussed offline ;)

src/porofluid_pressure_based/4C_porofluid_pressure_based_dyn.cpp
lkoeglmeier: resolved because discussed offline ;)

bgoderbauer: As discussed offline, this acts just as an interface between the current setup based on Global::Problem and the module implementation. Here's the last time Global::Problem is called, making it easier in the future to separate the functionality and maintain the correct dependency chain.

PR #1847, opened 2026-03-11
[SPH] Optimize reading of the additional particle parameters

vovannikov:
## Description and Context
I decided to remove the repeating if branches with a loop over the map of possible options, just to reduce code repetition. I am planning to add another parameter here later.

### DIFF
--- a/src/particle/src/engine/4C_particle_engine_particlereader.cpp
+++ b/src/particle/src/engine/4C_particle_engine_particlereader.cpp
@@ -61,46 +61,22 @@ void Particle::read_particles(Core::IO::InputFile& input, const std::string& sec
         Particle::StateEnum particlestate;
         std::vector<double> state;
 
+        // optional particle parameters
+        const std::unordered_map<std::string, Particle::ParticleState> additional_states = {
+            {"RAD", Particle::Radius}, {"RIGIDCOLOR", Particle::RigidBodyColor},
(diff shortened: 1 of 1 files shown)


src/particle/src/engine/4C_particle_engine_particlereader.cpp
```diff
@@ -61,46 +61,22 @@ void Particle::read_particles(Core::IO::InputFile& input, const std::string& sec
...
         while (!parser.at_end())
         {
-          auto next = parser.peek();
-          // optional particle radius
-          if (next == "RAD")
-          {
-            particlestate = Particle::Radius;
-            parser.consume("RAD");
+          const auto next = parser.peek();
+          statelabel = next;
```
ppraegla, line 72: `statelabel` was never set before. Thanks for fixing this 👍

src/particle/src/engine/4C_particle_engine_particlereader.cpp
georghammerl, line 88: Out of curiosity: Is this `continue` a valid path through the code? Does this mean, a key (e.g. RAD, RIGIDCOLOR,...) is given but no value is following in this input line?

vovannikov: Looks so. The logic seem to say that if the state value is not provided, then skip this state, as it is optional. However, if you forget to provide it, the parser will crash earlier.

PR #1848, opened 2026-03-11
Relocate viscoplastic material utilities to specific service files

dragos-ana:
- Moved several helper types and structs for the viscoplastic  material to the service utilities under a specific namespace.
A .cpp service file was added since several other utilities with lengthier implementations will be added in the future.
Accordingly, the viscoplastic material now stores instances of such helper objects instead of their declarations / implementations.
- Viscoplastic laws: updated overflow error checks with maximum plastic strain increments (and derivative increments), that can be specified by the user in the input file.
- Update input parameters and their types: new parameters (matrix function methods), and other more-straightforward types (consistent enums, instead of strings that are then converted to booleans, see MAT_BEHAVIOR instead of ANISOTROPY). 
- Added lengthier error message for the viscoplastic material that can be used subsequently for debugging.
- Added further evaluation types for state quantities StateQuantities and their derivatives StateQuantityDerivatives.
These can be used for partial evaluations, i.e., to stop once the relevant information has been evaluated (for example when we only need the equivalent stress or the plastic strain rate).
They will be used in subsequent PRs.
- Bugfix: additional NaN checks for stress and plastic strain to avoid overflow errors in that case.
- Bugfix: linearization now explicitly re-evaluates the state quantities. This is explicitly required in the case that the
inverse inelastic defgrad evaluation is circumvented (see #131).
- Refresh affected unit tests and YAML input files to match the new refactored code structure. 
In the unit test of the inelastic deformation factors, we now explicitly name the utilized viscoplastic law (Reformulated Johnson-Cook), since further tests may be added with other upcoming viscoplastic laws.

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
@@ -24,6 +24,7 @@
 #include "4C_porofluid_pressure_based_elast_scatra_input.hpp"
 
 #include <filesystem>
+#include <optional>
 #include <string>
 
 FOUR_C_NAMESPACE_OPEN
@@ -2795,19 +2796,19 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
             parameter<int>(
                 "FIBER_READER_ID", {.description = "MAT ID of the used fiber direction reader for "
                                                    "transversely isotropic behavior"}),
-            parameter<double>(
+            parameter<std::optional<double>>(
                 "YIELD_COND_A", {.description = "transversely isotropic version of the Hill(1948) "
                                                 "yield condition: parameter A, "
                                                 "following "
                                                 "the notation in Dafalias 1989, International "
                                                 "Journal of Plasticity, Vol. 5"}),
-            parameter<double>(
+            parameter<std::optional<double>>(
                 "YIELD_COND_B", {.description = "transversely isotropic version of the Hill(1948) "
                                                 "yield condition: parameter B, "
                                                 "following "
                                                 "the notation in Dafalias 1989, International "
                                                 "Journal of Plasticity, Vol. 5"}),
-            parameter<double>(
+            parameter<std::optional<double>>(
                 "YIELD_COND_F", {.description = "transversely isotropic version of the Hill(1948) "
                                                 "yield condition: parameter F, "
                                                 "following "
--- a/src/mat/4C_mat_inelastic_defgrad_factors.cpp
+++ b/src/mat/4C_mat_inelastic_defgrad_factors.cpp
@@ -37,6 +37,7 @@
 
 #include <map>
 #include <memory>
+#include <optional>
 #include <stdexcept>
 #include <string>
 #include <utility>
@@ -413,7 +414,6 @@ namespace
         Core::LinAlg::EigenvalInterpolationType::LOG, interp_param_list};
   }
 
-
 }  // namespace
 
 
@@ -589,9 +589,9 @@ Mat::PAR::InelasticDefgradTransvIsotropElastViscoplast::
     : Parameter(matdata),
       viscoplastic_law_id_(matdata.parameters.get<int>("VISCOPLAST_LAW_ID")),
       fiber_reader_gid_(matdata.parameters.get<int>("FIBER_READER_ID")),
-      yield_cond_a_(matdata.parameters.get<double>("YIELD_COND_A")),
-      yield_cond_b_(matdata.parameters.get<double>("YIELD_COND_B")),
-      yield_cond_f_(matdata.parameters.get<double>("YIELD_COND_F")),
+      yield_cond_a_(matdata.parameters.get<std::optional<double>>("YIELD_COND_A").value_or(0.0)),
+      yield_cond_b_(matdata.parameters.get<std::optional<double>>("YIELD_COND_B").value_or(0.0)),
+      yield_cond_f_(matdata.parameters.get<std::optional<double>>("YIELD_COND_F").value_or(0.0)),
       mat_behavior_(matdata.parameters.get<ViscoplastUtils::MatBehavior>("MAT_BEHAVIOR")),
       timint_type_(
           matdata.parameters.get<ViscoplastUtils::TimIntType>("TIME_INTEGRATION_HIST_VARS")),
@@ -612,11 +612,25 @@ Mat::PAR::InelasticDefgradTransvIsotropElastViscoplast::
           matdata.parameters.get<Core::LinAlg::GenMatrixLogFirstDerivCalcMethod>(
               "MATRIX_LOG_DERIV_CALC_METHOD"))
 {
-  // consistency checks
+  // consistency check: number of substepping halving procedures
   if (max_substepping_halve_num_ < 0)
     FOUR_C_THROW(
         "Maximum number of times the global time step can be halved in the substepping procedure "
         "must be >= 0!");
+
+  // consistency check: yield parameters in case of transversely-isotropic behavior
+  const bool all_yield_cond_param_specified =
+      matdata.parameters.get<std::optional<double>>("YIELD_COND_A").has_value() &&
+      matdata.parameters.get<std::optional<double>>("YIELD_COND_B").has_value() &&
+      matdata.parameters.get<std::optional<double>>("YIELD_COND_F").has_value();
+  if (mat_behavior_ ==
+          InelasticDefgradTransvIsotropElastViscoplastUtils::MatBehavior::transv_isotropic &&
+      !all_yield_cond_param_specified)
+  {
+    FOUR_C_THROW(
+        "You are attempting to simulate transversely isotropic behavior but have not specified all "
+        "yield function parameters!");
+  }
 }
 
 
@@ -1618,30 +1632,8 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::InelasticDefgradTransvIsotrop
   // set minimum substep length
   time_step_tracker_.min_dt = 0.0;
 
-  // ----- set last_ and current_ variables referring to values at different time instants
-  // ----- for now: the number of Gauss points is unknown -> we set the values only for 1
-  // Gauss point and update the number of Gauss points in the setup method
-
-  // default values of the inverse plastic deformation gradient: unit tensor
-  time_step_quantities_.last_plastic_defgrad_inverse.resize(1, const_non_mat_tensors.id3x3);
-  time_step_quantities_.current_plastic_defgrad_inverse.resize(
-      1, const_non_mat_tensors.id3x3);  // value irrelevant at this point
-  time_step_quantities_.last_substep_plastic_defgrad_inverse.resize(1, const_non_mat_tensors.id3x3);
-
-  // update last_ and current_ values of the plastic strain
-  time_step_quantities_.last_plastic_strain.resize(1, 0.0);
-  time_step_quantities_.current_plastic_strain.resize(1, 0.0);  // value irrelevant at this point
-  time_step_quantities_.last_substep_plastic_strain.resize(1, 0.0);
-
-  // default values of the right CG tensor: unit tensor
-  time_step_quantities_.last_rightCG.resize(1, const_non_mat_tensors.id3x3);
-  time_step_quantities_.current_rightCG.resize(
-      1, const_non_mat_tensors.id3x3);  // value irrelevant at this point
-
-  // default value for the current deformation gradient: zero tensor \f$ \boldsymbol{0} f$ (to make
-  // sure that the inverse inelastic deformation gradient is evaluated in the first method call)
-  time_step_quantities_.current_defgrad.resize(
-      1, Core::LinAlg::Matrix<3, 3>{Core::LinAlg::Initialization::zero});
+  // initialize time step quantities
+  time_step_quantities_.init();
 }
 
 
@@ -1667,14 +1659,8 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::pre_evaluate(
   time_step_tracker_.min_dt =
       time_step_tracker_.dt / std::pow(2.0, parameter()->max_halve_number());
 
-  // set last substep values (last converged state) as the last time step values --> required, as
-  // these are used in the EvaluateAdditionalCMat method (in the case where there is no plastic
-  // deformation, these would not be updated correctly otherwise)
-  time_step_quantities_.last_substep_plastic_defgrad_inverse[gp_] =
-      time_step_quantities_.last_plastic_defgrad_inverse[gp_];
-  time_step_quantities_.last_substep_plastic_strain[gp_] =
-      time_step_quantities_.last_plastic_strain[gp_];
-
+  // call pre_evaluate method of the time step quantities
+  time_step_quantities_.pre_evaluate(gp);
   // call preevaluate method of the viscoplastic law
   viscoplastic_law_->pre_evaluate(gp);
 }
@@ -1683,7 +1669,7 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::pre_evaluate(
  *--------------------------------------------------------------------*/
 void Mat::InelasticDefgradTransvIsotropElastViscoplast::calculate_gamma_delta(
     const Core::LinAlg::Matrix<3, 3>& CeM, Core::LinAlg::Matrix<3, 1>& gamma,
-    Core::LinAlg::Matrix<8, 1>& delta)
+    Core::LinAlg::Matrix<8, 1>& delta) const
 {
   // compute principal values
   Core::LinAlg::Matrix<3, 1> prinv(Core::LinAlg::Initialization::zero);
@@ -1715,9 +1701,11 @@ ViscoplastUtils::StateQuantities
 Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
     const Core::LinAlg::Matrix<3, 3>& CM, const Core::LinAlg::Matrix<3, 3>& iFinM,
     const double plastic_strain, ViscoplastUtils::ErrorType& err_status, const double dt,
-    const ViscoplastUtils::StateQuantityEvalType& eval_type)
+    const ViscoplastUtils::StateQuantityEvalType& eval_type) const
 {
   ViscoplastUtils::StateQuantities state_quantities{};
+  state_quantities.eval_type = eval_type;
+
 
   // auxiliaries
   Core::LinAlg::Matrix<1, 1> temp1x1(Core::LinAlg::Initialization::zero);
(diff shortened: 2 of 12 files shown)


unittests/mat/vplast/4C_vplast_reform_johnsoncook_test.cpp
```diff
@@ -91,16 +92,16 @@ namespace
...
-    const bool log_substep = true;
+    // declare error status
+    Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType err_status =
+        Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType::no_errors;
 
     // compute solution from the viscoplasticity law
     double plastic_strain_rate_reformulated_JC =
         vplast_law_reformulated_JC_->evaluate_plastic_strain_rate(
-            equiv_stress_, equiv_plastic_strain_, 1.0, log_substep, err_status, false);
+            equiv_stress_, equiv_plastic_strain_, 1.0, 1.0e30, err_status, false);
```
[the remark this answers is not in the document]
dragos-ana: Adapted it to the default value `std::exp(30.0)`.

unittests/mat/vplast/4C_vplast_reform_johnsoncook_test.cpp
```diff
@@ -116,16 +117,16 @@ namespace
...
-    const bool log_substep = true;
+    // declare error status
+    Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType err_status =
+        Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType::no_errors;
 
     // compute solution from the viscoplasticity law
     Core::LinAlg::Matrix<2, 1> deriv_plastic_strain_rate_reformulated_JC =
         vplast_law_reformulated_JC_->evaluate_derivatives_of_plastic_strain_rate(
-            equiv_stress_, equiv_plastic_strain_, 1.0, log_substep, err_status, false);
+            equiv_stress_, equiv_plastic_strain_, 1.0, 1.0e30, err_status, false);
```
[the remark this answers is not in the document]
dragos-ana: See comment above

c-p-schmidt:
Can you please double-check why two commits are displayed?
The first one is a commit that is already in main (see https://github.com/4C-multiphysics/4C/pull/1832/commits), but with a different commit hash, unless I confused something.

So this looks strange. Can you pull the current main, rebase your branch against it, and then check that only your commit is there? As a final step, please force-push this corrected branch.

Thank you!

dragos-ana:
@c-p-schmidt Should be fine now! Thanks for noticing it

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2793,28 +2795,74 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                                "possible overflow "
+                                "errors",
+                    .default_value = std::exp(30.0)}),
+            parameter<int>("MAX_SUBSTEPPING_HALVE_NUM",
+                {.description = "maximum number of times the global time step can "
+                                "be halved in the substepping procedure (default: 10)",
+                    .default_value = 10}),
+            parameter<Core::LinAlg::MatrixExpCalcMethod>("MATRIX_EXP_CALC_METHOD",
+                {.description = "chosen computation method for matrix exponential",
+                    .default_value = Core::LinAlg::MatrixExpCalcMethod::default_method}),
```
c-p-schmidt: Is it possible to improve the naming? `default_method` doesn't indicate what is done.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2793,28 +2795,74 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+            parameter<Core::LinAlg::GenMatrixExpFirstDerivCalcMethod>(
+                "MATRIX_EXP_DERIV_CALC_METHOD",
+                {.description = "chosen computation method for the first derivative of the matrix "
+                                "exponential w.r.t. matrix",
+                    .default_value =
+                        Core::LinAlg::GenMatrixExpFirstDerivCalcMethod::default_method}),
```
c-p-schmidt: Same here.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2851,8 +2899,8 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
             parameter<int>("IDMATZEROSC", {.description = "material for lambda equal to zero"}),
             parameter<int>("IDMATUNITSC", {.description = "material for lambda equal to one"}),
         },
-        {.description =
-                "integration point based and scalar dependent interpolation between to materials"});
+        {.description = "integration point based and scalar dependent interpolation between to "
```
c-p-schmidt: Ah, it was only reformatted. Nevertheless, probably 2 is meant instead of to. ;-)
```suggestion
        {.description = "integration point based and scalar dependent interpolation between two "
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -20,60 +29,359 @@ FOUR_C_NAMESPACE_OPEN
...
+      no_errors,                ///< no errors
+      negative_plastic_strain,  ///< negative plastic strain which does not allow for evaluations
+                                ///< inside the viscoplasticity laws
+      overflow_error,  ///< overflow error of the term \f$ \Delta t \dot{\varepsilon}^{\text{p}} \f$
+                       ///< (and \f$ \mathsymbol{E}^{\text{p}}  = \exp(- \Delta t
+                       ///< \dot{\varepsilon}^{\text{p}} \mathsymbol{N}^{\text{p}}) \f$)
+      no_flow_resistance,            ///< the material has no flow resistance anymore, such that the
+                                     ///< evaluations model non-physical phenomena
+      no_plastic_incompressibility,  ///< no plastic incompressibility, meaning that the determinant
+                                     ///< of the inelastic defgrad is far from 1
```
c-p-schmidt: What does "far from 1" mean? Can you quantify it?

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -20,60 +29,359 @@ FOUR_C_NAMESPACE_OPEN
...
+
+      //! set tensors for a given fiber direction \f$ \boldsymbol{m} \f$
+      void set_material_const_tensors(const Core::LinAlg::Matrix<3, 1>& m);
+    };
+
+    //! struct with local substepping utilities
+    struct LocalSubsteppingUtils
+    {
+      //! current time parameter ranging from 0 to the problem time step \f$ \Delta t \f$
+      double t_;
```
c-p-schmidt: I like that you structured the different quantities as structs, but I think we do not use `_` at the end for struct members.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -286,39 +288,72 @@ namespace
...
-      inelastic_defgrad_isotrop_elast_viscoplast_data.add("FIBER_READER_ID", 5);
-      inelastic_defgrad_isotrop_elast_viscoplast_data.add("YIELD_COND_A", 1.0);
-      inelastic_defgrad_isotrop_elast_viscoplast_data.add("YIELD_COND_B", 2.0);
-      inelastic_defgrad_isotrop_elast_viscoplast_data.add("YIELD_COND_F", 2.5);
-      inelastic_defgrad_isotrop_elast_viscoplast_data.add("ANISOTROPY", std::string("isotrop"));
-      inelastic_defgrad_isotrop_elast_viscoplast_data.add("LOG_SUBSTEP", true);
-      inelastic_defgrad_isotrop_elast_viscoplast_data.add("MAX_HALVE_NUM_SUBSTEP", 10);
-      params_isotrop_elast_viscoplast_ =
+      Core::IO::InputParameterContainer inelastic_defgrad_isotrop_vplast_refJC_data;
+      inelastic_defgrad_isotrop_vplast_refJC_data.add("FIBER_READER_ID", 5);
```
c-p-schmidt, line 331: You also need a fiber reader for the isotropic case? Or is this just not used then?

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
dragos-ana: This is rather difficult to properly rename. The `default_method` that is used depends on the matrix characteristics. In my opinion, `automatic` describes it slightly better, and I adapted it so. 
Moreover, I added the info regarding automatic method selection depending on matrix characteristics to the parameter description.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
dragos-ana: See my comment above.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: I appended `the determinant of the inelastic deformation gradient deviates from 1 beyond a set tolerance`, to avoid having to always update the exact number when I update the tolerance.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: There are some examples where it's done in the code.
But you are right, in the majority of the cases it is not done. 
I adapted it.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
dragos-ana: It is not used. However, tests fail without it since they require it to be explicitly set in the parameter container.

c-p-schmidt:
Can someone have an additional look at this PR, please?

dragos-ana:
Yes, that would be great. It looks scary due to the number of lines, but it is really just relocation and some rather small changes

rjoussen:
@dragos-ana @c-p-schmidt I can have a look, I assume I'm working with the new version of these files already anyway :)

dragos-ana:
> @dragos-ana @c-p-schmidt I can have a look, I assume I'm working with the new version of these files already anyway :)

Yes, thanks a lot @rjoussen!

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2793,28 +2795,75 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                                                "yield condition: parameter B, "
+                                                "following "
+                                                "the notation in Dafalias 1989, International "
+                                                "Journal of Plasticity, Vol. 5"}),
+            parameter<double>(
+                "YIELD_COND_F", {.description = "transversely isotropic version of the Hill(1948) "
+                                                "yield condition: parameter F, "
+                                                "following "
+                                                "the notation in Dafalias 1989, International "
+                                                "Journal of Plasticity, Vol. 5"}),
```
rjoussen: Not part of this PR, but I think it would be really good if these were optionals and only required if the material behavior is anisotropic.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -1747,12 +1711,13 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::calculate_gamma_delta(
...
 /*--------------------------------------------------------------------*
  *--------------------------------------------------------------------*/
-Mat::InelasticDefgradTransvIsotropElastViscoplast::StateQuantities
+ViscoplastUtils::StateQuantities
 Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
     const Core::LinAlg::Matrix<3, 3>& CM, const Core::LinAlg::Matrix<3, 3>& iFinM,
-    const double plastic_strain, Mat::ViscoplastErrorType& err_status, const double dt)
+    const double plastic_strain, ViscoplastUtils::ErrorType& err_status, const double dt,
+    const ViscoplastUtils::StateQuantityEvalType& eval_type)
 {
```
rjoussen, line 1705: I think this function could be marked `const`.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -1761,192 +1726,239 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
...
  *--------------------------------------------------------------------*/
-Mat::InelasticDefgradTransvIsotropElastViscoplast::StateQuantityDerivatives
+ViscoplastUtils::StateQuantityDerivatives
 Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantity_derivatives(
     const Core::LinAlg::Matrix<3, 3>& CM, const Core::LinAlg::Matrix<3, 3>& iFinM,
-    const double plastic_strain, Mat::ViscoplastErrorType& err_status, const double dt,
-    const bool eval_state)
+    const double plastic_strain, ViscoplastUtils::ErrorType& err_status, const double dt,
+    const ViscoplastUtils::StateQuantityDerivEvalType& eval_type, const bool eval_state)
 {
```
rjoussen, line 1948: Also this could be marked `const` I think.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2560,34 +2640,34 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::setup(const int numgp,
...
-  time_step_quantities_.last_rightCG_.resize(numgp, time_step_quantities_.last_rightCG_[0]);
-  time_step_quantities_.current_rightCG_.resize(
-      numgp, time_step_quantities_.last_rightCG_[0]);  // value irrelevant at this point
+  time_step_quantities_.last_rightCG.resize(numgp, time_step_quantities_.last_rightCG[0]);
+  time_step_quantities_.current_rightCG.resize(
+      numgp, time_step_quantities_.last_rightCG[0]);  // value irrelevant at this point
 
   // default values of the deformation gradient
-  time_step_quantities_.current_defgrad_.resize(numgp, time_step_quantities_.current_defgrad_[0]);
+  time_step_quantities_.current_defgrad.resize(numgp, time_step_quantities_.current_defgrad[0]);
```
rjoussen: Could the `TimeStepQuantities` struct have a little `resize(int numgp)` function which does that for all members? That makes it easier to append stuff to TimestepQuantities since it only needs to be changed in one place.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -20,60 +29,360 @@ FOUR_C_NAMESPACE_OPEN
...
+      FullEval,  ///< full evaluation (full call of the evaluate_state_quantity_derivatives
+                 ///< method)
+      PlasticStrainRateDerivsOnly,  ///< return in evaluate_state_quantity_derivatives once the
+                                    ///< derivatives of the plastic strain rate have been
+                                    ///< evaluated
+      EquivStressDerivsOnly,  ///< return in evaluate_state_quantities once the derivatives of the
+                              ///< equivalent stress has been evaluated
+    };
+
+  }  // namespace InelasticDefgradTransvIsotropElastViscoplastUtils
```
rjoussen: I understand the idea of these, but I am a little worried that if we end up using them, that means that there are `StateQuantities` and `StateQuantitityDerivatives` going around which are incomplete. If somewhere else you want to access them, it could be a huge mess to debug what's going wrong. Maybe both struct should store their `EvalType` at least?

tests/input_files/mat_iso_viscoplast_refJC_log_substep.4C.yaml
```diff
@@ -52,9 +52,9 @@ MATERIALS:
       YIELD_COND_A: 0
       YIELD_COND_B: 0
       YIELD_COND_F: 0
-      ANISOTROPY: "isotrop"
-      LOG_SUBSTEP: true
-      MAX_HALVE_NUM_SUBSTEP: 10
+      MAT_BEHAVIOR: "isotrop"
```
rjoussen: I'm just noticing this here: I think we should change `"isotrop"` to `"isotropic"`. Isotrop is German.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
dragos-ana: Good idea! I have implemented this now and removed the parameters from the isotropic tests ;)

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Great idea! 
Based on your comment, I decided to go one step even further and provide `init`, `resize`, `update`, `pack` and `unpack` functions directly to the `TimeStepQuantities` struct. 
The PR has exploded anyway in terms of changes, might as well do this further restructuring, which in my opinion makes the code structure slightly more comprehensible.

tests/input_files/mat_iso_viscoplast_refJC_log_substep.4C.yaml
dragos-ana: I tried to shorten the names, that's why "transv_isotrop" and "isotrop" :) 
But I agree, it becomes clearer when using "transv_isotropic" and "isotropic", which I have done now.

dragos-ana:
Thanks for the detailed reviews, @rjoussen  and @c-p-schmidt!
I implemented the further suggestions from @rjoussen into a further commit to keep it logically separated from the first commit; however, I can still squash if that's better in regards to our PR policies.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -281,9 +333,28 @@ namespace Mat
...
+    enum class StateQuantityDerivEvalType
+    {
+      FullEval,  ///< full evaluation (full call of the evaluate_state_quantity_derivatives
+                 ///< method)
+      PlasticStrainRateDerivsOnly,  ///< return in evaluate_state_quantity_derivatives once the
+                                    ///< derivatives of the plastic strain rate have been
+                                    ///< evaluated
+      EquivStressDerivsOnly,  ///< return in evaluate_state_quantities once the derivatives of the
+                              ///< equivalent stress has been evaluated
     };
```
rjoussen, line 354: Do they need to be different from `StateQuantityEvalType`? I could imagine that this adds a lot of if statements later when we use them. But I assume that depends on what the exact plan is to use them, so its up to you.

rjoussen (approved):
Looks really nice now! I added one small last comment.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: I would keep them different for now, but it's a good point. I guess it will depend on the state of the code after I merge my remaining changes, so I will keep this in mind.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: Amazing, thank you! It does look a lot cleaner now with all the timestep quantity management sourced out.

PR #1849, opened 2026-03-11
[discretization] Simplify element volume calculation

maxfirmbach:
## Description and Context

As the title states.

In addition this PR currently adds:
- length calculation for 1D Nurbs elements
- area calculation for 2D Nurbs elements
- volume calculation for 3D Nurbs elements
- Updates documentation of all functions involved

### DIFF
--- a/src/core/fem/src/geometry/4C_fem_geometry_element_volume.hpp
+++ b/src/core/fem/src/geometry/4C_fem_geometry_element_volume.hpp
@@ -292,7 +292,7 @@ namespace Core::Geo
         return element_volume<Core::FE::CellType::nurbs27>(xyze, knots, weights);
       default:
         FOUR_C_THROW(
-            "Element volume calculation is current not implemented for the given cell type.");
+            "Element volume calculation is currently not implemented for the given cell type.");
     }
   }
 


src/core/fem/src/geometry/4C_fem_geometry_element_volume.hpp
```diff
@@ -139,98 +131,52 @@ namespace Core::Geo
...
 
-    d.update(1.0, x1, -1.0);
-    return d.norm2();
+    return vol;
   }
 
 
-  /** \brief calculates the area of a surface element in given configuration
-   */
+  //! calculates the volume of an element in given configuration
```
c-p-schmidt: I think it would be good to add more documentation on the behavior of this method: for 1D elements, the volume is identified with the length; for 2D elements, with the surface area; and for 3D elements, with the actual volume of the element.

If I do not read anything in the docu, I would assume that the volume of a 1D element is calculated by its length and an assigned cross-sectional area, as for example with trusses or beams.

maxfirmbach: I did update the documentation.

maxfirmbach:
Ready for review.

PR #1850, opened 2026-03-11
Introduce specific solver for the calculation of the initial potential field

c-p-schmidt:
## Description and Context
While working on simulations for the HPSF community summit, it turned out to be valuable to allow different preconditioner settings for the calculation of the initial potential field compared to all other solves later.

Thus, this PR enables explicitly setting a different solver for this. If no such solver is set, just the solver for the main part is reused. And adapts an input file that does this.

### DIFF
--- a/src/ssi/4C_ssi_input.cpp
+++ b/src/ssi/4C_ssi_input.cpp
@@ -225,7 +225,11 @@ std::vector<Core::IO::InputSpec> SSI::valid_parameters()
 
           parameter<bool>("INITPOTCALC",
               {.description = "Automatically calculate initial field for electric potential",
-                  .default_value = false})},
+                  .default_value = false}),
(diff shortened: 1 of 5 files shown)


src/ssi/4C_ssi_input.cpp
```diff
@@ -225,7 +225,12 @@ std::vector<Core::IO::InputSpec> SSI::valid_parameters()
 
           parameter<bool>("INITPOTCALC",
               {.description = "Automatically calculate initial field for electric potential",
-                  .default_value = false})},
+                  .default_value = false}),
+          parameter<int>("INIT_POT_CALC_LINEAR_SOLVER",
+              {.description = "ID of linear solver for global system of equations for calculation "
+                              "of the initial potential field",
+                  .default_value = -1}),
```
davidrudlstorfer: If the default value is -1 it would probably make more sense to use an optional int?

maxfirmbach (approved):

PR #1851, opened 2026-03-11
Fix DEM minimum radius check

ppraegla:
The error check for the minimum radius failed to trigger when MIN_RADIUS is not set (default value 0.0), and the radius of a particle is 0.0 (e.g., missing RAD property in particle definition). MIN_RADIUS is made optional because it is only needed for a radius distribution. For the other forms of radius input, it is only checked if the radius is greater than zero.

If the particle input is not from a distribution, MIN_RADIUS is just ignored. This is not ideal, but the other solutions I considered required more restructuring of the input.

### DIFF
--- a/src/particle/src/4C_particle_input.cpp
+++ b/src/particle/src/4C_particle_input.cpp
@@ -9,6 +9,7 @@
 
 #include "4C_fem_condition_definition.hpp"
 #include "4C_io_input_spec_builders.hpp"
+#include "4C_io_input_spec_validators.hpp"
 FOUR_C_NAMESPACE_OPEN
 
 /*---------------------------------------------------------------------------*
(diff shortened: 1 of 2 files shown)


src/particle/src/interaction/4C_particle_interaction_dem.cpp
```diff
@@ -291,11 +289,11 @@ void Particle::ParticleInteractionDEM::set_initial_radius()
...
         // safety checks
-        if (container->get_min_value_of_state(Particle::Radius) < r_min)
-          FOUR_C_THROW("minimum particle radius smaller than minimum allowed particle radius!");
+        FOUR_C_ASSERT_ALWAYS(container->get_min_value_of_state(Particle::Radius) > 0,
+            "the minimum particle radius is smaller than zero. Fix the particle input.");
 
-        if (container->get_max_value_of_state(Particle::Radius) > r_max)
-          FOUR_C_THROW("maximum particle radius larger than maximum allowed particle radius!");
+        FOUR_C_ASSERT_ALWAYS(container->get_max_value_of_state(Particle::Radius) <= r_max,
+            "the maximum particle radius larger than maximum allowed particle radius!");
```
slfuchs:
```suggestion
            "the maximum particle radius is larger than the maximum allowed particle radius!");
```

slfuchs (approved):
Looks good from my side!

src/particle/src/4C_particle_input.cpp
```diff
@@ -484,10 +485,14 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
               {.description = "type of (random) surface energy distribution",
                   .default_value = SurfaceEnergyDistribution::Constant}),
 
-          parameter<double>("MIN_RADIUS",
-              {.description = "minimum allowed particle radius", .default_value = 0.0}),
+          parameter<std::optional<double>>("MIN_RADIUS",
+              {.description =
+                      "minimum allowed particle radius: used as cutoff for a radius distribution",
```
georghammerl:
```suggestion
                      "minimum allowed particle radius: used as lower cutoff for a radius distribution",
```

src/particle/src/4C_particle_input.cpp
```diff
@@ -484,10 +485,14 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
...
+                  .validator = null_or(positive<double>())}),
           parameter<double>("MAX_RADIUS",
-              {.description = "maximum allowed particle radius", .default_value = 0.0}),
+              {.description = "maximum allowed particle radius: used as cutoff for a radius "
```
georghammerl:
```suggestion
              {.description = "maximum allowed particle radius: used as upper cutoff for a radius "
```

PR #1817, opened 2026-03-06
[SPH] Store both Dirichlet and Neumann open boundries in a single vector

vovannikov:
## Description and Context
I have started gradually working on a feature I proposed in #1816. Instead of having 2 pointers referring to 1 Neumann and 1 Dirichlet open boundary each, all the open boundaries can be stored in a vector and their number is not limited to 2. With the current design it is actually very easy to generalize handling of the open boundaries.

## Related Issues and Pull Requests
#1816

Issue #1816, opened 2026-03-05
Introduce multiple open boundaries for SPH


vovannikov:
## Description
Currently, one can have only one inlet and one outlet in SPH benchmarks. This does not seem convenient. Recently, a colleague of mine requested multiple inlets for his simulations.

## Possible Solution and Definition of Done
My suggestion is to modify the input file section as follows.

Old:
```yaml
PARTICLE DYNAMIC/SPH:
...
  DIRICHLETBOUNDARYTYPE: "DirichletNormalToPlane"
  DIRICHLET_FUNCT: 1
  DIRICHLET_OUTWARD_NORMAL: "-1.0 0.0 0.0"
  DIRICHLET_PLANE_POINT: "-0.1 0.0 0.0"
  NEUMANNBOUNDARYTYPE: "NeumannNormalToPlane"
  NEUMANN_OUTWARD_NORMAL: "1.0 0.0 0.0"
  NEUMANN_PLANE_POINT: "0.1 0.0 0.0"
```

New (here I have already 2 Dirichlet boundaries as an example):
```yaml
PARTICLE DYNAMIC/INITIAL AND BOUNDARY CONDITIONS:
...
  OPEN_BOUNDARIES:
    - BOUNDARY: 0
      DIRICHLET:
        TYPE: "DirichletNormalToPlane"
        FUNCT: 1
        OUTWARD_NORMAL: "-1.0 0.0 0.0"
        PLANE_POINT: "-0.1 0.0 0.0"
    - BOUNDARY: 1
      DIRICHLET:
        TYPE: "DirichletNormalToPlane"
        FUNCT: 1
        OUTWARD_NORMAL: "-1.0 0.0 0.0"
        PLANE_POINT: "-0.1 0.0 0.0"
    - BOUNDARY: 2
      NEUMANN:
        TYPE: "NeumannNormalToPlane"
        OUTWARD_NORMAL: "1.0 0.0 0.0"
        PLANE_POINT: "0.1 0.0 0.0"
```

Then I would eliminate `dirichletphase` and `neumannphase` and replace them with `openphase`. Consequently, I would remove `dirichletopenboundary_` and `neumannopenboundary_` members from `Particle::ParticleInteractionSPH` and create a vector of `std::unique_ptr<Particle::SPHOpenBoundaryBase>`. Subsequently, a particle data would get transformed from

```yaml
- "TYPE dirichletphase POS -0.105 -0.045 0.0"
```

into
```yaml
[...]

PR #1856, opened 2026-03-13
[SPH] Multiple open (Dirichlet and Neumann) boundaries

vovannikov:
## Description and Context
As proposed in #1816, this PR allows one to define multiple Dirichlet and Neumann boundaries. The boundaries are defined in the input file as follows:
```yaml
PARTICLE DYNAMIC/OPEN BOUNDARIES:
  DIRICHLET_BOUNDARIES:
    - BOUNDARY: 0
      DirichletNormalToPlane:
        FUNCT: 1
        OUTWARD_NORMAL: "-1.0 0.0 0.0"
        PLANE_POINT: "-0.1 0.0 0.0"
    - BOUNDARY: 1
      DirichletNormalToPlane:
        FUNCT: 1
        OUTWARD_NORMAL: "-1.0 0.0 0.0"
        PLANE_POINT: "-0.1 0.0 0.0"
  NEUMANN_BOUNDARIES:
    - BOUNDARY: 0
      NeumannNormalToPlane:
        OUTWARD_NORMAL: "1.0 0.0 0.0"
        PLANE_POINT: "0.1 0.0 0.0"
```
The old way to define the open boundaries is also preserved, i.e. this is equivalent:
```yaml
PARTICLE DYNAMIC/SPH:
  DIRICHLETBOUNDARYTYPE: "DirichletNormalToPlane"
  DIRICHLET_FUNCT: 1
  DIRICHLET_OUTWARD_NORMAL: "-1.0 0.0 0.0"
  DIRICHLET_PLANE_POINT: "-0.1 0.0 0.0"
  NEUMANNBOUNDARYTYPE: "NeumannNormalToPlane"
  NEUMANN_OUTWARD_NORMAL: "1.0 0.0 0.0"
  NEUMANN_PLANE_POINT: "0.1 0.0 0.0"
```
I could have removed it, but decided to keep not to break the user experience.

There is one nuance in the implementation which I am not absolutely happy about.

For the case when there is a single boundary of the same type, nothing changes, but for _N_ boundaries of the same type, the loop over the particles in a container runs _N_ times each time picking only the particles belonging to the different boundaries according to the provided `boundary_id`, something like this:
```c++
  for (int particle_i = 0; particle_i < container_i->particles_stored(); ++particle_i)
  {
    // get particle boundary id
    const double* boundary_id_i =
        container_i->get_ptr_to_state(Particle::OpenBoundaryId, particle_i);
    if (static_cast<int>(*boundary_id_i) != boundary_id_) continue;
    ...
  }
```
Given that the number of boundary particles is smaller than the number of fluid particles and that we usually have single inlet and/or outlie, this overhead is not significant. However, I got an idea how this can be improved. To this end, we could alter `states_` property here:
https://github.com/4C-multiphysics/4C/blob/e6e2d7e36ad6a91ec4a1be1c364befaf1cbdcbf9/src/particle/src/engine/4C_particle_engine_container.hpp#L413-L414
to make it `std::vector<std::vector<std::vector<double>>>`. This would also allow us to actually later add more fluid phases. The rigid body or PD body states can be also stacked separately depending on the `RIGIDCOLOR` or `PDBODY` property.

#1816, #1817


src/particle/src/engine/4C_particle_engine_enums.hpp
```diff
@@ -96,7 +96,10 @@ namespace Particle
     Young,                  //!< young's modulus
     InitialConnectedBonds,  //!< initial active bonds
     CurrentConnectedBonds,  //!< remained active bonds
-    PDDamageVariable        //!< damage level at pd node
+    PDDamageVariable,       //!< damage level at pd node
+    /*---------------------------------------------------------------------------*/
+    // open boundary phases
```
slfuchs:
```suggestion
    // particle states for open boundaries
```

src/particle/src/interaction/4C_particle_interaction_sph_open_boundary.cpp
```diff
@@ -27,11 +27,12 @@ FOUR_C_NAMESPACE_OPEN
 /*---------------------------------------------------------------------------*
  | definitions                                                               |
  *---------------------------------------------------------------------------*/
-Particle::SPHOpenBoundaryBase::SPHOpenBoundaryBase(const Teuchos::ParameterList& params)
-    : params_sph_(params),
```
slfuchs, line 31: Why not leave it as it is and just hand over the parameter list? This is how it is done for most classes in the particle framework ...

vovannikov: In my opinion, this is the thing that probably should be improved in the framework. First of all, I tend to write the code as explicit as possible. If a class needs some parameters - why not to provide these parameters directly? This improves readability and simplifies maintenance.

Of course, sometimes you have too many parameters, then it makes sense to provide the parameters list. In my opinion, one needs to be reasonable about it. Here, for instance, a single parameter is actually required from the whole list. Moreover, this is a base class, and a user is somewhat not supposed to instantiate it, providing again parameters list to descendants.

Another thing, which I normally try to avoid, is storing the parameter lists or references to them as internal members. This again creates a tight implicit dependency which makes it more difficult to reason about the code. What parameters are going to be used? This aspect remains unclear just looking at the ctor. I then need to look through the whole class to figure it out. For this reason I think it is better to deal with the parameters list only in the ctor - pull all the necessary parameters, initialize everything that is needed and then not to reuse the parameters list at all.

vovannikov:
@slfuchs, thank you for your review!

> I would prefer to remove the old way to define the open boundaries. Let's be realistic, at this state it is way better for the code to just have one way to define open boundaries and not try to maintain backward compatibility.
> 

I fully agree here. If no one objects this decision, I will happily remove the old definitions.

> For me it is absolutely fine to iterate over the particles in the container and check for the open boundary ID. As you said, there are just few open boundary particles compared to fluid particles, and the open boundary particles are in its own container anyway. So I am sure you won't even notice a speed up when trying to change that. Besides that, I am against changing to much in the particle container and its states.

I agree that this is not critical, I just wanted to mention this aspect in order not to forget about it in the future, if we encounter a strong need to have more than 2 liquid phases or something similar. Just keep for the history that all these demands could be handled by, at least, this modification, but probably other options could be brought to the table. At the moment, I do not see a lot of sense in making this modification.

georghammerl:
I would also prefer to have just one version of providing Dirichlet and Neumann BCs. Hence, deletion of the previous way is fine for me. (However, I do not have that much different input files in use).

src/particle/src/interaction/4C_particle_interaction_sph.cpp
```diff
@@ -45,12 +45,12 @@ Particle::ParticleInteractionSPH::ParticleInteractionSPH(
     MPI_Comm comm, const Teuchos::ParameterList& params)
     : Particle::ParticleInteractionBase(comm, params), params_sph_(params.sublist("SPH"))
 {
-  initialize_members();
+  initialize_members(params);
```
ppraegla: It's not necessary to pass `params` because you can also access `params_` from the base class

tests/input_files/particle_sph_2d_openboundary_straight_channel_two_dirichlet_boundaries.4C.yaml
```diff
@@ -0,0 +1,649 @@
...
+  - "TYPE dirichletphase POS -0.105 0.025 0.0 BOUNDARYID 1"
+  - "TYPE dirichletphase POS -0.115 0.025 0.0 BOUNDARYID 1"
+  - "TYPE dirichletphase POS -0.125 0.025 0.0 BOUNDARYID 1"
+  - "TYPE dirichletphase POS -0.105 0.035 0.0 BOUNDARYID 1"
+  - "TYPE dirichletphase POS -0.115 0.035 0.0 BOUNDARYID 1"
+  - "TYPE dirichletphase POS -0.125 0.035 0.0 BOUNDARYID 1"
+  - "TYPE dirichletphase POS -0.105 0.045 0.0 BOUNDARYID 1"
+  - "TYPE dirichletphase POS -0.115 0.045 0.0 BOUNDARYID 1"
+  - "TYPE dirichletphase POS -0.125 0.045 0.0 BOUNDARYID 1"
+  - "TYPE neumannphase POS 0.105 -0.045 0.0"
```
ppraegla: Is `BOUNDARYID 0` missing here

src/particle/src/interaction/4C_particle_interaction_sph_open_boundary.cpp
ppraegla: I'm strongly in favor of removing the ParameterList where possible. As @vovannikov already said, it makes it hard to determine which parameters the class actually needs. 
If many parameters are needed, it makes sense to bundle them in a struct. The struct can also be directly filled from the InputSpec.

src/particle/src/4C_particle_input.cpp
```diff
@@ -126,6 +126,53 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
...
+      parameter<int>("BOUNDARY"),
+      one_of(
+          std::vector<Core::IO::InputSpec>{
+              group("NoDirichletOpenBoundary", {}, {.description = "no dirichlet open boundary"}),
+              group("DirichletNormalToPlane",
+                  {
+                      parameter<int>(
+                          "FUNCT", {.description = "number of function governing velocity "
+                                                   "condition on dirichlet open boundary",
+                                       .default_value = -1}),
```
ppraegla: Does it make sense to remove the default here when there is no reasonable default? Also for OUTWARD_NORMAL, PLANE_POINT, and the Neumann boundary.

```suggestion
                          "FUNCT", {.description = "number of function governing velocity "
                                                   "condition on dirichlet open boundary"}),
```

src/particle/src/4C_particle_input.cpp
```diff
@@ -126,6 +126,53 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
...
+                      parameter<std::string>("OUTWARD_NORMAL",
+                          {.description = "direction of outward normal on dirichlet open boundary",
+                              .default_value = "0.0 0.0 0.0"}),
```
ppraegla: Remove the default here as well?

And, what do you think about explicitly reading a vector instead of a string?

```suggestion
                      parameter<std::vector<double>>("OUTWARD_NORMAL",
                          {.description = "direction of outward normal on dirichlet open boundary",
                              .size = 3}),
```

This would simplify the reading to 
```c++
outwardnormal_ = params.get<std::vector<double>>("PLANE_POINT");
```

The input would look like
```yaml
- BOUNDARY: 2
      DirichletNormalToPlane:
        FUNCT: 3
        OUTWARD_NORMAL: [0.0, 0.0, 0.0] # flow style
        PLANE_POINT:
          - 0.0
          - 0.0
          - 0.0 # block style
```
Yaml sequences can be specified in flow or block style. For me, VScode uses block style when auto-completing, which I personally find less readable for a vector with 3 entries. So, I'm unsure about this change. However, the vector input would be more consistent with what we expect from the input.

georghammerl:
> Thanks for this improvement, @vovannikov . I have some suggestions.
> 
> From my side, you can remove the old open boundary input. I'm not using it and there should not be two ways to specify the open boundary. @georghammerl Do you know of anyone else who could be affected by this?

@mairehenke Is this relevant for you?

georghammerl:
What happens if the referenced boundary id is not availble, i.e. an input file containing `TYPE dirichletphase POS -0.105 0.045 0.0 BOUNDARYID 1` does not contain the corresponding `boundary id 1` ? Is there already an error message included?

mairehenke:
It’s relevant for me, but I’m fine with the new version. I’d also prefer having just one way to specify the open boundary instead of keeping two versions. The input file changes are pretty minor, so adapting should be easy.

src/particle/src/interaction/4C_particle_interaction_sph_open_boundary.cpp
slfuchs: Sounds reasonable to me, so let's do it as you proposed.

slfuchs:
> I agree that this is not critical, I just wanted to mention this aspect in order not to forget about it in the future, if we encounter a strong need to have more than 2 liquid phases or something similar. Just keep for the history that all these demands could be handled by, at least, this modification, but probably other options could be brought to the table. At the moment, I do not see a lot of sense in making this modification.

If we need to implement more fluid phases I would tend to still leave the particle container untouched and rather modify the particle container bundle. Currently the particle phase enum is used to index the corresponding particle container in the particle container bundle. So one could for example come up with a solution, that defines the index in another way.

tests/input_files/particle_sph_2d_openboundary_straight_channel_two_dirichlet_boundaries.4C.yaml
vovannikov: Strictly speaking yes, but this is not necessary, since the default value 0 is used if `BOUNDARYID` is not provided, as similarly done for `RIGIDCOLOR` parameter for `rigidphase`. But I think it is worth adding it here for consistency with the `dirichletphase`.

src/particle/src/4C_particle_input.cpp
vovannikov: Indeed, it makes sense to remove the default value for the Dirichlet case, whereas we should keep it optional for the Neumann, since in the latter case it just means a zero pressure condition, which is actually a commonly used one.

src/particle/src/4C_particle_input.cpp
vovannikov: I like this idea, thank you for the hint.

vovannikov:
> What happens if the referenced boundary id is not availble, i.e. an input file containing `TYPE dirichletphase POS -0.105 0.045 0.0 BOUNDARYID 1` does not contain the corresponding `boundary id 1` ? Is there already an error message included?

@georghammerl No, there was no error emitted in this case and such particles will simply be skipped in the relevant open boundary particle loops. I have added it in the recent update of this PR.

src/particle/src/interaction/4C_particle_interaction_sph.cpp
```diff
@@ -168,6 +165,57 @@ void Particle::ParticleInteractionSPH::setup(
...
+  std::vector<int> available_boundary_ids(openboundaries_.size());
+  std::transform(openboundaries_.begin(), openboundaries_.end(), available_boundary_ids.begin(),
+      [](const auto& boundary) { return boundary->get_boundary_id(); });
+
+  for (const auto& type_id : particlecontainerbundle_->get_particle_types())
+  {
+    // skip if this is not an open boundary type
+    if (types_to_check.find(type_id) == types_to_check.end()) continue;
+
+    // vector of missing bouindary ids
```
georghammerl:
```suggestion
    // vector of missing boundary ids
```

georghammerl (dismissed):

tests/input_files/particle_sph_2d_openboundary_straight_channel_two_dirichlet_boundaries.4C.yaml
```diff
@@ -28,18 +28,18 @@ PARTICLE DYNAMIC/OPEN BOUNDARIES:
     - BOUNDARY: 0
       DirichletNormalToPlane:
         FUNCT: 1
-        OUTWARD_NORMAL: "-1.0 0.0 0.0"
-        PLANE_POINT: "-0.1 0.0 0.0"
+        OUTWARD_NORMAL: [-1.0, 0.0, 0.0]
+        PLANE_POINT: [-0.1, 0.0, 0.0]
```
ppraegla, line 32: Looks much nicer now. Thanks 👍

src/particle/src/interaction/4C_particle_interaction_sph_open_boundary.cpp
```diff
@@ -193,43 +196,30 @@ void Particle::SPHOpenBoundaryBase::check_open_boundary_phase_change(
...
+
+Particle::SPHOpenBoundaryDirichlet::SPHOpenBoundaryDirichlet(
+    const Teuchos::ParameterList& params, double initialparticlespacing, int boundary_id)
+    : SPHOpenBoundaryBase::SPHOpenBoundaryBase(initialparticlespacing, boundary_id)
 {
-  prescribedstatefunctid_ = params_sph_.get<int>("DIRICHLET_FUNCT");
+  prescribedstatefunctid_ = params.get<int>("FUNCT");
 
-  if (not(prescribedstatefunctid_ > 0)) FOUR_C_THROW("no function id of prescribed state set!");
+  FOUR_C_ASSERT(prescribedstatefunctid_ > 0, "no function id of prescribed dirichlet state set!");
```
ppraegla: Nice that you adapted this. But you need 

```suggestion
  FOUR_C_ASSERT_ALWAYS(prescribedstatefunctid_ > 0, "no function id of prescribed dirichlet state set!");
```
to keep the same behavior as before. FOUR_C_ASSERT is not active in a release build.

Same for the other assertions

src/particle/src/4C_particle_input.cpp
```diff
@@ -126,6 +126,50 @@ std::vector<Core::IO::InputSpec> Particle::valid_parameters()
...
               {.description = "consider peridynamic body interaction", .default_value = false})},
       {.required = false}));
+
+  auto all_dirichlet_boundaries = all_of({
+      parameter<int>("BOUNDARY"),
+      one_of(
+          std::vector<Core::IO::InputSpec>{
+              group("NoDirichletOpenBoundary", {}, {.description = "no dirichlet open boundary"}),
+              group("DirichletNormalToPlane",
+                  std::vector<Core::IO::InputSpec>{
```
ppraegla: I would remove `std::vector<Core::IO::InputSpec>` to make the spec slightly more readable.
```suggestion
                  {
```

src/particle/src/interaction/4C_particle_interaction_sph_open_boundary.cpp
vovannikov: Sure! Missed it, done.

georghammerl (approved):

slfuchs:
Thanks nice work 👍

davidrudlstorfer:
@vovannikov currently the FourCIPP pipeline is broken (see https://github.com/4C-multiphysics/fourcipp/actions/runs/23281822587/job/67696780242)

I think that this PR is the culprit
```

line_list = ['0']
keyword_casting = {'PDBODYID': functools.partial(<function _extract_entry at 0x7fda24df7880>, extractor=<class 'float'>), 'POS': functoo...float'>), 'RIGIDCOLOR': functools.partial(<function _extract_entry at 0x7fda24df7880>, extractor=<class 'float'>), ...}

    def inline_dat_read(line_list: list, keyword_casting: LineCastingDict) -> dict:
        """Read inline dat to dict.
    
        Note: This function is not able to read nested containers such as groups. This would diminish
        performance and is not needed for the legacy sections.
    
        Args:
            line_list: List to extract the entries
            keyword_casting: Dict with the casting
    
        Returns:
            Entry as dict
        """
        entry: dict = {}
        while line_list:
            key = line_list.pop(0)
            # Raises Error if an entry was provided twice
            if key in entry:
                raise KeyError(f"The entry {key} was provided already: {entry}")
>           entry[key] = keyword_casting[key](line_list)
                         ^^^^^^^^^^^^^^^^^^^^
E           KeyError: 'BOUNDARYID'

src/fourcipp/legacy_io/inline_dat.py:222: KeyError

```

The corresponding failing test is: `particle_sph_2d_openboundary_straight_channel_two_dirichlet_boundaries.4C.yaml`


Maybe you could check out what is going wrong:)

ppraegla:
@vovannikov I found what's missing. I can push a quick fix

PR #1857, opened 2026-03-13
Centralize Global::Problem in remaining porofluid_pressure_based

bgoderbauer:
## Description and Context

Continues isolating `Global::Problem()` calls in the three remaining porofluid_pressure_based folders.
- porofluid_pressure_based_ele:
  - Removed `Global::Problem` access from element internals.
  - Passed spatial dimension explicitly into the element factory.
  - Threaded FunctionManager via element parameter context for Neumann/domain-integral function lookup.
- porofluid_pressure_based_elast:
  - Added `PorofluidElastAlgorithmDeps`.
  - Centralized `Global::Problem` access in composition (dyn) and injected callbacks/data into base/monolithic/partitioned/utils paths.
  - Replaced direct singleton usage in clone/material/solver/result-test/restart-related code with injected deps/callbacks.
- porofluid_pressure_based_elast_scatra:
  - Added `PorofluidElastScatraAlgorithmDeps` and artery-coupling dependency structs/builders.
  - Injected dependencies into base/monolithic/utils and artery-coupling internals (including pair init/function lookup/rank/dimension handling).
  - Kept compatibility wrapper overloads where external artery strategy callsites still rely on the legacy signature.

Follows #1842

### DIFF
--- a/src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.hpp
+++ b/src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.hpp
@@ -21,6 +21,7 @@
 
 #include <Sacado.hpp>
 
+#include <functional>
 #include <memory>
 
 // define Fad object for evaluation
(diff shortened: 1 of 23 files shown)


src/porofluid_pressure_based_elast_scatra/4C_porofluid_pressure_based_elast_scatra_artery_coupling_pair.hpp
```diff
@@ -89,7 +90,10 @@ namespace PoroPressureBased
         const std::vector<int>& coupled_dofs_artery,
         const std::vector<std::vector<int>>& scale_vector,
         const std::vector<std::vector<int>>& funct_vector, std::string condition_names,
-        double penalty_parameter, std::string coupling_type = "", int eta_ntp = 0) = 0;
+        double penalty_parameter, std::string coupling_type = "", int eta_ntp = 0,
```
c-p-schmidt, line 93: You could remove the defaults, at least I saw in one case that you already explicitly specified them anyway. However, there might be other occurrences left.

```suggestion
        double penalty_parameter, std::string coupling_type, int eta_ntp,
```

bgoderbauer: There are a few more places throughout the module. I think it doesn't really matter whether this is cleaned up like this or not :smile:, I would leave it to the people working deeper in the code.

PR #1865, opened 2026-03-17
Check for early stoppage in struct field

tuchpaul:
In growth and remodeling simulations, a prestressing algorithm is often used to obtain a desired configuration. Therefore, an additional convergence criterion is introduced that checks whether the displacement falls below a prescribed tolerance.

To stop the time loop for this additional criterion, an `early_stopping()` mechanism is implemented (see `bool Solid::IMPLICIT::PreStress::early_stopping() const` in `4C_structure_new_impl_prestress.cpp`).

This PR adds the additional convergence check for 2-way coupling (2wc), 1-way coupling (1wc), and monolithic coupling via `structure_field()->not_finished()`. The test `ssi_homogenized_constraint_mixture_prestress_qcyl.4C.yaml` checks the functionality for the 2wc case.

Since there are additional early-stopping checks in electrochemistry simulations handled by `scatra_field()->not_finished()`, I suggest always checking the SSI Control via `not_finished()` as well as the corresponding fields `scatra_field()->not_finished()` and `structure_field()->not_finished()` before the respective step.

NOTE: For `ssi_OneWay_SolidToScatra` tests, `scatra_field()->not_finished()` is called in the `prepare_time_step()` method.

---

While implementing, I went through some test files and noticed the following:

If different time step sizes are used, the corresponding time loop checks for the flag `DIFFTIMESTEPSIZE: true`. If enabled, the SSI base initializes the time integrators with the time parameters of the respective field. If disabled, both fields receive the global time parameters from SSI Control.

In some tests, `DIFFTIMESTEPSIZE: true` was set despite both fields sharing the same step size. I fixed this and removed the redundant double initialization.


src/ssi/4C_ssi_monolithic.cpp
```diff
@@ -909,7 +909,7 @@ void SSI::SsiMono::timeloop()
   if (step() == 0) prepare_time_loop();
 
   // time loop
-  while (not_finished() and scatra_field()->not_finished())
+  while (not_finished() and scatra_field()->not_finished() and structure_field()->not_finished())
```
amgebauer, line 912: I have a rather general question to the state before your change. Why do we ask here the coupling algorithm and the scatra field whether the simulation is finished? Why does the scatra field needs to be asked in a monolithic scheme?

I see why you need to ask also the structure field now (which makes this consistent actually), however, I'm not sure whether this is the right way to do it.

tuchpaul: As mentioned in the comment above, there are also checks used to determine whether an additional convergence criterion is met for the scatra field. Similar to the prestress check in the structure field, certain electrochemistry simulations rely on an additional stopping criterion on the scatra integration side:

```
//! true, when all half cycles are completed
bool not_finished() const { return nhalfcycles_ >= ihalfcycle_; };
```

For checks that are tied to a specific field and depend on members within the corresponding time integration, I think this split makes sense. 

But I am happy to discuss alternative approaches.

amgebauer: Ah sorry for not reading the full description, now it makes sense.

I think instead of this inheritance structure of the `not_finished()` function, we could add a `std::function` that we store at the algorithm level which determines whether the simulation is finished or not. We can then set this function to behave as we want it, in the general case via timestep size and max time/steps, or anything else for special algorithms (like iterative prestressing or elch). But we would need to do a cleanup before. I'm fine with doing this in a followup.

PR #1869, opened 2026-03-18
Update superlu_dist dependency to version 9.2.1

c-p-schmidt:
## Description and Context
We have constantly failing nightly tests that use Superlu. Thus, @amgebauer, @ischeider, and I looked into that issue a bit using `valgrind` and found that, with the current Superlu_dist version (9.2.1), not all issues are solved, but some are. Consequently, updating the Superlu_dist version sounds like a good idea.

I've attached the outputs of two test cases: one with the old superlu_dist version and one with the new one.
[homogenized-valgrind-output-old-superlu-dist.txt](https://github.com/user-attachments/files/26082707/homogenized-valgrind-output-old-superlu-dist.txt)
[homogenized-valgrind-output-new-superlu-dist.txt](https://github.com/user-attachments/files/26082708/homogenized-valgrind-output-new-superlu-dist.txt)
[valgrind-output-new-superlu-dist.txt](https://github.com/user-attachments/files/26082709/valgrind-output-new-superlu-dist.txt)
[valgrind-output-old-superlu-dist.txt](https://github.com/user-attachments/files/26082710/valgrind-output-old-superlu-dist.txt)

FYI: I'll update the documentation accordingly and push after the manually triggered Docker build was successful.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -16,7 +16,7 @@ jobs:
   gcc13_assertions_build:
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:718a74e5
+      image: ghcr.io/4c-multiphysics/4c-dependencies-ubuntu24.04:24b228a5
       options: --user root --env OMPI_ALLOW_RUN_AS_ROOT=1 --env OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
     defaults:
       run:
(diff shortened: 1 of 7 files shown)


mayrmt:
We picked version 7.2.0, since it was the only version working with Amesos and Amesos2. @maxfirmbach, is there further need to stick to this version? I am not aware, that we still need to maintain compatibility with Amesos, though a confirmation by @maxfirmbach  would be great.

maxfirmbach:
I think we can try updating. Currently I don't see issues, not sure though.

c-p-schmidt:
In the previous push, some info on ArborX was accidentally removed; it's back now, sorry.

c-p-schmidt:
@4C-multiphysics/developers: As soon as this PR is merged, superlu_dist version 9.2.1 is tested in the pipeline and is also the recommended version for use, as it showed fewer issues than the old version 7.2.0 (see valgrind outputs above). To use the same version on your workstation or cluster, you (or an admin at your institute) need to do the following:

1. Build the correct version of superlu_dist. An install script can be found in `dependencies/current/superlu_dist/install.sh`
2. Compile Trilinos, but change the path of superlu_dist to the path of the updated version. In the trilinos install script, these lines look like:
```bash
  -D TPL_ENABLE_SuperLUDist:BOOL=ON \
    -D SuperLUDist_INCLUDE_DIRS:PATH="$INSTALL_DIR/include" \
    -D SuperLUDist_LIBRARY_DIRS:PATH="$INSTALL_DIR/lib" \
```
3. Compile 4C with the newly compiled Trilinos from step 2

PR #1870, opened 2026-03-18
Make control file writing crash safe

georghammerl:
As reported in https://github.com/4C-multiphysics/4C/issues/1476, the control file is not written properly in case a floating point exception occurs. Catching the FPE and trying to flush the control file buffer before shutdown is illegal as described in the issue.

A possible remedy is to flush the control file content regularly with each writing not to loose the latest part in case of an FPE. Not to deteriorate performance, asynchronous writing is chosen (not 100% safe, however first test were all successful. Thanks to @vovannikov for this good hint.) Performance loss (with way too many restart writings and a small test case) was < 1%.

The `sync_file_range()` call is only available on linux, hence currently guarded by `#ifdef __linux__`. We might drop this guard?!

@mairehenke @rjoussen Please have a try whether this solves the corrupt control file writing in your applications.

https://github.com/4C-multiphysics/4C/issues/1476

### DIFF
--- a/src/core/io/src/4C_io_control.cpp
+++ b/src/core/io/src/4C_io_control.cpp
@@ -58,7 +58,12 @@ struct Core::IO::ControlFileWriter::Impl
   void flush_and_reset_tree()
   {
     // Only write if there is anything to write. Otherwise, we get an ugly "[]" in the output.
-    if (tree_.rootref().num_children() > 0) out_ << tree_ << "\n" << std::flush;
+    if (tree_.rootref().num_children() > 0)
+    {
+      // Seek back to where this group started to overwrite the last partial snapshot written
+      if (group_write_start_ >= 0) out_.seekp(group_write_start_);
+      out_ << tree_ << "\n" << std::flush;
+    }
 
     // Clear the tree and reset the current node to the root
     tree_.clear();
@@ -66,6 +71,7 @@ struct Core::IO::ControlFileWriter::Impl
     current_node_id_ = tree_.rootref().id();
     group_level = 0;
     tree_.rootref() |= ryml::SEQ;
+    group_write_start_ = -1;
   }
 
   template <typename T>
@@ -77,10 +83,18 @@ struct Core::IO::ControlFileWriter::Impl
     node << ryml::key(key_str);
 
     emit_value_as_yaml(YamlNodeRef{node, ""}, value);
+
+    // Seek back to where this group started and rewrite the entire (now larger) tree snapshot
+    if (group_write_start_ >= 0) out_.seekp(group_write_start_);
+    out_ << tree_ << std::flush;
   }
 
   void start_group(std::string_view key)
   {
+    // Record the stream position before the outermost group's first byte so write() can seek
+    // back here and overwrite the previous partial snapshot with the latest (larger) one.
+    if (group_level == 0) group_write_start_ = out_.tellp();
+
     ++group_level;
 
     ryml::NodeRef outer = (current().is_seq()) ? current().append_child() : current();
@@ -117,6 +131,10 @@ struct Core::IO::ControlFileWriter::Impl
 
   //! output stream for the control file
   std::ostream& out_;
+
+  //! Stream position of the start of the current outermost group, used for seek-and-overwrite.
+  //! -1 means no group is currently open.
+  std::streamoff group_write_start_{-1};
 };
 
 
--- a/src/core/io/tests/4C_io_control_test.cpp
+++ b/src/core/io/tests/4C_io_control_test.cpp
@@ -144,4 +144,26 @@ namespace
     EXPECT_EQ("", stream.str());
   }
 
+  // Verify that each write() immediately places data in the stream, even before end_group() is
+  // called. This is the core crash-safety guarantee: if end_group() is never reached (e.g. due to
+  // a floating-point exception), all previously written key-value pairs are already in the kernel
+  // buffer and will survive the crash.
+  TEST(ControlFileWriterTest, FPESafetyDataInStreamBeforeEndGroup)
+  {
+    std::stringstream stream;
+    Core::IO::ControlFileWriter writer(true, stream);
+    writer.start_group("field");
+
+    writer.write("step", 1);
+    EXPECT_THAT(stream.str(), testing::HasSubstr("step: 1"));
+
+    writer.write("time", 0.5);
+    EXPECT_THAT(stream.str(), testing::HasSubstr("step: 1"));
+    EXPECT_THAT(stream.str(), testing::HasSubstr("time:"));
+
+    // end_group() is intentionally never called; the assertions above already verified that all
+    // written keys are in the stream. In a real crash the destructor would not run either, but
+    // the data is already in the kernel buffer at each write() call.
+  }
+
 }  // namespace


src/core/io/src/4C_io_control.cpp
```diff
@@ -299,6 +348,13 @@ Core::IO::OutputControl::OutputControl(MPI_Comm comm, std::string problemtype,
   }
 }
 
+Core::IO::OutputControl::~OutputControl()
+{
+#ifdef __linux__
+  if (control_sync_fd_ >= 0) ::close(control_sync_fd_);
```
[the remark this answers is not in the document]
georghammerl: Interesting aspect I did not pay attention. I adapted the destructor calls accordingly.

src/core/io/src/4C_io_control.cpp
```diff
@@ -53,19 +57,32 @@ struct Core::IO::ControlFileWriter::Impl
...
+    {
+      // Seek back to where this group started to overwrite the last partial snapshot written
+      if (group_write_start_ >= 0) out_.seekp(group_write_start_);
+      out_ << tree_ << "\n" << std::flush;
+    }
+
+    // Trigger async writeback to disk such that data survives abnormal termination, e.g. FPE
+#ifdef __linux__
+    if (sync_fd_ >= 0) sync_file_range(sync_fd_, 0, 0, SYNC_FILE_RANGE_WRITE);
+#endif
```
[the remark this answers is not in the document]
georghammerl: done

src/core/io/src/4C_io_control.cpp
```diff
@@ -77,10 +94,21 @@ struct Core::IO::ControlFileWriter::Impl
     node << ryml::key(key_str);
 
     emit_value_as_yaml(YamlNodeRef{node, ""}, value);
+
+    // Seek back to where this group started and rewrite the entire (now larger) tree snapshot
+    if (group_write_start_ >= 0) out_.seekp(group_write_start_);
+    out_ << tree_ << std::flush;
+#ifdef __linux__
+    if (sync_fd_ >= 0) sync_file_range(sync_fd_, 0, 0, SYNC_FILE_RANGE_WRITE);
+#endif
```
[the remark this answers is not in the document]
georghammerl: added error check

src/core/io/src/4C_io_control.cpp
```diff
@@ -77,10 +94,21 @@ struct Core::IO::ControlFileWriter::Impl
...
   }
```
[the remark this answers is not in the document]
georghammerl: Added two unit tests, one which can always run, the second which tests explicitly the FPE behavior which is linux specific.

jeremylt:
With respect to testing this - does 4C have a reliable way to trigger FPE (that most users know not to do) such as zeroing a local parameter value to trigger a divide by zero?

If not, could such a feature be added? With a warn directive and perhaps an error directive if 4C is not built in debug mode? 

With such a feature, then you could write a test that intentionally triggers an FPE on one MPI rank, then the test harness would check that the test terminated with the correct error message/code and that all other ranks correctly wrote out their control file contents.

rjoussen:
@georghammerl Thanks for working on this. I tried my use case, and it works as expected. Btw, this is not only an issue for floating point exceptions, but (I think) any GP level errors that throw.

Regarding the performance, and I'm also not noticing any huge losses, at least for reasonable values for `RESTARTEVERY`.

rjoussen:
> With respect to testing this - does 4C have a reliable way to trigger FPE (that most users know not to do) such as zeroing a local parameter value to trigger a divide by zero?

@jeremylt There definitely is, but I wouldnt rely on that for testing. In fact, every input parameter combination that reliably lead to FPE should be caught before it reaches the gauss point.

jeremylt:
Ah, that makes sense. I'd then personally try to create a way to bypass those safeguards and guard that header with warnings

src/core/io/tests/4C_io_control_test.cpp
```diff
@@ -144,4 +148,60 @@ namespace
...
+
+    writer.write("time", 0.5);
+    EXPECT_THAT(stream.str(), testing::HasSubstr("step: 1"));
+    EXPECT_THAT(stream.str(), testing::HasSubstr("time:"));
+
+    // Destructor runs without end_group() which simulates a crash that bypasses normal cleanup.
+    // The stream must already contain all written keys at this point.
+  }
+
+#ifdef __linux__
```
sebproell: Please use `GTEST_SKIP` in this class and call it in non-Linux case. The code should at least compile for non-Linux.

src/core/io/src/4C_io_control.hpp
```diff
@@ -59,6 +60,11 @@ namespace Core::IO
      */
     void end_all_groups_and_flush();
 
+    /**
+     * Set a file descriptor used for async writeback after each flush.
+     */
+    void set_sync_fd(int fd);
```
sebproell: It's 2026 and setters are so 2000s 😛 Let's just pass this in the constructor. You probably want to create a small struct with the options and set an invalid default there. Also, consider the case where somebody sets this on a non-writing rank. I would FOUR_C_ASSERT this does not happen.

src/core/io/src/4C_io_control.hpp
```diff
@@ -39,7 +39,8 @@ namespace Core::IO
     ControlFileWriter(bool do_write, std::ostream& stream);
 
     /**
-     * Delete the writer object. Flush any pending writes.
+     * Flush any pending writes and destroy the writer.
+     * Must be defined in the .cpp where Impl is a complete type (pimpl requirement).
```
sebproell: Not sure we need to explain C++, but I agree that it is a lesser known thing. I would probably not add this comment, but I am OK with you leaving it if you find this useful.

src/core/io/src/4C_io_control.cpp
```diff
@@ -299,6 +355,19 @@ Core::IO::OutputControl::OutputControl(MPI_Comm comm, std::string problemtype,
...
 }
 
+Core::IO::OutputControl::~OutputControl()
+{
+  control_file_.end_all_groups_and_flush();
+#ifdef __linux__
+  if (control_sync_fd_ >= 0)
+  {
+    control_file_.set_sync_fd(-1);
+    ::close(control_sync_fd_);
```
sebproell: I see, you take ownership of the file and close it in the destructor. That belongs in the docs of set_fd(), or the constructor (see my other comment), IMO.

src/core/io/tests/4C_io_control_test.cpp
```diff
@@ -144,4 +148,60 @@ namespace
...
+    // Subprocess writes partial data to a real file, then crashes with SIGFPE.
+    EXPECT_DEATH(
+        {
+          std::ofstream stream(tmp_path);
+          Core::IO::ControlFileWriter writer(true, stream);
+          writer.start_group("field");
+          writer.write("step", 1);
+          writer.write("time", 0.5);
+          // Crash without calling end_group(); data must already be in the kernel buffer.
+          std::raise(SIGFPE);
```
sebproell: Why go this complicated route? I think the other test is fine to demonstrate the feature works. What is the benefit of specifically triggering SIGFPE?

src/core/io/tests/4C_io_control_test.cpp
```diff
@@ -144,4 +148,60 @@ namespace
...
+        },
+        ""  // accept any crash output
+    );
+
+    // Parent reads the file and verifies the partial data survived.
+    std::ifstream file(tmp_path);
+    ASSERT_TRUE(file.is_open());
```
sebproell: Hard to tell with async operations, but I think we are not guaranteed that the syscall has finished when we reach this point.

src/core/io/src/4C_io_control.cpp
```diff
@@ -77,10 +98,25 @@ struct Core::IO::ControlFileWriter::Impl
...
 
     emit_value_as_yaml(YamlNodeRef{node, ""}, value);
+
+    // Seek back to where this group started and rewrite the entire (now larger) tree snapshot
+    if (group_write_start_ >= 0) out_.seekp(group_write_start_);
+    out_ << tree_ << std::flush;
+#ifdef __linux__
+    if (sync_fd_ >= 0)
+    {
+      int err = sync_file_range(sync_fd_, 0, 0, SYNC_FILE_RANGE_WRITE);
```
sebproell: Please comment why 0 and 0 are the correct arguments here. Not obvious to me.

georghammerl:
> > Not to deteriorate performance, asynchronous writing is chosen (not 100% safe, however first test were all successful.
> 
> Are you sure that we need the performance here? I imagine we write a handful of entries into the control file per time step, so I doubt that the speed of a single write is going to matter much. I would vote against using `sync_file_range`.
> 
> The real fix to the underlying problem is IMO to not write the control file piecemeal, but rather write a whole step at once.

As we are not able to set proper `end_group()` calls throughout the code, this is an engineer's solution which seems to work for 3 out of 3 people (I am aware of) having problems with improper control file writing.

I played around with `fsync()` and `fdatasync()` which are synchronous calls (and thus 100% safe), however  both were massive performance killers (127 sec. -> 149 sec. and 139 sec. -> 264 sec. for two different test cases). Hence, I am quite happy with the current asynchronous solution and vote for using `sync_file_range` :-)

src/core/io/tests/4C_io_control_test.cpp
georghammerl: Definitely true, however this is the interesting aspect that decides on success or not of this approach. If this works repeatedly, it would be great. I can delete this test if it does not run successfully.

src/core/io/tests/4C_io_control_test.cpp
georghammerl: This unexpected occurence of FPE (or similar) down on Gauss Point level is the reason for the changes in this PR. Therefore, I try to mimic this within a unit test.

src/core/io/tests/4C_io_control_test.cpp
georghammerl: Good point. Thanks

src/core/io/src/4C_io_control.hpp
georghammerl: Fine for me. I remove it.

src/core/io/src/4C_io_control.cpp
georghammerl: I do not get it. `OutputControl` and `ControlFileWriter` both use the `fd_` and therefore one of them creates it (= `OutputControl` ) and the second one (=`ControlFileWriter`) gets a copy of `fd_` via the setter. The creator closes the `fd_` in the destructor. What shall be changed?

src/core/io/src/4C_io_control.hpp
georghammerl: Where to set the check? In `set_sync_fd()`? Neither `ControlFileWriter` nor its implementation know the rank. 
The `control_file_stream_` and `fd_` are both created within the `if (write_binary_output_ && myrank_ == 0)` clause and thus exclusively on rank 0.

georghammerl:
> @georghammerl Thanks for working on this. I tried my use case, and it works as expected. Btw, this is not only an issue for floating point exceptions, but (I think) any GP level errors that throw.
> 
> Regarding the performance, and I'm also not noticing any huge losses, at least for reasonable values for `RESTARTEVERY`.

@rjoussen Thanks for this feedback. The proposed solution is expected to work for every crash whatever/whereever it originates from.

src/core/io/src/4C_io_control.cpp
georghammerl: This just means the complete file. I added a comment stating that the whole file is syncd.

rjoussen:
> As we are not able to set proper end_group() calls throughout the code, this is an engineer's solution which seems to work for 3 out of 3 people (I am aware of) having problems with improper control file writing.

I agree, I tried using `end_group()` for the pure solid problems but even there it did not really work (see #1484), and for coupled problems it becomes harder. 
In my opinion, the only good solution would be to have FPE or any crashs communicated to all other ranks properly such that rank 0 calls the control file writer destructor and flushes. This would indeed be super helpful for other reasons too I think.
But right now the main symptom is that the control file gets easily corrupted and this PR fixes that, so I am fine to go along with it.

georghammerl:
> In my opinion, the only good solution would be to have FPE or any crashs communicated to all other ranks properly such that rank 0 calls the control file writer destructor and flushes. This would indeed be super helpful for other reasons too I think.

@rjoussen True, however technically impossible as, e.g., streams get corrupted as soon as an FPE occurs, nicely described by @sebproell  in https://github.com/4C-multiphysics/4C/issues/1476 . There is no safe way to recover from an FPE event.

rjoussen:
@georghammerl Just out of curiosity: Would the "right" way then to catch any FPE on the rank where it happens directly? And then send/communicate that to rank 0 and only throw there?

georghammerl:
> @georghammerl Just out of curiosity: Would the "right" way then to catch any FPE on the rank where it happens directly? And then send/communicate that to rank 0 and only throw there?

When you catch it is already too late. You need to check (for division-by-zero exception) before each division whether it will be a division by zero and then do not perform this division. Instead inform the other procs and write proper control file and shut down the simulation. I would say: This is not feasible.

georghammerl:
I wonder about the failing "unittests_core" due to time out. On my machine the newly introduced tests pass in under 0.5 sec which might be okay for a death test?! Help is appreciated.

src/core/io/src/4C_io_control.cpp
sebproell: If you pass a file descriptor around it is not obvious that whoever receives it should also close the file. I would add a comment for that in the constructor that receives it.

src/core/io/src/4C_io_control.hpp
sebproell: Yeah, I see that this whole class is still quite messy. I guess I only pulled out the bare minimum of the implementation. Not super important

src/core/io/tests/4C_io_control_test.cpp
sebproell: Well, in reality the failure would not be right behind the last sync call, so the chances increase there was enough time. I would still remove the second test -- see also the issues you already face in the Github actions.

sebproell:
Regarding `fsync`: I would expect `std::flush` to trigger something like that under the hood but haven't checked.

src/core/io/src/4C_io_control.cpp
georghammerl: The file descriptor `control_sync_fd_` is opened in the constructor of `OutputControl` and closed in the destructor of `OutputControl` . The copy given to the `ControlFileWriter` (via the setter) is used to perform sync operations only. I am still confused where to put which comment?!

sebproell: Ooooh, nvm. Sorry I did not read this properly. You are right, no comment would be needed.

src/core/io/tests/4C_io_control_test.cpp
georghammerl: I see your point. Removed the second unit test.

georghammerl:
> Regarding `fsync`: I would expect `std::flush` to trigger something like that under the hood but haven't checked.

I tried for my problematic example and it worked without `sync_file_range()`, `std::flush` seems to be "strong" enough.
@mairehenke & @rjoussen Might you please try for your problematic test cases whether control file is still written as expected when you comment out the two calls of `sync_file_range(sync_fd_, 0, 0, SYNC_FILE_RANGE_WRITE)`

I am happy if we can omit this syscall and still have a somehow "reliable" solution.

sebproell:
@georghammerl Nice if that works. To make the life of everyone easier why not drop the syscall in this PR.

Also, I am confused now: is this not the solution you tried initially that gives poor performance? If not, how does it differ?

georghammerl:
> @georghammerl Nice if that works. To make the life of everyone easier why not drop the syscall in this PR.
> 
> Also, I am confused now: is this not the solution you tried initially that gives poor performance? If not, how does it differ?

I revisited my first version: It contains `std::flush()` and `fsync()` which is the safe but poorly performing variant. The better performance came with the async syscall (together with `flush`). However, I have never tried without the syscalls.

Hence, flushing in each writing step (compared to flushing at the end of the simulation as it was before) might be solution. I'll briefly commit the changes. (cleaning/rebasing still open)

mairehenke:
@georghammerl Thanks for working on this! I have now tested my use case and everything works as expected. The performance has not changed much with this version. It went from 139s to about 140s, even though I enabled a restart at every timestep over a total of 200 timesteps. For reasonable restart writing, I did not observe any significant performance overhead

rjoussen:
I also tested my example (tsi with 200 timesteps and `RESTARTEVERY: 1`). The cpu time stayed exactly the same at 128.2sec, even with this frequent restart writing. The flushing worked as expected when I triggered GP failure. I'm really happy with this solution now!

georghammerl:
@mairehenke @rjoussen Thank you for your feedback. I cleaned the implementation. We are now at just a handfull of lines which seem to work as expected. Solution is to `std::flush` in every `write()`. 
@sebproell Thanks for your trivial but very helpful comment.

PR #1874, opened 2026-03-19
Fix missing BOUNDARYID in create_particle_spec

ppraegla:
The BOUNDARYID was introduced in #1856 . This should fix 4C-multiphysics/fourcipp#120


ppraegla:
> @ppraegla thanks for the quick fix 🙏🏻
> 
> have you tried to run the fourcipp testsuite with the new metadata file as well?

No :grimacing: . But I will do it now

ppraegla:
@davidrudlstorfer the tests pass with the new metadata file

davidrudlstorfer:
thanks for checking :)

PR #1875, opened 2026-03-19
Require gmsh in csv comparison test for gmsh input

amgebauer:
I added the `REQUIRED_DEPENDENCIES GMSH` to the csv comparison tests with gmsh mesh input. Note, this test depends on another test that already declares the `REQUIRED_DEPENDENCIES GMSH`. In the nightly pipeline, the test is marked as skipped (see https://github.com/4C-multiphysics/4C/actions/runs/23270351600/job/67670838748), however, on our configurations at LNM, the test still runs, which fails, of course.

~~I don't understand why we actually need this locally and not in our pipeline :thinking: I will try to investigate it further.~~


amgebauer:
@rjoussen I think I understand the issue now. In the `no_optional_dependencies`, we have disabled python and this test implicitly depends on python. Hence, this test is skipped because of missing Python and not because of the skipped FIXTURE. So, I think there is no way around adding the dependency to GMSH also to this test. The respective CMake issue is here https://gitlab.kitware.com/cmake/cmake/-/issues/17980

davidrudlstorfer (approved):
Thanks for fixing this, and also thanks for finding the actual root cause.

Nice that at least one part of our pipeline works again ;)

PR #1898, opened 2026-03-27
Add external python predictor

danielwolff1:
## Description and Context
With this PR, I'm proposing a new predictor class which allows to defer the actually implementation of the predictor to python. We need this feature as we want to study the behavior of Neural Network-based structural predictors. 

### Implementation details
I implemented this feature similarly to what has been done in the context of a python surrogate model for rough surface contact laws as merged in #1735. The general idea of the implementation are:
- Only static cases are supported so far, i.e., load stepping is fine, but no dynamics.
- `4C` execution might happen in parallel, yet the computations in python only happen in a single interpreter, i.e., in a single process.
- Process with id 0 communicates with python.
- The logic for communicating with python via `pybind11` functionality is completely encapsulated in an `Implementation` class so that it does not interfere with the actual predictor.
- As for my personal application case I need to evaluate the neural network on the computational mesh, I need a lightweight representation of the discretization in python. Since this is a special requirement of my own research, the feature is implemented in a way that the communication of the mesh happens in an optional *setup* step, which will simply be skipped, if the python file does not provide a `setup()` method. That way the user can easily control whether to have this once-per-simulation overhead of communicating the mesh or whether they only need the "regular" information that also the C++ predictors see.

### Testing
I have implemented two integration tests where I test the two major different control flows, i.e., once the python script has a setup method and the mesh needs to be communicated and once not. In both cases, the python script implements the simple logic of a `ConstDis` predictor and the corresponding values in the `RESULTS DESCRIPTION` section have been generated by running the same input scripts however with `PREDICT = ConstDis`.

I'm still planning to look into unit tests for this feature but the general implementation already works and won't change anymore, so I already wanted to put it out for review.

I appreciate any comments and insights for potential improvements 😊


src/inpar/4C_inpar_structure.cpp
```diff
@@ -349,23 +349,18 @@ namespace Inpar
...
-                      {"TangDis", pred_tangdis},
-                      {"TangDisConstFext", pred_tangdis_constfext},
+                  {{"Vague", pred_vague}, {"ConstDis", pred_constdis}, {"ConstVel", pred_constvel},
+                      {"ConstAcc", pred_constacc}, {"ConstDisVelAcc", pred_constdisvelacc},
+                      {"TangDis", pred_tangdis}, {"TangDisConstFext", pred_tangdis_constfext},
                       {"ConstDisPres", pred_constdispres},
                       {"ConstDisVelAccPres", pred_constdisvelaccpres},
-                      {"PythonSurrogate", pred_python_surrogate}
-                  },
+                      {"PythonSurrogate", pred_python_surrogate}},
```
mayrmt: The new predictor will be executed in Python, but it is not necessarily a surrogate. ;-)

src/inpar/4C_inpar_structure.cpp
```diff
@@ -349,23 +349,18 @@ namespace Inpar
...
                   {.description = "Type of predictor", .default_value = pred_constdis}),
 
               // File providing the python implementation of the predictor
               parameter<std::filesystem::path>("PYTHON_PREDICTOR_FILE",
-                  {.description = "Path to the Python script implementing the PythonSurrogate predictor.",
+                  {.description =
+                          "Path to the Python script implementing the PythonSurrogate predictor.",
```
mayrmt: Is this the absolute or relative path? Does it matter? I as a user would care about this. If it does not matter, "Absolute or relative path to the ..." would be fine.

src/structure_new/src/predict/4C_structure_new_predict_python_surrogate.cpp
```diff
@@ -0,0 +1,624 @@
...
+  pybind11::dict build_mesh_snapshot(
+      const std::shared_ptr<Core::FE::Discretization>& discret_ptr, unsigned int problem_dim) const;
+
+  /*!
+   * \brief This function builds a dictionary with information that will be passed to python in the
+   * predictor step.
+   *
+   * For now, the dictionary mostly contains the global displacement, velocity, and acceleration
+   * vectors, as well as some iteration depending information such as the current time/load step.
+   * Since the global state vectors are generally stored in a partitioned fashion, this function
```
mayrmt:
```suggestion
   * Since the global state vectors are generally stored in a distributed-memory fashion, this function
```

This also applies to other lines in this PR.

src/structure_new/src/predict/4C_structure_new_predict_python_surrogate.cpp
```diff
@@ -0,0 +1,624 @@
...
+   */
+  pybind11::array_t<double> copy_vector_to_numpy(const Core::LinAlg::Vector<double>& source) const;
+
+  /*!
+   * \brief Convenience function for copying a python-compatible pybind11::array_t<double> to a
+   * Core::LinAlg::Vector<double>
+   *
+   * @param[in] source The python-compatible array instance whose data should be converted to a C++
+   * vector.
+   * @param[out] target The C++ vector instance which holds the data provided by source.
```
mayrmt: To avoid confusion with the other C++ vector, `std::vector<double>`:
```suggestion
   * @param[in] source The python-compatible array instance whose data should be converted to a Core::LinAlg::Vector<double>
   * vector.
   * @param[out] target The Core::LinAlg::Vector<double> instance which holds the data provided by source.
```

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
```diff
@@ -0,0 +1,624 @@
...
+  bool has_setup_ = false;
+  //! Flag indicating whether an Implementation instance has been set up.
+  bool is_setup_ = false;
+};
+
+Solid::Predict::PythonSurrogate::Implementation::Implementation(
+    const std::filesystem::path& python_filename)
+{
+  // Initialize interpreter once per process.
+  static pybind11::scoped_interpreter py_guard;
```
mayrmt: You mentioned that only rank 0 should interact with Python. So far, this is not checked here. Just an observation. I am not sure yet, if an actual check is required.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
```diff
@@ -0,0 +1,624 @@
...
+    pybind11::object result = compute_(state);
+    // update 4C's internal state based on the result from the python script
+    apply_prediction(result, gathered, global_state_ptr);
+  }
+  else
+  {
+    // all other processes enter apply_prediction and wait for python to be finished.
+    // Once python is done, the results will be communicated back to each processor.
+    apply_prediction(pybind11::none(), gathered, global_state_ptr);
+  }
```
mayrmt: I would have expected a comm barrier for all processes to wait for proc 0 to finish... I do not fully understand yet, why and how this works.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
```diff
@@ -0,0 +1,624 @@
...
+  std::map<int, std::vector<int>> local_element_conn;
+
+  // iterate over all nodes on the current processor and fill node-based data
+  for (int i = 0; i < discret_ptr->num_my_row_nodes(); ++i)
+  {
+    const Core::Nodes::Node* node = discret_ptr->l_row_node(i);
+    const int node_id = node->id();
+
+    // copy the node's coordinates into a simple STL container
+    std::vector<double> coords(3, 0.0);
```
mayrmt: This could probably be an `std::array`, since the dimension is always 3.

src/structure_new/src/predict/4C_structure_new_predict_python_surrogate.cpp
```diff
@@ -0,0 +1,624 @@
...
+      FOUR_C_THROW("Node {} has only {} DOFs, expected at least {} displacement DOFs.", node_id,
+          dofs.size(), problem_dim);
+    // copy only problem_dim entries of the dofs
+    local_node_dofs[node_id] = std::vector<int>(dofs.begin(), dofs.begin() + problem_dim);
+  }
+
+  // iterate over all elements and fill element-based data
+  for (int i = 0; i < discret_ptr->num_my_row_elements(); ++i)
+  {
+    const Core::Elements::Element* element = discret_ptr->l_row_element(i);
```
mayrmt:
```suggestion
  for (auto element : discret_ptr_->my_row_element_range())
  {
```

src/structure_new/src/predict/4C_structure_new_predict_python_surrogate.cpp
```diff
@@ -0,0 +1,624 @@
...
+{
+  // STL containers for a lightweight per-processor mesh representation
+  std::map<int, std::vector<double>> local_node_coords;
+  std::map<int, std::vector<int>> local_node_dofs;
+  std::map<int, std::vector<int>> local_element_conn;
+
+  // iterate over all nodes on the current processor and fill node-based data
+  for (int i = 0; i < discret_ptr->num_my_row_nodes(); ++i)
+  {
+    const Core::Nodes::Node* node = discret_ptr->l_row_node(i);
```
mayrmt:
```suggestion
  for (auto node : discret_ptr_->my_row_node_range())
  {
```

src/structure_new/src/predict/4C_structure_new_predict_python_surrogate.cpp
```diff
@@ -0,0 +1,624 @@
...
+Solid::Predict::PythonSurrogate::~PythonSurrogate() = default;
+
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+void Solid::Predict::PythonSurrogate::setup()
+{
+  check_init();
+
+  const auto sdyn_params = Global::Problem::instance()->structural_dynamic_params();
+  // perform a sanitiy check and warn the user if they try to use this class
```
mayrmt:
```suggestion
  // perform a sanity check and warn the user if they try to use this class
```

src/structure_new/src/predict/4C_structure_new_predict_python_surrogate.cpp
```diff
@@ -0,0 +1,624 @@
...
+ *----------------------------------------------------------------------------*/
+void Solid::Predict::PythonSurrogate::compute(::NOX::Abstract::Group& grp)
+{
+  check_init_setup();
+
+  // Compute new disnp_ptr, velnp_ptr, accnp_ptr via call to python script
+  python_implementation_->compute(global_state_ptr());
+
+  impl_int().model_eval().predict(get_type());
+}
```
mayrmt: `grp` is not used in this function. Do you really need to pass it as input argument?

tests/input_files/structure_new_python_wrapper_predictor_constdis.4C.yaml
```diff
@@ -0,0 +1,146 @@
...
+  TIMESTEP: 0.5
+  NUMSTEP: 2
+  MAXTIME: 1
+  PREDICT: "PythonSurrogate"
+  PYTHON_PREDICTOR_FILE: "./structure_new_python_surrogate_predictor_constdis.py"
+  TOLDISP: 1e-12
+  TOLRES: 1e-12
+  LINEAR_SOLVER: 1
+  NLNSOL: "fullnewton"
+  MAXITER: 20
```
mayrmt: This performs a full Newton solve, so the final displacement values in the result tests will be independent of the outcome of the predictor. Hence, this test might not test the outcome of the predictor, just that the code path is not broken.

tests/input_files/structure_new_python_wrapper_predictor_constdis_with_setup.4C.yaml
```diff
@@ -0,0 +1,146 @@
...
+  TIMESTEP: 0.5
+  NUMSTEP: 2
+  MAXTIME: 1
+  PREDICT: "PythonSurrogate"
+  PYTHON_PREDICTOR_FILE: "./structure_new_python_surrogate_predictor_constdis_with_setup.py"
+  TOLDISP: 1e-12
+  TOLRES: 1e-12
+  LINEAR_SOLVER: 1
+  NLNSOL: "fullnewton"
+  MAXITER: 20
```
mayrmt: This performs a full Newton solve, so the final displacement values in the result tests will be independent of the outcome of the predictor. Hence, this test might not test the outcome of the predictor, just that the code path is not broken.

mayrmt (commented):
@danielwolff1 I left some comments and raised some questions, that I could not answer for myself right away. For me, it looks like this is going into the right direction.

src/inpar/4C_inpar_structure.cpp
danielwolff1: Fair point, I renamed it to `PythonWrapper`.

src/inpar/4C_inpar_structure.cpp
danielwolff1: So far, both seems to be working fine (I used an absolute path while I was developing the feature and now the tests use a relative path), so I adapted it accordingly.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
danielwolff1: This is a good point. I've thought about it the last couple of days and my explanation is: 
I think it would also work if I rewrote it like this:
```C++
// create an empty result object (necessary for processes with non-zero ID)
pybind11::object result = pybind11::none();
// Only process with rank 0 communicates with python
if (is_rank_zero(global_state_ptr->get_dis_n()->get_comm()))
{
    // make it safe to touch python objects for the rest of this function
    pybind11::gil_scoped_acquire gil;
    // invoke the python predictor step in the python script
    result = compute_(state);
    // update 4C's internal state based on the result from the python script
    apply_prediction(result, gathered, global_state_ptr);
}
// explicitly synchronise all processes, i.e., wait for process 0 to be done with python communication
Core::Communication::barrier(global_state_ptr->get_discret()->get_comm());
// Once processor 0 is done with python, the results will be communicated back to each processor.
apply_prediction(result, gathered, global_state_ptr);
```
Right now, the communication happening in `apply_prediction`, i.e., the scattering of the results that process 0 received from python involves an implicit barrier, i.e., the current implementation let's all processes already entering the receiving state, but of course they will only receive the data, once processor 0 starts scattering them, i.e., after python communication is finished. So the communication in `apply_prediction` involves an implicit barrier. 

Does that make sense or do you disagree?

If you prefer, I can adapt the implementation to make the synchronisation more explicit.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
danielwolff1: True, but then I would also need to adapt the type of the container `local_node_coords` which is currently declared as `std::map<int, std::vector<double>>` because  `Core::LinAlg::gather()` is already overloaded for a container of type `std::map<int, std::vector<double>>` (see [here](https://github.com/4C-multiphysics/4C/blob/main/src/core/linalg/src/dense/4C_linalg_utils_densematrix_communication.hpp#L275)) which simplifies the communication. But I can also check if `gather` would work for a container of type ``std::map<int, std::array<double, 3>>``, at least there seems to be a more general templated version available [here](https://github.com/4C-multiphysics/4C/blob/main/src/core/linalg/src/dense/4C_linalg_utils_densematrix_communication.hpp#L351).

src/structure_new/src/predict/4C_structure_new_predict_python_surrogate.cpp
danielwolff1: I think so unfortunately, since the signature of this method is inherited from the parent class `Solid::predict::Generic`, see [here](https://github.com/4C-multiphysics/4C/blob/main/src/structure_new/src/predict/4C_structure_new_predict_generic.hpp#L82).

tests/input_files/structure_new_python_wrapper_predictor_constdis.4C.yaml
danielwolff1: Yeah I thought that's what the integration tests are for and I wanted to provide a unit test for testing the actual functionality of the predictor. However, this is not as straightforward as I initially thought: In my opinion it would make more sense to test `solid::predict::PythonWrapper::Implementation` because this class provides the actual functionality, rather than `solid::predict::PythonWrapper` itself, which is only a wrapper. However, the `Implementation` class is currently private within `PythonWrapper`, which makes it very inconvenient to test. 

So I guess I will adapt this test like we discussed and modify the newton iteration such that I can actually "see" the influence of my predictor, unless you have a better idea? 🤔

danielwolff1 (commented):
Thanks for the fast review, I appreciate the input! I still have to look into some of the points you mentioned and will do so next week probably.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
```diff
@@ -0,0 +1,625 @@
...
+{
+  // create a new NumPy array
+  pybind11::array_t<double> target(source.local_length());
+  auto target_view = target.mutable_unchecked<1>();
+  const double* source_view = source.get_values();
+  // copy data from vector into the NumPy array
+  for (int i = 0; i < source.local_length(); ++i) target_view(i) = source_view[i];
+
+  return target;
+}
```
amgebauer: Can we avoid to copy the data? If I'm not mistaken, we could just return a view to the data

```suggestion
pybind11::array_t<const double> Solid::Predict::PythonWrapper::Implementation::copy_vector_to_numpy(
    const Core::LinAlg::Vector<double>& source) const
{
   return pybind11::array_t<const double>(
            {source.local_length()},
            {sizeof(double)},
            source.get_values()
        );
}
```

This might even be general enough so that we can move this into a helper function close to our `Core::LinAlg::Vector`.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
```diff
@@ -0,0 +1,625 @@
...
+
+void Solid::Predict::PythonWrapper::Implementation::copy_numpy_to_vector(
+    const pybind11::array_t<double>& source, Core::LinAlg::Vector<double>& target) const
```
amgebauer: You could also do it the other way around, i.e., directly writing into the data of our `Core::LinAlg::Vector` (same as above, just non-const)

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
danielwolff1: Thanks for raising this point! Took me a while to investigate it and find a solution.

In the first draft of the implementation, communication was indeed restricted to rank 0, i.e., only rank 0 would call the python interpreter, however, `pybind11` data types were still used on other processes. Yet, in order to allow this, python needed to be initialized on all processes (which in itself was fine, as it didn't result in weird/undefined behavior during runtime, but as you rightfully pointed out, the implementation was not strictly following the documentation, as indeed a python interpreter was initialized on each process, thus creating an overhead as ultimately only rank 0 meaningfully used it later on).

With commit 4ba3d3c11b99a2460d35563bd5e3825ed4f53a72 I revised the implementation so that it now strictly follows the self-imposed "only rank 0 communicates with python"-rule. That is, initialization of the python interpreter now only happens on rank 0. All processes participate in communication, but only rank 0 interacts with python and only rank 0 actually touches `pybind11` data types, meaning all conversion between 4C-internal and python-compatible data types happen on rank 0.

Overall, the implementation should be much cleaner now and some of the bigger functions (like `build_mesh_snapshot()`) have now been split up in two, namely gathering data from other processes in a first step and converting the data from C++ to python-compatible data types in a second step.

tests/input_files/structure_new_python_wrapper_predictor_constdis.4C.yaml
mayrmt: Yes, I agree that one would ideally test the `Implementation`, though that is probably hard to do.

If you want to test not only the success of the simulation, but the actual outcome of the predictor within an integration test, then you need to "bypass" the Newton solver.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
danielwolff1: Thanks for the comment! I adapted the `copy_vector_to_numpy` function as you suggested. If you think it's general enough to implement it as a utility function, I'd be happy to move the implementation somewhere (and possible rename it to something more expressive like `create_numpy_view()` or similar).

As for your remark regarding the other way around: I'm not sure I entirely understand. Since the return value is created by a computation in python, the only way I would see to prevent a copy here would be if we allowed python to directly write into the memory of the `Core::LinAlg::Vector` owned by C++ which would mean we would already have to provide a non-const view to python via the `copy_vector_to_numpy` function?. Is that what you had in mind?

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
```diff
@@ -0,0 +1,687 @@
...
+  // Helper structs for data encapsulation
+
+  //! Convenience struct for encapsulating the gathered mesh information in one place.
+  struct GatheredMesh
+  {
+    // std::vector<int> node_ids;
+    std::map<int, std::vector<double>> node_coords;
+    std::map<int, std::vector<int>> node_dofs;
+    // std::vector<int> element_ids;
+    std::map<int, std::vector<int>> element_conn;
```
isteinbrecher: Is there a reason you store this in maps and not vectors?

isteinbrecher (dismissed):
Thanks for this feature @danielwolff1 ! Nice to see how easy it is to incorporate python within 4C.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
danielwolff1: Good question! So maybe that's overly complicated but I was trying to account for the fact that when I gather the mesh information from all other processes on rank 0, the resulting gathered vector does not need to contain the nodes ordered according to their global node IDs, which is why I decided to go for maps here, so that I can store the data in the form of `(global_node_id, node_coordinates)` (and similarly for the element connectivity and dof information). Alternatively, I could probably change the maps to vectors and include additional vectors that just contain the indices that are currently stored in the `.first` member of each map entry.

isteinbrecher: Thanks for the explanation, that makes a lot of sense!

tests/input_files/structure_new_python_wrapper_predictor_constdis.4C.yaml
danielwolff1: Done, please have a look and let me know if this is how you envisioned is. 

I basically now implemented a dummy predictor in the python script, which predicts at each node of the computational mesh a displacement of $\mathbf{d}^{(n+1)}= \mathbf{d}^{(n)}+(1,-2,3)^\top$ over a total of 2 time/load steps, thus resulting in an expected solution displacement of $\mathbf{d}=(2,-4,6)^\top$ in each node. This is checked in the tests and works.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
amgebauer: yes, I would provide both, a const version and a non-const version. So if you give the create_numpy_view a `const Vector<double>()` it should create a const view, and if it is a `Vector<double>() `(non-const), it should be a non-const view that python can change. The underlying memory are just contiguous doubles, so that it is perfectly compatible with numpy.

danielwolff1: okay, thanks for the clarification, I'll add the functions and test it!

danielwolff1: Done with commit 0ce386e406899edc581ae05a607cdbf628ee102c. 
In the end, I needed to create an additional `pybind11::capsule` object when creating the view which apparently tells a `pybind11` object that its lifetime is tied to an externally owned instance and generates a mutable view. Without that, modifications performed in python would not affect the C++ side. 
Moreover, simply trying to generate a const view by returning a `pybind11::array_t<const double>` did not mark the object as not writable in python. I added an additional line which explicitly sets a `writable` flag to `false` and now trying to modify this array from within a python script results in a `ValueError`.

src/structure_new/src/implicit/4C_structure_new_impl_generic.hpp
```diff
@@ -176,6 +176,9 @@ namespace Solid
       //! Return linear error coefficient of velocities
       virtual double method_lin_err_coeff_vel() const = 0;
 
+      //! Return a pointer to the container of the structural dynamic data (read only)
```
amgebauer:
```suggestion
      //! Return a reference to the container of the structural dynamic data (read only)
```

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
```diff
@@ -0,0 +1,677 @@
...
+
+  /*!
+   * \brief Convenience function for creating a non-constant view of a Core::LinAlg::Vector<double>
+   * which will be wrapped into a python-compatible pybind11::array_t<double>.
+   *
+   * @param[in] source The C++ vector instance whose data should be returned in a python-compatible,
+   * mutable format.
+   * \return The mutable python-compatible representation of the C++ data.
+   */
+  pybind11::array_t<double> create_numpy_view(Core::LinAlg::Vector<double>& source) const;
```
amgebauer: Can you move these into core close to where we define `Core::LinAlg::Vector`?

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
```diff
@@ -0,0 +1,677 @@
...
+  GatheredState gathered = gather_state_to_rank_zero(global_state_ptr);
+
+  // Only process with rank 0 communicates with python
+  if (is_rank_zero())
+  {
+    // make it safe to touch python objects for the rest of this function
+    pybind11::gil_scoped_acquire gil;
+    // convert the gathered data to a python-compatible dictionary that will be passed to the python
+    // interpreter
+    const pybind11::dict state = build_python_state(global_state_ptr, gathered);
```
amgebauer: Instead of a dict, it would be nicer if you create a small struct and bind each members directly. In that case, your ide can tell you what variables are part of the struct (and don't rely on strings at runtime). Similar in some other cases.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
```diff
@@ -0,0 +1,677 @@
...
+  context["input_file"] = output_control->input_file_name();
+  context["mesh"] =
+      convert_gathered_mesh_to_python_dict(gathered_mesh, global_state_ptr->get_dim());
+  context["problem_dim"] = global_state_ptr->get_dim();
+  context["step_max"] = sdyn.get_step_max();
+  context["time_max"] = sdyn.get_time_max();
+  return context;
+}
+
+pybind11::dict Solid::Predict::PythonWrapper::Implementation::convert_gathered_mesh_to_python_dict(
```
amgebauer: Also here, can't we just write a binding for the GatheredMesh?

isteinbrecher (dismissed):

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
danielwolff1: I decided against bindings for the `GatheredMesh` directly as this is merely a structure meant for aiding communication to rank 0 and still contains purely C++ data, but I added separate structs which contain `pybind11` data and are sent to python. Access still works as before, and the data of the state vectors (displacement, velocity, acceleration) is still not copied but provided as a (const, where applicable) view.

src/structure_new/src/predict/4C_structure_new_predict_python_wrapper.cpp
danielwolff1: Done, I created a new file `src/core/linalg/src/sparse/4C_linalg_vector_numpy_utils.hpp` and placed the functionality there.

isteinbrecher (approved):

amgebauer (approved):

mayrmt (approved):

mayrmt:
@danielwolff1 You could squash the commits.

amgebauer:
@danielwolff1 could you have a look at the failing nightly pipeline? https://github.com/4C-multiphysics/4C/actions/runs/24694150094/job/72223568409

danielwolff1:
After a brief look, the tests fail because `numpy` is not known to the python interpreter used to run the tests, which essentially means that the runner uses a different python environment than the one that is generally available in 4C (because this features `numpy` as a dependency). I encountered this error already earlier which is why I adapted the workflow file [here](https://github.com/danielwolff1/4C/blob/78dc93c7b03f23499afea7268973bb9591c720cb/.github/workflows/buildtest.yml#L80) in the attempt to make the installed python dependencies known to the executing python interpreter. 

I'll have to investigate this more closely and will do so as quickly as possible, but unfortunately I'm at a conference this week, so it might take a couple of days.

amgebauer:
> I encountered this error already earlier which is why I adapted the workflow file [here](https://github.com/danielwolff1/4C/blob/78dc93c7b03f23499afea7268973bb9591c720cb/.github/workflows/buildtest.yml#L80) in the attempt to make the installed python dependencies known to the executing python interpreter.

That explains why the nightly pipeline if failing. These tests are defined in `nightly_tests.yml` and not in `buildtest.yml`. However, I'm not sure whether the added modification of the python path is what we want actually.

Edit: I think it makes more sense to just activate the corresponding python environment so that the correct environment is in the path during testing. Then, 4C should pick the correct one and should find all dependencies.

danielwolff1:
okay perfect, I didn't know that we already have a python environment (or how to activate it). Do you know if I can simply run `. venv/bin/activate` instead of modifying the paths as I do at the moment? I will push a quick fix then.

PR #1918, opened 2026-03-31
Feature: Support named point sets in VTU files

bennoschoenstein:
## Summary

Closes #1907

VTU files can contain named integer PointData arrays (e.g. `X_0`, `N_00`)
that represent node sets. Previously, only arrays named `point_set_N` were
recognized. This PR adds support for arbitrarily named arrays, enabling
`NODE_SET_NAME` references in 4C input files.

## Changes

- `4C_io_vtu_reader.cpp`: New helper `is_integral_array`, struct `PointSetEntry`,
  and function `get_point_set_arrays` replace the old `get_numbered_arrays_with_prefix`.
  Named arrays are assigned IDs starting after the highest existing numeric ID
  (backward compatible with `point_set_N` convention).
- New regression test `solid_vtu_named_point_sets` verifying `NODE_SET_NAME` works
  with VTU input files.


src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -232,30 +232,63 @@ namespace
...
+  struct PointSetEntry
+  {
+    int id;
+    std::optional<std::string> name;
+    std::reference_wrapper<vtkDataArray> array;
+  };
+
+  // Returns all single-component integer PointData arrays as point-set entries.
+  // Arrays named "point_set_<N>" keep their numeric ID (backward compatibility).
+  // All other arrays are assigned IDs starting after the highest numeric ID found.
```
isteinbrecher: Is this the approach we desire? Do we have cases where we have point data that we don't want to end up as node sets? Maybe a saver approach would be to interpret all point sets named like `point_set_here_comes_the_name` as node sets.

bennoschoenstein: In my RVE simulations I would like to refer to the corner nodes and the surfaces with names rather than an arbitraty ID, since I have 14 IDs in 3D and its hard to remember which is which. 
But of course I could also name it `point_set_arbitrary_name`, but for new users I thought it would be more intuitive to just use `arbitrary_name` since then you don't even have to know about the specifics of these kind of things, such as `point_set_XX`

src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -232,30 +232,63 @@ namespace
...
-      arrays.emplace(set_id, data);
+      if (array_name.size() > prefix.size() && array_name.starts_with(prefix))
+      {
+        const std::string suffix = array_name.substr(prefix.size());
+        if (!suffix.empty() && suffix.find_first_not_of("0123456789") == std::string::npos)
+        {
+          int id = std::stoi(suffix);
+          max_numeric_id = std::max(max_numeric_id, id);
+          numbered.push_back({id, std::nullopt, array_ref});
+          continue;
```
amgebauer, line 283: I think we should also allow to refer to `point_set_1` as name to a pointset. It would be weird if those names magically disappear.

amgebauer (commented):
Thanks for your contribution! I also though about adding this when @rjoussen enabled name-based input

src/core/io/src/4C_io_vtu_reader.cpp
bennoschoenstein: It is still possible to just use 'point_set_'.
Bascially how it works, if you define 'point_set_1', 'point_set_2' and let's say we skip 'point_set_3' and then add 'point_set_4'. Then the IDs 1,2 and 4 are gone. The ID 3 is just skipped.
If we then want to add extra point_sets but with names like 'traction', then this will get the highest point_set ID +1 (internally), so in this case the ID with the number 5.

The corresponding code is in line 261-262.

Does this answer your concern?

amgebauer: No, I mean sth different. If I'm not mistaken, in your current approach, if a point set is called point_set_1, it gets the id 1 but does not get a name. I would rather suggest that it can also be referred with the name "point_set_1" (for consistency).

src/core/io/src/4C_io_vtu_reader.cpp
rjoussen: I agree with @bennoschoenstein. Having to know that point sets need to start with `point_set` adds unnecessary complexity. Plus, its never harmful to read a field as a node set even if that was not the intent. If its not referenced in the boundary conditions it is just dropped.

src/core/io/src/4C_io_vtu_reader.cpp
rjoussen: > I think we should also allow to refer to point_set_1 as name to a pointset. It would be weird if those names magically disappear.

This might be a breaking change though, since how it is set up at the moment, the node_set_name is used for writing monitor_dbc files. So these files wil be named differently afterwards.

amgebauer: I would still do it. Vtu input is still very young in 4C so I don't expect a large impact with this breaking change.

src/core/io/src/4C_io_vtu_reader.cpp
isteinbrecher: > If its not referenced in the boundary conditions it is just dropped.

If that's the case I am fine with using the full point data name.

amgebauer:
@bennoschoenstein I really like this feature. Is there any blocking item, or anything I can help you with so that we can get this merged?

bennoschoenstein:
@amgebauer Thank you for your comment! I will try to finish it this week, I was just very busy in preparing the lecture material for this new semester.

bennoschoenstein:
I addressed the open review points:

@amgebauer `point_set_<N>` arrays are now also referenceable by name.
The entry's `name` field is populated with the full array name for all
entries (numbered and named), so `NODE_SET_NAME: point_set_1` now works
in addition to `NODE_SET_ID: 1`.

Additionally: 
- Named arrays are sorted alphabetically
by array name before IDs are assigned. This makes the ID-to-name mapping reproducible across runs and platforms.
- An entry in `mesh.point_sets` is only created when at least one point is actually marked as part of the set. Arrays that contain only zeros (or don't match any point) no longer create empty pointsets.

src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -232,30 +233,69 @@ namespace
...
+          numbered.push_back({id, array_name, array_ref});
+          continue;
+        }
+      }
+      named.push_back({0, array_name, array_ref});
     }
 
-    return arrays;
+    std::sort(named.begin(), named.end(),
+        [](const PointSetEntry& a, const PointSetEntry& b) { return a.name < b.name; });
```
amgebauer, line 290: Don't they appear in a specific order in the vtu file, which we can just use to define the ids?
Note, this will not sort the names in the standard human-way since it is case sensitive. I'm fine with that since I don't think that we should/need to specify how we assign internal id's on pointsets without an id. Making it consistent across runs is a good idea though.

src/core/io/src/4C_io_vtu_reader.cpp
```diff
@@ -232,30 +233,69 @@ namespace
...
 
-      arrays.emplace(set_id, data);
+      if (array_name.size() > prefix.size() && array_name.starts_with(prefix))
+      {
+        const std::string suffix = array_name.substr(prefix.size());
+        if (!suffix.empty() && suffix.find_first_not_of("0123456789") == std::string::npos)
+        {
+          int id = std::stoi(suffix);
+          max_numeric_id = std::max(max_numeric_id, id);
+          numbered.push_back({id, array_name, array_ref});
```
amgebauer: Just a minor comment: You can also directly create it in-place (and not creating it and then pushing it on the vector).
```suggestion
          numbered.emplace_back(id, array_name, array_ref);
```

The same applies to a few other places as well.

amgebauer (dismissed):
Thank you for finishing it! I only have a few minor comments. I'm also fine with merging it as is.

rjoussen (dismissed):

src/core/io/src/4C_io_vtu_reader.cpp
bennoschoenstein: I have implemented this now too.

amgebauer (approved):

rjoussen (approved):

PR #1956, opened 2026-04-09
Refactor Return Mapping (Viscoplastic Material)

dragos-ana:
- Refactors return mapping and the Local Newton Loop for the transversely isotropic viscoplastic
material.
- Local Newton:
    * Added new logic for divergence handling: stop, continue, continue with safeguards
    * Added several possibilities for convergence checking (residual, increment ratio, and combined)
    * Related parameters are grouped
    * Added unit tests for the new logic mechanisms.
    * Added a function to manage evaluation after residual, Jacobian, and linear system depending on the encountered error.
    
    
    Update after review:
        

-  Substepping and one-step procedures were accommodated into a viscoplastic-correction function, which calls the Local Newton loop for each substep.
-  The Local Newton loop was cleaned up further and divided into several functions.
-  Moreover, a group was added for the local substepping parameters.
- Both LocalNewtonManager and LocalSubsteppingUtils are now classes instead of structs.
- Substepping is now also tested using a small unit test, which shows that a challenging deformation scenario can be solved by employing local substepping, and a framework test based on a one-step test utilizing a higher global time step.


src/mat/4C_mat_inelastic_defgrad_factors_service.cpp
```diff
@@ -289,6 +296,54 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::TimeStepQuantities:
...
+      "You already called resize for the Local Newton manager! The number of current GP is {} and "
+      "you attempt to set it to {}",
+      curr_num_iters.size(), numgp);
+
+  // resize arrays
+  curr_num_iters.resize(numgp, curr_num_iters[0]);
 
 
+  resize_called_ = true;
+}
```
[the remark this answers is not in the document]
dragos-ana: This compiles without problems. But it is still inconsistent in regards to how I named the other variables; I have therefore adapted the name.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -1725,6 +1750,19 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::pre_evaluate(
...
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+void Mat::InelasticDefgradTransvIsotropElastViscoplast::prepare_return_mapping(
+    const Core::LinAlg::Matrix<3, 3>& defgrad)
+{
+  // pre-evaluate viscoplastic law
+  viscoplastic_law_->pre_evaluate(params_, gp_);
+
+  // set local iteration to 0
+  local_newton_manager_.iter = 0;
```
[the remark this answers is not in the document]
dragos-ana: Removed the defgrad as an input parameter. I would still leave the cheap viscoplastic law pre-evaluation there, since it is safe and required for return mapping, depending on the viscoplastic law employed.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2992,14 +3065,12 @@ Core::LinAlg::Matrix<10, 1> Mat::InelasticDefgradTransvIsotropElastViscoplast::l
 {
   // auxiliaries
   Core::LinAlg::Matrix<10, 10> temp10x10(Core::LinAlg::Initialization::zero);
+  Core::LinAlg::Matrix<10, 1> temp10x1(Core::LinAlg::Initialization::zero);
```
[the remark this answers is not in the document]
dragos-ana: I now use temp10x10 for solution instead of JacMat directly.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3030,60 +3101,154 @@ Core::LinAlg::Matrix<10, 1> Mat::InelasticDefgradTransvIsotropElastViscoplast::l
...
-        if (!new_substep_status) return sol;  // return with error
-        continue;
-      }
+      ++local_newton_manager_.iter;
 
       // compute residual
       residual = calculate_local_newton_loop_residual(curr_CM, sol,
           time_step_quantities_.last_substep_plastic_defgrad_inverse[gp_],
           time_step_quantities_.last_substep_plastic_strain[gp_], local_substepping_utils_.curr_dt,
           err_status);
```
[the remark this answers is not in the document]
dragos-ana: Since the error type does not seem particularly relevant, I removed it entirely for now, and may introduce it again in the future.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2900,7 +2905,40 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                                 "logarithm w.r.t. matrix",
                     .default_value =
                         Core::LinAlg::GenMatrixLogFirstDerivCalcMethod::pade_part_fract}),
-        },
+            group("LOCAL_NEWTON",
+                {parameter<
+                     Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LocalNewtonConvCheck>(
+                     "CONV_CHECK", {.description = "convergence check type"}),
+                    parameter<int>("MAX_ITER", {.description = "maximum number of iterations"}),
+                    parameter<double>(
```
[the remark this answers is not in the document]
dragos-ana: I added validators for all relevant parameters. Also, I made the entire group optional for now, although I am not sure whether I will not retract this in the future. The poro viscoplastic tests do not contain the group anymore within their yaml files.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -413,6 +413,27 @@ namespace
...
+            .get<ViscoplastUtils::LocalNewtonConvCheck>("CONV_CHECK"),
+        .diver_cont = matdata.parameters.group("LOCAL_NEWTON")
+            .get<ViscoplastUtils::LocalNewtonDiverCont>("DIVER_CONT"),
+        .max_iter = static_cast<unsigned int>(
+            matdata.parameters.group("LOCAL_NEWTON").get<int>("MAX_ITER")),
+        .max_exceedance_fact_res_tol =
+            matdata.parameters.group("LOCAL_NEWTON").get<double>("MAX_EXCEEDANCE_FACT_RES_TOL"),
+        .max_exceedance_fact_incr_tol =
+            matdata.parameters.group("LOCAL_NEWTON").get<double>("MAX_EXCEEDANCE_FACT_INCR_TOL"),
+    };
```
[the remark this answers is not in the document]
dragos-ana: Group is optional now, and MAX_ITER is validated.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3112,46 +3277,199 @@ Core::LinAlg::Matrix<10, 1> Mat::InelasticDefgradTransvIsotropElastViscoplast::l
...
+          case FourC::Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LocalNewtonDiverCont::
+              ContinueWithSafeGuard:
+          {
+            // throw warning
+            std::cout << "WARNING: The Local Newton Loop for ele_gid = " << ele_gid_
+                      << ", gp = " << gp_ << " did not reach convergence after "
+                      << local_newton_manager_.params.max_iter
+                      << " iterations: residualNorm2 = " << residualNorm2
+                      << ", increment = " << rel_sol_incr_norm << std::endl;
+
```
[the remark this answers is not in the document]
dragos-ana: I now only show the warning on the owner rank, determined by searching for the discretization 'structure'. If it is not found, then a message is displayed to signal that something may be wrong with the simulated problem, since the structure discretization has to be available for this material. 
I explicitly did not throw here: the unit test `4C_inelastic_defgrad_factors_test.cpp` would have failed otherwise, because the discretization is not available, and adding a dummy discretization only to show warnings seems rather ugly.
I am open to other proposals here.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3304,6 +3625,47 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+  {
+    const bool new_substep_status = prepare_new_substep(sol, curr_CM);
+    if (!new_substep_status)
+    {
+      std::cout << get_error_info(
+                       "Could not find a suitable substep which converges for the given "
+                       "settings!")
+                << std::endl;
+      err_action = Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorAction::
+          return_solution_with_errors;  // return with error
```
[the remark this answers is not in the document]
dragos-ana: This is removed now; I don't return with errors anymore, but directly throw.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -622,7 +682,8 @@ Mat::PAR::InelasticDefgradTransvIsotropElastViscoplast::
       !all_yield_cond_param_specified)
   {
     FOUR_C_THROW(
-        "You are attempting to simulate transversely isotropic behavior but have not specified all "
+        "You are attempting to simulate transversely isotropic behavior but have not specified "
+        "all "
         "yield function parameters!");
```
rjoussen:
```suggestion
        "all yield function parameters!");
```

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2858,6 +2858,11 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                         "evolution equation for the plastic deformation gradient -> default)",
                     .default_value = Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::
                         TimIntType::logarithmic}),
+            parameter<bool>("USE_SUBSTEPPING",
+                {.description =
+                        "boolean: use substepping in the Local Newton Loop? (true: yes, false: no)",
```
rjoussen: the `yes`/`no` explanation seems outdated, I think the simple fact that it is a boolean is enough.
```suggestion
                        "Use substepping in the Local Newton Loop?",
```

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2900,7 +2905,50 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
-        },
+            group("LOCAL_NEWTON",
+                {parameter<
+                     Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LocalNewtonConvCheck>(
+                     "CONV_CHECK",
+                     {.description = "convergence check type",
+                         .default_value = Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::
+                             LocalNewtonConvCheck::residual_and_increment_ratio}),
+                    parameter<int>("MAX_ITER", {.description = "maximum number of iterations",
+                                                   .default_value = 100,
```
rjoussen, line 2918: Is `100` a good default here? It seems rather large.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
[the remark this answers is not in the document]
rjoussen: I think, for ease of use, it would be nice if this group stayed optional.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -10,6 +10,7 @@
 #include "4C_global_data.hpp"
 #include "4C_inpar_structure.hpp"
 #include "4C_io_input_field.hpp"
+#include "4C_io_input_parameter_container.hpp"
```
```suggestion
```

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2900,7 +2905,50 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                    parameter<double>(
+                        "INCR_TOL", {.description = "increment tolerance ("
+                                                    "ratio of |increment| / |solution|)",
+                                        .default_value = 1.0e-8,
+                                        .validator = positive<double>()}),
+                    parameter<double>("MAX_EXCEEDANCE_FACT_INCR_TOL",
+                        {.description =
+                                "maximum exceedance factor for the specified increment tolerance "
+                                "(Local Newton divergence safeguard for "
+                                "continuing the simulation, if desired so by the user)",
```
rjoussen: Maybe like this? (If that's what is happening). And then maybe switch the order such that `DIVER_CONT` comes first, since the `MAX_EXCEEDANCE*` parameters conceptually depend on it. 
```suggestion
                                "continuing the simulation, if specified by the user via DIVER_CONT)",
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -49,11 +49,8 @@ namespace Mat
       overflow_error,  ///< overflow error of the term \f$ \Delta t \dot{\varepsilon}^{\text{p}} \f$
                        ///< (and \f$ \mathsymbol{E}^{\text{p}}  = \exp(- \Delta t
                        ///< \dot{\varepsilon}^{\text{p}} \mathsymbol{N}^{\text{p}}) \f$)
```
rjoussen: Kind of a minor tweak.
```suggestion
///< (and \f$ \boldsymbol{E}^{\text{p}}  = \exp(- \Delta t
///< \dot{\varepsilon}^{\text{p}} \boldsymbol{N}^{\text{p}}) \f$)
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -413,6 +414,63 @@ namespace
...
+      // DEBUG
+      std::cout << std::format("gid: {}, lid: {}, owner = {}, proc = {} \n", ele_gid,
+          structure_dis->element_row_map()->lid(ele_gid),
+          structure_dis->element_row_map()->lid(ele_gid) >= 0,
+          Core::Communication::my_mpi_rank(structure_dis->get_comm()));
+
+
+      return structure_dis->element_row_map()->lid(ele_gid) >= 0;
+    }
+  }
```
rjoussen, line 465: Do we really need this whole function? It seems to me like this is mainly a debugging functionality. Also, it looks like the `cout` titled with `// DEBUG` could leak into the standard output of "normal" simulations.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2742,6 +2854,8 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::unpack_inelastic(
     extract_from_pack(buffer, m_);
     // unpack time step quantities
     time_step_quantities_.unpack(buffer);
+    // unpack the Local Newton manager
+    local_newton_manager_.unpack(buffer);
```
rjoussen, line 2840: Does the `local_newton_manager` need to be packed and unpacked?
I think reinitializing it would be sufficient.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2900,7 +2905,50 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                            .validator = positive_or_zero<double>()}),
+                    parameter<double>(
+                        "INCR_TOL", {.description = "increment tolerance ("
+                                                    "ratio of |increment| / |solution|)",
+                                        .default_value = 1.0e-8,
+                                        .validator = positive<double>()}),
+                    parameter<double>("MAX_EXCEEDANCE_FACT_INCR_TOL",
+                        {.description =
+                                "maximum exceedance factor for the specified increment tolerance "
+                                "(Local Newton divergence safeguard for "
```
c-p-schmidt, line 2940: Just to make sure that I understood the doc correctly. Does this mean that the convergence criterion can be violated by this factor if you use the safeguard setting?

src/mat/4C_mat_inelastic_defgrad_factors.hpp
```diff
@@ -1721,6 +1754,19 @@ namespace Mat
         const Core::LinAlg::Matrix<10, 1>& x,
         InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType& err_status);
 
+    /*!
+     * @brief Performs return mapping (elastic predictor - viscoplastic plastic corrector procedure)
```
c-p-schmidt:
```suggestion
     * @brief Performs return mapping (elastic predictor - viscoplastic/plastic corrector procedure)
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -413,6 +414,63 @@ namespace
         Core::LinAlg::EigenvalInterpolationType::LOG, interp_param_list};
   }
 
+  Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LocalNewtonParams
+  retrieve_local_newton_params(const Core::Mat::PAR::Parameter::Data& matdata)
+  {
+    FOUR_C_ASSERT_ALWAYS(matdata.parameters.has_group("LOCAL_NEWTON"),
```
c-p-schmidt: Does this have an effect now? Isn't it always there now, since you always have default values?

And if not, wouldn't we rather check that before calling this method and only call this method if it is available?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -413,6 +414,63 @@ namespace
...
+        "The input parameters do not contain the required group 'LOCAL_NEWTON'!");
+
```
c-p-schmidt: I think readability of the lower part would be better if you would store `matdata.parameters.group("LOCAL_NEWTON")` to something like `const auto local_newton_params`

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -413,6 +414,63 @@ namespace
...
+    }
+    else
+    {
+      FOUR_C_ASSERT_ALWAYS(structure_dis->have_global_element(ele_gid),
+          "Inconsistency: the element with global id {} cannot be found in the structure "
+          "discretization!",
+          ele_gid);
+
+      // DEBUG
+      std::cout << std::format("gid: {}, lid: {}, owner = {}, proc = {} \n", ele_gid,
```
c-p-schmidt: Did you forget to remove this as it is marked with DEBUG?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3030,60 +3142,150 @@ Core::LinAlg::Matrix<10, 1> Mat::InelasticDefgradTransvIsotropElastViscoplast::l
...
+      temp10x1.update(1.0, sol, 0.0);
+      manage_evaluation(err_status, sol, curr_CM, eval_action);
+      switch (eval_action)
+      {
+        case (ViscoplastUtils::EvaluationAction::continue_current_iteration):
+        {
+          // continue evaluation
+          break;
+        }
+        case (ViscoplastUtils::EvaluationAction::go_to_next_iteration):
```
c-p-schmidt: Maybe?
```suggestion
        case (ViscoplastUtils::EvaluationAction::continue_with_next_iteration):
```

src/mat/4C_mat_inelastic_defgrad_factors_service.cpp
```diff
@@ -289,6 +293,74 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::TimeStepQuantities:
...
+  // resize arrays
+  curr_num_iters.resize(numgp, curr_num_iters[0]);
+
+
+  resize_called = true;
+}
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+void Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LocalNewtonManager::post_lnl(
```
c-p-schmidt: A more descriptive name, like `update_iteration_counter` or similar, would be better.
```suggestion
void Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LocalNewtonManager::update_iteration_counter(
```

src/mat/4C_mat_inelastic_defgrad_factors_service.cpp
```diff
@@ -289,6 +293,74 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::TimeStepQuantities:
...
+  // increment number of LNL iterations for the current timestep at the
+  // current GP
+  curr_num_iters[gp] += iter;
+}
+
+
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+void Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LocalNewtonManager::update()
```
c-p-schmidt: Shouldn't his method be rather called reset?
```suggestion
void Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LocalNewtonManager::reset()
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -451,6 +454,115 @@ namespace Mat
...
+                                     ///< residual and the ratio of solution increment to current
+                                     ///< solution
+    };
+
+
+    /// enum: strategy in dealing with divergence of the Local Newton Loop
+    enum class LocalNewtonDiverCont
+    {
+      stop,          ///< stop the simulation entirely
+      continue_sim,  ///<  continue the simulation, and display warning in regards to the current
```
c-p-schmidt, line 523: I think this is enough.
```suggestion
      continue,  ///<  continue the simulation, and display warning in regards to the current
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -451,6 +454,115 @@ namespace Mat
...
+
+      //! maximum exceedance factor for the solution increment tolerance (to be used when
+      //! employing the divergence management strategy for continuation with
+      //! safeguard)
+      const double max_exceedance_fact_incr_tol;
+    };
+
+    //! struct for managing the Local Newton loop, containing the utilized parameters and iteration
+    //! data
+    struct LocalNewtonManager
```
c-p-schmidt: It obviously works, but aren't you using it more like a class? So why not make it a class?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: Is it possible to break this function down a little bit more? I feel like all the new (and nice!) functionality made this function blow up quite a bit. Right now, it especially seems to be responsible for substepping/predictor logic **and** the actual newton solve and error handling. You can probably assess better if it would make sense to split this more, so I'll leave it to you.

src/mat/4C_mat_inelastic_defgrad_factors.hpp
```diff
@@ -352,7 +352,10 @@ namespace Mat
           const
       {
         return mat_behavior_;
-      }
+      };
+      //! get boolean: use substepping in the time integration scheme? (true: yes, false: no)
```
rjoussen: see my other comment.

src/mat/4C_mat_inelastic_defgrad_factors.hpp
rjoussen, line 1805: Maybe it would be worth mentioning here that this function also performs substepping (if requested)?

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
rjoussen: I noticed that there are now some parameters (definitely this one and `USE_SUBSTEPPING`) which are used exclusively in the `local_newton_loop()`, yet they are not in the `LOCAL NEWTON` input group. I am not sure whether they should be in that group or outside; if they are outside, the code should maybe distinguish more between substepping and local Newton.

tests/input_files/mat_transviso_viscoplast_refJC_log_timint.4C.yaml
```diff
@@ -55,6 +55,15 @@ MATERIALS:
...
       MAX_SUBSTEPPING_HALVE_NUM: 10
+      LOCAL_NEWTON:
+        CONV_CHECK: residual_and_increment_ratio
+        MAX_ITER: 100
+        RES_TOL: 1e-8
+        INCR_TOL: 1e-8
+        MAX_EXCEEDANCE_FACT_RES_TOL: 10
+        MAX_EXCEEDANCE_FACT_INCR_TOL: 10
+        DIVER_CONT: stop
+
```
rjoussen, line 65: I like the look of this new input group a lot!

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -20,11 +20,13 @@
 #include "4C_mat_so3_material.hpp"
 #include "4C_mat_vplast_law.hpp"
 #include "4C_mat_vplast_reform_johnsoncook.hpp"
+#include "4C_solid_3D_ele.hpp"
```
rjoussen:
```suggestion
```

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -20,11 +20,13 @@
 #include "4C_mat_so3_material.hpp"
 #include "4C_mat_vplast_law.hpp"
```
rjoussen:
```suggestion
```

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
rjoussen, line 11:
```suggestion
```

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
dragos-ana: Exactly!

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: True, it is always there, but I don't think that this additional safety layer damages the code ;) 
The second question is rather a matter of taste, I guess

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Yes, thanks for spotting this

rjoussen (commented):
Looks nice, I like all the new input functionality a lot and how the `local_newton_manager` handles things in the code.

Something I noticed: None of the tests seem to use `USE_SUBSTEPPING`, but they all define `MAX_SUBSTEPPING_HALVE_NUM`.

src/mat/4C_mat_inelastic_defgrad_factors_service.cpp
dragos-ana: Here, I tried to make it more general since there may be some additional tasks to be run after the Local Newton, e.g., things related to line search that may be implemented in the future. 

But the name `post_lnl` is cryptic, here I agree. I would rather use `update_after_local_newton`, or something like this...

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: Does not seem possible, since `continue` is a reserved keyword. I would therefore leave it as is for now. I adapted `continue_with_safeguard` to `continue_sim_with_safeguard` for consistency.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: I have adapted it to a class now

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
dragos-ana: Yes, it's generally a good idea, particularly when the Local Newton Guess Interpolation will be introduced, which can effectively restart the Local Newton with an updated estimate.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: The same happens with the global DIVER_CONT strategy: if the simulation is continued even if the global Newton has diverged, a warning is shown. I try to do the same here, that's why it is not really a debugging functionality in my opinion, since it is important for the user to know that convergence could not be achieved locally.

But this // DEBUG part is of course a problem, which I forgot to remove prior to pushing.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
dragos-ana: I don't really get this comment. What do you mean exactly?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Not really. But since it may grow with other variables when I will also implement line search algorithms, I would leave it like this for now. Your comment is, of course, valuable.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
rjoussen: Yes, it shows up weirdly, I meant that `#include "4C_io_input_parameter_container.templates.hpp"` can be removed.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
dragos-ana: I like the idea! I have integrated both substepping under the Local Newton group, since they are directly associated. They are now regarded as Local Newton params in the code.

src/mat/4C_mat_inelastic_defgrad_factors_service.cpp
c-p-schmidt: If you already know that something will come in this direction, it's fine for me to directly use a more generic name.
I just mentioned it because having too generic names always comes with the cost we see in the new solid time integrator, where we have lots of generic names that often do nothing.
Since such a name is not set in stone and can change with feature changes, I'm more a fan of naming it based on the current functionality (if it is not already clear that it will change in the very near future).

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: That makes sense, and there should definitely be a warning in that case. What I am mainly wondering about is: does this function return false ever in normal runs (not unittests)? I would expect that it can only be called by ranks that own the current `ele_gid`.

dragos-ana: I verified this prior to the warning implementation - the method is also called for ghosted elements.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
dragos-ana: Removing it leads to linkage errors, because it contains the `add` template method for more complex parameter types, such as the many enum classes I use.

tests/list_of_tests.cmake
```diff
@@ -861,7 +861,7 @@ four_c_test(TEST_FILE mat_gtn_solid_necking_bar_with_f0_fc_h8.4C.yaml NP 2 RETUR
 four_c_test_restart(BASED_ON ${current} SAME_FILE NP 2 RESTART_STEP 2)
 four_c_test(TEST_FILE mat_holzapfelcardio.4C.yaml NP 2)
 four_c_test(TEST_FILE mat_humphreycardio.4C.yaml NP 2)
-four_c_test(TEST_FILE mat_iso_viscoplast_refJC_log_substep.4C.yaml RETURN_AS current)
+four_c_test(TEST_FILE mat_iso_viscoplast_refJC_log_timint.4C.yaml RETURN_AS current)
```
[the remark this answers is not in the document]
dragos-ana: The files have been renamed, the comment makes no sense

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -592,7 +642,10 @@ Mat::PAR::InelasticDefgradTransvIsotropElastViscoplast::
       max_plastic_strain_incr_(matdata.parameters.get<double>("MAX_PLASTIC_STRAIN_INCR")),
       max_plastic_strain_deriv_incr_(
           matdata.parameters.get<double>("MAX_PLASTIC_STRAIN_DERIV_INCR")),
-      max_substepping_halve_num_(matdata.parameters.get<int>("MAX_SUBSTEPPING_HALVE_NUM")),
+      use_local_substepping_(
+          matdata.parameters.group("LOCAL_SUBSTEPPING").get<bool>("USE_SUBSTEPPING")),
+      max_local_substepping_halve_num_(static_cast<unsigned int>(
+          matdata.parameters.group("LOCAL_SUBSTEPPING").get<int>("MAX_SUBSTEPPING_HALVE_NUM"))),
```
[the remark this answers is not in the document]
dragos-ana: This defeats the purpose of the input validators; I will leave it as it is for now.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: I had planned this for a future PR, but since this PR has already got out of proportion, I did it here in a new commit ;)

I would squash the commits after the PR is approved.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2600,16 +2664,41 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_inverse_inelast
...
+  }
+}
 
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+Core::LinAlg::Matrix<3, 3> Mat::InelasticDefgradTransvIsotropElastViscoplast::return_mapping(
+    const Core::LinAlg::Matrix<3, 3>& FredM)
+{
+  // declare output: inverse inelastic deformation gradient (after return mapping)
+  Core::LinAlg::Matrix<3, 3> iFinM{Core::LinAlg::Initialization::zero};
```
rjoussen, line 2681:
```suggestion
  // declare output: inverse inelastic deformation gradient and plastic strain (after return mapping)
  Core::LinAlg::Matrix<3, 3> iFinM{Core::LinAlg::Initialization::zero};
  double plastic strain = 0.0;
```
I would initialize plastic strain here as well, for two reasons:
1. Handle it in parallel with `iFinM`, they kinda belong together anyway.
2. This later enables setting `plastic_strain = plastic_strain_pred;` if the prediction is the solution and `plastic_strain = sol(9)` in the corrector branch. Consequently, the lengthy `update_hist_var_` block could move after the if/else and would be needed only once.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
-      jacMat = calculate_jacobian(curr_CM, sol,
-          time_step_quantities_.last_substep_plastic_defgrad_inverse[gp_],
-          time_step_quantities_.last_substep_plastic_strain[gp_], local_substepping_utils_.curr_dt,
-          err_status);
-      if (err_status != ViscoplastUtils::ErrorType::no_errors)
+    // update solution vector and relative increment
+    sol.update(1.0, dx, 1.0);
+    const double sol_norm = sol.norm2();
+    const double dx_norm = dx.norm2();
+    FOUR_C_ASSERT_ALWAYS(sol_norm >= 1.0e-8,
```
c-p-schmidt, line 3396: Do you really want to hard-code the tolerance here?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+}
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+bool Mat::InelasticDefgradTransvIsotropElastViscoplast::is_local_newton_stuck(
+    const ViscoplastUtils::LocalNewtonConvQuantities& conv_quantities)
+{
+  // check for "stuck" Local Newton, i.e., the increment does not change much but there is not a
+  // converged state (check only feasible after the first iteration, since dx must be available)
+  if ((local_newton_manager_.iter() > 1) && (conv_quantities.increment_norm < 1.0e-15))
```
c-p-schmidt, line 3436: Also here

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3304,6 +3720,45 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
 }
 
 
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+void Mat::InelasticDefgradTransvIsotropElastViscoplast::manage_evaluation(
```
c-p-schmidt, line 3714: Aren't you rather doing some error handling in here?
```suggestion
void Mat::InelasticDefgradTransvIsotropElastViscoplast::error_handling(
```

And you do not use `curr_CM` and `sol` in this method, so these arguments are unused. Please remove them.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -268,29 +275,70 @@ namespace Mat
...
+
+      //! halve current substep and update relevant quantities
+      void halve_substep();
+
+      //! get substep size
+      [[nodiscard]] double get_substep_size() const { return curr_dt_; }
+
+      //! retrieve the normalized time parameter computed currently
+      //! (1.0 corresponds to the full problem time step dt)
+      [[nodiscard]] double get_normalized_tnp(const double dt) const
```
c-p-schmidt:
```suggestion
      [[nodiscard]] double get_normalized_time_param(const double dt) const
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -451,6 +499,146 @@ namespace Mat
...
+    class LocalNewtonManager
+    {
+     public:
+      /*!
+       * @brief Constructor
+       *
+       * @param[in] lnl_params Local Newton parameters
+       *
+       */
+      LocalNewtonManager(const LocalNewtonParams& lnl_params);
```
c-p-schmidt: One argument constructors should be marked explicit.
```suggestion
      explicit LocalNewtonManager(const LocalNewtonParams& lnl_params);
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
-  // declare error status for tensor interpolation
-  Core::LinAlg::TensorInterpolationErrorType tensor_interpolator_err_status{
-      Core::LinAlg::TensorInterpolationErrorType::NoErrors};
+  // local substepping procedures
+  if (parameter()->use_local_substepping())
+  {
+    while (!local_substepping_utils_.end_substepping())
+    {
+      // interpolate right Cauchy-Green tensor if we use local substepping
+      curr_CM = tensor_interpolator_.get_interpolated_matrix(ref_matrices_, ref_locs_,
```
rjoussen:
```suggestion
      curr_CM = tensor_interpolator_.get_interpolated_matrix({time_step_quantities_.last_rightCG[gp_], CM}, {0.0, 1.0},
```
this still works without any copying due to const reference. This makes the `ref_matrices_` and `ref_locs_` member variables obsolete.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+
+        // update the values of history variables at the last converged state (if we have
+        // not reached the end of substepping yet)
+        if (!local_substepping_utils_.end_substepping())
+        {
+          time_step_quantities_.last_substep_plastic_defgrad_inverse[gp_] =
+              extract_inverse_inelastic_defgrad(sol);
+          time_step_quantities_.last_substep_plastic_strain[gp_] = sol(9);
+          // update last substep history variables of the viscoplastic flow rule
+          viscoplastic_law_->update_gp_state(gp_);
```
rjoussen: Will this do anything in the future? Because right now it does nothing.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+      // update substep
+      if (err_status == InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType::no_errors)
+      {
+        // this means the current substep has converged: we increment the substep and update the
+        // values relevant for substepping, including history data
+        local_substepping_utils_.increment_substep();
+
+        // update the values of history variables at the last converged state (if we have
+        // not reached the end of substepping yet)
+        if (!local_substepping_utils_.end_substepping())
```
rjoussen: Is this if statement necessary? `last_substep_plastic_defgrad_inverse` is always the right value even if substepping will end. I also dont see who it would be used afterwards still.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
 
 /*--------------------------------------------------------------------*
  *--------------------------------------------------------------------*/
-Core::LinAlg::Matrix<10, 1> Mat::InelasticDefgradTransvIsotropElastViscoplast::local_newton_loop(
+Core::LinAlg::Matrix<10, 1>
+Mat::InelasticDefgradTransvIsotropElastViscoplast::viscoplastic_correction(
```
rjoussen, line 3087: I like this new/refactored function a lot now!

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
       }
+      case (ViscoplastUtils::EvaluationAction::exit_with_error):
+      {
+        // exit with the set error status
+        return;
+      }
+      default:
+      {
+        FOUR_C_THROW("Invalid evaluation action {} for error status {} after residual evaluation",
+            EnumTools::enum_name(eval_action), EnumTools::enum_name(err_status));
```
rjoussen, line 3247: I dont think you need `enum_name` in `FOUR_C_THROW`. It should print the name directly. Also applies to a lot of other places.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: I could theoretically introduce a variable similar to `zero_plastic_strain_increment` in `InelasticDefgradTransvIsotropElastViscoplastUtils`. However, the zero plastic strain increment is something that appears in multiple places, and that's why it is saved as a single variable. 

In contrast, the variable / safeguard here appears only in this single function. I would therefore neither introduce an input parameter for it, nor a dedicated const variable in the utilities namespace as for the zero plastic strain increment. 
We can discuss in private if you have particular suggestions for improvement - but to me it seems, that there are no clear guidelines in 4C for such variables that appear in single checks. I could at most name it and declare it somewhere in the beginning of the function, but in my opinion it does not improve the situation considerably. I am however open to other suggestions that I may have missed.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Same as above.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: It depends on how you look at it. The function is called regardless of whether there is an error or not; that's why it is more evaluation management than error handling, in my opinion.

But I fully agree on the arguments, thanks!

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: True, thanks for spotting it!

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: I adapted it to `get_normalized_next_time_param`, and also provided a more detailed documentation of what is retrieved.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -78,6 +76,15 @@ namespace Mat
     };
 
 
+    /// enum class for evaluation management actions in the iterations of the
+    /// Local Newton loop
+    enum class EvaluationAction
+    {
+      continue_current_iteration,    ///< continue current iteration
+      continue_with_next_iteration,  ///< go to next iteration after performing certain reset steps
```
rjoussen, line 84: `continue_with_next_iteration` is not used anywhere (i.e., no variable is ever assigned this value). Is this a planned feature?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
c-p-schmidt: You are the expert on using these kinds of materials. I just thought from a user perspective it might be cumbersome not being able to set such a value from input.

But this might not be the case.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
-          time_step_quantities_.last_substep_plastic_defgrad_inverse[gp_],
-          time_step_quantities_.last_substep_plastic_strain[gp_], local_substepping_utils_.curr_dt,
-          err_status);
-      if (err_status != ViscoplastUtils::ErrorType::no_errors)
+      case (ViscoplastUtils::EvaluationAction::continue_with_next_iteration):
       {
-        new_substep_status = prepare_new_substep(sol, curr_CM);
-        if (!new_substep_status) return sol;  // return with error
+        // recompute dx after conducting adjustments to solution vector
+        dx.update(1.0, sol, -1.0, temp10x1, 0.0);
```
rjoussen, line 3233: Due to line 3230, `temp10x1` is `sol`, so this is essentially setting `dx` to 0?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+    const bool is_converged = is_local_newton_converged(conv_quantities);
+
+    if (is_converged)
+    {
+      return;
+    }
+    else
+    {
+      // check if maximum iteration is exceeded
+      if (local_newton_manager_.iter() > local_newton_manager_.params().max_iter)
```
rjoussen:
```suggestion
    else if (local_newton_manager_.iter() > local_newton_manager_.params().max_iter)
      // check and return if maximum iteration is exceeded
```
or maybe just `if` since we return anyway before.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+            {
+              FOUR_C_THROW(
+                  "Invalid evaluation action {} for error status {} after verification of stuck "
+                  "Local Newton",
+                  EnumTools::enum_name(eval_action), EnumTools::enum_name(err_status));
+            }
+          }
         }
+      }
+    }
```
rjoussen, line 3325:
```suggestion
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+          verify_local_newton_exit(conv_quantities, err_status);
+          return;
+        }
+      }
+      else
+      {
+        // check whether the Local Newton is 'stuck'
+        const bool is_stuck = is_local_newton_stuck(conv_quantities);
+        if (is_stuck)
+        {
```
rjoussen, line 3296:
```suggestion
      else if (is_local_newton_stuck(conv_quantities))
      {
```
Or maybe just if since the iterations exceeded branch returns anyway

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+          {
+            case (ViscoplastUtils::EvaluationAction::continue_current_iteration):
+            {
+              // continue evaluation
+              break;
+            }
+            case (ViscoplastUtils::EvaluationAction::continue_with_next_iteration):
+            {
+              // recompute dx after conducting adjustments to solution vector
+              dx.update(1.0, sol, -1.0, temp10x1, 0.0);
```
rjoussen: same as https://github.com/4C-multiphysics/4C/pull/1956/changes#r3100709917

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+    {
+      case (ViscoplastUtils::EvaluationAction::continue_current_iteration):
+      {
+        // continue evaluation
         break;
       }
+      case (ViscoplastUtils::EvaluationAction::continue_with_next_iteration):
+      {
+        // proceed with next iteration after performing adjustments due
+        // to errors
```
rjoussen, line 3342: Is this happening here? To me, it seems like all the `continue_with_next_iteration` cannot be reached at the moment anyway, so it is hard to assess.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
-          err_status);
-      if (err_status != ViscoplastUtils::ErrorType::no_errors)
+    // update solution vector and relative increment
+    sol.update(1.0, dx, 1.0);
+    const double sol_norm = sol.norm2();
+    const double dx_norm = dx.norm2();
+    FOUR_C_ASSERT_ALWAYS(sol_norm >= 1.0e-8,
+        "The solution vector in local iteration {} is nearly 0, with 2-norm: {}! Something went "
+        "wrong, since such mechanical states are not expected!",
+        local_newton_manager_.iter(), sol_norm);
```
rjoussen, line 3399: Does that happens sometimes? It feels strange to do a physical/mechanically motivated assert so deep in the local newton loop. I guess it asserts that `iFinM` is not around 0, which is indeed unphysical (even though not impossible!), but here the computed `sol` is definitely the correct solution of the local linear system, so its not "wrong" in that sense.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+    conv_quantities.increment_norm = dx_norm / sol_norm;
+  }
+}
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+bool Mat::InelasticDefgradTransvIsotropElastViscoplast::is_local_newton_converged(
```
rjoussen, line 3406: I think this should be a member function of the `LocalNewtonManager`.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+              conv_quantities.increment_norm <= local_newton_manager_.params().incr_tol);
+      break;
+    default:
+      FOUR_C_THROW("You should not be here (convergence checking of the Local Newton Loop)");
+  }
+}
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+bool Mat::InelasticDefgradTransvIsotropElastViscoplast::is_local_newton_stuck(
```
rjoussen, line 3431: I think this should be a member function of the `LocalNewtonManager.`

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
-
-      // solve loop equation
-      dx.clear();                                      // reset
-      solver_10_10_1.set_matrix(jacMat);               // set A=jacMat
-      solver_10_10_1.set_vectors(dx, residual);        // set dx=increment, residual=RHS
-      solver_10_10_1.factor_with_equilibration(true);  // "some easy type of preconditioning"
-      int err2 = solver_10_10_1.factor();              // factoring
-      int err = solver_10_10_1.solve();                // X = A^-1 B
-      if ((err != 0) || (err2 != 0))
+  return false;
```
rjoussen, line 3460: I think it is easier if you first do the `conv_check` switch, an then assemble all requirements into the return. But that's somewhat personal taste.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2986,219 +3086,550 @@ Core::LinAlg::Matrix<10, 10> Mat::InelasticDefgradTransvIsotropElastViscoplast::
...
+}
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+void Mat::InelasticDefgradTransvIsotropElastViscoplast::verify_local_newton_exit(
```
rjoussen, line 3465: This function also feels like it would like to be in `LocalNewtonManager`, but I guess here it is hard because of the error message?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen, line 3714: Will this function have more functionality in the future? Considering that where ever it is used at the moment, it is just followed by another switch statement, it feels like the error management could also happen directly there.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3304,6 +3720,45 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
 }
 
 
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+void Mat::InelasticDefgradTransvIsotropElastViscoplast::manage_evaluation(
+    const Core::LinAlg::Matrix<3, 3>& curr_CM,
+    const InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType& err_status,
+    Core::LinAlg::Matrix<10, 1>& sol,
```
rjoussen:
```suggestion
    const Core::LinAlg::Matrix<10, 1>& sol,
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3304,6 +3720,45 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+    InelasticDefgradTransvIsotropElastViscoplastUtils::EvaluationAction& eval_action)
```
rjoussen:
```suggestion
    InelasticDefgradTransvIsotropElastViscoplastUtils::EvaluationAction& eval_action) const
```

rjoussen (commented):
I really like the control flow through return_mapping -> corrector -> local_newton now!Its very easy to follow now in my opinion.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: This is a left-over from a time where I tried different linearizations in case of substepping. I am not sure what the exact motivation was. But I would drop it for now as proposed.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: I will leave it anyway since I know exactly what gets into the error message. But thanks for the note!

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: Planned feature for the upcoming PR. Sorry for already introducing this here.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: It will not be. In the case of continue_with_next_iteration, sol will be updated and will not be the same as temp10x1.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Here I think it is a matter of taste, in contrast to the previous case where the return was immediate. I will leave it like this for more clarity.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Yes, my bad, because I introduced this enum already in this PR. This will become clearer in the next PR.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: If that happens (which of course it should not in normal mechanical scenarios), then a division by 0 follows directly. At least we detect that the mechanical state has become extremely weird with this assert.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: I understand the motivation, but I don't want to nest too much information into the LocalNewtonManager. I currently see it rather as a bookkeeper of iterations, rather than a standalone Local Newton performer.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Same as above.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Yes, this is rather subjective, but thanks for the note.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: This has been removed entirely. For now, because it will be introduced again to enable `continue_with_next_iteration`.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Same motivation as above. The Local Newton manager is only a bookkeeper, currently. I try do the as much computation as I can directly inside the material to keep it clear.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Yes, of course, see my comments above on `continue_with_next_iteration`. 

It is definitely my own doing that this PR has become so extreme, because I should have introduced less preparation functionalities here.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: It does something for the Anand law, which will be added in the future.

rjoussen (approved):

dragos-ana:
@rjoussen @c-p-schmidt I squashed the commits now.

c-p-schmidt (approved):

PR #1958, opened 2026-04-09
refactor: rename cmake test functions for better overview

davidrudlstorfer:
## Description and Context

renames the cmake test functions with two underscores at the beginning. This creates a much better overview within our tests. Follow up of #1952 

With two underscores we also differentiate between internal cmake functions and the testing functions we need.

Originally discussed here https://github.com/4C-multiphysics/4C/pull/1448#discussion_r2478110945

Part of #56


amgebauer (dismissed):

c-p-schmidt (dismissed):
Sounds reasonable to me.

amgebauer (approved):

c-p-schmidt (approved):

PR #1959, opened 2026-04-10
Fix bugs in old w1-poro-p1 element

amgebauer:
This should fix a few bugs in the w1-poro-p1 element. I found those as I wanted to migrate all solid-poro tests to the new solid-poro framework.

The bugs are:

* Thickness is missing during integration
* Removed one extra J in the linearization of solid-poro-brinkman-formulation
* And the same bug as in https://github.com/4C-multiphysics/4C/pull/1924 (we multiplied two times with the integration factor)


lkoeglmeier (approved):
Thanks a lot for looking into this and fixing it :pray:

c-p-schmidt (approved):

PR #1960, opened 2026-04-10
[ReducedLung] Implement model registry and factory for airway models

bgoderbauer:
## Description and Context

Refactor airway model creation to use a registry + factory pattern, aligned with the terminal-unit approach introduced in #1929.  
Flow-resistance and wall-mechanics models are now instantiated via type-based dispatch, and compile-time state-count traits ensure each model combination initializes the correct number of equations/DOFs.  
This is a structural refactor only. The model equations and the intended numerical behavior remain unchanged.

#1929


maxiludwig (approved):
Great! Thanks for taking care of this 👏

davidrudlstorfer (approved):

PR #1968, opened 2026-04-14
Move all constr2D tests to new solid elements

amgebauer:
This PR moves all constr2D tests tot he new solid elements.

I had to slightly adapt a few results since the SOLID-elements use a different formulation for EAS than the old Wall elements.


src/solid_3D_ele/4C_solid_3D_ele_line.cpp
```diff
@@ -107,6 +138,85 @@ void Discret::Elements::SolidLine<dim>::print(std::ostream& os) const
...
+
+
+      // We are not interested in volume of ghosted elements
+      if (Core::Communication::my_mpi_rank(discretization.get_comm()) != owner()) return 0;
+
+      auto current_nodal_coordinates =
+          evaluate_current_coordinates<celltype, dim>(*this, discretization, lm);
+
+      elevector3[0] = 0.5 * (current_nodal_coordinates[0](1) + current_nodal_coordinates[1](1)) *
+                      (current_nodal_coordinates[1](0) - current_nodal_coordinates[0](0));
```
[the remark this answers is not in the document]
amgebauer: This is how it is implemented in the old wall elements.

src/core/legacy_enum_definitions/4C_legacy_enum_definitions_element_actions.hpp
```diff
@@ -41,6 +41,8 @@ namespace Core::Elements
     struct_calc_reset_istep,  //!< reset elementwise internal variables, during iteration to last
                               //!< converged state
     struct_calc_energy,       //!< compute internal energy
+    calc_struct_constrarea,
```
c-p-schmidt, line 44: Maybe `calc_struct_area_constr` and `calc_struct_area_constr_stiff` for better readability?

c-p-schmidt (approved):

src/core/legacy_enum_definitions/4C_legacy_enum_definitions_element_actions.hpp
amgebauer: I would leave it like so for now since it is also used in a couple of other places. In a cleanup, one could probably move this out of the element completely and remove the action completely.

davidrudlstorfer (approved):

PR #1971, opened 2026-04-14
Input: support name-based element block definition

rjoussen:
## Description and Context

This adds support for defining element blocks by the new input parameter `NAME` instead of `ID` when reading external meshes. For name-based resolution to work, the name must occur only once in the external mesh file. This looks as follows in the input:

```yaml
STRUCTURE GEOMETRY:
  SHOW_INFO: "detailed_summary"
  ELEMENT_BLOCKS:
    - NAME: FirstBox
      SOLID:
        HEX8:
          MAT: 1
          KINEM: nonlinear
``` 

For this, the `read_external_mesh` function is refactored: it loops over the element block definition in the input instead of the cell blocks in the mesh file now, which made it easier to allow for both name- and id-based input. In addition, this allowed to introduce more strict error handling:
   - An error is thrown if an input element block is not present in the external mesh.
   - An error is thrown when an element block is referenced more than once in the input. This was not the case so far (hence breaking)

I added a new integration test with gmsh and modified the 2D fsi tutorial to test the exodus case.

Note: this should also work for vtu input if vtu input interprets integer cell arrays as element block by default (similar to #1918).

Part of #825

### DIFF
--- a/src/core/io/src/4C_io_meshreader.cpp
+++ b/src/core/io/src/4C_io_meshreader.cpp
@@ -25,12 +25,15 @@
 #include "4C_rebalance.hpp"
 #include "4C_rebalance_graph_based.hpp"
 #include "4C_rebalance_print.hpp"
+#include "4C_utils_exceptions.hpp"
 
 #include <Teuchos_TimeMonitor.hpp>
 
+#include <iostream>
 #include <set>
 #include <string>
 #include <utility>
+#include <vector>
 
 FOUR_C_NAMESPACE_OPEN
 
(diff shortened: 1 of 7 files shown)


src/global_legacy_module/4C_global_legacy_module_validparameters.cpp
```diff
@@ -180,17 +180,24 @@ namespace CommonParameters
...
         list("ELEMENT_BLOCKS",
             all_of({
-                parameter<int>(
-                    "ID", {.description = "ID of the element block in the exodus file."}),
+                parameter<std::optional<int>>(
+                    "ID", {.description = "ID of the element block in the external mesh file. Only "
+                                          "allowed if NAME is not specified."}),
+                parameter<std::optional<std::string>>(
+                    "NAME", {.description = "Name of the element block in the external mesh file. "
+                                            "Only allowed if ID is not specified."}),
```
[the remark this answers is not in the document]
rjoussen: `one_of` doesnt work here because there are already parallel `one_of`s in the element definitions.

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -564,87 +565,153 @@ namespace
...
+    FOUR_C_ASSERT_ALWAYS(!relevant_blocks.empty(),
+        "None of the element blocks specified in the input file could be found in the external "
+        "mesh. Please check your input and mesh files.");
+
+    mesh_reader.filtered_mesh_on_rank_zero.emplace(
+        mesh.filter_by_cell_block_ids({relevant_blocks.begin(), relevant_blocks.end()}));
+
+    // Rank zero provides the actual data.
+    mesh_reader.target_discretization.fill_from_mesh(
+        *mesh_reader.filtered_mesh_on_rank_zero, user_elements, {}, parameters);
```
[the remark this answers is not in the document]
rjoussen: This actually does correctly deduce to call the range constructor, but I updated it anyway.

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -564,87 +567,153 @@ namespace
...
-          ele_count++;
-          cell_id_in_block++;
-        }
+      // No matching block found in the external mesh for the given ID/NAME in the input
+      if (not found)
+      {
+        std::cout << "Warning: Element block with " << block_label
+                  << " matches no definition in the external mesh file and will be ignored.\n";
+        continue;  // to next element block definition in input file
       }
```
amgebauer: Why ignoring it? We should throw an error if we specify a block that does not exist in the input file.

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -564,87 +567,153 @@ namespace
...
+      const auto input_id = current_element_block_input.get<std::optional<int>>("ID");
+      const auto input_name = current_element_block_input.get<std::optional<std::string>>("NAME");
 
-        FOUR_C_ASSERT_ALWAYS(cell_type == cell_block.cell_type(),
-            "Element block '{}' has cell type '{}' but your given element definition for '{}' has "
-            "cell type '{}'.",
-            eb_id, cell_block.cell_type(), element_name, cell_type);
+      // Resolve the block ID from either the given ID or NAME
+      const auto& [found, resolved_eb_id, block_label] =
+          find_element_block(input_id, input_name, mesh_block_ids, mesh_block_ids_by_name);
```
amgebauer: Can we make it explicit here instead of implicit in find_element_block, i.e., having a function that obtains the element block by an id and another one that finds it by a name.

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -564,87 +567,153 @@ namespace
...
   }
 
+  struct FoundElementBlock
+  {
+    bool found;
+    int id;
+    std::string label;
+  };
+
+  FoundElementBlock find_element_block(const std::optional<int>& input_id,
```
amgebauer: Can't we just return the reference to the respective block? (and throw if it was not found)

amgebauer (commented):
Thank you @rjoussen for making the mesh input less error prone for users!

src/core/io/src/4C_io_meshreader.cpp
rjoussen: My idea was that it could be a convenient way to use the same input file across different mesh files, e.g., when some mesh files contain one or two bodies. But the better way to do this is probably to use the same mesh file in both cases and just omit one of the blocks. I'll change it and update the PR description, since this is another breaking change.

src/core/io/src/4C_io_meshreader.cpp
rjoussen: You mean one function `find_element_block_by_name` and another one `find_element_block_by_id`?

src/core/io/src/4C_io_meshreader.cpp
rjoussen: If we throw in cases when the input element block was not found (see your other comment), this should be possible, yes.

ischeider:
You gave it the label "breaking change", but in fact all old input files work as before, right? IMO it's just an extension, not breaking anything, but maybe I don't see the break.

src/core/io/src/4C_io_meshreader.cpp
amgebauer: Yes, I would find that cleaner

rjoussen:
@ischeider There are corner cases in which old input files don't work as before anymore, e.g. when a user defined an element block twice, silently only the first one got picked up. Additionally, so far, it was possible to define an element block that was not present in the mesh file, which now throws an error. I guess neither of these was a feature to begin with, so feel free to remove the label.

src/core/io/src/4C_io_meshreader.cpp
rjoussen: Done

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -564,87 +567,150 @@ namespace
...
+      const auto input_name = current_element_block_input.get<std::optional<std::string>>("NAME");
+
+      FOUR_C_ASSERT_ALWAYS((input_id.has_value() != input_name.has_value()),
+          "Element block definition must specify exactly one of ID or NAME but got {}.",
+          input_id.has_value()
+              ? "ID: " + std::to_string(*input_id) + " and NAME: '" + *input_name + "'"
+              : "neither");
+
+      std::string block_label;
+      const auto& cell_block = [&]() -> const Core::IO::MeshInput::CellBlockReference<3, true>&
```
rjoussen, line 655: The reason why this is a lambda is that `CellBlockReference` is not default constructible, and otherwise I would need to use an initially null pointer.

rjoussen:
I updated the following:
- First Commit moves all `MeshInput` references out of the internal namespace, so its less awkward to use outside.
- The second commit is essentially the same as my initial version with the unified `find_element_block` (I just wanted to keep it around for now for possible revisions)
- The third commit uses seperate name/ID-based find_element_block functions and makes the look-up tables more symmetric.

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -564,87 +567,150 @@ namespace
     }
   }
 
+  const auto& find_element_block_by_id(const int id,
+      const std::unordered_map<Core::IO::MeshInput::ExternalIdType,
+          Core::IO::MeshInput::CellBlockReference<3, true>>& mesh_blocks_by_id,
+      std::string& block_label)
```
amgebauer: Filling the block_label within this function is weird. I would move it to the place where you actually query the element block (same for the case below)

src/core/io/src/4C_io_meshreader.cpp
amgebauer: That is perfectly fine here (and much better than first default constructing the type) :+1:

src/core/io/src/4C_io_meshreader.cpp
rjoussen: I moved it out of the functions. It's now only constructed in the actual error messages.

amgebauer (approved):

rjoussen:
Could someone else have a look, please? The large diff is mainly due to moving the references out of the internal namespace.

bgoderbauer (approved):

PR #1972, opened 2026-04-14
[beaminteraction] Improve error message for beam-to-solid

isteinbrecher:
Improve error messages for beam-to-solid (geometry) segmentation errors.

### DIFF
--- a/src/geometry_pair/4C_geometry_pair_utility_classes.hpp
+++ b/src/geometry_pair/4C_geometry_pair_utility_classes.hpp
@@ -402,7 +402,14 @@ namespace GeometryPair
         // The segments are equal.
       }
       else
-        FOUR_C_THROW("The two segments are overlapping. This is fatal!");
+      {
+        FOUR_C_THROW(
+            "The two segments are overlapping. This is fatal! lhs: [{}, {}], rhs: [{}, {}]",
+            Core::FADUtils::cast_to_double(lhs.get_eta_a()),
+            Core::FADUtils::cast_to_double(lhs.get_eta_b()),
+            Core::FADUtils::cast_to_double(rhs.get_eta_a()),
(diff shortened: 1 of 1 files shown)


davidrudlstorfer (approved):

dharinib98 (approved):

src/geometry_pair/4C_geometry_pair_utility_classes.hpp
```diff
@@ -402,7 +402,14 @@ namespace GeometryPair
         // The segments are equal.
       }
       else
-        FOUR_C_THROW("The two segments are overlapping. This is fatal!");
+      {
+        FOUR_C_THROW(
+            "The two segments are overlapping. This is fatal! lhs: [{}, {}], rhs: [{}, {}]",
```
maxfirmbach, line 407: Is `lhs` and `rhs` the correct wording in this context? I have the linear algebra bias here, but these words seem weird to me.

src/geometry_pair/4C_geometry_pair_utility_classes.hpp
```diff
@@ -427,7 +434,14 @@ namespace GeometryPair
...
```
maxfirmbach, line 439: Same here.

maxfirmbach (commented):
LGTM. I just have one comment.

src/geometry_pair/4C_geometry_pair_utility_classes.hpp
isteinbrecher: I think the original idea was that `lhs` and `rhs` correspond to the left and right hand side of the `<` operator, so I would keep this unless we can come up with a much better naming.

maxfirmbach (approved):

PR #1979, opened 2026-04-16
Allow to export timings as yaml

amgebauer:
This PR allows to export the time monitor values as a yaml format.

The output looks like this (shortened):

```yaml
# Teuchos::TimeMonitor report
---
Output mode: spacious
Number of processes: 2
Time unit: s
Statistics collected: 
  - MinOverProcs
  - MeanOverProcs
  - MaxOverProcs
  - MeanOverCallCounts
Timer names: 
  - Calculation
  - "Core::FE::Discretization::Evaluate"
Total times: 
  Calculation: 
    MinOverProcs: 0.0478823
    MeanOverProcs: 0.0478858
    MaxOverProcs: 0.0478894
    MeanOverCallCounts: 0.0478858
  "Core::FE::Discretization::Evaluate": 
    MinOverProcs: 0.0130819
    MeanOverProcs: 0.013161
    MaxOverProcs: 0.0132401
    MeanOverCallCounts: 0.000387088
Call counts:
  Calculation: 
    MinOverProcs: 1
    MeanOverProcs: 1
    MaxOverProcs: 1
    MeanOverCallCounts: 1
  "Core::FE::Discretization::Evaluate": 
    MinOverProcs: 34
    MeanOverProcs: 34
    MaxOverProcs: 34
    MeanOverCallCounts: 34
```
We can read this afterwards, extract and report the timings in a generated html report.


davidrudlstorfer (dismissed):
Would it make sense to use an input parameter from the input file for that rather than a CLI arg?

Maybe some variable in the global IO section?

amgebauer:
> Would it make sense to use an input parameter from the input file for that rather than a CLI arg?

I'm also fine with that. I personally would enable it by default. Any other opinions about that?

rjoussen:
Looks really nice!

> Would it make sense to use an input parameter from the input file for that rather than a CLI arg?

I would prefer that as well. 

> I personally would enable it by default.

I would disable it, to be honest, it creates an extra file which most users will very rarely look at.

apps/global_full/4C_global_full_main.cpp
```diff
@@ -245,6 +247,14 @@ void run(CommandlineArguments& cli_args, Core::Communication::Communicators& com
   }
 
   write_timemonitor(communicators.local_comm());
```
rjoussen, line 249: Should timing information still be written to standard output if a more readable format is available?

amgebauer: yes, I would keep it since you directly see a summary of the timings in the terminal when running simulations manually.

amgebauer:
> I would disable it, to be honest, it creates an extra file which most users will very rarely look at.

In my eyes, the timings provide valuable information that is worth to look at in many situations. That's why, I would write it by default. But I'm open to other opinions here.

davidrudlstorfer (approved):

src/inpar/4C_inpar_io.cpp
```diff
@@ -28,6 +28,12 @@ std::vector<Core::IO::InputSpec> Inpar::IO::valid_parameters()
               {.description = "Flag indicating whether quantities related to the spring dashpot "
                               "conditions should be written to output.",
                   .default_value = false}),
+          parameter<bool>("WRITE_TIMINGS",
+              {.description =
+                      "Flag indicating whether timings of the time monitor should be written to a "
+                      "yaml file. The The timing information is written to "
+                      "<outputprefix>-timings.yaml.",
```
ppraegla, line 35:
```suggestion
              {.description =
                      "Flag indicating whether timings of the time monitor should be written to a "
                      "yaml file. The timing information is written to "
                      "<outputprefix>-timings.yaml.",
```

ppraegla (changes requested):

apps/global_full/4C_global_full_main.cpp
```diff
@@ -245,6 +247,14 @@ void run(CommandlineArguments& cli_args, Core::Communication::Communicators& com
...
+
+  if (Global::Problem::instance()->io_params().get<bool>("WRITE_TIMINGS"))
+  {
+    std::filesystem::path filename(
+        Global::Problem::instance()->output_control_file()->file_name() + "-timings.yaml");
```
isteinbrecher, line 254: Now the naming reflects only that computational times are measured. Should we change the names to something like "performance_stats", so we can also output memory usage to that file in the future?

isteinbrecher (commented):
Thanks for this nice extension!

> In my eyes, the timings provide valuable information that is worth to look at in many situations. That's why, I would write it by default. But I'm open to other opinions here.

I agree, especially if one does not want to rerun the simulation later on.

I suspect that this feature will be used by many users, can we somehow test this? Obviously it makes no sense to test the actual timings, but we cold test the structure of the file by simply allowing a very high `float` tolerance.

rjoussen:
> I agree, especially if one does not want to rerun the simulation later on.

Okay that is actually quite convincing and speaks for the default 👍

c-p-schmidt (approved):
I'm also in favor of activating it by default.

apps/global_full/4C_global_full_main.cpp
amgebauer: I think we can rename it later once we also measure memory usage. This would require to change the structure of the file and is anyway not straight forward. We are currently use the Teuchos functions for the timers, which does not allow to access the timers in C++, so we cannot extract the timers and write it into a file with our own structure.

davidrudlstorfer (approved):

ppraegla (approved):

rjoussen (approved):

PR #1980, opened 2026-04-16
Node numbering in node set printout when external mesh files are read

ischeider:
# Description

If an external mesh (exodus, gmesh, ...) is read in, one can print out very detailed mesh information using
```
STRUCTURE GEOMETRY:
  FILE: <external_mesh_file>
  SHOW_INFO: full
```
In the case of `full` information, the node numbers of the node sets did not correlate to the external ids (which are for example relevant for `RESULT DESCRIPTION`), but were the internal numbers.

This fix now prints the external node ids also for the node sets (for the printout of the nodes itself it was correct before already).

BTW: It is always checked whether an external ID for a node exist (could also be `invalid_external_id`), but I don't know when this happens and whether these nodes without external ID can be part of a node set. Maybe someone can explain that to me.

### DIFF
--- a/src/core/io/src/4C_io_mesh.cpp
+++ b/src/core/io/src/4C_io_mesh.cpp
@@ -236,23 +236,29 @@ void Core::IO::MeshInput::print(const Mesh<dim>& mesh, std::ostream& os, Verbosi
   {
     if (verbose == VerbosityLevel::full)
     {
+      os << std::format(" {:^23} |{:^33}|\n", "node ID", "coordinates");
+      os << std::format(
+          "{:^12}|{:^12}|{:>10},{:>10},{:>10} |\n", "external", "internal(+1)", "X", "Y", "Z");
+      os << std::format("{:-<12}|{:-<12}|{:-<33}|\n", "", "", "");
       for (const auto& point : mesh.points_with_data())
       {
+        int internal_id = point.id() + 1;
         if (point.external_id() != invalid_external_id)
         {
-          os << "  " << point.external_id() << ": [";
+          os << std::format(" {:10} | {:10} |", point.external_id(), internal_id);
         }
         else
         {
-          os << "  [";
+          os << std::format(" {:10} | {:10} |", " ", internal_id);
         }
+        std::string coord_str;
         for (const auto& coord : point.coordinate())
         {
-          os << std::format("{:10.6g},", coord);
+          if (!coord_str.empty()) coord_str += ",";
+          coord_str += std::format("{:10.6g}", coord);
         }
-        os << "]\n";
+        os << coord_str << " |\n";
       }
-      os << "\n";
     }
     os << "cell-blocks:\n";
     for (const auto& cell_block : mesh.cell_blocks())
@@ -264,7 +270,7 @@ void Core::IO::MeshInput::print(const Mesh<dim>& mesh, std::ostream& os, Verbosi
     }
     os << "\n";
 
-    os << "point-sets:\n";
+    os << "point-sets (internal node numbers +1):\n";
     for (const auto& [ps_id, ps] : mesh.point_sets())
     {
       os << "  point-set " << ps_id;
@@ -285,6 +291,7 @@ void Core::IO::MeshInput::print(
 
   if (verbose == VerbosityLevel::full)
   {
+    os << "    (connectivity contains internal node numbers +1)\n";
     std::size_t i = 0;
     for (const auto& connectivity : block.cells())
     {
@@ -297,8 +304,13 @@ void Core::IO::MeshInput::print(
       {
         os << "    [";
       }
-      for (const auto& id : connectivity) os << id << ", ";
-      os << "]\n";
+      std::string connectivity_str;
+      for (const auto& id : connectivity)
+      {
+        if (!connectivity_str.empty()) connectivity_str += ", ";
+        connectivity_str += std::to_string(id + 1);
+      }
+      os << connectivity_str << "]\n";
       i++;
     }
     os << "\n";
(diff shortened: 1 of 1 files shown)


mayrmt:
@ischeider are you sure, that the result description uses the external ID? Should the difference just be an offset of 1 due to starting counting at 0 (internally) and 1 (externally)?

amgebauer:
The ids of elements and nodes are broken in 4C. We cannot use the external ids provided by the input mesh since we have a lot of the id-1 assumptions all over the code base. In #733, @sebproell tried to clean that up, but faced many problems where the results depended on the internal ids (which is super weird). 

Printing the external id is is quite error prone, since they are not the ones that are used for the result tests. There, we use the internal id + 1 currently. I think there is no easy fix for this (except, taking the hard way and finishing #733)

ischeider:
@mayrmt @amgebauer Okay, thanks for the very informative messages. I only noticed that the numbers in the point sets did not fit to the node coordinates. Now I wonder how people figured out the internal node number for using `RESULT DESCRIPTION`. 

How about having something like that in the mesh information output:
```

        node number       |
  external   | internal   |        node coordinates         |
  ---------- | ---------- |---------------------------------|
          20 |          0 |      -4.5,      -4.5,       2.5,|
         201 |          1 |      -4.5,      -4.5,       1.5,|
         273 |          2 |      -4.5,      -3.5,       1.5,|
          28 |          3 |      -4.5,      -3.5,       2.5,|
```
At least, there is a chance to find out the internal node numbering.

rjoussen:
@ischeider I think printing information like this could be a nice intermediate step, as long as we cannot use external IDs internally. Maybe as a hint for figuring out what internal ID 4C assigned to a node you want to perform a result test on: I normally just find the node again in the vtk output visually. The vtk output (not input!) node numbering corresponds to the internal node numbering in my experience.

amgebauer:
I agree, we could add this to the output. However, I would print the id+1 since this is the id that is typically printed in all 4C output

ischeider:
For `SHOW_INFO: full`, the following type of output is generated (taken from a personal example): 

```
Mesh consists of 6014 points and 3272 cells organized in 1 cell-blocks and 3 point-sets.

       node number       [        node coordinates         ]
  external  |internal(+1)[         X,         Y,         Z,]
------------|------------[---------------------------------]
         98 |          1 [         0,         6,        20,]
...
cell-blocks:
  cell-block 1 (): 3272 cells of type hex8
    (connectivity contains internal node numbers +1)
    1: [1, 2, 3, 4, 5, 6, 7, 8, ]
...
point-sets (internal node numbers +1):
  point-set 1 (top): 389 points
   5630, 5635, 5658, 5694, 5697, 5704, 5703, 5712, 5710, 5722, 5721, 5732,
```

src/core/io/src/4C_io_mesh.cpp
```diff
@@ -236,23 +236,27 @@ void Core::IO::MeshInput::print(const Mesh<dim>& mesh, std::ostream& os, Verbosi
   {
     if (verbose == VerbosityLevel::full)
     {
+      os << std::format(" {:^23} \n", "node number");
```
rjoussen:
```suggestion
      os << std::format(" {:^23} \n", "node ID");
```

src/core/io/src/4C_io_mesh.cpp
```diff
@@ -236,23 +236,27 @@ void Core::IO::MeshInput::print(const Mesh<dim>& mesh, std::ostream& os, Verbosi
...
+      os << std::format(
+          "{:^12}|{:^12}[{:^33}]\n", "external", "internal(+1)", "  node coordinates  ");
```
rjoussen:
```suggestion
          "{:^12}|{:^12}[{:^33}]\n", "external", "internal(+1)", "coordinates");
```

src/core/io/src/4C_io_mesh.cpp
```diff
@@ -236,23 +236,27 @@ void Core::IO::MeshInput::print(const Mesh<dim>& mesh, std::ostream& os, Verbosi
...
+      os << std::format(" {:-<11}| {:-<11}[{:-<33}]\n", "", "", "");
```
rjoussen: shouldnt this fill in the XYZ header?

ischeider: Indeed, it should. I missed to add the latest change into the commit. Fixed (together with the other comments. Thanks!

src/core/io/src/4C_io_mesh.cpp
```diff
@@ -236,23 +236,27 @@ void Core::IO::MeshInput::print(const Mesh<dim>& mesh, std::ostream& os, Verbosi
...
+        int internal_id = point.id() + 1;
         if (point.external_id() != invalid_external_id)
         {
-          os << "  " << point.external_id() << ": [";
+          os << std::format(" {:10} | {:10} [", point.external_id(), internal_id);
         }
         else
         {
-          os << "  [";
+          os << std::format(" {:10} | {:10} [", " ", internal_id);
```
rjoussen: can we make this look more like your initial suggestion?
```
        node number       |
  external   | internal   |        node coordinates         |
  ---------- | ---------- |---------------------------------|
          20 |          0 |      -4.5,      -4.5,       2.5,|
         201 |          1 |      -4.5,      -4.5,       1.5,|
         273 |          2 |      -4.5,      -3.5,       1.5,|
          28 |          3 |      -4.5,      -3.5,       2.5,|
```
I think the brackets `[]` look kinda strange. (Although they have been there before, so you can also leave them if you prefer, this is already a nice improvement!)

ischeider: Was also my impression. Changed it (and also deleted the comma after the last coordinate value)

rjoussen (approved):
Looks nice now!

mayrmt (approved):

PR #1209, opened 2025-09-02
Implement a wrapper around Teuchos::SerialDenseMatrix and hide Trilinos access in user code

bgoderbauer:
This PR changes our `Core::LinAlg::SerialDenseMatrix`, so it doesn't inherit from `Teuchos::SerialDenseMatrix` and instead builds a wrapper around the internal object. This way, Trilinos functions aren't directly exposed to the user. I changed the calls to Trilinos functions accordingly (if possible). 

There are a few steps left to do to completely hide the Trilinos SerialDense objects in our code:
- Wrapping of SerialDenseVector in the same manner
- Wrapping of SerialDenseSolver
- Some missing multiply calls

Overall, I've seen a lot of exchanges between SerialDenseVector and SerialDenseMatrix in the code. This isn't a problem with the Trilinos objects, but it didn't always work with the new wrapper.

PR #1983, opened 2026-04-17
Implement composition-based wrapper for SerialDenseSolver

bgoderbauer:
## Description and Context

This PR wraps `Teuchos::SerialDenseSolver` and thus hides user code exposure to Trilinos. Now, all these SerialDense objects are handled by our own implementation.

#1923, #1209


bgoderbauer:
Should I add a unittest for the solver? This would only ensure that nothing in the Trilinos backend suddenly stops working as expected. I don't know whether it is necessary here.

mayrmt (approved):

c-p-schmidt (approved):
I think it is not necessary to unit test in 4C.

PR #1988, opened 2026-04-21
Add performance testing framework

amgebauer:
This PR adds a performance testing framework to 4C.

Each performance test consists of an input file and two mesh files (a full one, and a minimal one). Both meshes should be the same problem with a different mesh resolution. The minimal one should be such that the test can run in the regular pipeline (it should run in a few seconds). The full mesh file is only used during performance testing. One can enable full performance testing with the option `FOUR_C_WITH_FULL_PERFORMANCE_TESTS`.

The tests run as usual during ctest. I added an extra fixture to the performance tests that collects the results in a (larger) json file. We can export this via the artifacts in the pipeline.

What is yet missing is the dedicated job for the performance testing (that runs in the nightly pipeline), and a job that collects the artifacts from the jobs and exports a report to Github pages.

Points to discuss/figure out:
 * Do we want to store the full meshes in the repo (each one might be a few MB)
 * Do we want to run the tests on Github hardware (might be less comparable over time), or on our own hardware (we can control the homogeneity of the hardware).
 * Where can we store historic performance data?

fyi @c-p-schmidt


davidrudlstorfer:
> Do we want to store the full meshes in the repo (each one might be a few MB)

https://docs.github.com/en/repositories/working-with-files/managing-large-files/configuring-git-large-file-storage 

@amgebauer maybe like this?

davidrudlstorfer:
> Do we want to run the tests on Github hardware (might be less comparable over time), or on our own hardware (we can control the homogeneity of the hardware).


And briefly regarding this; at least with the "performance" tests of BeamMe we see really large deviations from time to time which can be attributed to the utilization of the Github runners (based on each runner being only a VM IIRC?). Also the hardware changes from time to time. So I would definitely aim for some selfhosted/futureproof hardware that stays consistent.

Maybe a dedicated repo that does all of that in combination with git LFS and then just displaying the results on the main 4C page?

amgebauer:
> Maybe a dedicated repo that does all of that in combination with git LFS and then just displaying the results on the main 4C page?

We had that in the past. The problem is that if it is not included in PR testing, it is a lot manual work to migrate these tests. We can think about having the minimal tests in the repo, and the larger meshes and it's testing somewhere else.

tests/performance_tests/meshes/.gitattributes
davidrudlstorfer, line 1: Would that make more sense on a top level? Then this could be easily identified rather than some random folder within a folder within a folder

amgebauer: I added it into the subdirectory to make sure that even if we at some point rename the directory, it still remains in git-lfs

cmake/functions/four_c_performance_testing_functions.cmake
```diff
@@ -0,0 +1,222 @@
...
+
+  assert_required_arguments(_parsed TEST_FILE MESH)
+
+  if(NOT DEFINED _parsed_NP_FULL)
+    # Query the system for the number of logical cores
+    cmake_host_system_information(RESULT _parsed_NP_FULL QUERY NUMBER_OF_LOGICAL_CORES)
+
+    # Fallback to 1 if the query fails or returns 0
+    if(NOT _parsed_NP_FULL OR _parsed_NP_FULL EQUAL 0)
+      set(_parsed_NP_FULL 1)
```
c-p-schmidt: Why do we use these fallback options here and below rather than throwing an error?

cmake/setup_tests.cmake
```diff
@@ -94,6 +94,23 @@ else()
   message(STATUS "Benchmark tests with Google Benchmark: disabled")
 endif()
 
+four_c_process_global_option(
+  FOUR_C_WITH_FULL_PERFORMANCE_TESTS
```
sebproell:
```suggestion
  FOUR_C_ENABLE_FULL_PERFORMANCE_TESTS
```

would be the conventional name I introduced for features that are not dependencies. Yes, it is somewhat wrong that we still have `FOUR_C_WITH_GOOGLETEST` and `FOUR_C_WITH_GOOGLE_BENCHMARK`, these should probably be changed.

tests/performance_tests/meshes/full/README.md
```diff
@@ -0,0 +1,8 @@
+# Large meshes for performance testing
+
+This directory contains large mesh files dedicated to performance benchmarking. Due to their
+significant size, these files are tracked using Git Large File Storage (LFS). To add or modify a
+mesh in this folder, you must have Git LFS installed on your local system.
```
sebproell: I guess I also need LFS to check out the actual file blob? Never used it.

tests/performance_tests/meshes/.gitattributes
amgebauer: @davidrudlstorfer I moved this file now into the folder with the large files itself. I don't think that it is important enough to be in the main directory. It only affects the files in the respective folder.

davidrudlstorfer (dismissed):

c-p-schmidt (dismissed):

amgebauer:
I slightly adapted the test case and added a hex8 and tet10 variant.

Here is the pipeline that I triggered on my fork to test all the changes: https://github.com/amgebauer/4C/actions/runs/25040541453

c-p-schmidt (approved):
Looks good from my side. Let's check whether the pipeline on your fork passes. 🤞🏼

davidrudlstorfer (approved):

PR #1989, opened 2026-04-21
Fix python virtual environment during testing

amgebauer:
@danielwolff1 I suggest to fix it like this.

### DIFF
--- a/cmake/functions/four_c_testing_functions.cmake
+++ b/cmake/functions/four_c_testing_functions.cmake
@@ -26,7 +26,30 @@ function(require_fixture name_of_test name_of_required_test)
 endfunction()
 
 function(set_environment name_of_test)
-  set_tests_properties(${name_of_test} PROPERTIES ENVIRONMENT "PATH=$ENV{PATH}")
+  if(FOUR_C_WITH_PYTHON)
+    # Add Python virtual environment
+    if(NOT DEFINED FOUR_C_PYTHON_VENV_BUILD OR "${FOUR_C_PYTHON_VENV_BUILD}" STREQUAL "")
+      message(FATAL_ERROR "Could not determine Python virtual environment for testing.")
+    endif()
+
+    # We also need to add the PYTHONPATH with the site-packages of the virtual environment
+    if(NOT DEFINED FOUR_C_PYTHON_VENV_BUILD_SITE_PACKAGES
+       OR "${FOUR_C_PYTHON_VENV_BUILD_SITE_PACKAGES}" STREQUAL ""
+       )
+      message(
+        FATAL_ERROR
+          "Could not determine site-packages directory of Python virtual environment for testing."
+        )
+    endif()
+
+    set_property(
+      TEST ${name_of_test}
+      APPEND
+      PROPERTY ENVIRONMENT_MODIFICATION
+               "PATH=path_list_prepend:${FOUR_C_PYTHON_VENV_BUILD}/bin"
+               "PYTHONPATH=path_list_prepend:${FOUR_C_PYTHON_VENV_BUILD_SITE_PACKAGES}"
+      )
+  endif()
 endfunction()
 
 # set fail expressions to this test (name_of_test)
@@ -235,6 +258,7 @@ function(_add_test_with_options)
   endif()
 
   require_fixture(${_parsed_NAME_OF_TEST} "${_parsed_ADDITIONAL_FIXTURE};test_cleanup")
+  set_environment(${_parsed_NAME_OF_TEST})
   set_processors(${_parsed_NAME_OF_TEST} ${_parsed_TOTAL_PROCS})
   define_setup_fixture(${_parsed_NAME_OF_TEST} ${_parsed_NAME_OF_TEST})
   set_timeout(${_parsed_NAME_OF_TEST} ${_parsed_TIMEOUT})
@@ -532,9 +556,7 @@ function(__four_c_test_restart)
     endif()
 
     set(name_of_restart_check "${name_of_test}-check_restart_step")
-    set(check_command
-        "${FOUR_C_PYTHON_VENV_BUILD}/bin/check-restart-step ${control_file} ${_parsed_ASSERT_RESTART_STEP}"
-        )
+    set(check_command "check-restart-step ${control_file} ${_parsed_ASSERT_RESTART_STEP}")
 
     # Ensure that Python is listed as required dependency
     list(APPEND _parsed_REQUIRED_DEPENDENCIES "Python")
@@ -604,7 +626,7 @@ function(__four_c_test_add_csv_yaml_comparison)
   get_test_property(${_parsed_BASED_ON} _internal_OUTPUT_DIR test_directory)
 
   set(csv_comparison_command
-      "${FOUR_C_PYTHON_VENV_BUILD}/bin/diff-with-tolerance ${test_directory}/${_parsed_RESULT_FILE} ${PROJECT_SOURCE_DIR}/tests/input_files/${_parsed_REFERENCE_FILE} ${_parsed_TOL_R} ${_parsed_TOL_A}"
+      "diff-with-tolerance ${test_directory}/${_parsed_RESULT_FILE} ${PROJECT_SOURCE_DIR}/tests/input_files/${_parsed_REFERENCE_FILE} ${_parsed_TOL_R} ${_parsed_TOL_A}"
       )
 
   # Ensure that Python is listed as required dependency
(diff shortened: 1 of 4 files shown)


cmake/functions/four_c_testing_functions.cmake
```diff
@@ -1229,27 +1236,31 @@ function(four_c_test_restarted_vtk)
...
     set(run2_cmd
         "mkdir -p ${sim2_dir} && cd ${sim2_dir} && echo changing pwd to: && pwd \
+        source ${FOUR_C_PYTHON_VENV_BUILD}/bin/activate && \
       && ${extra_env}${MPIEXEC_EXECUTABLE} ${_mpiexec_all_args_for_testing} -np ${base_NP} $<TARGET_FILE:${FOUR_C_EXECUTABLE_NAME}> \
       ${input_path} ${sim2_dir}/xxx --restartfrom=${relative_path_from_sim2_to_sim1}/xxx --restart=${_parsed_RESTART_STEP}"
         )
   elseif(_parsed_RESTARTFROM_PATHTYPE STREQUAL "relative_from_parent"
          )# change into root_dir, then restartfrom folder_name_sim1
     set(run2_cmd
         "mkdir -p ${sim2_dir} && cd ${root_dir} && echo changing pwd to: && pwd \
```
danielwolff1, line 1268:
```suggestion
        "mkdir -p ${sim2_dir} && cd ${root_dir} && echo changing pwd to: && pwd && \
```

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -1229,27 +1236,31 @@ function(four_c_test_restarted_vtk)
...
+        source ${FOUR_C_PYTHON_VENV_BUILD}/bin/activate && \
       && ${extra_env}${MPIEXEC_EXECUTABLE} ${_mpiexec_all_args_for_testing} -np ${base_NP} $<TARGET_FILE:${FOUR_C_EXECUTABLE_NAME}> \
       ${input_path} ${sim2_dir}/xxx --restartfrom=${folder_name_sim1}/xxx --restart=${_parsed_RESTART_STEP}"
         )
   elseif(_parsed_RESTARTFROM_PATHTYPE STREQUAL "same_directory")
     # restartfrom sim1_dir (same directory), output is still written in sim2_dir for later vtk comparison
     set(run2_cmd
         "mkdir -p ${sim2_dir} && cd ${sim1_dir} && echo staying in: && pwd \
```
danielwolff1, line 1275:
```suggestion
        "mkdir -p ${sim2_dir} && cd ${sim1_dir} && echo staying in: && pwd && \
```

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -1259,6 +1270,7 @@ function(four_c_test_restarted_vtk)
     set(sim2_dir ${sim1_dir})
     set(run2_cmd
         "cd ${sim1_dir} && echo staying in: && pwd \
```
danielwolff1, line 1284:
```suggestion
        "cd ${sim1_dir} && echo staying in: && pwd && \
```

danielwolff1 (commented):
Thanks a lot for providing a fix so swiftly! 🙏 

The suggested changes look good to me, Copilot already spotted most missing `&&` in the added shell commands and I tried to mark the remaining ones.

amgebauer:
I now chose a slightly different approach: I added the bin dir of the correct virtual environment to the path variable in each ctest run.

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -26,7 +26,16 @@ function(require_fixture name_of_test name_of_required_test)
 endfunction()
 
 function(set_environment name_of_test)
-  set_tests_properties(${name_of_test} PROPERTIES ENVIRONMENT "PATH=$ENV{PATH}")
+  # Add Python venv and path variable
+  if(FOUR_C_WITH_PYTHON
```
danielwolff1: Is the `FOUR_C_WITH_PYTHON` flag always set for testing? Or can it happen that `FOUR_C_WITH_PYTHON=OFF` but `FOUR_C_WITH_PYBIND11=ON`?

danielwolff1 (dismissed):
Looks good to me!

cmake/functions/four_c_testing_functions.cmake
amgebauer: Python is required if you configure 4C with pybind11: https://github.com/4C-multiphysics/4C/blob/main/cmake/configure/configure_pybind11.cmake#L8-L10

c-p-schmidt (dismissed):
Looks good! Thank you for the fix!

amgebauer:
hm, my last approach did not work since it is not enough to set the correct python executable in the path. The python executable is determined already at link-time and is set to `${Python_EXECUTABLE}` which points to system python and not to the virtual environment. I think this is the correct approach since the we should link 4C against the Python that is found (or set by the user) during configure time.

During testing, we, however, need to use our own environment. So, I now determine the site-packages and set the PYTHON path accordingly in the testing framework. Let's see whether this works now.

danielwolff1 (dismissed):

tests/list_of_tests.cmake
```diff
@@ -2958,4 +2957,8 @@ four_c_test_restarted_vtk(TESTNAME correct-restarted-vtk-files-for-solid_runtime
 four_c_test_restarted_vtk(TESTNAME correct-restarted-vtk-files-for-solid_runtime_tet4 FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory PVD_RESULTFILENAME xxx-structure.pvd PVD_REFERENCEFILENAME ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-06 TIMEOUT 10 TIME_STEPS 0.5)
 four_c_test_restarted_vtk(TESTNAME correct-restarted-vtk-files-for-solid_runtime_tet4 FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory_implicit PVD_RESULTFILENAME xxx-1-structure.pvd PVD_REFERENCEFILENAME ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-06 TIMEOUT 10 TIME_STEPS 0.5)
 four_c_test_restarted_vtk(TESTNAME thermo3D_FBC_ost.4C.yaml-restarted-vtk-with-initial-state FILE thermo3D_FBC_ost.4C.yaml RESTART_STEP 3 RESTARTFROM_PATHTYPE relative_from_parent PVD_RESULTFILENAME xxx-thermo.pvd PVD_REFERENCEFILENAME ref/thermo3D_FBC_ost-thermo.pvd TOLERANCE 1e-06)
+
+# Create fixture to cleanup the test run
+four_c_test_cleanup()
```
davidrudlstorfer, line 2962: Does it make sense to move this to a function we have to deliberately call it? Looking at #1988 and more lists of tests we then need to add this specifically. Is there any benefit of moving this to a function rather than calling it directly?

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -1291,7 +1316,7 @@ function(four_c_test_restarted_vtk)
     set(result_dir "${sim2_dir}/${_parsed_PVD_RESULTFILENAME}")
 
     set(compare_cmd
-        "${FOUR_C_PYTHON_VENV_BUILD}/bin/vtk-compare ${result_dir} ${ref_path} ${_parsed_TOLERANCE} --points_in_time ${steps_joined}"
```
davidrudlstorfer, line 1294: Nice that this is finally gone!

tests/list_of_tests.cmake
amgebauer: Yes, I see two benefits:

First, we load the `testing_functions.cmake` very early in the configure process where the environment is not yet defined. So, at that time, we cannot set the environment since Python is not yet found. In theory, that would be ok since this cleanup does not need python, still, this is not clean.

Second, We magically add tests in a cmake file that is only supposed to define utility functions for testing. This is not very obvious actually.

Defining this as a function and calling it at a specific place ensures that we have control when the respective test is added.

davidrudlstorfer: Sounds valid:)

davidrudlstorfer (approved):

c-p-schmidt (approved):

amgebauer:
It took a while to get everything correct, but now all tests pass without manual adaptions during the workflow, so I think this should fix the nightly pipeline.

PR #1997, opened 2026-04-24
Move xfem input from inpar to xfem

maxiludwig:
## Description and Context

- Move XFEM input definitions from `Inpar::XFEM` in `src/inpar/` to the XFEM module in `src/xfem/`.
- Relocate input files (`4C_inpar_xfem.*` -> `4C_xfem_input.*`) and update all affected includes, qualifiers, and global registration wiring.
- Migrate XFEM qualifiers from `Inpar::XFEM` to `XFEM` across dependent modules.
- Rename the former `XFluidTimeInt` enum type to avoid a name conflict with the existing `XFEM::XFluidTimeInt` class.

## Related Issues
Part of #59


src/inpar/CMakeLists.txt
```diff
@@ -19,4 +19,3 @@ set(_dependencies
     )
 
 four_c_add_internal_dependency(${AUTO_DEFINED_MODULE_NAME} ${_dependencies})
-four_c_unity_build_compile_separately(${AUTO_DEFINED_MODULE_NAME} 4C_inpar_xfem.cpp)
```
davidrudlstorfer, line 22: Nice that this is gone

src/xfem/4C_xfem_input.hpp
```diff
@@ -0,0 +1,326 @@
...
+#include <vector>
+
+FOUR_C_NAMESPACE_OPEN
+
+// forward declaration
+namespace Core::Conditions
+{
+  class ConditionDefinition;
+}  // namespace Core::Conditions
+namespace XFEM
```
davidrudlstorfer, line 26: Maybe separate the actual implementation from the forward declaration with a newline, but not super important and we can merge as is

src/xfem/CMakeLists.txt
```diff
@@ -24,3 +24,4 @@ set(_dependencies
     )
 
 four_c_add_internal_dependency(${AUTO_DEFINED_MODULE_NAME} ${_dependencies})
+four_c_unity_build_compile_separately(${AUTO_DEFINED_MODULE_NAME} 4C_xfem_input.cpp)
```
davidrudlstorfer: what's the reason that we actually need this 🧐 further up I thought we could completely get rid of this

davidrudlstorfer (dismissed):

davidrudlstorfer:
Thanks a lot for tackling more inpar stuff @maxiludwig 🫶

c-p-schmidt (dismissed):
I agree with @davidrudlstorfer's comments.
Thank you for improving this!

src/xfem/CMakeLists.txt
maxiludwig: I just migrated this call together with the `inpar/input` file. I don't know why this is actually necessary.

amgebauer: This function is required for our Code Climate job, which utilizes a unity build to significantly reduce execution time. During this process, CMake concatenates a batch of .cpp files into a single compilation unit. While this speeds up the build, combining multiple units can occasionally lead to naming collisions. Therefore, the function above is not an indicator of poor code quality, but rather a necessary workaround to ensure the unity build compiles successfully.

Maybe, through the move, it will end up in a different compilation unit without a naming collusion, so that it might now not be necessary anymore. Might be worth a try.

maxiludwig: Thanks for the explanation! I removed it and it still seems to compile. :+1:

c-p-schmidt (approved):

davidrudlstorfer (approved):

PR #2005, opened 2026-04-28
Renaming and some input refactoring in viscoelasthyper toolbox

bgoderbauer:
## Description and Context

A first step towards cleaning the `viscoelasthyper` toolbox. 

### Most important changes
- The `GenMax` viscoelastic material (a standard linear solid model in this implementation) is **deprecated** in favor of the generalized Maxwell model.
- The generalized Maxwell model now has its proper name and a reworked input structure with less nesting.
- Adds a test for the exponential time discretization scheme of the generalized Maxwell model (previously unused)

<details>
<summary>Input changes</summary>

### Renamings
- `VISCO_Fract` -> `VISCO_FSLS` (see A. Birzle (2019), A viscoelastic nonlinear compressible material model of lung parenchyma – Experiments and numerical identification)
- `VISCO_GeneralizedGenMax` -> `VISCO_GeneralizedMaxwell`
    - `VISCO_BRANCH` -> `VISCO_GeneralizedMaxwellBranch`
    - `VISCO_PART` -> Removed, relaxation time `TAU` is now stored in `VISCO_GeneralizedMaxwellBranch`
    - Time discretization schemes:
        - `OST` -> `OneStepTheta`
        - `CONVOL` -> `ExponentialTimeDiscretization`

</details>


bgoderbauer:
This PR is not yet complete, I will add the renaming/deprecation of classes and methods. This contains only input changes for now.

tests/input_files/visco_generalized_genmax_ost_h8.4C.yaml
```diff
@@ -38,41 +38,32 @@ MATERIALS:
...
-      MATIDS: [40, 41]
+    VISCO_GeneralizedMaxwellBranch:
+      TAU: 0.5
+      MATID: 40
   - MAT: 40
     ELAST_CoupNeoHooke:
       YOUNG: 300
-  - MAT: 41
-    VISCO_PART:
-      TAU: 0.5
```
bgoderbauer, line 54: This is the biggest structural change inside the generalized Maxwell model. `VISCO_PART` is removed and directly integrated into the Maxwell branch.

ischeider: I haven't used this model so far, but wouldn't be the advantage of having a separate model for the viscous part that one could use a different (probably more complex) PDE?

bgoderbauer: Thanks for bringing this up! I want to share my perspective on this. Please feel free to correct me if I'm mistaken.

In theory, the current structure could be extended to specify nonlinear damping instead of only constant relaxation times. However, I think this is not something we should target with this material model.
In my opinion, there are two major reasons for this refactoring:
- From the code side, we only allow the constant relaxation time at the moment, so no utilities are changed. There is simply no other option implemented. The code design of this viscoelastic toolbox is also quite complex. It isn't easy to generalize for arbitrary nonlinear damping. I hope that by stripping it down step by step, it becomes easier to extend again in the future if desired. The changes in this PR simply focus on the existing tools.
- From a modeling perspective, nonlinear damping is also most likely a wrong model choice for a viscoelastic material. "Nonlinear" looking damping, as far as I know, is associated with materials that exhibit history- and/or state-dependent behavior. Mathematically, this just means that the relaxation function is not an exponential function, but rather any completely monotone, decreasing function. From viscoelastic theory, this can always be represented by a linear generalized Maxwell model with a sufficient number of branches. Since we allow different nonlinear constitutive models for each elastic spring of the generalized Maxwell branches, we can therefore already create any nonlinear anisotropic material model with arbitrary anisotropic relaxation processes.

bgoderbauer:
This would be ready for review :smile:

src/mat/elast/4C_mat_elast_summand.hpp
```diff
@@ -693,12 +692,10 @@ namespace Mat
...
       //! Indicator for the chosen viscoelastic formulations
       virtual void specify_visco_formulation(
-          bool& isovisco,     ///< global indicator for isotropic, split and viscous formulation
-          bool& viscogenmax,  ///< global indicator for viscous contribution according to the
-                              ///< SLS-Model
-          bool& viscogeneralizedgenmax,  ///< global indicator for viscoelastic contribution
-                                         ///< according to the generalized Maxwell Model
-          bool& viscofract  ///< global indicator for viscous contribution according the FSLS-Model
+          bool& visco_iso_rate,  ///< global indicator for isotropic rate-dependent visco response
+          bool& visco_legacy_sls_removed,   ///< global indicator for removed legacy SLS model
```
c-p-schmidt: Why do we still need it, if it is removed (according to the comment)?

src/mat/elast/4C_mat_elast_visco_fsls.hpp
```diff
@@ -25,13 +25,13 @@ namespace Mat
        * @brief material parameters for viscous contribution according the FSLS-model
        *
        * <h3>Input line</h3>
-       * MAT 1 VISCO_Fract TAU 0.1 ALPHA 0.5 BETA 1
+       * MAT 1 VISCO_FSLS TAU 0.1 ALPHA 0.5 BETA 1
        */
-      class Fract : public Core::Mat::PAR::Parameter
+      class Fsls : public Core::Mat::PAR::Parameter
```
c-p-schmidt, line 30: Is `Fsls` established in literature? Otherwise, I would vote for a more descriptive name.

src/mat/4C_mat_material_factory.cpp
```diff
@@ -802,17 +801,20 @@ std::unique_ptr<Core::Mat::PAR::Parameter> Mat::make_parameter(
...
-    case Core::Materials::mes_fract:
+    case Core::Materials::mes_fsls:
     {
-      return make_parameter_impl<Mat::Elastic::PAR::Fract>(id, type, input_data);
+      return make_parameter_impl<Mat::Elastic::PAR::Fsls>(id, type, input_data);
     }
     case Core::Materials::mes_genmax:
     {
-      return make_parameter_impl<Mat::Elastic::PAR::GenMax>(id, type, input_data);
+      FOUR_C_THROW(
```
c-p-schmidt: Do we really want to keep these and throw an error? I can get the motivation, as you can provide a nice error message, but is it relevant to have that in this case?

src/mat/elast/4C_mat_elast_summand.hpp
bgoderbauer: Just removed it!

src/mat/elast/4C_mat_elast_visco_fsls.hpp
bgoderbauer: Yes, either this name (fractional standard linear solid) or fractional Zener is used. FSLS is the name of the model in Anna Birzle's publication.

src/mat/4C_mat_material_factory.cpp
bgoderbauer: Removed :+1:

c-p-schmidt (approved):

davidrudlstorfer (approved):

PR #2006, opened 2026-04-28
Add a completion-oriented schema wrapper

### LINKED ISSUE #2002: vs-code .4C.yaml top-level auto-completion is bad
I dont know if that is my problem exclusively, but whenever I write an input file and want to add a section on the top-level, i.e. the autocompletion list only uses the first 10 entries of our schema, see below:

Image

Most of the time, these are not the sections I wanted to add. Even if I start typing, it searches for the right section only within these 10 section names:

Image

which is fairly useless. Once I typed in the right section name, it is recognized though and I get correct suggestions for downstream parameters.

After trying for too long to make the Red Hat YAML extension consume more values of our schema, I couldn't really find a setting that fixes that. Is anyone having the same problem? Or knows how to fix it?

I could imagine our schema is simply too long, which is why it's not working. That's why I looked into how to wrap our schema into a lighter version which states the property names explicity first, before the detailed descriptions from our normal schema. But I dont want to suggest that if there is an easier fix.

rjoussen:
## Description and Context

This adds a completion-oriented JSON schema that essentially wraps our normal schema and, for me at least, is much better at auto-completing section names (see #2002). 

Basically, a new schema (called `4C_schema.completion.json`) is created that lists all section names as `propertyNames`, provides a `defaultSnippet` for the function section, and references the normal schema for the actual validation. The completion-oriented schema can (but doesn't have to!) be referenced in VS Code to benefit from better top-level auto-completion.

- The existing schema remains unchanged and retains its name.
- This new schema is tested in PRs by validating all input files against it, exactly as the normal schema is tested too.
- The new schema is uploaded in the nightly tests as an artifact -> to be consumed by fourcipp.
- The documentation is updated where needed.

Closes #2002

### DIFF
--- a/doc/documentation/src/installation/installation.rst
+++ b/doc/documentation/src/installation/installation.rst
@@ -318,7 +318,7 @@ add a JSON schema file, which is automatically created during the build process,
 1. In File :math:`\to` Settings, search for "JSON schema", which will bring you to **JSON Schema Mappings**.
 2. Add a new entry by clicking on the "+" sign.
 3. Give it a descriptive name.
-4. Select the Schema file ``4C_schema.json`` from your local build directory.
+4. Select the Schema file ``4C_schema_completion.json`` from your local build directory.
 5. Select Schema version *JSON Schema v4*.
 6. Add the file name pattern ``*.4C.yaml``, to which this schema is applied.
 
@@ -537,7 +537,7 @@ In order to use the Yaml schema in VS Code, which simplifies editing |FOURC|  in
    ::
 
        "yaml.schemas": {
-         "<path_to>/4C_schema.json": "*.4C.yaml",
+         "<path_to>/4C_schema_completion.json": "*.4C.yaml",
        },
 
 The schema is now automatically applied on Yaml files that end with ``.4C.yaml``.
--- a/utilities/four_c_python/src/four_c_metadata/json_schema.py
+++ b/utilities/four_c_python/src/four_c_metadata/json_schema.py
@@ -300,6 +300,115 @@ def validator_to_schema(
             return None
 
 
+def create_function_section_default_snippet(
+    function_section_name: str, function_id_label: str
+) -> Schema:
+    return Schema(
+        schema_type=Types.object(
+            applicators=[
+                Applicators.defaultSnippets(
+                    [
+                        {
+                            "label": function_section_name,
+                            "description": "Insert a new FUNCT definition.",
+                            "body": {
+                                function_section_name.replace(
+                                    function_id_label,
+                                    f"${{1:{function_id_label}}}",
+                                ): ["${2:}"]
+                            },
+                        }
+                    ]
+                )
+            ]
+        ),
+    )
+
+
+def create_completion_schema(
+    top_level_property_names: list[str],
+    json_schema_path: pathlib.Path,
+    function_regex: str,
+    function_section_name: str,
+    function_id_label: str,
+) -> Schema:
+    """Wraps the full schema specified by json_schema_path in a completion-oriented
+    schema that exposes all top-level keys via propertyNames and provides a default
+    snippet for function sections.
+
+    Args:
+        top_level_property_names: list of all top-level property names
+        json_schema_path (pathlib.Path): Path to the full JSON schema to reference
+            in the completion schema.
+        function_regex (str): Regex pattern to validate function sections
+            (e.g., "^FUNCT[1-9][0-9]*$").
+        function_section_name (str): label of the function section
+        function_id_label (str): label for the function id in the snippet
+            (e.g., "<n>" for "FUNCT<n>")
+
+    Returns:
+        Schema: Returns a JSON schema object that can be used for completion-oriented
+        validation of 4C input files.
+    """
+    allowed_property_names = Schema(
+        schema_type=Types.string(
+            applicators=[
+                Applicators.anyOf(
+                    [
+                        Schema(
+                            schema_type=Types.string(
+                                validators=[
+                                    Validators.enum(sorted(top_level_property_names))
+                                ]
+                            )
+                        ),
+                        Schema(
+                            schema_type=Types.string(
+                                validators=[Validators.pattern(function_regex)]
+                            )
+                        ),
+                    ]
+                )
+            ]
+        )
+    )
+    completion_schema = Schema(
+        description=(
+            "Completion-oriented wrapper schema for 4C input files. "
+            "Uses propertyNames to surface all top-level keys and delegates "
+            "validation to the generated full schema."
+        ),
+        schema_type=Types.object(
+            applicators=[
+                Applicators.allOf(
+                    [
+                        Schema(
+                            schema_type=Types.object(
+                                applicators=[
+                                    Applicators.propertyNames(allowed_property_names)
+                                ]
+                            )
+                        ),
+                        create_function_section_default_snippet(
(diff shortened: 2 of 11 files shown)


utilities/four_c_python/src/four_c_metadata/json_schema.py
```diff
@@ -385,6 +495,21 @@ def main(fourc_metadata_yaml_path: str, json_schema_path: str):
...
+        function_section_name=function_section_name,
+        function_id_label=function_id_label,
+    )
+
+    json_completion_schema_path = json_schema_path.with_stem(
+        completion_schema_stem(json_schema_path.stem)
+    )
+    json_completion_schema_path.write_text(
+        json.dumps(completion_schema.to_dict(), indent=2)
+    )
```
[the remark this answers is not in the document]
rjoussen: Since the normal schema remains unchanged, this probably won't be a problem, @davidrudlstorfer?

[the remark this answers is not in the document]
davidrudlstorfer: Unfortunately, I do not know 100%, maybe you can just briefly test it.

As you've said, as long as the old schema is the same it will (99,999% sure) work.

But probably it would make sense to add some sort of testing on that schema either here in 4C or in FourCIPP?

Furthermore, this schema should also be added to FourCIPP so each version is available like the current schema (<= This is the most important point). This would also necessitate to upload the schema during build like the other schema and metadata file. See https://github.com/4C-multiphysics/4C/blob/dd555b4c0f019899ccb5db7f246191bc8f1caac4/.github/workflows/nightly_tests.yml#L128-L141

utilities/four_c_python/src/four_c_metadata/json_schema.py
```diff
@@ -300,6 +306,109 @@ def validator_to_schema(
...
+
+
+def create_completion_schema(
+    top_level_property_names: list[str],
+    json_schema_path: pathlib.Path,
+    function_regex: str,
+    function_section_name: str,
+    function_id_label: str,
+) -> Schema:
+    """Wraps the full schema specified by json_schema_path in a completion-oriented schema that exposes all top-level keys via propertyNames and provides a default snippet for function sections.
```
davidrudlstorfer: Could you maybe wrap this long line to our standard line length

utilities/four_c_python/src/four_c_metadata/json_schema.py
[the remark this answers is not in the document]
rjoussen: It is now uploaded as an artifact of the nightly pipeline and will be tested in fourcipp.

davidrudlstorfer (dismissed):
Tested it locally and it works as expected, thanks @rjoussen!

Only thing I already mentioned, maybe it makes sense to rename the original one to `4C_schema_full.json` or `4C_schema_complete.json` and the new one to `4C_schema_top_level.json` or something along those lines.

davidrudlstorfer:
And also let's not forget the changes in FourCIPP 🙏🏻

rjoussen:
> Only thing I already mentioned, maybe it makes sense to rename the original one to `4C_schema_full.json` or `4C_schema_complete.json` and the new one to `4C_schema_top_level.json` or something along those lines.

I decided to leave the old schema unchanged, mainly because it would require many changes at once and probably cause failing pipelines at first.


The new schema is now called `4C_schema.completion.json`.

We should also consider removing `4C_schema_partial.json`, as it seems pretty redundant and is not used/tested anywhere at the moment.

rjoussen:
[https://github.com/4C-multiphysics/4C/actions/runs/26362001518/job/77599351300?pr=2006](https://github.com/4C-multiphysics/4C/actions/runs/26362001518/job/77599351300?pr=2006) now validates all input files against the completion schema already.

rjoussen:
> And also let's not forget the changes in FourCIPP 🙏🏻

I have a version for that already, which I will push once this PR is merged.

doc/documentation/src/installation/installation.rst
```diff
@@ -333,7 +333,7 @@ Visual Studio Code
 `Visual Studio Code <https://code.visualstudio.com/>`_ is a code editor optimized for building and debugging modern web and cloud applications.
 It can also be used for developing |FOURC|.
 Visual Studio Code can connect to a remote computer so you can work on your home computer via SSH, see `here <https://code.visualstudio.com/docs/remote/remote-overview>`_.
-If you want to include a YAML schema for easier writing of input files within the project directory tree, see :ref:`below<vscode_yaml_schema>`.
+If you want to include the YAML completion schema for easier writing of input files within the project directory tree, see :ref:`below<vscode_yaml_schema>`.
```
davidrudlstorfer: Does it have any advantage to now update this everywhere to "completion schema"? I think in the end it is our YAML schema and not a completion schema? I think it would be more straightforward to keep the current naming in the docs with just "schema"

utilities/four_c_python/src/four_c_metadata/json_schema.py
```diff
@@ -385,6 +497,21 @@ def main(fourc_metadata_yaml_path: str, json_schema_path: str):
...
+    completion_schema = create_completion_schema(
+        top_level_property_names=list(properties.schemas.keys()),
+        json_schema_path=json_schema_path,
+        function_regex=function_regex,
+        function_section_name=function_section_name,
+        function_id_label=function_id_label,
+    )
+
+    json_completion_schema_path = json_schema_path.with_stem(
+        json_schema_path.stem + ".completion"
```
davidrudlstorfer: i am super torn here, does it make sense to add another "file extension" as the name? basically we could also append 


```suggestion
        json_schema_path.stem + "_completion"
```

like suggested above. Adding random extensions does not always play super nicely with IDEs and also maybe different OS (windows)

Is there any benefit of adding the new file type "*.completion.json"?

davidrudlstorfer (commented):
Two final questions, sorry that this takes so many rounds

Probably it also makes sense to then open some discussion/announcement on the new feature so users are aware of it

doc/documentation/src/installation/installation.rst
rjoussen: Okay, I will revert the changes except for the parts vscode/clion setup explanation.

utilities/four_c_python/src/four_c_metadata/json_schema.py
rjoussen: I'll rename it to `4C_schema_completion.json`.

davidrudlstorfer (approved):
Thanks a lot for this nice addition @rjoussen! 

(Sorry that the review rounds took that long from my side)

doc/documentation/src/installation/installation.rst
rjoussen: I changed it in moste places. The only thing that remains is that we do not reference the exact name of the schema in some places that refer to the installation guide anyway. Otherwise, it would be confusing if a tutorial says use the `4C_schema.json`, see the installation guide, and the installation guide says use `4C_schema_completion.json`

rjoussen:
> (Sorry that the review rounds took that long from my side)

Don't worry! You could make up for it by finding someone else to review :)

rjoussen:
> Probably it also makes sense to then open some discussion/announcement on the new feature so users are aware of it

Ill do that as soon as the changes are in fourcipp as well and everything works in the nightly tests

bgoderbauer (approved):

PR #2007, opened 2026-04-29
Refactor the drucker prager model to use tensors instead of matrices

ischeider:
## Description and Context

Now also the Drucker Prager Material model is supposed to use tensors instead of matrices.
The changes are significant though (all the peculiar scalar types are gone, and the FAD routines vanished then as well).
GitHub Copilot assisted a lot ;-)

## Interested parties
@BishrMaradni

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
@@ -19,6 +19,7 @@
 #include "4C_mat_fluidporo_singlephase.hpp"
 #include "4C_mat_inelastic_defgrad_factors_service.hpp"
 #include "4C_mat_micromaterial.hpp"
+#include "4C_mat_plasticdruckerprager.hpp"
 #include "4C_mat_scatra_growth_remodel.hpp"
 #include "4C_mat_scatra_nonlocal_stimulus.hpp"
 #include "4C_porofluid_pressure_based_elast_scatra_input.hpp"
@@ -1153,24 +1154,42 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
   /*----------------------------------------------------------------------*/
   // Plastic linear elastic St.Venant Kirchhoff / Drucker Prager plasticity
   {
+    using namespace Core::IO::InputSpecBuilders::Validators;
     known_materials[Core::Materials::m_pldruckprag] = group("MAT_Struct_DruckerPrager",
         {
-            parameter<double>("YOUNG", {.description = "Young's modulus"}),
-            parameter<double>("NUE", {.description = "Poisson's ratio"}),
-            parameter<double>("DENS", {.description = "Density"}),
-            parameter<double>("ISOHARD", {.description = "linear isotropic hardening"}),
-            parameter<double>("TOL", {.description = "Local Newton iteration tolerance"}),
-            parameter<double>("C", {.description = "cohesion"}),
-            parameter<double>("ETA", {.description = "Drucker Prager Constant Eta"}),
-            parameter<double>("XI", {.description = "Drucker Prager Constant Xi"}),
-            parameter<double>("ETABAR", {.description = "Drucker Prager Constant Etabar"}),
-            parameter<std::string>("TANG", {.description = "Method to compute the material tangent",
-                                               .default_value = "consistent"}),
-            parameter<int>(
-                "MAXITER", {.description = "Maximum Iterations for local Neutron Raphson",
-                               .default_value = 50}),
+            parameter<double>(
+                "YOUNG", {.description = "Young's modulus", .validator = positive<double>()}),
+            parameter<double>(
+                "NUE", {.description = "Poisson's ratio (must be in (-1, 0.5) for stability)",
+                           .validator = in_range(excl(-1.0), excl(0.5))}),
+            parameter<double>(
+                "DENS", {.description = "Density", .validator = positive_or_zero<double>()}),
+            parameter<double>(
+                "ISOHARD", {.description = "linear isotropic hardening modulus $H_\\mathrm{iso}$",
+                               .validator = positive_or_zero<double>()}),
+            parameter<double>(
+                "C", {.description = "cohesion $c$", .validator = positive<double>()}),
+            parameter<double>("ETA", {.description = "Drucker Prager Constant $\\eta$"}),
+            parameter<double>("XI", {.description = "Drucker Prager Constant $\\xi$"}),
+            parameter<double>("ETABAR",
+                {.description = "Drucker Prager Constant $\\overline{\\eta}$ "
+                                "(set $\\overline{\\eta} = \\eta$ for associative flow rule)"}),
+            parameter<Mat::PAR::PlasticDruckerPrager::TangentType>("TANG",
+                {.description = "Method to compute the material tangent (consistent / elastic)",
+                    .default_value = Mat::PAR::PlasticDruckerPrager::TangentType::consistent}),
+            parameter<double>("TOL", {.description = "Local Newton iteration tolerance",
+                                         .default_value = 1.0e-08,
+                                         .validator = positive<double>()}),
+            parameter<int>("MAXITER", {.description = "Maximum Iterations for local Newton Raphson",
+                                          .default_value = 50,
+                                          .validator = positive<int>()}),
         },
-        {.description = "elastic St.Venant Kirchhoff / plastic drucker prager"});
+        {.description = "Linear-elasto-plastic material with a Drucker-Prager yield surface: "
+                        "$\\phi = \\sqrt{J_2} + \\eta \\, \\mathrm{tr}(\\mathbf{\\sigma}) - "
+                        "\\xi\\, (c + H_\\mathrm{iso}\\varepsilon_\\mathrm{p}^\\mathrm{acc}$ "
+                        "and a potentially non-associated plastic potential: = "
+                        "\\sqrt{J_2} + \\overline{\\eta} \\, \\mathrm{tr}(\\mathbf{\\sigma}) - "
+                        "\\xi\\, (c + H_\\mathrm{iso}\\varepsilon_\\mathrm{p}^\\mathrm{acc}$"});
   }
 
   /*----------------------------------------------------------------------*/
--- a/src/mat/4C_mat_plasticdruckerprager.cpp
+++ b/src/mat/4C_mat_plasticdruckerprager.cpp
@@ -12,16 +12,11 @@
 #include "4C_global_data.hpp"
 #include "4C_linalg_FADmatrix_utils.hpp"
 #include "4C_linalg_fixedsizematrix.hpp"
-#include "4C_linalg_fixedsizematrix_voigt_notation.hpp"
 #include "4C_linalg_tensor.hpp"
-#include "4C_linalg_tensor_conversion.hpp"
-#include "4C_linalg_tensor_generators.hpp"
-#include "4C_mat_material_factory.hpp"
 #include "4C_mat_par_bundle.hpp"
 #include "4C_mat_stvenantkirchhoff.hpp"
 #include "4C_material_base.hpp"
 #include "4C_material_parameter_base.hpp"
-#include "4C_structure_new_input.hpp"
 #include "4C_utils_enum.hpp"
 #include "4C_utils_local_newton.hpp"
 
@@ -39,7 +34,7 @@ Mat::PAR::PlasticDruckerPrager::PlasticDruckerPrager(const Core::Mat::PAR::Param
       eta_(matdata.parameters.get<double>("ETA")),
       xi_(matdata.parameters.get<double>("XI")),
       etabar_(matdata.parameters.get<double>("ETABAR")),
-      tang_(matdata.parameters.get<std::string>("TANG")),
+      tang_(matdata.parameters.get<PAR::PlasticDruckerPrager::TangentType>("TANG")),
       itermax_(matdata.parameters.get<int>("MAXITER"))
 {
 }
@@ -72,25 +67,17 @@ void Mat::PlasticDruckerPrager::pack(Core::Communication::PackBuffer& data) cons
   int matid = -1;
   if (params_ != nullptr) matid = params_->id();
   add_to_pack(data, matid);
-  int histsize = initialized() ? strainpllast_.size() : 0;
-  add_to_pack(data, histsize);
-  for (int var = 0; var < histsize; ++var)
-  {
-    // insert last converged states
-    add_to_pack(data, strainpllast_.at(var));
-    add_to_pack(data, strainbarpllast_.at(var));
+  // insert last converged states
+  add_to_pack(data, strainpllast_);
+  add_to_pack(data, strainbarpllast_);
 
-    // insert current iteration states
-    add_to_pack(data, strainplcurr_.at(var));
-    add_to_pack(data, strainbarplcurr_.at(var));
-  }
+  // insert current iteration states
+  add_to_pack(data, strainplcurr_);
+  add_to_pack(data, strainbarplcurr_);
 }
 
 void Mat::PlasticDruckerPrager::unpack(Core::Communication::UnpackBuffer& buffer)
 {
-  isinit_ = true;
-
-
   Core::Communication::extract_and_assert_id(buffer, unique_par_object_id());
 
   int matid;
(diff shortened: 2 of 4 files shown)


src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -115,24 +115,24 @@ void Mat::PlasticDruckerPrager::unpack(Core::Communication::UnpackBuffer& buffer
...
       strainbarpllast_.push_back(tmp_scalar);
 
       // current iteration states are unpacked
-      extract_from_pack(buffer, tmp_vector);
-      strainplcurr_.push_back(tmp_vector);
+      extract_from_pack(buffer, tmp_tensor);
+      strainplcurr_.push_back(tmp_tensor);
       extract_from_pack(buffer, tmp_scalar);
       strainbarplcurr_.push_back(tmp_scalar);
     }
```
amgebauer: You could also directly extract (and pack) the full vectors instead of each component separately. This would simplify this pack logic a lot.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -141,11 +141,17 @@ void Mat::PlasticDruckerPrager::unpack(Core::Communication::UnpackBuffer& buffer
...
-  strainbarpllast_.resize(numgp);
-  strainbarplcurr_.resize(numgp);
+  strainbarpllast_.assign(numgp, 0.0);
+  strainbarplcurr_.assign(numgp, 0.0);
+
+  for (int i = 0; i < numgp; ++i)
+  {
+    strainpllast_.at(i).fill(0.0);
+    strainplcurr_.at(i).fill(0.0);
+  }
```
amgebauer: I don't thunk that you need this, the vector should ale´ready be initialized with zero tensors.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -155,7 +161,7 @@ void Mat::PlasticDruckerPrager::update()
   strainpllast_ = strainplcurr_;
   strainbarpllast_ = strainbarplcurr_;
 
-  std::for_each(strainplcurr_.begin(), strainplcurr_.end(), [](auto& item) { item.clear(); });
+  std::for_each(strainplcurr_.begin(), strainplcurr_.end(), [](auto& item) { item.fill(0.0); });
```
amgebauer: I don't know the details of the material, but do you really need to reset the plastic strain here? Often it makes sense to keep it as an initial value so that the local Newton algorithm is started close to the solution. But I haven't looked into the details of the material, so I might be wrong here.

unittests/mat/4C_druckerprager_test.cpp
```diff
@@ -194,51 +193,48 @@ namespace
...
-        ref_cmat(i, j) = result_stress(i).dx(j);
-      }
-    }
-    FOUR_C_EXPECT_NEAR(result_cmat_view, ref_cmat, 1.0e-12);
+    input_strain(0, 0) = 1.0;
+    input_strain(1, 1) = 1.0;
+    input_strain(2, 2) = 1.0;
+    input_strain(0, 1) = 0.0;
+    input_strain(1, 2) = 0.0;
+    input_strain(0, 2) = 0.0;
```
amgebauer:
```suggestion
  input_strain = Core::LinAlg::TensorGenerators::identity<double, 3, 3>;
```

the same applies at some other places.

amgebauer (dismissed):
Thank you @ischeider for migrating this material to tensors!

src/mat/4C_mat_plasticdruckerprager.cpp
BishrMaradni: Based on the reference they set it to 0 at setup, but I think this is not necessary as the plastic strain is calculated as the total strain - elastic strain. At the same time, the newton iterator is solving for the elastic component of the strain as far as I recall.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -42,6 +39,9 @@ Mat::PAR::PlasticDruckerPrager::PlasticDruckerPrager(const Core::Mat::PAR::Param
       tang_(matdata.parameters.get<std::string>("TANG")),
       itermax_(matdata.parameters.get<int>("MAXITER"))
 {
+  FOUR_C_ASSERT_ALWAYS(cohesion_ > 1.0e-14, "Cohesion must be positive, but got {}.", cohesion_);
+  FOUR_C_ASSERT_ALWAYS(poissonratio_ < 0.5 - 1.0e-14,
+      "Poisson's ratio must be strictly less than 0.5, but got {}.", poissonratio_);
```
rjoussen: These checks should be input validators. Also, I assume almost all of the values need some sort of validation.

Especially `tang_` should definitely be an enum (though not strictly part of this PR)

ischeider: I never used the validators before. Thanks for suggesting these.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -171,220 +174,124 @@ void Mat::PlasticDruckerPrager::setup_cmat(
...
+
+  const double J2 = 0.5 * ddot(devstress, devstress);
+  double strainbar_p = strainbarpllast_.at(gp);
+  const double Phi_trial = std::sqrt(J2) + eta * p_trial - xi * cohesion - xi * Hiso * strainbar_p;
+
+  double Dgamma = 0.0;
+  double dstrainv = 0.0;
+
+  if (Phi_trial / std::abs(cohesion) > params_->abstol_)
   {
```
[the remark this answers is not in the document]
ischeider: input validators included.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -171,220 +174,124 @@ void Mat::PlasticDruckerPrager::setup_cmat(
...
     { return this->return_to_cone_funct_and_deriv(Dgamma_init, G, kappa, Phi_trial); };
 
-    const double tol = params_->abstol_;
-    Dgamma =
-        Core::Utils::solve_local_newton(returnToConeFunctAndDeriv, Dgamma, tol * cohesion, itermax);
-    strainbar_p = (strainbarpllast_.at(gp)) + xi * Dgamma;
-    devstress.scale(1.0 - (G * Dgamma / std::sqrt(J2)));
+    Dgamma = Core::Utils::solve_local_newton(
+        returnToConeFunctAndDeriv, Dgamma, params_->abstol_ * cohesion, itermax);
+    strainbar_p = strainbarpllast_.at(gp) + xi * Dgamma;
```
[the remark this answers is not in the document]
ischeider: input validator included as well.

src/mat/4C_mat_plasticdruckerprager.cpp
ischeider: This is not clear to me. If you think I should pack/unpack the variables for all gauss points at once: I looked at other materials, but they all do it this way. Can you give any example for a leaner method?

src/mat/4C_mat_plasticdruckerprager.cpp
ischeider: I don't think it's necessary either ;-)

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -171,220 +174,124 @@ void Mat::PlasticDruckerPrager::setup_cmat(
...
+    devstress *= (1.0 - G * Dgamma / std::sqrt(J2));
     p = p_trial - kappa * etabar * Dgamma;
```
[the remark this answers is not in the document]
ischeider: Since sqrt(J2) - G*Dgamma/abs(cohesion) > abstol, J2 cannot be close to zero.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -171,220 +174,124 @@ void Mat::PlasticDruckerPrager::setup_cmat(
...
-      cmat_view(i, k) += epfac2 * D(i) * D(k);
-    }
-  }
+  constexpr auto id2 = Core::LinAlg::TensorGenerators::identity<double, 3, 3>;
+  constexpr auto Is = Core::LinAlg::TensorGenerators::symmetric_identity<double, 3, 3, 3, 3>;
+  constexpr auto Id = Is - 1.0 / 3.0 * dyadic(id2, id2);
 
-  epfac3 = -sqrt(2) * G * A * Kappa;
+  const double nd = std::sqrt(ddot(devstrain, devstrain));
+  const auto D = devstrain / nd;
```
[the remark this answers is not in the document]
ischeider: This only happens in the cone region of plastic state, where the stress state cannot be hydrostatic.

src/mat/4C_mat_plasticdruckerprager.cpp
amgebauer: You can just write

```c++
extract_from_pack(buffer, strainpllast_);
extract_from_pack(buffer, strainbarpllast_);
extract_from_pack(buffer, strainplcurr_);
extract_from_pack(buffer, strainbarplcurr_);
```

and correspondingly for packing
```c++
add_to_pack(data, strainpllast_);
add_to_pack(data, strainbarpllast_);
add_to_pack(data, strainplcurr_);
add_to_pack(data, strainbarplcurr_);
```

This avoids the manual loop over all gauss points.

Note, previously, this was often not possible. That's why other materials often also do this manual loop.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1205,22 +1205,30 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
             parameter<std::string>("TANG", {.description = "Method to compute the material tangent",
                                                .default_value = "consistent"}),
-            parameter<int>(
-                "MAXITER", {.description = "Maximum Iterations for local Neutron Raphson",
-                               .default_value = 50}),
+            parameter<int>("MAXITER", {.description = "Maximum Iterations for local Newton Raphson",
+                                          .default_value = 50,
+                                          .validator = positive<int>()}),
         },
         {.description = "elastic St.Venant Kirchhoff / plastic drucker prager"});
```
```suggestion
        {.description = "Linear-elasto-plastic material with a Drucker-Prager yield surface: $\\phi = \\sqrt{J_2} + \\eta*\\mathrm{tr}(\\mathbf{\\sigma}) - \\xi\\cdot(c + H_\\mathrm{iso}\\varepsilon_\\mathrm{p}^\\mathrm{acc}$ and a potentially non-associated plastic potential: = \\sqrt{J_2} + \\overline{\\eta}*\\mathrm{tr}(\\mathbf{\\sigma}) - \\xi\\cdot(c + H_\\mathrm{iso}\\varepsilon_\\mathrm{p}^\\mathrm{acc}$"});
```
I believe this is what is happening in the material.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1205,22 +1205,30 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
-            parameter<double>("TOL", {.description = "Local Newton iteration tolerance"}),
-            parameter<double>("C", {.description = "cohesion"}),
+            parameter<double>(
+                "YOUNG", {.description = "Young's modulus", .validator = positive<double>()}),
+            parameter<double>(
+                "NUE", {.description = "Poisson's ratio (must be in (-1, 0.5) for stability)",
+                           .validator = in_range(excl(-1.0), excl(0.5))}),
+            parameter<double>(
+                "DENS", {.description = "Density", .validator = positive_or_zero<double>()}),
+            parameter<double>("ISOHARD", {.description = "linear isotropic hardening",
```
```suggestion
            parameter<double>("ISOHARD", {.description = "Linear isotropic hardening modulus $H_\\mathrm{iso}$",
```
same for the other parameters.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1205,22 +1205,30 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                                             .validator = positive_or_zero<double>()}),
+            parameter<double>("TOL", {.description = "Local Newton iteration tolerance",
+                                         .validator = positive<double>()}),
+            parameter<double>("C", {.description = "cohesion (must be strictly positive)",
+                                       .validator = positive<double>()}),
             parameter<double>("ETA", {.description = "Drucker Prager Constant Eta"}),
             parameter<double>("XI", {.description = "Drucker Prager Constant Xi"}),
             parameter<double>("ETABAR", {.description = "Drucker Prager Constant Etabar"}),
             parameter<std::string>("TANG", {.description = "Method to compute the material tangent",
```
rjoussen: the `"TANG"` parameter should be an enum.

src/mat/4C_mat_plasticdruckerprager.cpp
rjoussen: I agree, this would make packing/unpacking `histsize` unnecessary as well.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -141,11 +138,11 @@ void Mat::PlasticDruckerPrager::unpack(Core::Communication::UnpackBuffer& buffer
...
-  strainplcurr_.resize(numgp);
+  strainpllast_ = std::vector<Core::LinAlg::SymmetricTensor<double, 3, 3>>(numgp);
+  strainplcurr_ = std::vector<Core::LinAlg::SymmetricTensor<double, 3, 3>>(numgp);
 
-  strainbarpllast_.resize(numgp);
-  strainbarplcurr_.resize(numgp);
+  strainbarpllast_.assign(numgp, 0.0);
+  strainbarplcurr_.assign(numgp, 0.0);
 
   isinit_ = true;
```
rjoussen: Setup is called unconditionally, so isinit_ is not needed here I think. The variable can vanish.

src/mat/4C_mat_plasticdruckerprager.hpp
```diff
@@ -123,82 +121,45 @@ namespace Mat
...
     double density() const override { return params_->density_; }
-    template <typename T>
-    std::pair<T, T> return_to_cone_funct_and_deriv(T Dgamma, T G, T kappa, T Phi_trial);
-    template <typename T>
-    std::pair<T, T> return_to_apex_funct_and_deriv(T dstrainv, T p, T kappa, T strainbar_p);
+    std::pair<double, double> return_to_cone_funct_and_deriv(
+        double Dgamma, double G, double kappa, double Phi_trial);
+    std::pair<double, double> return_to_apex_funct_and_deriv(
+        double dstrainv, double p, double kappa, double strainbar_p);
     bool initialized() const { return (isinit_ and !strainplcurr_.empty()); }
```
rjoussen: I think the `initialized()` function is unnecessary

src/mat/4C_mat_plasticdruckerprager.hpp
```diff
@@ -123,82 +121,45 @@ namespace Mat
...
     Core::Mat::PAR::Parameter* parameter() const override { return params_; }
     double density() const override { return params_->density_; }
-    template <typename T>
-    std::pair<T, T> return_to_cone_funct_and_deriv(T Dgamma, T G, T kappa, T Phi_trial);
-    template <typename T>
-    std::pair<T, T> return_to_apex_funct_and_deriv(T dstrainv, T p, T kappa, T strainbar_p);
+    std::pair<double, double> return_to_cone_funct_and_deriv(
+        double Dgamma, double G, double kappa, double Phi_trial);
+    std::pair<double, double> return_to_apex_funct_and_deriv(
+        double dstrainv, double p, double kappa, double strainbar_p);
```
rjoussen, line 167:
```suggestion

    std::pair<double, double> return_to_cone_funct_and_deriv(
        const double Dgamma, const double G, const double kappa, const double Phi_trial);

    std::pair<double, double> return_to_apex_funct_and_deriv(
        const double dstrainv, const double p, const double kappa, const double strainbar_p);
```

src/mat/4C_mat_plasticdruckerprager.hpp
```diff
@@ -123,82 +121,45 @@ namespace Mat
...
-        const Core::LinAlg::Matrix<NUM_STRESS_3D, 1, ScalarT>& linstrain,
-        const Teuchos::ParameterList& params,
-        Core::LinAlg::Matrix<NUM_STRESS_3D, 1, ScalarT>& stress,
-        Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3>& cmat, int gp, int eleGID);
-    template <typename T>
-    void stress(const T p, const Core::LinAlg::Matrix<NUM_STRESS_3D, 1, T>& devstress,
-        Core::LinAlg::Matrix<NUM_STRESS_3D, 1, T>& stress) const;
+        Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3>& cmat, int gp, int eleGID) override;
 
     void setup_cmat(Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3>& cmat) const;
```
rjoussen: I would remove this and use `Mat::StVenantKirchhoff::evaluate_stress_linearization(young, nu)` directly.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -171,220 +165,127 @@ void Mat::PlasticDruckerPrager::setup_cmat(
 
 void Mat::PlasticDruckerPrager::setup_cmat_elasto_plastic_cone(
     Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3>& cmat, double Dgamma, double G, double Kappa,
-    Core::LinAlg::Matrix<NUM_STRESS_3D, 1>& devstrain, double xi, double Hiso, double eta,
-    double etabar) const
+    const Core::LinAlg::SymmetricTensor<double, 3, 3>& devstrain, double xi, double Hiso,
+    double eta, double etabar) const
```
rjoussen:
```suggestion
Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3> Mat::PlasticDruckerPrager::setup_cmat_elasto_plastic_cone(
    const double Dgamma, const double G, const double Kappa,
    const Core::LinAlg::SymmetricTensor<double, 3, 3>& devstrain, const double xi, const double Hiso,
    const double eta, const double etabar) const
```

I think this would be clearer, same for the other cmat function.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -11,13 +11,10 @@
 #include "4C_comm_parobject.hpp"
 #include "4C_global_data.hpp"
 #include "4C_inpar_structure.hpp"
-#include "4C_linalg_FADmatrix_utils.hpp"
 #include "4C_linalg_fixedsizematrix.hpp"
```
rjoussen, line 14:
```suggestion
#include "4C_global_data.hpp"
#include "4C_linalg_fixedsizematrix.hpp"
```
`#include "4C_inpar_structure.hpp"`  is not needed, same for #include `"4C_utils_enum.hpp"`

src/mat/4C_mat_plasticdruckerprager.hpp
```diff
@@ -13,18 +13,16 @@
 #include "4C_comm_parobjectfactory.hpp"
 #include "4C_inpar_structure.hpp"
 #include "4C_linalg_tensor_conversion.hpp"
-#include "4C_mat_material_factory.hpp"
+#include "4C_linalg_tensor_generators.hpp"
 #include "4C_mat_so3_material.hpp"
```
rjoussen:
```suggestion
#include "4C_inpar_structure.hpp"
#include "4C_mat_so3_material.hpp"
```
`#include "4C_linalg_tensor_conversion.hpp"` and `#include "4C_linalg_tensor_generators.hpp"` are not needed.

rjoussen (dismissed):
Looks good to me now, thanks for the effort!

ischeider:
Small polishing changes (adhering also the copilot suggestions), also squashed two commits into one.

unittests/mat/4C_druckerprager_test.cpp
```diff
@@ -194,51 +192,38 @@ namespace
...
-      for (int j = 0; j < 6; j++)
-      {
-        ref_cmat(i, j) = result_stress(i).dx(j);
-      }
-    }
-    FOUR_C_EXPECT_NEAR(Core::FADUtils::cast_to_double(result_stress),
-        Core::FADUtils::cast_to_double(ref_stress), 1.0e-12);
+    input_strain = 0.2 * Core::LinAlg::TensorGenerators::identity<double, 3, 3>;
+    druckprag_->evaluate(nullptr, input_strain, paras, context, result_stress, result_cmat, 0, 0);
+    // After large plastic loading, a small strain should be elastic unloading
```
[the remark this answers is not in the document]
ischeider: Refactored this test, so that it now tests reverse loading.

src/mat/4C_mat_plasticdruckerprager.cpp
```diff
@@ -110,285 +97,161 @@ void Mat::PlasticDruckerPrager::unpack(Core::Communication::UnpackBuffer& buffer
...
-  Core::LinAlg::Matrix<Mat::NUM_STRESS_3D, 1> plstrain(Core::LinAlg::Initialization::zero);
-
-  ScalarT young = params_->youngs_;
-  ScalarT nu = params_->poissonratio_;
-  ScalarT Hiso = params_->isohard_;
-  ScalarT cohesion = params_->cohesion_;
-  ScalarT eta = params_->eta_;
-  ScalarT xi = params_->xi_;
-  ScalarT etabar = params_->etabar_;
+  const double young = params_->youngs_;
```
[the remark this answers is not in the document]
ischeider: Not common in 4C, didn't implement.

rjoussen (approved):
Looks good to me!

georghammerl (approved):

PR #2009, opened 2026-04-29
[beaminteraction]Dual Hermite shapefunctions for Lagrange Multipliers

dharinib98:
## Description and Context

This is an initial working version of the implementation of dual Hermite shape functions in beaminteraction. The current implementation still requires substantial refactoring to improve code efficiency and maintainability.

## Interested Parties
@mayrmt @isteinbrecher @maxfirmbach


src/geometry_pair/4C_geometry_pair_element_shape_functions.hpp
```diff
@@ -185,6 +185,58 @@ namespace GeometryPair
...
+     * @param xi (in) Parameter coordinate on the element.
+     * @param shape_function_data (in) Shape function data container.
+     */
+    template <typename V, typename T>
+    static void evaluate_deriv2(
+        V& ddN, const T& xi, const ShapeFunctionData<t_hermite_dual>& shape_function_data)
+    {
+      Core::FE::shape_function_dual_hermite_1d_deriv2(
+          ddN, xi, shape_function_data.ref_length_, t_hermite::discretization_);
+    }
```
isteinbrecher: You don't need `evaluate_deriv1` and `evaluate_deriv2`. We never need derivatives of the dual shape functions.

src/geometry_pair/4C_geometry_pair_element_shape_functions.hpp
```diff
@@ -185,6 +185,58 @@ namespace GeometryPair
...
+     *
+     * @param N (out) shape functions.
+     * @param xi (in) Parameter coordinate on the element.
+     * @param shape_function_data (in) Shape function data container.
+     */
+    template <typename V, typename T>
+    static void evaluate(
+        V& N, const T& xi, const ShapeFunctionData<t_hermite_dual>& shape_function_data)
+    {
+      Core::FE::shape_function_dual_hermite_1d(
```
isteinbrecher: The dual shape functions can directly be implemented here.

src/geometry_pair/4C_geometry_pair_element.hpp
```diff
@@ -154,7 +165,7 @@ namespace GeometryPair
    * of the element
    */
   template <Core::FE::CellType discretization, unsigned int values_per_node,
-      unsigned int spatial_dim = 3>
+      unsigned int spatial_dim = 3, BasisType type = BasisType::standard>
```
isteinbrecher: I don't understand why we would need this additional `type`.

src/geometry_pair/4C_geometry_pair_element.hpp
```diff
@@ -188,6 +202,8 @@ namespace GeometryPair
 
   //! 1D elements
   using t_hermite = ElementDiscretizationBase<Core::FE::CellType::line2, 2>;
+  using t_hermite_dual =
```
isteinbrecher: A general point, you are now adding the dual shape functions to the geometry pair. They are not related in any way to the geometry pair. Historically we simply use the functions implemented here for the LM field, because they were the same as the primal shape functions.

For the dual shape functions this is not the case and it does not really make sense to implement them here.

The good news, since everything is templated you can simply create a header containing all the information of the dual shape functions in the beaminteraction module. I think you don'y need to change anything in this file, you can simply move all the dual specific stuff into that header.

src/geometry_pair/4C_geometry_pair_element_shape_functions.hpp
```diff
@@ -185,6 +185,58 @@ namespace GeometryPair
     }
   };
 
+  /**
+   * \brief Specialization for Hermite elements
+   */
+  template <>
+  struct EvaluateShapeFunction<t_hermite_dual>
```
isteinbrecher: This should also be in the header mentioned in the previous comment.

src/core/fem/src/general/utils/4C_fem_general_utils_fem_shapefunctions.hpp
```diff
@@ -3024,6 +3024,113 @@ namespace Core::FE
     return;
   }
 
+  /*!
+   \brief Fill a matrix of type VectorType with 1D dual hermite shape functions
+   */
+  template <class VectorType, typename NumberType1, typename NumberType2>
+  void shape_function_dual_hermite_1d(
```
isteinbrecher: These are super specific shape functions for a beam-interaction problem, I don't think we should add them to `Core`, it makes more sense to add them directly to the shape function definition, as mentioned in a later comment.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar.cpp
```diff
@@ -113,6 +113,13 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPairMortar<Beam, Solid,
   {
     // Setup variables.
     GeometryPair::ElementData<Mortar, double> element_data_lambda;
+
+    if constexpr (std::is_same_v<Mortar, GeometryPair::t_hermite_dual>)
+    {
+      element_data_lambda.shape_function_data_.ref_length_ =
+          this->ele1pos_.shape_function_data_.ref_length_;
+    }
```
isteinbrecher: This is something we don't want to do. The code here is valid for all LM Ansatz functions, we don't want an `if` for a special case here.

What you should to is something like:
```c++
const auto element_data_lambda = GeometryPair::InitializeElementData<ElementType, double>::initialize(beam_element.get());
```
You can then specialise the `InitializeElementData` function for the dual hermite template.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar.cpp
```diff
@@ -330,8 +337,19 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPairMortar<Beam, Solid, Mortar>:
...
+        mortar_shape_function_data.ref_length_ = this->ele1pos_.shape_function_data_.ref_length_;
+
+        GeometryPair::EvaluateShapeFunction<GeometryPair::t_hermite_dual>::evaluate(
+            N_mortar, projected_gauss_point.get_eta(), mortar_shape_function_data);
+      }
+      else
+      {
+        GeometryPair::EvaluateShapeFunction<Mortar>::evaluate(
+            N_mortar, projected_gauss_point.get_eta());
+      }
```
isteinbrecher: Similar to the previous comment. You can create the shape function data here
```c++
ShapeFunctionData<ElementType> shape_function_data{};
SetShapeFunctionData<ElementType>::set(shape_function_data, element);
```
This will then automatically work for all shape function types.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar_rotation.cpp
```diff
@@ -651,14 +655,23 @@ namespace BeamInteraction
   initialize_template_beam_to_solid_volume_meshtying_pair_mortar_rotation(t_line2, t_line2);
   initialize_template_beam_to_solid_volume_meshtying_pair_mortar_rotation(t_line2, t_line3);
   initialize_template_beam_to_solid_volume_meshtying_pair_mortar_rotation(t_line2, t_line4);
+  initialize_template_beam_to_solid_volume_meshtying_pair_mortar_rotation(t_line2, t_hermite_dual);
```
isteinbrecher: This shows a shortcoming of the rotational mortar pair implementation. At the moment, we always use the same shape function for positional and rotational LM - in the dual case, this will obviously not work any more. For the first PR, let's only change the pure positional coupling. Then we can see if we also want rotations and how we could achieve that.

dharinib98:
@isteinbrecher I moved the dual Hermite implementation to a separate header and addressed most of the other comments. Two points are still open: I still instantiate the rotational variants as well, since the same templated factory path is used for both positional and rotational mortar pairs. Otherwise, configurations with t_hermite_dual can still be generated and lead to linker errors.
The shape-function equations are also still defined through the FEM utility function. Moving them fully to the header complained about some array-bound issues, I am not sure although exactly why. I think both these can be cleaned up once the rotational dual Hermite path is handled properly.  Please let me know what you think about the current state of this PR.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+  template <>
+  struct EvaluateShapeFunction<BeamInteraction::t_hermite_dual>
+  {
+    template <typename V, typename T>
+    static void evaluate(V& N, const T& xi,
+        const ShapeFunctionData<BeamInteraction::t_hermite_dual>& shape_function_data)
+    {
+      Core::FE::shape_function_dual_hermite_1d(
+          N, xi, shape_function_data.ref_length_, t_hermite::discretization_);
```
isteinbrecher: You should be able to directly implement the explicit shape function calculation here.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  template <typename Mortar, typename V, typename T>
+  inline void evaluate_beam_to_solid_mortar_shape_function(
+      V& N, const T& eta, const GeometryPair::ElementData<Mortar, double>& mortar_element_data)
+  {
+    GeometryPair::EvaluateShapeFunction<Mortar>::evaluate(
+        N, eta, mortar_element_data.shape_function_data_);
+  }
+}  // namespace BeamInteraction
```
isteinbrecher: This is not ideal, you should use the existing functions from `GEOMETRYPAIR` and not reimplement them here. This should work straight away, you just have to specialise the `SetShapeFunctionData` functionality for `t_hermite_dual` in this file.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar_rotation.cpp
```diff
@@ -553,8 +556,12 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPairMortarRotation<Beam, Solid,
...
 
       // Evaluate shape functions.
-      GeometryPair::EvaluateShapeFunction<MortarRot>::evaluate(
-          lambda_shape_functions, projected_gauss_point.get_eta());
+      GeometryPair::ElementData<MortarRot, double> mortar_rot_element_data;
+      BeamInteraction::set_beam_to_solid_mortar_shape_function_data(
+          mortar_rot_element_data, this->ele1pos_);
+
+      BeamInteraction::evaluate_beam_to_solid_mortar_shape_function(
+          lambda_shape_functions, projected_gauss_point.get_eta(), mortar_rot_element_data);
```
isteinbrecher:
```suggestion
      GeometryPair::ShapeFunctionData<MortarRot> shape_function_data;
      GeometryPair::SetShapeFunctionData<MortarRot>::set(shape_function_data, this->element1());
      GeometryPair::EvaluateShapeFunction<MortarRot>::evaluate(
          lambda_shape_functions, projected_gauss_point.get_eta(), shape_function_data);
```

This should work here and I all cases where you had to touch how shape functions are evaluated.

isteinbrecher (commented):
Thanks fort the changes @dharinib98 this is already looking a lot better.

Please have a look at my comments and letzt discuss if there are any questions.

Also, can you add an input file test with the dual shape functions?

dharinib98:
@isteinbrecher @maxfirmbach @mayrmt This is ready for review again! The core unit test fails in the pipeline here, because of the unit test I added for the dual hermite shape functions. However locally these tests pass for me and I not quite sure why they are problematic here.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_input.hpp
```diff
@@ -153,7 +153,9 @@ namespace BeamToSolid
     //! Quadratic.
     line3,
     //! Cubic.
-    line4
+    line4,
+    //! dual hermite
```
maxfirmbach: Well, this comment just replicates the variable name. For documentation purposes it might make sense to elaborate maybe a bit more on the polynomial degree etc. 😉.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include "4C_fem_general_utils_fem_shapefunctions.hpp"
+#include "4C_geometry_pair_element.hpp"
+#include "4C_geometry_pair_element_shape_functions.hpp"
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace BeamInteraction
+{
+  struct HermiteDual : public GeometryPair::ElementDiscretizationBase<Core::FE::CellType::line2, 2>
```
maxfirmbach, line 22: @isteinbrecher Is this what we want here? Of course using a `struct` forces us to a specific naming scheme, which looks a bit off in the template initialization later on. I don't have a strong opinion on this, just want to know if this would make sense for the other template parameters too in the future.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,88 @@
...
+          FOUR_C_THROW("distype unknown\n");
+          break;
+      }
+
+      return;
+    }
+  };
+
+  template <>
+  struct PrintElementData<BeamInteraction::HermiteDual>
```
maxfirmbach: I don't know ... is this function well-placed here? Seems more like a utility thing and kinda hardcoded. I don't have any objections though.

maxfirmbach (commented):
@dharinib98 Looking forward to use this feature too 😉. Looks alright to me ... I just have some minor comments and questions.

maxfirmbach:
Failing tests produce this output:
```terminal
The difference between Phi(i) and Phi_ref[i] is 0.043972369034521064, which exceeds 1e-10, where
Phi(i) evaluates to -0.052766842841425188,
Phi_ref[i] evaluates to -0.0087944738069041239, and
1e-10 evaluates to 1e-10.

/__w/4C/4C/src/core/fem/tests/utils/4C_fem_general_utils_fem_shapefunctions_test.cpp:56: Failure
The difference between Phi(i) and Phi_ref[i] is 16.761378432263328, which exceeds 1e-10, where
Phi(i) evaluates to -14.366895799082851,
Phi_ref[i] evaluates to 2.3944826331804752, and
1e-10 evaluates to 1e-10.

[  FAILED  ] ElementShapeFunctionsTest.TestDualHermiteLine2 (0 ms)
[ RUN      ] ElementShapeFunctionsTest.TestDualHermiteLine2Length03
/__w/4C/4C/src/core/fem/tests/utils/4C_fem_general_utils_fem_shapefunctions_test.cpp:75: Failure
The difference between Phi(i) and Phi_ref[i] is 0.14657456344840356, which exceeds 1e-10, where
Phi(i) evaluates to -0.17588947613808398,
Phi_ref[i] evaluates to -0.029314912689680415, and
1e-10 evaluates to 1e-10.

/__w/4C/4C/src/core/fem/tests/utils/4C_fem_general_utils_fem_shapefunctions_test.cpp:75: Failure
The difference between Phi(i) and Phi_ref[i] is 55.871261440877753, which exceeds 1e-10, where
Phi(i) evaluates to -47.889652663609503,
Phi_ref[i] evaluates to 7.9816087772682511, and
1e-10 evaluates to 1e-10.
```

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
dharinib98: I added this for the sake of completeness since this function was there for the other shape functions. Maybe we don't need this. I am not sure though

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar.cpp
```diff
@@ -113,6 +114,7 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPairMortar<Beam, Solid,
   {
     // Setup variables.
     GeometryPair::ElementData<Mortar, double> element_data_lambda;
+
```
isteinbrecher: Not needed white space change.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar_rotation.cpp
isteinbrecher, line 1: No changes should be required in this file.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_conditions.cpp
```diff
@@ -376,6 +380,37 @@ BeamInteraction::create_beam_to_solid_volume_pair_mortar(const Core::FE::CellTyp
   return create_beam_to_solid_volume_pair_shape<BtsClass, BtsMortarTemplateArguments...>(shape);
 }
 
+template <template <typename...> class BtsClass, typename PosMortar>
+std::shared_ptr<BeamInteraction::BeamContactPair> create_beam_to_solid_volume_pair_mortar_rotation(
+    const Core::FE::CellType shape,
+    const BeamToSolid::BeamToSolidMortarShapefunctions mortar_shape_function_rotation)
```
isteinbrecher: I think it might be possible to do this without having to create a new factory function. Can you try if something similar to this works:
```c++

template <template <typename...> class T>

inline constexpr bool is_rotation_pair_v = false;

template <>

inline constexpr bool
    is_rotation_pair_v<BeamInteraction::BeamToSolidVolumeMeshtyingPairMortarRotation> = true;

/**
 *
 */
template <template <typename...> class BtsClass, typename... BtsMortarTemplateArguments,
    typename... BtsMortarShape>
std::shared_ptr<BeamInteraction::BeamContactPair>
BeamInteraction::create_beam_to_solid_volume_pair_mortar(const Core::FE::CellType shape,
    const BeamToSolid::BeamToSolidMortarShapefunctions mortar_shape_function,
    BtsMortarShape... other_mortar_shape_function)
{
  switch (mortar_shape_function)
  {
    case BeamToSolid::BeamToSolidMortarShapefunctions::line2:
      return create_beam_to_solid_volume_pair_mortar<BtsClass, BtsMortarTemplateArguments...,
          GeometryPair::t_line2>(shape, other_mortar_shape_function...);
    case BeamToSolid::BeamToSolidMortarShapefunctions::line3:
      return create_beam_to_solid_volume_pair_mortar<BtsClass, BtsMortarTemplateArguments...,
          GeometryPair::t_line3>(shape, other_mortar_shape_function...);
    case BeamToSolid::BeamToSolidMortarShapefunctions::line4:
      return create_beam_to_solid_volume_pair_mortar<BtsClass, BtsMortarTemplateArguments...,
          GeometryPair::t_line4>(shape, other_mortar_shape_function...);
  }

  if constexpr (not is_rotation_pair_v<BtsClass>)
  {
    return create_beam_to_solid_volume_pair_mortar<BtsClass, BtsMortarTemplateArguments...,
        BeamInteraction::HermiteDual>(shape, other_mortar_shape_function...);
  }


  FOUR_C_THROW("Wrong mortar shape function.");
  return nullptr;
}
```
The main think is that we don't want to initialize the templates for rotational coupling AND dual hermite shape functions. This would be the cleanses approach.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
isteinbrecher: I agree @maxfirmbach. I am not sure what the best way is, we could use an alias to define a name like `t_hermite_dual` after creating the struct.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  template <>
+  struct EvaluateShapeFunction<BeamInteraction::HermiteDual>
+  {
+    template <typename V, typename T>
+    static void evaluate(V& N, const T& xi,
+        const ShapeFunctionData<BeamInteraction::HermiteDual>& shape_function_data)
+    {
+      static_assert(!std::is_integral_v<T>);
```
isteinbrecher: Why do we need this?

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+      switch (t_hermite::discretization_)
```
isteinbrecher: This `switch` is not needed.

src/core/fem/tests/utils/4C_fem_general_utils_fem_shapefunctions_test.cpp
isteinbrecher, line 1: Nice that we have unites for this! But they don't belong here, since the shape functions are not implemented in `core`. If you want to keep the unit tests, lets implement them in the beaminteraction unit tests.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,88 @@
...
+      constexpr auto max_precision{std::numeric_limits<double>::digits10 + 1};
+      out << std::setprecision(max_precision);
+      out << "\nElement reference length: " << element_data.shape_function_data_.ref_length_;
+      out << "\nElement state vector: ";
+      element_data.element_position_.print(out);
+    }
+  };
+}  // namespace GeometryPair
+
+FOUR_C_NAMESPACE_CLOSE
```
isteinbrecher, line 92: What I am missing in this file is the overload of
```c++
GeometryPair::SetShapeFunctionData
```
I think you are not setting `ref_length_` correctly. Can you check that this is not some default value?

Please add a input file test (preferably with a vtu check for the mortar functions) to be sure that you don't break your implementation (later on and also during this PR).

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_conditions.cpp
dharinib98: @isteinbrecher I tried similar implementations and this still complains that the rotations for hermite dual case is not instantiated. I tried to solve this using an if statement in the dual hermite switch case. Let me know is this is okay.

isteinbrecher: That part looks good now.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
dharinib98: @isteinbrecher Will the unit tests actually be enough to check if the `ref_length_` is set correctly for now since we have two lengths being checked for in there. Also do we want to add an input file test and a vtu check already in this PR or make it a follow up PR?

dharinib98: I corrected the overload issue and actually added the tests here. Let me know if that is okay.

src/core/fem/tests/utils/4C_fem_general_utils_fem_shapefunctions_test.cpp
```diff
@@ -34,5 +34,4 @@ namespace
         -9.2684042199908543, 3.8579485013272361};
     for (std::size_t i = 0; i < ddN_ref.size(); ++i) EXPECT_NEAR(ddN(i), ddN_ref[i], 1e-10);
   }
-
```
isteinbrecher, line 37: Not needed change.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_conditions.cpp
```diff
@@ -1000,4 +1021,4 @@ BeamInteraction::BeamToLineCondition::create_contact_pair_internal(
       Core::FE::cell_type_to_string(beam_element->shape()), Core::FE::cell_type_to_string(shape));
 }
 
-FOUR_C_NAMESPACE_CLOSE
+FOUR_C_NAMESPACE_CLOSE
```
isteinbrecher: Not needed change.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar_rotation.cpp
```diff
@@ -661,4 +661,4 @@ namespace BeamInteraction
   initialize_template_beam_to_solid_volume_meshtying_pair_mortar_rotation(t_line4, t_line4);
 }  // namespace BeamInteraction
 
-FOUR_C_NAMESPACE_CLOSE
+FOUR_C_NAMESPACE_CLOSE
```
isteinbrecher: Not needed change.

src/beaminteraction/tests/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite_test.cpp
```diff
@@ -0,0 +1,54 @@
...
+    GeometryPair::ShapeFunctionData<BeamInteraction::HermiteDual> shape_function_data;
+    shape_function_data.ref_length_ = ref_length;
+
+    GeometryPair::EvaluateShapeFunction<BeamInteraction::HermiteDual>::evaluate(
+        phi, xi, shape_function_data);
+
+    const std::vector<double> phi_ref = {
+        -0.24634578918994066, -0.17588947613808397, -1.1393423701836305, -47.889652663609505};
+
+    for (std::size_t i = 0; i < phi_ref.size(); ++i) EXPECT_NEAR(phi(i), phi_ref[i], 1.0e-10);
```
isteinbrecher: Why do we need two test cases here? Is the only difference the reference length? Then let's only use the one with a reference length != 1

tests/input_files/ref/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_dual_hermite-vtk-files/beam-to-solid-volume-mortar-continuous-00001-0.vtu
isteinbrecher, line 1: Do we need to check in 2 time steps of reference vtu files? I think it should be fine to only check the one at the final time step.

tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_dual_hermite.4C.yaml
```diff
@@ -0,0 +1,580 @@
+TITLE:
+  - "Description: Simulate a beam coupled to a solid via penalty mortar coupling."
+  - "The shape functions for the Lagrange multplicators are line3."
```
isteinbrecher: This is not up to date.

isteinbrecher:
Thanks for the changes @dharinib98!

> @isteinbrecher Will the unit tests actually be enough to check if the ref_length_ is set correctly for now since we have two lengths being checked for in there. Also do we want to add an input file test and a vtu check already in this PR or make it a follow up PR?

We still need the vtu test, because we previously saw that the unit test alone is not enough. Did you verify that the vtu tests produce the correct result?

One more point that we can maybe discuss in person, if we should add a runtime test that the D matrix is actually diagonal when using the dual shape functions.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_conditions.cpp
```diff
@@ -339,6 +340,13 @@ BeamInteraction::create_beam_to_solid_volume_pair_shape_no_nurbs(const Core::FE:
   }
 }
 
+template <template <typename...> class T>
+inline constexpr bool is_rotation_pair_v = false;
+
+template <>
+inline constexpr bool
+    is_rotation_pair_v<BeamInteraction::BeamToSolidVolumeMeshtyingPairMortarRotation> = true;
+
```
isteinbrecher: Let's add some documentation what this structure does. Also, can we move this to the header file of the rotational coupling pair? It would be more fitting there.

mayrmt:
@isteinbrecher I have discussed the idea of testing for a diagonal D with @dharinib98 some time ago, though it is not clear to me, yet, how to achieve this in 4C. It probably must be through solve form of unit test...

Most importantly, I agree that this property should be tested.

dharinib98:
> @isteinbrecher I have discussed the idea of testing for a diagonal D with @dharinib98 some time ago, though it is not clear to me, yet, how to achieve this in 4C. It probably must be through solve form of unit test...
> 
> Most importantly, I agree that this property should be tested.

yeah! I am not 100 percent sure either, how this can be done or rather how this can be done correctly. We can discuss about this during the week.

tests/input_files/ref/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_dual_hermite-vtk-files/beam-to-solid-volume-mortar-continuous-00001-0.vtu
dharinib98: I think we need to check in all the time steps that are being saved, otherwise the test does not pass.

isteinbrecher: I am pretty sure that we have tests where not all time steps are checked in. Please have a look at other output tests in the repo.

c-p-schmidt: You can specify for which times it should be checked like here:
`__four_c_test_vtk(BASED_ON ${current} PVD_RESULT xxx-structure-beams.pvd PVD_REFERENCE ref/beam3r_herm2line3_backweuler_browndyn_crosslinking_beam3rline2-vtk/structure-beams.pvd TOLERANCE 1e-08 TIME_STEPS 0.0025 0.0085)`

tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_dual_hermite.4C.yaml
```diff
@@ -0,0 +1,581 @@
...
+  DISPLACEMENT: true
+  ELEMENT_OWNER: true
+BEAM INTERACTION:
+  REPARTITIONSTRATEGY: "Everydt"
+BEAM INTERACTION/BEAM TO SOLID VOLUME MESHTYING:
+  CONTACT_DISCRETIZATION: mortar
+  CONSTRAINT_STRATEGY: lagrange
+  LAGRANGE_FORMULATION: saddlepoint
+  MORTAR_SHAPE_FUNCTION: dual_hermite
+  PENALTY_PARAMETER: 10
```
isteinbrecher: Not needed

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_manager.cpp
```diff
@@ -548,6 +548,43 @@ void BeamInteraction::BeamToSolidMortarManager::evaluate_and_assemble_global_cou
...
+      const int num_entries = constraint_lin_beam_->num_global_entries(lambda_gid);
+
+      std::vector<double> values(num_entries);
+      std::vector<int> col_gids(num_entries);
+
+      int num_extracted = 0;
+
+      constraint_lin_beam_->extract_global_row_copy(
+          lambda_gid, num_entries, num_extracted, values.data(), col_gids.data());
+
```
mayrmt: Check if `num_entries == 1` here. If yes, we are done and do not have to do any further checks.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_manager.cpp
```diff
@@ -548,6 +548,43 @@ void BeamInteraction::BeamToSolidMortarManager::evaluate_and_assemble_global_cou
...
+      const int lambda_gid = lambda_dof_rowmap_->gid(lid);
+
+      const int num_entries = constraint_lin_beam_->num_global_entries(lambda_gid);
+
+      std::vector<double> values(num_entries);
+      std::vector<int> col_gids(num_entries);
+
+      int num_extracted = 0;
+
+      constraint_lin_beam_->extract_global_row_copy(
```
mayrmt: Since we only want to read entries, it should be sufficient to extract a view instead of a copy.

mayrmt: Since we never use the pair of global row and column indices here, we do not need to operate in global IDs. It should be sufficient to work with local IDs directly.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar_rotation.hpp
```diff
@@ -127,6 +127,20 @@ namespace BeamInteraction
...
+  template <template <typename...> class T>
+  inline constexpr bool is_rotation_pair_v = false;
+
+  /**
+   * \brief Mark the rotational mortar pair implementation as a rotational coupling pair.
+   */
+  template <>
+  inline constexpr bool
+      is_rotation_pair_v<BeamInteraction::BeamToSolidVolumeMeshtyingPairMortarRotation> = true;
+
```
isteinbrecher, line 140:
```suggestion
  /**
   * \brief Compile time flag whether a beam-to-solid volume mortar pair includes rotational coupling
   */
  template <template <typename...> class T>
  inline constexpr bool is_rotation_pair_v = false;
  template <>
  inline constexpr bool
      is_rotation_pair_v<BeamInteraction::BeamToSolidVolumeMeshtyingPairMortarRotation> = true;

```

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_manager.cpp
```diff
@@ -548,6 +548,43 @@ void BeamInteraction::BeamToSolidMortarManager::evaluate_and_assemble_global_cou
   kappa_->complete();
   lambda_active_->complete();
   constraint_->complete();
+
+  if (parameters_.mortar_shape_function ==
+      BeamToSolid::BeamToSolidMortarShapefunctions::dual_hermite)
```
isteinbrecher: Did you check if this is actually ever executed?

I think this is never executed at the moment, because `parameters_.mortar_shape_function` is always the default value.

See also my comment below.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_manager.cpp
```diff
@@ -548,6 +548,43 @@ void BeamInteraction::BeamToSolidMortarManager::evaluate_and_assemble_global_cou
...
+  {
+    const double tol = 1.0e-13;
+
+    for (int lid = 0; lid < lambda_dof_rowmap_->num_my_elements(); ++lid)
+    {
+      const int lambda_gid = lambda_dof_rowmap_->gid(lid);
+
+      const int num_entries = constraint_lin_beam_->num_global_entries(lambda_gid);
```
isteinbrecher: You check this for `constraint_lin_beam_`. However, what is with `constraint_beam_lin_`? This also has to be checked.

What you can do is create a utility function that checks a matrix for diagonal-like structure (@mayrmt should we move such a thing directly into core or leave if here for now?) and then call it here like:


```c++
if (parameters_.check_diagonal_d_matrix)
{
  check_diagonal_like_structure(constraint_lin_beam_);
  check_diagonal_like_structure(constraint_beam_lin_);
}
```

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_manager.hpp
```diff
@@ -77,6 +77,10 @@ namespace BeamInteraction
     BeamToSolid::BeamToSolidLagrangeFormulation lagrange_formulation =
         BeamToSolid::BeamToSolidLagrangeFormulation::none;
 
+    //! Mortar shape function used for beam-to-solid coupling.
+    BeamToSolid::BeamToSolidMortarShapefunctions mortar_shape_function =
+        BeamToSolid::BeamToSolidMortarShapefunctions::none;
```
isteinbrecher: I would not pass the shape function here, but instead a flag like `check_diagonal_d_matrix` that indicates if the relevant coupling matrices should be checked for diagonal-like structure.

Right now, this variable always has its default value. You need to make sure that this variable is set where needed. Likely in:
```c++
std::shared_ptr<BeamInteraction::SubmodelEvaluator::BeamContactAssemblyManager>
BeamInteraction::BeamToSolidCondition::create_indirect_assembly_manager(
    const std::shared_ptr<const Core::FE::Discretization>& discret)
```

src/core/fem/tests/utils/4C_fem_general_utils_fem_shapefunctions_test.cpp
```diff
@@ -35,4 +35,4 @@ namespace
     for (std::size_t i = 0; i < ddN_ref.size(); ++i) EXPECT_NEAR(ddN(i), ddN_ref[i], 1e-10);
   }
 
-}  // namespace
```
isteinbrecher, line 38: Not needed change.

src/beaminteraction/tests/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite_test.cpp
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+}  // namespace
```
isteinbrecher: Missing new line at the end.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_manager.cpp
mayrmt: In order to move it to core, I think it actually needs to check for diagonal entries, not only for "maximum one entry per row". If we achieve that, we should move it to core. If the final results is a method of the matrix itself, that i can query, e.g.

```cxx
bool matrix_is_diagonal = my_matrix.is_diagonal();
```

then it should move into the matrix wrapper of course.

mayrmt:
@isteinbrecher @dharinib98 A basic question: if we use dual shape functions and we have ensured through unit testing, that the implementation actually satisfies the biorthogonality conditions, why can't we exploit that when assembling `D`? We know, that there is only one value, so it should be sufficient to compute just this value and put it into the correct location in the matrix. Then, it will be "diagonal" by construction and we do not have to test, if it is diagonal. Furthermore, this avoids to compute many matrix entries, that we expect to evaluate to zero anyways.

What are your thoughts on this?

dharinib98:
> @isteinbrecher @dharinib98 A basic question: if we use dual shape functions and we have ensured through unit testing, that the implementation actually satisfies the biorthogonality conditions, why can't we exploit that when assembling `D`? We know, that there is only one value, so it should be sufficient to compute just this value and put it into the correct location in the matrix. Then, it will be "diagonal" by construction and we do not have to test, if it is diagonal. Furthermore, this avoids to compute many matrix entries, that we expect to evaluate to zero anyways.
> 
> What are your thoughts on this?

@mayrmt The main issue with this would be (since we only deal with positional coupling and not full coupling as yet) to put the computed values into the correct location, since accessing the positional dofs and the lambda translation dofs are not easy with the current tools available. Another personal concern,  might be that we would be needing a lot of extra lines of code just for this special case and might make the the current assembly structure that is in place in beaminteraction look a bit messy.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_manager.cpp
dharinib98: Since the current implementation does not qualify for moving it to core, I would leave it here.

isteinbrecher: Then let's put it into `4C_beaminteraction_contact_beam_to_solid_utils`

isteinbrecher (commented):
Thanks for addressing all the points @dharinib98 . From my side only two things are missing, the move of the diagonal check to the beaminteratcion utils and the unittest.

isteinbrecher:
> @isteinbrecher @dharinib98 A basic question: if we use dual shape functions and we have ensured through unit testing, that the implementation actually satisfies the biorthogonality conditions, why can't we exploit that when assembling D? We know, that there is only one value, so it should be sufficient to compute just this value and put it into the correct location in the matrix. Then, it will be "diagonal" by construction and we do not have to test, if it is diagonal. Furthermore, this avoids to compute many matrix entries, that we expect to evaluate to zero anyways.

In theory we could do that. However, with the current code structure, we are able to combine all the different cases into one general implementation. I suspect the overhead of computing the zero entries negligible. Also, if at some point we will want to consider beams that are not fully embedded and then we will need integration again, also for dual shape functions.

dharinib98:
@isteinbrecher @mayrmt The current version is good enough for straight beams, both Euler-Bernoulli and Simo-Reissner, and should be ready for merging.
However, I have also implemented the equations from Alex’s thesis that we discussed as a possible solution for curved beams. With this implementation, I do obtain a diagonal (D), but the code is currently still in a very draft state and will require some additional time before it can be merged.
So now the question is should we go ahead with this merge or wait for the additional implementation as well?

isteinbrecher (approved):
Looks good to me, thanks for the extra effort @dharinib98.

> So now the question is should we go ahead with this merge or wait for the additional implementation as well?

Definitely merge now, let's do the curved beams in a follow up PR.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+      N(0) = -35.0 / 4.0 * xi3 + 15.0 / 4.0 * xi2 + 15.0 / 4.0 * xi - 3.0 / 4.0;
+
+      N(1) = (105.0 * xi3 - 45.0 / 2.0 * xi2 - 60.0 * xi + 15.0 / 2.0) /
+             shape_function_data.ref_length_;
+
+      N(2) = 35.0 / 4.0 * xi3 + 15.0 / 4.0 * xi2 - 15.0 / 4.0 * xi - 3.0 / 4.0;
+
+      N(3) = (105.0 * xi3 + 45.0 / 2.0 * xi2 - 60.0 * xi - 15.0 / 2.0) /
+             shape_function_data.ref_length_;
```
mayrmt, line 72:
```suggestion
      N(0) = -8.75 * xi3 + 3.75 * (xi2 + xi) - 0.75;

      N(1) = (105.0 * xi3 - 22.5 * xi2 - 60.0 * xi + 7.5) /
             shape_function_data.ref_length_;

      N(2) = 8.75 * xi3 + 3.75 * (xi2 - xi) - 0.75;

      N(3) = (105.0 * xi3 + 22.5 * xi2 - 60.0 * xi - 7.5) /
             shape_function_data.ref_length_;
```

mayrmt (approved):
Just an idea to simplify the mathematical expressions and save some floating point operations.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_mortar_shape_functions_dual_hermite.hpp
isteinbrecher: I would keep the fractions, makes it easier to check with manual derivations. The compiler will fill in the evaluated fractions anyway.

mayrmt: Ok, then let's merge this!

PR #2010, opened 2026-04-29
Fix distance threshold bug ScalarInterp

dragos-ana:
Fixed calculation of normalized weights within the scalar interpolation framework, if the interpolation location is within a distance threshold wrt a reference location.
In that case, the weight of that specific reference point is directly set to 1.0 AND the other weights are reset to 0.0. The previous approach also enabled multiple 1.0 weights, and only the first detected 1.0 was kept.
Unit tests were added to test the bugfix; these tests would have previously failed because the weights would have always been {1.0, 0.0} regardless of the interpolation location 0.0 or 1.0.

### DIFF
--- a/src/core/linalg/tests/4C_linalg_utils_scalar_interpolation_test.cpp
+++ b/src/core/linalg/tests/4C_linalg_utils_scalar_interpolation_test.cpp
@@ -28,7 +28,7 @@ namespace
   {
     using namespace Core::LinAlg;
 
-    // 1D
+    // Setup dimension of interpolation locations
     constexpr unsigned int loc_dim = 3;
 
     // Reference locations: -1.0, 1.0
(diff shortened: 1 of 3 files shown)


src/core/linalg/tests/4C_linalg_utils_scalar_interpolation_test.cpp
```diff
@@ -64,6 +64,58 @@ namespace
     for (size_t i = 0; i < weights.size(); ++i) ASSERT_NEAR(weights[i], weights_ref[i], 1e-8);
   }
 
+
+  TEST(LinalgScalarInterpolationTest, CalculateNormalizedWeights1DInverseDistanceAtRefLocations)
+  {
+    using namespace Core::LinAlg;
+
+    // 1D
```
c-p-schmidt: Leftover from development?

src/core/linalg/tests/4C_linalg_utils_scalar_interpolation_test.cpp
```diff
@@ -109,6 +161,57 @@ namespace
     for (size_t i = 0; i < weights.size(); ++i) ASSERT_NEAR(weights[i], weights_ref[i], 1e-8);
   }
 
+
+  TEST(LinalgScalarInterpolationTest, CalculateNormalizedWeights1DExponentialAtRefLocations)
+  {
+    using namespace Core::LinAlg;
+
+    // 1D
```
c-p-schmidt: Same as above.

c-p-schmidt (dismissed):

c-p-schmidt (approved):

src/core/linalg/tests/4C_linalg_utils_scalar_interpolation_test.cpp
dragos-ana: Not really, I copied the same framework from the other 1D tests. 
But I agree that it is not explicit / misplaced, and have therefore adapted the comment across all 1D tests.

bgoderbauer (approved):

PR #2015, opened 2026-05-05
Prepare viscoplastic factor for thermo-coupling

rjoussen:
## Description and Context

This prepares making the viscoplastic factor, and hence the multiplicative split material, temperature-dependent. 

The changes in this PR are pure refactoring and renaming, namely:

- Refactor `return_mapping` to return both the inelastic defgrad and the plastic strain, in a `HistoryVariables` struct
- `evaluate_od_stiff_mat` gets the reduced deformation gradient (reduced by preceding factors)
- Rename Mandel stress `Me` to thermo-elastic Mandel stress `Mtheta`. Note: the transversly isotropic contributions by the transvisotropic **elastic** summands retain their e-subscript, since the consistent thermo-coupling will (for now) not be supported for anisotropic elastic summands. Additional contributions due to the transversely isotropic yield criterion are renamed.
- Introduce helper for the defgrad difference norm -> will be reused to check the last evaluated solid state against the incoming thermo state.
- Make `NpV` either stress or strain form more explicitly
- Move `StressFactors` to the service file
- Rename and refactor the setup_local_newton_material testing utility function to allow to define the viscoplastic factor with different parameters more easily.

### DIFF
--- a/src/mat/4C_mat_inelastic_defgrad_factors.cpp
+++ b/src/mat/4C_mat_inelastic_defgrad_factors.cpp
@@ -2560,7 +2560,7 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
             ViscoplastUtils::LinearizationType::perturbation_based ||
         err_status != ViscoplastUtils::ErrorType::no_errors)
     {
-      evaluate_additional_cmat_perturb_based(FredM, cmatadd, iFin_other, dSdiFinj);
+      evaluate_additional_cmat_perturb_based(FredM, cmatadd, dSdiFinj);
 
       return;
     }
@@ -2581,7 +2581,7 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
 
     if (err_status != ViscoplastUtils::ErrorType::no_errors)
     {
-      evaluate_additional_cmat_perturb_based(FredM, cmatadd, iFin_other, dSdiFinj);
+      evaluate_additional_cmat_perturb_based(FredM, cmatadd, dSdiFinj);
 
       return;
     }
@@ -2590,7 +2590,7 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
     if (abs(jacMat.determinant()) < 1.0e-10)
     {
       err_status = ViscoplastUtils::ErrorType::singular_jacobian;
-      evaluate_additional_cmat_perturb_based(FredM, cmatadd, iFin_other, dSdiFinj);
+      evaluate_additional_cmat_perturb_based(FredM, cmatadd, dSdiFinj);
 
       return;
     }
@@ -2645,7 +2645,7 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
     if ((err != 0) || (err2 != 0))
     {
       err_status = ViscoplastUtils::ErrorType::failed_solution_analytic_linearization;
-      evaluate_additional_cmat_perturb_based(FredM, cmatadd, iFin_other, dSdiFinj);
+      evaluate_additional_cmat_perturb_based(FredM, cmatadd, dSdiFinj);
       return;
     }
 
@@ -3648,7 +3648,7 @@ bool Mat::InelasticDefgradTransvIsotropElastViscoplast::halve_and_prepare_new_su
 
 void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat_perturb_based(
     const Core::LinAlg::Matrix<3, 3>& FredM, Core::LinAlg::Matrix<6, 6>& cmatadd,
-    const Core::LinAlg::Matrix<3, 3>& iFin_other, const Core::LinAlg::Matrix<6, 9>& dSdiFinj)
+    const Core::LinAlg::Matrix<6, 9>& dSdiFinj)
 {
   // ----- FD-based linearization ----- //
   // approximation using perturbations of the right Cauchy-Green deformation tensor,
@@ -3676,7 +3676,6 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
 
   // declare perturbed variables
   Core::LinAlg::Matrix<3, 3> perturbed_FM(Core::LinAlg::Initialization::zero);
-  Core::LinAlg::Matrix<3, 3>* pointer_perturbed_FM = &perturbed_FM;
   Core::LinAlg::Matrix<3, 3> perturbed_CM(Core::LinAlg::Initialization::zero);
   Core::LinAlg::Matrix<3, 3> perturbed_iFinM(Core::LinAlg::Initialization::zero);
 
@@ -3706,7 +3705,7 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
     perturbed_CM.multiply_tn(1.0, perturbed_FM, perturbed_FM, 0.0);
 
     // get corresponding inverse inelastic defgrad
-    evaluate_inverse_inelastic_def_grad(pointer_perturbed_FM, iFin_other, perturbed_iFinM);
+    perturbed_iFinM = return_mapping(perturbed_FM).inv_plastic_defgrad;
     Core::LinAlg::Matrix<9, 1> perturbed_iFinV(Core::LinAlg::Initialization::zero);
     Core::LinAlg::Voigt::matrix_3x3_to_9x1(perturbed_iFinM, perturbed_iFinV);
 
--- a/src/mat/4C_mat_inelastic_defgrad_factors.cpp
+++ b/src/mat/4C_mat_inelastic_defgrad_factors.cpp
@@ -52,6 +52,14 @@ namespace
   static ViscoplastUtils::ConstNonMatTensors const_non_mat_tensors =
       ViscoplastUtils::ConstNonMatTensors::instance();
 
+  double defgrad_difference_norm(
+      const Core::LinAlg::Matrix<3, 3>& lhs, const Core::LinAlg::Matrix<3, 3>& rhs)
+  {
+    Core::LinAlg::Matrix<3, 3> diff{Core::LinAlg::Initialization::zero};
+    diff.update(1.0, lhs, -1.0, rhs, 0.0);
+    return diff.norm2();
+  }
+
   // read input parameter container of parent material (i.e., underlying
   // multiplicative_split_defgrad_elasthyper material)
   Core::IO::InputParameterContainer get_parameters_of_parent_material(const int mat_id,
@@ -972,8 +980,9 @@ void Mat::InelasticDefgradLinScalarIso::evaluate_additional_cmat(
 /*--------------------------------------------------------------------*
  *--------------------------------------------------------------------*/
 void Mat::InelasticDefgradLinScalarIso::evaluate_od_stiff_mat(
-    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFinjM,
-    const Core::LinAlg::Matrix<6, 9>& dSdiFinj, Core::LinAlg::Matrix<6, 1>& dstressdc)
+    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFin_other,
+    const Core::LinAlg::Matrix<3, 3>& iFinjM, const Core::LinAlg::Matrix<6, 9>& dSdiFinj,
+    Core::LinAlg::Matrix<6, 1>& dstressdc)
 {
   static Core::LinAlg::Matrix<9, 1> id9x1(Core::LinAlg::Initialization::zero);
   // prepare id9x1 (identity matrix written as a 9x1 vector)
@@ -1098,8 +1107,9 @@ void Mat::InelasticDefgradLinScalarAniso::evaluate_additional_cmat(
 /*--------------------------------------------------------------------*
  *--------------------------------------------------------------------*/
 void Mat::InelasticDefgradLinScalarAniso::evaluate_od_stiff_mat(
-    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFinjM,
-    const Core::LinAlg::Matrix<6, 9>& dSdiFinj, Core::LinAlg::Matrix<6, 1>& dstressdc)
+    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFin_other,
+    const Core::LinAlg::Matrix<3, 3>& iFinjM, const Core::LinAlg::Matrix<6, 9>& dSdiFinj,
+    Core::LinAlg::Matrix<6, 1>& dstressdc)
 {
   // static variables
   static Core::LinAlg::Matrix<3, 3> tmp(Core::LinAlg::Initialization::zero);
@@ -1204,8 +1214,9 @@ void Mat::InelasticDefgradPolyIntercalFracIso::evaluate_additional_cmat(
 /*--------------------------------------------------------------------*
  *--------------------------------------------------------------------*/
 void Mat::InelasticDefgradPolyIntercalFracIso::evaluate_od_stiff_mat(
-    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFinjM,
-    const Core::LinAlg::Matrix<6, 9>& dSdiFinj, Core::LinAlg::Matrix<6, 1>& dstressdc)
+    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFin_other,
+    const Core::LinAlg::Matrix<3, 3>& iFinjM, const Core::LinAlg::Matrix<6, 9>& dSdiFinj,
+    Core::LinAlg::Matrix<6, 1>& dstressdc)
 {
   static Core::LinAlg::Matrix<9, 1> id9x1(Core::LinAlg::Initialization::zero);
   // prepare id9x1 (identity matrix written as a 9x1 vector)
@@ -1343,8 +1354,9 @@ void Mat::InelasticDefgradPolyIntercalFracAniso::evaluate_additional_cmat(
 /*--------------------------------------------------------------------*
  *--------------------------------------------------------------------*/
 void Mat::InelasticDefgradPolyIntercalFracAniso::evaluate_od_stiff_mat(
-    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFinjM,
-    const Core::LinAlg::Matrix<6, 9>& dSdiFinj, Core::LinAlg::Matrix<6, 1>& dstressdc)
+    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFin_other,
+    const Core::LinAlg::Matrix<3, 3>& iFinjM, const Core::LinAlg::Matrix<6, 9>& dSdiFinj,
+    Core::LinAlg::Matrix<6, 1>& dstressdc)
 {
   // static variables
   static Core::LinAlg::Matrix<3, 3> tmp(Core::LinAlg::Initialization::zero);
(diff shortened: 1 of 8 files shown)


src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2177,44 +2187,42 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantity_deriv
 
   // compute the relevant derivatives of the symmetric part of the Mandel stress
 
-  // \f$ \frac{\partial \boldsymbol{M}^{\text{e}}_{\text{sym}} }{\partial
-  // \boldsymbol{F}^{\text{in}^{-1}}_{}} \f$ (Voigt stress-form)
```
c-p-schmidt, line 2181: I think it would be valuable to keep the information that the rows are stored in Voigt stress-form.
Even better would be obviously to use tensors directly, but that is out of the scope of this PR.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.cpp
```diff
@@ -1050,8 +1056,10 @@ void Mat::MultiplicativeSplitDefgradElastHyper::evaluate_od_stiff_mat(PAR::Inela
         Core::LinAlg::FourTensorOperations::add_non_symmetric_product(
             1.0, producta, productb, diFindiFinj);
         dSdiFinj.multiply(1.0, dSdiFin, diFindiFinj, 0.0);
+
+        defgrad_reduced.multiply_nn(1.0, *defgrad, productb, 0.0);
         facdefgradin[i].second->evaluate_od_stiff_mat(
-            defgrad, iFinjM[i].second, dSdiFinj, dstressdx);
+            &defgrad_reduced, iFinjM[i].second, dSdiFinj, dstressdx);
```
c-p-schmidt: This change is problematic since inelastic defgrad factors like `InelasticDefgradLinScalarIso` and `InelasticDefgradLinScalarAniso` rely on the fact that the complete defgrad is provided as an explicit detF is in the formulation.

So if you need the reduced defgrad, we need to change the interface and add this parameter.

It just did not show up in the test suite, as these are normally only used "alone", as we have not yet combined them with temperature effects or plasticity effects.

And one more question. You only modified that for `evaluate_od_stiff_mat`, so you only need that for the linearization?

rjoussen: I overlooked that dependence because of the double inheritance of these factors I assume.

I will revert it to the original interface. I thought all factors consume the reduced defgrad only anyway, thats why I changed it. The reduced defgrad can be calculated already with defgrad*iFin_other, thats how it is done in the other evaluate functions.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: I agree with the tensors, that would make a lot of the code more readable!

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.cpp
rjoussen: Okay, nevermind, the reason it was changed is that in contrast to the other evaluate functions, evaluate_od_stiff_mat doesnt carry iFin_other yet. So I do need to change the interface, but then it will be in line with evaluate_additional_cmat, for instance, at least.

c-p-schmidt (dismissed):

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -381,7 +381,7 @@ namespace Mat
       Core::LinAlg::Matrix<6, 6> curr_dSedCe{Core::LinAlg::Initialization::zero};
 
       //! deviatoric, symmetric part of the Mandel stress tensor
-      Core::LinAlg::Matrix<3, 3> curr_Me_dev_sym_M{Core::LinAlg::Initialization::zero};
+      Core::LinAlg::Matrix<3, 3> curr_Mtheta_dev_sym_M{Core::LinAlg::Initialization::zero};
```
dragos-ana: Although we (you, me, @c-p-schmidt) definitely know what theta means in this context, I think it is not directly apparent that it refers to the thermo-elastic part of the specific tensors. 
We could improve this by adding it as a hint to the documentation of the tensors, or by calling Mtheta maybe M_thermoelast or something like this (although the variable names will become very large then, which is currently a problem of the framework anyway...).

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -1098,8 +1107,9 @@ void Mat::InelasticDefgradLinScalarAniso::evaluate_additional_cmat(
 /*--------------------------------------------------------------------*
  *--------------------------------------------------------------------*/
 void Mat::InelasticDefgradLinScalarAniso::evaluate_od_stiff_mat(
-    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFinjM,
-    const Core::LinAlg::Matrix<6, 9>& dSdiFinj, Core::LinAlg::Matrix<6, 1>& dstressdc)
+    const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<3, 3>& iFin_other,
+    const Core::LinAlg::Matrix<3, 3>& iFinjM, const Core::LinAlg::Matrix<6, 9>& dSdiFinj,
+    Core::LinAlg::Matrix<6, 1>& dstressdc)
```
dragos-ana, line 1112: Not part of this PR for sure, but we should maybe do something at some point for the easily-swappable adjacent parameters iFin_other and iFinjM, e.g. a struct, which may also include the defgrad.
I am guilty of this myself, having introduced it in the evaluation routines of the viscoplastic material.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2636,6 +2663,16 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+{
+  const HistoryVariables history_variables = evaluate_history_variables(defgrad, iFin_other);
+  iFinM = history_variables.inv_plastic_defgrad;
+}
+
+/*--------------------------------------------------------------------*
+ *--------------------------------------------------------------------*/
+Mat::InelasticDefgradTransvIsotropElastViscoplast::HistoryVariables
+Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_history_variables(
+    const Core::LinAlg::Matrix<3, 3>* defgrad, const Core::LinAlg::Matrix<3, 3>& iFin_other)
```
dragos-ana: This is a very nice way to use the computed information effectively (i.e., the plastic strain as well)!

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2646,42 +2683,32 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_inverse_inelast
...
 }
 
 /*--------------------------------------------------------------------*
  *--------------------------------------------------------------------*/
-Core::LinAlg::Matrix<3, 3> Mat::InelasticDefgradTransvIsotropElastViscoplast::return_mapping(
+Mat::InelasticDefgradTransvIsotropElastViscoplast::HistoryVariables
+Mat::InelasticDefgradTransvIsotropElastViscoplast::return_mapping(
     const Core::LinAlg::Matrix<3, 3>& FredM)
 {
   // declare output: inverse inelastic deformation gradient (after return mapping)
```
dragos-ana:
```suggestion
  // declare output: history variables (after return mapping)
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2246,69 +2261,74 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantity_deriv
...
+  \boldsymbol{C}^{}_{}} \f$ (Voigt stress-stress form)
+  */
+  state_quantity_derivatives.curr_dMtheta_dev_sym_dC.multiply_nn(
+      1.0, const_non_mat_tensors.dev_op, dMtheta_sym_dC, 0.0);
 
   // plastic flow direction in Voigt strain notation
-  Core::LinAlg::Matrix<6, 1> NpV(Core::LinAlg::Initialization::zero);
-  Core::LinAlg::Voigt::Strains::matrix_to_vector(NpM, NpV);
+  Core::LinAlg::Matrix<6, 1> NpVstrainform(Core::LinAlg::Initialization::zero);
+  Core::LinAlg::Voigt::Strains::matrix_to_vector(NpM, NpVstrainform);
```
dragos-ana, line 2316: Thanks for adding the stress-strain formulations, which makes the subsequent operations much easier to understand

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -1968,32 +1982,32 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
...
   {
     // determine required components for the computation of the plastic flow direction
-    Core::LinAlg::Matrix<3, 1> Me_dev_sym_m(Core::LinAlg::Initialization::zero);
-    Me_dev_sym_m.multiply_nn(1.0, state_quantities.curr_Me_dev_sym_M, m_, 0.0);
-    Core::LinAlg::Matrix<3, 3> m_dyad_Me_dev_sym_m(Core::LinAlg::Initialization::zero);
-    m_dyad_Me_dev_sym_m.multiply_nt(1.0, m_, Me_dev_sym_m, 0.0);
-    Core::LinAlg::Matrix<3, 3> Me_dev_sym_A_dyad_A(Core::LinAlg::Initialization::zero);
-    Me_dev_sym_A_dyad_A.multiply_nt(1.0, Me_dev_sym_m, m_, 0.0);
+    Core::LinAlg::Matrix<3, 1> Mtheta_dev_sym_m(Core::LinAlg::Initialization::zero);
+    Mtheta_dev_sym_m.multiply_nn(1.0, state_quantities.curr_Mtheta_dev_sym_M, m_, 0.0);
```
dragos-ana, line 1986: I am a bit confused by the description of the PR where it says that the transversely-isotropic contributions retain their _e index instead of theta. 
Did I miss anything?

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -1193,18 +1206,22 @@ namespace
...
         std::shared_ptr<Mat::PAR::InelasticDefgradTransvIsotropElastViscoplast>&
             local_newton_material_params,
-        std::shared_ptr<Mat::InelasticDefgradTransvIsotropElastViscoplast>& local_newton_material,
-        const bool use_substepping = false, const unsigned int max_substepping_halve_num = 0)
+        const bool use_substepping = false, const unsigned int max_substepping_halve_num = 0,
+        const Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LinearizationType
+            linearization_type =
+                Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LinearizationType::analytic,
+        const std::optional<ReformulatedJohnsonCookParameters>& viscoplastic_law_params =
+            std::nullopt)
```
dragos-ana: Nice addition, this will make testing within this file easier for sure.
I think that we can make `max_substepping_halve_num` optional and remove `use_substepping` to reduce the number of parameters.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -1340,6 +1386,9 @@ namespace
     std::shared_ptr<Mat::InelasticDefgradTransvIsotropElastViscoplast> isotrop_vplast_refJC_;
     // pointer to InelasticDefgradTransvIsotropElastViscoplast (isotropic, logarithmic substepping)
     std::shared_ptr<Mat::InelasticDefgradTransvIsotropElastViscoplast> isotrop_elast_viscoplast_;
+    // pointer to InelasticDefgradTransvIsotropElastViscoplast (isotropic, logarithmic substepping,
+    // reformulated Johnson-Cook thermo-viscoplasticity)
+    std::shared_ptr<Mat::InelasticDefgradTransvIsotropElastViscoplast> isotrop_thermo_vplast_refJC_;
```
dragos-ana: Not used yet, but will for sure be used in the future ;)

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -1365,6 +1414,8 @@ namespace
     Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::StateQuantityDerivatives
         state_quantity_derivatives_solution_isotrop_;
 
+    std::vector<std::shared_ptr<Core::Mat::PAR::Parameter>> owned_viscoplastic_law_params_;
```
dragos-ana: Just a question: is this vector purely for lifetime management i.e., keeping the parameter objects alive for the duration of the test fixture, or will it provide any other significant benefits in your future thermo-viscoplastic tests? Or would it make sense to always create the viscoplastic law in the place it is used instead?

dragos-ana (commented):
Thanks for the nice refactoring in preparation for the thermo enhancement of the viscoplastic material. 
I left some smaller comments / suggestions.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
rjoussen: Ill add a hint to its documentation. It probably should be called `thermoelastic_symmetric_deviatoric_mandel_stress`, the `curr_` prefix and `_M` matrix indicator can probably be dropped... But I think a larger renaming makes only sense if we then also switch to tensors.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: I'm somewhat in favor of @c-p-schmidt's suggestion to pass the reduced defgrad, as well as the determinant of the full defgrad.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: As far as I understand, there are two distinct sources of anisotropy: The anisotropic yield criterion and the anisotropic elastic summands (`potsumel_transviso` I think). Only if there are such anisotropic elastic summands, I won't account for temperature yet, since this is kind of complicated and seems not really necessary at this point. The anisotropic behavior through the yield criterion, which is what we see here, will be covered.

TLDR: I will only throw if there are anisotropic elastic summands, but not if the material is anisotropic.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
rjoussen: Yes, I will test the analytic linearizations against perturbation-based linearizations in the unittests.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
rjoussen: I am actually not sure; the whole test setup is still a bit of a mess because I didn't wanna change too much in the existing tests, yet add more functionality. I will double-check if this could be simplified further.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Yes, I understand that. But I think that the current PR description is misleading, since there is no index _e  involved in the code; I mean this point specifically:
`Rename Mandel stress Me to thermo-elastic Mandel stress Mtheta. Note: the transversly isotropic contributions by the transvisotropic elastic summands retain their e-subscript, since the consistent thermo-coupling will (for now) not be supported for anisotropic elastic summands.`

Internally they are of course elastic, and not thermo-elastic, since you will throw after detecting transv.-iso. elastic summands. Maybe that note is best removed for now entirely, since this is confusing even for us, and you don't yet enforce the condition on thermo-elasticity anyway.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Yes, I missed that suggestion, but I find it better than an additional struct

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: Yeah, except for that, I realized that now adds an unnecessary layer.  I am going to remove `evaluate_history_variables` again, since other material routines can simply call `return_mapping` directly and still receive the plastic strain with the new return type. The stuff that previously happened in `evaluate_inverse_inelastic_def_grad` (compute `FredM` and check for already available state will not be necessary from other material routines.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -1872,25 +1885,25 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
...
       state_quantities.curr_gamma(1), CeCeM, 0.0);
-  Me_sym_M.update(state_quantities.curr_gamma(2), const_non_mat_tensors.id3x3, 1.0);
+  Mtheta_sym_M.update(state_quantities.curr_gamma(2), const_non_mat_tensors.id3x3, 1.0);
   if (parameter()->mat_behavior() == ViscoplastUtils::MatBehavior::transv_isotropic)
   {
     Core::LinAlg::Matrix<3, 3> addMeM(Core::LinAlg::Initialization::zero);
     temp3x3.multiply_nn(1.0, state_quantities.curr_CeM, state_quantities.curr_SeM, 0.0);
     addMeM.update(1.0 / 2.0, temp3x3, 0.0);
     temp3x3.multiply_tn(1.0, state_quantities.curr_SeM, state_quantities.curr_CeM, 0.0);
     addMeM.update(1.0 / 2.0, temp3x3, 1.0);
```
rjoussen, line 1898: This block shows an example where I decided to keep the `_e` subscript @dragos-ana. But I am open to change that still

rjoussen:
The first commit now incorporates the smaller revisions. 

In the second commit, I am using the private `return_mapping` instead of the public `evaluate_inverse_inelastic_defgrad` for the perturbation-based linearization. While doing that, I noticed an error related to the `iFin_other` usage that I pointed out in a separate comment.

The third commit restructures the factors test further; Setting up the viscoplastic factor now always goes through the new `set_up_viscoplastic_material` which removes a lot of duplicate code in the Test `SetUp()` and is easier customizable.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3706,7 +3705,7 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
     perturbed_CM.multiply_tn(1.0, perturbed_FM, perturbed_FM, 0.0);
 
     // get corresponding inverse inelastic defgrad
-    evaluate_inverse_inelastic_def_grad(pointer_perturbed_FM, iFin_other, perturbed_iFinM);
```
rjoussen, line 3693: So far, we perturbed the reduced defgrad here (as can be seen by the function signature in L3650). Yet, the actual `iFin_other` coming from `evaluate_additional_cmat` was passed here and used again for the reduced defgrad calculation in `evaluate_inverse_inelastic_defgrad`. With the new approach, we use the perturbed reduced defgrad directly for `return_mapping`. This function no longer needs `iFin_other`, it is hence removed.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: It's fine for me now, maybe this can be further clarified once the thermo-coupling is introduced.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -445,10 +445,10 @@ namespace Mat
 
       //! derivative of the deviatoric, symmetric part of the Mandel stress tensor w.r.t. the
       //! inverse inelastic deformation gradient (Voigt stress form)
```
dragos-ana: Maybe we can add thermo-elastic here as well as you have done above? Not necessary, but maybe more explicit / consistent.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -1729,14 +1583,35 @@ namespace
...
 
   TEST_F(InelasticDefgradFactorsTest, TestEvaluateStateQuantities)
   {
+    auto transv_isotropic_material = set_up_viscoplastic_material(
+        {.mat_behavior = ViscoplastUtils::MatBehavior::transv_isotropic, .numgp = 8})
+                                         .material;
+    auto isotropic_material = set_up_viscoplastic_material({.yield_cond_a = std::nullopt,
+                                                               .yield_cond_b = std::nullopt,
+                                                               .yield_cond_f = std::nullopt,
+                                                               .numgp = 8})
```
dragos-ana: We can leave it at 1 Gauss point, although it was also previously set at 8, right?
Does not change anything in the code, but maybe it is slightly confusing why we need 8 Gauss points for evaluating the state at 1 Gauss point.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -32,6 +33,62 @@
...
+    ViscoplastUtils::LinearizationType linearization_type =
+        ViscoplastUtils::LinearizationType::analytic;
+    std::optional<double> yield_cond_a = 1.0;
+    std::optional<double> yield_cond_b = 2.0;
+    std::optional<double> yield_cond_f = 2.5;
+    int numgp = 1;
+    ReformulatedJohnsonCookParameters viscoplastic_law_params{};
+  };
+
+  struct ViscoplasticTestMaterial
```
dragos-ana, line 73: I like this entire approach a lot, thanks for taking the time to refactor this and make the unit tests more accessible!

dragos-ana (dismissed):

c-p-schmidt (approved):

dragos-ana (approved):

PR #2018, opened 2026-05-05
Introduce central manager for viscoelastic states

bgoderbauer:
## Description and Context

This PR extracts viscoelastic state handling from the `ViscoElastHyper` god-class into a specific struct `ViscoElastState` with consistent naming, ownership, and guardrails. This struct is now responsible for initializing, storing, updating, packing, and unpacking history data at the Gauss-point level needed to evaluate viscoelastic constitutive laws. This is just an infrastructure refactoring.

Long-term goals I'm working towards:
- `ViscoElastHyper` should only be an easily understandable orchestration layer that takes elasticity information and calls the subroutines for updating the constitutive law with the viscous contributions.
- Models should be added via a registry + factory pattern, not handled via excessive bool flags or nullptrs.
- State handling, constitutive model evaluation, and shared infrastructure should be logically separated to ease extension with new models.

#2005


bgoderbauer:
I extended the refactoring step further:
- All files belonging to the `ViscoElastHyper` toolbox were moved to a folder `mat/viscoelast`. The remaining viscoelasticity implementations are separate and can be unified with this framework in the future.
- I exchanged the nested if-cases in `ViscoElastHyper`that were checking for active models with a dedicated active model descriptor, and included safeguards to ensure that the pipeline is executed in the same order in all places (setup, evaluate, pack, unpack).
- A similar active state descriptor is added to `ViscoElastState`

bgoderbauer:
I know this is quite a large PR, but it is only infrastructure refactoring and it doesn't change any details for simulations with viscoelastic materials. Actual modernizations of the material kernels will follow after this is merged.

davidrudlstorfer (approved):

amgebauer (approved):

PR #2019, opened 2026-05-07
Introduce non-local growth remodeling

tuchpaul:
# Description and Context
This PR introduces the non-local stimulus $\psi$ as input for the local growth evolution equation of MixtureConstituentRemodelFiberSsi. Instead of driving growth/remodeling locally by the fiber Cauchy stress deviation ($\sigma - \sigma_h$) at each Gauss point, $\psi$ is computed as the Helmholtz-smoothed version of that quantity in the SSI scalar transport field:

$$\psi - \ell_c^2\Delta\psi = \sigma - \sigma_h$$

The structure side then integrates the growth/remodeling ODEs explicitly at each Gauss point, using $\psi$ in place of $\sigma - \sigma_h$. This makes the G&R driver non-local with characteristic length scale $\ell_c$.

# What's added

1. MAT_scatra_nl_stimulus as scatra material containing relevat parameters: CHAR_LENGTH_SQ, STRUCTURE_MAT_ID.
2. ScaTraEleCalcNonlocalStimulus (TYPE: "NLS" on SOLIDSCATRA elements) — assembles $\psi - \ell_c^2\Delta\psi = \sigma - \sigma_h$ as a stationary PDE, with the source term $\sigma - \sigma_h$ from the corresponding mixture constituent at each GP.

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
@@ -20,6 +20,7 @@
 #include "4C_mat_inelastic_defgrad_factors_service.hpp"
 #include "4C_mat_micromaterial.hpp"
 #include "4C_mat_scatra_growth_remodel.hpp"
+#include "4C_mat_scatra_nonlocal_stimulus.hpp"
 #include "4C_porofluid_pressure_based_elast_scatra_input.hpp"
 
 #include <filesystem>
@@ -397,6 +398,22 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
         },
         {.description = "Simple transport material with linear reaction used in growth "
                         "and remodeling"});
+  }
+
+  /*----------------------------------------------------------------------*/
+  // scalar transport material for non-local G&R stimulus (Helmholtz smoothing)
+  {
+    known_materials[Core::Materials::m_scatra_nl_stimulus] = group("MAT_scatra_nl_stimulus",
+        {
+            parameter<double>("CHAR_LENGTH_SQ",
+                {.description =
+                        "Squared characteristic length scale lc for the Helmholtz equation"}),
+            parameter<int>("STRUCTURE_MAT_ID",
+                {.description = "Material ID of the MixtureConstituentRemodelFiberSsi constituent "
+                                "that provides the local stimulus"}),
+        },
+        {.description = "Helmholtz equation for non-local G&R stimulus: "
+                        "psi - lc^2 * laplace(psi) = (sigma-sigma_h)"});
   }
 
   /*----------------------------------------------------------------------*/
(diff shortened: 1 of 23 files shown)


src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -399,6 +400,21 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                         "and remodeling"});
   }
 
+  /*----------------------------------------------------------------------*/
+  // scalar transport material for non-local G&R stimulus (Helmholtz smoothing)
+  {
+    known_materials[Core::Materials::m_scatra_nl_stimulus] = group("MAT_scatra_nl_stimulus",
+        {
+            parameter<double>("CHAR_LENGTH_SQ",
+                {.description = "Squared characteristic length scale for the Helmholtz equation"}),
```
c-p-schmidt: This would make it even more explicit.
```suggestion
                {.description = "Squared characteristic length scale lc for the Helmholtz equation"}),
```

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -4333,15 +4349,22 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
                 {.description =
-                        "Index of the corresponding growth scalar material in the scatra matlist",
-                    .validator = positive_or_zero<int>()}),
-            parameter<int>(
-                "REMODELING_SCALAR_ID", {.description = "Index of the corresponding remodeling "
-                                                        "scalar material in the scatra matlist",
-                                            .validator = positive_or_zero<int>()}),
+                        "Index of the corresponding growth scalar material in the scatra matlist "
+                        "(Default: -1 = disabled)",
+                    .default_value = -1}),
```
c-p-schmidt: What about using an `std::optional<int>` instead of a default value and then keeping the validator?

src/scatra_ele/4C_scatra_ele_calc_nonlocal_stimulus.hpp
```diff
@@ -0,0 +1,81 @@
...
+        const int numdofpernode, const int numscal, const std::string& disname);
+
+    void get_material_params(const Core::Elements::Element* ele, std::vector<double>& densn,
+        std::vector<double>& densnp, std::vector<double>& densam, double& visc,
+        const int iquad) override;
+
+    void materials(const std::shared_ptr<const Core::Mat::Material> material, const int k,
+        double& densn, double& densnp, double& densam, double& visc, const int iquad) override;
+
+    void mat_scatra_nls(const std::shared_ptr<const Core::Mat::Material> material, const int k,
```
c-p-schmidt, line 64: Can you add some docu to methods that are not overridden? The overridden methods are hopefully already documented in the parent. ;-)

c-p-schmidt (approved):

mairehenke (approved):
Looks good to me!

PR #2020, opened 2026-05-07
Generate performance report

amgebauer:
To visualize the performance tests, this PR adds a small python utility to generate a performance test report. Once available, we could add a link to the README to make the results accessible.

I will run a pipeline on my fork to verify that the changes work.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,12 +2,17 @@ name: deploy
 
 on:
   workflow_run:
-    workflows: [documentation, coverage, nightly_tests]
+    workflows: [documentation, coverage, nightly_tests, performance_report]
     types: [completed]
     branches: [main]
 
 permissions:
   contents: read
+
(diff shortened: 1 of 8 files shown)


bgoderbauer (dismissed):

davidrudlstorfer (dismissed):
Just to better understand without digging super deep into the code:

We always just append the newest performance data to the old one and then save it as an artifact again. Do we have a max number of entries so the artifact is not exploding in size?

c-p-schmidt (dismissed):
Thank you for adding the infrastructure to track and display the performance tests.

amgebauer:
Good point @davidrudlstorfer. I added an expiration of 365 days. The test runs on my fork.

Note, performance report that run now since a few days on my fork can be found here: https://amgebauer.github.io/4C/performance_report/report.html

davidrudlstorfer (approved):

c-p-schmidt (approved):

PR #2021, opened 2026-05-07
Fix version of rapidyaml

amgebauer:
The latest version seems to break our documentation and our schema validation in the pipeline. Let's fix the version of rapidyaml until they have fixed the underlying problem.

### DIFF
--- a/utilities/four_c_python/requirements.txt
+++ b/utilities/four_c_python/requirements.txt
@@ -1,4 +1,4 @@
-rapidyaml
+rapidyaml==0.11.1
 jsonschema-rs
 
 numpy


davidrudlstorfer:
@amgebauer do you actually know what is going wrong/failing? The last release was yesterday => https://github.com/biojppm/rapidyaml/releases/tag/v0.12.0 and it was only a release which targeted stuff that shouldn't affect us

mayrmt (approved):
@amgebauer can you please open an issue, so that we have a reminder to check, whether we can revert fixing of the version?

davidrudlstorfer:
On a first glance this looks like our own fault rather than a rapidyaml version problem 😅

https://github.com/4C-multiphysics/4C/actions/runs/25496470180/job/74818518883 only the 3 xfsi tests failed to parse?

amgebauer:
I opened the issue here: https://github.com/biojppm/rapidyaml/issues/597

amgebauer:
The problem is fixed upstream.

davidrudlstorfer:
@amgebauer thanks a lot for looking into this:)

PR #2022, opened 2026-05-07
[TSI] Fix solver increment handling in partitioned solver

rjoussen:
## Description and Context

The increment vectors need to store the current states before an outer iteration. This is done in every `outer_iteration_loop` except for `IterStagg` with temperature as the coupling variable.

Consequently, the corresponding tests `tsi_lincompression_iterstaggtemp.4C.yaml-p1` so far never detected that the outer iteration loop is actually converged, as can be seen by its output:

```shell
**************************************************************
      OUTER ITERATION LOOP 
      Time Step   9/2000 
**************************************************************
|    95/100    |   1.000E-06[L_2 ]      |  1.641E+02         |  1.704E+00         |
|    96/100    |   1.000E-06[L_2 ]      |  1.039E+03         |  5.085E-03         |
|    97/100    |   1.000E-06[L_2 ]      |  1.641E+02         |  1.704E+00         |
|    98/100    |   1.000E-06[L_2 ]      |  1.039E+03         |  5.085E-03         |
|    99/100    |   1.000E-06[L_2 ]      |  1.641E+02         |  1.704E+00         |
|   100/100    |   1.000E-06[L_2 ]      |  1.039E+03         |  5.085E-03         |
```

With this change, we get:
```shell
**************************************************************
      OUTER ITERATION LOOP 
      Time Step   9/2000 
**************************************************************
|     1/100    |   1.000E-06[L_2 ]      |  1.751E+01         |  1.704E-01         |
|     2/100    |   1.000E-06[L_2 ]      |  5.660E-02         |  5.426E-04         |
|     3/100    |   1.000E-06[L_2 ]      |  1.829E-04         |  1.754E-06         |
|     4/100    |   1.000E-06[L_2 ]      |  5.911E-07         |  5.667E-09         |
```

### DIFF
--- a/src/tsi/4C_tsi_partitioned.cpp
+++ b/src/tsi/4C_tsi_partitioned.cpp
@@ -478,6 +478,12 @@ void TSI::Partitioned::outer_iteration_loop()
         // get current temperatures due to solve thermo step, like predictor in FSI
         if (itnum != 1) temp_ = thermo_field()->tempnp();
 
+        // Store the current outer iterates before the sub-solves overwrite them.
+        // convergence_check() turns these into increments by subtracting them
+        // from the updated field solutions.
+        tempincnp_->update(1.0, *thermo_field()->tempnp(), 0.0);
+        dispincnp_->update(1.0, *structure_field()->dispnp(), 0.0);
+
         // begin nonlinear solver / outer iteration ***************************
 
         // ---------------------------------------------- structure field


maxfirmbach (approved):
Makes sense. @rjoussen FYI: Partitioned TSI is in generally in a really bad shape ...

maxfirmbach:
@rjoussen Can you check if `tsi_lincompression_iterstaggtemp.4C.yaml-p1` also runs on two processors and gives the same result?

rjoussen:
> Makes sense. @rjoussen FYI: Partitioned TSI is in generally in a really bad shape ...

Yeah that's my impression too, but it's at least conceptually doing the right things o believe. So when I want to make a material tsi-ready it's still worth to check. But yeah it needs an overhaul, just the algorithm class itself has a lot of duplicate code. Same for the thermo element's tsi capabilities

rjoussen:
> @rjoussen Can you check if `tsi_lincompression_iterstaggtemp.4C.yaml-p1` also runs on two processors and gives the same result?

It does! So I'd leave it as is for now.

maxfirmbach (approved):

mayrmt (approved):

PR #2027, opened 2026-05-10
Fix 2D solid output of out-of-plane components

bennoschoenstein:
## Summary

Out-of-plane stress and strain components (`sigma_zz`, `sigma_xz`, `sigma_yz`, `eps_zz`, `eps_xz`, `eps_yz`) were hardcoded to zero in the VTK runtime output of 2D solid elements. This PR extracts them from the consistent 3D material state.

## What this changes

The 2D solid element pipeline already evaluates the 3D material internally, but discards the out-of-plane components before reaching the runtime VTK output - which then hardcodes them to zero. This PR keeps the 3D state alive end-to-end.

**1. Extended the 2D `Stress` container** ([`4C_solid_ele_calc_lib.hpp`](src/solid_ele/4C_solid_ele_calc_lib.hpp)) to also carry `gl_strain_3d_` and `defgrd_3d_`, alongside the `pk2_3d_` field added in the initial commit.

**2. Filled the new fields in the three material-evaluation paths** ([`4C_solid_ele_calc_lib_plane.cpp`](src/solid_ele/4C_solid_ele_calc_lib_plane.cpp)):
- `evaluate_material_plane_strain` already operates on 3D internally (F_zz = 1, ε_zz = 0) — just propagate.
- `evaluate_material_plane_stress` already iterates ε_zz via local Newton; reconstruct `defgrd_3d` from it via `compute_deformation_gradient_from_gl_strains`. The function takes `defgrd` as an additional parameter.
- `evaluate_material_stress` StVK shortcut bypasses the 3D material call, so the 3D state is computed analytically: σ_zz = λ_3D · tr(ε) for plane strain (0 for plane stress), ε_zz = −ν/(1−ν) · tr(ε_2D) for plane stress (0 for plane strain), F_zz = √(1 + 2ε_zz).

**3. Switched the 2D output branches to use the 3D state** ([`4C_solid_ele_calc_lib_io.hpp`](src/solid_ele/4C_solid_ele_calc_lib_io.hpp)):
- `assemble_stress_type_to_matrix_row`: PK2 writes `pk2_3d_` directly; Cauchy delegates to the existing `Solid::Utils::pk2_to_cauchy` helper with the 3D state, removing the inlined push-forward from the initial commit.
- `assemble_strain_type_to_matrix_row`: GL, Euler-Almansi, log all reroute through `stress.gl_strain_3d_` / `stress.defgrd_3d_`. Signature gained a `Stress<celltype>&` parameter; both call sites in `4C_solid_ele_calc.cpp` and `4C_solid_scatra_ele_calc.cpp` updated.

## Side effect

Cauchy stress and Euler-Almansi strain output now work for plane stress as well, since the helpers are called with 3D quantities.

## Tests

The four existing `solid_ele_quad4_*_plane_*` tests are extended with `RESULT DESCRIPTION` entries at NODE 9 for the six out-of-plane components. Trivial-zero entries (per definition of the plane assumption) are tested with TOLERANCE 1e-12; non-trivial entries against numerical reference values.


maxfirmbach:
@bennoschoenstein Looking good. One thing though: We should add a vtk output test with the respective quantities, just to be sure.

amgebauer (dismissed):
Thank you @bennoschoenstein for fixing the stress and strain output for 2d solids.

@maxfirmbach I think a direct runtime output is not necessary since the added result test for the stresses is based on the same underlying data structures as the runtime output.

bennoschoenstein:
@amgebauer Thank you for taking a look at this PR. I just added @maxfirmbach's vtk output test, just to make sure.

maxfirmbach (approved):

amgebauer (approved):

PR #2029, opened 2026-05-12
Clean parallel issues

georghammerl:
## Description and Context
- Take more care which bonds are sent around in parallel in peridynamic simulations. Duplicate entries may lead to trouble, so filtering is necessary.
  - Deduplicate bond entries during parallel sending and receiving in `PDNeighborPairs::communicate_bond_list()`
  - Error is thrown if still duplicate entry occurs in  `PDNeighborPairs::setup_peridynamic_pair_hashes()` -> reason needs further investigation in that scenario
- Also, the bin size check has been adapted such that slightly stretched bonds are still found in the preliminary geometry based binning search process.
- Ensure equal grid spacing for peridynamic and SPH particles (crucial for proper physics).

### DIFF
--- a/src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
+++ b/src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
@@ -24,6 +24,7 @@
 
 #include <Teuchos_TimeMonitor.hpp>
 
+#include <algorithm>
 #include <tuple>
 
 FOUR_C_NAMESPACE_OPEN
@@ -31,10 +32,43 @@ FOUR_C_NAMESPACE_OPEN
 // forward declarations
 namespace
 {
+  // Peridynamic bonds are identified by the unordered pair of particle global ids.
+  // Canonicalize to (min, max) whenever a bond is serialized or deduplicated.
+  std::pair<int, int> canonical_pd_bond_ids(const int i, const int j)
+  {
+    return {std::min(i, j), std::max(i, j)};
+  }
+
   long compute_pd_pair_hash_key(const long i, const long j)
   {
     // create same hashes independent of order of i and j
-    return (0.5 * (i + j) * (i + j + 1)) + std::min(i, j);
+    const long sum = i + j;
+    return (sum * (sum + 1)) / 2 + std::min(i, j);
+  }
+
+  Particle::LocalGlobalIndexTuple make_default_peridynamic_bond_tuple(const int globalid)
+  {
+    // Initialize a bond tuple where only the global id is known so far. Type and status remain
+    // unresolved until the particle is found in a local container, and local id = -1 marks that
+    // "not yet resolved on this rank" state
+    return std::make_tuple(
+        Particle::UninitializedType, Particle::UninitializedStatus, -1, globalid);
+  }
+
+  std::set<long> build_known_peridynamic_bond_hashes(const std::shared_ptr<std::vector<
+          std::pair<Particle::LocalGlobalIndexTuple, Particle::LocalGlobalIndexTuple>>>& bondlist)
+  {
+    std::set<long> known_bond_hashes;
+
+    for (const auto& [first, second] : *bondlist)
+    {
+      const int globalid_i = std::get<3>(first);
+      const int globalid_j = std::get<3>(second);
+
+      known_bond_hashes.insert(compute_pd_pair_hash_key(globalid_i, globalid_j));
+    }
+
+    return known_bond_hashes;
   }
 
   bool is_valid_peridynamic_bond_entry(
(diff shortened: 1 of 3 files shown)


src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -53,6 +53,12 @@ Particle::SPHPeridynamic::SPHPeridynamic(const Teuchos::ParameterList& particle_
   FOUR_C_ASSERT_ALWAYS(not particle_params.get<bool>("RIGID_BODY_MOTION"),
       "Peridynamic interaction is not available in combination with rigid body motion!");
 
+  // A mismatch produces physically inconsistent forces
+  const double dx_sph = particle_params.sublist("SPH").get<double>("INITIALPARTICLESPACING");
+  FOUR_C_ASSERT_ALWAYS(std::abs(dx_pd_ - dx_sph) <= 1.0e-10 * dx_pd_,
+      "PERIDYNAMIC_GRID_SPACING (%.10f) must equal INITIALPARTICLESPACING (%.10f). ", dx_pd_,
```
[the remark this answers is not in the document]
georghammerl: done

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
```diff
@@ -31,10 +31,41 @@ FOUR_C_NAMESPACE_OPEN
...
+    return {std::min(i, j), std::max(i, j)};
+  }
+
   long compute_pd_pair_hash_key(const long i, const long j)
   {
     // create same hashes independent of order of i and j
-    return (0.5 * (i + j) * (i + j + 1)) + std::min(i, j);
+    const long sum = i + j;
+    return (sum * (sum + 1)) / 2 + std::min(i, j);
+  }
```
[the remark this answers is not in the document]
georghammerl: included header

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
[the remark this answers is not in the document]
georghammerl: just use `abs`

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
```diff
@@ -458,14 +491,8 @@ void Particle::PDNeighborPairs::communicate_bond_list(
...
-
-  Particle::TypeEnum type_i;
-  Particle::StatusEnum status_i;
-  int particle_i, globalid_i;
-
-  Particle::TypeEnum type_j;
-  Particle::StatusEnum status_j;
-  int particle_j, globalid_j;
+  std::vector<std::set<std::pair<int, int>>> communicated_bonds_per_target(
+      Core::Communication::num_mpi_ranks(comm_));
```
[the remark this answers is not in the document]
georghammerl: Memory overhead is marginal compared to algorithmic overhead to store only the to_rank data in some map. Won't adapt.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
```diff
@@ -53,6 +53,12 @@ Particle::SPHPeridynamic::SPHPeridynamic(const Teuchos::ParameterList& particle_
   FOUR_C_ASSERT_ALWAYS(not particle_params.get<bool>("RIGID_BODY_MOTION"),
       "Peridynamic interaction is not available in combination with rigid body motion!");
 
+  // A mismatch produces physically inconsistent forces
+  const double dx_sph = particle_params.sublist("SPH").get<double>("INITIALPARTICLESPACING");
+  FOUR_C_ASSERT_ALWAYS(abs(dx_pd_ - dx_sph) <= 1.0e-10 * dx_pd_,
+      "PERIDYNAMIC_GRID_SPACING ({:.10f}) must equal INITIALPARTICLESPACING ({:.10f}). ", dx_pd_,
```
slfuchs, line 59: Do we need this input parameter then, or could we use the initial particle spacing? PD kind of lives within the SPH framework anyway.

slfuchs (commented):
The parallel stuff is hard to review 😆 but from what i saw this should be fine.

src/particle/src/interaction/4C_particle_interaction_sph_peridynamic.cpp
georghammerl: I was not aware during initial setup of PD that this link is so important. Now, one parameter is enough. However, I prefer to have it in a separate PR as all input files need to be touched.

slfuchs (dismissed):

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
```diff
@@ -24,17 +24,49 @@
...
     // create same hashes independent of order of i and j
-    return (0.5 * (i + j) * (i + j + 1)) + std::min(i, j);
+    const long sum = i + j;
+    return (sum * (sum + 1)) / 2 + std::min(i, j);
+  }
+
+  Particle::LocalGlobalIndexTuple make_default_peridynamic_bond_tuple(const int globalid)
+  {
+    return std::make_tuple(
+        Particle::UninitializedType, Particle::UninitializedStatus, -1, globalid);
```
vovannikov, line 55: I would add a comment here on how the bound is initialized. What is -1? A bit cryptic without looking that the other parts of the code.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
```diff
@@ -24,17 +24,49 @@
...
+  {
+    std::set<long> known_bond_hashes;
+
+    for (const auto& [first, second] : *bondlist)
+    {
+      int globalid_i, globalid_j;
+      std::tie(std::ignore, std::ignore, std::ignore, globalid_i) = first;
+      std::tie(std::ignore, std::ignore, std::ignore, globalid_j) = second;
+
+      known_bond_hashes.insert(compute_pd_pair_hash_key(globalid_i, globalid_j));
```
vovannikov: It is a matter of taste, but is not it less verbose to use std::get here instead of structural binding with so many ignore's? The same applies to other places below.

vovannikov (dismissed):
Fine for me. Some improvements are possible, but those are optional.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
georghammerl: Good point. Added some comment.

src/particle/src/interaction/4C_particle_interaction_pd_neighbor_pairs.cpp
georghammerl: Sure. Changed to `std::get<3>` here and elsewhere.

vovannikov (approved):

slfuchs (approved):

PR #2030, opened 2026-05-12
Add more verbose error message to particle code

ppraegla:
Follow-up of #1783 . The calculation of the maximum position increment is adjusted to output additional information in case the increment is larger than the minimum bin size.

I moved the old logic into a lambda function that is called a second time to output additional debug information in case we terminate the simulation. Since the if condition is known at compile time, the compiler should optimize away the if branch in case we only want the maximum position increment.

The new error message looks like the following.
```
2 particle(s) traveled more than one bin on this processor.
  Minimum bin size: 0.025
  Maximum position increment: 0.99999975 for particle with
    gid: 0 of type phase1
    Position: (0.98499975, 0, 0)
    Velocity: (999.99975, 0, 0)
  Particle type: number of violating particles
    phase1: 1
    phase2: 1
This can have numerous causes and can hint at an unstable simulation. Carefully check the time step size and other interaction related parameters.
```

### DIFF
--- a/src/particle/src/algorithm/4C_particle_algorithm.cpp
+++ b/src/particle/src/algorithm/4C_particle_algorithm.cpp
@@ -22,6 +22,7 @@
 #include "4C_particle_engine.hpp"
 #include "4C_particle_engine_communication_utils.hpp"
 #include "4C_particle_engine_container.hpp"
+#include "4C_particle_engine_enums.hpp"
 #include "4C_particle_engine_object.hpp"
 #include "4C_particle_input.hpp"
 #include "4C_particle_interaction_base.hpp"
@@ -37,7 +38,10 @@
 #include <Teuchos_StandardParameterEntryValidators.hpp>
 #include <Teuchos_TimeMonitor.hpp>
 
+#include <cstddef>
+#include <map>
 #include <memory>
+#include <sstream>
 
 FOUR_C_NAMESPACE_OPEN
 
@@ -719,8 +723,7 @@ bool Particle::ParticleAlgorithm::check_max_position_increment()
   }
 
   // get max particle position increment since last transfer
-  double maxparticlepositionincrement = 0.0;
-  get_max_particle_position_increment(maxparticlepositionincrement);
+  double maxparticlepositionincrement = get_max_particle_position_increment();
 
   // get max wall position increment since last transfer
   double maxwallpositionincrement = 0.0;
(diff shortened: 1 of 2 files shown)


georghammerl:
@ppraegla Thank you for the additional output.

My current debug output contains the position and velocity at which the too fast particles occur. This is very helpful to me, because it is easier to detect what is in the neighborhood there and in which direction is the problematic particle traveling. Here is a patch file of my current debug output. Maybe, it can be unioned with yours?!
[verbose_output.patch](https://github.com/user-attachments/files/27640351/verbose_output.patch)

My experience is that only few particles hit this error such that an even more verbose answer is acceptable.

src/particle/src/algorithm/4C_particle_algorithm.cpp
```diff
@@ -739,63 +742,121 @@ bool Particle::ParticleAlgorithm::check_max_position_increment()
...
+    {
+      if (value == 0) continue;
+
+      total_number_of_violating_particles += value;
+      ss << "    " << enum_to_type_name(key) << ": " << value << "\n";
+    }
+    FOUR_C_THROW(
+        "{} particle(s) traveled more than one bin on this processor.\n  Maximum position "
+        "increment: {}\n  Minimum bin size: {}\n  Particle type: number of violating "
+        "particles\n{}This can have numerous causes and can hint at an unstable simulation. Often "
```
slfuchs: I like this Info, for us developers this may be clear, but this message is really helpful for new users of the framework.

src/particle/src/algorithm/4C_particle_algorithm.cpp
```diff
@@ -739,63 +742,121 @@ bool Particle::ParticleAlgorithm::check_max_position_increment()
...
+        "it helps to reduce the time step size.",
```
slfuchs: Often it is not just the time step size but an interaction related parameter or something else. So this sentence might be misleading. What about: "Carefully check the time step size and other interaction related parameters."

slfuchs (dismissed):
Thanks for your effort! Looks good and improves the quality of this check a lot.

ppraegla:
I updated the error message
```
2 particle(s) traveled more than one bin on this processor.
  Minimum bin size: 0.025
  Maximum position increment: 0.99999975 for particle with
    gid: 0 of type phase1
    Position: (0.98499975, 0, 0)
    Velocity: (999.99975, 0, 0)
  Particle type: number of violating particles
    phase1: 1
    phase2: 1
This can have numerous causes and can hint at an unstable simulation. Carefully check the time step size and other interaction related parameters.
```

slfuchs (approved):

georghammerl (approved):

PR #2034, opened 2026-05-15
Minor rework of MAT_electrode

c-p-schmidt:
## Description and Context
Minor improvements to the electrode material implementation and the user interface. The main points are:
- add validators for maximum concentration and corresponding lithation state
- remove irrelevant OCP_PARA_NUM from Redlich-Kister OCP
- make X_MIN and X_MAN optional and remove -1 as a fake optional value
- If OCP is calculated by a function, the function is only assigned once on the first usage and not always taken from the global problem (unfortunately, we can not directly assign it in the constructor, as there is a crazy dependency of the functions on the materials, thus the functions are not yet read once the materials are created).
- improve the documentation
- all input files are adapted accordingly


src/mat/4C_mat_electrode.cpp
```diff
@@ -23,39 +23,69 @@ Mat::PAR::Electrode::Electrode(const Core::Mat::PAR::Parameter::Data& matdata)
...
+    FOUR_C_ASSERT_ALWAYS(xmax_.has_value(),
+        "If lower bound (X_MIN) of range of validity for ocp calculation model is set the upper "
+        "bound (X_MAX) also needs to be set. Both have to be in the range [0,1] since X is "
+        "calculated as c/c_max!");
+    FOUR_C_ASSERT_ALWAYS(xmin_.value() >= 0.0,
+        "Lower bound (X_MIN: {}) of range of validity for ocp calculation model cannot be negative "
+        "since X is calculated as c/c_max!",
+        xmin_.value());
+    FOUR_C_ASSERT_ALWAYS(xmin_.value() < 1.0,
+        "Lower bound (X_MIN: {}) of range of validity for ocp calculation model cannot be larger "
```
dragos-ana: These checks could be combined into a single one, referencing the interval [0, 1].

src/mat/4C_mat_electrode.cpp
```diff
@@ -23,39 +23,69 @@ Mat::PAR::Electrode::Electrode(const Core::Mat::PAR::Parameter::Data& matdata)
...
-        "the screen if bounds are violated throughout the simulation time!");
+    FOUR_C_ASSERT_ALWAYS(xmin_.has_value(),
+        "If upper bound (X_MAX) of range of validity for ocp calculation model is set the lower "
+        "bound (X_MIN) also needs to be set. Both have to be in the range [0,1] since X is "
+        "calculated as c/c_max!");
+    FOUR_C_ASSERT_ALWAYS(xmax_.value() > 0.0,
+        "Upper bound (X_MAX: {}) of range of validity for ocp calculation model cannot be negative "
+        "since X is calculated as c/c_max!",
+        xmax_.value());
+    FOUR_C_ASSERT_ALWAYS(xmax_.value() <= 1.0,
```
dragos-ana: Same here

src/mat/4C_mat_electrode.cpp
```diff
@@ -133,34 +176,29 @@ double Mat::Electrode::compute_open_circuit_potential(
 
   // print warning to screen if prescribed interval of validity for ocp calculation model is given
   // but not satisfied
-  if (((X < params_->xmin_) or (X > params_->xmax_)) and params_->xmax_ >= 0.0)
+  if (params_->xmin_.has_value() and (X < params_->xmin_.value() or X > params_->xmax_.value()))
```
dragos-ana:
```suggestion
  if ((params_->xmin_.has_value() and (X < params_->xmin_.value()) or (params_->xmax_.has_value() and (X > params_->xmax_.value())))
```

src/mat/4C_mat_electrode.cpp
```diff
@@ -133,34 +176,29 @@ double Mat::Electrode::compute_open_circuit_potential(
...
   {
     std::cout << "WARNING: intercalation fraction X = c/c_max is violating prescribed bounds of "
                  "ocp calculation model. Calculated values might therefore not be reasonable!"
               << '\n';
-    std::cout << "X: " << X << " lower bound is: " << params_->xmin_
-              << " upper bound is: " << params_->xmax_ << '\n'
-              << '\n';
+    std::cout << "X: " << X << " lower bound is: " << params_->xmin_.value()
+              << " upper bound is: " << params_->xmax_.value() << "\n\n";
```
dragos-ana: Maybe it would be better to split the check above into one for xmin, and one for xmax since they are both optional now? Retrieving both values might not always be possible with the new logic.

src/mat/4C_mat_electrode.cpp
```diff
@@ -351,19 +385,16 @@ double Mat::Electrode::compute_d2_open_circuit_potential_d_concentration_d_conce
...
-      // cubic spline interpolation of *.csv data points
+      // second derivative of the open circuit potential w.r.t. concentration, obtained from
+      // function
       case Mat::PAR::OCPModels::function:
       {
-        const int ocp_function_number = params_->ocpfunctnum_;
-        d2_ocp_dX2 = Global::Problem::instance()
-                         ->function_by_id<Core::Utils::FunctionOfScalar>(ocp_function_number)
-                         .evaluate_derivative(X, 2);
+        d2_ocp_dX2 = params_->ocp_function().evaluate_derivative(X, 2);
```
dragos-ana, line 393: All these value retrievals are much nicer now with `ocp_function()`!

dragos-ana (dismissed):
Looks good to me! Left some comments / suggestions

src/mat/4C_mat_electrode.cpp
c-p-schmidt: This is actually not true, as during setup, it is ensured that either both or neither are set.
However, this is currently not well represented in the code. For that, one would need to define a struct that clusters both limits and make this construct optional.
Is it worth looking into this in your opinion?

dragos-ana: True, I missed that. I would say that your proposal is actually a better representation, if either both limits need to be set or not. And this would maybe also simplify the checks

src/mat/4C_mat_electrode.cpp
```diff
@@ -22,40 +22,44 @@ Mat::PAR::Electrode::Electrode(const Core::Mat::PAR::Parameter::Data& matdata)
     : ElchSingleMat(matdata),
       cmax_(matdata.parameters.get<double>("C_MAX")),
       chimax_(matdata.parameters.get<double>("CHI_MAX")),
-      ocpmodel_(matdata.parameters.group("OCP_MODEL").get<OCPModels>("OCP_MODEL")),
-      xmin_(matdata.parameters.group("OCP_MODEL").get<double>("X_MIN")),
-      xmax_(matdata.parameters.group("OCP_MODEL").get<double>("X_MAX"))
+      ocpmodel_(matdata.parameters.group("OCP_MODEL").get<OCPModels>("OCP_MODEL"))
 {
```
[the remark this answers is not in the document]
c-p-schmidt: @dragos-ana: What do you think? Should we keep such a lower limit? Currently, it is only a positive double, but we could switch to a
`in_range(1.0e-12, std::numeric_limits<double>::max())`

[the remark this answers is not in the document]
dragos-ana: This sounds good to me!

dragos-ana (approved):

lkoeglmeier (approved):

PR #2035, opened 2026-05-15
Add possibility to use functions for bodyforces in porofluid_pressure_based* problems

lkoeglmeier:
This pr adds the possibility to use functions for the bodyforce contribution in the fluid field for porofluid_pressure_based* problem types.


c-p-schmidt:
Thank you for going the extra mile. Looks good to me.

However, avoiding the const_cast as suggested by Copilot and replacing `Core::Utils::FunctionOfSpaceTime*` with `const Core::Utils::FunctionOfSpaceTime*` to make this work would be a reasonable improvement.

c-p-schmidt (approved):

engstlerL (approved):

PR #2036, opened 2026-05-15
Fix monitor_dbc DOF iteration in 2D

### LINKED ISSUE #2032: [BUG] MonitorDbc::get_reaction_moment produces garbage moments in 2D problems
## Bug Description

`MonitorDbc::get_reaction_moment` assumes 3D throughout. The spatial dimension is hard-coded as `const static unsigned DIM = 3` in `4C_structure_new_monitor_dbc.hpp:60`.

In a 2D problem, the third iteration accesses a non-existent z-DOF. The read returns uninitialised memory, which then contaminates `node_position[2]` and flows into the `cross_product` that builds the reaction moment.

Two failure modes, depending on what is in the memory cell:

1. **Small garbage** -> silent bug. The `m_x` and `m_y` columns in the monitor output are non-zero, even though they must be exactly 0 in 2D (the only non-trivial 2D moment is `m_z = x·F_y - y·F_x`).
2. **Large garbage** -> `cross_product` overflows. The structural predictor aborts with `Floating Point Exception: OVERFLOW`. I hit this in 4 out of 200 jobs of a 2D-RVE convergence study.

Related Issues: 
- #1283  same fix pattern, applied to ~30 other files; monitor_dbc.cpp was missed
- #1228 broader Discretization geometry cleanup
- #1858 same family of bug (dimension hardcoding)

bennoschoenstein:
## Problem
In 2D the code reads past the end of two-element arrays. Most of the time the garbage that comes back is silently written into the moment columns of the CSV. Occasionally it is large enough to overflow the cross product that builds the moment, aborting the simulation. In practice this caused random crashes in my 2D RVE convergence studies.

## Fix
The DOF-indexed loops now use the actual spatial dimension of the discretisation. The 3-component storage matrices stay as they are because the cross product is only defined in 3D — the out-of-plane components then come out exactly zero in 2D, which is the physically correct answer (`m_z = x·F_y − y·F_x`).

A code review surfaced two more functions with the same pattern; all three are fixed here.

## Validation
- Existing `solid_monitor_dbc` regression tests still pass.
- New 2D regression test pins the formerly-corrupt CSV columns to exact zero
  under `TOL_A = 1e-12`.

Fixes #2032.

**Vorher:**
grafik

**Nachher:**
grafik


maxfirmbach:
I think the current state introduces a bit of ambiguity. Now we have `DIM` and `n_dim`, which in my opinion makes things unnecessarily confusing. I would prefer to modify the data structures in a way, such that they have the correct size right at the beginning and not have two different variables, which determine the size and how long I loop over that size. 

Edit: This would also avoid comments in the actual implementation, which should be self explaining.

tests/input_files/solid_monitor_dbc_2d.4C.yaml
```diff
@@ -0,0 +1,78 @@
+TITLE:
+  - "Regression test for the monitor_dbc moment-loop bug in 2D problems."
```
maxfirmbach: I honestly don't like this description that much. It describes a bug, which should be fixed with this PR, so no need to elaborate on it here. In addition, the description is quite detailed referring to the actual implementation and is again past tense ... so referring to something that _has been_ like that. We should add information what is actually tested and not what bug is triggered (because it isn't).

maxfirmbach:
@bennoschoenstein Closes https://github.com/4C-multiphysics/4C/issues/2032. Right?

bennoschoenstein:
@maxfirmbach thank you for your comments, and you are right, it would be better to be able to solve the dimensional problem beforehand. I will think of a way how to solve this problem more elegantly and reduce the unnecessary long comments. And yes, it would also solve/close Issue #2032 once its merged

bennoschoenstein:
- I removed `const static unsigned DIM = 3;` since it was misleading, since in 2D it does not represent the actual dimension.
- I store the reaction maps in a `std::vector` so we can use `.size()` and iterate over the real number of DOF directions.
- Cleaned up some unused code.
- Changed the return type of `get_reaction_force`/`get_reaction_moment` from `double` to `void`, since the returned value was never used.

m-frey (commented):
Looks good, thanks for fixing this

isteinbrecher (approved):

m-frey (approved):

PR #2037, opened 2026-05-19
[MultSplit Material] Add thermal stress evaluation helper

rjoussen:
## Description and Context

Add a `ThermalQuantities` struct to hold quantities related to the thermal deformation. This will be used by both the multiplicative split material and the factor. Those quantities are evaluated with `evaluate_thermal_quantities`.

Added two new service functions:
- `add_thermal_stress_contribution` updates the second PK using thermal stress factors.
- `calculate_thermal_stress_derivative` calculates the partial derivative of the 2PK stress with respect to temperature. 

A unittest covers `ThermalQuantities` evaluation and the stress computation helpers.

Thanks again to @dragos-ana who initially started working on this.

### DIFF
--- a/src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
+++ b/src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
@@ -12,6 +12,9 @@
 #include "4C_linalg_fixedsizematrix.hpp"
 #include "4C_linalg_fixedsizematrix_tensor_products.hpp"
 #include "4C_linalg_fixedsizematrix_voigt_notation.hpp"
+#include "4C_linalg_symmetric_tensor.hpp"
+#include "4C_linalg_tensor_conversion.hpp"
+#include "4C_linalg_tensor_generators.hpp"
 #include "4C_mat_elasthyper_service.hpp"
 #include "4C_mat_service.hpp"
 
@@ -28,6 +31,37 @@ namespace Mat
     Core::LinAlg::Matrix<3, 1> gamma{Core::LinAlg::Initialization::zero};
     /// Constitutive tensor factors, cf. Holzapfel p. 261.
     Core::LinAlg::Matrix<8, 1> delta{Core::LinAlg::Initialization::zero};
+  };
+
+  /// Thermal stretch quantities used by multiplicative-split thermoelastic stress evaluation.
+  struct ThermalQuantities
+  {
+    // ----- variables of thermal quantities ----- //
+    /// thermal right Cauchy-Green deformation tensor \f$ \mathbf{C}_T \f$ stored as 6x1
+    /// vector (stress-form!)
+    Core::LinAlg::Matrix<6, 1> CTV{Core::LinAlg::Initialization::zero};
+    /// inverse thermal right Cauchy-Green deformation tensor \f$ \mathbf{C}_T^{-1} \f$ stored as
+    /// 6x1 vector (stress-form)
+    Core::LinAlg::Matrix<6, 1> iCTV{Core::LinAlg::Initialization::zero};
+    /// \f$ \mathbf{F}_{\text{in}}^{-1} \mathbf{C}_T \mathbf{F}_{\text{in}}^{-T} \f$ stored as 6x1
+    /// vector (stress-form!)
+    Core::LinAlg::Matrix<6, 1> iFinCTiFinTV{Core::LinAlg::Initialization::zero};
+    /// \f$ \mathbf{F}_{\text{in}}^{-1} \mathbf{C}_T^{-1} \mathbf{F}_{\text{in}}^{-T} \f$ stored
+    /// as 6x1 vector (stress-form!)
+    Core::LinAlg::Matrix<6, 1> iFiniCTiFinTV{Core::LinAlg::Initialization::zero};
+    /// derivative of thermal right Cauchy-Green deformation tensor wrt temperature \f$ \mathrm{d}
+    /// \mathbf{C}_T / \mathrm{d} T \f$ stored as 6x1 vector (strain-form!)
+    Core::LinAlg::Matrix<6, 1> dCTdTV{Core::LinAlg::Initialization::zero};
+
+    /// principal invariants of the thermal right Cauchy-Green tensor
+    Core::LinAlg::Matrix<3, 1> prinv{Core::LinAlg::Initialization::zero};
+
+    // ----- derivatives of principal invariants ----- //
+
+    /// first derivatives of principal invariants
+    Core::LinAlg::Matrix<3, 1> dPI{Core::LinAlg::Initialization::zero};
+    /// second derivatives of principal invariants
+    Core::LinAlg::Matrix<6, 1> ddPII{Core::LinAlg::Initialization::zero};
   };
 
   inline void evaluate_ce(const Core::LinAlg::Matrix<3, 3>& F,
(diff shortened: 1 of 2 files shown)


src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
```diff
@@ -128,6 +162,167 @@ namespace Mat
     Core::LinAlg::FourTensorOperations::add_holzapfel_product(cmat, iCinv, delta(7));
   }
 
+  /*!
+   * @brief Subtracts the thermal contribution from the 2nd Piola-Kirchhoff stress
+   *
+   * \f[
+   * \texttt{stress} \mathrel{{-}{=}} \det(\boldsymbol{F}_\text{in}) \boldsymbol{F}_\text{in}^{-1}
```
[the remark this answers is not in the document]
rjoussen: api-wise, this is an addition with a negative sign, so I'll leave it.

rjoussen:
Also, I noticed that the `4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp` header currently defines `elast_hyper_evaluate_elastic_part` (which is also tested in `4C_mat_multiplicative_split_defgrad_elasthyper_service_test.cpp` as expected), however it is not actually used at all in the multiplicative split material. Instead, it is only used by some of the `4C_mixture_constituent_elasthyper*` files. The function should probably be moved closer to there. I also assume that the service file header's namespace is not protective enough (its just `Mat::`), which is probably why this happened in the first place.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
```diff
@@ -128,6 +162,166 @@ namespace Mat
...
+   * \boldsymbol{S}_\text{he}[\boldsymbol{C}_\text{he} = \boldsymbol{C}_T]
```
dragos-ana, line 170: Maybe a small reference on what we mean by $(\cdot)_{\text{he}}$ would be nice here

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
```diff
@@ -128,6 +162,166 @@ namespace Mat
...
+   *   : \frac{\partial \boldsymbol{C}_T}{\partial T}
+   * \right)
+   * \boldsymbol{F}_{\text{in}}^{-T}.
+   * \f]
+   *
+   * @param[in] iFinM inverse inelastic deformation gradient
+   * @param[in] thermal_quant Thermal stretch quantities and temperature derivative
+   * @return derivative of the thermal 2nd Piola-Kirchhoff stress w.r.t. temperature
+   */
+  inline Core::LinAlg::Matrix<6, 1> calculate_thermal_stress_derivative(
```
dragos-ana:
```suggestion
  inline Core::LinAlg::Matrix<6, 1> calculate_thermal_stress_temperature_derivative(
```

Including partial would also be nice, but then the name explodes.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
```diff
@@ -128,6 +162,166 @@ namespace Mat
...
+      const Core::LinAlg::Matrix<3, 3>& iFinM, const ThermalQuantities& thermal_quant)
+  {
+    Core::LinAlg::Matrix<6, 1> thermal_stress_deriv{Core::LinAlg::Initialization::zero};
+
+    const Core::LinAlg::Matrix<6, 1>& dCTdTV = thermal_quant.dCTdTV;
+    const double detFin = 1.0 / iFinM.determinant();
+
+    // clear variables
+    thermal_stress_deriv.clear();
```
dragos-ana: Not required, since you initialized it as zero above

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
```diff
@@ -128,6 +162,166 @@ namespace Mat
...
+    Core::LinAlg::Matrix<3, 3> iFin_pStheta_pT{Core::LinAlg::Initialization::zero};
+    iFin_pStheta_pT.multiply_nn(1.0, iFinM, pStheta_pT, 0.0);
+    Core::LinAlg::Matrix<3, 3> iFin_pStheta_pT_iFinT{Core::LinAlg::Initialization::zero};
+    iFin_pStheta_pT_iFinT.multiply_nt(1.0, iFin_pStheta_pT, iFinM, 0.0);
+    Core::LinAlg::Matrix<6, 1> iFin_pStheta_pT_iFinT_V{Core::LinAlg::Initialization::zero};
+    Core::LinAlg::Voigt::Stresses::matrix_to_vector(iFin_pStheta_pT_iFinT, iFin_pStheta_pT_iFinT_V);
+
+    // thermal derivative
+    thermal_stress_deriv.update(-1.0, iFin_pStheta_pT_iFinT_V, 0.0);
+    thermal_stress_deriv.scale(detFin);
```
dragos-ana:
```suggestion
    thermal_stress_deriv.update(-detFin, iFin_pStheta_pT_iFinT_V, 0.0);
```

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
```diff
@@ -30,6 +33,37 @@ namespace Mat
     Core::LinAlg::Matrix<8, 1> delta{Core::LinAlg::Initialization::zero};
   };
 
+  /// Thermal stretch quantities used by multiplicative-split thermoelastic stress evaluation.
+  struct ThermalQuantities
+  {
+    // ----- variables of thermal quantities ----- //
+    /// thermal right Cauchy-Green deformation tensor \f$ \mathbf{C}_T \f$ stored as 6x1
+    /// vector (stress-form!)
+    Core::LinAlg::Matrix<6, 1> CTV{Core::LinAlg::Initialization::zero};
```
dragos-ana: The "T" could be lightly misunderstood as transposed in the code. Maybe something like `CtemperV` would solve this problem?

unittests/mat/4C_multiplicative_split_defgrad_elasthyper_service_test.cpp
```diff
@@ -156,4 +157,118 @@ namespace
...
+    S_ref(0, 2) = 0.0000000000000000;
+    S_ref(1, 0) = -32.4795185513294626;
+    S_ref(1, 1) = -103.2384696810115088;
+    S_ref(1, 2) = 0.0000000000000000;
+    S_ref(2, 0) = 0.0000000000000000;
+    S_ref(2, 1) = 0.0000000000000000;
+    S_ref(2, 2) = -37.1194497729479664;
+    //**************************************************
+    Core::LinAlg::Matrix<3, 3> dS_dT_ref_{Core::LinAlg::Initialization::zero};
+    dS_dT_ref_(0, 0) = -0.0000941798851085;
```
dragos-ana: This is very inconsistent throughout our current code, but should we maybe start denoting partial derivatives with p instead of d? 
@rjoussen, @c-p-schmidt: We should discuss this to find a more descriptive notation

dragos-ana (commented):
Looks very nice! I left some small suggestions

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
rjoussen: What about
```suggestion
  inline Core::LinAlg::Matrix<6, 1> compute_partial_d_stress_d_temperature(
```
?
`thermal` and `temperature` are not needed here.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
rjoussen: fair

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
rjoussen: I agree, but I dont like `CtemperV` either. In [16b0a70](https://github.com/4C-multiphysics/4C/pull/2037/commits/16b0a7092eb4b767cbeaec1b1b6fc58d334f8268), I renamed the quantities to their actual names (not trying to imitate math by variable names). Do you like this better?

unittests/mat/4C_multiplicative_split_defgrad_elasthyper_service_test.cpp
rjoussen: I changed it to p for now, though I don't love this solution either. First, p is kind of non-standard to denote a derivative; second, in many places, it is ambiguous if something is a total or partial derivative. I think the only good solution would be to use full names for variable names as much as possible (at least for all public variables), and try to work with structs more to prevent the names from exploding. Also, many of the really long and complicated names are intermediate variables, which are a product or a sum of other variables. I think a lot of that is probably not needed anymore if we ever switch to tensors (which would also mean that we don't need to specify stress/strain form anymore, which we don't really do atm anyway)

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
c-p-schmidt: Hm, so you only know what exactly it is via the corresponding struct, since "stressform" alone doesn't tell me a lot.

When I think about the code I would like to read, it is rather close to how you would also write it in the paper. However, for that, we obviously would need tensors again.

Is this already a hint that it is not so helpful to put much effort into naming now, but rather to switch to tensors as soon as possible? What do you think?

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
```diff
@@ -128,6 +166,175 @@ namespace Mat
...
+      const Core::LinAlg::Matrix<3, 3>& iFinM, const ThermalQuantities& thermal_quant)
+  {
+    Core::LinAlg::Matrix<6, 1> thermal_stress_deriv{Core::LinAlg::Initialization::zero};
+
+    const Core::LinAlg::Matrix<6, 1>& dCTdT_strainform =
+        thermal_quant.thermal_cauchy_green.partial_deriv_wrt_temperature_strainform;
+    const double detFin = 1.0 / iFinM.determinant();
+
+    // evaluate purely hyperelastic stiffness with the thermal right CG tensor as input
+    Core::LinAlg::SymmetricTensor<double, 3, 3> hyperelast_stress{};
```
c-p-schmidt, line 221: Is there a specific reason why you implement all of this directly in the hpp and not in a cpp?

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
dragos-ana: Although the idea is good, it becomes rather strange when you have a struct for a tensor, which then also contains elements such as "pulled_back_by"...

I would, for now, still be in favor of renaming the specific variables, maybe also including the variable names rather than the math for such "single" tensors. Maybe to something like `right_thermal_cg_stressform`?

@c-p-schmidt: Maybe I don't understand this correctly, but how would switching to tensors solve the underlying naming problem? That this would for example avoid naming the "multiplied" tensors?

rjoussen: I dont think that would be a problem necessarily, since it is clear what it mean inside the container it is stored in. But I also find the names too lenghty this way, so I am going to revert the name change.
@dragos-ana Are you okay if I just leave the name as CT? The doxygen comments clearly state what is meant.

@c-p-schmidt I agree, I think a lot of the naming issues could be solved by using tensors also (Just because a lot fewer names are needed). Switching to tensors won't be easy, though. If we want to do that, we really need to think about where to start.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
rjoussen: Mainly because everything else was in the header before already. Since it is a service file, I think it should be intended to be included by one file only, so it shouldnt matter. However, this is clearly violated already, see my other comment. Do you want me to put the new implementations in a new cpp file?

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
dragos-ana: Yes, you can leave it like this until we find a better solution for naming the variables.

I also fully agree that switching to tensors is not straightforward. Hence, my upcoming PRs will use matrices for now, and we will need to do a separate transition to tensors where we also adapt the variable names.

rjoussen: Done.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
c-p-schmidt: I prefer headers that are rather clean. However, since this file is also affected by upcoming restructurings, we can address this during those restructurings.

c-p-schmidt (approved):

dragos-ana (approved):

PR #2039, opened 2026-05-20
Collect benchmark test data and generate report

amgebauer:
Collect benchmark data over time and generate a report using the same infrastructure as for our performance tests.

To standardize data reporting, I modified the performance test collection script to use our own collection file format. Note that this has the big benefit of not silently breaking our report if Trilinos decides to change how it writes time monitor data. Additionally, we can reuse all the report-generation infrastructure for any other kind of data (as we do here for benchmark tests).

The generated report looks like so: https://amgebauer.github.io/4C/benchmark_test_report/report.html


c-p-schmidt (dismissed):
Looks reasonable to me.
Thank you for enabling this for the benchmark test data as well!

cmake/functions/four_c_auto_define_benchmark_tests.cmake
```diff
@@ -30,15 +30,30 @@ function(_set_up_benchmark_test_target _module_under_test _target)
 
   target_link_libraries(${_target} PRIVATE unittests_common)
 
+  # configure the respective input test with the respective mesh
+  if(FOUR_C_ENABLE_FULL_BENCHMARK_TESTS)
+    set(benchmark_test_additional_options "")
+    set(benchmark_test_timeout 600) # 10 minutes
```
davidrudlstorfer: does it make sense to hardcode this timeout here? maybe use the same approach and define the timeout in the same place as the unittest_timeout?

utilities/four_c_python/src/four_c_building_testing/collect_benchmark_test_results.py
davidrudlstorfer, line 1: I do not have a really strong opinion here, maybe it makes sense to add a few line breaks in the really really long doc strings or f strings

davidrudlstorfer (dismissed):
Looks nice, especially the custom "file format"/file structure without relying on Trilinos.

I just have two minor comments, but you can happily merge immediately:)

amgebauer:
Thanks for your hints @davidrudlstorfer. I added new lines for better readability and moved the timeout definition into `setup_tests.cmake`, as we do for the unit tests.

c-p-schmidt (approved):

davidrudlstorfer (approved):
Thanks a lot for going the extra mile:)

PR #2040, opened 2026-05-21
Enable possibility to use functions for body force contributions in porofluid_pressure_based_elast* framework

lkoeglmeier:
This pr enables the use of functions for adding body force contributions to the solid equation in the porofluid_pressure_based_elast* framework.

### DIFF
--- a/src/porofluid_pressure_based_elast/4C_porofluid_pressure_based_elast_input.cpp
+++ b/src/porofluid_pressure_based_elast/4C_porofluid_pressure_based_elast_input.cpp
@@ -48,9 +48,10 @@ std::vector<Core::IO::InputSpec> PoroPressureBased::valid_parameters_porofluid_e
               {.required = false}),
 
           // body force contribution for solid part
-          parameter<std::optional<std::vector<double>>>("body_force",
-              {.description = "External body force contribution vector. The vector must "
-                              "have the same dimension as the problem (2D/3D)."}),
+          parameter<std::optional<int>>("body_force_function",
+              {.description = "Function defining the external body force contribution vector. The "
+                              "vector must have the same dimension (number of components) as the "
+                              "problem (2D/3D)."}),
 
           // nonlinear solver
           group("nonlinear_solver",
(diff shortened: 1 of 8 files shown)


src/porofluid_pressure_based_elast/4C_porofluid_pressure_based_elast_input.cpp
```diff
@@ -48,9 +48,11 @@ std::vector<Core::IO::InputSpec> PoroPressureBased::valid_parameters_porofluid_e
               {.required = false}),
 
           // body force contribution for solid part
-          parameter<std::optional<std::vector<double>>>("body_force",
-              {.description = "External body force contribution vector. The vector must "
-                              "have the same dimension as the problem (2D/3D)."}),
+          parameter<std::optional<int>>("body_force_function",
+              {.description =
+                      "Function defining the external body force contribution vector. The vector "
+                      "must "
```
lauraengelhardt: There seems to be a formatting issue here.

src/porofluid_pressure_based_elast/4C_porofluid_pressure_based_elast_input.cpp
```diff
@@ -48,9 +48,11 @@ std::vector<Core::IO::InputSpec> PoroPressureBased::valid_parameters_porofluid_e
...
+                      "have the same dimension (number of components) as the problem (2D/3D)."}),
```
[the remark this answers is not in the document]
c-p-schmidt: You can decide best, whether it is valuable to support both, probably not as a constant function can reproduce the original behavior.
However, it is a breaking change, so I think the corresponding label should be added.

src/solid_poro_ele/4C_solid_poro_ele_pressure_based.hpp
```diff
@@ -220,9 +215,6 @@ namespace Discret::Elements
     //! flag, whether the post setup of materials is already called
     bool material_post_setup_ = false;
 
-    //! optional body force contribution
-    std::optional<Core::LinAlg::Tensor<double, dim>> bodyforce_contribution_{};
```
c-p-schmidt, line 224: I think it makes sense to store the function now instead. Then you do not need to deduce it every time from the global problem, but only once at the beginning.

lauraengelhardt: I would do that as well and then simplify by removing the function `get_bodyforce_function_from_input`. You can then just directly use `bodyforce_contribution_function_`.

c-p-schmidt (approved):
Looks good to me!

bgoderbauer (approved):

PR #2041, opened 2026-05-21
Add thermal effects to the viscoplastic factor

rjoussen:
## Description

This PR adds thermal effects to the viscoplastic inelastic deformation-gradient factor, including thermal expansion and Taylor-Quinney mechanical dissipation.

The new `evaluate_mechanical_dissipation` method returns the Taylor-Quinney heat source

$$
R_\mathrm{TQ} = \xi_\mathrm{TQ}\cdot\sigma_\mathrm{eq}\cdot\dot{\varepsilon}_\mathrm{p}
$$

together with its linearizations with respect to temperature and the right Cauchy-Green tensor. In the viscoplastic state evaluation, the mechanical driving stress is changed from the elastic Mandel stress to the thermo-elastic Mandel stress, i.e. the thermal-expansion stress contribution is subtracted.

The main challenge is that `evaluate_mechanical_dissipation` adds another public evaluation path to the material, called from the thermo field. Depending on the coupling algorithm, for example monolithic vs. partitioned, this call may happen before or after the solid-side evaluations and may be requested for a different deformation-gradient/temperature state. Therefore, the thermo-side evaluation must be independent of the solid-side call order. At the same time, re-running return mapping or the history-variable linearization for an unchanged state would be wasteful.

To handle this, the PR adds a `ThermoMechanicalCouplingCache` that stores all quantities needed for thermo-mechanical linearization per Gauss point. A changed incoming state invalidates the cache, and both solid-side and thermo-side evaluations can populate it. The new `evaluate_history_variables_wrt*` helpers either fill the cache or return cached values if they are already valid. `evaluate_additional_cmat`, `evaluate_od_stiff_mat`, and `evaluate_mechanical_dissipation` all use this shared path.

The new `ThermoViscoplastPublicEvaluationsCanBeCalledInAnyOrder` test checks that solid and thermo evaluations can be called in arbitrary order, and that `evaluate_mechanical_dissipation` safely reruns return mapping when the incoming state requires it.

All relevant new linearizations are covered by unit tests against finite differences. For this comparison, `evaluate_additional_cmat` was switched to central finite differences in the perturbation-based path. The thermo-viscoplastic `state_quantities` and `state_quantity_derivatives` are also tested against analytically obtained reference values.

### Current limitation
- thermoelastic coupling with transversely isotropic elastic summands is not supported yet for non-reference temperatures (it throws).
- So far, these are only changes to the factor, it is not wired into the parent material yet. This will be the next PR.

Thanks to @dragos-ana for providing especially the linearizations of the history variables wrt temperature.

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
@@ -2652,6 +2652,12 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                         .default_value = std::vector{0},
                         .size = from_parameter<int>("NUMFACINEL")}),
                 parameter<double>("DENS", {.description = "material mass density"}),
+                parameter<double>("REF_TEMPERATURE",
+                    {.description = "reference temperature for thermoelastic expansion.",
+                        .default_value = 0.0}),
+                parameter<double>("THERMAL_EXPANSION_COEFFICIENT",
+                    {.description = "coefficient of thermal expansion $\\alpha_T$",
+                        .default_value = 0.0}),
             },
             {.description = "multiplicative split of deformation gradient"});
   }
@@ -2831,6 +2837,12 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 "FIBER_READER_ID", {.description = "MAT ID of the used fiber direction reader for "
                                                    "transversely isotropic behavior",
                                        .validator = positive<int>()}),
+            parameter<double>("TAYLOR_QUINNEY_COEFFICIENT",
+                {.description =
+                        "Taylor-Quinney coefficient $\\xi_{TQ}$ modeling the internal dissipation",
+                    .default_value = 0.0,
+                    .validator =
+                        Core::IO::InputSpecBuilders::Validators::positive_or_zero<double>()}),
             parameter<std::optional<double>>(
                 "YIELD_COND_A", {.description = "transversely isotropic version of the Hill(1948) "
                                                 "yield condition: parameter A, following the "
--- a/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
@@ -2654,7 +2654,8 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 parameter<double>("DENS", {.description = "material mass density"}),
                 parameter<double>("REF_TEMPERATURE",
                     {.description = "reference temperature for thermoelastic expansion.",
-                        .default_value = 0.0}),
+                        .default_value = 0.0,
+                        .validator = Validators::positive_or_zero<double>()}),
                 parameter<double>("THERMAL_EXPANSION_COEFFICIENT",
                     {.description = "coefficient of thermal expansion $\\alpha_T$",
                         .default_value = 0.0}),
@@ -2841,8 +2842,7 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 {.description =
                         "Taylor-Quinney coefficient $\\xi_{TQ}$ modeling the internal dissipation",
                     .default_value = 0.0,
-                    .validator =
-                        Core::IO::InputSpecBuilders::Validators::positive_or_zero<double>()}),
+                    .validator = positive_or_zero<double>()}),
             parameter<std::optional<double>>(
                 "YIELD_COND_A", {.description = "transversely isotropic version of the Hill(1948) "
                                                 "yield condition: parameter A, following the "
--- a/src/mat/4C_mat_inelastic_defgrad_factors.cpp
+++ b/src/mat/4C_mat_inelastic_defgrad_factors.cpp
@@ -2906,7 +2906,7 @@ Mat::HeatSource
 Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_taylor_quinney_heat_source(
     const Mat::EvaluationContext<3>& context, const int gp, const int eleGID,
     const Core::LinAlg::Matrix<3, 3>* defgrad, const Core::LinAlg::Matrix<3, 3>& iFin_other,
-    const double& temperature)
+    const double temperature)
 {
   const ReducedKinematics reduced_kinematics = evaluate_reduced_kinematics(*defgrad, iFin_other);
 
--- a/src/mat/4C_mat_inelastic_defgrad_factors.cpp
+++ b/src/mat/4C_mat_inelastic_defgrad_factors.cpp
@@ -8,12 +8,15 @@
 #include "4C_mat_inelastic_defgrad_factors.hpp"
 
 #include "4C_comm_mpi_utils.hpp"
+#include "4C_fem_discretization.hpp"
 #include "4C_global_data.hpp"
 #include "4C_legacy_enum_definitions_materials.hpp"
 #include "4C_linalg_fixedsizematrix.hpp"
 #include "4C_linalg_fixedsizematrix_solver.hpp"
 #include "4C_linalg_fixedsizematrix_tensor_products.hpp"
 #include "4C_linalg_fixedsizematrix_voigt_notation.hpp"
+#include "4C_linalg_four_tensor.hpp"
+#include "4C_linalg_serialdensematrix.hpp"
 #include "4C_linalg_symmetric_tensor.hpp"
 #include "4C_linalg_tensor.hpp"
 #include "4C_linalg_tensor_conversion.hpp"
@@ -26,6 +29,7 @@
 #include "4C_mat_electrode.hpp"
 #include "4C_mat_inelastic_defgrad_factors_service.hpp"
 #include "4C_mat_multiplicative_split_defgrad_elasthyper.hpp"
+#include "4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp"
 #include "4C_mat_par_bundle.hpp"
 #include "4C_mat_vplast_law.hpp"
 #include "4C_utils_enum.hpp"
@@ -58,6 +62,21 @@ namespace
     Core::LinAlg::Matrix<3, 3> diff{Core::LinAlg::Initialization::zero};
     diff.update(1.0, lhs, -1.0, rhs, 0.0);
     return diff.norm2();
+  }
+
+  struct ReducedKinematics
+  {
+    Core::LinAlg::Matrix<3, 3> defgrad{Core::LinAlg::Initialization::zero};
+    Core::LinAlg::Matrix<3, 3> right_cauchy_green{Core::LinAlg::Initialization::zero};
+  };
+
+  ReducedKinematics evaluate_reduced_kinematics(const Core::LinAlg::Matrix<3, 3>& defgrad,
+      const Core::LinAlg::Matrix<3, 3>& inverse_other_inelastic_defgrad)
+  {
+    ReducedKinematics result;
+    result.defgrad.multiply_nn(1.0, defgrad, inverse_other_inelastic_defgrad, 0.0);
+    result.right_cauchy_green.multiply_tn(1.0, result.defgrad, result.defgrad, 0.0);
+    return result;
   }
 
   // read input parameter container of parent material (i.e., underlying
@@ -367,6 +386,32 @@ namespace
     return diFin_dC_V;
   }
 
+  /**
+   * @brief Extract the derivative of the plastic strain w.r.t. right CG tensor from the
+   solution of the linear system of equations arising in the additional cmat calculation.
+   *
+   * @param SOL
+   * @return Core::LinAlg::Matrix<1, 6> \f[ \frac{\mathrm{d} \varepsilon_p}{\mathrm{d}
+   \boldsymbol{C}}
+   \f] in Voigt Notation
+   */
+  Core::LinAlg::Matrix<1, 6> extract_derivative_of_plastic_strain(
+      const Core::LinAlg::Matrix<10, 6>& SOL)
+  {
+    // declare output derivative
+    Core::LinAlg::Matrix<1, 6> depsp_dC_V(Core::LinAlg::Initialization::zero);
+
+    // set its components
+    depsp_dC_V(0, 0) = SOL(9, 0);
+    depsp_dC_V(0, 1) = SOL(9, 1);
+    depsp_dC_V(0, 2) = SOL(9, 2);
+    depsp_dC_V(0, 3) = SOL(9, 3);
+    depsp_dC_V(0, 4) = SOL(9, 4);
+    depsp_dC_V(0, 5) = SOL(9, 5);
+
+    return depsp_dC_V;
+  }
+
   // wrap inverse inelastic defgrad and plastic strain to a vector of unknowns for the Local
   // Newton Loop (helper function: InelasticDefgradTransvIsotropElastViscoplast)
   Core::LinAlg::Matrix<10, 1> wrap_unknowns(
@@ -638,6 +683,7 @@ Mat::PAR::InelasticDefgradTransvIsotropElastViscoplast::
     InelasticDefgradTransvIsotropElastViscoplast(const Core::Mat::PAR::Parameter::Data& matdata)
     : Parameter(matdata),
       viscoplastic_law_id_(matdata.parameters.get<int>("VISCOPLAST_LAW_ID")),
+      taylor_quinney_coefficient_(matdata.parameters.get<double>("TAYLOR_QUINNEY_COEFFICIENT")),
       fiber_reader_gid_(matdata.parameters.get<int>("FIBER_READER_ID")),
       yield_cond_a_(matdata.parameters.get<std::optional<double>>("YIELD_COND_A").value_or(0.0)),
       yield_cond_b_(matdata.parameters.get<std::optional<double>>("YIELD_COND_B").value_or(0.0)),
@@ -823,9 +869,15 @@ std::shared_ptr<Mat::InelasticDefgradFactors> Mat::InelasticDefgradFactors::fact
         }
       }
 
+      const double thermal_expansion_coefficient =
+          parentmat_input_params.get<double>("THERMAL_EXPANSION_COEFFICIENT");
+
+      const auto ref_temperature = parentmat_input_params.get<double>("REF_TEMPERATURE");
+
       // return shared pointer to the inelastic factor
-      return std::make_shared<InelasticDefgradTransvIsotropElastViscoplast>(
-          params, viscoplastic_law, fiber_reader, potsumel, potsumel_transviso);
+      return std::make_shared<InelasticDefgradTransvIsotropElastViscoplast>(params,
+          viscoplastic_law, fiber_reader, potsumel, potsumel_transviso,
+          thermal_expansion_coefficient, ref_temperature);
     }
 
     default:
@@ -1731,8 +1783,11 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::InelasticDefgradTransvIsotrop
     Core::Mat::PAR::Parameter* params, std::shared_ptr<Mat::Viscoplastic::Law> viscoplastic_law,
     Mat::Elastic::CoupTransverselyIsotropic fiber_reader,
     std::vector<std::shared_ptr<Mat::Elastic::Summand>> pot_sum_el,
-    std::vector<std::shared_ptr<Mat::Elastic::CoupTransverselyIsotropic>> pot_sum_el_transv_iso)
+    std::vector<std::shared_ptr<Mat::Elastic::CoupTransverselyIsotropic>> pot_sum_el_transv_iso,
+    const double thermal_expansion_coefficient, const double ref_temperature)
     : InelasticDefgradFactors(params),
+      thermal_expansion_coefficient_(thermal_expansion_coefficient),
+      ref_temperature_(ref_temperature),
       potsumel_(std::move(pot_sum_el)),
       potsumel_transviso_(std::move(pot_sum_el_transv_iso)),
       viscoplastic_law_(std::move(viscoplastic_law)),
@@ -1750,7 +1805,7 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::InelasticDefgradTransvIsotrop
   time_step_tracker_.min_dt = 0.0;
 
   // initialize time step quantities
-  time_step_quantities_.init();
+  time_step_quantities_.init(ref_temperature_);
 }
 
 
@@ -1768,6 +1823,23 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::pre_evaluate(
 
   // set element ID
   ele_gid_ = eleGID;
+
+
+  const double previous_temperature = time_step_quantities_.current_temperature[gp];
+
+  // set current absolute temperature
+  time_step_quantities_.current_temperature[gp] = ref_temperature_;
+  if (params.isParameter("temperature"))
+  {
+    time_step_quantities_.current_temperature[gp] = params.get<double>("temperature");
+  }
+
+  // invalidate the caches if the temperature changed
+  if (std::abs(time_step_quantities_.current_temperature[gp] - previous_temperature) > 1.0e-16)
+  {
+    time_step_quantities_.current_defgrad[gp].clear();
+    thermo_mechanical_coupling_cache_.reset(gp);
+  }
 
   // set time step
   FOUR_C_ASSERT(context.time_step_size, "Time step size not given in evaluation context.");
@@ -1835,8 +1907,9 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::calculate_gamma_delta(
  *--------------------------------------------------------------------*/
 ViscoplastUtils::StateQuantities
 Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
-    const Core::LinAlg::Matrix<3, 3>& CM, const Core::LinAlg::Matrix<3, 3>& iFinM,
-    const double plastic_strain, ViscoplastUtils::ErrorType& err_status, const double dt,
+    const Core::LinAlg::Matrix<3, 3>& CM, const double temperature,
+    const Core::LinAlg::Matrix<3, 3>& iFinM, const double plastic_strain,
+    ViscoplastUtils::ErrorType& err_status, const double dt,
     const ViscoplastUtils::StateQuantityEvalType& eval_type) const
 {
   ViscoplastUtils::StateQuantities state_quantities{};
@@ -1873,6 +1946,9 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
     Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3> dSedCe{};
     for (const auto& p : potsumel_transviso_)
     {
+      FOUR_C_ASSERT_ALWAYS(std::abs(temperature - ref_temperature_) <= 1.0e-16,
+          "Thermoelastic coupling with transversely isotropic elastic summands is not supported "
+          "yet. Use isotropic elastic summands only.");
       p->add_stress_aniso_principal(CeV, dSedCe, SeV, param_list, gp_, ele_gid_);
     }
     state_quantities.curr_dSedCe.clear();
@@ -1884,11 +1960,38 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
   Core::LinAlg::Matrix<3, 3> CeCeM(Core::LinAlg::Initialization::zero);
   CeCeM.multiply_nn(1.0, state_quantities.curr_CeM, state_quantities.curr_CeM, 0.0);
 
-  // compute symmetric part of Mandel stress tensor
+  // thermal quantities and stress factors
+  Mat::ThermalQuantities thermal_quantities =
+      Mat::evaluate_thermal_quantities(temperature - ref_temperature_,
+          thermal_expansion_coefficient_, iFinM, gp_, ele_gid_, potsumel_);
+  Mat::StressFactors thermal_stress_factors;
+  Mat::calculate_gamma_delta(thermal_stress_factors.gamma, thermal_stress_factors.delta,
+      thermal_quantities.prinv, thermal_quantities.dPI, thermal_quantities.ddPII);
+
+  // compute mixed thermo-elastic tensors required for reducing the Mandel stress based on
+  // temperature
+  Core::LinAlg::Matrix<3, 3> CT{Core::LinAlg::Initialization::zero};
+  Core::LinAlg::Voigt::Stresses::vector_to_matrix(thermal_quantities.CTV, CT);
+  Core::LinAlg::Matrix<3, 3> iCT{Core::LinAlg::Initialization::zero};
+  Core::LinAlg::Voigt::Stresses::vector_to_matrix(thermal_quantities.iCTV, iCT);
+  Core::LinAlg::Matrix<3, 3> CeCT{Core::LinAlg::Initialization::zero};
+  CeCT.multiply_nn(1.0, state_quantities.curr_CeM, CT, 0.0);
+  Core::LinAlg::Matrix<3, 3> CeiCT{Core::LinAlg::Initialization::zero};
+  CeiCT.multiply_nn(1.0, state_quantities.curr_CeM, iCT, 0.0);
+
+
+  /**
+  compute symmetric part of thermo-elastic Mandel stress tensor
+  \f[\boldsymbol{M}_{\theta,\text{sym}}  = \boldsymbol{C}_e \cdot \left( \boldsymbol{S}_{e} -
+  \boldsymbol{S}_T \right)\f]
+   */
   Core::LinAlg::Matrix<3, 3> Mtheta_sym_M(Core::LinAlg::Initialization::zero);
   Mtheta_sym_M.update(state_quantities.curr_gamma(0), state_quantities.curr_CeM,
       state_quantities.curr_gamma(1), CeCeM, 0.0);
   Mtheta_sym_M.update(state_quantities.curr_gamma(2), const_non_mat_tensors.id3x3, 1.0);
+  Mtheta_sym_M.update(-1.0 * thermal_stress_factors.gamma(0), state_quantities.curr_CeM, 1.0);
+  Mtheta_sym_M.update(-1.0 * thermal_stress_factors.gamma(1), CeCT, 1.0);
+  Mtheta_sym_M.update(-1.0 * thermal_stress_factors.gamma(2), CeiCT, 1.0);
   if (parameter()->mat_behavior() == ViscoplastUtils::MatBehavior::transv_isotropic)
   {
     Core::LinAlg::Matrix<3, 3> addMeM(Core::LinAlg::Initialization::zero);
@@ -2078,8 +2181,9 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
  *--------------------------------------------------------------------*/
 ViscoplastUtils::StateQuantityDerivatives
 Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantity_derivatives(
-    const Core::LinAlg::Matrix<3, 3>& CM, const Core::LinAlg::Matrix<3, 3>& iFinM,
-    const double plastic_strain, ViscoplastUtils::ErrorType& err_status, const double dt,
+    const Core::LinAlg::Matrix<3, 3>& CM, const double temperature,
+    const Core::LinAlg::Matrix<3, 3>& iFinM, const double plastic_strain,
+    ViscoplastUtils::ErrorType& err_status, const double dt,
     const ViscoplastUtils::StateQuantityDerivEvalType& eval_type, const bool eval_state) const
 {
   ViscoplastUtils::StateQuantityDerivatives state_quantity_derivatives{};
@@ -2091,13 +2195,14 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantity_deriv
   Core::LinAlg::Matrix<6, 9> temp6x9(Core::LinAlg::Initialization::zero);
   Core::LinAlg::Matrix<9, 6> temp9x6(Core::LinAlg::Initialization::zero);
   Core::LinAlg::Matrix<9, 9> temp9x9(Core::LinAlg::Initialization::zero);
+  Core::LinAlg::FourTensor<3> temp_four_tensor(true);
 
   // check whether to reevaluate the state or to keep the available state quantity values
   ViscoplastUtils::StateQuantities relevant_state_quantities = state_quantities_;
   if (eval_state)
   {
-    relevant_state_quantities = evaluate_state_quantities(CM, iFinM, plastic_strain, err_status, dt,
-        ViscoplastUtils::StateQuantityEvalType::full_eval);
+    relevant_state_quantities = evaluate_state_quantities(CM, temperature, iFinM, plastic_strain,
+        err_status, dt, ViscoplastUtils::StateQuantityEvalType::full_eval);
   }
 
   // get the state quantities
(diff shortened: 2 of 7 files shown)


src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2524,49 +2710,374 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+    RHS(9) = rhs_epsp_V(0);
+
+    // solve the linear system of equations
+    Core::LinAlg::Matrix<10, 1> SOL(Core::LinAlg::Initialization::zero);
+    Core::LinAlg::FixedSizeSerialDenseSolver<10, 10, 1> solver;
+    solver.set_matrix(jacMat);     // set A = jacM
+    solver.set_vectors(SOL, RHS);  // set X=SOL, B=RHS
+    solver.factor_with_equilibration(true);
+    int err = solver.solve();  // X = A^-1 B
+    int err2 = solver.factor();
```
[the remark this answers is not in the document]
rjoussen: @dragos-ana Do you know if that's right? It's really hard to find out what this is actually doing tbh.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -823,9 +869,15 @@ std::shared_ptr<Mat::InelasticDefgradFactors> Mat::InelasticDefgradFactors::fact
         }
       }
 
+      const double thermal_expansion_coefficient =
+          parentmat_input_params.get<double>("THERMAL_EXPANSION_COEFFICIENT");
```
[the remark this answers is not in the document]
rjoussen: Both values have default values in the parent material, so this is not an issue. I changed the `REF_TEMPERATURE` to `get<double>` now also.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -2229,6 +2354,9 @@ namespace
         set_up_viscoplastic_material({.mat_behavior = ViscoplastUtils::MatBehavior::isotropic})
             .material;
 
+    // refernence temperature with which the materials were created.
```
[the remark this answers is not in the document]
rjoussen: I changed it to reading the reference temperature from the default setup now.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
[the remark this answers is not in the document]
dragos-ana: I think factoring is a pre-processing / decomposition step before solving, so I would say that factor - solve is the right sequence. It's also like this in most of the code, AFAIK.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2524,49 +2710,374 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+                             .inv_plastic_defgrad_wrt_cauchy_green;
+
+  if (err_status != ViscoplastUtils::ErrorType::no_errors)
+  {
+    // if we get an error in evaluating the history variable derivatives: perform
+    // perturbation-based linearization
+    std::cout << "Warning: Error encountered in evaluating history variable derivatives: "
+              << ViscoplastUtils::get_detailed_error_message_for_error_type(err_status) << "\n"
+              << "Performing perturbation-based linearization for the additional stiffness "
+                 "contribution.\n";
```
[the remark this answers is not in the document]
rjoussen: I put this warning because I was quite surprised that this was a fallback even, and I think if this fails (but the perturbation-based linearization succeeds), it's definitely a hint that something is wrong. Hence the warning. In my experience, though, the linearization calculation isn't the issue most of the time, so maybe we could remove this fallback altogether.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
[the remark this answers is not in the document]
rjoussen: Then I'll change it in the linearization wrt to Cauchy Green as well.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2831,6 +2837,12 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 "FIBER_READER_ID", {.description = "MAT ID of the used fiber direction reader for "
                                                    "transversely isotropic behavior",
                                        .validator = positive<int>()}),
+            parameter<double>("TAYLOR_QUINNEY_COEFFICIENT",
+                {.description =
+                        "Taylor-Quinney coefficient $\\xi_{TQ}$ modeling the internal dissipation",
+                    .default_value = 0.0,
+                    .validator =
+                        Core::IO::InputSpecBuilders::Validators::positive_or_zero<double>()}),
```
c-p-schmidt: Just this should also work as for the parameter above.
```suggestion
                        positive_or_zero<double>()}),
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -60,6 +64,21 @@ namespace
     return diff.norm2();
   }
 
+  struct ReducedKinematics
```
c-p-schmidt, line 71: Some documentation of what `ReducedKinematics` is or stands for might help. Without further description, the names defgrad and also right Cauchy green are dangerous because I would expect for defgrad F and for right Cauchy Green C=F^TF, while here you store F' = F*Fin^[-1] and C' = F'^T F'.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -367,6 +386,32 @@ namespace
     return diFin_dC_V;
   }
 
+  /**
+   * @brief Extract the derivative of the plastic strain w.r.t. right CG tensor from the
+   solution of the linear system of equations arising in the additional cmat calculation.
+   *
+   * @param SOL
```
c-p-schmidt: Some docu what SOL is would be nice.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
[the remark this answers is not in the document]
c-p-schmidt: I would leave that decision up to you and @dragos-ana, as you are mostly using this implementation.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2524,49 +2710,374 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+  ViscoplastUtils::ErrorType err_status = ViscoplastUtils::ErrorType::no_errors;
+
+  gp_ = gp;
+
+  if (parameter()->linearization_type() == ViscoplastUtils::LinearizationType::perturbation_based)
+  {
+    return evaluate_mechanical_dissipation_perturb_based(reduced_kinematics.defgrad, temperature);
+  }
+
+  const double& tol = ViscoplastUtils::thermo_mechanical_state_equality_tolerance;
```
c-p-schmidt: You don't need the reference, but very minor. ;-)
```suggestion
  constexpr double tol = ViscoplastUtils::thermo_mechanical_state_equality_tolerance;
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3698,31 +4188,189 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+    const Core::LinAlg::Matrix<3, 3>& FredM, const double temperature,
+    Core::LinAlg::Matrix<6, 1>& dstressdT, const Core::LinAlg::Matrix<6, 9>& dSdiFinj)
+{
+  update_hist_var_ = false;
+  const auto unperturbed_state_quantities = state_quantities_;
+  const auto unperturbed_time_step_quantities = time_step_quantities_;
+
+  const double perturbation_factor = 1.0e-7;
+  const double delta_T_perturbation =
+      std::max(perturbation_factor * std::abs(temperature), perturbation_factor);
```
c-p-schmidt: Can the temperature be negative here?

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2652,6 +2652,12 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                         .default_value = std::vector{0},
                         .size = from_parameter<int>("NUMFACINEL")}),
                 parameter<double>("DENS", {.description = "material mass density"}),
+                parameter<double>("REF_TEMPERATURE",
+                    {.description = "reference temperature for thermoelastic expansion.",
+                        .default_value = 0.0}),
```
dragos-ana: The reference temperature should be >= 0, right? For our applications also the thermal expansion coefficient, but there we can leave it open still.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -172,18 +259,24 @@ namespace Mat
...
-      /*!
+      /**
        * @brief Set meaningful initial values. Done first for one single Gauss point (extended later
        * on using the resizing function).
        *
+       * @param ref_temperature Reference temperature used to set initial values of last/current
+       * temperature.
        */
-      void init();
+      void init(const double& ref_temperature);
```
dragos-ana:
```suggestion
      void init(const double ref_temperature);
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -442,18 +535,25 @@ namespace Mat
...
+      //! derivatives of the equivalent tensile stress w.r.t. the temperature
+      double curr_dequiv_stress_dT{0.0};
 
       //! derivative of the deviatoric, symmetric part of the thermo-elastic Mandel stress tensor
       //! w.r.t. the inverse inelastic deformation gradient (Voigt stress form)
       Core::LinAlg::Matrix<6, 9> curr_dMtheta_dev_sym_diFin{Core::LinAlg::Initialization::zero};
       //! derivative of the deviatoric, symmetric part of the thermo-elastic Mandel stress tensor
       //! w.r.t. the right Cauchy-Green deformation tensor (Voigt stress-stress form)
       Core::LinAlg::Matrix<6, 6> curr_dMtheta_dev_sym_dC{Core::LinAlg::Initialization::zero};
+      //! derivative of the deviatoric, symmetric part of the Mandel stress tensor w.r.t. the
```
dragos-ana:
```suggestion
      //! derivative of the deviatoric, symmetric part of the thermo-elastic Mandel stress tensor w.r.t. the
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -494,16 +602,189 @@ namespace Mat
...
-      double deriv_temperature;
+      double deriv_temperature = 0.0;
+    };
+
+    /// Derivatives of the history variables wrt. the right Cauchy-Green deformation tensor
+    struct HistoryVariablesDerivativesWrtCauchyGreen
+    {
+      //! derivative of the inverse plastic deformation gradient w.r.t. the right Cauchy-Green
+      //! tensor \f$ \frac{\mathrm{d}\boldsymbol{F}_\mathrm{p}^{-1}}{\mathrm{d}\boldsymbol{C}} \f$
+      //! in Voigt notation (second dimension in in stress-form)
```
dragos-ana:
```suggestion
      //! derivative of the inverse plastic deformation gradient w.r.t. the right Cauchy-Green
      //! tensor \f$ \frac{\mathrm{d}\boldsymbol{F}_{\mathrm{p}^{-1},n+1}}{\mathrm{d}\boldsymbol{C}_{n+1}} \f$
      //! in Voigt notation (second dimension in stress-form)
```
Just for more clarity, since this is an algorithmic / discretized derivative

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -494,16 +602,189 @@ namespace Mat
...
+      auto quantities()
+      {
+        return std::tie(state, state_derivatives, history_variables_wrt_cauchy_green,
+            history_variables_wrt_temperature);
+      }
+    };
+
+
+    /**
+     * Returns the derivative of the taylor-quinney term wrt. the right Cauchy-Green tensor
```
dragos-ana:
```suggestion
     * Returns the derivative of the Taylor-Quinney term wrt. the right Cauchy-Green tensor
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -494,16 +602,189 @@ namespace Mat
...
+     * =\xi_\mathrm{TQ}\left(
+     * \frac{\mathrm{d}\bar{\sigma}}{\mathrm{d}\mathbf{C}}\,\dot{\varepsilon}_\mathrm{p}
+     * +\bar{\sigma}\,\frac{\mathrm{d}\dot{\varepsilon}_\mathrm{p}}
+     * {\mathrm{d}\mathbf{C}}\right)\f]
+     *
+     * with
+     * \f[\frac{\mathrm{d}\bar{\sigma}}{\mathrm{d}\mathbf{C}}
+     * =\frac{\partial \bar{\sigma}}{\partial \mathbf{C}}
+     * +\frac{\partial \bar{\sigma}}{\partial \mathbf{F}_\mathrm{p}^{-1}}
+     * :\frac{\partial \mathbf{F}_\mathrm{p}^{-1}}{\partial \mathbf{C}}\f]
```
dragos-ana: Same as above in regards to the algorithmic derivatives, although your documentation is really great already!

src/mat/4C_mat_inelastic_defgrad_factors_service.cpp
```diff
@@ -306,13 +338,63 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::TimeStepQuantities:
...
+{
+  Core::LinAlg::Matrix<1, 6> total_dequiv_stress_dC =
+      state_derivatives.equiv_stress_wrt_cauchy_green;
+  total_dequiv_stress_dC.multiply(1.0, state_derivatives.equiv_stress_wrt_inverse_plastic_defgrad,
+      history_variables_derivatives.inv_plastic_defgrad_wrt_cauchy_green, 1.0);
+
+  Core::LinAlg::Matrix<1, 6> total_dpsr_dC = total_dequiv_stress_dC;
+  total_dpsr_dC.update(state_derivatives.plastic_strain_rate_derivs.deriv_plastic_strain,
+      history_variables_derivatives.plastic_strain_wrt_cauchy_green,
+      state_derivatives.plastic_strain_rate_derivs.deriv_equiv_stress);
```
dragos-ana: This is a bit difficult to understand in my opinion. Maybe initializing it as 0 first, and then using two update calls for both derivative components would be more readable?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -1769,6 +1824,23 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::pre_evaluate(
...
+  time_step_quantities_.current_temperature[gp] = ref_temperature_;
+  if (params.isParameter("temperature"))
+  {
+    time_step_quantities_.current_temperature[gp] = params.get<double>("temperature");
+  }
+
+  // invalidate the caches if the temperature changed
+  if (std::abs(time_step_quantities_.current_temperature[gp] - previous_temperature) > 1.0e-16)
+  {
+    time_step_quantities_.current_defgrad[gp].clear();
```
dragos-ana: This seems a bit strange, why does a new temperature invalidate the stored deformation gradient?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2190,10 +2295,29 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantity_deriv
...
+  Mat::ThermalQuantities thermal_quantities =
+      Mat::evaluate_thermal_quantities(temperature - ref_temperature_,
+          thermal_expansion_coefficient_, iFinM, gp_, ele_gid_, potsumel_);
+  Core::LinAlg::SymmetricTensor<double, 3, 3> hyperelast_stress_CT{};
+  Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3> hyperelast_stiffness_CT{};
+  Mat::elast_hyper_add_isotropic_stress_cmat(hyperelast_stress_CT, hyperelast_stiffness_CT,
+      Core::LinAlg::make_symmetric_tensor_from_stress_like_voigt_matrix(thermal_quantities.CTV),
+      Core::LinAlg::make_symmetric_tensor_from_stress_like_voigt_matrix(thermal_quantities.iCTV),
+      thermal_quantities.prinv, thermal_quantities.dPI, thermal_quantities.ddPII);
+  const Core::LinAlg::Matrix<3, 3> S_T =
```
dragos-ana, line 2322:
```suggestion
  const Core::LinAlg::Matrix<3, 3> S_theta_T =
```

Maybe like this for more clarity?

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2211,9 +2335,21 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantity_deriv
   dMtheta_sym_diFin.multiply_nt(delta(4), CeCeV, CiFiniCeV, 1.0);
   dMtheta_sym_diFin.multiply_nt(delta(4), const_non_mat_tensors.id6x1, CiFinCeV, 1.0);
   dMtheta_sym_diFin.multiply_nt(delta(5), const_non_mat_tensors.id6x1, CiFiniCeV, 1.0);
+  // thermal contribution
+  temp_four_tensor.clear();
+  Core::LinAlg::FourTensorOperations::multiply_matrix_four_tensor_by_second_index<3>(
+      temp_four_tensor, S_T, dCediFin_FourTensor, true);
+  Core::LinAlg::Matrix<6, 9> dMT_sym_dFin{Core::LinAlg::Initialization::zero};
```
dragos-ana, line 2356:
```suggestion
  Core::LinAlg::Matrix<6, 9> dM_theta_T_sym_dFin{Core::LinAlg::Initialization::zero};
```
would maybe be slightly more accurate. But tbh, the naming of the variables is very suboptimal anyway, this will need to be addressed in the future when we switch to tensors.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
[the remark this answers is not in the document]
dragos-ana: I think that this was an issue for substepping at that time, where evaluating the linearization "in one go" (time n -> n+1) could pose problems, as opposed to evaluating the derivatives with smaller substeps and other previous states which was done so in the Local Newton (particularly for the Anand law where we perform a smaller Local Newton for the flow resistance on the viscoplastic law level). A completely correct approach for linearizing when substepping is used for the Local Newton would actually require building the linearization sequentially from the substep values WITHIN the Local Newton, and then simply returning the computed linearization here. 

Since substepping is, in my current experience, suboptimal anyway, we can remove the fallback for now by asserting an error-free status. If we encounter any problems we can think about better strategies in the future.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2524,49 +2710,374 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+  // compute additional term to stiffness matrix additional_cmat
+  cmatadd.multiply_nn(2.0, dSdiFinj, diFinjdC, 1.0);
+}
+
+ViscoplastUtils::HistoryVariablesDerivativesWrtTemperature
+Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_history_variables_wrt_temperature(
+    const Core::LinAlg::Matrix<3, 3>& CredM, const double temperature,
+    InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType& err_status)
+{
+  if (thermo_mechanical_coupling_cache_.history_variables_wrt_temperature.is_evaluated(gp_))
```
dragos-ana, line 2757: You could set the error status to no_errors here to ensure that this is not called with a "dirty" error state. 
I have a cleanup coming up where I ensure that all routines are called with a clean error state, and I will also include this part, but for now, it would be nice if setting to no_errors is already done here.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2524,49 +2710,374 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+    Core::LinAlg::Matrix<10, 10> jacMat(Core::LinAlg::Initialization::zero);
+    viscoplastic_law_->pre_evaluate(params_, gp_);  // set last_substep <- last_
+    jacMat = evaluate_local_newton_jacobian(CredM, temperature, current_sol,
+        time_step_quantities_.last_plastic_strain[gp_],
+        time_step_quantities_.last_plastic_defgrad_inverse[gp_], time_step_tracker_.dt, err_status);
+    FOUR_C_ASSERT_ALWAYS(
+        err_status == InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType::no_errors,
+        "Could not evaluate Jacobian in off-diagonal stiffness evaluation!");
+
+    // if we get singular Jacobian: throw exception -> go to FD-based linearization
```
dragos-ana:
```suggestion
    // if we get singular Jacobian: throw exception 
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2524,49 +2710,374 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
 
-      return;
+      // calculate RHS of the equation for the plastic strain
+      /** \f$\texttt{rhs\_epsp\_V} =\frac{\partial r_{\varepsilon_{\text{p}}}}{\partial T_{n+1}}
+      = - \Delta t \cdot \left(\underbrace{\frac{\partial \dot{\varepsilon}_{\text{p}}}{\partial
+      \sigma_{\text{yield}}} \cdot
+      \frac{\partial\sigma_{\text{yield}}}{\partial T}}_\texttt{curr\_dpsr\_dT} + \frac{\partial
+      v_{\text{p}}}{\partial \sigma_\text{eq}}
+      \cdot \frac{\partial \sigma_\text{eq}}{\partial T}\right)\f$
+      */
```
dragos-ana, line 2818:
```suggestion
      // calculate RHS of the equation for the plastic strain
      /** \f$\texttt{rhs\_epsp\_V} = - \frac{\partial r_{\varepsilon_{\text{p}}}}{\partial T_{n+1}}
      = \Delta t \cdot \left(\underbrace{\frac{\partial \dot{\varepsilon}_{\text{p}}}{\partial
      \sigma_{\text{yield}}} \cdot
      \frac{\partial\sigma_{\text{yield}}}{\partial T}}_\texttt{curr\_dpsr\_dT} + \frac{\partial
      v_{\text{p}}}{\partial \sigma_\text{eq}}
      \cdot \frac{\partial \sigma_\text{eq}}{\partial T}\right)\f$
      */
```

I like the addition of formulae very much!

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2524,49 +2710,374 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+    solver.set_matrix(jacMat);     // set A = jacM
+    solver.set_vectors(SOL, RHS);  // set X=SOL, B=RHS
+    solver.factor_with_equilibration(true);
+    int err2 = solver.factor();
+    int err = solver.solve();  // X = A^-1 B
+
+    if ((err != 0) || (err2 != 0))
+    {
+      err_status = ViscoplastUtils::ErrorType::failed_solution_analytic_linearization;
+      FOUR_C_THROW("Evaluation of linear system for off-diagonal stiffness has failed!");
```
dragos-ana:
```suggestion
      err_status = ViscoplastUtils::ErrorType::failed_solution_analytic_linearization;
      FOUR_C_THROW("Evaluation of linear system for off-diagonal stiffness has failed!: {}", get_detailed_error_message_for_error_type(err_status));
```

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
```diff
@@ -18,10 +19,39 @@
...
+
 FOUR_C_NAMESPACE_OPEN
 
 namespace Mat
 {
+  /*!
+   * @brief Mechanical heat source contribution produced by the material and the linearizations
+   * needed by thermomechanical coupling. Default constructed with zero values.
+   */
+  struct MechanicalDissipation
```
dragos-ana: If I understand this correctly, then this is the Taylor-Quinney heating along with its derivatives right?

I think that the term MechanicalDissipation can be slightly confusing here, because the Taylor-Quinney heating is 
$R_{TQ} = D_{\text{mech}} - T \frac{\partial D_{\text{mech}}}{\partial T} $, where $D_{\text{mech}}$ denotes the mechanical dissipation.
AFAIK this is also very confusing in the current TSI framework.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3698,31 +4188,189 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
     temp3x3(std::get<1>(indices_array[i]), std::get<0>(indices_array[i])) += pert_fact / 2.0;
     delta_perturbed_defgrads[i].multiply_tn(1.0, iFredM, temp3x3, 0.0);
 
-    // get perturbed defgrad
-    perturbed_FM.update(1.0, FredM, 1.0, delta_perturbed_defgrads[i], 0.0);
+    for (const double delta_sign : {1.0, -1.0})
+    {
```
dragos-ana, line 4206: Does this really solve any problems based on your experience with the perturbation-based approach? Because it certainly adds even more computation effort...

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3698,31 +4188,189 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_additional_cmat
...
+
+    // run return-mapping with the perturbed temperature
+    const HistoryVariables perturbed_history_variables =
+        return_mapping(FredM, perturbed_temperature);
+
+    Core::LinAlg::Matrix<9, 1> perturbed_iFinV(Core::LinAlg::Initialization::zero);
+    Core::LinAlg::Voigt::matrix_3x3_to_9x1(
+        perturbed_history_variables.inv_plastic_defgrad, perturbed_iFinV);
+
+    diFindT_FD.update(delta_sign / (2.0 * delta_T_perturbation), perturbed_iFinV, 1.0);
```
dragos-ana: Maybe I'm missing something, but isn't the perturbation of iFin required instead of the perturbed IFin?

ischeider:
Nice work! There are other plasticity models (`MAT_struct_ThrPlasticHyperElast, MAT_Plastic_ElastHyper, MAT_Struct_ThrPlasticLineElast`) which are used for TSI analyses, and MAT_PlasticElastHyper even has a taylor quinney heat generation included. Thus, I wonder, whether some components of your code use the same code as the old one does (I couldn't find any, but maybe it did not dig deep enough). Since heat generation does not depend so much on the material model, it may be used by a number of dissipative models.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
```diff
@@ -2334,5 +2463,604 @@ namespace
...
+    ASSERT_GT(std::abs(analytic_dissipation.derivative_wrt_temperature), 0.0);
+
+    ASSERT_DOUBLE_EQ(analytic_dissipation.value, fd_dissipation.value);
+    expect_near_relative(analytic_dissipation.derivative_wrt_cauchy_green,
+        fd_dissipation.derivative_wrt_cauchy_green, 1.0e-7, 1.0e-16);
+    expect_near_relative(analytic_dissipation.derivative_wrt_temperature,
+        fd_dissipation.derivative_wrt_temperature, 1.0e-8, 1.0e-16);
+  }
+
+  TEST_F(InelasticDefgradFactorsTest, ThermoViscoplastPublicEvaluationsCanBeCalledInAnyOrder)
```
dragos-ana, line 2797: A small documentation of this test would be great. 
It's very long, and it's definitely understandable once you get the gist of it. 
But it would be nice to see here in 2-3 bullet points what one can expect.

dragos-ana (commented):
Really great work!

I left some suggestions, and there are also some unclarities in my understanding of some parts of the code.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
rjoussen: agreed. Negative expansion would be physically strange, but otherwise, no problem.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
rjoussen: I'm redesigning the whole test because, thinking it over again, I realized it wasn't entirely testing what I wanted it to test. Lmk if its understanable now. Since I am using `pre_evaluate` inside `evaluate_mechanical_dissipation` now, they are truly independent an that is also what I am testing.

src/mat/4C_mat_inelastic_defgrad_factors_service.cpp
rjoussen: makes sense

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
rjoussen: It still approximates the conversion from plastic deformation to heat, which people normally think of as dissipation. I agree that it is definitely not the internal dissipation. I was thinking of a more general name like `ThermoSolidHeatSource`, this way, the material could decide what they mean by it. But probably conflicting with the fact that we also have the $\frac{1}{2}\cdot T \cdot\boldsymbol{S}_{;T} : \dot{\boldsymbol{C}}$ term. But technically, that would be the right abstraction for me: The thermo element should just know that it can ask solid materials for an additional heat source plus its linearizations, not what exactly that heat source is.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
rjoussen: I don't think that is needed here. For example, this struct is not concerned with whether it is timestep `n` or `n+1`. So I think only the usage sites would need to document that more accurately if it is unclear there.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: In `evaluate_inverse_inelastic_defgrad`, we currently do
```cpp
// check whether we have already evaluated the inverse inelastic deformation gradient for
  // the given reduced deformation gradient (this check should only be
  // performed for the first repetition, since we want to repeat the
  // return mapping under the same conditions, and not simply return the
  // computed value)
  if (defgrad_difference_norm(
          reduced_kinematics.defgrad, time_step_quantities_.current_defgrad[gp_]) <= 1.0e-16)
  {
    // just set the already computed value, no further computation
    iFinM = time_step_quantities_.current_plastic_defgrad_inverse[gp_];
    return;
  }
```
which, in my understanding, protects against double-evaluation. However, if the temperature changed, even the same defgrad would need to be evaluated again. I don't think this happens at the moment, because it would mean that the solid field only changes temperature but not displacement in between two evaluations, but it is still safer to invalidate the current defgrad here.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: I started using $S_T$ as the direct counterpart of $S_e$ in #2037 I think, so I would leave it like this now. I think it is very unlikely to mix it up with the thermal contribution to the 2PK stress pulled back by $\boldsymbol F_p$

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: Same argument as above.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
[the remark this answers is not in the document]
rjoussen: Yes that sounds very unfeasible for the substepping. I am asserting now.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: I agree, they kind of add a second check sometimes. I shifted the minus sign

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: It definitely does not, but with this, I was able to reduce the tolerances a little in the `ThermoViscoplastAdditionalCmatLinearizationMatchesFD`. Plus, it is the same now as in the off-diagonal stiffness matrix. I understand your concern, but don't we use the perturbation linearization only for debugging anyway? Because it is already way too expensive. If we ever switch to tensors, it could be worthwhile thinking about using auto-diff for that, I think.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: It used to be, because we started implementing this for temperature increments only, but now it can't. Theoretically, the perturbed temperature could become negative here, I adapted that.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: I am approximating the derivative by 
$
\frac{d\mathbf{F}_p^{-1}}{dT}\bigg|_{T}
\approx
\frac{\mathbf{F}_p^{-1}(T+\Delta T)-\mathbf{F}_p^{-1}(T-\Delta T)}{2\Delta T}
$
So I think it is right? I don't completely understand your comment tbh.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
rjoussen: I'm not sure here as well if this is the right place to document that. The function could do the exact same at any timestep.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: done. Also in `evaluate_history_variables_wrt_cauchy_green`

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
rjoussen: Done in [fecb48a](https://github.com/4C-multiphysics/4C/pull/2041/commits/fecb48ad39340f7b34bcece59ff8cae30fcbff82)

rjoussen:
> Nice work! There are other plasticity models (`MAT_struct_ThrPlasticHyperElast, MAT_Plastic_ElastHyper, MAT_Struct_ThrPlasticLineElast`) which are used for TSI analyses, and MAT_PlasticElastHyper even has a taylor quinney heat generation included. Thus, I wonder, whether some components of your code use the same code as the old one does (I couldn't find any, but maybe it did not dig deep enough). Since heat generation does not depend so much on the material model, it may be used by a number of dissipative models.

@ischeider: Currently, this material is self-contained and does not reuse or expose any existing heat generation code. In the next step, the multiplicative split parent material will expose the Taylor-Quinney heating introduced here to the thermo element. 

That said, the layout of heat generation across different materials could indeed use a rework, in my opinion. Currently, pretty much every material is treated completely differently in the thermo element, which is super confusing and introduces a lot of code that doesn't belong there. As I said in one of the review comments ([here](https://github.com/4C-multiphysics/4C/pull/2041#discussion_r3318616304)), the thermo element should not know the reason for a heat source. That especially means, though, that the decision about the type of heat source also needs to remain in the solid material, and from there it's hard to unify. But I am open to ideas :)

rjoussen:
In the latest commits, I refactored the call order unittests, addressed some of the comments, and fixed an issue with substepping: The temperature needs to be interpolated as well.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: True that it does not matter which time instance it is. But the derivative, as you have written it, does not really exist in an analytical sense. It only exists in a discretized sense specifically for timesteps n -> n+1 with initial values for the internal variables at time n. That's why it would be clearer in my opinion to keep this logic also within the code.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: Same as above

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Ok, nice catch! Can you maybe also add a comment there with this nice reasoning , to make it more easily accessible?

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
rjoussen: Okay, I am convinced. I will change all occurrences.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
rjoussen: done

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: Done

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper_service.hpp
rjoussen: In https://github.com/4C-multiphysics/4C/pull/2041/commits/bf27691ce838755afbfbb904073fe4c11b9e07f3, I renamed this struct to `HeatSource` because this is essentially what it is and what the thermo element needs in the next step. All functions in the viscoplastic factor are renamed to something with `taylor_quinney`, because they specialize the kind of heat source.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: That's true, I don't use the perturbation approach at all anymore ;) you can leave it like this if this is better for your computations, just wanted to understand whether this has any physical reasoning.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: True, now I get it, this is already solved with the changing delta_sign. Thanks for the clarification.

unittests/mat/4C_inelastic_defgrad_factors_test.cpp
dragos-ana: Looks very nice with the changed logic and the added documentation.

src/mat/4C_mat_inelastic_defgrad_factors.hpp
```diff
@@ -1500,17 +1512,40 @@ namespace Mat
...
+     * @param[in] iFin_other Already computed inverse inelastic deformation gradient
+     *              (from already computed inelastic factors in the multiplicative split material)
+     * @param[in] current_temperature Absolute current temperature
+     * @return mechanical dissipation heat source and derivatives w.r.t. temperature and the right
+     *         Cauchy-Green tensor
+     */
+    [[nodiscard]] HeatSource evaluate_taylor_quinney_heat_source(
+        const EvaluationContext<3>& context, const int gp, const int eleGID,
+        const Core::LinAlg::Matrix<3, 3>* defgrad, const Core::LinAlg::Matrix<3, 3>& iFin_other,
+        const double& temperature);
```
c-p-schmidt:
```suggestion
        const double temperature);
```

src/mat/4C_mat_inelastic_defgrad_factors.hpp
```diff
@@ -1500,17 +1512,40 @@ namespace Mat
...
+     * this inelastic factor.
+     *
+     * @param[in] context Evaluation context, providing access to the current timestep and total
+     * time
+     * @param[in] gp Gauss point
+     * @param[in] eleGID global element ID
+     * @param[in] defgrad Deformation gradient
+     * @param[in] iFin_other Already computed inverse inelastic deformation gradient
+     *              (from already computed inelastic factors in the multiplicative split material)
+     * @param[in] current_temperature Absolute current temperature
```
c-p-schmidt:
```suggestion
     * @param[in] temperature Absolute current temperature
```

c-p-schmidt (dismissed):

c-p-schmidt (approved):

dragos-ana (approved):

PR #2042, opened 2026-05-22
Remove custom implemented functions from elchsinglemat

c-p-schmidt:
## Description and Context
This PR is the first step to remove custom implemented functions from the `elchsinglemat`. Different functions can be chosen by „magic“ negative numbers in the function definition with a combination of paramters which is a nightmare for usage. Thus, the commits of this PR are structured as:
1. some cleanup (improve docu, add nodiscard, ..)
2. remove custom implemented functions that were unsused
3. replace the „inverse linear“ function by usage of the function framework


src/mat/4C_mat_elchsinglemat.hpp
```diff
@@ -85,162 +85,130 @@ namespace Mat
...
-    static constexpr int NYMAN = -8;
-    static constexpr int DEBYE_HUECKEL = -9;
-    static constexpr int KOHLRAUSCH_SQUAREROOT = -10;
     static constexpr int GOLDIN = -11;
-    static constexpr int STEWART_NEWMAN = -12;
-    static constexpr int TDF = -13;
     static constexpr int ARRHENIUS = -14;
-    static constexpr int INVERSE_LINEAR = -15;
-    //@}
+    ///@}
```
dragos-ana, line 135: I think that it would be much nicer if these were contained within an enum class, with enums directly referenced within the input files.
Although this would be associated with some input file refactoring, it would definitely be much clearer.

If this path is unfeasible, then at least we could make the indices here consecutive.

c-p-schmidt: I should have put this in the description. This PR is just the first step. When I'm finished, all these values are gone. Thus, I also would not change them to enum classes or similar.
First, because I hope to have them removed soon, and second, because this clutters the commits with unnecessary information. Now, when I only change function by function in a commit, you can directly see what you have to change in the input files. When I combine this with renaming / or reindexing (changing the assigned numbers), I also have quite some changes only due to this, which makes it less transparent in my opinion.

rjoussen: I agree, but since (as I understand it) different laws require different numbers of function parameters, it could make sense to have one input group per law, with the parameters as entries in that group. Not a concern of this PR though

dragos-ana: Ok, that's fine for me if the end goal is to remove them altogether. I was slightly confused with the PR because I also thought based on our discussion yesterday that everything should be removed.

rjoussen (approved):
Changes look good to me

dragos-ana (approved):

PR #2044, opened 2026-05-23
Enable monitor dbc in monolithic solvers

rjoussen:
## Description and Context

This PR aims to enable DBC reaction forces to be output in monolithic solvers for multifield problems (as in pure solid problems via `TAG: monitor_reaction` and `IO/MONITOR STRUCTURE DBC`).

#### Remarks

1. If there are multiple dofsets in the discretization, `.dof()` access expects the that the number of the dofset is specified, otherwise it throws (at least in debug mode). Since there is no easy way to communicate the displacement dofset ID to the dbc writer, it remains hardcoded to 0, which I think is a reasonable choice for now. 
2. Specifically tsi creates the dbc writer twice for some reason (it calls the structure setup twice), which was a problem since the dbc writer created reaction conditions on every init call and directly added them to the discretization. If this is done twice, there ends up being too many reaction conditions. This is circumvented by building the reaction conditions from the tagged conditions and then **replacing** all existing reaction conditions.

I added tests for ssi and tsi, though I think this should work now for other problem types as well.


maxfirmbach:
@rjoussen Is the default node set not always zero? Might be worth checking ... or I am missing something important 😄.

Edit: I understand now ... took a bit 😅.

maxfirmbach (dismissed):

maxfirmbach (approved):

c-p-schmidt (approved):
Having to assume the dofset is 0 is not ideal, but since we are in the `structure_new_dbc_monitor` and referring to the structure discretization, this should always be justified.

rjoussen:
> Having to assume the dofset is 0 is not ideal, but since we are in the `structure_new_dbc_monitor` and referring to the structure discretization, this should always be justified.

That's what I was thinking. As far as I could see, the structure discretization never explicitly declares displacements as dofset 0, so this would have needed to change as well.

PR #2058, opened 2026-05-31
[Particle] Fix probe for the message with the size tag

vovannikov:
## Description and Context
I observed a very strange bug on a system with 2 EPYC 7713 processors. When running on 32 ranks, the SPH simulation worked fine, but it crashes for larger numbers with the following errors:
```
[a0sv2300080:231845] *** An error occurred in MPI_Waitall
[a0sv2300080:231845] *** reported by process [3095068673,11]
[a0sv2300080:231845] *** on communicator MPI COMMUNICATOR 3 SPLIT FROM 0
[a0sv2300080:231845] *** MPI_ERR_TRUNCATE: message truncated
[a0sv2300080:231845] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[a0sv2300080:231845] ***    and potentially your MPI job)
```

It turned out the bug was related to `MPI_Probe` race condition.

The size-reception loop used `MPI_ANY_TAG` when probing for size messages (tag 1234). With 64 ranks spanning 2 NUMA domains, execution speeds diverge significantly. A fast rank can complete its size reception, post `MPI_Irecv`s, and begin sending data (tag 5678) back to a slow rank — while that slow rank is still probing for size messages. MPI does not guarantee inter-tag ordering, so the data message can be matched by the probe before all size messages arrive. This corrupts the communication protocol and causes cascading `MPI_ERR_TRUNCATE` errors on other ranks during `MPI_Waitall`.

With 32 ranks on a single socket, execution speeds are more uniform, making this race too unlikely to trigger.

Additionally, there was a UB with `msgsizetosend` variable that goes out of scope for a non-blocking `MPI_Isend`.

@georghammerl, @slfuchs, @ppraegla

### DIFF
--- a/src/particle/src/engine/4C_particle_engine_communication_utils.cpp
+++ b/src/particle/src/engine/4C_particle_engine_communication_utils.cpp
@@ -41,6 +41,7 @@ void Particle::ParticleUtils::immediate_recv_blocking_send(
   // ---- send size of messages to receiving processors ----
   std::vector<MPI_Request> sizerequest(numsendtoprocs);
   std::vector<int> sizetargets(numsendtoprocs);
+  std::vector<int> msgsizestosend(numsendtoprocs);
   int counter = 0;
   for (const auto& p : sdata)
   {
@@ -50,14 +51,15 @@ void Particle::ParticleUtils::immediate_recv_blocking_send(
 
     sizetargets[counter] = torank;
 
-    int msgsizetosend = static_cast<int>((p.second).size());
+    msgsizestosend[counter] = static_cast<int>((p.second).size());
 
     // check sending size of message
-    if (not(msgsizetosend > 0))
-      FOUR_C_THROW("sending non-positive message size {} to proc {}!", msgsizetosend, torank);
+    if (not(msgsizestosend[counter] > 0))
+      FOUR_C_THROW(
+          "sending non-positive message size {} to proc {}!", msgsizestosend[counter], torank);
 
     // perform non-blocking send operation
-    MPI_Isend(&msgsizetosend, 1, MPI_INT, torank, 1234, comm, &sizerequest[counter]);
+    MPI_Isend(&msgsizestosend[counter], 1, MPI_INT, torank, 1234, comm, &sizerequest[counter]);
 
     ++counter;
   }
@@ -67,17 +69,12 @@ void Particle::ParticleUtils::immediate_recv_blocking_send(
   std::vector<MPI_Request> recvrequest(numrecvfromprocs);
   for (int rec = 0; rec < numrecvfromprocs; ++rec)
   {
-    // probe for any message to come
+    // probe for message with size tag to come
     MPI_Status status;
-    MPI_Probe(MPI_ANY_SOURCE, MPI_ANY_TAG, comm, &status);
+    MPI_Probe(MPI_ANY_SOURCE, 1234, comm, &status);
 
-    // get message sender and tag
+    // get message sender
     int const msgsource = status.MPI_SOURCE;
-    int const msgtag = status.MPI_TAG;
-
-    // check message tag
-    if (msgtag != 1234)
-      FOUR_C_THROW("received data on proc {} with wrong tag from proc {}", myrank, msgsource);
 
     // get message size
     int msgsize = -1;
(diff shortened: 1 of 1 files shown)


src/particle/src/engine/4C_particle_engine_communication_utils.cpp
```diff
@@ -67,17 +69,12 @@ void Particle::ParticleUtils::immediate_recv_blocking_send(
...
+    MPI_Probe(MPI_ANY_SOURCE, 1234, comm, &status);
 
-    // get message sender and tag
+    // get message sender
     int const msgsource = status.MPI_SOURCE;
-    int const msgtag = status.MPI_TAG;
-
-    // check message tag
-    if (msgtag != 1234)
-      FOUR_C_THROW("received data on proc {} with wrong tag from proc {}", myrank, msgsource);
```
ppraegla, line 80: Do you know why the if didn't catch the wrong tag and threw an error?

ppraegla (approved):
Great work @vovannikov ! Thanks for fixing this bug.

georghammerl (approved):

src/particle/src/engine/4C_particle_engine_communication_utils.cpp
vovannikov: That is a very good question. Most likely, this exception did fire on a rank that has encountered the mismatch. This rank then starts stack unwinding due to the exception thrown, which is not fast. And this happens in the middle of the rather complex MPI communication pattern that gets broken. I guess, that another rank then detects the breakage of the communication and reports `MPI_ERR_TRUNCATE` and since this error is emitted not by the C++ exception mechanism but rather MPI, it just functions much faster and leads to aborting the application without any chances for the rank that actually detected the original mismatch to ever complete the proper exception handling.

This is my understanding of the issue. I did not make detailed tests actually, apart from checking the fix on the simulations with which I actually encountered this problem.

vovannikov: After further thinking, I decided to query Opus 4.6 to analyze the crash mechanics and it has a similar opinion:

Here's the sequence with 64 ranks:

1. Rank X (slow, still in the Probe loop) matches a tag-`5678` data message from a fast rank that already progressed to its data-sending phase.
2. The `msgtag != 1234` check triggers and `FOUR_C_THROW` throws a C++ exception (not `MPI_Abort`).
3. Stack unwinding on rank X destroys the local `std::vector<MPI_Request>` objects (`sizerequest`, `recvrequest`, `sendrequest`) - but MPI doesn't know about C++ destructors. The active requests become orphaned: MPI still holds references to send/receive buffers that are now being freed.
4. Rank X's outstanding `MPI_Isend` operations (sizes it promised to send, data it promised to send) never complete properly. Rank X's posted `MPI_Irecv` buffers are freed while MPI is still trying to write into them.
5. Rank Y, which is further along (already in `MPI_Waitall` for data receives), encounters corrupted MPI state on the shared communicator. The orphaned operations on rank X have disrupted message matching — rank Y receives a message into a buffer that doesn't match (either because the matching got confused, or because rank X's partial cleanup sent garbage). This manifests as `MPI_ERR_TRUNCATE`.
6. With `MPI_ERRORS_ARE_FATAL`, rank Y prints the truncation error and aborts. Rank X's `FOUR_C_THROW` output (which went through `Core::IO::cout`, not `stderr`) is lost in the chaotic multi-rank abort - **only rank Y's MPI error handler output survives because it writes directly to `stderr`**.

**Why 32 ranks worked**: With fewer ranks on a single socket, execution speeds are more uniform. All ranks progress through the Probe loop at similar rates, so no rank reaches the data-sending phase (tag 5678) before others finish receiving sizes (tag `1234`). The race window is effectively zero.

**Why 64 ranks fails**: With ranks spanning 2 NUMA domains (2× AMD EPYC 7713), cross-socket communication latency creates significant speed divergence. Fast ranks on one socket complete their entire size-exchange and begin sending data while slow ranks on the other socket are still probing — opening the race window for the MPI_ANY_TAG probe to match wrong-tag messages.

_PS from me_: I marked in bold the place where I could have been not fully correct. We might have seen the original error if it had been written to `stderr` and in general had had a simpler error handling mechanics. `FOUR_C_THROW` is actually quite complicated.

ppraegla: Thanks for the detailed explanation. Definitely makes sense. But either way, good that you fixed it 🙌

PR #2060, opened 2026-06-01
Introduce output verbosity levels in reduced lung

maxiludwig:
## Description and Context

This PR adds configurable runtime output verbosity for the `reduced_lung` module with three modes: `minimal`, `medium`, and `high`.
Core solution fields are always written. If a quantity is not defined for a given element/model pair (for example `q_out` in terminal units), we now write `NaN` explicitly so missing values are clear.
Additional model-specific outputs are handled directly in model-level output evaluators. This keeps minimal output compact, while detailed output can include internal model quantities when needed for analysis/debugging.

### DIFF
--- a/src/reduced_lung/src/4C_reduced_lung_helpers.cpp
+++ b/src/reduced_lung/src/4C_reduced_lung_helpers.cpp
@@ -661,12 +661,19 @@ namespace ReducedLung
       Core::IO::DiscretizationVisualizationWriterMesh& visualization_writer,
       const AirwayContainer& airways, const TerminalUnitContainer& terminal_units,
       const Core::LinAlg::Vector<double>& locally_relevant_dofs,
-      const Core::LinAlg::Map* element_row_map)
+      const Core::LinAlg::Map* element_row_map, ReducedLungParameters::OutputVerbosity verbosity)
   {
     Core::LinAlg::Vector<double> pressure_in(*element_row_map, true);
     Core::LinAlg::Vector<double> pressure_out(*element_row_map, true);
     Core::LinAlg::Vector<double> flow_in(*element_row_map, true);
     Core::LinAlg::Vector<double> flow_out(*element_row_map, true);
+
+    const double nan = std::numeric_limits<double>::quiet_NaN();
+    pressure_in.put_scalar(nan);
+    pressure_out.put_scalar(nan);
+    flow_in.put_scalar(nan);
+    flow_out.put_scalar(nan);
+
     for (const auto& model : airways.models)
     {
       const bool has_q_out = model.data.n_state_equations == 2;
@@ -693,8 +700,6 @@ namespace ReducedLung
             locally_relevant_dofs.local_values_as_span()[model.data.lid_p2[i]]);
         flow_in.replace_local_value(model.data.local_element_id[i],
             locally_relevant_dofs.local_values_as_span()[model.data.lid_q[i]]);
-        flow_out.replace_local_value(model.data.local_element_id[i],
-            locally_relevant_dofs.local_values_as_span()[model.data.lid_q[i]]);
       }
     }
     visualization_writer.append_result_data_vector_with_context(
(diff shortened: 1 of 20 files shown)


src/reduced_lung/src/4C_reduced_lung_helpers.cpp
```diff
@@ -705,6 +710,33 @@ namespace ReducedLung
         flow_in, Core::IO::OutputEntity::element, {"q_in"});
     visualization_writer.append_result_data_vector_with_context(
         flow_out, Core::IO::OutputEntity::element, {"q_out"});
+
+    if (verbosity > ReducedLungParameters::OutputVerbosity::minimal)
+    {
+      RuntimeOutputCollector collector(*element_row_map);
+
+      for (const auto& model : airways.models)
```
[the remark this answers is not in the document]
maxiludwig: I would propose to extend the integration testing of `reduced_lung` in a future PR and and add testing of the output verbosity levels there.

src/reduced_lung/src/4C_reduced_lung_helpers.cpp
```diff
@@ -693,8 +700,6 @@ namespace ReducedLung
             locally_relevant_dofs.local_values_as_span()[model.data.lid_p2[i]]);
         flow_in.replace_local_value(model.data.local_element_id[i],
             locally_relevant_dofs.local_values_as_span()[model.data.lid_q[i]]);
-        flow_out.replace_local_value(model.data.local_element_id[i],
```
maxiludwig, line 696: Please note that the `q_out` of terminal units is now `NaN`

src/reduced_lung/src/4C_reduced_lung_input.hpp
```diff
@@ -22,6 +23,13 @@ namespace ReducedLung
 {
   struct ReducedLungParameters
   {
+    enum class OutputVerbosity : std::uint8_t
+    {
+      minimal,
+      standard,
+      detailed
```
maxiludwig: @bgoderbauer what do you think about the names?

src/reduced_lung/src/4C_reduced_lung_input.cpp
```diff
@@ -402,6 +402,12 @@ Core::IO::InputSpec ReducedLung::valid_parameters()
                           .store = in_struct(
                               &ReducedLungParameters::Dynamics::nonlinear_increment_tolerance),
                       }),
+                  parameter<ReducedLungParameters::OutputVerbosity>("output_verbosity",
+                      {
+                          .description = "Output verbosity level.",
+                          .default_value = ReducedLungParameters::OutputVerbosity::minimal,
```
maxiludwig, line 408: Maybe its confusing to have `minimal` as the default, because `standard` sounds like a classical word describing a default. 
We could rename `standard` to `medium`.

src/reduced_lung/src/4C_reduced_lung_input.hpp
davidrudlstorfer: Isn't it possible to reuse the already existing verbosity levels of 4C instead of creating even more distributed verbosity levels?

maxiludwig: I am not sure if this is a good idea. Isn't the Core::IO::Verbositylevel mainly used for command line output verbosity? 
For reduced_lung we would like to change the verbosity of the vtk output

bgoderbauer: I think the implementation is fine, but I agree with @davidrudlstorfer that, if possible, we should use existing naming when available, so we don't end up with different modules with different specifications for the same use cases in the long run. @c-p-schmidt, do you know whether there exist different VTK output levels somewhere in the code base? Then we can just switch based on them without changing your current implementation details

src/reduced_lung/src/4C_reduced_lung_input.cpp
bgoderbauer: Let's see if we can fall back to common names, otherwise I would opt for medium as well

src/reduced_lung/src/4C_reduced_lung_input.hpp
bgoderbauer: I just talked to @c-p-schmidt and it likely makes most sense to keep this local in our module. Having different vtk-verbosity levels is quite unique to our use case and not really relevant to the other 4C modules.

bgoderbauer:
Thanks for implementing this! I'm looking forward to not needing 10 different filters in paraview again and again :smile:. Let's just rename 'standard' to 'medium,' I would say.

src/reduced_lung/src/4C_reduced_lung_input.hpp
c-p-schmidt: @bgoderbauer: As just spoken in person, something like this is not available. And from the fields, I use such a level would be rather unintuitive to use. In these fields, you can choose that, e.g., the velocity or strain output is on or off, but this is then always explicitly stated and not a "category".

However, in your case, this sounds like the way to go with a reasonable description of what the levels mean.

bgoderbauer (approved):

davidrudlstorfer (approved):

PR #2064, opened 2026-06-02
Remove custom implemented functions: Part 3

c-p-schmidt:
## Description and Context
This PR is the third step toward removing custom-implemented functions from elchsinglemat. Different functions can be chosen by „magic“ negative numbers in the function definition, along with a combination of parameters, which is a nightmare to use. The commits of this PR are structured as:

1. Replaces the Arrhenius function with an implementation in scatra functions that is now available for all scatra problems with a nicer user interface
2. Some minor cleanup of the scl material as a byproduct, as I had to touch it within the first commit anyway

Now, only the most useless but most used function (the constant function) is left, which will be removed in the next PR with some restructuring.

Follows #2042 #2061


lauraengelhardt (approved):
Nice, looks good to me!

bgoderbauer (approved):

PR #2065, opened 2026-06-02
Modify vtu vector storage for 2D node output

ischeider:
# Description

For 2D solid analyses, the  displacements, velocities, etc., was stored as 3-component vectors, and the third component was saved as "NaN". The reason was given in a comment:

```
        // We have to make sure that this dof is actually available at this node. It can happen that
        // this is not the case e.g. in multi-scale simulations where different nodes in the mesh
        // have a different number of dofs.
        // Since this is not supported by vtu, we add a NaN in this case, see else branch

```

Since a Warp filter cannot be applied in Paraview, if the third component is NaN, I changed it to 0.0. I hope there are no other side effects.

## Related issues, PR, etc.
#2027

## Interested parties:
@bennoschoenstein

### DIFF
--- a/src/core/io/src/4C_io_element_append_visualization.hpp
+++ b/src/core/io/src/4C_io_element_append_visualization.hpp
@@ -209,7 +209,7 @@ namespace Core::IO
         // We have to make sure that this dof is actually available at this node. It can happen that
         // this is not the case e.g. in multi-scale simulations where different nodes in the mesh
         // have a different number of dofs.
-        // Since this is not supported by vtu, we add a NaN in this case, see else branch
+        // Keep the requested vector size and pad missing components with 0.0, see else branch
         if (nodedofs.size() > read_result_data_from_dofindex + idof)
         {
           const int lid =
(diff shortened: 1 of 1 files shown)


c-p-schmidt:
Can you please provide a bit more information? I just tried it with the test `solid_runtime_vtk_2d.4C.yaml`.
With the current main and with the proposed changes, filters like WarpByVector can not be applied.
I guess that the root cause is that e.g. the displacement vector has only 2 dofs per node. Thus in both cases the entry in the vtu file looks like this for me.

```bash
  <PointData>
        <DataArray type="Float64" Name="displacement" NumberOfComponents="2" format="ascii">
2.058292735896430e-04 2.458540727946142e-04
... some more lines
```

ischeider:
Sorry, obviously my branch was screwed up. There was a time when I always got three components, but one of them was nan. However, I cannot reproduce this output, now I also get two components. So the PR is to be closed.
Anyway, since warp by vector cannot be applied if the vector has only two components I recommend to add a third component to the displacements if possible. I'll check.

PR #2067, opened 2026-06-02
Monolithic non-local stimulus

tuchpaul:
## Monolithic SSI for non-local stimulus growth and remodeling

Enables the monolithic coupling for the non-local stimulus (NLS) hCMM variant introduced in #2019. 

### What this PR adds

**Implicit local ODE integration** (`MixtureConstituentRemodelFiberSsi`, `RemodelFiberImplementation`)

- New `integrate_local_evolution_equations_implicit_with_nonlocal_stimulus(psi, dt)`: integrates growth_scalar $\kappa$ in closed form (Crank-Nicolson exact because the ODE is linear in $\kappa$) and $\lambda_r$ via a scalar Newton. Stores the sensitivities $\partial\kappa_g^{n+1}/\partial\psi$ and $\partial\lambda_r^{n+1}/\partial\lambda_f^2$ needed for the off-diagonal blocks.
- New `IMPLICIT` input parameter on `MIX_Constituent_SsiRemodelFiber` (default `false`). Implicit mode is required for a consistent monolithic tangent; explicit mode remains available for staggered solvers.
- Coressponding integration call sites: implicit integration is called inside `evaluate`/`evaluate_elastic_part` (every Newton iteration), explicit only inside `update`/`update_elastic_part` (after convergence).
- `FOUR_C_ASSERT_ALWAYS` in `GrowthRemodelMixtureRule::evaluate_d_stress_d_scalars` enforces implicit mode when the monolithic $K_{u\psi}$ path is entered.

**Off-diagonal block** $K_{\psi u}$ (`ScaTraEleCalcNonlocalStimulus`)

New `sysmat_od_mesh_nls` assembles the ScaTra–displacement block from the source-term linearization:

$$-N_{v_i} \cdot \frac{d\sigma}{d\lambda_f^2} \cdot 2\mathbf{b}_\alpha \nabla_x N_{u_j}$$

The Eulerian structural tensor is:

$$\mathbf{b}_\alpha = \mathbf{F}\mathbf{A}_\alpha\mathbf{F}^T$$

which needed the construction of the deformation gradient $F$ in `ScaTraEleCalcNonlocalStimulus`.

Here $d\sigma/d\lambda_f^2$ includes the implicit $\lambda_r$ sensitivity via the stored `d_lambda_r_d_lambda_f_sq_`.

**Off-diagonal block $K_{u\psi}$** (`SolidScatraEleCalc`, `GrowthRemodelMixtureRule`)

- `evaluate_d_stress_d_scalars` extended to the multi-scalar case (previously single-scalar only) 
- Located in `Mat::Mixture` to `GrowthRemodelMixtureRule`, as the evaluate call for $K_{uu}$
- `MonolithicSolidScalarMaterial::evaluate_d_stress_d_scalars` added as the multi-scalar virtual interface; the default wraps the existing single-scalar `evaluate_d_stress_d_scalar` for backward compatibility.

**Monolithic solver**

- Removed the single-scalar restriction in `SsiMono::setup` that prevented NLS (which uses one ScaTra DOF per fiber).

**Test**

- New test `ssi_homogenized_constraint_mixture_qcyl_nonlocal_mono`: quarter-cylinder hCMM with NLS, monolithic SSI.

### Related
- #2019 -- non-local stimulus staggered variant (base implementation)

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
@@ -4368,6 +4368,11 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 {.description = "Index of the non-local stimulus scalar in the scatra matlist "
                                 "(leave unset to disable)",
                     .validator = null_or(positive_or_zero<int>())}),
+            parameter<bool>("IMPLICIT_INTEGRATION",
+                {.description =
+                        "Integrate growth_scalar and lambda_r implicitly at each Newton iteration. "
+                        "Default: false.",
+                    .default_value = false}),
         },
         {.description =
                 "A 1D constituent where the g&r evolution ODEs can be solved either as own scatra "
--- a/src/mat/4C_mat_monolithic_solid_scalar_material.hpp
+++ b/src/mat/4C_mat_monolithic_solid_scalar_material.hpp
@@ -13,7 +13,10 @@
 #include "4C_linalg_fixedsizematrix.hpp"
 #include "4C_linalg_symmetric_tensor.hpp"
 #include "4C_mat_so3_material.hpp"
+#include "4C_utils_exceptions.hpp"
 #include "4C_utils_parameter_list.fwd.hpp"
+
+#include <vector>
 
 FOUR_C_NAMESPACE_OPEN
 
@@ -39,6 +42,27 @@ namespace Mat
         const Core::LinAlg::SymmetricTensor<double, 3, 3>& glstrain,
         const Teuchos::ParameterList& params, const EvaluationContext<3>& context, int gp,
         int eleGID) = 0;
+
+    /*!
+     * @brief Evaluates dS/dc_k for all scalar DOFs per node.
+     *
+     * Returns a vector of num_scalars tensors. Entry k holds dS/dc_k.
+     * The default wraps evaluate_d_stress_d_scalar and places the result in entry 0;
+     * all other entries are zero (scalars not affecting the stress).
+     * Override for materials where the stress is affected by more than one scalar.
+     */
+    [[nodiscard]] virtual std::vector<Core::LinAlg::SymmetricTensor<double, 3, 3>>
+    evaluate_d_stress_d_scalars(const Core::LinAlg::Tensor<double, 3, 3>& defgrad,
+        const Core::LinAlg::SymmetricTensor<double, 3, 3>& glstrain,
+        const Teuchos::ParameterList& params, const EvaluationContext<3>& context, int num_scalars,
+        int gp, int eleGID)
+    {
+      FOUR_C_ASSERT(num_scalars > 0, "num_scalars must be positive");
+      std::vector<Core::LinAlg::SymmetricTensor<double, 3, 3>> result(num_scalars);
+      for (auto& t : result) t.fill(0.0);
+      result[0] = evaluate_d_stress_d_scalar(defgrad, glstrain, params, context, gp, eleGID);
+      return result;
+    }
   };
 }  // namespace Mat
 
--- a/src/mixture/src/4C_mixture_remodelfiber.cpp
+++ b/src/mixture/src/4C_mixture_remodelfiber.cpp
@@ -338,16 +338,16 @@ void Mixture::Implementation::RemodelFiberImplementation<numstates,
   const Core::LinAlg::Matrix<2, 1, T> d_residuum_d_lambda_f_sq = std::invoke(
       [&]()
       {
-        Core::LinAlg::Matrix<2, 1, T> d_residuum_d_lambfa_f_sq;
+        Core::LinAlg::Matrix<2, 1, T> d_residuum_d_lambda_f_sq;
 
-        d_residuum_d_lambfa_f_sq(0, 0) =
+        d_residuum_d_lambda_f_sq(0, 0) =
             evaluate_d_current_growth_evolution_implicit_time_integration_residuum_d_lambda_f_sq(
                 dt);
-        d_residuum_d_lambfa_f_sq(1, 0) =
+        d_residuum_d_lambda_f_sq(1, 0) =
             evaluate_d_current_remodel_evolution_implicit_time_integration_residuum_d_lambda_f_sq(
                 dt);
 
-        return d_residuum_d_lambfa_f_sq;
+        return d_residuum_d_lambda_f_sq;
       });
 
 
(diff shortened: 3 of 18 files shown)


src/mixture/src/4C_mixture_remodelfiber.cpp
```diff
@@ -384,6 +384,69 @@ void Mixture::Implementation::RemodelFiberImplementation<numstates,
...
+  const T lambda_f = states_.back().lambda_f;
+  const T lambda_ext = states_.back().lambda_ext;
+
+  // get linear ODE coefficient c
+  const T c = evaluate_growth_evolution_equation_dt_with_nonlocal_stimulus(
+      psi, lambda_f, states_[0].lambda_r, lambda_ext, T(1.0));
+
+  // growth_scalar^{n+1} = growth_scalar^n * (1 + dt*(1-theta)*c) / (1 - dt*theta*c)
+  constexpr double theta = ImplicitIntegration<numstates, T>::theta;
+  const T B = T(1.0) - dt * theta * c;
```
c-p-schmidt: What is B standing for?
```suggestion
  const T denominator = T(1.0) - dt * theta * c;
```

src/mixture/src/4C_mixture_rule.hpp
```diff
@@ -230,6 +230,26 @@ namespace Mixture
...
     }
 
+    /*!
+     * @brief Evaluates the derivative of the stress w.r.t. the scalar DOFs of the mixture
+     * constituents
+     *
+     * @param defgrad (in) : Deformation gradient
+     * @param glstrain (in) : Green-Lagrange strain
+     * @param params (in) : ParameterList for additional parameters
+     * @param gp (in) : Gauss points
```
c-p-schmidt: Docu does not fit to the implementation

src/scatra_ele/4C_scatra_ele_calc_nonlocal_stimulus.cpp
```diff
@@ -191,6 +193,103 @@ Discret::Elements::ScaTraEleCalcNonlocalStimulus<distype, probdim>::get_struct_m
...
+    Core::LinAlg::SerialDenseMatrix& elemat2, Core::LinAlg::SerialDenseVector& elevec1,
+    Core::LinAlg::SerialDenseVector& elevec2, Core::LinAlg::SerialDenseVector& elevec3)
+{
+  // base class call
+  const int result = my::evaluate_action_od(
+      ele, params, discretization, action, la, elemat1, elemat2, elevec1, elevec2, elevec3);
+
+  // nls specific contribution
+  if (action == ScaTra::Action::calc_scatra_mono_odblock_mesh)
+    sysmat_od_mesh_nls(elemat1, my::nsd_);
```
c-p-schmidt, line 210: nls is the abbreviation for nonlocal stimulus, or something else?

src/scatra_ele/4C_scatra_ele_calc_nonlocal_stimulus.cpp
```diff
@@ -191,6 +193,103 @@ Discret::Elements::ScaTraEleCalcNonlocalStimulus<distype, probdim>::get_struct_m
...
+        // Eulerian structural tensor: b_alpha = F * A_alpha * F^T
+        const Core::LinAlg::Matrix<my::nsd_, my::nsd_> A_alpha_mat = Core::LinAlg::make_matrix(
+            Core::LinAlg::get_full(constituent_[k]->get_structural_tensor(iquad)));
+        Core::LinAlg::Matrix<my::nsd_, my::nsd_> FA;
+        FA.multiply(defgrd, A_alpha_mat);
+        Core::LinAlg::Matrix<my::nsd_, my::nsd_> b_alpha;
+        b_alpha.multiply_nt(FA, defgrd);
+
+        // K_psi_u^{vi*numdof+k, ui*ndofmesh+d}
+        //   += -fac * d_sigma_d_lf2 * 2 * (b_alpha_{d,c} * derxy_(c,ui)) * funct_(vi)
```
c-p-schmidt: Would be nice if the comment and the implementation would "match", i.e., with the current implementation:

```bash
+= -fac * d_sigma_d_lf2 * funct_(vi) * 2 * (b_alpha_{d,c} * derxy_(c,ui)) 
```

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -4368,6 +4368,11 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 {.description = "Index of the non-local stimulus scalar in the scatra matlist "
                                 "(leave unset to disable)",
                     .validator = null_or(positive_or_zero<int>())}),
+            parameter<bool>("IMPLICIT",
```
maxfirmbach: Would `INTEGRATE_IMPLICIT` make sense here? Otherwise it feels just like a random word.

src/mat/4C_mat_monolithic_solid_scalar_material.hpp
```diff
@@ -39,6 +42,26 @@ namespace Mat
...
+    [[nodiscard]] virtual std::vector<Core::LinAlg::SymmetricTensor<double, 3, 3>>
+    evaluate_d_stress_d_scalars(const Core::LinAlg::Tensor<double, 3, 3>& defgrad,
+        const Core::LinAlg::SymmetricTensor<double, 3, 3>& glstrain,
+        const Teuchos::ParameterList& params, const EvaluationContext<3>& context, int num_scalars,
+        int gp, int eleGID)
+    {
+      FOUR_C_ASSERT(num_scalars > 0, "num_scalars must be positive");
+      std::vector<Core::LinAlg::SymmetricTensor<double, 3, 3>> result(num_scalars);
+      result[0] = evaluate_d_stress_d_scalar(defgrad, glstrain, params, context, gp, eleGID);
+      return result;
```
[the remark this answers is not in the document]
maxfirmbach: That is a really good comment from copilot 😃.

src/mixture/src/4C_mixture_remodelfiber.cpp
```diff
@@ -384,6 +384,69 @@ void Mixture::Implementation::RemodelFiberImplementation<numstates,
   d_lambda_r_d_lambda_f_sq_ = 0.0;
 }
 
+template <int numstates, typename T>
```
maxfirmbach, line 387: I have to say the `typename T` makes things really weird to read in the code ... but I do not have a strong opinion on this. I guess it is also done like this in other places ...

src/scatra_ele/4C_scatra_ele_calc_nonlocal_stimulus.cpp
```diff
@@ -191,6 +193,103 @@ Discret::Elements::ScaTraEleCalcNonlocalStimulus<distype, probdim>::get_struct_m
...
+    Core::FE::IntPointsAndWeights<Core::FE::dim<distype>> intpoints(
+        ScaTra::DisTypeToOptGaussRule<distype>::rule);
+
+    for (int iquad = 0; iquad < intpoints.ip().nquad; ++iquad)
+    {
+      const double fac = my::eval_shape_func_and_derivs_at_int_point(intpoints, iquad);
+      my::set_internal_variables_for_mat_and_rhs();
+
+      // Reference coordinates: xyze_ - edispnp_
+      Core::LinAlg::Matrix<my::nsd_, my::nen_> XYZe(my::xyze_);
```
maxfirmbach: Can we not call the reference coordinates what they actually are? `XYZe` is a really cryptic way of expressing that (I also know that this appears a lot in 4C).

src/solid_scatra_ele/4C_solid_scatra_ele_calc.cpp
```diff
@@ -521,30 +528,32 @@ void Discret::Elements::SolidScatraEleCalc<celltype, SolidFormulation>::evaluate
...
+              for (int k = 0; k < scatra_column_stride; ++k)
               {
-                const double BdSdc_rowi = BdSdc(rowi, 0);
-                // loop over columns
-                for (int coli = 0; coli < Core::FE::num_nodes(celltype); ++coli)
+                BdSdc.put_scalar(0.0);
+                Discret::Elements::add_internal_force_vector(
+                    jacobian_mapping, deformation_gradient, dSdc[k], integration_factor, BdSdc);
+
+                // loop over rows
```
maxfirmbach: Isn't it clear from the loop index `row` and `col` that we loop over rows and cols? So no need for these comments. Would `row` instead of `rowi` make more sense? Or `row_i` etc.

src/scatra_ele/4C_scatra_ele_calc_nonlocal_stimulus.cpp
```diff
@@ -191,6 +193,103 @@ Discret::Elements::ScaTraEleCalcNonlocalStimulus<distype, probdim>::get_struct_m
...
+            Core::LinAlg::get_full(constituent_[k]->get_structural_tensor(iquad)));
+        Core::LinAlg::Matrix<my::nsd_, my::nsd_> FA;
+        FA.multiply(defgrd, A_alpha_mat);
+        Core::LinAlg::Matrix<my::nsd_, my::nsd_> b_alpha;
+        b_alpha.multiply_nt(FA, defgrd);
+
+        // K_psi_u^{vi*numdof+k, ui*ndofmesh+d}
+        //   += -fac * d_sigma_d_lf2 * 2 * (b_alpha_{d,c} * derxy_(c,ui)) * funct_(vi)
+
+        // loop over the scatra nodes with usually one numdofpernode_
```
maxfirmbach, line 267: I think with better variable naming one could avoid the comments in general, but not extremely important for now.

maxfirmbach (commented):
@tuchpaul I left you some comments. Hopefully they are helpful ... Still, this PR looks really good 👍.

maxfirmbach:
This PR also has a very nice description, well done.

src/scatra_ele/4C_scatra_ele_calc_nonlocal_stimulus.cpp
tuchpaul: yes, I changed the comment to be more precise
`// nonlocal stimulus specific contribution to the off-diagonal block`

src/mixture/src/4C_mixture_remodelfiber.cpp
tuchpaul: The template `typename T' is used throughout the `RemodelFiberImplementation` and enables the support for automatic differentiation. But I agree, it makes the code harder to read..

src/scatra_ele/4C_scatra_ele_calc_nonlocal_stimulus.cpp
tuchpaul: Since there are quite some loops in the assembly block, I would keep the comments to better follow along.

src/mixture/src/4C_mixture_rule.hpp
```diff
@@ -230,6 +230,26 @@ namespace Mixture
...
+     * @return std::vector<Core::LinAlg::SymmetricTensor<double, 3, 3>>
+     */
+    [[nodiscard]] virtual std::vector<Core::LinAlg::SymmetricTensor<double, 3, 3>>
+    evaluate_d_stress_d_scalars(const Core::LinAlg::Tensor<double, 3, 3>& defgrad,
+        const Core::LinAlg::SymmetricTensor<double, 3, 3>& glstrain,
+        const Teuchos::ParameterList& params, const Mat::EvaluationContext<3>& context,
+        int num_scalars, int gp, int eleGID) const
+    {
+      return std::vector<Core::LinAlg::SymmetricTensor<double, 3, 3>>(num_scalars);
+    }
```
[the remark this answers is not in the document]
tuchpaul: `std::vector<T>(n)` value-initializes its elements. Since the constructor of `TensorInternal` is default (non-user-provided) the tensor should be zero-initialized. But I added the explicit `fill(0.0)` for clarity and consistency.

c-p-schmidt (approved):

maxfirmbach (approved):

PR #2072, opened 2026-06-03
Adapt the workflow schedules

c-p-schmidt:
## Description and Context
I realized recently that our nightly workflows finish rather late (around 10 - 11 a.m.), which interferes with PR pipelines when a PR is opened in the morning.
Thus, I suggest triggering the nightly workflows earlier and explicitly setting the timezone so one does not need to read the documentation to find the default time zone.
With these changes, the pipeline should be triggered around 4 hours earlier, i.e., the estimated starting time would be 9 p.m., and the estimated finish time would be 6-7 a.m.

See here why the workflows are usually triggered later than set in the definition (https://github.com/orgs/community/discussions/147369), thank you @davidrudlstorfer.

And here is the doc on how to set the timezone: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -2,7 +2,9 @@ name: coverage
 
 on:
   schedule:
-    - cron: '0 1 * * *'
+    # GitHub recommends avoiding scheduling at the start of the hour to avoid delays; thus, a random minute is set.
+    - cron: '29 0 * * *'
+      timezone: "Europe/Berlin"
 
 jobs:
   clang18_coverage_build:
--- a/.github/workflows/checkcode.yml
+++ b/.github/workflows/checkcode.yml
@@ -6,7 +6,9 @@ on:
     branches:
       - 'main'
   schedule:
-    - cron: '0 22 * * *'
+    # GitHub recommends avoiding scheduling at the start of the hour to avoid delays; thus, a random minute is set.
+    - cron: '44 20 * * *'
+      timezone: "Europe/Berlin"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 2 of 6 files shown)


.github/workflows/coverage.yml
```diff
@@ -2,7 +2,8 @@ name: coverage
 
 on:
   schedule:
-    - cron: '0 1 * * *'
+    - cron: '0 0 * * *'
```
davidrudlstorfer: Should we go even earlier with this one?

davidrudlstorfer (dismissed):
Thanks for the changes :)

.github/workflows/coverage.yml
c-p-schmidt: I don't think it would help, as it depends on completing the nightly_tests action. This is also why I only reduced the time by 1 hour (which, in combination with the updated timezone, is 3 hours).

rjoussen:
Looks good! While I was working on the fourcipp workflows, I read somewhere that GitHub recommends not starting workflows at the full hour, because they often get delayed then, apparently because everyone wants to start their workflows then. I can't remember where exactly I read that. Might be worth a try?

c-p-schmidt:
> Looks good! While I was working on the fourcipp workflows, I read somewhere that GitHub recommends not starting workflows at the full hour, because they often get delayed then, apparently because everyone wants to start their workflows then. I can't remember where exactly I read that. Might be worth a try?

It is actually directly mentioned in the GitHub doc that I linked in the description. Thank you for the hint. I'll generate a random number and put it as the minute specifier. 😉

c-p-schmidt:
I added random numbers as the minute specifier for the "relevant" workflows scheduled every night, following the recommendation in the GitHub doc.

davidrudlstorfer (approved):

rjoussen (approved):
Perfect!

PR #2063, opened 2026-06-01
[Particle] Reduce complexity of `Particle::PDNeighborPairs::communicate_bond_list()`

vovannikov:
## Description and Context
The original implementation used 3 nested loops with the resultant complexity of $O(P \times M \times B)$. Prebuilding the bond indices and storing them in an `unordered_map` significantly reduced the time spent in this function, see the table with the timing results I obtained when running the scalability studies of an SPH-PD simulation with ~5,000,000 particles. Observe the spike in time spent in `Particle::ParticleAlgorithm::distribute_load_among_procs` for 8 ranks which calls `Particle::PDNeighborPairs::communicate_bond_list()`. This is the time measured from the restart point that was computed running a simulation on 24 ranks. The table shows the maximum time over all ranks in seconds. The tests were performed on a system with 2 AMD EPYC 7713 processors, the same as used in #2058.

| ranks  | old version, s | new version, s |
| -- | -- | -- |
| 2 | 3.7 | 3.7 |
| 4 | 2.1 | 2.4 |
| 8 | 1827 | 3.3 |
| 16 | 486 | 1.6 |
| 32 | 463 | 1.2 |
| 64 | 234 | 1.1 |
| 128 | 109 | 3.5 |

For reference, here are the original job timings:
[2ranks-job-timings.yaml](https://github.com/user-attachments/files/28483745/2ranks-job-timings.yaml)
[4ranks-job-timings.yaml](https://github.com/user-attachments/files/28483754/4ranks-job-timings.yaml)
[8ranks-job-timings.yaml](https://github.com/user-attachments/files/28483757/8ranks-job-timings.yaml)
[16ranks-job-timings.yaml](https://github.com/user-attachments/files/28483752/16ranks-job-timings.yaml)
[32ranks-job-timings.yaml](https://github.com/user-attachments/files/28483753/32ranks-job-timings.yaml)
[64ranks-job-timings.yaml](https://github.com/user-attachments/files/28483756/64ranks-job-timings.yaml)
[128ranks-job-timings.yaml](https://github.com/user-attachments/files/28483751/128ranks-job-timings.yaml)

@georghammerl, @slfuchs, @ppraegla

PR #2074, opened 2026-06-04
[Particle] Cache senders and receivers of the refreshed particles

vovannikov:
When performing the scalability studies I have identified a few places which can be optimized in order to improve performance. Most of those changes are related to optimizing the MPI communications: mainly reducing their number and the amount of data to be transferred.

As the result of this study, multiple modifications are proposed and split into 6 PRs. This whole sequence of optimizations was performed using Opus 4.6 assisting in reasoning about the bottlenecks.

This PR (and the follow-up PRs too) is tested using the ~5,000,000 particles SPH-PD simulation previously studied in #2063. The tests were again performed on a system with 2 AMD EPYC 7713 processors.

### 1. Cache senders and receivers to reuse the ghosts topology when communicating particles

This PR eliminates the per-step target-discovery collective inside the refresh path by reusing the directed-ghosting topology already known at ghost-rebuild time.

It also replaces an expensive `Allreduce` communication with small fixed point-to-point exchanges over a peer set.

The caches are updated every time `communicate_direct_ghosting_map()` runs.

This is the effect of this modification:

image

### DIFF
--- a/src/particle/src/engine/4C_particle_engine.cpp
+++ b/src/particle/src/engine/4C_particle_engine.cpp
@@ -362,8 +362,8 @@ void Particle::ParticleEngine::refresh_particles() const
   // determine particles that need to be refreshed
   determine_particles_to_be_refreshed(particlestosend);
 
-  // communicate particles
-  communicate_particles(particlestosend, particlestoinsert);
+  // communicate refreshed particles using cached communication graph
+  communicate_refreshed_particles(particlestosend, particlestoinsert);
 
   // insert refreshed particles received from other processors
   insert_refreshed_particles(particlestoinsert);
@@ -383,8 +383,8 @@ void Particle::ParticleEngine::refresh_particles_of_specific_states_and_types(
   determine_specific_states_of_particles_of_specific_types_to_be_refreshed(
       particlestatestotypes, particlestosend);
 
-  // communicate particles
-  communicate_particles(particlestosend, particlestoinsert);
+  // communicate refreshed particles using cached communication graph
+  communicate_refreshed_particles(particlestosend, particlestoinsert);
 
   // insert refreshed particles received from other processors
   insert_refreshed_particles(particlestoinsert);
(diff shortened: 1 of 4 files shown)


src/particle/src/engine/4C_particle_engine.hpp
```diff
@@ -739,6 +753,12 @@ namespace Particle
 
     //! flag denoting validity of direct ghosting
     bool validdirectghosting_;
+
+    //! cached set of procs to send refreshed particle data to
+    std::set<int> refresh_send_procs_;
```
georghammerl:
```suggestion
    std::set<int> cached_procs_to_send_ghost_data_to_;
```
I'd prefer to have the variable contained the "cache" property and to make clearer the "sending to" and "receiving from" aspect.

src/particle/src/engine/4C_particle_engine.hpp
```diff
@@ -739,6 +753,12 @@ namespace Particle
...
+
+    //! cached set of procs to receive refreshed particle data from
+    std::set<int> refresh_recv_procs_;
```
georghammerl:
```suggestion
    std::set<int> cached_procs_to receive_ghost_data_from_;
```
same here

src/particle/src/engine/4C_particle_engine.hpp
```diff
@@ -604,6 +604,20 @@ namespace Particle
     void communicate_particles(std::vector<std::vector<ParticleObjShrdPtr>>& particlestosend,
         std::vector<std::vector<std::pair<int, ParticleObjShrdPtr>>>& particlestoreceive) const;
 
+    /*!
+     * \brief communicate refreshed particles using cached communication graph
+     *
+     * Uses the pre-computed refresh_send_procs_ and refresh_recv_procs_ to exchange refreshed
+     * particle data without an MPI_Allreduce for target discovery.
```
georghammerl:
```suggestion
     * particle.
```
The MPI_Allreduce  is not obvious from the code, I assume to happen it under the hood of the non-blocking communication?!

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1727,6 +1727,113 @@ void Particle::ParticleEngine::communicate_particles(
...
+    Core::Communication::UnpackBuffer buffer(rmsg);
+    while (!buffer.at_end())
+    {
+      std::shared_ptr<Core::Communication::ParObject> object(Core::Communication::factory(buffer));
+      ParticleObjShrdPtr particleobject = std::dynamic_pointer_cast<ParticleObject>(object);
+      FOUR_C_ASSERT(particleobject, "received object is not a particle object!");
+
+      particlestoreceive[particleobject->return_particle_type()].push_back(
+          std::make_pair(msgsource, particleobject));
+    }
```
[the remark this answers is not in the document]
ppraegla: Use `FOUR_C_ASSERT_ALWAYS` instead

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1727,6 +1727,113 @@ void Particle::ParticleEngine::communicate_particles(
...
+          comm_, &recvrequest[i]);
+    }
+    else
+    {
+      recvrequest[i] = MPI_REQUEST_NULL;
+    }
+  }
+
+  // wait for size sends to complete, then send data
+  MPI_Waitall(numsendtoprocs, sizesendrequest.data(), MPI_STATUSES_IGNORE);
```
ppraegla: Is this wait correct? Above, you already use MPI_Waitall to wait for the size receive requests. I would think that the size requests can only be completed when all size sends have completed.

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1727,6 +1727,113 @@ void Particle::ParticleEngine::communicate_particles(
...
+  }
+
+  // clear after all particles are packed
+  particlestosend.clear();
+
+  const int numsendtoprocs = static_cast<int>(refresh_send_procs_.size());
+  const int numrecvfromprocs = static_cast<int>(refresh_recv_procs_.size());
+
+  // send size of messages to ALL known refresh targets (0 if no data)
+  std::vector<MPI_Request> sizesendrequest(numsendtoprocs);
```
ppraegla: Does it make sense to put the following communication logic into its own function in `ParticleUtils`? This would make `communicate_refresh_particles` easier to read and would potentially allow to reuse the communication function (sending and receiving particles between a specific set of known procs). Maybe something like this
```c++
void immediate_recv_blocking_send(MPI_Comm comm, std::map<int, std::vector<char>>& sdata,
      std::map<int, std::vector<char>>& rdata, std::set<int> procs_to_send_data_to,
      std::set<int> procs_to_receive_data_from);
```

src/particle/src/engine/4C_particle_engine.cpp
slfuchs: When touching this file with this commit can you please also remove this blank lines here.

src/particle/src/engine/4C_particle_engine.cpp
slfuchs: I would also propose to move everything related to the MPI functionality into the file `4C_particle_engine_communication_utils.cpp`.

So copy the method
```c++
void Particle::ParticleUtils::immediate_recv_blocking_send(
    MPI_Comm comm, std::map<int, std::vector<char>>& sdata, std::map<int, std::vector<char>>& rdata)
```
and replace/remove everything that is not needed but known from the cached send to respectively receive from sets. The new function can look as:
```c++
void Particle::ParticleUtils::cached_immediate_recv_blocking_send(
    MPI_Comm comm, std::map<int, std::vector<char>>& sdata, std::map<int, std::vector<char>>& rdata, std::set<int> send_to_procs, std::set<int> receive_from_procs)
```

src/particle/src/engine/4C_particle_engine.cpp
slfuchs:
```c++
if not validdirectghosting_
  ParticleUtils::immediate_recv_blocking_send(comm_, sdata, rdata);
else
  ParticleUtils::cached_immediate_recv_blocking_send(comm_, sdata, rdata, cached_procs_to_send_ghost_data_to_, cached_procs_to_receive_ghost_data_from_);
```

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1727,6 +1727,113 @@ void Particle::ParticleEngine::communicate_particles(
   }
 }
 
+void Particle::ParticleEngine::communicate_refreshed_particles(
```
slfuchs: This method is not needed when introducing the if-else-case in ```Particle::ParticleEngine::communicate_particles``` and a separate method in ```4C_particle_engine_communication_utils.cpp``` as proposed above and below.

slfuchs (commented):
Very nice idea to cache the procs sending data to and receiving data from. However, to keep things clean I would propose to refactor a little bit as described in the comments.

slfuchs:
... and sorry that it took some while for reviewing! This week was quite busy. I am looking forward to see the remaining PR you mentioned.

src/particle/src/engine/4C_particle_engine.cpp
vovannikov: This wait is correct, since we need to wait for the requests to get completed on both sides: receivers and senders.

src/particle/src/engine/4C_particle_engine.hpp
vovannikov: You are right, I have removed this part of the comment as suggested.

src/particle/src/engine/4C_particle_engine.cpp
vovannikov: Very good idea, done!

src/particle/src/engine/4C_particle_engine.cpp
vovannikov: This will not work as expected.

`communicate_particles()` is also called from `ghost_particles()`, and at that point `validdirectghosting_` could still be `true` from the previous time step (it only gets invalidated after `communicate_particles()` returns, inside `insert_ghosted_particles()`). This would incorrectly route the ghosting communication through the stale cached proc sets.

However, after moving the MPI related functionality to utils, `communicate_particles()` can be modified such that `communicate_refreshed_particles()` gets removed as you have suggested, have a look at the result.

src/particle/src/engine/4C_particle_engine.cpp
vovannikov: I have removed the method and refactored `communicate_particles()` accordingly.

vovannikov:
Thank you @slfuchs, @georghammerl and @ppraegla for the feedback. Very nice proposals, they have improved this PR quite a lot.

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1788,6 +1807,13 @@ void Particle::ParticleEngine::communicate_direct_ghosting_map(
 
   // validate flags denoting validity of direct ghosting
   validdirectghosting_ = true;
+
+  // cache procs to which we send refreshed particle data
```
georghammerl:
```suggestion
  // cache procs to which this proc sends refreshed particle data
```

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1737,6 +1751,11 @@ void Particle::ParticleEngine::communicate_direct_ghosting_map(
   // invalidate flags denoting validity of direct ghosting
   validdirectghosting_ = false;
 
+  // cache procs from which we receive refreshed particle data: these are the procs that sent
```
georghammerl:
```suggestion
  // cache procs from which this proc receives refreshed particle data: these are the procs that sent
```

ppraegla (dismissed):
Nice work @vovannikov . Looking forward to the next PRs

slfuchs (dismissed):
Looks good from my side!

src/particle/src/engine/4C_particle_engine.cpp
vovannikov: Fixed

vovannikov:
I have updated the comments as suggested by @georghammerl, so ready to merge I guess.

slfuchs (approved):

ppraegla (approved):

PR #2075, opened 2026-06-05
Remove thermo trait from thermo solid trait

rjoussen:
## Description and Context

This removes the `Trait::Thermo` and hence the thermo base material from the `Trait::ThermoSolid`. Before, thermo-solid materials, which are mostly used to indicate that a solid material has thermo-mechanical coupling, needed to implement all functions that a normal thermo material (= the fourier thermo material) needs. All current thermo-solid materials solved that by wrapping an internal thermo-material with the input parameter `THERMOMAT`, which was, though, completely unused except for one occurrence in thermo-fluid-structure-interaction.

The main motivation for removing this is to allow new solid materials that want to implement thermo-mechanical coupling to avoid implementing many functions they don't need -> see #2083.

The commits are split as follows:

1. The first commit fixes a completely independent bug in the ale fluid restart code: the grid velocity wasnt correctly re-initialized, which led to a numerical drift when restarting for example tfsi problems.
2. Address the usage of the internally wrapped thermo-material of the thermo-SVK material. The scatra2 discretization now learns the solid material (of which it needs the density) from a material map clone. The tfsi test is adapted accordingly. This also works for restarts now, I added a test for that.
3. The thermo-solid specific reinit function is temporarily replaced by the pure thermo reinit function. All existing materials used this function only to get the current temperature from the thermo field anyway. This thermo reinit function becomes non-inherited once thermo trait inheritance is removed (in commit 4).
4. The `THERMOMAT` input parameter is removed. It is no longer needed.
5.  The ThermoSolid trait is no longer inheriting from the pure Thermo trait. All implemented functions are removed. The unnecessary extra inheritance layer `ThermoMechanicalMaterial` is removed as well.
6. The Solid trait is removed. It was empty anyway, and the thermo solid trait was the only place that used it.


src/fs3i/4C_fs3i_utils.cpp
```diff
@@ -0,0 +1,34 @@
...
+    Core::Elements::Element* targetele = targetdis.l_col_element(i);
+    const int gid = targetele->id();
+
+    Core::Elements::Element* sourceele = sourcedis.g_element(gid);
+
+    // For matching-grid coupling we need the structural material on the scatra element and vice
+    // versa.
+    targetele->add_material(sourceele->material());
+    sourceele->add_material(targetele->material());
+  }
```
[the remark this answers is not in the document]
rjoussen: changed to `set_material`

src/fs3i/4C_fs3i_partitioned.cpp
```diff
@@ -53,6 +54,12 @@ FS3I::PartFS3I::PartFS3I(MPI_Comm comm) : FS3IBase(), comm_(comm)
   return;
 }
 
+void FS3I::PartFS3I::validate_structure_scatra_clone_configuration(const Core::FE::Discretization&,
```
c-p-schmidt: As you do not want to do anything here, you could also just „implement“ it in the header.

src/fs3i/4C_fs3i_utils.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+  for (int i = 0; i < numelements; ++i)
+  {
+    Core::Elements::Element* targetele = targetdis.l_col_element(i);
+    const int gid = targetele->id();
+
+    Core::Elements::Element* sourceele = sourcedis.g_element(gid);
+
+    // For matching-grid TFSI coupling we need the structural material on the scatra element.
+    targetele->set_material(1, sourceele->material());
```
c-p-schmidt, line 28: This assignment is often done in the clonestrategy elsewhere, I thought. Maybe not here, because the SSI clonestrategy is (mis) because there might be no FS3I clone strategy.

src/fs3i/4C_fs3i_partitioned.cpp
rjoussen: makes sense. I think I first had it do something, that's why it is here. Im removing the implementation here.

src/fs3i/4C_fs3i_utils.cpp
rjoussen: I mainly looked into tsi for reference on how to do that, because I happened to understand it better than ssi. As far as I know, other places set materials in both directions (not needed here I think) or use `add_material`, which is dangerous in my opinion, sinde the index of the material actually matters. Does that answer your question?

maxfirmbach (approved):

c-p-schmidt (approved):

maxfirmbach:
@rjoussen Great to see this merged 👍 . Good job.

rjoussen:
Thank you @maxfirmbach!

PR #2076, opened 2026-06-08
Implement quasilinear generalized Maxwell model for structural dynamics

bgoderbauer:
## Description and Context

Adds an independent `VISCO_QuasiLinearGeneralizedMaxwell` material for Fung-type/QLV generalized Maxwell viscoelasticity in `MAT_ViscoElastHyper`.
The new model uses one isotropic hyperelastic base response and defines each Maxwell branch as a scalar multiple of that base response via $\beta_i$ and the respective relaxation time $\tau_i$. It also supports an optional parallel dashpot through `VISCOSITY`.

### Model Equations

The isotropic hyperelastic base law (from `ElastHyper`) provides the second Piola-Kirchhoff stress and Green-Lagrange strain

```math
\mathbf{S}_0(\mathbf{C}) = 2 \frac{\partial \Psi_0}{\partial \mathbf{C}},
\qquad
\mathbf{E} = \frac{1}{2}(\mathbf{C} - \mathbf{I}).
```

Each parallel Maxwell branch with internal stress state $\mathbf{Q}_i$ evolves according to

```math
\dot{\mathbf{Q}}_i+\frac{1}{\tau_i}\mathbf{Q}_i
=
\beta_i\dot{\mathbf{S}}_0.
```
The full material response is a sum of hyperelastic base and viscoelastic internal variables plus the additional parallel dashpot:
```math
\mathbf{S}_{vis} =\mathbf{S}_0 + \sum_i \mathbf{Q}_i + \mathbf{S}_{\eta},\qquad\mathbf{S}_{\eta} = \eta \dot{\mathbf{E}}.
```

### Time discretization

I've implemented two time discretization schemes, each with its own input file test
- One-step theta, with theta directly taken over from `STRUCTURAL DYNAMIC/ONESTEPTHETA`
- Exponential time discretization (first order method with implicit Euler for $\dot{\mathbf{S}}_0$ and $\dot{\mathbf{E}}$): this scheme exactly integrates the homogeneous decay of the internal variables $\mathbf{Q}_i$ and is thus well-suited for problems with many relaxation times spanning orders of magnitude.

### Usage in input file

It is used within `ViscoElastHyper`, whose combined elastic summands define the hyperelastic base response.

```yaml
- MAT: 1
    MAT_ViscoElastHyper:
      NUMMAT: 2
      MATIDS: [2, 3]
      DENS: 0.01
- MAT: 2 # or also more hyperelastic materials
    ELAST_CoupNeoHooke:
      YOUNG: 400
      NUE: 0.3
```

Material definition:

```yaml
- MAT: 3
VISCO_QuasiLinearGeneralizedMaxwell:
    BETA: [2.0, 0.1, 1.4, ...]
    TAU: [0.002, 4.2, 2.6, ...]
    SOLVE: "OneStepTheta" # or "ExponentialTimeDiscretization"
    VISCOSITY: 6.0
```

`BETA` and `TAU` may contain any number of branches, with non-negative `BETA` and positive `TAU`.


src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2471,6 +2471,25 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
 
+  /*--------------------------------------------------------------------*/
+  // quasi-linear Fung-type generalized Maxwell model
+  {
+    known_materials[Core::Materials::mes_quasilineargeneralizedmaxwell] =
+        group("VISCO_QuasiLinearGeneralizedMaxwell",
+            {
+                parameter<std::vector<double>>(
+                    "BETA", {.description = "relative branch spring weights"}),
+                parameter<std::vector<double>>("TAU", {.description = "branch relaxation times"}),
```
lauraengelhardt: I would use descriptive names for those parameters as well, as you have done for the VISCOSITY as well.

lauraengelhardt: You may also use the validators, as you wrote in the PR description, that beta >= 0 and tau > 0.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2471,6 +2471,25 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                parameter<std::string>("SOLVE",
+                    {.description = "Solution for evolution equation: OneStepTheta (default) or "
+                                    "ExponentialTimeDiscretization (ETD1 backwards forcing)",
```
lauraengelhardt: What means EDT1?

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2471,6 +2471,25 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                        .default_value = "OneStepTheta"}),
+                parameter<double>("VISCOSITY",
+                    {.description = "parallel dashpot viscosity", .default_value = 0.0}),
+            },
+            {.description = "Fung-type quasi-linear generalized Maxwell visco summand"});
```
lauraengelhardt:
```suggestion
            {.description = "Fung-type quasi-linear generalized Maxwell viscoelastic summand"});
```

tests/input_files/visco_quasilinear_generalized_maxwell_etd_h8_small.4C.yaml
```diff
@@ -0,0 +1,119 @@
...
+  ELEMENTS: 1
+  NODES: 8
+  MATERIALS: 3
+  NUMDF: 6
+PROBLEM TYPE:
+  PROBLEMTYPE: "Structure"
+DISCRETISATION:
+  NUMFLUIDDIS: 0
+  NUMALEDIS: 0
+  NUMTHERMDIS: 0
```
lauraengelhardt: I would remove all unnecessary sections from the test files.

tests/input_files/visco_quasilinear_generalized_maxwell_etd_h8_small.4C.yaml
```diff
@@ -0,0 +1,119 @@
+TITLE:
+  - "Small QLV generalized Maxwell ETD1/dashpot regression test on one HEX8 element."
```
lauraengelhardt: I would write ETD1 here not as an abbreviation. Some more description would also be nice, i.e., that this test is for testing the material `VISCO_QuasiLinearGeneralizedMaxwell` with the SOLVE option `ExponentialTimeDiscretization`.

src/mat/viscoelast/4C_mat_viscoelast_quasilineargeneralizedmaxwell.hpp
```diff
@@ -0,0 +1,179 @@
...
+            "Mat::ViscoElast::Summand::Factory.");
+        return nullptr;
+      };
+    };
+  }  // namespace PAR
+
+
+  /**
+   * \brief Parameter-backed summand for quasi-linear generalized Maxwell viscoelasticity.
+   */
```
lauraengelhardt, line 76: Some more documentation on the material would be really helpful. You could e.g. provide the equations you also included in the PR description. Maybe also clarify shortly what is the difference to the `GeneralizedMaxwell` model?

src/mat/viscoelast/4C_mat_viscoelast_quasilineargeneralizedmaxwell.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+namespace Mat::ViscoElast
+{
+  namespace PAR
+  {
+    /**
+     * \brief Parameters for quasi-linear Fung-type generalized Maxwell viscoelasticity.
+     *
+     * The branch stiffnesses are scalar multiples of the isotropic invariant response of the
+     * parallel hyperelastic base law. Each branch is defined by a non-negative relative spring
```
lauraengelhardt: I have to admit that I do not really get this part of the description: "isotropic invariant response of the parallel hyperelastic base law". What is meant with response?

src/mat/viscoelast/4C_mat_viscoelast_quasilineargeneralizedmaxwell.hpp
```diff
@@ -0,0 +1,179 @@
...
+   public:
+    explicit QuasiLinearGeneralizedMaxwell(PAR::QuasiLinearGeneralizedMaxwell* params);
+
+    Core::Materials::MaterialType material_type() const override
+    {
+      return Core::Materials::mes_quasilineargeneralizedmaxwell;
+    }
+
+    void read_material_parameters(const std::vector<double>*& beta, const std::vector<double>*& tau,
+        std::string& solve, double& viscosity) const;
```
lauraengelhardt:
```suggestion
    void read_material_parameters(const std::vector<double>& beta, const std::vector<double>& tau,
        std::string& solve, double& viscosity) const;
```
I don't think you want to be able to reassign the pointer.

src/mat/viscoelast/4C_mat_viscoelast_state.hpp
```diff
@@ -83,6 +85,27 @@ namespace Mat
       BranchHistory branch_elastic_stress_previous_;
     };
 
+    /// \brief Previous/current branch and dashpot history for quasi-linear generalized Maxwell.
+    struct QuasiLinearGeneralizedMaxwellState
```
lauraengelhardt, line 89: I took a look at the `ViscoElastState` struct. It seems like you tried to integrate your material into the existing framework, but the design is a bit awkward at the moment. The idea behind `ViscoElastState` appears to be to provide a common container for state-dependent data. However, in practice the implementation still ends up defining separate structs for each material anyway, without exposing a proper common interface.

So at this point, I would either:
- clean up the design and introduce an actual shared material-state interface, or
- implement the state structs directly inside your own material instead of relying on this convoluted implementation.

src/mat/viscoelast/4C_mat_viscoelast_quasilineargeneralizedmaxwell.hpp
```diff
@@ -0,0 +1,179 @@
...
+        const std::string& solve, int visco_mat_id, int gp, int ele_gid);
+    [[nodiscard]] const Metadata& require_metadata(const char* context, int gp, int ele_gid) const;
+    [[nodiscard]] const RuntimeContext& require_runtime_context(
+        const char* context, int gp, int ele_gid) const;
+
+    void build_metadata(const QuasiLinearGeneralizedMaxwellSetupContext& context);
+    void build_runtime_context(const ContributionSetupContext& context);
+
+    std::optional<Metadata> metadata_;
+    std::optional<RuntimeContext> runtime_context_;
```
lauraengelhardt, line 199: Some varibles/functions in here need documentation.

src/mat/viscoelast/4C_mat_viscoelast_quasilineargeneralizedmaxwell.cpp
```diff
@@ -0,0 +1,603 @@
...
+  const QuasiLinearGeneralizedMaxwellContribution::Metadata&
+  QuasiLinearGeneralizedMaxwellContribution::require_metadata(
+      const char* context, const int gp, const int ele_gid) const
+  {
+    FOUR_C_ASSERT_ALWAYS(metadata_.has_value(),
+        "Missing VISCO_QuasiLinearGeneralizedMaxwell metadata cache while {} in "
+        "MAT_ViscoElastHyper (GP {}, ELE {}). Run setup() before evaluation.",
+        context, gp, ele_gid);
+
+    FOUR_C_ASSERT_ALWAYS(!metadata_->tau.empty() || metadata_->viscosity > 0.0,
```
lauraengelhardt: If you use the validators in the input, could you skip this?

lauraengelhardt (commented):
I left some comments regarding documentation/naming and one larger comment regarding the overall design.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
bgoderbauer: True, I forgot the input validators! I'll add them. Regarding descriptive names: TAU could be easily replaced with RELAXATION_TIMES, but for BETA I can't really think of a short and descriptive name, since it is a model-specific parameter. Many models have characteristic variables where you don't know what they are doing by their name, but that's what the description is for, right? This would feel very different from all the other material implementations I have seen if I were to change to long descriptive names here. I can add a little more explanation to the description.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
bgoderbauer: First order Exponential Time Differencing/Discretization. Just the short name of the numerical scheme.

src/mat/viscoelast/4C_mat_viscoelast_quasilineargeneralizedmaxwell.hpp
bgoderbauer: I'll extend the documentation. Generally, it means that this material needs an isotropic hyperelastic material law as a basis, and the material law of the springs of the different Maxwell branches are scaled versions of it.

src/mat/viscoelast/4C_mat_viscoelast_state.hpp
bgoderbauer: I extracted this functionality, covering everything related to history data, in #2018 directly from the ViscoElastHyper class (a place where it definitely doesn't belong). It was a design choice there to reduce what this orchestration layer is allowed to do. I agree that this is not a perfect abstraction for handling viscoelastic history independent of the model, but I'd say the current status plays an important role in splitting the general infrastructure. It centralizes the lifecycle of history states, including setup, packing, communication, and history updates, so that this is clearly separated from the orchestration layer and the constitutive updates of the different models. Importantly, we allow mixed viscoelastic materials, so something within ViscoElastHyper must manage all states in a distinct order. This functionality cannot be easily implemented in the different material implementations currently; you would have to carefully duplicate the lifecycle contracts everywhere.

I could see this as a future refactoring step, but it is out of scope for this PR.

bgoderbauer:
I also added a test with a porofluid problem, it seemed to work out of the box :). @lkoeglmeier could you please check whether the parameters make sense here?

bgoderbauer:
Thanks @lkoeglmeier for helping me refine the test and @lauraengelhardt for the detailed review! :smiley:

lkoeglmeier (approved):

lauraengelhardt (approved):
Nice, thanks for improving the documentation and the other changes.

PR #2079, opened 2026-06-12
Remove custom implemented functions: Part 4

c-p-schmidt:
## Description and Context

This PR is the fourth and final step toward removing custom-implemented functions from elchsinglemat and related materials. Different functions can be chosen by „magic“ negative numbers in the function definition, along with a combination of parameters, which is a nightmare to use. This final step now removes the constant function in several steps, as it was quite a lot to do. The commits of this PR are structured as:

- Adapts the definition of the diffusion coefficient and conductivity
- Adapts the definition of the thermodynamic factor
- Adapts the definition of the transference number
- Does a final cleanup by removing the now unused methods, as only the function framework and constant values are used now   

Follows #2042 #2061 #2064


src/mat/4C_mat_elchsinglemat.cpp
```diff
@@ -18,363 +18,305 @@ FOUR_C_NAMESPACE_OPEN
...
-void Mat::PAR::ElchSingleMat::check_provided_params(
-    const int functnr, const std::vector<double>& functparams)
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+const Core::Utils::FunctionOfTime&
+Mat::PAR::ElchSingleMat::diffusion_coefficient_concentration_scaling_funct()
 {
-  // name of specified curve
-  std::string functionname;
+  FOUR_C_ASSERT(has_diffusion_coefficient_concentration_scaling(),
```
dragos-ana, line 39: Shouldn't this be always asserted via 'FOUR_C_ASSERT_ALWAYS'?

src/mat/4C_mat_elchsinglemat.cpp
```diff
@@ -18,363 +18,305 @@ FOUR_C_NAMESPACE_OPEN
...
-      {
-        FOUR_C_THROW("Curve number {} is not implemented", functnr);
-      }
-    }
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+const Core::Utils::FunctionOfTime&
+Mat::PAR::ElchSingleMat::diffusion_coefficient_temperature_scaling_funct()
+{
+  FOUR_C_ASSERT(has_diffusion_coefficient_temperature_scaling(),
```
dragos-ana, line 59: Same here.

dragos-ana (dismissed):
Looks good to me, added just a small question.

src/mat/4C_mat_elchsinglemat.cpp
c-p-schmidt: As this is a check for proper code usage and cannot be triggered by "wrong" usage in the input, but only by developers during development, I prefer having the check only in assert (debug) mode and not in production runs (release without assertions).

src/mat/4C_mat_elchsinglemat.cpp
c-p-schmidt: The same applies here.

src/mat/4C_mat_scl.cpp
```diff
@@ -39,12 +39,26 @@ Mat::PAR::Scl::Scl(const Core::Mat::PAR::Parameter::Data& matdata)
...
-    FOUR_C_THROW("number of materials {} does not fit to size of material vector {}",
-        transnrparanum_, transnr_.size());
+}
+
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+const Core::Utils::FunctionOfScalar&
+Mat::PAR::Scl::transference_number_concentration_scaling_funct()
+{
+  FOUR_C_ASSERT(has_transference_number_concentration_scaling(),
```
lauraengelhardt, line 49: Shouldn't this be a throw?

src/mat/4C_mat_newman.cpp
```diff
@@ -17,32 +17,62 @@
...
+  }
+  return transference_number_concentration_scaling_funct_->get();
+}
+
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+const Core::Utils::FunctionOfTime&
+Mat::PAR::Newman::thermodynamic_factor_concentration_scaling_funct()
+{
+  FOUR_C_ASSERT(has_thermodynamic_factor_concentration_scaling(),
```
lauraengelhardt, line 60: Also this, should it be a throw?

src/mat/4C_mat_newman.cpp
```diff
@@ -17,32 +17,62 @@
...
-  check_provided_params(transnrcurve_, transnrpara_);
-  check_provided_params(thermfaccurve_, thermfacpara_);
 }
 
+/*----------------------------------------------------------------------*
+ *----------------------------------------------------------------------*/
+const Core::Utils::FunctionOfTime&
+Mat::PAR::Newman::transference_number_concentration_scaling_funct()
+{
+  FOUR_C_ASSERT(has_transference_number_concentration_scaling(),
```
lauraengelhardt, line 40: Same here.

src/mat/4C_mat_scl.cpp
c-p-schmidt: I don't think so. See my reply to Dragos: https://github.com/4C-multiphysics/4C/pull/2079#discussion_r3411473825.
Do you agree?

lauraengelhardt: I have to admit that I don't quite understand. The error message says: `"You try to access the function defining the concentration scaling of the transference "
      "number, but no function number has been set in the input file! Check your input file and "
      "set a positive integer for 'TRANSFERENCE_NR_CONC_SCALE_FUNCT' to use a function from the "
      "function framework."`, so I would assume that will throw/not throw depending on the input (which would only be available during runtime).

I'm not familiar with the framework though, so maybe I am getting something wrong here :)

c-p-schmidt:
After talking to @lauraengelhardt we decided that the wording of the error message in the assert should be different. This has changed with the last push.

lauraengelhardt (approved):

bgoderbauer (approved):

PR #2082, opened 2026-06-14
Further cleanup in preparation for the Adaptive Estimate Interpolation algorithm

dragos-ana:
- The term 'constitutive update' is now used instead of 'return mapping' to avoid confusion, since our constitutive update also comprises the elastic predictor verification. Several frameworks in literature use 'return mapping' to comprise both stages, but return mapping is actually the second corrector stage only.  We want to make this distinction clear in the code.
- Added struct for tracking all deformation tensors used for the local integration routines. These are then employed consistently throughout the constitutive update routines.
- Improved error handling by adding error registration settings, and only outputting detailed
errors. Plastic strain rate derivative checks are now disabled by default, since the detecting overflow here is not straightforward and can overconstrain the computation. 
This has some implications for the substepping test -> changed results.
- Several functions are guarded to only be evaluated if there are no evaluation errors currently.
- Added tracking of plastic flow at each Gauss point to improve plasticity checks, such as the ones in
the linearization and the update procedures.
- The update procedures are run separately per Gauss point instead of over all Gauss points, to only loop once over the GP in the update method of the material.
- Also added characteristic for viscoplastic laws: do they use yield surface or not? Important for the AEI procedures subsequently.
- Incorporated further logic from the viscoplastic factor to the dedicated Local Newton manager, such as the solution, along with the tracking and checking of convergence quantities.
- Added a dedicated unit tests file for the service methods of the inelastic deformation gradient factors.
- In preparation for the AEI: a method dedicated to determining the Local Newton estimate was used. This method will serve as the entry point for the AEI, along with manage_evaluation, which will be responsible for re-estimation.
- Substepping: now the deformation gradient is interpolated, instead of the right CG tensor. This is also done in preparation for the AEI, to enable combining it with substepping.


src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2906,46 +2906,38 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
                                 "automatic method selection based on matrix characteristics)",
                     .default_value = Core::LinAlg::MatrixExpCalcMethod::automatic}),
             parameter<Core::LinAlg::GenMatrixExpFirstDerivCalcMethod>(
                 "MATRIX_EXP_DERIV_CALC_METHOD",
-                {.description = "chosen computation method for the first derivative of the matrix "
-                                "exponential w.r.t. matrix (default: automatic method selection "
+                {.description = "chosen computation method for the first derivative of the "
+                                "matrix "
+                                "exponential w.r.t. matrix (default: automatic method "
+                                "selection "
```
lauraengelhardt: There seems to be something wrong with the formatting here and in some other places. I think quite some changes in the diff are due to this. Maybe you could fix that first such that the actual changes are not obscured by these.

dragos-ana: Thank you for the suggestion. I think it should be fine now ;)

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -3007,6 +2996,31 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                {parameter<bool>("REGISTER_PLASTIC_STRAIN_INCR_OVERFLOW",
+                     {.description = "should overflow error be registered via ErrorType when the "
+                                     "plastic strain increment exceeds the specified tolerance?",
+                         .default_value = true}),
+                    parameter<double>("MAX_PLASTIC_STRAIN_INCR",
+                        {.description = "maximum evaluable plastic strain increment "
+                                        "used for registering overflow errors",
+                            .default_value = std::exp(30.0),
+                            .validator = positive<double>()}),
+                    parameter<bool>("REGISTER_PLASTIC_STRAIN_INCR_DERIVS_OVERFLOW",
```
c-p-schmidt: Why did you not rename the flag in a similar manner as the above?
```suggestion
                    parameter<bool>("REGISTER_PLASTIC_STRAIN_DERIV_INCR_OVERFLOW",
```

src/mat/4C_mat_inelastic_defgrad_factors.hpp
```diff
@@ -2028,11 +2000,14 @@ namespace Mat
      *
      *
      * @param[in] err_status error status
-     * @param[out] eval_action action to be performed subsequently in the Local Newton Loop
+     * @param[in] deftensors deformation tensors used for local time integration
```
c-p-schmidt: Why did you add these? They are not arguments of the method, at least how it is displayed on GitHub.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -1856,9 +1875,9 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::pre_evaluate(
   }
 
   // set time step
-  FOUR_C_ASSERT(context.time_step_size, "Time step size not given in evaluation context.");
+  FOUR_C_ASSERT_ALWAYS(context.time_step_size, "Time step size not given in evaluation context.");
```
c-p-schmidt, line 1845: To me, the original assertion is fine as well. This is the developer's responsibility to set, and it's not dependent on the input file alone, I guess.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3283,6 +3319,9 @@ void Mat::InelasticDefgradTransvIsotropElastViscoplast::setup(const int numgp,
   // of Gauss points
   local_newton_manager_.resize(numgp);
 
+  // resize plastic flow tracking vector with the correct number of Gauss points
+  is_plastic_gp_.resize(numgp, is_plastic_gp_[0]);
```
c-p-schmidt, line 3278: Don't you want to use assign? This would also override all original data from is_plastic_gp_ in case something undesired would be in there.
```suggestion
  is_plastic_gp_.assign(numgp, false);
```

unittests/mat/vplast/4C_vplast_reform_johnsoncook_test.cpp
```diff
@@ -108,17 +108,45 @@ namespace
...
+    const double max_plastic_strain_increments = std::exp(30.0);
     double plastic_strain_rate_reformulated_JC =
-        vplast_law_reformulated_JC_->evaluate_plastic_strain_rate(
-            equiv_stress_, equiv_plastic_strain_, 1.0, std::exp(30.0), err_status, false);
+        vplast_law_reformulated_JC_->evaluate_plastic_strain_rate(equiv_stress_,
+            equiv_plastic_strain_, 1.0,
+            {.register_plastic_strain_incr_overflow = true,
+                .max_plastic_strain_incr = max_plastic_strain_increments,
+                .register_plastic_strain_incr_derivs_overflow = true,
+                .max_plastic_strain_deriv_incr = max_plastic_strain_increments},
```
c-p-schmidt: I guess you intended to set the derivative to the same value as well?

src/mat/4C_mat_inelastic_defgrad_factors.hpp
dragos-ana: They will be added in the next PR, and I forgot to remove their docs.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: True, but I would keep it like this, since the general strategy I use within the framework is to always assert.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: This is not important to be honest, it's just a resizing. The real values will get set at each GP during the local integration anyway. Since the entire setup function currently works with `resize`, I would keep it like this.

unittests/mat/vplast/4C_vplast_reform_johnsoncook_test.cpp
dragos-ana: Yes, they should be the same value here. 
I adapted the name to `max_plastic_strain_incr_and_deriv_incr` to make this more clear, thank you for the question.

tests/input_files/mat_iso_viscoplast_refJC_log_timint_substepping.4C.yaml
```diff
@@ -125,19 +125,19 @@ RESULT DESCRIPTION:
...
       QUANTITY: "stress_zz"
-      VALUE: 9.30776106020185466e+02
+      VALUE: 9.31099439482890148e+02
       TOLERANCE: 0.00098
   - STRUCTURE:
       DIS: "structure"
       NODE: 8
       QUANTITY: "stress_zz"
-      VALUE: 9.30776106016615358e+02
+      VALUE: 9.31099439481397212e+02
```
lauraengelhardt, line 140: Just for my understanding: Can I ask what is the reason for the change of these values. I wouldn't expect the result to change that much just from a pure cleanup.

lauraengelhardt (commented):
Thanks for cleaning up the formatting first 👍 That already removed a large part of the diff. I have one short question considering the test results. 

I haven’t had time to fully review the rest of the PR yet, since all changes are contained in a single large commit (and also its late 😉). IMO, it would be very helpful to split changes into smaller commits - for example one for the error handling improvements, one for adding the unit tests, and one for the AEI preparation. This would make reviewing much easier 🙂 No need to do it here, but maybe something to keep in mind for the next PRs.

tests/input_files/mat_iso_viscoplast_refJC_log_timint_substepping.4C.yaml
dragos-ana: We trigger local substepping if overflow is identified based on the plastic strain increments and the increments of their derivatives. Previously, we registered overflow for both plastic strain increments and derivative increments; however, I have seen in previous weeks that this strategy can overly constrain the local time integration. I have therefore adapted the formulation to only check overflow based on the plastic strain increments by default, and not also based on the derivative increments. This means that the overflow checking mechanism has changed, which leads to differences in the triggered substepping. Hence the observed differences in the test results.

The changes are substantial, I agree. But this is a weakness of the substepping formulation, which is very fragile and inconsistent anyway in my opinion, and should be removed or deprecated in some way in upcoming PRs on this material.

rjoussen: The result changes make sense to me and are also changes in the "right" direction, compared to the non-substepping test

src/mat/vplast/4C_mat_vplast_law.hpp
```diff
@@ -78,7 +78,11 @@ namespace Mat
       static std::shared_ptr<Law> factory(int matnum);
 
       /// provide material type
-      virtual Core::Materials::MaterialType material_type() const = 0;
+      [[nodiscard]] virtual Core::Materials::MaterialType material_type() const = 0;
+
+      /// does the viscoplastic law use a yield surface formulation, or is it a no-yield-surface
+      /// law?
+      [[nodiscard]] virtual bool uses_yield_surface() const = 0;
```
rjoussen, line 91: I guess I'll need to see how this is actually used. Currently, I don't fully understand why viscoplastic laws without a strict yield surface require a different approach in the parent material. I would rather think the important distinction is whether or not viscoplastic laws manage their own history variables.

src/mat/vplast/4C_mat_vplast_law.hpp
```diff
@@ -187,17 +192,20 @@ namespace Mat
...
+      virtual void update(const unsigned int gp) = 0;
 
       /*!
        * @brief Update the history variables for a specific GP after a converged substep
        * (substepping procedure)
        *
        * @param[in] gp      Gauss point
        */
-      virtual void update_gp_state(int gp) = 0;
+      virtual void update_gp_state(const unsigned int gp) = 0;
```
rjoussen: This was of course here before, but these methods should either be unified if possible or have more expressive names. But its hard to assess without any code in these functions.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -3007,6 +2996,31 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 },
                 {.description = "Parameters used in the Local Newton--Raphson procedure "
                                 "(viscoplastic corrector stage)",
+                    .required = false}),
+            group("ERROR_REGISTRATION_SETTINGS",
```
```suggestion
            group<ErrorRegistrationSettings>("ERROR_REGISTRATION_SETTINGS",
```
I think you can do something like this here and then use the `.store` function to get the constructed struct directly

src/mat/vplast/4C_mat_vplast_reform_johnsoncook.cpp
```diff
@@ -178,8 +179,12 @@ double Mat::Viscoplastic::ReformulatedJohnsonCook::evaluate_plastic_strain_rate(
...
   // verify whether the maximum plastic strain increment is larger than 0, since we will take its
   // logarithm
-  FOUR_C_ASSERT_ALWAYS(max_plastic_strain_incr > 0.0,
-      "Maximum plastic strain increment must be > 0: current value = {}", max_plastic_strain_incr);
+  if (error_registration_settings.register_plastic_strain_incr_overflow)
+  {
+    FOUR_C_ASSERT_ALWAYS(error_registration_settings.max_plastic_strain_incr > 0.0,
+        "Maximum plastic strain increment must be > 0: current value = {}",
+        error_registration_settings.max_plastic_strain_incr);
+  }
```
rjoussen: This is coming directly from the input, where it is validated already. So I dont think another assert is needed here.

src/mat/vplast/4C_mat_vplast_reform_johnsoncook.cpp
```diff
@@ -281,17 +289,22 @@ Mat::Viscoplastic::ReformulatedJohnsonCook::evaluate_derivatives_of_plastic_stra
...
     // will take its logarithm
-    FOUR_C_ASSERT_ALWAYS(max_plastic_strain_deriv_incr > 0.0,
-        "Maximum plastic strain derivative increment must be > 0: current value = {}",
-        max_plastic_strain_deriv_incr);
+    if (error_registration_settings.register_plastic_strain_deriv_incr_overflow)
+    {
+      FOUR_C_ASSERT_ALWAYS(error_registration_settings.max_plastic_strain_deriv_incr > 0.0,
+          "Maximum plastic strain derivative increment must be > 0: current value = {}",
+          error_registration_settings.max_plastic_strain_deriv_incr);
+    }
```
rjoussen: see my other comment.

src/mat/vplast/4C_mat_vplast_law.hpp
```diff
@@ -128,15 +132,14 @@ namespace Mat
...
-       * @note?
+       * @param[in] error_registration_settings Settings used for error registration.
+       * @param[out] err_status Output variable: error due the term considered in @note?
        * @return Equivalent plastic strain rate \f$ \dot{\varepsilon}^{\text{p}} \f$
        */
       virtual double evaluate_plastic_strain_rate(const double equiv_stress,
-          const double equiv_plastic_strain, const double dt, const double max_plastic_strain_incr,
+          const double equiv_plastic_strain, const double dt,
+          const InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorRegistrationSettings&
+              error_registration_settings,
```
rjoussen: Are the error_registrations_settings actively changed by the parent material/the caller? Because if they're not, the should be set for the viscoplastic law once in the constructur and then reused.

src/mat/4C_mat_inelastic_defgrad_factors.hpp
```diff
@@ -2028,11 +2000,11 @@ namespace Mat
      *
      *
      * @param[in] err_status error status
-     * @param[out] eval_action action to be performed subsequently in the Local Newton Loop
+     * @param[out] eval_action action to be performed subsequently in the local Newton Loop
      */
     void manage_evaluation(
         const InelasticDefgradTransvIsotropElastViscoplastUtils::ErrorType& err_status,
-        InelasticDefgradTransvIsotropElastViscoplastUtils::EvaluationAction& eval_action) const;
+        InelasticDefgradTransvIsotropElastViscoplastUtils::EvaluationAction& eval_action);
```
rjoussen, line 1996: I think this methods can stay const for now.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3178,20 +3205,20 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_history_variables_wr
...
-  CredM.multiply_tn(1.0, FredM, FredM, 0.0);
+  // construct struct containing deformation tensors used for local integration
+  ViscoplastUtils::LocalIntegrationDeformationTensors deftensors(
+      FredM, time_step_quantities_.last_plastic_defgrad_inverse[gp_]);
 
   // perform non-repeatable pre-evaluation tasks (non-repeatable: not
   // called in the redundant evaluate call, which is already handled -> direct return
   // without calling this function)
-  prepare_return_mapping();
+  prepare_constitutive_update();
```
rjoussen: Will `prepare_constitutive_update()` do more in the future? Currently it is so small that I would be in favor of just doing the necessary things here explicitly. This way, its less hidden. On top of that, prepare* sound like something that needs to be done before the actual function call, which is not the case here.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3201,13 +3228,13 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::return_mapping(
...
   time_step_quantities_.current_temperature[gp_] = temperature;
   thermo_mechanical_coupling_cache_.reset(gp_);
   // check whether the predictor is the solution (no plastic strain during this time step)
-  bool pred_is_sol =
-      check_elastic_predictor(CredM, temperature, iFinM_pred, plastic_strain_pred, err_status);
+  bool pred_is_sol = check_elastic_predictor(
+      deftensors.right_cg, temperature, iFinM_pred, plastic_strain_pred, err_status);
   if ((err_status == ViscoplastUtils::ErrorType::no_errors) && (pred_is_sol))
   {
     // update inverse inelastic defgrad and plastic strain
```
rjoussen, line 3195:
```suggestion
    // update inverse inelastic defgrad and plastic strain
    is_plastic_gp_[gp_] = false;
```

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3693,48 +3759,63 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::viscoplastic_correction(
...
+    {
+      return Core::LinAlg::Matrix<10, 1>{Core::LinAlg::Initialization::zero};
+    }
+    // without substepping, we throw directly
+    else
+    {
+      FOUR_C_THROW(
+          "{}", get_error_warning_info("Could not compute initial estimate for the local Newton!"));
+    }
+  }
```
rjoussen, line 3759: this only makes sense if `determine_local_newton_init_estimate` will be able to set an error status.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -3693,48 +3759,63 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::viscoplastic_correction(
...
-      .residual_norm = 1.0, .increment_norm = 1.0};
-
   // initialize evaluation management action
   ViscoplastUtils::EvaluationAction eval_action{
       ViscoplastUtils::EvaluationAction::continue_current_iteration};
 
-  // reset Local Newton iteration count
-  local_newton_manager_.set_iteration_count(0);
+  // initialize local Newton
+  local_newton_manager_.reset_iter();
```
rjoussen, line 3741: Would it be possible to construct a local (in a programming sense) local_newton_manager in every local newton loop? This would probably prevent a lot of resetting everywhere. The only thing that would need to change is that the number of iterations is (only once in the end) written to a class variable per gp, such that they can be plotted.

rjoussen (commented):
Exciting to see that we're moving towards AEI! If that helps you with merging the next step, I would also be fine with removing substepping first, and then add AEI on a cleaner state.

I left some comments and also have a more general question regarding overflow errors:

What exactly is the idea behind the "manual" overflow checking in the first place? In my simulations, I often end up setting these values super high so the simulation doesn't over-cautiously stop. As far as I know/remember, resulting nan-values are caught later on anyway. Or I guess formulated differently: Why are we setting a limit on these values when we could also just abort/trigger substepping/estimate adaptation when overflow actually happens?

dragos-ana:
> Exciting to see that we're moving towards AEI! If that helps you with merging the next step, I would also be fine with removing substepping first, and then add AEI on a cleaner state.
> 
> I left some comments and also have a more general question regarding overflow errors:
> 
> What exactly is the idea behind the "manual" overflow checking in the first place? In my simulations, I often end up setting these values super high so the simulation doesn't over-cautiously stop. As far as I know/remember, resulting nan-values are caught later on anyway. Or I guess formulated differently: Why are we setting a limit on these values when we could also just abort/trigger substepping/estimate adaptation when overflow actually happens?
@rjoussen:

It's fine if we proceed with substepping now, I implemented the AEI in a way such that this is also possible for substepping ;)  As I said, substepping is used in the literature a lot, so I would leave it for now within the codebase.

Regarding your question on the manual overflow checking: I think that this is required to cleanly identify overflow in the different evaluated quantities, such as plastic strain rates and plastic velocity gradients. As far as I see, the nan checks are only for the stress and the plastic strain, so non-nan values could still lead to overflow in the quantities above. Not sure that I understand your last question correctly, but both substepping and estimate adaptation are triggered when overflow / other errors are identified. So they need to be identified somehow before overflow actually happens in the code, because this will simply throw. At some point I had the entire procedure using `try--catch` logic, but this was not that easy to manage and less explicit. Does that answer your question?

src/mat/vplast/4C_mat_vplast_law.hpp
dragos-ana: It's required in the AEI, which has different logic when determining the plastic predictor for laws with and without yield surfaces. I can remove it though if it's confusing here and add it later on.

src/mat/vplast/4C_mat_vplast_law.hpp
dragos-ana: Good catch, I renamed it to `update_gp_state_after_substep`

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
dragos-ana: Nice,  I wasn't aware of this, thank you! I adapted this for both `LocalNewtonParams` and `ErrorRegistrationSettings`

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Removed.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Not required because this is already set to false previously.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: It will once the AEI is incorporated. I have removed it for now.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
dragos-ana: Theoretically possible, but I don't really the see the benefit, or rather the problem with the current approach, since the entire reset procedure is just reset of the iteration count and saving the initial estimate.
I would like the manager to manage the quantities to be visualized at all Gauss points rather than putting this into the material directly (currently only the number of iterations, but things such as the residuals or the increments over the local iterations could also be interesting at some point).

dragos-ana:
@c-p-schmidt @rjoussen: I added another commit which integrates the recent feedback. Once this is approved, I'll squash and then this can be merged.

src/mat/vplast/4C_mat_vplast_law.hpp
rjoussen: Its okay for me to keep it now.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2962,51 +2952,121 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                 {.description = "Settings for the usage of local substepping to integrate the "
                                 "viscoplastic evolution equations",
                     .required = false}),
-            group("LOCAL_NEWTON",
+            group<Mat::InelasticDefgradTransvIsotropElastViscoplastUtils::LocalNewtonParams>(
```
rjoussen: Maybe define `namespace ViscoplastUtils = Mat::InelasticDefgradTransvIsotropElastViscoplastUtils;` at the beginning of this scope to make this a little prettier?

rjoussen (dismissed):

src/mat/4C_mat_inelastic_defgrad_factors.cpp
rjoussen: Okay, that works for me.

c-p-schmidt (dismissed):

rjoussen (approved):

c-p-schmidt (approved):

PR #2083, opened 2026-06-15
Add thermal effects to the multiplicative split material

rjoussen:
## Description and Context
For now, the last PR to make the viscoplastic factor thermo-coupled:
- Heat source due to plastic dissipation is wired from the viscoplastic factor to the thermo element implementation.
- Stress contribution due to thermal expansion is added
- Two new regression tests are added, for partitioned and monolithic TSI, respectively. Including both is because the call order and, hence, the caching behavior are quite different for the two algorithms. They use identical result tests to demonstrate consistency. Though it should be noted that using a partitioned algorithm for the viscoplastic material in its current form is not recommended, since this leads to many re-evaluations of the material without a valid cache being present and consequently repeated return-mappings (or "constitutive updates" in the future), which probably mitigates all advantages a partitioned algorithm might have.

Current Limitations:
- Supported only for a single viscoplastic factor.
- The thermo-elastic contribution to the thermo equation via `stress_temperature_modulus_and_deriv` is neglected so far.

following #2041 
Doesn't touch any factor code, so independent of #2082 (@dragos-ana) Edit: I ended up re-designing some of the `ThermalQuantities` struct, mainly due to performance concerns, which now lead to some changes in the factor, but quite limited. The thermal expansion evaluations now also use tensors instead of matrices.


src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.cpp
```diff
@@ -45,7 +47,9 @@ Mat::PAR::MultiplicativeSplitDefgradElastHyper::MultiplicativeSplitDefgradElastH
       matids_elast_(matdata.parameters.get<std::vector<int>>("MATIDSEL")),
       numfac_inel_(matdata.parameters.get<int>("NUMFACINEL")),
       inel_defgradfacids_(matdata.parameters.get<std::vector<int>>("INELDEFGRADFACIDS")),
-      density_(matdata.parameters.get<double>("DENS"))
+      density_(matdata.parameters.get<double>("DENS")),
+      ref_temperature_(matdata.parameters.get_or<double>("REF_TEMPERATURE", 0.0)),
+      thermal_expansion_coefficient(matdata.parameters.get<double>("THERMAL_EXPANSION_COEFFICIENT"))
```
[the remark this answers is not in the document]
rjoussen: both have a default value.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.hpp
```diff
@@ -65,6 +71,12 @@ namespace Mat
       /// material mass density
       const double density_;
 
+      /// reference temperature for thermal expansion
+      const double ref_temperature_;
+
+      /// thermal expansion coefficient
+      const double thermal_expansion_coefficient;
```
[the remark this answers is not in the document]
rjoussen: agreed

src/thermo/src/element/4C_thermo_ele_impl.cpp
```diff
@@ -2116,13 +2130,39 @@ void Discret::Elements::TemperImpl<distype>::nonlinear_dissipation_fint_tang(
...
+      // with MechDiss := sqrt(2/3) . sigma_y(T_{n+1}) . Dgamma
+      Dmech = thermoplhyperelast->mech_diss(iquad) / stepsize;
+    }
+    else if (structmat->material_type() ==
+             Core::Materials::m_multiplicative_split_defgrad_elasthyper)
+    {
+      if constexpr (nsd_ == 3)
+      {
+        heat_source = multiplicative_split_defgrad_elast_hyper_ptr->evaluate_additional_heat_source(
+            eval_context, iquad, ele->id(), &defgrd, current_temperature(0));
```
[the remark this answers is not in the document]
rjoussen: This mirrors the behavior elsewhere in the thermo element implementation.

src/thermo/src/element/4C_thermo_ele_impl.cpp
```diff
@@ -2116,13 +2130,39 @@ void Discret::Elements::TemperImpl<distype>::nonlinear_dissipation_fint_tang(
     // coordinates
     eval_shape_func_and_derivs_at_int_point(intpoints, iquad, ele->id());
 
+    // (material) deformation gradient F
+    // F = d xcurr / d xrefe = xcurr^T . N_XYZ^T
+    defgrd.multiply_tt(xcurr, derxy_);
+
+    current_temperature.multiply_tn(funct_, etempn_);
```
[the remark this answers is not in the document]
rjoussen: This mirrors the behavior elsewhere in the thermo element implementation.

unittests/mat/4C_multiplicative_split_defgrad_elasthyper_test.cpp
```diff
@@ -547,8 +550,10 @@ namespace
     stress_fact.gamma = gamma_ref_;
     stress_fact.delta = delta_ref_;
 
-    Core::LinAlg::Matrix<6, 9> dSdiFin =
-        multiplicative_split_defgrad_->evaluated_sdi_fin(kinemat_quant, stress_fact);
+    Mat::ThermalQuantities
+        thermal_quantities;  // empty thermal quantities are unphysical but sufficient here.
+    Core::LinAlg::Matrix<6, 9> dSdiFin = multiplicative_split_defgrad_->evaluated_sdi_fin(
+        kinemat_quant, stress_fact, thermal_quantities);
```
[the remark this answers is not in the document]
rjoussen: refactored.

src/thermo/src/element/4C_thermo_ele_impl.cpp
```diff
@@ -2064,50 +2070,58 @@ void Discret::Elements::TemperImpl<distype>::nonlinear_dissipation_fint_tang(
...
+  Core::LinAlg::Matrix<nen_, nsd_> xcurr;      // current  coord. of element
+  Core::LinAlg::Matrix<nen_, nsd_> xcurrrate;  // current  velocity of element
+
+  std::vector<double> vel(disp.size(), 0.0);  // dummy velocity vector
+
+  initial_and_current_nodal_position_velocity(ele, disp, vel, xcurr, xcurrrate);
+
   // initialise the deformation gradient w.r.t. material configuration
   Core::LinAlg::Matrix<nsd_, nsd_> defgrd(Core::LinAlg::Initialization::uninitialized);
+  Core::LinAlg::Matrix<1, 1> current_temperature(Core::LinAlg::Initialization::uninitialized);
```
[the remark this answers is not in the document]
rjoussen: This mirrors the behavior elsewhere in the thermo element implementation.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.cpp
```diff
@@ -1066,6 +1142,48 @@ void Mat::MultiplicativeSplitDefgradElastHyper::evaluate_od_stiff_mat(PAR::Inela
...
+  if (num_contributions == 1)
+  {
+    if (const auto viscoplastic_factor =
+            std::dynamic_pointer_cast<Mat::InelasticDefgradTransvIsotropElastViscoplast>(
+                facdefgradin[0].second))
+    {
+      heat_source = viscoplastic_factor->evaluate_taylor_quinney_heat_source(context, gp, eleGID,
+          current_defgrad, Core::LinAlg::identity_matrix<3>(), current_temperature);
+    }
+  }
```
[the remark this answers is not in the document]
rjoussen: Heat source initializes to 0.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.cpp
```diff
@@ -983,8 +1060,7 @@ void Mat::MultiplicativeSplitDefgradElastHyper::evaluate_od_stiff_mat(PAR::Inela
     const Core::LinAlg::Matrix<3, 3>* const defgrad, const Core::LinAlg::Matrix<6, 9>& dSdiFin,
     Core::LinAlg::Matrix<6, 1>& dstressdx)
 {
-  // clear variable
-  dstressdx.clear();
+  // dstressdx might already contain a thermal contribution.
```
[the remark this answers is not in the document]
rjoussen: Refactored

rjoussen:
Thanks for running copilot @c-p-schmidt, mine ran out of tokens apparently :)

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.hpp
```diff
@@ -276,6 +287,112 @@ namespace Mat
         const Teuchos::ParameterList& params, const EvaluationContext<3>& context, int gp,
         int eleGID) override;
 
+    void reinit(const Core::LinAlg::Tensor<double, 3, 3>* defgrd,
+        const Core::LinAlg::SymmetricTensor<double, 3, 3>& glstrain, double temperature,
+        unsigned gp) override { /* do nothing */ };
+
```
dragos-ana, line 293: This has nothing to do with your PR, but `reinit` is a very strange /  non-descriptive name for what the function does. 

Maybe a matter for private discussion, but shouldn't we at least include the stress-temperature modulus into the residual and neglect the linearization first?

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.hpp
```diff
@@ -276,6 +287,112 @@ namespace Mat
...
+    void stress_temperature_modulus_and_deriv(Core::LinAlg::SymmetricTensor<double, 3, 3>& stm,
+        Core::LinAlg::SymmetricTensor<double, 3, 3>& stm_dT, const int gp) override;
+
+    /*!
+     * @brief Evaluate the mechanical dissipation produced by this material. Currently, this is only
+     * enabled for the viscoplastic factor.
```
dragos-ana:
```suggestion
     * @brief Evaluate the heat source produced by this material. Currently, this is only
     * enabled for the viscoplastic factor.
```
To prevent confusion as also noted in a previous PR ;)  It's still my opinion that this is better named as the Taylor-Quinney heating, but I am also fine with heat source.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.cpp
```diff
@@ -1064,8 +1128,52 @@ void Mat::MultiplicativeSplitDefgradElastHyper::evaluate_od_stiff_mat(PAR::Inela
...
+            std::dynamic_pointer_cast<Mat::InelasticDefgradTransvIsotropElastViscoplast>(
+                facdefgradin[0].second))
+    {
+      heat_source = viscoplastic_factor->evaluate_taylor_quinney_heat_source(context, gp, eleGID,
+          current_defgrad, Core::LinAlg::identity_matrix<3>(), current_temperature);
+    }
+  }
+  else if (num_contributions > 1)
+  {
+    FOUR_C_THROW("Mechanical Dissipation is only implemented for a single factor so far.");
```
dragos-ana:
```suggestion
    FOUR_C_THROW("Evaluation of additional heat source is only implemented for a single factor so far.");
```

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
```diff
@@ -467,6 +468,9 @@ namespace Mat
       //! isotropic constitutive tensor factors
       Core::LinAlg::Matrix<8, 1> curr_delta{Core::LinAlg::Initialization::zero};
 
+      //! thermal contribution due to thermal expansion to the thermo-elastic 2PK stress
+      TensorAndTemperatureDerivative curr_ST{};
```
dragos-ana, line 472:
```suggestion
      TensorAndTemperatureDerivative curr_ST_and_temp_deriv{};
```
Maybe?

tests/input_files/mat_iso_thermoviscoplast_refJC_log_timint_tsi_monolithic.4C.yaml
```diff
@@ -0,0 +1,346 @@
...
+  NEGLECTINERTIA: true
+  LINEAR_SOLVER: 1
+THERMAL DYNAMIC:
+  DYNAMICTYPE: OneStepTheta
+  RESULTSEVERY: 1
+  RESTARTEVERY: 1
+  INITIALFIELD: "field_by_function"
+  INITFUNCNO: 2
+  TIMESTEP: 0.05
+  NUMSTEP: 200
```
dragos-ana: TSI Dynamic governs the time settings, but it would be nice if they match the set values in both the dedicated structure and thermal sections.

tests/input_files/mat_iso_thermoviscoplast_refJC_log_timint_tsi_monolithic.4C.yaml
```diff
@@ -0,0 +1,346 @@
...
+  MERGE_TSI_BLOCK_MATRIX: true
+  TOLINC: 1e-08
+  CONVTOL: 1e-08
+  NORM_RESF: Abs
+SOLVER 1:
+  SOLVER: "UMFPACK"
+SOLVER 2:
+  SOLVER: Belos
+  SOLVER_XML_FILE: "xml/linear_solver/iterative_gmres_template.xml"
+  PRECONDITIONER_XML_FILE: "xml/block_preconditioner/thermo_solid.xml"
```
dragos-ana: If I see this right, Belos is not used here at all? Maybe it should be removed if that is not the case.

src/thermo/src/element/4C_thermo_ele_impl.cpp
```diff
@@ -2064,50 +2070,58 @@ void Discret::Elements::TemperImpl<distype>::nonlinear_dissipation_fint_tang(
...
+  // store possible pointers for specific material types for later use
+  std::shared_ptr<Mat::ThermoPlasticHyperElast> thermoplhyperelast;
+  std::shared_ptr<Mat::MultiplicativeSplitDefgradElastHyper>
+      multiplicative_split_defgrad_elast_hyper_ptr;
+
+  if (structmat->material_type() == Core::Materials::m_thermoplhyperelast)
   {
-    FOUR_C_THROW("So far dissipation only for ThermoPlasticHyperElast material!");
+    thermoplhyperelast = std::dynamic_pointer_cast<Mat::ThermoPlasticHyperElast>(structmat);
+    // true: error if cast fails
```
dragos-ana: Since the dynamic pointer cast does not throw by default, it could be helpful to explicitly assert that it is not a nullptr after the cast.

dragos-ana (commented):
Looks great!
Particularly, your use of tensors and the newly added `TensorAndTemperatureDerivative` have simplified the code a lot! 
I added some minor comments / suggestions

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.hpp
rjoussen: > This has nothing to do with your PR, but reinit is a very strange / non-descriptive name for what the function does.

I agree, especially because it is esclusively used to pass the thermo element's current temperature before calling `stress_temperature_modulus_and_deriv`. So I think this function should be removed entirely in the future.

> Maybe a matter for private discussion, but shouldn't we at least include the stress-temperature modulus into the residual and neglect the linearization first?

I tried that, but in my tests, it worsened global convergence a lot, while not having a big impact on the solution. There might be an option to include at least parts of the linearization, but then I think it's nicer to include it consistently directly.

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.hpp
rjoussen: I changed it to heat source. It specializes to Taylor-Quinney heating in the viscoplastic factor, where it actually is the Taylor-Quinney heat source. Before that, it could be any type (and maybe will be in the future)

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
rjoussen: I would argue that the type documents it sufficiently. That's exactly the reason why the type is useful in my opinion. With your name suggestion, we end up having expressions like `curr_ST_and_temp_deriv.temperature.derivative`, which is actually worse than just splitting the variable into two.

tests/input_files/mat_iso_thermoviscoplast_refJC_log_timint_tsi_monolithic.4C.yaml
rjoussen: true. I removed it

src/thermo/src/element/4C_thermo_ele_impl.cpp
rjoussen: I adapted it here and in a couple of other places.

tests/input_files/mat_iso_thermoviscoplast_refJC_log_timint_tsi_monolithic.4C.yaml
rjoussen: I deleted the timestep control parameters in both sections, as well as some other parameters that are obviously overwritten.

src/mat/4C_mat_inelastic_defgrad_factors_service.hpp
dragos-ana: Using your argument, one could argue that the temperature derivatives could simply be called `.deriv` inside of the struct, leading to the name `curr_ST_and_temp_deriv_.deriv`, which would arguably be more descriptive.

You can leave it as it is ;) It's just that, if I see `curr_ST` somewhere in the code, I would expect it to be ST and not also its temperature derivative, so I would need to go the definition of the struct to identify what it actually contains

rjoussen: We can revisit the naming once we use tensors more broadly, then also in the context of partial/total derivatives :)

dragos-ana (dismissed):

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.hpp
```diff
@@ -337,11 +459,13 @@ namespace Mat
      * @param[in] kinemat_quant struct containing kinematic quantities
      * @param[in] stress_fact struct containing \f$ \gamma_i \f$ and \f$ \delta_i \f$ stress
      * factors, as presented in Holzapfel - Nonlinear Solid Mechanics
+     * @param[in] SthetaT thermal part of the thermo-elastic stress (due to thermal expansion)
      * @return derivative \f$ \frac{\partial \mathsymbol{S}}{\partial
      * \mathsymbol{F}^{-1}_{\text{in}}} \f$
      */
-    Core::LinAlg::Matrix<6, 9> evaluated_sdi_fin(
-        const KinematicQuantities& kinemat_quant, const Mat::StressFactors& stress_factors) const;
+    [[nodiscard]] Core::LinAlg::Matrix<6, 9> evaluated_sdi_fin(
```
c-p-schmidt: As you are anyway touching this method that used to be renamed wrongly (back when naming conventions have been applied to the whole repo), you could also correct the naming here to e.g.:
```suggestion
    [[nodiscard]] Core::LinAlg::Matrix<6, 9> evaluate_ds_d_ifin(
```

c-p-schmidt (dismissed):

src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.hpp
rjoussen: In line with `evaluate_d_stress_d_scalar`, I renamed it to `evaluate_d_stress_d_ifin`.

c-p-schmidt (approved):

dragos-ana (approved):

PR #2084, opened 2026-06-16
Tensor migration 

tuchpaul:
## Migrate elastic right CG tensor derivatives to the tensor API

Part of the ongoing `FourTensorOperations` → tensor-API migration.

### Migrated
- `elast_hyper_get_derivs_of_elastic_right_cg_tensor` now returns full
  `Tensor<double, 3, 3, 3, 3>` for `dCedC` and `dCediFin`, computed directly with
  inline `einsum` expressions instead of the legacy `FourTensor` / 9x9-Voigt round-trip:
  - `dCedC = 0.5 * (einsum<"ik","jl">(iFinT, iFinT) + einsum<"il","jk">(iFinT, iFinT))`
  - `dCediFin = einsum<"ad","bc">(id, iFinTC) + einsum<"ac","bd">(iFinTC, id)`
- Callers convert the result to Voigt via the new helpers.

### Added
- Conversion helpers in `4C_linalg_tensor_conversion.hpp`:
  `make_6x9 / make_9x6 / make_6x6_voigt_matrix_from_tensor`, with unit tests.

### Removed
- The  6x9 `add_kronecker_tensor_product` overload — it had
  no callers.
- `contract_matrix_matrix`, replaced by `Core::LinAlg::ddot` at its call site.

### DIFF
--- a/src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.cpp
+++ b/src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.cpp
@@ -150,120 +150,6 @@ void Core::LinAlg::FourTensorOperations::add_kronecker_tensor_product(Core::LinA
       scalar_this * C(5, 5) + scalar_AB_half * (A(0, 0) * B(2, 2) + A(0, 2) * B(2, 0));  // C1313
 }
 
-
-void Core::LinAlg::FourTensorOperations::add_kronecker_tensor_product(Core::LinAlg::Matrix<6, 9>& C,
-    const double scalar_AB, const Core::LinAlg::Matrix<3, 3>& A,
-    const Core::LinAlg::Matrix<3, 3>& B, const double scalar_this)
-{
-  const double scalar_AB_half = scalar_AB * 0.5;
-  C(0, 0) = scalar_this * C(0, 0) + scalar_AB * (A(0, 0) * B(0, 0));  // C1111
-  C(0, 1) = scalar_this * C(0, 1) + scalar_AB * (A(0, 1) * B(0, 1));  // C1122
-  C(0, 2) = scalar_this * C(0, 2) + scalar_AB * (A(0, 2) * B(0, 2));  // C1133
-  C(0, 3) =
-      scalar_this * C(0, 3) + scalar_AB_half * (A(0, 0) * B(0, 1) + A(0, 1) * B(0, 0));  // C1112
-  C(0, 4) =
-      scalar_this * C(0, 4) + scalar_AB_half * (A(0, 1) * B(0, 2) + A(0, 2) * B(0, 1));  // C1123
-  C(0, 5) =
-      scalar_this * C(0, 5) + scalar_AB_half * (A(0, 0) * B(0, 2) + A(0, 2) * B(0, 0));  // C1113
-  C(0, 6) =
-      scalar_this * C(0, 6) + scalar_AB_half * (A(0, 1) * B(0, 0) + A(0, 0) * B(0, 1));  // C1121
-  C(0, 7) =
-      scalar_this * C(0, 7) + scalar_AB_half * (A(0, 2) * B(0, 1) + A(0, 1) * B(0, 2));  // C1132
-  C(0, 8) =
-      scalar_this * C(0, 8) + scalar_AB_half * (A(0, 2) * B(0, 0) + A(0, 0) * B(0, 2));  // C1131
-
-
-  C(1, 0) = scalar_this * C(1, 0) + scalar_AB * (A(1, 0) * B(1, 0));  // C2211
-  C(1, 1) = scalar_this * C(1, 1) + scalar_AB * (A(1, 1) * B(1, 1));  // C2222
-  C(1, 2) = scalar_this * C(1, 2) + scalar_AB * (A(1, 2) * B(1, 2));  // C2233
-  C(1, 3) =
-      scalar_this * C(1, 3) + scalar_AB_half * (A(1, 0) * B(1, 1) + A(1, 1) * B(1, 0));  // C2212
-  C(1, 4) =
-      scalar_this * C(1, 4) + scalar_AB_half * (A(1, 1) * B(1, 2) + A(1, 2) * B(1, 1));  // C2223
-  C(1, 5) =
-      scalar_this * C(1, 5) + scalar_AB_half * (A(1, 0) * B(1, 2) + A(1, 2) * B(1, 0));  // C2213
-  C(1, 6) =
-      scalar_this * C(1, 6) + scalar_AB_half * (A(1, 1) * B(1, 0) + A(1, 0) * B(1, 1));  // C2221
-  C(1, 7) =
-      scalar_this * C(1, 7) + scalar_AB_half * (A(1, 2) * B(1, 1) + A(1, 1) * B(1, 2));  // C2232
-  C(1, 8) =
-      scalar_this * C(1, 8) + scalar_AB_half * (A(1, 2) * B(1, 0) + A(1, 0) * B(1, 2));  // C2231
-
-
-
-  C(2, 0) = scalar_this * C(2, 0) + scalar_AB * (A(2, 0) * B(2, 0));  // C3311
-  C(2, 1) = scalar_this * C(2, 1) + scalar_AB * (A(2, 1) * B(2, 1));  // C3322
-  C(2, 2) = scalar_this * C(2, 2) + scalar_AB * (A(2, 2) * B(2, 2));  // C3333
-  C(2, 3) =
-      scalar_this * C(2, 3) + scalar_AB_half * (A(2, 0) * B(2, 1) + A(2, 1) * B(2, 0));  // C3312
-  C(2, 4) =
-      scalar_this * C(2, 4) + scalar_AB_half * (A(2, 1) * B(2, 2) + A(2, 2) * B(2, 1));  // C3323
-  C(2, 5) =
-      scalar_this * C(2, 5) + scalar_AB_half * (A(2, 0) * B(2, 2) + A(2, 2) * B(2, 0));  // C3313
-  C(2, 6) =
-      scalar_this * C(2, 6) + scalar_AB_half * (A(2, 0) * B(2, 1) + A(2, 1) * B(2, 0));  // C3321
-  C(2, 7) =
-      scalar_this * C(2, 7) + scalar_AB_half * (A(2, 2) * B(2, 1) + A(2, 1) * B(2, 2));  // C3332
-  C(2, 8) =
-      scalar_this * C(2, 8) + scalar_AB_half * (A(2, 2) * B(2, 0) + A(2, 0) * B(2, 2));  // C3331
-
-
-
-  C(3, 0) = scalar_this * C(3, 0) + scalar_AB * (A(0, 0) * B(1, 0));  // C1211
-  C(3, 1) = scalar_this * C(3, 1) + scalar_AB * (A(0, 1) * B(1, 1));  // C1222
-  C(3, 2) = scalar_this * C(3, 2) + scalar_AB * (A(0, 2) * B(1, 2));  // C1233
-  C(3, 3) =
-      scalar_this * C(3, 3) + scalar_AB_half * (A(0, 0) * B(1, 1) + A(0, 1) * B(1, 0));  // C1212
-  C(3, 4) =
-      scalar_this * C(3, 4) + scalar_AB_half * (A(0, 1) * B(1, 2) + A(0, 2) * B(1, 1));  // C1223
-  C(3, 5) =
-      scalar_this * C(3, 5) + scalar_AB_half * (A(0, 0) * B(1, 2) + A(0, 2) * B(1, 0));  // C1213
-  C(3, 6) =
-      scalar_this * C(3, 6) + scalar_AB_half * (A(0, 1) * B(1, 0) + A(0, 0) * B(1, 1));  // C1221
-  C(3, 7) =
-      scalar_this * C(3, 7) + scalar_AB_half * (A(0, 2) * B(1, 1) + A(0, 1) * B(1, 2));  // C1232
-  C(3, 8) =
-      scalar_this * C(3, 8) + scalar_AB_half * (A(0, 2) * B(1, 0) + A(0, 0) * B(1, 2));  // C1231
-
-
-
-  C(4, 0) = scalar_this * C(4, 0) + scalar_AB * (A(1, 0) * B(2, 0));  // C2311
-  C(4, 1) = scalar_this * C(4, 1) + scalar_AB * (A(1, 1) * B(2, 1));  // C2322
-  C(4, 2) = scalar_this * C(4, 2) + scalar_AB * (A(1, 2) * B(2, 2));  // C2333
-  C(4, 3) =
-      scalar_this * C(4, 3) + scalar_AB_half * (A(1, 0) * B(2, 1) + A(1, 1) * B(2, 0));  // C2312
-  C(4, 4) =
-      scalar_this * C(4, 4) + scalar_AB_half * (A(1, 1) * B(2, 2) + A(1, 2) * B(2, 1));  // C2323
-  C(4, 5) =
-      scalar_this * C(4, 5) + scalar_AB_half * (A(1, 0) * B(2, 2) + A(1, 2) * B(2, 0));  // C2313
-  C(4, 6) =
-      scalar_this * C(4, 6) + scalar_AB_half * (A(1, 1) * B(2, 0) + A(1, 0) * B(2, 1));  // C2321
-  C(4, 7) =
-      scalar_this * C(4, 7) + scalar_AB_half * (A(1, 2) * B(2, 1) + A(1, 1) * B(2, 2));  // C2332
-  C(4, 8) =
-      scalar_this * C(4, 8) + scalar_AB_half * (A(1, 2) * B(2, 0) + A(1, 0) * B(2, 2));  // C2331
-
-
-
-  C(5, 0) = scalar_this * C(5, 0) + scalar_AB * (A(0, 0) * B(2, 0));  // C1311
-  C(5, 1) = scalar_this * C(5, 1) + scalar_AB * (A(0, 1) * B(2, 1));  // C1322
-  C(5, 2) = scalar_this * C(5, 2) + scalar_AB * (A(0, 2) * B(2, 2));  // C1333
-  C(5, 3) =
-      scalar_this * C(5, 3) + scalar_AB_half * (A(0, 0) * B(2, 1) + A(0, 1) * B(2, 0));  // C1312
-  C(5, 4) =
-      scalar_this * C(5, 4) + scalar_AB_half * (A(0, 1) * B(2, 2) + A(0, 2) * B(2, 1));  // C1323
-  C(5, 5) =
-      scalar_this * C(5, 5) + scalar_AB_half * (A(0, 0) * B(2, 2) + A(0, 2) * B(2, 0));  // C1313
-  C(5, 5) =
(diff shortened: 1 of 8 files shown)


src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.cpp
```diff
@@ -254,14 +255,26 @@ void Core::LinAlg::FourTensorOperations::add_kronecker_tensor_product(Core::LinA
...
       scalar_this * C(5, 5) + scalar_AB_half * (A(0, 0) * B(2, 2) + A(0, 2) * B(2, 0));  // C1313
-  C(5, 5) =
-      scalar_this * C(5, 5) + scalar_AB_half * (A(0, 0) * B(2, 2) + A(0, 2) * B(2, 0));  // C1313
   C(5, 6) =
       scalar_this * C(5, 6) + scalar_AB_half * (A(0, 1) * B(2, 0) + A(0, 0) * B(2, 1));  // C1321
   C(5, 7) =
       scalar_this * C(5, 7) + scalar_AB_half * (A(0, 2) * B(2, 1) + A(0, 1) * B(2, 2));  // C1332
   C(5, 8) =
-      scalar_this * C(5, 7) + scalar_AB_half * (A(0, 2) * B(2, 0) + A(0, 0) * B(2, 2));  // C1331
+      scalar_this * C(5, 8) + scalar_AB_half * (A(0, 2) * B(2, 0) + A(0, 0) * B(2, 2));  // C1331
```
rjoussen: Nice catch!

src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.cpp
```diff
@@ -254,14 +255,26 @@ void Core::LinAlg::FourTensorOperations::add_kronecker_tensor_product(Core::LinA
...
+}
+
+template <typename T>
+Core::LinAlg::SymmetricTensor<T, 3, 3, 3, 3> Core::LinAlg::FourTensorOperations::kronecker_product(
+    const Core::LinAlg::SymmetricTensor<T, 3, 3>& A)
+{
+  return 0.5 * (einsum_sym<"ik", "jl">(A, A) + einsum_sym<"il", "jk">(A, A));
+}
```
rjoussen: This operation returns a symmetric tensor regardless of whether `A` is symmetric. We should probably overload this or use `SquareTensor` concept directly here. This would also allow sizes different than 3.

src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.cpp
rjoussen: This tensor product seems to be unused. It probably makes sense to remove this in favor of the new kronecker product functions

src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.cpp
```diff
@@ -682,6 +695,13 @@ void Core::LinAlg::FourTensorOperations::add_adbc_tensor_product(const double fa
   out(2, 2) += fac * A(2, 2) * B(2, 2);
 }
 
+template <typename T>
+Core::LinAlg::Tensor<T, 3, 3, 3, 3> Core::LinAlg::FourTensorOperations::adbc_tensor_product(
+    const Core::LinAlg::Tensor<T, 3, 3>& A, const Core::LinAlg::Tensor<T, 3, 3>& B)
+{
+  return einsum<"ad", "bc">(A, B);
+}
+
```
rjoussen:
```suggestion

```
Since this purely wraps einsum, which is clearer and easier to understand in my opinion, I would not implement this. Users can just use einsum directly

src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.cpp
```diff
@@ -777,29 +797,11 @@ void Core::LinAlg::FourTensorOperations::add_non_symmetric_product(double const&
...
-        {
-          for (int m = 0; m < dim; ++m)
-          {  // C^ijkl = A^ijkm * B_m^l
-            four_tensor_result(i, j, k, l) += four_tensor(i, j, k, m) * matrix(m, l);
-          }
-        }
-      }
-    }
-  }
+  return einsum<"ac", "db">(A, B);
```
rjoussen: Same here, it is much clearer to use einsum directly, I think.

src/core/linalg/src/dense/4C_linalg_tensor_conversion.hpp
rjoussen, line 1: These new conversion methods will help a lot in the future, thanks for implementing them!

src/mat/4C_mat_elasthyper_service.cpp
```diff
@@ -580,29 +580,22 @@ void Mat::elast_hyper_get_derivs_of_elastic_right_cg_tensor(
...
+  dCedC = Core::LinAlg::make_6x6_voigt_matrix_from_tensor(
+      Core::LinAlg::FourTensorOperations::kronecker_product(iFinT, iFinT));
+
+  // \frac{\partial C^e}{\partial F_{in}^{-1}} = Id o (F_in^{-T} C) + (F_in^{-T} C) o Id
+  const Core::LinAlg::Tensor<double, 3, 3> iFinTC =
+      Core::LinAlg::dot(iFinT, Core::LinAlg::get_full(CM));
+  const Core::LinAlg::Tensor<double, 3, 3, 3, 3> dCediFin_tensor =
+      Core::LinAlg::FourTensorOperations::adbc_tensor_product(id, iFinTC) +
+      Core::LinAlg::FourTensorOperations::non_symmetric_product(iFinTC, id);
+  dCediFin = Core::LinAlg::make_6x9_voigt_matrix_from_tensor(dCediFin_tensor);
```
rjoussen: I would do this instead:
```cpp
void Mat::elast_hyper_get_derivs_of_elastic_right_cg_tensor(
    const Core::LinAlg::Tensor<double, 3, 3>& iFinM,
    const Core::LinAlg::SymmetricTensor<double, 3, 3>& CM, Core::LinAlg::Tensor<double, 3, 3, 3, 3>& dCedC,
    Core::LinAlg::Tensor<double, 3, 3, 3, 3>& dCediFin)
{
  // F_in^{-T} and the 2nd order identity tensor
  const auto iFinT = Core::LinAlg::transpose(iFinM);
  const auto id =
     Core::LinAlg::TensorGenerators::identity<double, 3, 3>;

  // \frac{\partial C^e}{\partial C} = F_in^{-T} o F_in^{-T} (self box product -> minor-symmetric)
  dCedC = Core::LinAlg::FourTensorOperations::kronecker_product(iFinT, iFinT);

  // \frac{\partial C^e}{\partial F_{in}^{-1}} = Id o (F_in^{-T} C) + (F_in^{-T} C) o Id
  const Core::LinAlg::Tensor<double, 3, 3> iFinTC =
      Core::LinAlg::dot(iFinT, CM);
  dCediFin =
      einsum<"ad", "bc">(id, iFinTC) +
      einsum<"ac", "bd">(iFinTC, id);
}
```
for the following reasons:
1. It makes this function purely tensor api, not mixed. The callers should be responsible for converting to Matrix if the need it.
2. Use `einsum` directly instead of adbc/non_symmetric special definitions.

rjoussen (commented):
Thanks for tackling this!

src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.hpp
```diff
@@ -119,6 +120,51 @@ namespace Core::LinAlg::FourTensorOperations
...
+   *
+   * @tparam T          Scalar type (e.g. double, FAD)
+   * @param[in] A       2nd order tensor \f$A_{IJ}\f$
+   * @param[in] B       2nd order tensor \f$B_{IJ}\f$
+   * @return            4th order tensor \f$\frac{1}{2}(A_{IK}B_{JL} + A_{IL}B_{JK})\f$
+   */
+  template <typename T>
+  Core::LinAlg::Tensor<T, 3, 3, 3, 3> kronecker_product(
+      const Core::LinAlg::Tensor<T, 3, 3>& A, const Core::LinAlg::Tensor<T, 3, 3>& B);
+
```
rjoussen: This is really a very special kind of kronecker product. For a normal kronecker product, I would probably expect something like $C = A\otimes B$ such that $C_{ijkl}=A_{ij}B_{kl}$. In other places, we refer to this exact product as Holzapfel product, but it's already inconsistent in the existing code.

Mainly, I am not sure whether we should introduce tensor helper functions with unclear semantics (unlike `dot` or `ddot` which are easy to understand). In contrast to the old `Matrix` api, specialized tensor products are easy to do on the fly when you need them, and the resulting code is probably easier to understand than having to look up what exactly a Kronecker product means here.

tuchpaul: Yes, I totally agree. I found myself quite confused with the terminology and implementation of the kronecker product (which clashes sometimes).
Therefore, I am also happy with dropping the helper functions, as the einsum is actually self-documenting. But I'm also happy to hear other opinions, from e.g. @dragos-ana and @c-p-schmidt.

dragos-ana: I agree as well. One could easily introduce "his" / "her" own helper functions directly in the cpp where the tensor products are required. 

However, I also find the idea of having some products already available quite nice. Since the naming is confusing, maybe this is more explicitly addressed by naming them based on the indices involved, e.g. like in `add_adbc_tensor_product`. For the composed products such as the Holzapfel product, this gets slightly uglier, e.g., `add_half_adbc_half_acbd_product`.

src/core/linalg/src/dense/4C_linalg_tensor_conversion.hpp
```diff
@@ -226,6 +227,91 @@ namespace Core::LinAlg
     }
   }
 
+  /*!
+   * @brief Creates a 6x9 Voigt matrix from a 4th order tensor that has the minor symmetry in its
+   * first index pair (C_ijkl = C_jikl).
```
dragos-ana, line 232: Really nice, the entire tensor conversion part. The conversion to 9x9 is missing as far as I see, but this is easy to implement based on your nice implementation here, if this will be required in the future.

src/mat/4C_mat_inelastic_defgrad_factors.cpp
```diff
@@ -2031,8 +2031,8 @@ Mat::InelasticDefgradTransvIsotropElastViscoplast::evaluate_state_quantities(
   // determine scalar quantities of invariants / pseudoinvariants needed to compute the
   // equivalent tensile stress
   double Mtheta_dev_sym_contract_Mtheta_dev_sym =
-      Core::LinAlg::FourTensorOperations::contract_matrix_matrix(
-          state_quantities.curr_Mtheta_dev_sym_M, state_quantities.curr_Mtheta_dev_sym_M);
+      Core::LinAlg::ddot(Core::LinAlg::make_tensor_view(state_quantities.curr_Mtheta_dev_sym_M),
```
dragos-ana, line 1993: Nice start to the tensor migration within the viscoplastic material ;)

src/core/linalg/tests/4C_linalg_tensor_conversion_test.cpp
```diff
@@ -314,6 +315,79 @@ namespace
...
+          }
+
+    const Core::LinAlg::Matrix<9, 6> new_voigt = Core::LinAlg::make_9x6_voigt_matrix_from_tensor(T);
+    Core::LinAlg::Matrix<9, 6> old_voigt(Core::LinAlg::Initialization::zero);
+    Core::LinAlg::Voigt::setup_9x6_voigt_matrix_from_four_tensor(old_voigt, ft);
+
+    FOUR_C_EXPECT_NEAR(new_voigt, old_voigt, 1e-12);
+  }
+
+  TEST(TensorConversionTest, MakeStressLikeVoigtMatrixFromTensorTest)
```
dragos-ana:
```suggestion
  TEST(TensorConversionTest, Make6x6StressLikeVoigtMatrixFromTensorTest)
```
Maybe slightly more descriptive and consistent with the other added tests?

dragos-ana (commented):
Looks very nice! I added some minor suggestions. 
I am not entirely sure what is the best approach now regarding the added tensor product counterparts for some of the matrix products such as the Kronecker product. 
The idea of having helper functions for these is nice, but this should be done with consistent, easily understandable names.

src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.hpp
rjoussen: > `add_adbc_tensor_product`. For the composed products such as the Holzapfel product, this gets slightly uglier, e.g., `add_half_adbc_half_acbd_product`.

Would such functions be used though? All these specialized tensor products are already used only a handful of times across the code. I think the main reason why some specialized tensor functions are useful is because they happen to appear often within certain contexts, like the `half_adbc_half_acbd_product` in the stiffness tensor computation. But then such a specialized product could be defined in elasthyper if necessary, maybe with a name that allows usage without knowing the exact index order. Whenever someone knows they have to use `half_adbc_half_acbd_product`, they apparently know the indices already and can use einsum in my opinion.

c-p-schmidt: I also agree that with tensors, we can get rid of many of these helper methods, as they are easier to use and read in place and less error-prone than the older matrix versions.

And if helpers make sense in a certain context, everyone can provide such helpers locally for relevant material implementations.

tuchpaul: As discussed I removed the additional helper functions. `elast_hyper_get_derivs_of_elastic_right_cg_tensor` now takes `dCedC` and `dCediFin` as `Core::LinAlg::Tensor`. 

Die relevant tensor operations are done inline.

src/core/linalg/src/dense/4C_linalg_fixedsizematrix_tensor_products.hpp
```diff
@@ -21,6 +21,8 @@
 #include "4C_linalg_four_tensor.hpp"
 #include "4C_linalg_symmetric_tensor.hpp"
 #include "4C_linalg_tensor.hpp"
+#include "4C_linalg_tensor_einstein.hpp"
+#include "4C_linalg_tensor_symmetric_einstein.hpp"
```
rjoussen: no longer needed

src/mat/4C_mat_elasthyper_service.cpp
```diff
@@ -577,32 +578,25 @@ void Mat::elast_hyper_check_polyconvexity(const Core::LinAlg::Tensor<double, 3,
...
-  const Core::LinAlg::Matrix<3, 3> C_mat = Core::LinAlg::make_matrix(Core::LinAlg::get_full(CM));
-  // auxiliaries
-  Core::LinAlg::Matrix<3, 3> id3x3(Core::LinAlg::Initialization::zero);
-  for (int i = 0; i < 3; ++i) id3x3(i, i) = 1.0;
-  Core::LinAlg::Matrix<9, 9> temp9x9(Core::LinAlg::Initialization::zero);
-  Core::LinAlg::FourTensor<3> tempFourTensor(true);
+  // F_in^{-T} and the 2nd order identity tensor
+  const Core::LinAlg::Tensor<double, 3, 3> iFinT = Core::LinAlg::transpose(iFinM);
+  const Core::LinAlg::Tensor<double, 3, 3> id =
+      Core::LinAlg::get_full(Core::LinAlg::TensorGenerators::identity<double, 3, 3>);
```
rjoussen:
```suggestion
  const Core::LinAlg::Tensor<double, 3, 3> id = Core::LinAlg::TensorGenerators::identity<double, 3, 3>;
```
I don't think we need the `get_full` here.

src/mat/4C_mat_elasthyper_service.cpp
```diff
@@ -577,32 +578,25 @@ void Mat::elast_hyper_check_polyconvexity(const Core::LinAlg::Tensor<double, 3,
...
   // \frac{\partial C^e}{\partial C}
-  dCedC.clear();
-  Core::LinAlg::Matrix<3, 3> iFinTM(Core::LinAlg::Initialization::zero);
-  iFinTM.multiply_nt(1.0, id3x3, iFinM_mat, 0.0);
-  Core::LinAlg::FourTensorOperations::add_kronecker_tensor_product(dCedC, 1.0, iFinTM, iFinTM, 0.0);
+  dCedC = 0.5 * (Core::LinAlg::einsum<"ik", "jl">(iFinT, iFinT) +
+                    Core::LinAlg::einsum<"il", "jk">(iFinT, iFinT));
+
+  const Core::LinAlg::Tensor<double, 3, 3> iFinTC =
+      Core::LinAlg::dot(iFinT, Core::LinAlg::get_full(CM));
```
rjoussen:
```suggestion
      Core::LinAlg::dot(iFinT, CM);
```
same here.

src/mat/4C_mat_elasthyper_service.cpp
```diff
@@ -577,32 +578,25 @@ void Mat::elast_hyper_check_polyconvexity(const Core::LinAlg::Tensor<double, 3,
...
+  const Core::LinAlg::Tensor<double, 3, 3> id =
+      Core::LinAlg::get_full(Core::LinAlg::TensorGenerators::identity<double, 3, 3>);
 
   // \frac{\partial C^e}{\partial C}
-  dCedC.clear();
-  Core::LinAlg::Matrix<3, 3> iFinTM(Core::LinAlg::Initialization::zero);
-  iFinTM.multiply_nt(1.0, id3x3, iFinM_mat, 0.0);
-  Core::LinAlg::FourTensorOperations::add_kronecker_tensor_product(dCedC, 1.0, iFinTM, iFinTM, 0.0);
+  dCedC = 0.5 * (Core::LinAlg::einsum<"ik", "jl">(iFinT, iFinT) +
+                    Core::LinAlg::einsum<"il", "jk">(iFinT, iFinT));
```
rjoussen:
```suggestion
  dCedC = 0.5 * assume_symmetry(Core::LinAlg::einsum<"ik", "jl">(iFinT, iFinT) +
                    Core::LinAlg::einsum<"il", "jk">(iFinT, iFinT));
```
should we assume symmetry here? Technically, dCedC is not necessarily symmetric, but we are only interested in the symmetric part since we contract it with a symmetric tensor anyway. The way we compute it here, we already only take the symmetric part, so returning a symmetric tensor would be more honest maybe. Returning a full fourth order tensor could indicate that we are returning something like $$\frac{\partial C^e_{il}}{\partial C_{mn}} = F^{-1}_{\text{in},mi} F^{-1}_{\text{in},nl}$$

src/core/linalg/src/dense/4C_linalg_tensor_conversion.hpp
```diff
@@ -226,6 +227,91 @@ namespace Core::LinAlg
...
+   * @brief Creates a 6x6 stress-like Voigt matrix from a full 4th order tensor by symmetrizing
+   * both index pairs independently.
+   *
+   * Both the first (IJ) and second (KL) index pairs are symmetrized and stored in stress-like
+   * 6-Voigt notation. The result may be a non-symmetric 6x6 matrix if the tensor lacks major
+   * symmetry (C_IJKL != C_KLIJ).
+   */
+  template <typename T>
+  Core::LinAlg::Matrix<6, 6, T> make_6x6_voigt_matrix_from_tensor(
+      const Core::LinAlg::Tensor<T, 3, 3, 3, 3>& tensor)
```
rjoussen:
```suggestion
      const Core::LinAlg::SymmetricTensor<T, 3, 3, 3, 3>& tensor)
```
maybe like this because this should only be done for symmetric tensors?

rjoussen (approved):
Thank you!

c-p-schmidt (approved):

PR #2086, opened 2026-06-16
Correct invalid names in journal files

### LINKED ISSUE #1973: Invalid names in exodus journal files
This might be rather niche, but I noticed that the following tutorial journal files:

- `tutorial_fsi_3d.jou`
- `tutorial_fsi_2d.jou`
- `tutorial_battery.jou`

use cubit names (via `nodeset <ID> name "<name>"` or `block <ID> name "<name>"`) that have spaces in them. Apparently, those names are considered invalid by cubit, and the spaces are silently replaced with underscores `_`. See [https://coreform.com/cubit_help/cubithelp.htm#t=geometry%2Fattributes%2Fentity_names.htm](https://coreform.com/cubit_help/cubithelp.htm#t=geometry%2Fattributes%2Fentity_names.htm)

This is a problem when we want to refer to node sets (or element blocks) with their names instead of their ids. 

I could just change the names in the journal files, but perhaps it is safer if someone who actually has a Cubit license runs the new journal files again.

c-p-schmidt:
## Description and Context
As reported in #1973, we have journal files with invalid names in the repo. This issue is addressed by changing the names in the journal files to valid names (in Coreform Cubit terms) and assigning the conditions using this name.

While changing this, I realized that we can only have 32-character names, as the Exodus API for reading the mesh truncates the name beyond 32 characters, which resulted in 4C not finding the corresponding nodeset. Thus, the first commit adds a check to ensure that the name is not longer than 32 characters. The others are changing the affected tutorials test by test.

Closes #1973


c-p-schmidt:
As Copilot noted. Along the way, I also adapted the battery tutorial a bit. ;-)

tests/tutorials/battery/tutorial_battery.4C.yaml
```diff
@@ -253,198 +277,206 @@ FUNCT6:
...
 
 DESIGN VOL TRANSPORT DIRICH CONDITIONS:
-  - E: 2
-    ENTITY_TYPE: element_block_id
-    NUMDOF: 2
-    ONOFF: [1, 0]
-    VAL: [1200, 0]
-    FUNCT: [0, 0]
   - E: 4
     ENTITY_TYPE: element_block_id
```
rjoussen, line 474: We could consider changing to name-based block definition as well. Especially for a tutorial that's nicer imo. But it's your tutorial, so your decision :)

rjoussen (dismissed):
Changes look good to me! I left one suggestion.

tests/tutorials/battery/tutorial_battery.4C.yaml
c-p-schmidt: I tried doing this as well but it didn't work. Maybe because they are not explicitly set as a nodeset? Or maybe I made another mistake? I can double check later.

tests/tutorials/battery/tutorial_battery.4C.yaml
rjoussen, line 686: I mixed it up. Here, it should definitely be possible to use the block names.

tests/tutorials/battery/tutorial_battery.4C.yaml
rjoussen: As I wrote in my other comment, I thought these were the block definitions, but they are not. Yes, currently, boundary applied to element blocks can't be identified by name, I didn't implement that. Mainly because I did not really see the need to not always use nodesets to apply boundary conditions in the first place.

tests/tutorials/battery/tutorial_battery.4C.yaml
c-p-schmidt: Nice, thank you for the hint. I just changed it.

src/core/fem/src/condition/4C_fem_condition_definition.cpp
```diff
@@ -86,6 +85,13 @@ void Core::Conditions::ConditionDefinition::read(
   {
     auto parsed_condition_data = read_condition_data(condition_data);
 
+    if (parsed_condition_data.node_set_name.has_value())
+    {
+      FOUR_C_ASSERT_ALWAYS(parsed_condition_data.node_set_name.value().size() <= 32,
+          "NODE_SET_NAME '{}' exceeds the maximum length of 32 characters.",
+          parsed_condition_data.node_set_name.value());
+    }
+
```
rjoussen: If I understand this correctly, this now always throws an error if there is a condition definition in the input with a name longer than 32 characters. Isn't this overly restrictive? Gmsh and vtu meshes have no problem with reading meshes with long node set names. Maybe we shouldn't check that here, but in the Exodus reader itself, when a nodeset was read that has exactly 32 characters? Or, we only display a note about this problem when a nodeset has not been found by name, and it has more than 32 characters.

c-p-schmidt: I also thought about that, but the problem is that I do not have the relevant information in the Exodus reader. One could only use the name length provided by the Exodus method as a proxy. If it has fewer than 32 characters, it was ok; if it has 32 characters, it might have been exactly the right length or too long. Which also felt like an unsatisfying solution to me.

However, you are right that this may be too strict for other mesh readers (I do not know). I just thought that 32 characters might be long enough anyway. But if you have doubts, we could only output a warning stating that for the Exodus reader, 32 characters is the maximum, and then the code would terminate later anyway, as the condition is not found in the mesh.

rjoussen: > we could only output a warning stating that for the Exodus reader, 32 characters is the maximum, and then the code would terminate later anyway, as the condition is not found in the mesh.

I would be in favor of that. This would only look a little strange if someone defines a name in exodus with exactly 32 characters and already knows about the limitation. In almost all scenarios, the user would see a warning about the length first and a crash immediately afterwards.

c-p-schmidt: It is now replaced by two warning outputs, one for the nodeset name length and one for the element block name length, only in the Exodus reader.

rjoussen (approved):

src/core/fem/src/condition/4C_fem_condition_definition.cpp
rjoussen: perfect, thanks

lkoeglmeier (approved):

PR #2088, opened 2026-06-17
feat: beaminteraction -> potential: add potential reduction functions

davidrudlstorfer:
## Description and Context

Add different potential reduction functions, i.e., polynomials of 5th and 7th order instead of the regular cosine potential reduction function

Follows #1957 
Two more small extensions will follow ASAP


src/beaminteraction/src/potential/4C_beaminteraction_potential_input.cpp
```diff
@@ -85,6 +85,13 @@ Core::IO::InputSpec BeamInteraction::Potential::valid_parameters()
                   .validator = null_or(positive<double>()),
                   .store = in_struct(&BeamPotentialParameters::potential_reduction_length)}),
 
+          parameter<ReductionFunction>("potential_reduction_function",
```
isteinbrecher, line 88: Does this only affect end point potential interactions? Maybe add “end_point” to indicate this, `end_point_potential_reduction_function`.

isteinbrecher: Side question, are any combinations between end point interactions and line to line interactions possible, or do they have to fulfill some consistency requirements?

isteinbrecher (approved):

src/beaminteraction/src/potential/4C_beaminteraction_potential_input.cpp
davidrudlstorfer: Thanks for the review:)

> Does this only affect end point potential interactions? Maybe add “end_point” to indicate this, end_point_potential_reduction_function.

This is only in the potential reduction region where this potential reduction function is active. I would keep the parametername like proposed because in the input file there is the 

`potential_reduction_length`

and now the 

`potential_reduction_function`

and with the next PRs new variables will follow which follow the same naming scheme. If then the naming scheme is consistent, the parameters will be sorted next to each other in the input files.

> Side question, are any combinations between end point interactions and line to line interactions possible, or do they have to fulfill some consistency requirements?

There are no specific end point interactions but only the beam to beam (line to line) interactions. The only problem is that with the SBIP approach (from Max Grill) there are end point effects missing.
With the already introduced endpoint augmentation approach these effects are no longer missing and we get an axial pull-off force.

So summing up, these effects are not standalone but extend the regular approach with a specific handling of end points.

isteinbrecher: Thanks for the explanation, looks good to me!

src/beaminteraction/src/potential/4C_beaminteraction_potential_input.hpp
```diff
@@ -77,6 +77,14 @@ namespace BeamInteraction::Potential
     bool write_uids{};
   };
 
+  /// function for potential reduction factor
+  enum class ReductionFunction
+  {
+    cosine,
+    polynomial_5,
+    polynomial_7
```
knarfnitram, line 85: Would it be better to introduce here a global function from the Input file specified with an ID and evaluate it with the derivatives accordingly? 
This would give you/other users the flexibility to test fast,  different potential reduction functions specified from the input file, without recompiling.

davidrudlstorfer: Here we need a (at least C1 continuous) function mapping from the beam end point till the potential reduction length from 0 to 1 (basically a smoothstep function).

In the method I came up with a few more functions (for example the sigmoid function does not fulfil all requirements) but they are not really usable in this approach. In the paper also these three functions will be discussed in more detail so I would keep the hardcoded functions in this case.

Further, the two polynomial are a rather "bad" choice but I wanna introduce the code into 4C because it is also discussed in the upcoming paper.

knarfnitram (approved):
Apart from that one suggestion, the changes looks good to me!!

src/beaminteraction/src/potential/4C_beaminteraction_potential_pair_beam_to_beam.cpp
```diff
@@ -3410,27 +3411,63 @@ bool BeamInteraction::BeamToBeamPotentialPair<numnodes, numnodalvalues,
...
+        pot_red_fac_2ndderiv_l_edge =
+            0.5 * std::numbers::pi * std::numbers::pi /
+            (potential_reduction_length.value() * potential_reduction_length.value()) *
+            std::cos(std::numbers::pi * length_to_edge / potential_reduction_length.value());
+      }
+      else if (params()->potential_reduction_function ==
+               BeamInteraction::Potential::ReductionFunction::polynomial_5)
+      {
+        Core::FE::Polynomial polynomial =
+            Core::FE::Polynomial({0, 0, 0, 10 / std::pow(potential_reduction_length.value(), 3),
```
isteinbrecher, line 3441: Out of interest, where do these factors 10, -15, 6 come from?

davidrudlstorfer: That's just the factors of the corresponding smoothstep polynomial (5th and 7th order)

For example here they are listed

https://en.wikipedia.org/wiki/Smoothstep#5th-order_equation

https://en.wikipedia.org/wiki/Smoothstep#7th-order_equation

src/beaminteraction/src/potential/4C_beaminteraction_potential_input.hpp
isteinbrecher: Thanks for the explanation, this also makes sense regarding my initial question on consistency.

PR #2092, opened 2026-06-21
Flush error output before MPI abort in main

rjoussen:
## Description and Context

Currently, we write the error message to `std::cout` and call `MPI_Abort` directly afterward. Sometimes, this leads to mpi abort before the error message makes it to the standard output. I had problems with that when running 4C with a python subprocess. The error message got lost in these cases. This now reliably flushes the messages to `std::cerr` before aborting.

### DIFF
--- a/apps/global_full/4C_global_full_main.cpp
+++ b/apps/global_full/4C_global_full_main.cpp
@@ -181,8 +181,10 @@ int main(int argc, char* argv[])
     }
     catch (Core::Exception& err)
     {
-      char line[] = "=========================================================================\n";
-      std::cout << "\n\n" << line << err.what_with_stacktrace() << "\n" << line << "\n" << '\n';
+      constexpr std::string_view line =
+          "\n=========================================================================\n";
+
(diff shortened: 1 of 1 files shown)


lauraengelhardt (dismissed):

apps/global_full/4C_global_full_main.cpp
```diff
@@ -181,8 +181,10 @@ int main(int argc, char* argv[])
     }
     catch (Core::Exception& err)
     {
-      char line[] = "=========================================================================\n";
-      std::cout << "\n\n" << line << err.what_with_stacktrace() << "\n" << line << "\n" << '\n';
+      std::string line =
```
tuchpaul:
```suggestion
      constexpr std::string_view line =
```

tuchpaul: Maybe we can use `constexpr std::string_view` here. In case the error is triggered by OOM, we then don't need to allocate new memory

tuchpaul (approved):

lauraengelhardt (approved):

PR #2094, opened 2026-06-22
Fix circular references in anisotropy framework

lauraengelhardt:
## Description and Context
First, the anisotropy extension had a circular reference problem: every `Anisotropy` object stored `shared_ptr<BaseAnisotropyExtension>` in its `extensions_` vector, and each extension stored a `shared_ptr<Anisotropy>` back-reference to the `Anisotropy`. I removed the anisotropy back-reference from extensions by removing the `shared_ptr<Anisotropy> anisotropy_` member from `BaseAnisotropyExtension` and the `set_anisotropy()` / `get_anisotropy()` accessors (first commit).

Second, we now store the anisotropy by value in all materials (just one change needed in the `MultiplicativeSplitDefgradElastHyper`) which is more safe (second commit).

Third, the `extensions_` vector was never cleared before re-registering ansiotropy extensions in the unpacking, so extensions that were recreated during unpack could still hold dangling pointers to the old (pre-unpack) Anisotropy object. A pragmatic solution to fix this is to clear the `extensions_` vector in the anisotropy unpacking (forth commit). 

## Disclaimer
I first want to see if all test pass (so this is still a draft) and am not entirely sure if this is the most optimal approach. The entire anisotropy framework would need a broader rework IMO, but this could be a step in the right direction. I would much appreciate any comments/improvements! 

While the code changes were done by myself I used opencode to actually figure out the circular reference problem.


src/mat/4C_mat_anisotropy_extension.hpp
```diff
@@ -266,6 +268,9 @@ namespace Mat
      */
     void compute_structural_tensors_stress();
 
+    /// Number of Gauss points supplied by the Anisotropy class during setup
+    int numgp_ = 0;
```
c-p-schmidt: Do you maybe want to initialize to something obviously wrong?
```suggestion
    int numgp_ = -1;
```

src/mat/4C_mat_anisotropy_extension_cylinder_cosy.cpp
```diff
@@ -68,10 +81,10 @@ Mat::CylinderCoordinateSystemAnisotropyExtension::get_cylinder_coordinate_system
 
   if (cosy_location_ == CosyLocation::ElementCosy)
   {
-    return get_anisotropy()->get_element_cylinder_coordinate_system();
+    return *element_cosy_;
```
c-p-schmidt, line 89: Do you need to make sure that this optional is actually set? Or is that already ensured elsewhere?

lauraengelhardt: We would just return the value before, so I assume there is a check before. This entire class is a mess, with the element_cosy_ and the gp_cosy_ and then the two flags... But I can add an assert for safety.

src/mat/4C_mat_anisotropy_extension.hpp
lauraengelhardt: Let's see if the tests pass :)

src/mat/4C_mat_anisotropy_extension_cylinder_cosy.cpp
```diff
@@ -24,36 +24,49 @@ void Mat::CylinderCoordinateSystemAnisotropyExtension::pack_anisotropy(
...
-void Mat::CylinderCoordinateSystemAnisotropyExtension::on_global_data_initialized()
+void Mat::CylinderCoordinateSystemAnisotropyExtension::on_global_data_initialized(
+    Mat::Anisotropy& anisotropy)
 {
-  if (get_anisotropy()->has_gp_cylinder_coordinate_system())
+  if (anisotropy.has_gp_cylinder_coordinate_system())
   {
     cosy_location_ = CosyLocation::GPCosy;
+    gp_cosy_.clear();
+    for (int gp = 0; gp < anisotropy.get_number_of_gauss_points(); ++gp)
```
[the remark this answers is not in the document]
lauraengelhardt: Was bad before and is bad now. Could be improved by reworking the cylinder cosy extension to only store one coordinate system (element OR gp). Out of scope here.

bgoderbauer (approved):

c-p-schmidt (approved):

PR #2095, opened 2026-06-22
Add 3D solid superposition material

lauraengelhardt:
## Description and Context
This PR implements a new material `MAT::SolidSuperposition` that superposes the stress and tangent contributions of multiple 3D solid materials. Each constituent material is evaluated independently and their responses are accumulated:

$$
\mathbf{S} = \sum_i \mathbf{S}_i, \qquad
\mathbb{C} = \sum_i \mathbb{C}_i
$$

`setup`, `pack`, `unpack`, etc. simply call the methods of the to be superposed materials in a loop. 

This is already possible for all materials within the `ElastHyper` framework and similarly for all materials within the `Mixture` framework but not for any solid materials that fall out of these two categories. 

**Background**: I specifically need it to add additional stress contributions to my muscle materials. I can currently hack this using the Mixture framework by choosing the same massfractions for all constituents (e.g. [0.5, 0.5] if two constituents] and then modifying the parameters that scale stresses, but this is extremely annoying/error prone and doesn't really reflect this simple stress addition. 

**Tests**
1. St. Venant Kirchhoff superposition: two SVK materials with Young's moduli 200 and 600; result matches a single SVK with modulus 800.
2. ElastHyper superposition: superposition of anisotropic CoupAnisoExpo + CoupNeoHooke pairs; equivalent to a single material with summed parameters.
3. Mixture + ElastHyper superposition: same as (2) but one constituent is wrapped in MAT_Mixture (single constituent, mass fraction 1.0) to verify that mixture framework works correctly within superposition.

**AI disclaimer**
opencode used for improving test descriptions

This PR depends on #2094. The only relevant commit for the PR is the last.


src/mat/4C_mat_solid_superposition.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+    [[nodiscard]] Core::Materials::MaterialType material_type() const override
+    {
+      return Core::Materials::m_superposition;
+    }
+
+    [[nodiscard]] std::shared_ptr<Core::Mat::Material> clone() const override
+    {
+      return std::make_shared<SolidSuperposition>(*this);
+    }
```
bgoderbauer, line 118: You create a new material with the same `std::vector<std::shared_ptr<Mat::So3Material>> materials_;`. So, both the old and the new superposition material point to the same material objects. Is this the desired scenario? It feels a little suspicious.

tests/input_files/mat_solid_superposition.4C.yaml
```diff
@@ -0,0 +1,185 @@
...
+    and 600 (summing to 800).
+  - The results match a single St. Venant Kirchhoff material with Young's modulus 800, verifying linear
+    superposition of stresses and tangents.
+PROBLEM TYPE:
+  PROBLEMTYPE: "Structure"
+IO:
+  STRUCT_STRESS: "2PK"
+  STRUCT_STRAIN: "GL"
+IO/RUNTIME VTK OUTPUT:
+  INTERVAL_STEPS: 1
```
bgoderbauer, line 16: I'm not so familiar with the output and result description. Do we need runtime VTK output here?

src/mat/4C_mat_solid_superposition.cpp
```diff
@@ -0,0 +1,187 @@
...
+    Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3>& cmat, int gp, int eleGID)
+{
+  stress = {};
+  cmat = {};
+
+  // evaluate stress and material tangent for all materials
+  for (const auto& m : materials_)
+  {
+    Core::LinAlg::SymmetricTensor<double, 3, 3> s_tmp{};
+    Core::LinAlg::SymmetricTensor<double, 3, 3, 3, 3> c_tmp{};
```
bgoderbauer: Is it possible to put allocation out of the loop over the materials and just reset entries in each iteration?

bgoderbauer (commented):
Looks very good! Definitely easier to understand than `ElastHyper`. Do you think it makes sense to unify them in the future? At least I would have thought that the functionality and application case of `SuperPosition` should be something that `ElastHyper` should be able to represent from its description (not the actual existing implementation, though).

lauraengelhardt:
Thanks for the review! :) 

> Do you think it makes sense to unify them in the future?
Short answer: Not sure.

Long answer: `ElastHyper` implements methods _specifically_ for hyperelastic constitutive laws and can, for example, compute stress and material tangent responses based on coefficients defined with respect to modified principal stretches. So you can implement your material by just implementing these coefficients and the framework will handle the rest (e.g. computing the derivatives of the principal stretches with respect to C; computing the stresses from these quantities; handling volumetric–isochoric split if specified, a.s.o.). This does not hold for all materials. The addition of stresses/cmat could maybe be unified...

> At least I would have thought that the functionality and application case of `SuperPosition` should be something that `ElastHyper` should be able to represent from its description (not the actual existing implementation, though).

`ElastHyper` cannot do what `Superposition` does as `ElastHyper` can only add any of its summands - not any arbitrary 3D solid materials.

tests/input_files/mat_solid_superposition.4C.yaml
lauraengelhardt: I'll remove for two of the three tests, and just keep it for one. I know I don't test against the vtk results (which is not really needed as we have vtk output tests for the solid framework), but would like to ensure the `evaluate_output_data` doesn't crash.

src/mat/4C_mat_solid_superposition.hpp
lauraengelhardt: We do this for all materials, but I do not know why. I can have a look in the next weeks.

bgoderbauer: I guess we rarely copy materials after creating them anyway

c-p-schmidt: I guess you are right. I just checked the coverage report for the multiplicative split material which is used rather often and this method has 0 calls:
https://4c-multiphysics.github.io/4C/coverage_report/coverage/__w/4C/4C/src/mat/4C_mat_multiplicative_split_defgrad_elasthyper.hpp.html#L260

lauraengelhardt: OK nice, maybe we can just remove it in a follow up then.

src/mat/elast/4C_mat_elast_coupanisoexposhear.cpp
```diff
@@ -69,12 +69,14 @@ Mat::Elastic::CoupAnisoExpoShearAnisotropyExtension::get_structural_tensor(int g
   return structural_tensors_[gp];
 }
 
-void Mat::Elastic::CoupAnisoExpoShearAnisotropyExtension::on_global_data_initialized()
+void Mat::Elastic::CoupAnisoExpoShearAnisotropyExtension::on_global_data_initialized(
+    Mat::Anisotropy& anisotropy)
 {
```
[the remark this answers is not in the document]
lauraengelhardt: See https://github.com/4C-multiphysics/4C/pull/2094

src/mat/4C_mat_anisotropy_extension_cylinder_cosy.cpp
```diff
@@ -24,36 +24,49 @@ void Mat::CylinderCoordinateSystemAnisotropyExtension::pack_anisotropy(
...
   else
   {
     cosy_location_ = CosyLocation::None;
   }
 }
 
-void Mat::CylinderCoordinateSystemAnisotropyExtension::on_global_element_data_initialized()
+void Mat::CylinderCoordinateSystemAnisotropyExtension::on_global_element_data_initialized(
+    Mat::Anisotropy& anisotropy)
 {
```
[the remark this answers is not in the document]
lauraengelhardt: See https://github.com/4C-multiphysics/4C/pull/2094

src/mat/4C_mat_solid_superposition.hpp
lauraengelhardt: We call it here https://github.com/4C-multiphysics/4C/blob/a532b5781f0aaa972ac822ba4114c4c0e16ba43f/src/core/fem/src/general/element/4C_fem_general_element.cpp#L149

So I suppose its not so easy to remove.
The coverage report says the branch is True: 5.38%, False: 94.62%.

bgoderbauer (dismissed):

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1293,6 +1293,23 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+    known_materials[Core::Materials::m_superposition] = group("MAT_Superposition",
+        {
+            parameter<std::vector<int>>("MATIDS",
+                {
+                    .description = "List of material IDs to be superimposed",
+                    .validator = all_elements(positive<int>()),
+                }),
+            parameter<double>("DENS", {.description = "mass density of superposition material"}),
+        },
+        {.description = "3D solid material superposition"});
```
c-p-schmidt: Maybe it would be good to expand the description to include what you wrote in the PR description.

3D solid material superposition. Each constituent material defined by MATIDS is evaluated independently, and its responses are accumulated: \mathbf{S} = \sum_i \mathbf{S}_i, \mathbb{C} = \sum_i \mathbb{C}_i

tests/input_files/mat_solid_superposition.4C.yaml
```diff
@@ -0,0 +1,185 @@
...
+  SOLVER: "UMFPACK"
+DESIGN SURF DIRICH CONDITIONS:
+  - E: 1
+    NUMDOF: 3
+    ONOFF: [1, 1, 1]
+    VAL: [0, 0, 0]
+    FUNCT: [0, 0, 0]
+DESIGN SURF NEUMANN CONDITIONS:
+  - E: 2
+    NUMDOF: 6
```
c-p-schmidt: Why 6 dofs? Isn't that a standard solid test case?

tests/input_files/mat_solid_superposition_elasthyper_anisotropic.4C.yaml
```diff
@@ -0,0 +1,213 @@
...
+  SOLVER: "UMFPACK"
+DESIGN SURF DIRICH CONDITIONS:
+  - E: 1
+    NUMDOF: 3
+    ONOFF: [1, 1, 1]
+    VAL: [0, 0, 0]
+    FUNCT: [0, 0, 0]
+DESIGN SURF NEUMANN CONDITIONS:
+  - E: 2
+    NUMDOF: 6
```
c-p-schmidt: Also here

tests/input_files/mat_solid_superposition_mixture_anisotropic.4C.yaml
```diff
@@ -0,0 +1,225 @@
...
+  SOLVER: "UMFPACK"
+DESIGN SURF DIRICH CONDITIONS:
+  - E: 1
+    NUMDOF: 3
+    ONOFF: [1, 1, 1]
+    VAL: [0, 0, 0]
+    FUNCT: [0, 0, 0]
+DESIGN SURF NEUMANN CONDITIONS:
+  - E: 2
+    NUMDOF: 6
```
c-p-schmidt: And here

c-p-schmidt (dismissed):

c-p-schmidt (dismissed):

bgoderbauer (approved):

c-p-schmidt (approved):

PR #2099, opened 2026-06-25
Cleanup of scatra ele boundary calc methods

c-p-schmidt:
## Description and Context
I need to implement some additional mechanics dependence for the scatra-scatra interface (s2i) kinetics and decided to kick this off with a little clean-up.

It mainly consists of:
- renaming s2_i to s2i (renaming mistake from global renaming to consistent naming)
- renaming *_rh_sand_* to *_rhs_and_* (renaming mistake from global renaming to consistent naming)
- add consts, nodiscard where reasonable
- remove unnecessary doc (in cpp, due to override)

### DIFF
--- a/src/scatra_ele/4C_scatra_ele_boundary_calc_elch.cpp
+++ b/src/scatra_ele/4C_scatra_ele_boundary_calc_elch.cpp
@@ -17,7 +17,6 @@
 FOUR_C_NAMESPACE_OPEN
 
 /*----------------------------------------------------------------------*
- | protected constructor for singletons                      fang 01/15 |
  *----------------------------------------------------------------------*/
 template <Core::FE::CellType distype, int probdim>
 Discret::Elements::ScaTraEleBoundaryCalcElch<distype, probdim>::ScaTraEleBoundaryCalcElch(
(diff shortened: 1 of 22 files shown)


lauraengelhardt (dismissed):

src/scatra_ele/4C_scatra_ele_boundary_calc_elch.cpp
```diff
@@ -114,18 +99,22 @@ void Discret::Elements::ScaTraEleBoundaryCalcElch<distype, probdim>::calc_elch_b
   // condition) but the electrode status is evaluated
   const auto zerocur = cond->parameters().get<int>("ZERO_CUR");
   if (nume < 0)
+  {
     FOUR_C_THROW(
         "The convention for electrochemical reactions at the electrodes does not allow \n"
         "a negative number of transferred electrons");
+  }
```
bgoderbauer:
```suggestion
FOUR_C_ASSERT_ALWAYS(nume >= 0, 
"The convention for electrochemical reactions at the electrodes does not allow \n"
"a negative number of transferred electrons");
```

If you're already in cleanup? 😄

bgoderbauer (dismissed):

src/scatra_ele/4C_scatra_ele_boundary_calc_elch.cpp
c-p-schmidt: I already thought about it, but then was to lazy to do it right away.
Thanks for the virtual push 😆

bgoderbauer (approved):

lauraengelhardt (approved):

PR #702, opened 2025-04-25
Add runtime output to 1d arteries

lkoeglmeier:
## Description and Context
This pr enables the vtk-based output in the 1D artery elements
The old output mechanism was deleted. At least as far as I have seen, it has only output the owner so far. Maybe I miss something. To check the new output, the variables were temporarily added to the `4C_post_processor_single_field_writer.cpp`.

Issue #212, opened 2025-01-08
Cleanup after unification of output writers


eulovi:
> This issue was originally created by @isteinbrecher in https://gitlab.lrz.de/baci/baci/-/issues/900 on 2022-11-16.

## Description

Cleanup after unification of output writers. This includes renaming functions and classes to names that best represent what something is doing (this might not be clear when the names are first introduced). Furthermore, this includes the removal of all `post_*` stuff, i.e., ensuring a unified way of writing output in BACI.

## Related Issues and Merge Requests

* Blocks
* Is blocked by
* Follows [899](https://gitlab.lrz.de/baci/baci/-/issues/899)
* Precedes
* Related to
* Part of [896](https://gitlab.lrz.de/baci/baci/-/issues/896)
* Composed of

## Interested Parties
@baci/team_output

PR #2100, opened 2026-06-25
Enable vtk based output in red-airways

lkoeglmeier:
This pr adds the new vtk-output for the reduced airways and as a little cleanup it removes the post-processor writers of the 1d arteries. This module already uses the new vtk-output since #702.
Additionally, we fixed the check in the visualization writer so that column and row maps are allowed for the element based result data. Thanks a lot @c-p-schmidt for your help.


src/red_airways/4C_red_airways_implicitintegration.hpp
```diff
@@ -151,8 +152,11 @@ namespace Airway
     \brief update configuration and output to file/screen
 
     */
-    void output(
-        bool CoupledTo3D = false, std::shared_ptr<Teuchos::ParameterList> CouplingParams = nullptr);
+    void output(bool CoupledTo3D = false,
+        const std::shared_ptr<Teuchos::ParameterList>& CouplingParams = nullptr);
```
c-p-schmidt: I think it makes sense to keep the original version. I know, clang-tidy likes to suggest this, but you are only handing in a pointer anyway.

c-p-schmidt (commented):
Looks good to me. Now you just need to find the problematic element quantity/ies.

tests/input_files/fsi_artery_coupled_0d_3d_0d.4C.yaml
```diff
@@ -9,6 +9,7 @@ REDUCED DIMENSIONAL AIRWAYS DYNAMIC:
   MAXITERATIONS: 40
   TOLERANCE: 0.0001
   LINEAR_SOLVER: 1
+  RESTARTEVERY: 1
```
c-p-schmidt: I guess this change is irrelevant, but it wouldn't matter if we keep it.

tests/input_files/red_airway_one_acinus_Exponential.4C.yaml
```diff
@@ -12,6 +12,8 @@ REDUCED DIMENSIONAL AIRWAYS DYNAMIC:
   MAXITERATIONS: 40
   TOLERANCE: 1e-07
   LINEAR_SOLVER: 1
+IO/RUNTIME VTK OUTPUT:
```
c-p-schmidt: Same here, I guess.

c-p-schmidt (dismissed):

lkoeglmeier:
We found that quantities that are written for restart after calling `write_mesh()` are not stored correctly in the result section of the `.control` file and can not be read by our restart reader. The problem is documented in #212. Thank you @c-p-schmidt for your help!

c-p-schmidt (approved):

bgoderbauer (approved):

PR #2102, opened 2026-06-29
Parallelize periodic RVE boundary conditions

m-frey:
## Description and Context

The periodic RVE boundary conditions in the constraint framework were
serial-only and aborted with a FOUR_C_THROW on more than one rank. This
enables them in parallel.

Changes:
- Match periodic node pairs with a distributed ArborX search
  (`global_collision_search`) instead of a rank-local BVH query.
- Ghost the matched partner and corner nodes onto the owning rank so all
  four nodes of a constraint are locally available.
- Assemble each constraint row on a single owner; derive `Q_dL`
  and the residual from a distributed mat-vec, so no off-processor
  assembly is needed.

### DIFF
--- a/src/constraint_framework/4C_constraint_framework_equation.cpp
+++ b/src/constraint_framework/4C_constraint_framework_equation.cpp
@@ -16,25 +16,11 @@ FOUR_C_NAMESPACE_OPEN
 /*----------------------------------------------------------------------------*
  *----------------------------------------------------------------------------*/
 void Constraints::SubmodelEvaluator::LinearCoupledEquation::evaluate_equation(
-    Core::LinAlg::SparseMatrix& Q_dd, Core::LinAlg::SparseMatrix& Q_dL,
-    Core::LinAlg::SparseMatrix& Q_Ld, Core::LinAlg::Vector<double>& constraint_vector,
-    const Core::LinAlg::Vector<double>& D_np1)
+    Core::LinAlg::SparseMatrix& Q_Ld)
 {
-  double constraintViolation = 0.;
-
-  // Iterate over the elements (coefficient, rowId, dofId) in equationData.
-  // Each element of equation data represents one term of the defined multipoint constraints
-  // The rowId is equivalent to the Number of the equation
-  for (const auto& [coefficient, rowId, dofId] : equation_data_)
-  {
-    // stiffness contribution
-    Q_dL.assemble(coefficient, dofId, rowId);
-    Q_Ld.assemble(coefficient, rowId, dofId);
-
-    // force contribution
-    constraintViolation = D_np1.get_values()[dofId] * coefficient;
-    constraint_vector.sum_into_global_values(1, &constraintViolation, &rowId);
-  }
+  // assemble the rows owned by this rank
+  for (const auto& [coefficient, row_id, dof_id] : equation_data_)
+    if (Q_Ld.row_map().my_gid(row_id)) Q_Ld.assemble(coefficient, row_id, dof_id);
 }
 /*----------------------------------------------------------------------------*
  *----------------------------------------------------------------------------*/
--- a/src/constraint_framework/4C_constraint_framework_submodelevaluator_base.hpp
+++ b/src/constraint_framework/4C_constraint_framework_submodelevaluator_base.hpp
@@ -15,6 +15,8 @@
 #include "4C_structure_new_enum_lists.hpp"
 #include "4C_structure_new_model_evaluator_generic.hpp"
 
+#include <vector>
+
 FOUR_C_NAMESPACE_OPEN
 
 namespace Constraints::SubmodelEvaluator
@@ -58,6 +60,9 @@ namespace Constraints::SubmodelEvaluator
     //! Return the Penalty-Parameter
     double& get_penalty_parameter_ptr() { return penalty_parameter_; }
 
+    //! Set the constraint rows owned by this rank
+    void set_owned_constraint_row_ids(std::vector<int> row_ids);
+
     //! Generate the runtime output
     virtual void runtime_output_step_state(std::pair<double, int> output_time_and_step) {};
 
@@ -69,6 +74,12 @@ namespace Constraints::SubmodelEvaluator
 
     //! Penalty parameter
     double penalty_parameter_;
+
+    //! Constraint rows owned by this rank
+    std::vector<int> owned_constraint_row_ids_;
+
+    //! Use owned_constraint_row_ids_ for the constraint map
+    bool use_explicit_constraint_row_ids_ = false;
 
    protected:
     //! Vector containing all constraint equation objects
--- a/src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
+++ b/src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
@@ -10,13 +10,12 @@
 #include "4C_constraint_framework_submodelevaluator_mpc.hpp"
 
 #include "4C_beam3_base.hpp"
+#include "4C_comm_mpi_utils.hpp"
 #include "4C_constraint_framework_equation.hpp"
 #include "4C_fem_condition.hpp"
 #include "4C_fem_discretization.hpp"
-#include "4C_geometric_search_access_traits.hpp"
 #include "4C_geometric_search_bounding_volume.hpp"
-#include "4C_geometric_search_bvh.hpp"
-#include "4C_geometric_search_input.hpp"
+#include "4C_geometric_search_distributed_tree.hpp"
 #include "4C_global_data.hpp"
 #include "4C_io.hpp"
 #include "4C_linalg_sparsematrix.hpp"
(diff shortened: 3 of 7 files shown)


src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
```diff
@@ -166,6 +207,17 @@ void Constraints::SubmodelEvaluator::RveMultiPointConstraintManager::check_input
   discret_ptr_->get_condition(
       "PointLinearCoupledEquation", point_linear_coupled_equation_conditions_);
 
+  const bool is_parallel = Core::Communication::num_mpi_ranks(discret_ptr_->get_comm()) > 1;
+  if (is_parallel &&
+      rve_ref_type_ == Constraints::MultiPoint::RveReferenceDeformationDefinition::manual)
+  {
+    FOUR_C_THROW("Manual RVE reference points are not implemented in parallel.");
```
georghammerl, line 214: May be use FOUR_C_ASSERT_ALWAYS here. Same below.

src/constraint_framework/4C_constraint_framework_submodelevaluator_base.hpp
```diff
@@ -58,6 +60,9 @@ namespace Constraints::SubmodelEvaluator
     //! Return the Penalty-Parameter
     double& get_penalty_parameter_ptr() { return penalty_parameter_; }
 
+    //! Set the constraint rows owned by this rank
```
mayrmt, line 63: Are these actually row IDs in a matrix or the global IDs of DOFs, that are subject to a constraint?

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
```diff
@@ -25,17 +24,62 @@
...
+  struct PeriodicRveMpcNodeSet
+  {
+    int plus_gid;
+    int minus_gid;
+    int ref_end_gid;
+    int ref_base_gid;
+  };
+
+  //! Suspend floating point exception trapping while in scope. ArborX may raise benign fp
+  //! exceptions on ranks whose local search input is empty.
```
mayrmt, line 49:
```suggestion
  //! Suspend floating point exception trapping while in scope. ArborX may raise benign floating point
  //! exceptions on ranks whose local search input is empty.
```

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
```diff
@@ -25,17 +24,62 @@
...
+}  // namespace
+
 /*----------------------------------------------------------------------------*
  *----------------------------------------------------------------------------*/
 Constraints::SubmodelEvaluator::RveMultiPointConstraintManager::RveMultiPointConstraintManager(
-    std::shared_ptr<const Core::FE::Discretization> disc_ptr, Core::LinAlg::SparseMatrix* st_ptr)
+    std::shared_ptr<Core::FE::Discretization> disc_ptr, Core::LinAlg::SparseMatrix* st_ptr)
 {
   discret_ptr_ = disc_ptr;
+  writable_discret_ = std::move(disc_ptr);
```
mayrmt, line 82: 1. Is `discret_ptr_` still a valid point with actual data at this point?
2. Is it still used anywhere?

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
```diff
@@ -874,6 +935,42 @@ void Constraints::SubmodelEvaluator::RveMultiPointConstraintManager::
...
+      ghosted_node_gids.data(), 0, writable_discret_->get_comm());
+  writable_discret_->export_column_nodes(ghosted_node_map);
+  writable_discret_->fill_complete();
+
+  // ghosting must not change the dof row map
+  const Core::LinAlg::Map& dof_row_map_after = *writable_discret_->dof_row_map();
+  const std::vector<int> dof_gids_after(dof_row_map_after.my_global_elements(),
+      dof_row_map_after.my_global_elements() + dof_row_map_after.num_my_elements());
+  if (dof_gids_before != dof_gids_after)
+    FOUR_C_THROW("Ghosting the periodic partner nodes changed the dof row map.");
```
mayrmt, line 972: That looks like code, that should be run in DEBUG mode only.

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
```diff
@@ -302,141 +354,150 @@ void Constraints::SubmodelEvaluator::RveMultiPointConstraintManager::build_perio
...
-              Core::IO::cout(Core::IO::verbose)
-                  << bounding_volumes_neg[i].first << "," << indices(i) << ","
-                  << rveCornerNodeIdMap[surf.second] << "," << rveCornerNodeIdMap["N1"]
-                  << Core::IO::endl;
-            }
-          }
+        std::map<int, std::vector<int>> positive_partners;
+        for (const auto& match : matches)
+          positive_partners[match.gid_predicate].push_back(match.gid_primitive);
+
```
bennoschoenstein, line 439: What about if nodes don't find any partner? 

The old code checked `nHits == 0` for every '-' node and threw. 
Now `positive_partners` only contains nodes that actually found a partner. A '-' node without a partner (mesh not perfectly periodic, or POINT_TOLERANCE too small) is now skipped - the simulation continues with a missing constraint and gives wrong results without any warning.

FIX:

```cpp
if (positive_partners.size() != shifted_negative_nodes.size())
        {
          std::string unmatched_gids;
          for (const auto& [minus_gid, bounding_volume] : shifted_negative_nodes)
            if (!positive_partners.contains(minus_gid))
              unmatched_gids += std::to_string(minus_gid) + " ";
          FOUR_C_THROW(
              "Periodic search on the '{}-' boundary found no partner for node(s): {}."
              " Check mesh periodicity or increase POINT_TOLERANCE.",
              axis, unmatched_gids);
        }
```

Maybe also worth a small test with an intentionally non-periodic mesh?

src/constraint_framework/4C_constraint_framework_equation.cpp
```diff
@@ -16,25 +16,11 @@ FOUR_C_NAMESPACE_OPEN
...
-    Q_dL.assemble(coefficient, dofId, rowId);
-    Q_Ld.assemble(coefficient, rowId, dofId);
-
-    // force contribution
-    constraintViolation = D_np1.get_values()[dofId] * coefficient;
-    constraint_vector.sum_into_global_values(1, &constraintViolation, &rowId);
-  }
+  // assemble the rows owned by this rank
+  for (const auto& [coefficient, row_id, dof_id] : equation_data_)
+    if (Q_Ld.row_map().my_gid(row_id)) Q_Ld.assemble(coefficient, row_id, dof_id);
```
bennoschoenstein, line 23: This `if` drops rows that are not in the row map. For the periodic constraints this never happens. But for `DESIGN POINT COUPLED DOF EQUATION CONDITIONS` together with `RVE_REFERENCE_POINTS: automatic` it does happen.

FIX:

```cpp
for (const auto& [coefficient, row_id, dof_id] : equation_data_)
{
    FOUR_C_ASSERT_ALWAYS(Q_Ld.row_map().my_gid(row_id),
        "Constraint equation row {} is not in the constraint row map and would "
        "be dropped.",
        row_id);
    Q_Ld.assemble(coefficient, row_id, dof_id);
  }
```

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
```diff
@@ -302,141 +354,150 @@ void Constraints::SubmodelEvaluator::RveMultiPointConstraintManager::build_perio
...
+          constraint_equations_.emplace_back(
+              std::make_shared<LinearCoupledEquation>(mpc_id++, pbc_dofs, pbc_coefficients));
+        }
+      }
+      set_owned_constraint_row_ids(std::move(owned_constraint_row_ids));
+
+      Core::IO::cout(Core::IO::verbose)
+          << "\nNumber of periodic constraint equations on this rank: "
+          << constraint_equations_.size() << Core::IO::endl;
+      return;
```
bennoschoenstein, line 500: This early `return` skips the "Ensure that no constraint equation is used twice" block further down, which the old automatic path still reached. 

Note: two duplicates can be created on different ranks. Maybe the simplest fix is to not create the duplicate in the first place, e.g. skip the known corner nodes in the '-' node loop for all axes but one

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
```diff
@@ -302,141 +354,150 @@ void Constraints::SubmodelEvaluator::RveMultiPointConstraintManager::build_perio
...
-            }
-          }
+        std::map<int, std::vector<int>> positive_partners;
+        for (const auto& match : matches)
+          positive_partners[match.gid_predicate].push_back(match.gid_primitive);
+
+        for (const auto& [minus_gid, partners] : positive_partners)
+        {
+          if (partners.size() != 1)
+            FOUR_C_THROW(
```
bennoschoenstein, line 443: If only one rank throws, the others can end up waiting instead of aborting

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
```diff
@@ -874,6 +935,42 @@ void Constraints::SubmodelEvaluator::RveMultiPointConstraintManager::
...
+  // the refill below is collective, so decide collectively whether to skip it
+  const int local_needs_ghosting =
+      static_cast<int>(ghosted_node_gids.size()) > node_col_map.num_my_elements() ? 1 : 0;
+  if (Core::Communication::max_all(local_needs_ghosting, writable_discret_->get_comm()) == 0)
+    return;
+
+  Core::LinAlg::Map ghosted_node_map(-1, static_cast<int>(ghosted_node_gids.size()),
+      ghosted_node_gids.data(), 0, writable_discret_->get_comm());
+  writable_discret_->export_column_nodes(ghosted_node_map);
+  writable_discret_->fill_complete();
```
bennoschoenstein, line 965: Note: this `fill_complete()` rebuilds the column maps and re-assigns the dofs. Anything that grabbed column-map data from this discretization *before* this constructor runs would silently be stale...

bennoschoenstein (changes requested):
Thanks a lot! Exactly what I need 🗡️

mayrmt:
@m-frey is there anything that is blocking this PR to move forward?

PR #2103, opened 2026-06-30
Move inpar_structure to structure_new_input

maxiludwig:
# Description and Context
- Move structure input definitions from `Inpar::Solid` in `src/inpar/` to the `structure_new` module in `src/structure_new/src/input/`.
- Relocate input files (4C_inpar_structure.* -> 4C_structure_new_input.*) and update all affected includes and global registration wiring.
- Migrate solid qualifiers from `Inpar::Solid` to `Solid` across dependent modules.

# Related Issues

Part of https://github.com/4C-multiphysics/4C/issues/59


src/structure_new/src/input/4C_structure_new_input.cpp
```diff
@@ -0,0 +1,661 @@
...
+
+            parameter<bool>("TIME_ADAPTIVITY",
+                {.description = "Enable adaptive time integration", .default_value = false}),
+
+            deprecated_selection<Solid::DynamicType>("DYNAMICTYPE",
+                {
+                    {"Statics", DynamicType::Statics},
+                    {"GenAlpha", DynamicType::GenAlpha},
+                    {"GenAlphaLieGroup", DynamicType::GenAlphaLieGroup},
+                    {"OneStepTheta", DynamicType::OneStepTheta},
```
bgoderbauer, line 87: I know you only moved these files, but I'm curious: do you know why everything is a `deprecated_selection` and not a `selection` here?

bgoderbauer (approved):
Thanks for the effort!

src/structure_new/src/input/4C_structure_new_input.cpp
maxiludwig: No I don't know unfortunately. @lauraengelhardt Do you have an idea?

lauraengelhardt: If I remember correctly, `selection` uses an enum to choose between `InputSpecs`. `deprecated_selection` uses a string-to-value map. If the enum names are identical to the strings its easy to switch (which should be the case here).

lauraengelhardt (approved):
I didn't review every single line ofc, but trust your search and replace skills ;)

src/structure_new/src/input/4C_structure_new_input.cpp
lauraengelhardt: I have to correct myself here: selection is for choosing between groups of parameters, not between simple values (e.g. either you give the group of parameters needed to work with model A or the group of parameters needed to work with model B). If you just want to switch between different enums you would do it like this: 

```
class MyChoice
{
  ABC,
  DEF,
};

parameter<MyChoice>("my_param", {.description = "Choose ABC or DEF."});
```

davidrudlstorfer:
@maxiludwig thanks for tackling this :)

PR #2108, opened 2026-07-01
Add dynamic rebalance to structure problems

LaSi5002:
This branch adds rebalancing of rank load mid simulation run based on element-wise `Core::Elements::Element::evaluate` times to structure problems.

```
STRUCTURAL DYNAMIC/DYNAMIC REBALANCE:
  ENABLED: true
  IMBALANCE_THRESHOLD: 1.35 # if the rank imbalance max/mean reaches this threshold, the rebalance is triggered.
  EDGE_WEIGHT_MULTIPLIER: 5 # multiplier for edge weights during redistribution to take communication costs into account
  WINDOW_STEPS: 3 # the imbalance trigger metric (max/mean) is averaged over the last 3 steps to avoid triggering on random spikes
  COOLDOWN_STEPS: 100 # minimum number of steps between rebalances
  IMBALANCE_TOL: 1.4 # allow vertex imbalance during redistribution. Higher values allow for more communication aware redistribution
  MIN_ELE_PER_PROC: 0
  REBALANCE_TYPE: Hypergraph # multijagged and monolithic also available
```

An example of a use case could be a stress test with the reformulated johnson cook model. Necking can lead to locally higher evaluation times which in turn leads to rank imbalances and idle times.
image
(Per element evaluation times of stress test with reformulated johnson cook)

.

.

image
(Rank evaluation times without rebalance)

.

.

image
(Rank evaluation times with rebalance)

.

.

image
(max/mean rank evaluation times without rebalance)

.

.

image
(max/mean rank evaluation times with rebalance)

.

.

Core::FE::Discretization::Evaluate took 441.1 s on the slowest rank without rebalance and 267.8 s with rebalance (39.28% reduction) The whole simulation took 1057.9 s. Rebalancing took 0.8466 s over 32 rebalances in total.


src/core/fem/src/discretization/4C_fem_discretization.cpp
```diff
@@ -684,10 +684,10 @@ void Core::FE::Discretization::reset_element_eval_timers()
 std::vector<double> Core::FE::Discretization::get_rank_eval_times_on_root() const
 {
   double my_eval_time_sum = 0.0;
-  for (auto* ele : elecolptr_)
+  for (auto* ele : elerowptr_)
```
mayrmt: Field discretizations in 4C loop over the column element map, i.e. elements on processor boundaries are evaluated by all participating processors, such that each processor can write into its rows of the matrix/vector (determined by the unique dof row map). This change seems to neglect this. If considering row elements only, then elements across processor boundaries are most likely not treated properly.

src/core/rebalance/src/4C_rebalance.hpp
```diff
@@ -13,6 +13,8 @@
 #include <mpi.h>
 #include <Teuchos_ParameterList.hpp>
 
+#include <memory>
```
mayrmt: It is not clear to me, why this inclusion is necessary.

src/structure_new/src/input/4C_structure_new_input.cpp
```diff
@@ -423,6 +424,48 @@ namespace Solid
...
+            parameter<bool>("ENABLED", {.description = "Enable dynamic rebalance based on "
+                                                       "Core::Elements::Element::evaluate() times "
+                                                       "for pure structure problems.",
+                                           .default_value = false}),
+            parameter<double>("IMBALANCE_THRESHOLD",
+                {.description = "Trigger rebalance if the rolling average of the "
+                                "max-to-mean rank evaluation time ratio exceeds this value.",
+                    .default_value = 1.5}),
+            parameter<int>("WINDOW_STEPS",
+                {.description = "Number of converged steps used in the rolling imbalance average.",
```
mayrmt: What do you mean by "converged step"?

src/structure_new/src/input/4C_structure_new_input.cpp
```diff
@@ -423,6 +424,48 @@ namespace Solid
...
+                    .default_value = 3}),
+            parameter<int>("COOLDOWN_STEPS",
+                {.description = "Minimum number of converged steps between rebalances.",
+                    .default_value = 3}),
+            parameter<double>("IMBALANCE_TOL",
+                {.description = "Target vertex imbalance passed to the repartitioner. Higher "
+                                "values allow edge weights to have greater effect.",
+                    .default_value = 1.1}),
+            parameter<int>("MIN_ELE_PER_PROC",
+                {.description = "Minimum number of elements per processor used for during dynamic "
```
mayrmt:
```suggestion
                {.description = "Minimum number of elements per processor used for dynamic "
```

mayrmt (commented):
@LaSi5002 Thanks for this contribution! I like the extra capability.

I did not have a detailed look, yet left some initial comments. My current main concern ist the switch from the element column map to the element row map, since this can lead to neglecting some elements at processor boundaries as far as I understand it. Please double-check.

src/core/fem/src/discretization/4C_fem_discretization.cpp
LaSi5002: Yes, you're right. I wanted to avoid double counting evaluation times, but the elements exist on every rank seperately with their own evaluation time and only store the evaluation time actually generated on their branch. I will change it tomorrow.

src/structure_new/src/input/4C_structure_new_input.cpp
LaSi5002: I just mean steps. Will clarify this tomorrow.

src/structure_new/src/input/4C_structure_new_input.cpp
LaSi5002: Thanks!

src/core/rebalance/src/4C_rebalance.hpp
LaSi5002: I added it while experimenting and forgot it there. I removed it in 8479b63fc63f8e7a746c082c4afc6df09cf9d7af

src/structure_new/src/input/4C_structure_new_input.cpp
mayrmt: Ok, that shifts my question to: what do you mean by "steps"? Nonlinear iterations? Time steps? Load steps?

LaSi5002: Im sorry, I mean time steps. At every time step the imbalance is computed and pushed back onto the history deque with max length `WINDOW_STEPS`

LaSi5002: Clarified in 30dfadf387fa7b60e581dda84186984bac588553

tests/list_of_tests.cmake
```diff
@@ -2023,6 +2023,25 @@ four_c_test(TEST_FILE solid_nodal_fiber.4C.yaml NP 2)
 four_c_test(TEST_FILE solid_nodal_fiber_eletypes.4C.yaml NP 2)
 four_c_test(TEST_FILE solid_nurbs27_hex8_ele_neumann_bcs.4C.yaml NP 1 RETURN_AS current)
 __four_c_test_vtk(BASED_ON ${current} PVD_RESULT xxx-structure.pvd PVD_REFERENCE ref/solid_nurbs27_hex8_ele_neumann_bcs-structure.pvd TOLERANCE 1e-08)
+four_c_test(TEST_FILE solid_dynamic_rebalance_simple_5x2x1.4C.yaml NP 2 RETURN_AS current REQUIRED_DEPENDENCIES TRILINOS_MUMPS)
+set(solid_dynamic_rebalance_simple_5x2x1_base ${current})
```
davidrudlstorfer: please check how this is done in the other tests, this addition should not be neccessary

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -658,6 +658,84 @@ function(__four_c_test_add_csv_yaml_comparison)
...
+#   RESULT_FILE:           name of the result file created by BASED_ON
+#   COMPARE_FILE:          name of the result file created by COMPARE_TO
+#   TOL_R:                 relative tolerance for comparison
+#   TOL_A:                 absolute tolerance for comparison
+#
+# optional parameters:
+#   LABELS:                add labels to the test
+#   REQUIRED_DEPENDENCIES: any required external dependencies. The test will be skipped if the dependencies are not met.
+#
+function(__four_c_test_add_csv_yaml_comparison_between_tests)
```
davidrudlstorfer, line 677: why do we need this new function? cannot this be handled/appended to the already existing functions?

src/core/fem/src/discretization/4C_fem_discretization.hpp
```diff
@@ -545,9 +545,12 @@ namespace Core::FE
     //! Reset stored element evaluation timers for all local column elements
     void reset_element_eval_timers();
 
-    //! Get per-rank evaluation times on rank 0, summed over all local column elements
+    //! Get per-rank evaluation times on rank 0, summed over all local row elements
```
c-p-schmidt: In the implementation, you are looping over all column elements, so why did you change the documentation?
If there is a specific reason, that only the row elements hold data, then an additional @note hint to document this would be helpful. This also applies to the newly introduced method `get_rank_eval_times()`.

src/core/rebalance/src/4C_rebalance.cpp
```diff
@@ -138,8 +146,14 @@ void Core::Rebalance::rebalance_discretization(Core::FE::Discretization& discret
...
-    std::tie(rowmap, colmap) = do_rebalance_discretization(
-        *graph, discretization, rebalanceMethod, rebalanceParams, parameters, comm);
+    const std::optional<Core::Rebalance::PartitionWeights> effective_partition_weights =
+        use_eval_time_weights
+            ? std::make_optional(Core::Rebalance::build_eval_time_partition_weights(
+                  discretization, *graph, parameters.edge_weight_multiplier))
+            : std::nullopt;
+    std::tie(rowmap, colmap) = do_rebalance_discretization(*graph, discretization, rebalanceMethod,
+        rebalanceParams, parameters, comm,
+        effective_partition_weights ? &effective_partition_weights.value() : nullptr);
```
c-p-schmidt: Would be a bit more expressive, I think.
```suggestion
        effective_partition_weights.has_value() ? &effective_partition_weights.value() : nullptr);
```

src/core/rebalance/src/4C_rebalance_graph_based.hpp
```diff
@@ -102,16 +102,29 @@ namespace Core::Rebalance
...
 
   @return Node and edge weights to be used for repartitioning
   */
-  std::pair<std::shared_ptr<Core::LinAlg::Vector<double>>,
-      std::shared_ptr<Core::LinAlg::SparseMatrix>>
-  build_weights(const Core::FE::Discretization& dis);
+  PartitionWeights build_static_partition_weights(const Core::FE::Discretization& dis);
+
+  /**
+   * Build repartitioning weights on the rebalance graph map.
```
c-p-schmidt:
```suggestion
   * @brief Build repartitioning weights on the rebalance graph map.
```

src/structure_new/src/implicit/4C_structure_new_impl_ost.cpp
```diff
@@ -75,6 +75,22 @@ void Solid::IMPLICIT::OneStepTheta::setup()
   issetup_ = true;
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+void Solid::IMPLICIT::OneStepTheta::remap_integrator_state_after_redistribution()
+{
+  check_init();
+
+  const_vel_acc_update_ptr_ = std::make_shared<Core::LinAlg::MultiVector<double>>(
```
c-p-schmidt, line 84: Is the `replace_map` method of the MultiVector not doing what you want?

src/structure_new/src/implicit/4C_structure_new_impl_genalpha_liegroup.cpp
```diff
@@ -43,6 +43,16 @@ void Solid::IMPLICIT::GenAlphaLieGroup::setup()
...
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+void Solid::IMPLICIT::GenAlphaLieGroup::remap_integrator_state_after_redistribution()
+{
+  GenAlpha::remap_integrator_state_after_redistribution();
+
+  accn_mod_ =
+      std::make_shared<Core::LinAlg::Vector<double>>(*global_state().dof_row_map_view(), true);
```
c-p-schmidt, line 53: Is the `replace_map` method of the Vector not doing what you want?

src/structure_new/src/implicit/4C_structure_new_impl_genalpha.cpp
```diff
@@ -125,6 +125,22 @@ void Solid::IMPLICIT::GenAlpha::setup()
   issetup_ = true;
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+void Solid::IMPLICIT::GenAlpha::remap_integrator_state_after_redistribution()
+{
+  check_init();
+
+  const_vel_acc_update_ptr_ = std::make_shared<Core::LinAlg::MultiVector<double>>(
```
c-p-schmidt, line 134: Is the `replace_map` method of the MultiVector not doing what you want?

src/structure_new/src/input/4C_structure_new_input.cpp
```diff
@@ -423,6 +424,48 @@ namespace Solid
...
+                                           .default_value = false}),
+            parameter<double>("IMBALANCE_THRESHOLD",
+                {.description = "Trigger rebalance if the rolling average of the "
+                                "max-to-mean rank evaluation time ratio exceeds this value.",
+                    .default_value = 1.5}),
+            parameter<int>("WINDOW_STEPS",
+                {.description = "Number of time steps used in the rolling imbalance average.",
+                    .default_value = 3}),
+            parameter<int>("COOLDOWN_STEPS",
+                {.description = "Minimum number of converged steps between rebalances.",
```
c-p-schmidt: Probably also?
```suggestion
                {.description = "Minimum number of converged time steps between rebalances.",
```

src/structure_new/src/input/4C_structure_new_input.cpp
```diff
@@ -423,6 +424,48 @@ namespace Solid
...
+                    .default_value = 1.1}),
+            parameter<int>("MIN_ELE_PER_PROC",
+                {.description = "Minimum number of elements per processor used for dynamic "
+                                "rebalances.",
+                    .default_value = 0}),
+            parameter<double>("EDGE_WEIGHT_MULTIPLIER",
+                {.description = "Multiplier applied graph edge weights during dynamic rebalance. "
+                                "Represents the inter-rank communication overhead per edge.",
+                    .default_value = 1.0}),
+            deprecated_selection<Core::Rebalance::RebalanceType>("REBALANCE_TYPE",
```
c-p-schmidt: Why didn't you use parameter<Core::Rebalance::RebalanceType> instead? Then the available options in the input file should be in lowercase.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
```diff
@@ -326,6 +343,37 @@ bool Solid::ModelEvaluator::Structure::initialize_inertia_and_damping()
   return eval_error_check();
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+bool Solid::ModelEvaluator::Structure::initialize_inertia_and_damping(
```
c-p-schmidt: Why do you need this new method?

src/structure_new/src/4C_structure_new_timint_base.cpp
```diff
@@ -120,6 +129,91 @@ void Solid::TimeInt::Base::post_setup()
   int_ptr_->post_setup();
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+void Solid::TimeInt::Base::post_output()
+{
+  check_init_setup();
+  maybe_perform_dynamic_rebalance();
```
c-p-schmidt: I know this isn't handled cleanly throughout 4C at the moment, but (optionally) performing the rebalancing in the post_output method doesn't feel right.

You could create a rebalance method on the top level and place it there. Then it is obvious what is happening.

src/structure_new/src/4C_structure_new_timint_base.cpp
```diff
@@ -120,6 +129,91 @@ void Solid::TimeInt::Base::post_setup()
...
+
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+void Solid::TimeInt::Base::maybe_perform_dynamic_rebalance()
+{
+  const auto& rebalance_config = datasdyn_->get_dynamic_rebalance_config();
+  if (!rebalance_config.enabled) return;
+
+  const std::vector<double> rank_eval_times =
+      dataglobalstate_->get_discret()->get_rank_eval_times();
```
c-p-schmidt, line 140: Can that even happen? You have checked before that rebalancing is activated, right?

tests/list_of_tests.cmake
LaSi5002: Yes, you're right. I simplified it in dd007632de6cfb729de175afb704928f4bfdf8fb

tests/list_of_tests.cmake
```diff
@@ -2023,6 +2023,23 @@ four_c_test(TEST_FILE solid_nodal_fiber.4C.yaml NP 2)
 four_c_test(TEST_FILE solid_nodal_fiber_eletypes.4C.yaml NP 2)
 four_c_test(TEST_FILE solid_nurbs27_hex8_ele_neumann_bcs.4C.yaml NP 1 RETURN_AS current)
 __four_c_test_vtk(BASED_ON ${current} PVD_RESULT xxx-structure.pvd PVD_REFERENCE ref/solid_nurbs27_hex8_ele_neumann_bcs-structure.pvd TOLERANCE 1e-08)
+four_c_test(TEST_FILE solid_dynamic_rebalance_simple_5x2x1.4C.yaml NP 2 RETURN_AS solid_dynamic_rebalance_simple_5x2x1_base REQUIRED_DEPENDENCIES TRILINOS_MUMPS)
+four_c_test(TEST_FILE solid_dynamic_rebalance_simple_5x2x1_rebal.4C.yaml NP 2 RETURN_AS solid_dynamic_rebalance_simple_5x2x1_rebal REQUIRED_DEPENDENCIES TRILINOS_MUMPS)
+__four_c_test_add_csv_yaml_comparison_between_tests(
```
c-p-schmidt: Why do you do this check?

c-p-schmidt (commented):
Nice work @LaSi5002!

cmake/functions/four_c_testing_functions.cmake
LaSi5002: As far a as I understand, in the existing functions, the csv file that is compared against is hardcoded and has to exist prior to any runs. The idea of this new function is to run two tests and compare both their csvs. This ensures that the change in the second yaml file does not alter the physical state of the solution. 

You could also run a simulation without rebalancing once before and hardcode that solution as the target csv, but for me there were two reasons to do it this way. On the one hand, with this new function we make it clear that we are comparing the results of two different yaml files, not just checking if the solution of one yaml file doesn't change, and on the other hand, we ensure that this test would also survive any later code changes that alter the solution. 
But I see why this might be overengineering and I could opt for the simpler solution.

src/core/fem/src/discretization/4C_fem_discretization.hpp
LaSi5002: Both `get_rank_eval_times` functions should sum over the column elements as the ghosted elements also contribute to the overall rank evaluation time. I forgot the comments in the hpp file, but fixed it now in 22dc8954197d4abfdf2d801cf4cadab4c738f524

src/structure_new/src/implicit/4C_structure_new_impl_ost.cpp
LaSi5002: `replace_map` does only work if the map sizes are the same. Here they can be different because the rebalance can change the number of entries on each rank.

src/structure_new/src/input/4C_structure_new_input.cpp
LaSi5002: Yes, that is a much cleaner option. Did it in 8243217b781ea8024504dafe844471ffba99e6f7

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
LaSi5002: I added it, because I needed to set displacement and velocity during rebalance and wanted to change as little running code as possible. In the new commit 547907223e24733eef1bc56c3c2f57fe7022f14d I now get the displacement directly from global_state, which should be zeros on startup, but work aswell during rebalance.
Do you think this is a robust approach?

src/structure_new/src/4C_structure_new_timint_base.cpp
LaSi5002: Yeah, that's right. This is unnecessary and misleading. Removed in fbf5ea116d81ee1c87ecb77f755f2f8049b563e1

src/structure_new/src/4C_structure_new_timint_base.cpp
LaSi5002: It now lives directly in `Adapter::StructureTimeLoop::integrate` 2c7d10cfbb1985c17c6379c5419e6e9b826dbffe

tests/list_of_tests.cmake
LaSi5002: This test was added to specifically only show, that a change does not alter the solution of simulation. 
It checks whether the yaml with rebalance produces the same solution as the yaml without rebalance.
See https://github.com/4C-multiphysics/4C/pull/2108#discussion_r3526981558

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
LaSi5002: Removed the function and added matrix redistribution to `redistribute_and_preserve_state`

tests/input_files/solid_dynamic_rebalance_simple_5x2x1.4C.yaml
```diff
@@ -0,0 +1,157 @@
...
+  TIMESTEP: 0.02
+  RESTARTEVERY: 120
+  NUMSTEP: 10
+  MAXTIME: 24.0
+  TOLDISP: 1.0e-6
+  TOLRES: 1.0e-6
+  PREDICT: ConstVel
+  MAXITER: 100
+  DIVERCONT: halve_step
+  MAXDIVCONREFINEMENTLEVEL: 8
```
rjoussen: This should be removed, same in the other test file.

src/structure_new/src/4C_structure_new_timint_base.cpp
```diff
@@ -45,14 +51,16 @@ Solid::TimeInt::Base::Base()
       datasdyn_(nullptr),
       dataglobalstate_(nullptr),
       int_ptr_(nullptr),
-      dbc_ptr_(nullptr)
+      dbc_ptr_(nullptr),
+      last_dynamic_rebalance_step_(std::numeric_limits<int>::min() / 2)
```
rjoussen: What is the ida here? Shouldnt this rather be initialized with 0?

src/structure_new/src/4C_structure_new_timint_base.cpp
```diff
@@ -120,6 +129,86 @@ void Solid::TimeInt::Base::post_setup()
   int_ptr_->post_setup();
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+void Solid::TimeInt::Base::post_output() { check_init_setup(); }
+
```
rjoussen: Why is this needed now?

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
```diff
@@ -132,6 +132,38 @@ std::shared_ptr<Core::Conditions::Condition> Solid::MonitorDbc::create_reaction_
...
+  {
+    const Core::Conditions::Condition& rcond = *rcond_ptr;
+    auto ipair = react_maps_.insert(std::make_pair(
+        rcond.node_set_name(), std::vector<std::shared_ptr<Core::LinAlg::Map>>(3, nullptr)));
+
+    if (not ipair.second)
+      FOUR_C_THROW("The reaction condition id #{} seems to be non-unique!", rcond.node_set_name());
+
+    create_reaction_maps(*discret_ptr_, rcond, ipair.first->second);
+  }
```
rjoussen: This is exactly the same as in `setup()`. I would do this in a small helper function. Also, you use hardcoded 3 as the vector size here, which should be `discret_ptr_->n_dim()`, as in setup.

src/structure_new/src/input/4C_structure_new_input.cpp
```diff
@@ -423,6 +424,40 @@ namespace Solid
                 {.description = "Function for Initial displacement", .default_value = -1})},
         {.required =
                 false})); /*--------------------------------------------------------------------*/
+    specs.push_back(group("STRUCTURAL DYNAMIC/DYNAMIC REBALANCE",
```
rjoussen:
```suggestion
    specs.push_back(group<Core::Rebalance::RebalanceParameters>("STRUCTURAL DYNAMIC/DYNAMIC REBALANCE",
```
You can do something like this and store the parameters directly in the struct. Then you could probably also make this an `std::optional<Core::Rebalance::RebalanceParameters>` and get rid of `ENABLED`, but just turn on rebalancing if the group is present.

src/structure_new/src/input/4C_structure_new_input.cpp
```diff
@@ -423,6 +424,40 @@ namespace Solid
...
+            parameter<double>("IMBALANCE_TOL",
+                {.description = "Target vertex imbalance passed to the repartitioner. Higher "
+                                "values allow edge weights to have greater effect.",
+                    .default_value = 1.1}),
+            parameter<int>("MIN_ELE_PER_PROC",
+                {.description = "Minimum number of elements per processor used for dynamic "
+                                "rebalances.",
+                    .default_value = 0}),
+            parameter<double>("EDGE_WEIGHT_MULTIPLIER",
+                {.description = "Multiplier applied graph edge weights during dynamic rebalance. "
```
dragos-ana:
```suggestion
                {.description = "Multiplier applied to graph edge weights during dynamic rebalance. "
```

src/core/rebalance/src/4C_rebalance.hpp
```diff
@@ -89,7 +98,8 @@ namespace Core::Rebalance
    * This is a collective call over all ranks in @p comm.
    */
   void rebalance_discretization(Core::FE::Discretization& discretization,
-      const Core::LinAlg::Map& row_elements, const RebalanceParameters& parameters, MPI_Comm comm);
+      const Core::LinAlg::Map& row_elements, const RebalanceParameters& parameters, MPI_Comm comm,
+      bool use_eval_time_weights = false);
```
dragos-ana, line 101: Please adapt the documentation to the new parameter introduced here.

src/structure_new/src/4C_structure_new_timint_base.cpp
LaSi5002: This initial value (0 or very negative) just determines whether rebalance can occur in the first `COOLDOWN_STEPS` steps. In my opinion early rebalance should be allowed, because `eval_time` dependent rebalance is of course not an option as the initial distribution.
And if `IMBALANCE_THRESHOLD` is already surpassed by the initial distribution rebalance should be allowed.

src/structure_new/src/4C_structure_new_timint_base.cpp
LaSi5002: Not needed anymore, thanks for noticing. 58ebf926bb4c37b136c0d119c91da0a3141b14ef

tests/input_files/solid_dynamic_rebalance_simple_5x2x1.4C.yaml
LaSi5002: 7daa4fce15266da520dde14838ad61cce6c16a18

src/structure_new/src/utils/4C_structure_new_monitor_dbc.cpp
LaSi5002: Yes, creating a helper function is cleaner here. bb45420bb10a9598f85658710ee06430aea5cbf9

src/structure_new/src/input/4C_structure_new_input.cpp
LaSi5002: `Core::Rebalance::RebalanceParameters` is a struct for the lower level preexisting redistrubution mechanism, not for dynamic rebalance parameters. For the higher level dynamic rebalance there is `DynamicRebalanceConfig`. I used the basic `group` with later parsing because that seemed to be the convention in the file, but I can wire it differently if you want.
Also I feel like it is more intuitive, if real data is only carried in parameters, so that groups in yaml are only organisational. But I can see how an `ENABLED` seems a bit redundant.

src/structure_new/src/4C_structure_new_timint_base.cpp
rjoussen: Okay, that makes sense, thank you! Then I guess I am only confused with the that it is divided by 2 (even though that shouldn't matter of course)

src/structure_new/src/input/4C_structure_new_input.cpp
rjoussen: Yes, I meant `DynamicRebalanceConfig`. I think I would prefer if the input group is that struct directly, for two reasons: 1. you can store the parameters directly in the struct (via `.store=in_struct())`, which means you don't need to "remember" the exact names of the parameters when you parse them, you just parse the entire struct at once. This removes the second source of thruth. 2. you can make this an `std::optional<DynamicRebalanceConfig>` with default `nullopt`. `ENABLED` is not needed then, but rebalancing is just turned on when the group is not null. I assume the main reason why it seems unconventional in this file is that most of this input is really old.

I would probably even be in favor of putting this group directly into STRUCTURAL `DYNAMIC`, not an additional sublist, since it is easier to find this (really cool!) feature then.

src/structure_new/src/4C_structure_new_timint_base.cpp
rjoussen: There are still some changes to `post_output` related code remaining now. Can you remove those? Specifically in `src/adapter/4C_adapter_str_timeada.hpp` and `src/adapter/4C_adapter_str_timeloop.hpp`
This way it won't have any conflicts with #2125

src/structure_new/src/4C_structure_new_timint_basedataglobalstate.hpp
```diff
@@ -109,6 +114,11 @@ namespace Solid
       /// setup of the new class variables
       virtual void setup();
 
+      /// Redistribute the discretization and remap all global state vectors to the new maps.
```
dragos-ana: Here you could also add some documentation for the parameter `use_eval_time_weights`

src/structure_new/src/4C_structure_new_timint_basedatasdyn.cpp
```diff
@@ -20,6 +20,23 @@
 
 FOUR_C_NAMESPACE_OPEN
 
+Solid::TimeInt::DynamicRebalanceConfig Solid::TimeInt::parse_dynamic_rebalance_config(
```
dragos-ana: Not needed anymore if you specify the input group as proposed by @rjoussen.

src/structure_new/src/4C_structure_new_timint_base.hpp
```diff
@@ -70,9 +73,21 @@ namespace Solid
...
+      /// Redistribute a pure structure discretization using measured evaluation times.
+      bool perform_dynamic_rebalance();
+
+      /// Compute rolling imbalance and trigger redistribution when needed.
+      void maybe_perform_dynamic_rebalance();
+
       /// tests if there are more time steps to do
       [[nodiscard]] bool not_finished() const override;
 
+      const Solid::TimeInt::DynamicRebalanceConfig& get_dynamic_rebalance_config() const
```
dragos-ana:
```suggestion
      [[nodiscard]] const Solid::TimeInt::DynamicRebalanceConfig& get_dynamic_rebalance_config() const
```

src/structure_new/src/4C_structure_new_timint_base.hpp
```diff
@@ -866,6 +885,9 @@ namespace Solid
 
       /// pointer to the dirichlet boundary condition handler
       std::shared_ptr<Solid::Dbc> dbc_ptr_;
+
+      std::deque<double> dynamic_rebalance_imbalance_history_;
```
dragos-ana, line 883: Some documentation would be nice here particularly for the history deque.

src/structure_new/src/4C_structure_new_timint_base.cpp
```diff
@@ -120,6 +129,82 @@ void Solid::TimeInt::Base::post_setup()
   int_ptr_->post_setup();
 }
 
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
+void Solid::TimeInt::Base::maybe_perform_dynamic_rebalance()
+{
+  const auto& rebalance_config = datasdyn_->get_dynamic_rebalance_config();
+
+  const std::vector<double> rank_eval_times =
```
dragos-ana, line 139: Having two functions `maybe_do_this` and `do_this` is a little strange in my opinion.
I would be in favor of having two separate functions: one that verifies whether dynamic rebalancing should be performed, and one that really performs it  (e.g. `verify_dynamic_rebalance_conditions` and `perform_dynamic_rebalance`). Both of them would be called sequentially in the timeloop, which is slightly more explicit in my opinion.

src/structure_new/src/4C_structure_new_integrator.hpp
```diff
@@ -380,6 +383,9 @@ namespace Solid
     //! reset the time step dependent parameters for the element evaluation
     virtual void reset_eval_params() {};
 
+    //! Refresh time-integrator-owned cached vectors and helpers after redistribution.
```
dragos-ana, line 383: Is there any way we could make the naming of this function more explicit? E.g., `remap_vectors_and_helpers` or something like this? To better differentiate it from `remap_after_redistribution`. Which also has `Restore integrator state` in its documentation, so the functions seem redundant from outside although they are not.

src/structure_new/src/utils/4C_structure_new_monitor_dbc.hpp
```diff
@@ -68,6 +68,8 @@ namespace Solid
         Core::FE::Discretization& discret, Solid::TimeInt::BaseDataGlobalState& gstate,
         Solid::Dbc& dbc);
 
+    void remap_reaction_maps();
```
dragos-ana, line 73: Some documentation would be nice, even if very short. At least referencing that this is required for dynamic rebalance.

dragos-ana (commented):
Nice work, @LaSi5002! I added some small comments / suggestions.

src/structure_new/src/4C_structure_new_timint_base.cpp
dragos-ana: It still requires more thought than necessary. Not sure about the optimal strategy, but I think that even introducing a private boolean variable tracking whether we are in the initial stage (before the minimum number of cooldown steps) is reached would be more explicit.

rjoussen: I think we could initialize it with 0 and then where it's actually being used do this:
```cpp
if (current_step - last_dynamic_rebalance_step_ < cooldown_steps
    and not (current_step - problem_->restart() < cooldown_steps)) return;
```
instead of just doing the first check.

This would also mean that a restart is allowed to rebalance immediately (which I think makes just as much sense then, since we redistribute at restart).
Then the tracking variable could also be `steps_since_last_dynamic_rebalance_`, but that doesn't matter that much.

LaSi5002: I think this is a good idea. I didn't think of restarts but it makes sense that a standard (re)distribution happens at restart. Which we have to take into account of course.

LaSi5002: 1af87686d48f620dad4d31e33eb4557df5be4308

src/structure_new/src/4C_structure_new_timint_base.cpp
LaSi5002: Alright! Done in 9ddbd9715b3ac202ad28902f705ad9a9b4080f3d

src/structure_new/src/4C_structure_new_timint_base.cpp
rjoussen: Thanks. The remaining conflict in the timeloop should be easy to fix now as well I think. Sorry for interfering with your, thanks to it I just noticed that post_output is empty anyway so I removed right away :)

src/structure_new/src/input/4C_structure_new_input.cpp
LaSi5002: Alright. Moved `DYNAMIC REBALANCE` into `STRUCTURAL DYNAMIC` and switched it to `std::optional<DynamicRebalanceConfig>` and `group<DynamicRebalanceConfig>` in 4468f6b421df45555c3ea561c93e252d3c075060
`ENABLED` is now gone and replaced by group presence.

src/core/rebalance/src/4C_rebalance.hpp
LaSi5002: 3c0ecc9fa2fde106be93d2cd88ca8a265244c0c8

src/structure_new/src/4C_structure_new_timint_basedataglobalstate.hpp
LaSi5002: c6f9ae43c4524fd366e875aa0feb848fab8710c1

src/structure_new/src/4C_structure_new_timint_basedatasdyn.cpp
LaSi5002: Removed in 4468f6b421df45555c3ea561c93e252d3c075060

src/structure_new/src/4C_structure_new_timint_base.hpp
LaSi5002: 854b9dc1799daef858c072a92803f32dfb71bd18

src/structure_new/src/4C_structure_new_timint_base.cpp
LaSi5002: Yes, I can see how you like this option more. Changed it in 8ed1790cbf0330e42bb840a0c5dbbec5a698f211

src/structure_new/src/4C_structure_new_integrator.hpp
LaSi5002: Yes, that makes sense. 4bb09f620da8123bc45297bd25c2f83b17d6e7b7

src/structure_new/src/utils/4C_structure_new_monitor_dbc.hpp
LaSi5002: 78f2b72ee331390a373eb65e8880638955af98cc

src/structure_new/src/4C_structure_new_timint_base.cpp
LaSi5002: No problem :) Rebased onto main and resolved conflicts.

src/structure_new/src/input/4C_structure_new_input.cpp
rjoussen: Looks really nice now!

LaSi5002: After the commits, two nested parallelism tests failed. The reason was, that because the mesh partitioning parameters were set as required by default, if a `STRUCTURAL DYNAMIC` parameter was accessed, without existing in a teuchos parameter list, no default value could be created (which led to the faliure). Even though the `DYNAMIC REBALANCE` group was optional. After also setting the mesh partitioning to unrequired, rebalance was triggered in many tests because a `DYNAMIC REBALANCE` group was created automatically when accessing other `STRUCTURAL DYNAMIC` parameters. 
I went back to the `ENABLED` option because it is safe and doesn't allow for triggering by default value construction when tests access the `STRUCTURAL DYNAMIC` group.

LaSi5002: Another option would be to check whether any `DYNAMIC REBALANCE` parameter is not its default value, but that would require to have prohibited values like -1 as default values which would defeat the purpose of a default value, so i would still advocate for keeping `ENABLED`

src/core/fem/src/discretization/4C_fem_discretization.hpp
```diff
@@ -548,6 +548,9 @@ namespace Core::FE
     //! Get per-rank evaluation times on rank 0, summed over all local column elements
     std::vector<double> get_rank_eval_times_on_root() const;
 
+    //! Get per-rank evaluation times on all ranks, summed over all local column elements
+    std::vector<double> get_rank_eval_times() const;
```
c-p-schmidt: Adding this would be nice.
```suggestion
    [[nodiscard]] std::vector<double> get_rank_eval_times() const;
```

src/core/rebalance/src/4C_rebalance.cpp
```diff
@@ -138,8 +146,14 @@ void Core::Rebalance::rebalance_discretization(Core::FE::Discretization& discret
...
-    std::tie(rowmap, colmap) = do_rebalance_discretization(
-        *graph, discretization, rebalanceMethod, rebalanceParams, parameters, comm);
+    const std::optional<Core::Rebalance::PartitionWeights> effective_partition_weights =
+        use_eval_time_weights
+            ? std::make_optional(Core::Rebalance::build_eval_time_partition_weights(
+                  discretization, *graph, parameters.edge_weight_multiplier))
+            : std::nullopt;
+    std::tie(rowmap, colmap) = do_rebalance_discretization(*graph, discretization, rebalanceMethod,
+        rebalanceParams, parameters, comm,
+        effective_partition_weights.has_value() ? &effective_partition_weights.value() : nullptr);
```
c-p-schmidt: Instead of converting this to a pointer or a null pointer, the method could just take a std::optional to indicate that this parameter is optional. And then doing the right thing internally.
But just a thought.

src/structure_new/src/4C_structure_new_timint_base.cpp
```diff
@@ -120,6 +128,91 @@ void Solid::TimeInt::Base::post_setup()
...
+ *----------------------------------------------------------------------------*/
+bool Solid::TimeInt::Base::should_perform_dynamic_rebalance()
+{
+  const auto& rebalance_config_opt = datasdyn_->get_dynamic_rebalance_config();
+  if (!rebalance_config_opt.has_value()) return false;
+  const auto& rebalance_config = *rebalance_config_opt;
+
+  const std::vector<double> rank_eval_times =
+      dataglobalstate_->get_discret()->get_rank_eval_times();
+  if (rank_eval_times.empty()) return false;
```
c-p-schmidt, line 141: Isn't that rather an error if dynamic rebalancing is activated but no evaluation times are available? Otherwise, don't you just neglect dynamic rebalancing?

c-p-schmidt (commented):
Looks good to me. Some minor comments I added. Could you please squash your changes and push so that we can hopefully merge soon?

tests/list_of_tests.cmake
```diff
@@ -2028,6 +2028,23 @@ four_c_test(TEST_FILE solid_nodal_fiber.4C.yaml NP 2)
...
+  TOL_R 1e-12
+  TOL_A 1e-12
+  )
+__four_c_test_timings(
+  BASED_ON ${solid_dynamic_rebalance_simple_5x2x1_rebal}
+  TIMINGS_FILE xxx-timings.yaml
+  TIMERS Rebalance::rebalance_graph Rebalance::rebalance_node_maps
+  MIN_CALLS 6 6
+  MAX_CALLS 6 6
+  )
```
rjoussen: Can you put these two tests in one line? SUch that it is the same as all other tests.

tests/input_files/solid_dynamic_rebalance_5x2x1_no_rebal.4C.yaml
rjoussen, line 1: Maybe rename this to to solid `solid_dynamic_rebalance_simple_5x2x1_no_rebal.4C.yaml`? Or add a description in the title. Also, what does the `simple` stand for?

src/structure_new/src/4C_structure_new_timint_basedatasdyn.hpp
```diff
@@ -511,6 +523,12 @@ namespace Solid
         check_init_setup();
         return start_func_no_;
       }
+
+      const std::optional<DynamicRebalanceConfig>& get_dynamic_rebalance_config() const
```
rjoussen:
```suggestion
      [[nodiscard]] const std::optional<DynamicRebalanceConfig>& get_dynamic_rebalance_config() const
```

src/structure_new/src/4C_structure_new_timint_basedataglobalstate.cpp
```diff
@@ -221,6 +225,98 @@ void Solid::TimeInt::BaseDataGlobalState::setup()
...
+{
+  check_init_setup();
+
+  FOUR_C_ASSERT(discret_->element_row_map() != nullptr,
+      "The discretization has to be fill_complete() before redistribution.");
+
+  struct RemappedVectorState
+  {
+    std::shared_ptr<Core::LinAlg::Vector<double>>* state;
+  };
```
rjoussen:
```suggestion
  using RemappedVectorState = std::shared_ptr<Core::LinAlg::Vector<double>>*;
```

src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_nox.hpp
```diff
@@ -67,6 +67,8 @@ namespace Solid
         //! derived from the base class
         void reset() override;
 
+        void refresh_after_redistribution() override { rebuild_problem_state(); }
```
rjoussen, line 70: I think the code inside `rebuild_problem_state()` could be directly inside the `refresh_after_redistribution()` implementation

src/adapter/4C_adapter_str_timeloop.cpp
```diff
@@ -62,6 +63,8 @@ int Adapter::StructureTimeLoop::integrate()
 
       // write output
       output();
+      if (auto* timint = dynamic_cast<Solid::TimeInt::Base*>(structure_.get()))
```
rjoussen, line 65: I think a small comment saying that dynamic rebalance is only implemented for the new time integration would be helpful here.

rjoussen (commented):
Only a few minor things. Overall, it looks really good now!

src/core/fem/src/discretization/4C_fem_discretization.hpp
LaSi5002: Yes, that makes sense.

src/core/rebalance/src/4C_rebalance.cpp
LaSi5002: Yeah, that's a bit cleaner. Done in 4148883ca91f4cc56bd47f5a043affbfdc38f8c1

src/structure_new/src/4C_structure_new_timint_base.cpp
LaSi5002: The history gets cleared after each rebalance.

tests/list_of_tests.cmake
LaSi5002: Done in 0cc6cf5566ad5827e21f2670840a68b3eb22d569

tests/input_files/solid_dynamic_rebalance_5x2x1_no_rebal.4C.yaml
LaSi5002: Done in 5d625a49698dcb16dbf911b76049cd69454f72c7

src/adapter/4C_adapter_str_timeloop.cpp
LaSi5002: Done in dfed7c2b674217fc0886b690cadbe11d7a135c4c

src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_nox.hpp
LaSi5002: `rebuild_problem_state` also gets called inside the constructor, so replacing it by `refresh_after_redistribution` would mean calling `refresh_after_redistribution` not after redistribution.

rjoussen: Okay, I probably missed that.

rjoussen (approved):
Looks great! Please squash the commits before merging.

c-p-schmidt (approved):

PR #2109, opened 2026-07-02
A hybrid Ogden-type material model with compression-tension asymmetry control

lauraengelhardt:
This PR implements the hyperelastic, isotropic material model proposed in:

> Kevin M. Moerman, Ciaran K. Simms, Thomas Nagel
> Control of tension–compression asymmetry in Ogden hyperelasticity with application to soft tissue modelling
> Journal of the Mechanical Behavior of Biomedical Materials,
> Volume 56, 2016, Pages 218–228,
> ISSN 1751-6161,
> https://doi.org/10.1016/j.jmbbm.2015.11.027
> https://www.sciencedirect.com/science/article/pii/S1751616115004452?via=ihub

It is a simple Ogden material model with the nice feature that you can control the stiffness, nonlinearity, and the degree of compression–tension asymmetry (CTA) all individually. In a classical Ogden material model, CTA is coupled to the nonlinearity parameter and cannot be controlled individually. 

Below you can see the stress-stretch curves for uniaxial loading for different CTA parameters q:
grafik

Assuming λ ≥ 1, you can see that 
- For q → 0: σ(1/λ) > σ(λ)
- For q → 1: σ(1/λ) < σ(λ)
- For q = 0.5: σ(1/λ) = σ(λ)


src/mat/4C_mat_ogden_cta.cpp
```diff
@@ -0,0 +1,193 @@
...
+      {
+        cmat +=
+            dSdlambda[i] * Core::LinAlg::assume_symmetry(Core::LinAlg::dyadic(
+                               Core::LinAlg::dyadic(N[i], N[i]), Core::LinAlg::dyadic(N[j], N[j])));
+      }
+      else
+      {
+        const double fac = (std::abs(lambda[i] - lambda[j]) < 1e-12)
+                               ? 0.5 * (dSdlambda[j] - dSdlambda[i])
+                               : (S[j] - S[i]) / (lambda[j] * lambda[j] - lambda[i] * lambda[i]);
```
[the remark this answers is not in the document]
lauraengelhardt: I will double check this.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -834,6 +834,28 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                         "multi-scale approach"});
   }
 
+  /*----------------------------------------------------------------------*/
+  // Hyperelastic Ogden material with compression-tension asymmetry control
+  {
+    using namespace Core::IO::InputSpecBuilders::Validators;
+
+    known_materials[Core::Materials::m_ogden_cta] = group("MAT_Ogden_CTA",
```
```suggestion
    known_materials[Core::Materials::m_ogden_cta] = group("MAT_Ogden_CompressionTensionAssymmetry",
```
maybe like this? Then its easier to understand what this material does when just scanning over them quickly.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -834,6 +834,28 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                "M", {.description = "nonlinearity parameter", .validator = positive<double>()}),
+            parameter<double>(
+                "Q", {.description = "compression-tension asymmetry control parameter",
+                         .validator = in_range(0.0, 1.0)}),
+            parameter<double>("KAPPA",
+                {.description = "incompressibility parameter", .validator = positive<double>()}),
+            parameter<double>(
+                "DENS", {.description = "density", .validator = positive_or_zero<double>()}),
+        },
+        {.description = "Hyperelastic Ogden material with compression-tension asymmetry control"});
```
rjoussen: Would it be possible to include the model equation in the description and/or reference the paper? This way it shows up in the documentation and people dont need to look into the code. This would also make the parameter names more clear

src/mat/4C_mat_ogden_tca.hpp
```diff
@@ -0,0 +1,153 @@
...
+   *
+   * The second Piola--Kirchhoff stress is computed as
+   * \f[ \mathbf{S} = \sum_{i=1}^{3}
+   * \frac{c}{m} (q \lambda_i^{m-2} - (1-q) \lambda_i^{-m-2}) \mathbf{N}_i \otimes \mathbf{N}_i
+   * + [\kappa J(J-1) + \frac{c}{m}(1-2q)] \mathbf{C}^{-1} \f]
+   *
+   * where \f$J=\det\mathbf{F}=\sqrt{\det\mathbf{C}}\f$,
+   * \f$\lambda_i\f$ are the principal stretches, and
+   * \f$\mathbf{N}_i\f$ are the corresponding principal directions of \f$\mathbf{C}\f$.
+   *
```
rjoussen, line 86: I really like the inclusion of proper equations!
This could go though (additionally or exclusively) to the material input parameter description, this way its accesible in the sphinx documenation as well.

src/mat/4C_mat_ogden_tca.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+  // parameters
+  const double c = params_->c_;
+  const double m = params_->m_;
+  const double q = params_->q_;
+  const double kappa = params_->kappa_;
+
+  // right Cauchy Green tensor C = F^{T} F
+  const Core::LinAlg::SymmetricTensor<double, 3, 3> C =
+      Core::LinAlg::assume_symmetry(Core::LinAlg::transpose(*defgrad) * *defgrad);
```
rjoussen, line 124:
```suggestion
      assume_symmetry(transpose(*defgrad) * *defgrad);
```
Idk if you are aware of that, but you can do this also and use argument dependent lookup. It looks cleaner in my opinion than having Core::LinAlg everywhere, but that's totally your decision.

src/mat/4C_mat_ogden_cta.cpp
```diff
@@ -0,0 +1,210 @@
...
+  for (int i = 0; i < 3; ++i)
+  {
+    for (int j = 0; j < 3; ++j)
+    {
+      if (i != j)
+      {
+        const auto dyadic_Ni_Nj = Core::LinAlg::dyadic(N[i], N[j]);
+        const auto dyadic_Nj_Ni = Core::LinAlg::dyadic(N[j], N[i]);
+
+        const double fac = (std::abs(lambda[i] - lambda[j]) < 1e-12)
```
rjoussen: could this be numeric limits? Or what are you trying to achieve here?

tests/list_of_tests.cmake
```diff
@@ -915,6 +915,8 @@ four_c_test(TEST_FILE mat_muscle_giantesio_hex.4C.yaml NP 2 RETURN_AS current)
 __four_c_test_restart(BASED_ON ${current} SAME_FILE NP 2 RESTART_STEP 140)
 four_c_test(TEST_FILE mat_muscle_weickenmeier_hex.4C.yaml NP 2 RETURN_AS current)
 __four_c_test_restart(BASED_ON ${current} SAME_FILE NP 2 RESTART_STEP 140)
+four_c_test(TEST_FILE mat_ogden_cta_hex.4C.yaml NP 2 RETURN_AS current)
+__four_c_test_restart(BASED_ON ${current} SAME_FILE NP 2 RESTART_STEP 25)
```
rjoussen: They need MUMPS as a dependeny

tests/input_files/mat_ogden_tca_hex.4C.yaml
```diff
@@ -0,0 +1,281 @@
...
+IO:
+  STRUCT_STRESS: "2PK"
+  STRUCT_STRAIN: "GL"
+  VERBOSITY: Standard
+IO/RUNTIME VTK OUTPUT:
+  INTERVAL_STEPS: 1
+IO/RUNTIME VTK OUTPUT/STRUCTURE:
+  OUTPUT_STRUCTURE: true
+  DISPLACEMENT: true
+  STRESS_STRAIN: true
```
rjoussen, line 18: VTK output is not needed here.

src/mat/4C_mat_ogden_tca.hpp
c-p-schmidt: I think it should be here in the header, but I agree it might also help users, so it's valuable information to include in the input parameter description.

tests/input_files/mat_ogden_tca_hex.4C.yaml
lauraengelhardt: Thought so as well but without the result tests cannot test for the stresses.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
lauraengelhardt: I would like to keep the shorter name. The abbreviation becomes clear in the first sentence of the description.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
lauraengelhardt: I included it, good idea!

src/mat/4C_mat_ogden_tca.cpp
lauraengelhardt: I didn't know. If I find the time I'll change that in one go for multiple materials :)

src/mat/4C_mat_ogden_cta.cpp
lauraengelhardt: I switch to a limit expression when the eigenvalues become nearly equal. I'm just using the same tolerance as the already existing Ogden material for consistency. I made it a bit more clear with the variable naming now.

lauraengelhardt:
Thanks for the review @rjoussen and @c-p-schmidt!
Took some time due to the WCCM to work on this again ;)

c-p-schmidt (dismissed):

src/mat/4C_mat_ogden_cta.cpp
```diff
@@ -0,0 +1,211 @@
...
+  // formula according to Holzapfel Eq. (6.52) and Eq. (6.180), respectively:
+  // stress = sum_i [S_i * N_i dyadic N_i] with S_i = 1/lambda_i * dPsi/dlambda_i
+  // cmat = cmat_1 + cmat_2 with
+  // cmat_1 = sum_i sum_j [1/lambda_j * dS_i/dlambda_j N_i dyadic N_i dyadic N_j dyadic N_j]
+  // cmat_2 = sum_i sum_j i!=j [(S_j - S_i)/(lambda_j^2 - lambda_i^2) (N_i dyadic N_j dyadic N_i
+  // dyadic N_j + N_i dyadic N_j dyadic N_j dyadic N_i)]
+  std::array<double, 3> S{0}, dSdlambda{0};
+  for (int i = 0; i < 3; ++i)
+  {
+    const auto dyadic_Ni_Ni = Core::LinAlg::dyadic(N[i], N[i]);
```
ischeider: There is a function self_dyadic. which provides a symmetric tensor automatically.

src/mat/4C_mat_ogden_cta.cpp
```diff
@@ -0,0 +1,211 @@
...
+  // lambda_i according to to Eq. (6.190): dS_j/d(lambda_j^2) - dS_i/d(lambda_j^2) =
+  // dS_j/d(lambda_j^2) = 1/(2*lambda_j) * dS_j/dlambda_j
+  for (int i = 0; i < 3; ++i)
+  {
+    for (int j = 0; j < 3; ++j)
+    {
+      if (i != j)
+      {
+        const auto dyadic_Ni_Nj = Core::LinAlg::dyadic(N[i], N[j]);
+        const auto dyadic_Nj_Ni = Core::LinAlg::dyadic(N[j], N[i]);
```
ischeider: Not sure, but I guess a transpose of dyadic_Ni_Nj is more efficient than doing the dyadic operation again.

```suggestion
        const auto dyadic_Nj_Ni = Core::LinAlg::transpose(dyadic_Ni_Nj);
```

ischeider (dismissed):
Nice work, I like the asymmetry model.
The only problem I have personally is that I heard the term tension-compression asymmetry often, but never compression-tension asymmetry. Maybe it's because of my bubble, therefore I asked the AI, but  it confirmed my impression (I asked unbiased):

I would prefer **"tension-compression asymmetry"**. This is the term most commonly used in the constitutive modeling and continuum mechanics literature for materials that exhibit different responses in tension and compression. It appears consistently in publications on hyperelasticity, plasticity, damage mechanics, fracture, and biological tissues.

While "compression-tension asymmetry" is occasionally encountered, it is much less common and is not the standard terminology. For consistency with the scientific literature, I recommend using:

    Tension-compression asymmetry

Examples:
- tension-compression asymmetric material
- tension-compression asymmetric response
- tension-compression asymmetry model

rjoussen (approved):

ischeider (approved):
Thank you very much for considering my comment!

PR #2117, opened 2026-07-08
[Particle] Do not use ParticleObjShrdPtr for commicating particle data

vovannikov:
This PR is the continuation of #2074. It further improves the performance of the SPH+PD benchmark analyzed in that previous PR.

Figure_1
 
### Goal

The particle refresh routines (`refresh_particles` and `refresh_particles_of_specific_states_and_types`) previously communicated data by wrapping every particle in a heap-allocated `ParticleObject` (`ParticleObjShrdPtr`), packing it, then reconstructing objects on the receiving side before copying their states into containers. This adds needless allocation, virtual dispatch through the `ParObject` factory, and a temporary object round-trip on a hot communication path that runs every time step.

The goal of this PR is to **eliminate `ParticleObject` creation entirely from the refresh path**, packing and unpacking particle state directly into/out of MPI byte buffers.

### How it is achieved

- **Direct packing into send buffers.** `determine_particles_to_be_refreshed` and `determine_specific_states_of_particles_of_specific_types_to_be_refreshed` are replaced by `pack_particles_to_be_refreshed` and `pack_specific_states_of_particles_to_be_refreshed`. These pack a small header (particle type + ghosted container index) followed by the particle states straight into `std::map<int, std::vector<char>>` send buffers, skipping `ParticleObject` construction.

- **State reuse per particle.** Each particle's states are packed once and appended for every target rank, avoiding repeated serialization when a particle is ghosted on multiple processors.

- **Direct unpacking into containers.** The new `communicate_refreshed_particles(sdata)` exchanges the pre-packed buffers over the cached send/receive graph (`immediate_send_recv_known_procs`) and unpacks type, index, and states directly into the ghosted containers via `replace_particle` — no factory, no `dynamic_pointer_cast`, no intermediate object list.

- **Removed the intermediate insert step.** `insert_refreshed_particles` is deleted, since received data is now written to containers inline during unpacking.

- **`communicate_particles` restored to its original form.** The refresh path no longer piggybacks on `communicate_particles`, so its optional `send_to_procs`/`receive_from_procs` overload parameters are removed, keeping that method focused on the general (collective) particle exchange.

### Result

- Fewer heap allocations and no virtual `ParObject` factory dispatch on the per-timestep refresh path.
- Simpler data flow: pack -> communicate -> unpack-into-container, with one fewer intermediate stage.
- No behavioral change to what data is exchanged; the cached communication graph (`cached_procs_send_ghost_data_to_` / `cached_procs_receive_ghost_data_from_`) is reused as before.

### DIFF
--- a/src/particle/src/engine/4C_particle_engine.cpp
+++ b/src/particle/src/engine/4C_particle_engine.cpp
@@ -355,19 +355,12 @@ void Particle::ParticleEngine::refresh_particles() const
 {
   TEUCHOS_FUNC_TIME_MONITOR("Particle::ParticleEngine::RefreshParticles");
 
-  std::vector<std::vector<ParticleObjShrdPtr>> particlestosend(
-      Core::Communication::num_mpi_ranks(comm_));
-  std::vector<std::vector<std::pair<int, ParticleObjShrdPtr>>> particlestoinsert(typevectorsize_);
+  // pack particles to be refreshed directly into send buffers
+  std::map<int, std::vector<char>> sdata;
+  pack_particles_to_be_refreshed(sdata);
 
-  // determine particles that need to be refreshed
-  determine_particles_to_be_refreshed(particlestosend);
-
-  // communicate particles using cached communication graph
-  communicate_particles(particlestosend, particlestoinsert, &cached_procs_send_ghost_data_to_,
-      &cached_procs_receive_ghost_data_from_);
-
-  // insert refreshed particles received from other processors
-  insert_refreshed_particles(particlestoinsert);
+  // communicate and unpack refreshed particles directly into containers
+  communicate_refreshed_particles(sdata);
 }
 
 void Particle::ParticleEngine::refresh_particles_of_specific_states_and_types(
@@ -376,20 +369,12 @@ void Particle::ParticleEngine::refresh_particles_of_specific_states_and_types(
   TEUCHOS_FUNC_TIME_MONITOR(
       "Particle::ParticleEngine::refresh_particles_of_specific_states_and_types");
 
-  std::vector<std::vector<ParticleObjShrdPtr>> particlestosend(
-      Core::Communication::num_mpi_ranks(comm_));
-  std::vector<std::vector<std::pair<int, ParticleObjShrdPtr>>> particlestoinsert(typevectorsize_);
+  // pack specific states of particles directly into send buffers
+  std::map<int, std::vector<char>> sdata;
+  pack_specific_states_of_particles_to_be_refreshed(particlestatestotypes, sdata);
 
-  // determine particles that need to be refreshed
-  determine_specific_states_of_particles_of_specific_types_to_be_refreshed(
-      particlestatestotypes, particlestosend);
-
-  // communicate particles using cached communication graph
-  communicate_particles(particlestosend, particlestoinsert, &cached_procs_send_ghost_data_to_,
-      &cached_procs_receive_ghost_data_from_);
-
-  // insert refreshed particles received from other processors
-  insert_refreshed_particles(particlestoinsert);
+  // communicate and unpack refreshed particles directly into containers
+  communicate_refreshed_particles(sdata);
 }
 
 void Particle::ParticleEngine::dynamic_load_balancing()
@@ -1583,8 +1568,8 @@ void Particle::ParticleEngine::determine_particles_to_be_ghosted(
   }
 }
 
-void Particle::ParticleEngine::determine_particles_to_be_refreshed(
-    std::vector<std::vector<ParticleObjShrdPtr>>& particlestosend) const
+void Particle::ParticleEngine::pack_particles_to_be_refreshed(
+    std::map<int, std::vector<char>>& sdata) const
 {
   // safety check
   if (not validdirectghosting_) FOUR_C_THROW("invalid direct ghosting!");
@@ -1603,9 +1588,14 @@ void Particle::ParticleEngine::determine_particles_to_be_refreshed(
     {
       int ownedindex = indexIt.first;
 
+      // get particle states once
       int globalid(0);
       ParticleStates states;
       container->get_particle(ownedindex, globalid, states);
+
+      // pre-pack states data (reused for all targets of this particle)
+      Core::Communication::PackBuffer statesdata;
+      add_to_pack(statesdata, states);
 
       // iterate over target processors
       for (const auto& targetIt : indexIt.second)
@@ -1613,18 +1603,23 @@ void Particle::ParticleEngine::determine_particles_to_be_refreshed(
         int sendtoproc = targetIt.first;
         int ghostedindex = targetIt.second;
 
-        // append particle to be send
-        particlestosend[sendtoproc].emplace_back(
-            std::make_shared<ParticleObject>(type, -1, states, -1, ghostedindex));
+        // pack type and ghosted index header
+        Core::Communication::PackBuffer header;
+        add_to_pack(header, static_cast<int>(type));
+        add_to_pack(header, ghostedindex);
+
+        // append header + pre-packed states to send buffer
+        auto& buf = sdata[sendtoproc];
+        buf.insert(buf.end(), header().begin(), header().end());
+        buf.insert(buf.end(), statesdata().begin(), statesdata().end());
       }
     }
   }
 }
 
-void Particle::ParticleEngine::
-    determine_specific_states_of_particles_of_specific_types_to_be_refreshed(
-        const StatesOfTypesToRefresh& particlestatestotypes,
-        std::vector<std::vector<ParticleObjShrdPtr>>& particlestosend) const
+void Particle::ParticleEngine::pack_specific_states_of_particles_to_be_refreshed(
+    const StatesOfTypesToRefresh& particlestatestotypes,
+    std::map<int, std::vector<char>>& sdata) const
 {
   // safety check
   if (not validdirectghosting_) FOUR_C_THROW("invalid direct ghosting!");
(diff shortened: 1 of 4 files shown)


src/particle/src/engine/4C_particle_engine.hpp
```diff
@@ -568,47 +570,55 @@ namespace Particle
         std::vector<std::vector<ParticleObjShrdPtr>>& particlestosend) const;
 
     /*!
-     * \brief determine particles that need to be refreshed
+     * \brief pack particles to be refreshed into send buffers
      *
-     * Determine particles that need to be refreshed on other processors based on the direct
-     * ghosting targets map.
+     * Pack all states of particles that need to be refreshed on other processors directly into
+     * send buffers, bypassing ParticleObject creation.
```
georghammerl, line 576:
```suggestion
     * send buffers.
```

src/particle/src/engine/4C_particle_engine.hpp
```diff
@@ -568,47 +570,55 @@ namespace Particle
...
-    void determine_particles_to_be_refreshed(
-        std::vector<std::vector<ParticleObjShrdPtr>>& particlestosend) const;
+    void pack_particles_to_be_refreshed(std::map<int, std::vector<char>>& sdata) const;
 
     /*!
-     * \brief determine specific states of particles of specific type that need to be refreshed
+     * \brief pack specific states of particles of specific types into send buffers
+     *
+     * Pack the information of specific states of particles of specific types directly into
+     * send buffers, bypassing ParticleObject creation.
```
georghammerl, line 587:
```suggestion
     * send buffers.
```

src/particle/src/engine/4C_particle_engine.hpp
```diff
@@ -568,47 +570,55 @@ namespace Particle
...
-        const std::set<int>* send_to_procs = nullptr,
-        const std::set<int>* receive_from_procs = nullptr) const;
+        std::vector<std::vector<std::pair<int, ParticleObjShrdPtr>>>& particlestoreceive) const;
+
+    /*!
+     * \brief communicate refreshed particles using cached communication graph
+     *
+     * Uses the pre-computed cached_procs_send_ghost_data_to_ and
+     * cached_procs_receive_ghost_data_from_ to exchange refreshed particle data. Unpacks received
+     * data directly into ghosted particle containers without ParticleObject creation.
```
georghammerl, line 630:
```suggestion
     * data directly into ghosted particle containers.
```

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1603,28 +1588,38 @@ void Particle::ParticleEngine::determine_particles_to_be_refreshed(
...
       int globalid(0);
       ParticleStates states;
       container->get_particle(ownedindex, globalid, states);
 
+      // pre-pack states data (reused for all targets of this particle)
+      Core::Communication::PackBuffer statesdata;
+      add_to_pack(statesdata, states);
+
       // iterate over target processors
       for (const auto& targetIt : indexIt.second)
```
georghammerl: What about a structured binding here:

```suggestion
      for (const auto& [sendtoproc, ghostedindex] : indexIt.second)
```

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -355,19 +355,12 @@ void Particle::ParticleEngine::refresh_particles() const
...
-
-  // determine particles that need to be refreshed
-  determine_particles_to_be_refreshed(particlestosend);
-
-  // communicate particles using cached communication graph
-  communicate_particles(particlestosend, particlestoinsert, &cached_procs_send_ghost_data_to_,
-      &cached_procs_receive_ghost_data_from_);
+  // pack particles to be refreshed directly into send buffers
+  std::map<int, std::vector<char>> sdata;
+  pack_particles_to_be_refreshed(sdata);
```
ppraegla: Does it make sense to return sdata instead of passing it as an argument? IMO, this makes it more explicit that sdata is an output.

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -376,20 +369,12 @@ void Particle::ParticleEngine::refresh_particles_of_specific_states_and_types(
...
-  // determine particles that need to be refreshed
-  determine_specific_states_of_particles_of_specific_types_to_be_refreshed(
-      particlestatestotypes, particlestosend);
-
-  // communicate particles using cached communication graph
-  communicate_particles(particlestosend, particlestoinsert, &cached_procs_send_ghost_data_to_,
-      &cached_procs_receive_ghost_data_from_);
+  // pack specific states of particles directly into send buffers
+  std::map<int, std::vector<char>> sdata;
+  pack_specific_states_of_particles_to_be_refreshed(particlestatestotypes, sdata);
```
ppraegla: Same as above. Return sdata?

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1583,8 +1568,8 @@ void Particle::ParticleEngine::determine_particles_to_be_ghosted(
   }
 }
 
-void Particle::ParticleEngine::determine_particles_to_be_refreshed(
-    std::vector<std::vector<ParticleObjShrdPtr>>& particlestosend) const
+void Particle::ParticleEngine::pack_particles_to_be_refreshed(
+    std::map<int, std::vector<char>>& sdata) const
 {
   // safety check
   if (not validdirectghosting_) FOUR_C_THROW("invalid direct ghosting!");
```
ppraegla: Since you make changes to the function anyway
```suggestion
  FOUR_C_ASSERT_ALWAYS(validdirectghosting_, "invalid direct ghosting!");
```

src/particle/src/engine/4C_particle_engine.hpp
ppraegla: I would keep the comments about the ParticleObject because they make it clear that this method handles communication differently. The other communication methods in the Particle code work with ParticleObjects.

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1603,28 +1588,38 @@ void Particle::ParticleEngine::determine_particles_to_be_refreshed(
...
-            std::make_shared<ParticleObject>(type, -1, states, -1, ghostedindex));
+        // pack type and ghosted index header
+        Core::Communication::PackBuffer header;
+        add_to_pack(header, static_cast<int>(type));
+        add_to_pack(header, ghostedindex);
+
+        // append header + pre-packed states to send buffer
+        auto& buf = sdata[sendtoproc];
+        buf.insert(buf.end(), header().begin(), header().end());
+        buf.insert(buf.end(), statesdata().begin(), statesdata().end());
```
ppraegla: With the ParticleObject, I liked that it was clear you have `ParticleObject::pack` and `ParticleObject::unpack`, and that both implementations are next to each other. So, it is less likely that you forget something when changing pack or unpack.

Can we do something similar here? Packing and unpacking now live in somewhat unrelated functions `pack_particles_to_be_refreshed` and `communicate_refreshed_particles`, plus additional duplicated packing in `pack_specific_states_of_particles_to_be_refreshed`. But they must fit together to work. However, I wasn't able to come up with a satisfying solution, as it is quite special that the states are only packed once and then reused. 

TL;DR: Maybe add a comment that the packed data must fit the unpacking in `communicate_refreshed_particles`. Or add a note in the doxygen comments that these functions must be called together.

ppraegla (commented):
Nice speed up! @vovannikov

src/particle/src/engine/4C_particle_engine.cpp
vovannikov: Yes, with the modern (N)RVO optimizations this is harmless, good point.

A would actually also implement another refactoring. Currently, `immediate_send_recv_known_procs()` mimics the behavior of `immediate_recv_blocking_send()` which existed previously, meaning that it receives a non-const reference to `sdata` and clears `sdata` at the end. Maybe I am missing something, but in my opinion this complicates reasoning about the code. It took me a while to understand that the only point for it being non-const is to perceive `immediate_recv_blocking_send()` as "send-and-its-gone" sequence. Why not to pass it const reference? This will clearly show the intention. And apply the same to `immediate_send_recv_known_procs()`. All these changes will simplify then the signatures of many other functions and make reasoning about them much easier. What do you guys think about it? Of course, this is the topic for a separate PR in order not to mess up the code.

src/particle/src/engine/4C_particle_engine.cpp
vovannikov: That is a good point. I have introduced a lightweight `ParticleRefreshEntry` struct that makes no harm for the proposed improvements overall but targets the aspect you mentioned. Have a look specifically at this commit to see how it was added 8faa70f32ae6489db8341f158f4252b4b5a4c5ee.

src/particle/src/engine/4C_particle_engine.hpp
vovannikov: I actually agree with @ppraegla. I would prefer to keep this remark for clarity of the intentions.

src/particle/src/engine/4C_particle_engine.cpp
vovannikov: Sure, done!

src/particle/src/engine/4C_particle_engine.hpp
georghammerl: Fine for me.

georghammerl (approved):

ppraegla (approved):
Thanks for the improvements. I like what you did with the pack and unpack in ParticleRefreshEntry.

src/particle/src/engine/4C_particle_engine.cpp
ppraegla: I like the idea to pass sdata as const reference. I think everywhere sdata will leave the scope after immediate_recv_blocking_send() and will be deleted anyway.

slfuchs:
Sorry for the (too) late reaction here. This week was quite busy so far. Nice optimization again. Good work 👍

vovannikov:
Just for the reference. Since the final version of this fix was a bit different from the one I originally had when running the profiling tests, I have rerun the profiling once again and compared it with my initial implementation to ensure that nothing got broken on the way. Seems fine.

4c_particle_perf_main_fix2

PR #2121, opened 2026-07-09
Resolve unreachable code in cut kernel

georghammerl:
## Description and Context
The Intel compiler pointed me to this line in the cut kernel code 
https://github.com/4C-multiphysics/4C/blob/55490c3d5d25b8ca5bbc653962be8ac2df47931a/src/cut/4C_cut_kernel.hpp#L2203
which evaluates always to false.

I am no expert of this piece of code. Please review my proposed solution.
The failing tests reveal that the proposed changes lead to different behavior. Tests with negative determinant seem to have passed in the past?!

Alternative: Delete the lines of code as they have never been reached in the past.

### DIFF
--- a/src/cut/4C_cut_kernel.hpp
+++ b/src/cut/4C_cut_kernel.hpp
@@ -2200,16 +2200,22 @@ namespace Cut::Kernel
         if (det < 1.0e-16) std::cout << "!!! determinant of jacobian is smaller than 1.0e-16 !!!\n";
       }
 
-      if (det < 1.0e-16 && det > 1.0e-16)
-#if EXTENDED_CUT_DEBUG_OUTPUT
-        std::cout << "Determinant in  compute position is very close to zero" << std::endl;
-#endif
-      if (Core::MathOperations<FloatType>::abs(det) == 0.0)  // here might lie problem for the cln
+      if (det < -1.0e-16)
       {
+        std::stringstream msg;
+        msg << "Determinant in compute position is negative: " << det;
+        FOUR_C_THROW("{}", msg.str());
+      }
+
+      if (Core::MathOperations<FloatType>::abs(det) < 1.0e-16)
+      {
+#if EXTENDED_CUT_DEBUG_OUTPUT
+        std::cout << "Determinant in compute position is very close to zero" << std::endl;
+#endif
         /* then calculation of a normal to a line like this makes no sense at
          * all! */
 #if EXTENDED_CUT_DEBUG_OUTPUT
-        std::cout << "Absolute value of the determinant is zero " << std::endl;
+        std::cout << "Absolute value of the determinant is close to zero" << std::endl;
 #endif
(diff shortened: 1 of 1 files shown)


src/cut/4C_cut_kernel.hpp
```diff
@@ -2200,16 +2200,22 @@ namespace Cut::Kernel
         if (det < 1.0e-16) std::cout << "!!! determinant of jacobian is smaller than 1.0e-16 !!!\n";
       }
 
-      if (det < 1.0e-16 && det > 1.0e-16)
+      if (det < -1.0e-16)
```
maxfirmbach: Should we use a numerical limit here?

src/cut/4C_cut_kernel.hpp
```diff
@@ -2200,16 +2200,22 @@ namespace Cut::Kernel
...
+      {
+        std::stringstream msg;
+        msg << "Determinant in compute position is negative: " << det;
+        FOUR_C_THROW("{}", msg.str());
+      }
+
```
maxfirmbach, line 2210: Same applies here.

maxfirmbach (commented):
@georghammerl I think the changes make a lot of sense. I would propose to go forward and fix the failing tests.

src/cut/4C_cut_kernel.hpp
```diff
@@ -2200,16 +2200,22 @@ namespace Cut::Kernel
         if (det < 1.0e-16) std::cout << "!!! determinant of jacobian is smaller than 1.0e-16 !!!\n";
       }
 
-      if (det < 1.0e-16 && det > 1.0e-16)
+      if (det < -1.0e-16)
+      {
+        std::stringstream msg;
+        msg << "Determinant in compute position is negative: " << det;
+        FOUR_C_THROW("{}", msg.str());
```
sebproell, line 2208: FOUR_C_THROW can format strings directly


```suggestion
        FOUR_C_THROW("Determinant in compute position is negative: {}", det);
```

georghammerl: Yes and no. 
A `std::formatter<T>` specialization is neccessary for every {} argument type. Since `FloatType` is a template parameter that can be instantiated with things like `double`, `Sacado::Fad::DFad<double>`, or `Core::CLN::ClnWrapper`, there is no generic formatter for it. To circumvent adding those at the moment, the workaround with the `operator<<` which is available for all template types.

vryy:
@georghammerl should we merge this soon? because I have the same problem in #1795

georghammerl:
I would like to merge, however the tests behave different and I have no clue on those details that matter in the cut code :-/

vryy:
Hi @georghammerl, I made the fix in #2161 and seems this check does not affect the test. I would suggest to merge #2161 and rebase this PR based on this for further adjustment. The only fix in #2161 is needed to continue with #1795.

PR #2124, opened 2026-07-12
Modify dual shape function construction to work for curved beams

dharinib98:
## Description and Context

In this PR, the currently implemented dual LM construction have been modified according to the equations (3.64 - 3.65) from Popp, A.: Mortar methods for computational contact mechanics and general interface
problems. Dissertation, Technische Universität München (2012) 

#2009


src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar.cpp
```diff
@@ -297,6 +298,14 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPairMortar<Beam, Solid, Mortar>:
...
   Core::LinAlg::Matrix<1, Solid::n_nodes_ * Solid::n_val_, double> N_solid(
       Core::LinAlg::Initialization::zero);
+  Core::LinAlg::Matrix<1, Beam::n_nodes_ * Beam::n_val_, double> N_mortar_primal(
+      Core::LinAlg::Initialization::zero);
+  Core::LinAlg::Matrix<Mortar::n_dof_, Mortar::n_dof_, double> local_D_mortar(
+      Core::LinAlg::Initialization::zero);
+  Core::LinAlg::Matrix<Mortar::n_dof_, Mortar::n_dof_, double> local_M_mortar(
+      Core::LinAlg::Initialization::zero);
+  Core::LinAlg::Matrix<Mortar::n_dof_, Mortar::n_dof_, double> local_A(
+      Core::LinAlg::Initialization::zero);
```
isteinbrecher: I am not a big fan of adding these matrices here for all cases, even tough they war only needed for one special case.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar.cpp
```diff
@@ -332,12 +466,40 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPairMortar<Beam, Solid, Mortar>:
...
-      GeometryPair::ShapeFunctionData<Mortar> shape_function_data;
-      GeometryPair::SetShapeFunctionData<Mortar>::set(shape_function_data, this->element1());
-      GeometryPair::EvaluateShapeFunction<Mortar>::evaluate(
-          N_mortar, projected_gauss_point.get_eta(), shape_function_data);
+
       GeometryPair::EvaluateShapeFunction<Beam>::evaluate(
-          N_beam, projected_gauss_point.get_eta(), this->ele1pos_.shape_function_data_);
+          N_beam, eta, this->ele1pos_.shape_function_data_);
+
+      if constexpr (std::is_same_v<Mortar, BeamInteraction::HermiteDual>)
```
isteinbrecher: With this proposed change, do we even need the explicit implementation of the dual shape functions any more?

src/beaminteraction/tests/4C_beaminteraction_beam_to_solid_volume_meshtying_mortar_test.cpp
```diff
@@ -116,7 +116,7 @@ namespace
       // Check the results for D.
       for (unsigned int i_row = 0; i_row < LambdaType::n_dof_; i_row++)
         for (unsigned int i_col = 0; i_col < BeamType::n_dof_; i_col++)
-          EXPECT_NEAR(local_D(i_row, i_col), result_local_D(i_row, i_col), 1e-11);
+          EXPECT_NEAR(local_D(i_row, i_col), result_local_D(i_row, i_col), 1e-11;
```
isteinbrecher: Is this a syntax error?

src/beaminteraction/tests/4C_beaminteraction_beam_to_solid_volume_meshtying_mortar_test.cpp
```diff
@@ -4427,160 +4427,439 @@ namespace
...
-    result_local_D(9, 6) = -0.00000000001687716633114178;
-    result_local_D(9, 9) = 0.2121320343565152;
-    result_local_D(10, 1) = 0.000000000003244959856374408;
-    result_local_D(10, 4) = -0.0000000000001378064329315976;
-    result_local_D(10, 7) = -0.00000000001687716633114178;
-    result_local_D(10, 10) = 0.2121320343565152;
-    result_local_D(11, 2) = 0.000000000003244959856374408;
-    result_local_D(11, 5) = -0.0000000000001378064329315976;
-    result_local_D(11, 8) = -0.00000000001687716633114178;
-    result_local_D(11, 11) = 0.2121320343565152;
```
isteinbrecher: Why do we need to adapt the test results?

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar.cpp
```diff
@@ -305,9 +314,132 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPairMortar<Beam, Solid, Mortar>:
   double segment_jacobian = 0.0;
   double beam_segmentation_factor = 0.0;
 
+  // compute local_A matrix
+  const unsigned int n_segments = this->line_to_3D_segments_.size();
+  if constexpr (std::is_same_v<Mortar, BeamInteraction::HermiteDual>)
```
isteinbrecher: We have to talk about this general procedure. This proposed solution should work fine, but I think will fail if the beam spans over multiple solid elements. We have to at least investigate what happens in this case.

dharinib98:
@isteinbrecher This is now again ready for review. I removed the dual hermite polynomials and moved all the implementation to the header. However, we are still required to adapt the results for the straight beam unit test added earlier, especially for the rows corresponding to the tangent dofs(I will check again why this is the case ).

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,190 @@
...
+  };
+
+  template <>
+  struct SetPairDependentShapeFunctionData<BeamInteraction::t_hermite_dual>
+  {
+    template <typename Beam, typename Segments, typename BeamReferencePosition,
+        typename BeamShapeFunctionData>
+    static void set(ShapeFunctionData<BeamInteraction::t_hermite_dual>& mortar_data,
+        const Segments& segments, const BeamReferencePosition& beam_reference_position,
+        const BeamShapeFunctionData& beam_shape_function_data)
```
isteinbrecher, line 73: We cannot use the segments of the pair. This will fail if a beam is only partly located inside a solid volume. 

So no need to pass the segments here, we need to integrate over the whole beam here.

This will likely result in a (very small) violation of conservation of momentum for curved beams, but we can investigate this.

Just a side note, this approach will fail, if the beam sticks out of the solid structure.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,190 @@
...
+        typename BeamShapeFunctionData>
+    static void set(ShapeFunctionData<Mortar>&, const Segments&, const BeamReferencePosition&,
+        const BeamShapeFunctionData&)
+    {
+      // No pair-dependent data required for standard mortar shape functions.
+    }
+  };
+
+  template <>
+  struct SetPairDependentShapeFunctionData<BeamInteraction::t_hermite_dual>
```
isteinbrecher: `SetBeamDependentShapeFunctionData` would be a better name.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,190 @@
...
+      Core::LinAlg::Matrix<n_shape, n_shape, double> M(Core::LinAlg::Initialization::zero);
+      Core::LinAlg::Matrix<3, 1, double> dr_beam_ref(Core::LinAlg::Initialization::zero);
+
+      Core::LinAlg::Matrix<1, n_shape, double> N_primal(Core::LinAlg::Initialization::zero);
+
+      for (const auto& segment : segments)
+      {
+        const double segment_factor = 0.5 * segment.get_segment_length();
+
+        for (const auto& gauss_point : segment.get_projection_points())
```
isteinbrecher: Here, you can simply loop over Gauss points in the interval [-1,1]. For now, we can even hardcode the number of integration points here.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar.cpp
```diff
@@ -334,6 +334,9 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPairMortar<Beam, Solid, Mortar>:
       N_solid.clear();
       GeometryPair::ShapeFunctionData<Mortar> shape_function_data;
       GeometryPair::SetShapeFunctionData<Mortar>::set(shape_function_data, this->element1());
+      GeometryPair::SetPairDependentShapeFunctionData<Mortar>::template set<Beam>(
+          shape_function_data, this->line_to_3D_segments_, this->ele1posref_,
+          this->ele1pos_.shape_function_data_);
```
isteinbrecher: Why are these definitions inside this loop? This will have a very bad impact on performance. We should be able to move them out of the loop.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,190 @@
...
+      // Phi_j = sum_k A_jk N_primal_k.
+      N.clear();
+
+      for (unsigned int j = 0; j < 4; ++j)
+      {
+        for (unsigned int k = 0; k < 4; ++k)
+        {
+          N(j) += shape_function_data.dual_transformation_(j, k) * N_primal(k);
+        }
+      }
```
isteinbrecher: Isn't this just a simple matrix multiplication? Why is this coded explicitly here?

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_mortar_shape_functions_dual_hermite.hpp
```diff
@@ -0,0 +1,190 @@
...
+  template <>
+  struct PrintElementData<BeamInteraction::t_hermite_dual>
+  {
+    template <typename ScalarType>
+    static void print(const ElementData<BeamInteraction::t_hermite_dual, ScalarType>& element_data,
+        std::ostream& out)
+    {
+      constexpr auto max_precision{std::numeric_limits<double>::digits10 + 1};
+      out << std::setprecision(max_precision);
+      out << "\nElement reference length: " << element_data.shape_function_data_.ref_length_;
```
isteinbrecher, line 180: We should also print the dual transformation matrix here.

src/beaminteraction/tests/4C_beaminteraction_beam_to_solid_volume_meshtying_mortar_test.cpp
```diff
@@ -4332,7 +4332,7 @@ namespace
   }
 
   /**
-   * \brief Test a non straight beam in a hex8 element, with dual hermite mortar shape functions.
+   * \brief Test a straight beam in a hex8 element, with dual hermite mortar shape functions.
```
isteinbrecher, line 4335: This confuses me. Is this a test for a straight beam or not? For a straight beam, the results should not change from the previous version to this one (or am I missing something?).

PR #2125, opened 2026-07-12
Remove `post_output()`

rjoussen:
## Description and Context

While reviewing #2108, I noticed that the `post_output` function is _almost_ completely inert. The only place across all ~25 implementations of post_output that actually did something was in `Solid::ModelEvaluator::BrownianDyn::post_output()`, in which `generate_gaussian_random_numbers()` is called, which, as far as I understand, generates random numbers for the next timestep. Since this is clearly unrelated to the output, I moved it to the predict() function instead.


mayrmt (dismissed):

c-p-schmidt (dismissed):

davidrudlstorfer (dismissed):
I am always a bit hesitant with the changes of the brown dyn framework, but here it lokos like all the tests still succeed :)

I haven't checked it in detail, but the predict step is executed before or after the post_output step @rjoussen ?

rjoussen:
> I haven't checked it in detail, but the predict step is executed before or after the post_output step @rjoussen ?

It needs to run at some point before evaluation, when time and step have already been updated. So yes, predict runs before post_output. For the first timestep, random number generation has only occurred in `setup` so far. With this change, `generate_gaussian_random_numbers` is actually called twice for the first timestep, which is no problem, though, since `generate_gaussian_random_numbers` itself guards against generating random numbers multiple times per timestep. I tried to remove `generate_gaussian_random_numbers` from `setup`, but this would lead to failing random crosslinking tests, since they also draw random numbers from the same generator, so the order is important (to keep the tests passing, not from a physical point of view, of course).

Since `generate_gaussian_random_numbers` is safe to call multiple times per timestep, we could also put it directly in `evaluate_brownian()`. This is probably the cleanest solution. Done in https://github.com/4C-multiphysics/4C/pull/2125/commits/040454e6144e859aee71115737ebe89314782bcd

davidrudlstorfer (approved):
@rjoussen thanks for looking into this again :)

mayrmt (approved):

PR #2128, opened 2026-07-15
[beaminteraction]Correct dual hermite shapefunctions

dharinib98:
## Description and Context

This pull request corrects the dual Hermite shape functions introduced in #2009.
The previous implementation accidentally used incorrect scaling. 

#2009


mayrmt (commented):
@dharinib98 I am fine with the current changes. Just one question: how do we know, that this version now is the correct version? How has this been verified?

dharinib98:
> @dharinib98 I am fine with the current changes. Just one question: how do we know, that this version now is the correct version? How has this been verified?
This is the version that satisfies partition of unity, hence the one we want. I checked that it satisfies this property for length = 1.

mayrmt:
What about other lengths?

dharinib98:
> What about other lengths?

For arbitrary L, we can't really check the standard unweighted partition of unity due to the difference in dimensionality in the derivative shape functions, however they satisfy the weighted relation
$\phi_1 + L\phi_2 + \phi_3 + L\phi_4 = 1$ 
 which I think is sufficient.

mayrmt (dismissed):

dharinib98:
Another small related issue that we include in this correction pull request. I encountered the issue that some rows in D are all zeros (or close to zero) in some examples. Hence I changed the diagonal check accordingly. Is this expected or hints at something going wrong ?

isteinbrecher:
> This is the version that satisfies partition of unity, hence the one we want. I checked that it satisfies this property for length = 1. 

@dharinib98 Are you sure this is correct? How did you introduce the additional scaling to ensure that we keep all the other envisioned properties of the bi-orthogonality condition? I let's have a short internal talk about this before we merge this.

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_utils.cpp
```diff
@@ -868,7 +868,7 @@ void BeamInteraction::check_diagonal_like_structure(
       }
     }
 
-    if (nnz_in_row != 1)
+    if (nnz_in_row > 1)
```
isteinbrecher: This now makes the check less strong - but it is to be expected the way we currently do the check.

The D matrix can have complete zero columns. We check $D$ and $D^T$, so with the current check we need this change. One alternative would be to add an option to this function if the rows or columns should be checked. Then we could keep the `!=1`, but this might require additional communication. So I am not 100% sure if we want to go down that road.

isteinbrecher (approved):

mayrmt (approved):

PR #2129, opened 2026-07-15
[beaminteraction]Add postprocess routine for Dual Lagrange Multipliers

dharinib98:
## Description and Context

This PR implements post-processing of the Lagrange multipliers obtained with the dual LM approach(see #2009), enabling the visualization of smooth Lagrange multiplier curves. The implementation follows the reconstruction described in Equations (34) and (35) of Seitz et al., “Isogeometric dual mortar methods for computational contact mechanics.”


mayrmt (commented):
@dharinib98 is it true, that this change is related to the runtime output? If yes, please add an output test.

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_pair_mortar.cpp
```diff
@@ -244,13 +247,47 @@ void BeamInteraction::BeamToSolidVolumeMeshtyingPairMortar<Beam, Solid,
...
+            }
+          }
 
           // Add to output data.
           for (unsigned int dim = 0; dim < 3; dim++)
           {
             point_coordinates.push_back(Core::FADUtils::cast_to_double(X(dim)));
             displacement.push_back(Core::FADUtils::cast_to_double(u(dim)));
             lambda_vis.push_back(Core::FADUtils::cast_to_double(lambda_discret(dim)));
+            lambda_postprocessed_vis.push_back(
```
isteinbrecher: Can we do this in the constextpression if before?

src/beaminteraction/src/contact/beam_to_solid/volume/4C_beaminteraction_contact_beam_to_solid_volume_meshtying_visualization_output_writer.cpp
```diff
@@ -87,6 +87,7 @@ BeamInteraction::BeamToSolidVolumeMeshtyingVisualizationOutputWriter::
       auto& visualization_data = visualization_writer->get_visualization_data();
       visualization_data.register_point_data<double>("displacement", 3);
       visualization_data.register_point_data<double>("lambda", 3);
+      visualization_data.register_point_data<double>("lambda_postprocessed", 3);
```
isteinbrecher: We now initialize this even, if we don't need the postprocessed lambda. We should not do this postprocessing in cases where we don't need it. We can either automatically do the post processing for the dual shapefuctions or have an input file option `postprocess_lambda` that allows to turn this on/off (I prefer the input file option).

isteinbrecher (commented):
Thanks for the contribution @dharinib98.

I have two additional questions:
- How is the post processing actually done? (Maybe we should talk about this internally)
- Why are there no adapted test results / failing tests with this change?

PR #2130, opened 2026-07-20
Implement a Mohr Coulomb plasticity model

ischeider:
## Description and Context

I implemented a Mohr-Coulomb plasticity model into 4C. It has a possibly non-associated flow rule and the typical parameters
`COHESION, FRICTION_ANGLE, DILATANCY_ANGLE`. In addition, there might be a linear or saturating hardening. of the cohesion. 

The main implementation was done by openAI GPT 5.6 sol, with little interaction, some review and correction by me. 

One main point is that the LLM automatically delivers some documentation and puts it into the developer guide. For now I left it there, so that we can discuss, whether this is useful or not (it is in my opinion), and whether it's the right place for it (I guess not).


vryy:
Hi @ischeider, the test results look fine to me, although for soil the MC is normally used in the context of small strain. I would suggest trying some structural tests to test the reliability of the implementation. The first one is the slope stability (sent to you separately via email). Other than that, one could test the non-associativity by setting dilatancy angle=1 (this situation typically occurs in soil investigation reports when the friction angle < 30).

PR #2135, opened 2026-07-27
[beaminteraction] Allow output of energy in beam-to-solid volume mortar simulations

isteinbrecher:
Allow to output energy for beam-to-solid simulations with mortar contact discretization. Each pair outputs zero energy, as the energy is evaluated globally in the mortar manager.


dharinib98 (approved):

maxfirmbach (approved):
If this is the behavior we want, I'm fine with it.

PR #2140, opened 2026-07-27
[geometric search] Allow beam bounding volume based on beams surface

isteinbrecher:
This PR consists of a few changes:
1. The data structure for geometric search is simplified by using features of the new input implementation.
1. Elements can now return multiple bounding volumes. This allows to fine-tune the geometric approximation in the ArborX search.
1. With the previous point, users can now specify the type of bounding volume for beam elements. A new variant is added that consists of integration points distributed on the beams surface.
For certain problems, e.g., 2D-3D beam-to-solid pairs, this drastically increases performance for large simulations, from $\mathcal{O}(n^2)$ to $\mathcal{O}(n^1)$.


isteinbrecher:
WIP for now

maxfirmbach:
I will have a look at this soonish.

davidrudlstorfer:
@isteinbrecher just retag me as a reviewer once this is ready :)

isteinbrecher:
@maxfirmbach @davidrudlstorfer this should now be ready. The approach changed compared to the first variant, so I updated the description.

src/beam3/src/4C_beam3_base.cpp
```diff
@@ -412,35 +412,87 @@ void Discret::Elements::Beam3Base::get_triad_of_binding_spot(Core::LinAlg::Matri
 
 /*--------------------------------------------------------------------------------------------*
  *--------------------------------------------------------------------------------------------*/
-Core::GeometricSearch::BoundingVolume Discret::Elements::Beam3Base::get_bounding_volume(
```
davidrudlstorfer, line 415: Just a quick question: does it make sense to keep the old method as well? Or is the new one superior in every sense?

src/constraint_framework/4C_constraint_framework_submodelevaluator_mpc.cpp
```diff
@@ -136,7 +137,7 @@ void Constraints::SubmodelEvaluator::RveMultiPointConstraintManager::check_input
       Teuchos::getIntegralValue<Constraints::MultiPoint::RveReferenceDeformationDefinition>(
           mpc_parameter_list, "RVE_REFERENCE_POINTS");
 
-  node_search_toler_ = Teuchos::getDoubleParameter(geom_search_parameter_list, "POINT_TOLERANCE");
```
davidrudlstorfer, line 139: Nice that more Teuchos goes away :)

davidrudlstorfer (approved):
I do not fully get the new kdop bounding volume of the beams with the circumferential gauss points but maybe you can explain that to me in person soonish :)

maxfirmbach (approved):
Looks good. Also nice to see, that a few parameter related files are gone and the overall code is much simpler now.

src/beam3/src/4C_beam3_base.cpp
isteinbrecher: Yes, the old method is still the default, because it includes the whole beam _volume_. The new method is an approximation of only the beam _surface_. So for special cases where we are only interested in interaction pairs with the surface the new method makes sense.

PR #2141, opened 2026-07-28
[Particle] Make sdata const and return rdata in particle utils

vovannikov:
As proposed in the discussion of #2117, I have decided to refactor the declarations of `immediate_recv_blocking_send()` and  `immediate_send_recv_known_procs()` such that these methods would receive constant references to `sdata` and also explicitly return `rdata` instead of filling it via an output parameter. This allowed to subsequently simplify the callers.

#2117


slfuchs:
Have you also checked the runtimes with this PR compared to main? Is there any difference?

ppraegla (approved):

georghammerl (approved):

vovannikov:
> Have you also checked the runtimes with this PR compared to main? Is there any difference?

I have performed the test and these changes did not hurt. They even show some positive influence, but I guess this can be the run-to-run noise since the changes here should not have any performance hits.

image

PR #2142, opened 2026-07-29
Add a performance test for reduced-lung

maxiludwig:
## Description 
Adds a `reduced_lung` performance test (7 generations as the minimal benchmark, 17 generations as the full nightly benchmark), run single-core with UMFPACK since this problem doesn't benefit from parallelism.

Rebased onto #2166, which moved the reduced lung to mesh based geometry input. That removed the need
for the generic `PLACEHOLDERS` mechanism this PR previously had to add to the performance test
infrastructure.

Developed with Claude Code (Claude Opus 5), used for generating the meshes and setting up the test;
the commit carries a `Co-Authored-By` trailer. All AI-assisted code has been reviewed and validated
by the author, who takes responsibility for the contribution.


tests/performance_tests/list_of_performance_tests.cmake
```diff
@@ -55,6 +55,22 @@ four_c_performance_test(
...
   )
+four_c_performance_test(
+  TEST_FILE
+  reduced_lung.4C.yaml.in
+  MESH
+  reduced_lung_tree_4C_fields.json
+  NP_MINIMAL
+  1
+  NP_FULL
+  1
```
c-p-schmidt, line 67: You really want to execute both tests on just one processor?

c-p-schmidt (dismissed):

tests/performance_tests/list_of_performance_tests.cmake
maxiludwig: Yes, it's the relevant use case for these reduced lung examples. With our custom SoE we require direct solvers and i checked that the big example doesn't benefit from more than one proc for the MUMPS solver. So Umfpack is faster here.

maxiludwig:
@bgoderbauer as you suggested, I've set `restart_every` and `results_every` to 76, so that we don't actually write results and restarts for the examples

tests/performance_tests/meshes/full/reduced_lung_tree_4C_fields.json
ppraegla, line 1: The file is huge with 30 MB. Can this file be generated during testing? If more performance tests will be added, the size of the input files can quickly grow

I also thought about adding a performance test for particles. But the input file will also be so large that it doesn't make sense to commit the raw input.

davidrudlstorfer: Git LFS is already set-up for the performance tests? This file should also move in there

EDIT: see https://github.com/4C-multiphysics/4C/blob/main/tests/performance_tests/meshes/full/README.md

ppraegla: Thanks. I didn't know. :+1:

cmake/functions/four_c_performance_testing_functions.cmake
```diff
@@ -13,7 +13,11 @@
 # enabled by enabling the option FOUR_C_ENABLE_FULL_PERFORMANCE_TESTS.
 #
 # required parameters:
-#   TEST_FILE:                    Name of the input file in the directory tests/performance_tests.
+#   TEST_FILE:                    Name of the input file in the directory tests/performance_tests. If this name
+#                                 ends in `.in`, the file is treated as a template and the suffix is stripped
+#                                 for the generated test name and the file actually handed to 4C. Use the `.in`
+#                                 suffix for input files that contain placeholders (e.g. @NUM_NODES@) that are
+#                                 not valid 4C-YAML input and would fail schema validation.
```
davidrudlstorfer: It is a nice feature but it is suuuuper specific. Is this really what we want?

Isn't the way that @amgebauer handled it and how it is already implemented a nice way to do this?

cmake/functions/four_c_performance_testing_functions.cmake
```diff
@@ -28,6 +32,18 @@
...
+#                                 FULL_VALUE". Every @NAME@ in the input file becomes MINIMAL_VALUE (minimal variant)
+#                                 or FULL_VALUE (full variant). Example:
+#                                   PLACEHOLDERS
+#                                   "NUM_NODES      192 134535"
+#                                   "NUM_ELEMENTS   191 134534"
+#                                 Use for problem-specific parameters that depend on the mesh/data variant, without
+#                                 teaching this generic function about any physics module's parameters.
+#                                 Each entry needs exactly 3 tokens, a valid/unique NAME that actually occurs as
+#                                 @NAME@ in the resolved test file. NAME "MESH_FILE" is reserved (it would collide
+#                                 with the mandatory @MESH_FILE@ substitution) and is rejected.
```
davidrudlstorfer: This is also really specific and cannot be reused anywhere else than the lung input file. Maybe a different approach would be nicer.

davidrudlstorfer (commented):
First of all cool that you add another performance test to 4C!

As I was involved into all the cmake rework and especially testing framework rework I would like to give my opinion here.

The proposed extension is super specific and cannot be reused anywhere else in the 4C framework. With the testing functions we want to provide functions that are generally applicable and not specified for one test only (that's what we had before and now everything is modularized). Therefore, I would urge to improve the logic a little bit of this approach.

One question that comes to my mind while looking at this PR. I know that the lung stuff mainly is touched by @bgoderbauer and you. Would it make sense to somehow switch the lung input to a vtu input file rather than the current rather arbitrary json input (which is also somehow misused I guess? Normally you use this for continous quantities and now you prescribe the radius in space?). In this way you could also open the input file in paraview (as a first benefit) and secondly you could also use the already available testing infrastructure?

Do not get me wrong that changes are necessary here but I want to propose some ideas of improvement. And with the "new" lung input we have problems on each corner of the testing framework (from unit tests to integration tests to performance tests). So maybe this should be thought about before going deeper. Sorry for the wall of text 👀

davidrudlstorfer:
And what I forgot, this should probably be just discussed by the @4C-multiphysics/maintainer. Thx!

c-p-schmidt:
Let me add some thoughts. When I first saw that such a large JSON file had been added, I was also rather skeptical and wondered whether a binary alternative was available or better.
However, when I briefly discussed with @maxiludwig, realizing that the file size for the large file (which is already about the largest it will get, as he is rather simulating fewer generations (leading to less data) than here) is only about 30 MB, and input reading takes around 4 seconds, I am not so sure whether this is really so relevant or pressing.
I agree that it would be nice to have a directly visualizable file format like vtu, but I am not sure whether the required input data fits that format. When I think about these airway trees, they remind me of a complex beam structure, so that you (@davidrudlstorfer) can maybe shed more light on that yourself.
Moreover, concerning the necessary changes to the CMake functionality of the performance tests. Yes, the original idea was that only the mesh needs to be replaced. However, @amgebauer and I are only working with conventional FE meshes and had these in mind when discussing how to design this. However, if this is too brittle and prevents other parts of 4C from being performance-tested, I do not really see an issue with adding more functionality if necessary.

cmake/functions/four_c_performance_testing_functions.cmake
lauraengelhardt: I think this is actually the correct approach for all template files within the testing framework. The schema validation for the old approach only worked because the expected data type for the mesh was a string, and @MESH_FILE@ is also a string. However, without replacing the placeholder, the file would still not be a runnable 4C input file (well...it is runnable, but would fail, hope this becomes clear).

I would even suggest adding the .in suffix to all template files to make it clear that these are templates rather than the actual input files.

cmake/functions/four_c_performance_testing_functions.cmake
lauraengelhardt: I'm rather wondering why this has to be treated differently than a substitution of @MESH_FILE@. Both are substitutions, why have a separate mechanism for PLACEHOLDERS and MESH_FILE and not just generalize it?

lauraengelhardt (commented):
Some more thoughts from my side 😉 
Anyhows, as @davidrudlstorfer said, it's really nice to see more performance tests being added!

cmake/functions/four_c_performance_testing_functions.cmake
ppraegla: I was not involved in the original setup of the performance tests and the replacement function. But what is the reason to do the replacement with cmake?

Since there is a need for more flexible replacements, I think Python and Jinja for the replacement would be better. The user only needs to provide a Python file/function that returns a dict containing the replacements. You would have a lot more flexibility with Jinja, e.g., replacing entire blocks (if else block with Jinja) or more complex replacements than one value.

maxiludwig:
With #2166 the reduced lung now reads its geometry from a mesh file, so both trees are just VTU meshes and the test registers through the standard `MESH` mechanism like every other performance test. The `PLACEHOLDERS` extension of `four_c_performance_test`, the `.4C.yaml.in` template file and the extra LFS patterns for the JSON field files are all gone.

For the future it might still be worth replacing the `sed` based `@MESH_FILE@` substitution by a proper jinja based template mechanism as suggested by @ppraegla, so that performance tests can parametrize their input files in a more standardized way.

bgoderbauer (dismissed):

tests/performance_tests/reduced_lung.4C.yaml
```diff
@@ -0,0 +1,396 @@
...
+      - id: 2
+        function_id: 2
+    flow:
+      - id: 1
+        function_id: 1
+SOLVER 1:
+  SOLVER: "UMFPACK"
+  NAME: "Reduced_lung_solver"
+FUNCT1:
+  - SYMBOLIC_FUNCTION_OF_TIME: 1000*flow_ml_s
```
c-p-schmidt, line 62: Not related to the changes of this PR, but you could think about using **CUBIC_SPLINE_FROM_CSV** instead of this gigantic list of values in the future for your input files.

c-p-schmidt (approved):
Looks good. Thank you for going the extra mile concerning the geometry and for adding a performance test.

tests/performance_tests/meshes/minimal/reduced_lung_tree.vtu
davidrudlstorfer, line 1: How many elements is this minimal tree?

If it is only a handful of elements it would make sense to save the vtu as human readable rather than compressed

davidrudlstorfer (approved):
Thanks @maxiludwig for implementing the vtu input ;)

This is now really nice compared to the highly complex code with the original json files

Thanks!

tests/performance_tests/meshes/minimal/reduced_lung_tree.vtu
maxiludwig: Its ~190 elements, so probably its nothing you can picture in your mind's eye.

tests/performance_tests/reduced_lung.4C.yaml
maxiludwig: True, thanks. I wasn't aware of this option 👍

PR #2144, opened 2026-07-30
[beaminteraction] Post filter penalty contribution

maxfirmbach:
## Description and Context

The assembly of penalty contributions results in explicit zero entries. Especially in the case, where one scaling parameter is explicitly set to `0.0`. The explicitly stored zeros destroy the graph structure and unnecessarily increase the computational effort, thus multigrid struggles.

With this PR we post filter the penalty contribution and remove explicit zeros (low entries).


tests/input_files/beam3eb_static_beam_to_solid_volume_meshtying_mortar_lagrange_individually_augmented_line3.4C.yaml
```diff
@@ -43,8 +43,8 @@ BEAM INTERACTION/BEAM TO SOLID VOLUME MESHTYING:
   CONSTRAINT_STRATEGY: lagrange
   LAGRANGE_FORMULATION: augmented
   PENALTY_PARAMETER: 1
-  AUGMENTATION_SCALING_PARAMETER_SOLID: 1
-  AUGMENTATION_SCALING_PARAMETER_BEAM: 100
+  AUGMENTATION_SCALING_PARAMETER_SOLID: 0.0
```
isteinbrecher, line 46: Why does this not break the test?

maxfirmbach: Why should it break the test? Do we theoretically divide with that value at some point? I don't think so.

mayrmt (approved):

mayrmt:
@maxfirmbach Makes sense. 

Out of curiosity: can you quantify the performance impact?

Should we note this as a technical remark in the paper?

tests/input_files/beam3eb_static_beam_to_solid_volume_meshtying_mortar_lagrange_individually_augmented_line3.4C.yaml
isteinbrecher: Didn't we setup this test with high non-linear solver tolerances, such that the effect of the augmentation parameters can be seen in the results?

isteinbrecher: But I remember that this was quite brittle with changing the variables. So please check that briefly and then let's merge this.

isteinbrecher (approved):

maxfirmbach:
@mayrmt The performance gain is kind of given by the timings of variant IIIa vs variant IIIb in the respective preprint.

PR #2145, opened 2026-07-30
[Particle] Improve bin weight estimate

vovannikov:
## Description and Context
This feature evaluates weights for particle bins based on the number of interaction pairs a bin contains. This information is already available, since we construct neighbors anyways. As the outcome, it improves the particle distribution and optimizes the load balance, especially when the work load per rank is rather large, e.g. >300,000 particles/rank. For smaller number of particles per rank, e.g. ~50,000, the effect is less significant but still present.

image

#2117, #2074

### DIFF
--- a/src/particle/src/algorithm/4C_particle_algorithm.cpp
+++ b/src/particle/src/algorithm/4C_particle_algorithm.cpp
@@ -140,23 +140,26 @@ void Particle::ParticleAlgorithm::setup()
   // ghost particles on other processors
   particleengine_->ghost_particles();
 
-  // build global id to local index map
-  particleengine_->build_global_id_to_local_index_map();
-
-  // build potential neighbor relation — populates bin_interaction_costs_ for cost-aware rebalancing
-  if (particleinteraction_) build_potential_neighbor_relation();
-
-  // second redistribution pass: redistribute load using interaction-pair counts as bin weights
-  // now that bin_interaction_costs_ is available from the neighbor build above
   if (particleinteraction_)
   {
+    // build particle-to-particle neighbors only, to populate bin_interaction_costs_ for
+    // cost-aware rebalancing below; the global id map and wall neighbor relations are skipped
+    // here since both would be invalidated by the redistribution pass right after anyway
+    build_potential_neighbor_relation(/*include_wall_neighbors=*/false);
+
+    // second redistribution pass: redistribute load using interaction-pair counts as bin weights
+    // now that bin_interaction_costs_ is available from the neighbor build above
     distribute_load_among_procs();
 
-    // re-ghost and rebuild after second redistribution
+    // re-ghost after second redistribution
     particleengine_->ghost_particles();
-    particleengine_->build_global_id_to_local_index_map();
-    build_potential_neighbor_relation();
   }
+
+  // build global id to local index map (once, for the final distribution)
+  particleengine_->build_global_id_to_local_index_map();
+
+  // build potential neighbor relation, including wall neighbors, for the final distribution
+  if (particleinteraction_) build_potential_neighbor_relation();
 
   // setup initial states
   if (not isrestarted_) setup_initial_states();
@@ -982,14 +985,14 @@ void Particle::ParticleAlgorithm::distribute_load_among_procs()
     Core::IO::cout(Core::IO::verbose) << "distribute load in step " << step() << Core::IO::endl;
 }
 
-void Particle::ParticleAlgorithm::build_potential_neighbor_relation()
+void Particle::ParticleAlgorithm::build_potential_neighbor_relation(bool include_wall_neighbors)
 {
   TEUCHOS_FUNC_TIME_MONITOR("Particle::ParticleAlgorithm::build_potential_neighbor_relation");
 
   // build particle to particle neighbors
   particleengine_->build_particle_to_particle_neighbors();
 
-  if (particlewall_)
+  if (particlewall_ && include_wall_neighbors)
   {
     // relate bins to column wall elements
     particlewall_->relate_bins_to_col_wall_eles();
--- a/src/particle/src/algorithm/4C_particle_algorithm.cpp
+++ b/src/particle/src/algorithm/4C_particle_algorithm.cpp
@@ -134,7 +134,7 @@ void Particle::ParticleAlgorithm::setup()
   // setup initial rigid bodies
   if (particlerigidbody_) setup_initial_rigid_bodies();
 
-  // distribute load among processors
+  // distribute load among processors (first pass: by particle count)
   distribute_load_among_procs();
 
   // ghost particles on other processors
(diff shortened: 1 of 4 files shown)


src/particle/src/algorithm/4C_particle_algorithm.cpp
```diff
@@ -134,16 +134,31 @@ void Particle::ParticleAlgorithm::setup()
...
   // ghost particles on other processors
   particleengine_->ghost_particles();
 
-  // build global id to local index map
+  if (particleinteraction_)
+  {
+    // build particle-to-particle neighbors only, to populate bin_interaction_costs_ for
+    // cost-aware rebalancing below; the global id map and wall neighbor relations are skipped
+    // here since both would be invalidated by the redistribution pass right after anyway
+    build_potential_neighbor_relation(/*include_wall_neighbors=*/false);
```
georghammerl, line 148: I wonder if the assumption of only the particle-particle interaction entering the cost function is fair? The evluation particle-wall is (what I would guess) way more expensive as the nonlinear contact point search needs to be performed (see `Core::Geo::nearest_3d_object_on_element` calls in `interaction_sph_neighbor_pairs` and `interaction_dem_neighbor_pairs`.

vovannikov: That is a very good point! I thought of this. But this issue existed even before this modification. Strictly speaking, that's why we have customizable weights per particle type. Here the weights are averaged within each contact, so if a user provided custom weights, then this information is also accounted for. But it would be nice to have this data provided in more automatic manner rather then user provided weights, this will then further resolve the imbalance issue.

slfuchs: The reason why I neglected particle-wall pairs respectively wall element counts for the bin distribution is, that in most cases the wall is surrounding the particle domain and coupling is performed only on the interface. So my aim was to optimize for ideal distribution of particles among procs rather than wall elements among procs. Note that the structural elements in the structural solver are distributed based on graph partitioning.  That said, and also for the sake of ideally cubic processor domains, only the particles are considered for the bin distribution. I think, that considering particle-wall pairs in the bin distribution might lead to more lengthy shaped processor domains, such that each processor gets a share of wall elements and consequently this leads to more communication between the processors due to the worse ratio of processor domain boundary to volume.

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -554,6 +555,11 @@ void Particle::ParticleEngine::build_particle_to_particle_neighbors()
           potentialparticleneighbors_.push_back(
               std::make_pair(std::make_tuple(type, Owned, ownedindex),
                   std::make_tuple(neighbortype, neighborstatus, neighborindex)));
+
+          // accumulate interaction cost for load balancing weight estimation, weighted by the
+          // average of the dynamic load balance factors of the two interacting particle types
+          bin_interaction_costs_[gidofbin] +=
```
slfuchs, line 569: I would prefer to move this to a separate function. I like to keep the concept, that one method is doing only one thing. So `build_particle_to_particle_neighbors()` should only build pontential particle neighbors.

vovannikov: The potential particle neighbors is exactly the data this criterium relies on. So in that sense `bin_interaction_costs_` is not something that is absolutely independent, on the contrary, now the coupling in the code is quite tight, which was the exact intent. However, I don't have a strong opinion on this.

What bothers me more - is actually the overhead when moving the population of `bin_interaction_costs_` into a different function. I could have just iterated over `potentialparticleneighbors_` as it contains the neighbors types (still, there is overhead here but not significant) but there is no `gidofbin` info available. Unless there is a quick way to get it from a `potentialparticleneighbors_` entry. Is something like this available?

src/particle/src/algorithm/4C_particle_algorithm.cpp
vovannikov: @slfuchs that makes sense. It would be actually interesting to measure the effect of particle-wall interactions on the load imbalance. The benchmark I have been using so far has no walls. I guess, the best way is to run the same geometry with the rigid boundary particles and the particle walls and measure the load imbalance. Anyways, this is one of those things that, in my opinion, needs profiling and measuring. And I would postpone it for a different PR.

src/particle/src/engine/4C_particle_engine.cpp
slfuchs: you are right, since the global id of the bin is needed it probably makes sense to leave it as it is.

jeremylt:
Two questions here

1) Do you have a sense or data on what this does for per-process heterogeneity? Roughly speaking, we'll get better GPU performance with better per-process particle heterogeneity, as long as we have enough particles per process.

2) Do you have the configurations for these tests somewhere I can use? I'd like to make sure my GPU porting efforts aren't impacting CPU-only runs negatively.

vovannikov:
> Two questions here
> 
> 1. Do you have a sense or data on what this does for per-process heterogeneity? Roughly speaking, we'll get better GPU performance with better per-process particle heterogeneity, as long as we have enough particles per process.
> 2. Do you have the configurations for these tests somewhere I can use? I'd like to make sure my GPU porting efforts aren't impacting CPU-only runs negatively.

@jeremylt 

1. This is a very good question. I have not really checked the costs of different particle types in details. What this algorithm does, it can indeed put a bit more particles on one rank and a fewer number on another based on how expensive the operations with the particles are. But this strategy can (and probably should) be changed together with the binning strategy.
2. I have sent you the benchmark and the scripts for running in it and postprocessing the measurements data.

vovannikov:
What do you guys think about this feature?

I have a feeling, that given the ongoing GPU transition, most probably we will need to reconsider the particle redistribution algorithm grouping the particles according to their types. In this context, this feature can serve as a good benchmark reference to measure the obtained imbalance for new approaches.

jeremylt:
This looks like a clear win for CPU workloads, and we can always adjust the binning strategy when running for the GPU if we determine it is a bottleneck.

slfuchs:
I have the same opinion than @jeremylt , so let's proceed here and get it into the master. Later on we can still refine or adapt if necessary.

slfuchs (dismissed):

georghammerl (dismissed):

jeremylt:
A note on the casting, we could do `EnumTools::enum_integer(foo)` instead of `static_cast<int>(foo)` if that's preferred, but it wouldn't save us any space so its whatever seems clearer to adopt as the general preference

ppraegla (dismissed):

ppraegla:
> A note on the casting, we could do `EnumTools::enum_integer(foo)` instead of `static_cast<int>(foo)` if that's preferred, but it wouldn't save us any space so its whatever seems clearer to adopt as the general preference

Good point. I suggest sticking to `static_cast<int>`. To me it seems to be the standard way to convert enum classes. And we do not rely on an external library for this simple task. magic_enum is better suited for obtaining the string representation.

slfuchs (dismissed):

vovannikov:
> > A note on the casting, we could do `EnumTools::enum_integer(foo)` instead of `static_cast<int>(foo)` if that's preferred, but it wouldn't save us any space so its whatever seems clearer to adopt as the general preference
> 
> Good point. I suggest sticking to `static_cast<int>`. To me it seems to be the standard way to convert enum classes. And we do not rely on an external library for this simple task. magic_enum is better suited for obtaining the string representation.

I agree. Explicit `static_cast` does the job.

Btw, I can get why that single check has been failing, looks like it is not related to the PR itself.

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1978,7 +1985,7 @@ void Particle::ParticleEngine::remove_particles_from_containers(
     // iterate in reversed order over particles to be removed
     std::set<int>::reverse_iterator rit;
     for (rit = particlestoremove[static_cast<int>(type)].rbegin();
-        rit != particlestoremove[static_cast<int>(type)].rend(); ++rit)
+         rit != particlestoremove[static_cast<int>(type)].rend(); ++rit)
```
ppraegla: This change causes the codecheck to fail.

ppraegla:
The following test failed
```
2975 - fsi/tutorial_fsi_monolithic.4C.yaml-p2-fw (Timeout)
```
Seems like a random failure to me. The other re-runs of the test failed because the 4C build used for the test is no longer available. You need to run the 4C build again. But it will automatically run when you fix the codecheck.

ppraegla (approved):

georghammerl (approved):

slfuchs (approved):

PR #2146, opened 2026-07-30
Extended documentation for solid materials

ischeider:
## Description and Context

The analysis guide is being extended by a description of the solid material models. The added doc includes documentation for all
model representing 

- elasticity
- plasticity
- viscoelasticity
- hyperelastic framework
- viscohyperelastic framework

No Mixture or Growth models are included so far (because I have no clue about them ;-) )

The LLM, which looked up for the available models and respective equations for me, also identified flaws in the output of two material models, which will be fixed in a separate PR.

### DIFF
--- a/doc/documentation/CMakeLists.txt
+++ b/doc/documentation/CMakeLists.txt
@@ -56,6 +56,24 @@ if(FOUR_C_ENABLE_DOCUMENTATION)
       ${PROJECT_SOURCE_DIR}/tests/input_files/constr2D_MPC_angle.4C.yaml
       ${PROJECT_SOURCE_DIR}/tests/input_files/elch_gaussian_hill_pbc_genalpha.4C.yaml
       ${PROJECT_SOURCE_DIR}/tests/input_files/rve3d_periodic_bcs.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/plastic_pressurisedcylinder.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/tsi_plastic_heating_monolithic.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_druckerprager_RetToCone.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_gtn_patch_test.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/plastic_necking_damage.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/tsi_pressurisedcylinder_robinson.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/plastic_necking_fbar_thrplast.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/plastic_necking_fbar_vcu.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_viscoplastic_no_yield_surface_1hex8.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_iso_viscoplast_refJC_log_timint.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/structure_bidomain_material_problem_torsion.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/tsi_lindilatation_geolin.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/f3_womersley.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/elasthyper_coupanisoexpo.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/elasthyper_isoogden.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/elasthyper_isoanisoexpo_DispersedTransverselyIsotropic_ost_h8.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/viscoAnisotropic_creep.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/viscoelasthyper.4C.yaml
       )
 
   set(_sphinx_FILES_FROM_TESTS_DIR "${_sphinx_OUT_DIR}/testfiles")
--- a/doc/documentation/CMakeLists.txt
+++ b/doc/documentation/CMakeLists.txt
@@ -74,6 +74,7 @@ if(FOUR_C_ENABLE_DOCUMENTATION)
       ${PROJECT_SOURCE_DIR}/tests/input_files/elasthyper_isoanisoexpo_DispersedTransverselyIsotropic_ost_h8.4C.yaml
       ${PROJECT_SOURCE_DIR}/tests/input_files/viscoAnisotropic_creep.4C.yaml
       ${PROJECT_SOURCE_DIR}/tests/input_files/viscoelasthyper.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_ogden_tca_hex.4C.yaml
       )
 
   set(_sphinx_FILES_FROM_TESTS_DIR "${_sphinx_OUT_DIR}/testfiles")
--- a/doc/documentation/CMakeLists.txt
+++ b/doc/documentation/CMakeLists.txt
@@ -56,6 +56,25 @@ if(FOUR_C_ENABLE_DOCUMENTATION)
       ${PROJECT_SOURCE_DIR}/tests/input_files/constr2D_MPC_angle.4C.yaml
       ${PROJECT_SOURCE_DIR}/tests/input_files/elch_gaussian_hill_pbc_genalpha.4C.yaml
       ${PROJECT_SOURCE_DIR}/tests/input_files/rve3d_periodic_bcs.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/plastic_pressurisedcylinder.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/tsi_plastic_heating_monolithic.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_druckerprager_RetToCone.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_gtn_patch_test.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/plastic_necking_damage.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/tsi_pressurisedcylinder_robinson.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/plastic_necking_fbar_thrplast.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/plastic_necking_fbar_vcu.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_viscoplastic_no_yield_surface_1hex8.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_iso_viscoplast_refJC_log_timint.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/structure_bidomain_material_problem_torsion.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/tsi_lindilatation_geolin.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/f3_womersley.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/elasthyper_isoogden.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/elasthyper_isoanisoexpo_DispersedTransverselyIsotropic_ost_h8.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/viscoAnisotropic_creep.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/visco_generalized_maxwell_etd_h8.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_ogden_tca_hex.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/cardiovascular0d_arterialproxdist_structure_direct_genalpha.4C.yaml
       )
 
   set(_sphinx_FILES_FROM_TESTS_DIR "${_sphinx_OUT_DIR}/testfiles")
--- a/doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
+++ b/doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
@@ -1,0 +1,150 @@
+{% set iso_stvenant_file = "solid_runtime_hex8.4C.yaml" %}
+{% set iso_stvenant = load_input_file(iso_stvenant_file) %}
+{% set ortho_stvenant_file = "structure_bidomain_material_problem_torsion.4C.yaml" %}
+{% set ortho_stvenant = load_input_file(ortho_stvenant_file) %}
+{% set thermo_stvenant_file = "tsi_lindilatation_geolin.4C.yaml" %}
+{% set thermo_stvenant = load_input_file(thermo_stvenant_file) %}
+{% set aaa_file = "f3_womersley.4C.yaml" %}
+{% set aaa = load_input_file(aaa_file) %}
+{% set elasthyper_file = "cardiovascular0d_arterialproxdist_structure_direct_genalpha.4C.yaml" %}
+{% set elasthyper = load_input_file(elasthyper_file) %}
+{% set ogden_tca_file = "mat_ogden_tca_hex.4C.yaml" %}
+{% set ogden_tca = load_input_file(ogden_tca_file) %}
+
+.. _elastic-materials:
+
+Elastic materials
+=================
+
+This page summarizes the elastic materials that can be used for 2D and 3D solid
+elements. Materials that are composed of hyperelastic energy summands are documented
+separately in :doc:`hyperelastic_framework`.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 18 42
+
+   * - Material
+     - Kinematics
+     - Application
+   * - :ref:`MAT_Struct_StVenantKirchhoff <elastic-stvenant>`
+     - Linear or nonlinear
+     - Isotropic St. Venant--Kirchhoff elasticity.
+   * - :ref:`MAT_Struct_StVenantKirchhoffOrthotropic <elastic-orthotropic-stvenant>`
+     - Linear or nonlinear
+     - Orthotropic St. Venant--Kirchhoff elasticity.
+   * - :ref:`MAT_Struct_ThermoStVenantK <elastic-thermo-stvenant>`
+     - Linear or nonlinear
+     - Temperature-dependent St. Venant--Kirchhoff elasticity.
+   * - :ref:`MAT_Struct_AAANeoHooke <elastic-aaa-neohooke>`
+     - Nonlinear
+     - Nonlinear arterial-wall elasticity.
+   * - :ref:`MAT_Ogden_TCA <elastic-ogden-tca>`
+     - Nonlinear
+     - Ogden hyperelasticity with independently controlled tension--compression asymmetry.
+   * - :ref:`MAT_ElastHyper <elastic-elasthyper>`
+     - Nonlinear
+     - General composition of hyperelastic energy summands.
+
+.. _elastic-stvenant:
+
+``MAT_Struct_StVenantKirchhoff``
+--------------------------------
+
+This isotropic material is linear in Green--Lagrange strain and second Piola--Kirchhoff stress.
+With ``KINEM: linear`` it represents standard small-strain elasticity. With
+``KINEM: nonlinear`` it retains the same constitutive law but includes geometric nonlinearity;
+the model is not suitable for arbitrary large strains because its energy loses physical
+realism under strong compression.
+
+The following material section is taken from ``{{ iso_stvenant_file }}``.
+See details about the parameters in :ref:`MAT_Struct_StVenantKirchhoff
+in the Input Parameter Reference <MATERIALS_MAT_Struct_StVenantKirchhoff>`.
+
+{{ section_dump(iso_stvenant, "MATERIALS") }}
+
+The material does not register model-specific Gauss-point output.
+
+.. _elastic-orthotropic-stvenant:
+
+``MAT_Struct_StVenantKirchhoffOrthotropic``
+-------------------------------------------
+
+This is the orthotropic counterpart of the St. Venant--Kirchhoff material. Three Young's moduli,
+three shear moduli, and three Poisson ratios define the material axes. It accepts both linear and
+nonlinear total-Lagrangian kinematics. The principal material directions are fixed to the
+Cartesian axes of the reference configuration; the material provides no input for rotating this
+orthotropic coordinate system. A different orientation therefore requires a correspondingly
+oriented reference mesh or an extension of the material model.
+
+The following material section is taken from ``{{ ortho_stvenant_file }}``.
+See details about the parameters in
+:ref:`MAT_Struct_StVenantKirchhoffOrthotropic in the Input Parameter Reference
+<MATERIALS_MAT_Struct_StVenantKirchhoffOrthotropic>`.
+
+<MATERIALS_MAT_Struct_StVenantKirchhoffOrthotropic>`.
+
+{{ section_dump(ortho_stvenant, "MATERIALS") }}
+
+The material does not register model-specific Gauss-point output.
+
+.. _elastic-thermo-stvenant:
+
+``MAT_Struct_ThermoStVenantK``
+------------------------------
+
+This thermoelastic St. Venant--Kirchhoff model adds isotropic thermal strain and allows a list of
+Young's moduli for temperature-dependent applications. It accepts linear and nonlinear
+total-Lagrangian kinematics.
+
+The following material section is taken from
+``{{ thermo_stvenant_file }}``. See details about the parameters in
+:ref:`MAT_Struct_ThermoStVenantK in the Input Parameter Reference <MATERIALS_MAT_Struct_ThermoStVenantK>`.
+
+{{ section_dump(thermo_stvenant, "MATERIALS") }}
+
+.. _elastic-aaa-neohooke:
+
+``MAT_Struct_AAANeoHooke``
+--------------------------
+
+This nonlinear isotropic material implements the arterial-wall model of Raghavan and Vorp. It
+requires ``KINEM: nonlinear`` and is intended for nearly incompressible soft-tissue response.
+
+The following material section is taken from ``{{ aaa_file }}``.
+See details about the parameters in
+:ref:`MAT_Struct_AAANeoHooke in the Input Parameter Reference
+<MATERIALS_MAT_Struct_AAANeoHooke>`.
+
+{{ section_dump(aaa, "MATERIALS") }}
+
+.. _elastic-ogden-tca:
+
+``MAT_Ogden_TCA``
+-----------------
+
+``MAT_Ogden_TCA`` is a nonlinear isotropic hyperelastic solid material representing the
+tension--compression asymmetric Ogden model of
+`Moerman et al. (2016) <https://doi.org/10.1016/j.jmbbm.2015.11.027>`_,
+where the tension-compression asymmetry is controlled by the ``Q`` parameter.
+
+The two materials in the following excerpt from ``{{ ogden_tca_file }}`` differ only in ``Q`` and
+therefore demonstrate opposite tension-compression asymmetry. See more details of the parameters
+in the :ref:`Input Parameter Reference for MAT_Ogden_TCA <MATERIALS_MAT_Ogden_TCA>`:
+
+{{ section_dump(ogden_tca, "MATERIALS") }}
+
+.. _elastic-elasthyper:
+
+``MAT_ElastHyper``
+------------------
+
+``MAT_ElastHyper`` is the general finite-strain elastic framework. It sums separately defined
+``ELAST_*`` energy contributions and requires ``KINEM: nonlinear``;
+see :ref:`MAT_ElastHyper in the Input Parameter Reference <MATERIALS_MAT_ElastHyper>` and
+:doc:`hyperelastic_framework` for further parameters, composition rules and the summand catalog.
+The following material section is taken from ``{{ elasthyper_file }}``:
+
+{{ section_dump(elasthyper, "MATERIALS") }}
+
+Anisotropic combinations can provide fiber directions and summand-specific visualization data.
--- a/doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
+++ b/doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
@@ -6,15 +6,17 @@
 {% set aaa = load_input_file(aaa_file) %}
 {% set elasthyper_file = "elasthyper_coupanisoexpo.4C.yaml" %}
 {% set elasthyper = load_input_file(elasthyper_file) %}
+{% set ogden_tca_file = "mat_ogden_tca_hex.4C.yaml" %}
+{% set ogden_tca = load_input_file(ogden_tca_file) %}
 
 .. _elastic-materials:
 
 Elastic materials
 =================
 
-This page summarizes the elastic materials that are assigned directly to three-dimensional solid
-elements. Hyperelastic energy summands are documented separately in
-:doc:`hyperelastic_framework`.
+This page summarizes the elastic materials that are assigned directly to 2D and 3D solid
+elements. Materials that are composed of hyperelastic energy summands are documented
+separately in :doc:`hyperelastic_framework`.
 
 .. list-table::
    :header-rows: 1
@@ -35,6 +37,9 @@ elements. Hyperelastic energy summands are documented separately in
    * - :ref:`MAT_Struct_AAANeoHooke <elastic-aaa-neohooke>`
      - Nonlinear
      - Nonlinear arterial-wall elasticity.
+   * - :ref:`MAT_Ogden_TCA <elastic-ogden-tca>`
+     - Nonlinear
+     - Ogden hyperelasticity with independently controlled tension--compression asymmetry.
    * - :ref:`MAT_ElastHyper <elastic-elasthyper>`
      - Nonlinear
      - General composition of hyperelastic energy summands.
@@ -110,6 +115,41 @@ ratio and density. See :ref:`MAT_Struct_AAANeoHooke in the Input Parameter Refer
 ``{{ aaa_file }}``:
 
 {{ section_dump(aaa, "MATERIALS") }}
+
+.. _elastic-ogden-tca:
+
+``MAT_Ogden_TCA``
+-----------------
+
+``MAT_Ogden_TCA`` is a nonlinear isotropic hyperelastic solid material representing the
+tension--compression asymmetric Ogden model of
+`Moerman et al. (2016) <https://doi.org/10.1016/j.jmbbm.2015.11.027>`_:
+
+.. math::
+
+   \Psi ={}& \frac{c}{m^2}\sum_{i=1}^{3}
+      \left[q\left(\lambda_i^m-1\right)
+      +(1-q)\left(\lambda_i^{-m}-1\right)\right] \\
+      &+\frac{c}{m}(1-2q)\ln J+\frac{\kappa}{2}(J-1)^2 ,
+
+where :math:`\lambda_i` are the principal stretches and :math:`J=\det\boldsymbol F`.
+``C`` (:math:`c>0`) scales the material stiffness and ``M`` (:math:`m>0`) controls its
+nonlinearity. ``Q`` (:math:`q\in[0,1]`) independently controls tension--compression asymmetry:
+
+- :math:`q=0.5` gives a tension--compression symmetric response,
+- :math:`q>0.5` makes the material stiffer in tension than in compression, and
+- :math:`q<0.5` makes the material stiffer in compression than in tension.
+
+``KAPPA`` (:math:`\kappa>0`) is the volumetric penalty parameter. Larger values produce a more
+nearly incompressible response but do not enforce exact incompressibility and may cause poor
+conditioning or volumetric locking. ``DENS`` specifies the non-negative density. The material
+requires nonlinear total-Lagrangian kinematics.
+
+See :ref:`MAT_Ogden_TCA in the Input Parameter Reference <MATERIALS_MAT_Ogden_TCA>`. The two
+materials in the following excerpt from ``{{ ogden_tca_file }}`` differ only in ``Q`` and
+therefore demonstrate opposite asymmetry:
+
+{{ section_dump(ogden_tca, "MATERIALS") }}
 
 .. _elastic-elasthyper:
 
--- a/doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
+++ b/doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
@@ -1,0 +1,130 @@
+{% set stvenant_file = "structure_bidomain_material_problem_torsion.4C.yaml" %}
+{% set stvenant = load_input_file(stvenant_file) %}
+{% set thermo_stvenant_file = "tsi_lindilatation_geolin.4C.yaml" %}
+{% set thermo_stvenant = load_input_file(thermo_stvenant_file) %}
+{% set aaa_file = "f3_womersley.4C.yaml" %}
+{% set aaa = load_input_file(aaa_file) %}
+{% set elasthyper_file = "elasthyper_coupanisoexpo.4C.yaml" %}
+{% set elasthyper = load_input_file(elasthyper_file) %}
+
+.. _elastic-materials:
+
+Elastic materials
+=================
+
+This page summarizes the elastic materials that are assigned directly to three-dimensional solid
+elements. Hyperelastic energy summands are documented separately in
+:doc:`hyperelastic_framework`.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 18 42
+
+   * - Material
+     - Kinematics
+     - Application
+   * - :ref:`MAT_Struct_StVenantKirchhoff <elastic-stvenant>`
+     - Linear or nonlinear
+     - Isotropic St. Venant--Kirchhoff elasticity.
+   * - :ref:`MAT_Struct_StVenantKirchhoffOrthotropic <elastic-orthotropic-stvenant>`
+     - Linear or nonlinear
+     - Orthotropic St. Venant--Kirchhoff elasticity.
+   * - :ref:`MAT_Struct_ThermoStVenantK <elastic-thermo-stvenant>`
+     - Linear or nonlinear
+     - Temperature-dependent St. Venant--Kirchhoff elasticity.
+   * - :ref:`MAT_Struct_AAANeoHooke <elastic-aaa-neohooke>`
+     - Nonlinear
+     - Nonlinear arterial-wall elasticity.
+   * - :ref:`MAT_ElastHyper <elastic-elasthyper>`
+     - Nonlinear
+     - General composition of hyperelastic energy summands.
+
+.. _elastic-stvenant:
+
+``MAT_Struct_StVenantKirchhoff``
+--------------------------------
+
+This isotropic material is linear in Green--Lagrange strain and second Piola--Kirchhoff stress.
+With ``KINEM: linear`` it represents standard small-strain elasticity. With
+``KINEM: nonlinear`` it retains the same constitutive law but includes geometric nonlinearity;
+the model is not suitable for arbitrary large strains because its energy loses physical
+realism under strong compression.
+
+``YOUNG``, ``NUE``, and ``DENS`` define the material. See
+:ref:`MAT_Struct_StVenantKirchhoff in the Input Parameter Reference
+<MATERIALS_MAT_Struct_StVenantKirchhoff>`. The following material section is taken from
+``{{ stvenant_file }}``:
+
+{{ section_dump(stvenant, "MATERIALS") }}
+
+The material does not register model-specific Gauss-point output.
+
+.. _elastic-orthotropic-stvenant:
+
+``MAT_Struct_StVenantKirchhoffOrthotropic``
+-------------------------------------------
+
+This is the orthotropic counterpart of the St. Venant--Kirchhoff material. Three Young's moduli,
+three shear moduli, and three Poisson ratios define the material axes. It accepts both linear and
+nonlinear total-Lagrangian kinematics. The principal material directions are fixed to the
+Cartesian axes of the reference configuration; the material provides no input for rotating this
+orthotropic coordinate system. A different orientation therefore requires a correspondingly
+oriented reference mesh or an extension of the material model.
+
+The vectors ``YOUNG``, ``SHEAR``, and ``NUE`` contain the directional elastic constants;
+``DENS`` sets density. See
+:ref:`MAT_Struct_StVenantKirchhoffOrthotropic in the Input Parameter Reference
+<MATERIALS_MAT_Struct_StVenantKirchhoffOrthotropic>`. Its definition is included in
+``{{ stvenant_file }}`` above.
+
+The material does not register model-specific Gauss-point output.
+
+.. _elastic-thermo-stvenant:
+
+``MAT_Struct_ThermoStVenantK``
+------------------------------
+
+This thermoelastic St. Venant--Kirchhoff model adds isotropic thermal strain and allows a list of
+Young's moduli for temperature-dependent applications. It accepts linear and nonlinear
+total-Lagrangian kinematics.
+
+``YOUNGNUM`` and ``YOUNG`` define the elastic-modulus data. ``NUE``, ``DENS``, ``THEXPANS``, and
+``INITTEMP`` specify Poisson's ratio, density, thermal expansion, and reference temperature. See
+:ref:`MAT_Struct_ThermoStVenantK in the Input Parameter Reference
+<MATERIALS_MAT_Struct_ThermoStVenantK>`. The following material section is taken from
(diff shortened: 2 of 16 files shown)


doc/documentation/src/analysis_guide_templates/viscoelastic_materials.rst.j2
```diff
@@ -0,0 +1,138 @@
...
+       =\dot{\boldsymbol S}^{\,e}_i`. Each
+       :math:`\boldsymbol S^{\,e}_i` comes from the elastic material referenced by its branch.
+   * - :ref:`VISCO_GeneralizedMaxwellBranch
+       <MATERIALS_VISCO_GeneralizedMaxwellBranch>`
+     - Branch referenced by a generalized Maxwell model.
+
+       No independent equation; ``MATID`` defines
+       :math:`\boldsymbol S^{\,e}_i` and ``TAU`` defines :math:`\tau_i` in the parent Maxwell law.
+   * - :ref:`VISCO_QuasiLinearGeneralizedMaxwell
+       <MATERIALS_VISCO_QuasiLinearGeneralizedMaxwell>`
```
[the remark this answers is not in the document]
ischeider: My github copilot respondsd to this recommendation:
No change is technically required. Multiline inline roles are valid reStructuredText; Sphinx 7.2.6/docutils 0.20.1 resolves this exact multiline :ref: successfully even with warnings treated as errors.
You can reply that the recommendation is incorrect. Reformatting may still appease a linter or improve readability, but it is not needed for a valid documentation build.

doc/documentation/src/analysis_guide_templates/plasticity_materials.rst.j2
```diff
@@ -0,0 +1,685 @@
...
+
+Input
+~~~~~
+
+The material requires ``KINEM: linear``. ``YOUNG``, ``NUE``, and ``DENS`` define elasticity and
+density. ``YIELD`` is the initial yield stress, ``ISOHARD`` and ``KINHARD`` are the isotropic and
+kinematic hardening moduli, and ``TOL`` controls the local constitutive iteration.
+
+See :ref:`MAT_Struct_PlasticLinElast in the Input Parameter Reference
+<MATERIALS_MAT_Struct_PlasticLinElast>` for the complete schema. The following definition is
```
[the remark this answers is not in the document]
ischeider: see above

doc/documentation/CMakeLists.txt
```diff
@@ -56,6 +56,24 @@ if(FOUR_C_ENABLE_DOCUMENTATION)
...
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_viscoplastic_no_yield_surface_1hex8.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/mat_iso_viscoplast_refJC_log_timint.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/structure_bidomain_material_problem_torsion.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/tsi_lindilatation_geolin.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/f3_womersley.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/elasthyper_coupanisoexpo.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/elasthyper_isoogden.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/elasthyper_isoanisoexpo_DispersedTransverselyIsotropic_ost_h8.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/viscoAnisotropic_creep.4C.yaml
+      ${PROJECT_SOURCE_DIR}/tests/input_files/viscoelasthyper.4C.yaml
```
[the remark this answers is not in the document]
ischeider: false positive. All test files are contained.

doc/documentation/src/analysis_guide_templates/plasticity_materials.rst.j2
```diff
@@ -0,0 +1,685 @@
...
+~~~~~
+
+The material requires ``KINEM: linear``. In addition to ``YOUNG``, ``NUE``, and ``DENS``, specify
+the thermal expansion coefficient ``THEXPANS`` and reference temperature ``INITTEMP``.
+``YIELD``, ``ISOHARD``, and ``KINHARD`` define linear hardening. For piecewise-linear isotropic
+hardening, ``SAMPLENUM`` sets the number of entries in ``SIGMA_Y`` and ``EPSBAR_P``. ``TOL``
+controls the local iteration.
+
+See :ref:`MAT_Struct_ThermoPlasticLinElast in the Input Parameter Reference
+<MATERIALS_MAT_Struct_ThermoPlasticLinElast>` for the complete schema. The following definition
```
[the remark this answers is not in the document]
ischeider: see above

doc/documentation/src/analysis_guide_templates/viscoelastic_materials.rst.j2
```diff
@@ -0,0 +1,138 @@
...
+   :widths: 35 65
+
+   * - Summand
+     - Model
+   * - :ref:`VISCO_IsoRateDep <MATERIALS_VISCO_IsoRateDep>`
+     - Isochoric rate-dependent contribution.
+
+       :math:`\Phi_\mathrm{v}=n\,\bar J_2(\bar I_1-3)`, with
+       :math:`\bar J_2=\frac12\dot{\bar{\boldsymbol C}}:
+       \dot{\bar{\boldsymbol C}}`. The rate is evaluated by a backward difference.
```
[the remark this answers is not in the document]
ischeider: False positive. Sphinx accepts an inline :math: role across physical lines when indentation remains within the same paragraph; the newline becomes harmless TeX whitespace.

ischeider:
Added the brand new material model by @lauraengelhardt  ;-)

doc/documentation/src/analysis_guide_templates/materials.rst.j2
lauraengelhardt: I think this should actually be `Solid material models`.
I know we are not really concise with the distinction between solid and structural mechanics in 4C.

doc/documentation/src/analysis_guide_templates/materials.rst.j2
```diff
@@ -51,9 +51,134 @@ Structural Material Models
...
+- :doc:`hyperelastic_framework` (hyperelastic materials composed of several strain-energy terms)
+- :doc:`viscoelastic_materials`
+- :doc:`plasticity_materials`
+
+Material directions
+~~~~~~~~~~~~~~~~~~~
+
+Direction-dependent materials usually obtain their material directions from element fiber
+vectors. For a native |FOURC| mesh, append consecutive ``FIBER1``, ``FIBER2``, and, where
+required, ``FIBER3`` vectors to each entry in ``STRUCTURE ELEMENTS``:
```
lauraengelhardt:
```suggestion
vectors. For a native |FOURC| mesh, append consecutive ``FIBER1``, and, where required ``FIBER2``, and ``FIBER3`` vectors to each entry in ``STRUCTURE ELEMENTS``:
```

doc/documentation/src/analysis_guide_templates/materials.rst.j2
```diff
@@ -51,9 +51,134 @@ Structural Material Models
...
 A material model in structural mechanics defines the connection between deformation (usually strain) and stress.
 Many material models are available, including (hyper-)elastic, elasto-plastic, visco-elastic, visco-plastic, and even damage models.
+The following pages describe the principal continuum material families:
+
+- :doc:`elastic_materials`
+- :doc:`hyperelastic_framework` (hyperelastic materials composed of several strain-energy terms)
+- :doc:`viscoelastic_materials`
+- :doc:`plasticity_materials`
+
+Material directions
```
lauraengelhardt: Everything written in this section only applies to materials that rely on the `Anisotropy` extension. There are several materials (the muscle materials and all mixture materials) that rely on fiber input through a vtu file (using `FiberInterpolation`). Maybe you can start the section somewhat like this: 
- how anisotropy is handled depends on the material itself
- there are two approaches:  `Anisotropy` and `FiberInterpolation`
- for `Anisotropy` the way to input is through native 4C or constant in the element block (just as you wrote)
- for `FiberInterpolation` I can supply a short description after ;)

doc/documentation/src/analysis_guide_templates/materials.rst.j2
```diff
@@ -51,9 +51,134 @@ Structural Material Models
...
+       - NAME: anisotropic_solid
+         SOLID:
+           HEX8:
+             MAT: 1
+             KINEM: nonlinear
+             FIBER1: [1.0, 0.0, 0.0]
+             FIBER2: [0.0, 1.0, 0.0]
+
+This applies the specified directions to the complete block. No regression input currently
+exercises fiber assignment through ``STRUCTURE GEOMETRY``.
```
lauraengelhardt: What do you mean with "No regression input"`?

doc/documentation/src/analysis_guide_templates/materials.rst.j2
```diff
@@ -51,9 +51,134 @@ Structural Material Models
...
+
+As an alternative, solid elements accept the local coordinate-system vectors ``RAD``, ``AXI``,
+and ``CIR``. All three must be given together. For materials using the default anisotropy
+framework, this coordinate system takes precedence over explicit ``FIBER*`` vectors. Some
```
lauraengelhardt: > For materials using the default anisotropy
framework, this coordinate system takes precedence over explicit ``FIBER*`` vectors

I'm not so sure about this. Do you mean when you give both then RAD/AXI/CIR is used?

doc/documentation/src/analysis_guide_templates/materials.rst.j2
```diff
@@ -51,9 +51,134 @@ Structural Material Models
...
+     - Direction source and restrictions
+   * - ``ELAST_CoupAnisoExpo``
+     - 1
+     - Element fiber or ``RAD``/``AXI``/``CIR``; ``FIBER_ID`` selects the element fiber.
+   * - ``ELAST_CoupAnisoNeoHooke``
+     - 1
+     - Element fiber or ``RAD``/``AXI``/``CIR``.
+   * - ``ELAST_CoupAnisoNeoHooke_VarProp``
+     - 1
+     - Element fiber or ``RAD``/``AXI``/``CIR``.
```
lauraengelhardt: Can we use a different name for ``RAD``/``AXI``/``CIR``? Or use  ``FIBER`` instead of Element fiber to make it symmetric?

doc/documentation/src/analysis_guide_templates/materials.rst.j2
```diff
@@ -51,9 +51,134 @@ Structural Material Models
...
+This applies the specified directions to the complete block. No regression input currently
+exercises fiber assignment through ``STRUCTURE GEOMETRY``.
+
+As an alternative, solid elements accept the local coordinate-system vectors ``RAD``, ``AXI``,
+and ``CIR``. All three must be given together. For materials using the default anisotropy
+framework, this coordinate system takes precedence over explicit ``FIBER*`` vectors. Some
+specialized materials only support one of these input methods; consult the table below and the
+material-specific Input Parameter Reference.
+
+.. list-table:: Direction requirements of the continuum materials documented in this guide
```
lauraengelhardt: As this list is not complete and will become outdated really quickly, I would rather have the information in the documentation of each material/summand.

doc/documentation/src/analysis_guide_templates/materials.rst.j2
```diff
@@ -51,9 +51,134 @@ Structural Material Models
 
 A material model in structural mechanics defines the connection between deformation (usually strain) and stress.
 Many material models are available, including (hyper-)elastic, elasto-plastic, visco-elastic, visco-plastic, and even damage models.
+The following pages describe the principal continuum material families:
+
+- :doc:`elastic_materials`
+- :doc:`hyperelastic_framework` (hyperelastic materials composed of several strain-energy terms)
+- :doc:`viscoelastic_materials`
+- :doc:`plasticity_materials`
```
lauraengelhardt: If I'm not mistaken then the only possibility to go to these pages is through these links. Would it be possible to have these subsections also appear in the navigation menu on the left?

doc/documentation/src/analysis_guide_templates/materials.rst.j2
lauraengelhardt: Just one question here as well: I assume we are aiming to document all materials that derive from `So3Material`, correct?

lauraengelhardt: I know you didn't touch this section but this only applies to certain collections - the MIXTURE e.g., does not need the NUMMAT. Not sure if we want to improve this in this PR though ;)

doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
```diff
@@ -0,0 +1,170 @@
...
+{% set elasthyper = load_input_file(elasthyper_file) %}
+{% set ogden_tca_file = "mat_ogden_tca_hex.4C.yaml" %}
+{% set ogden_tca = load_input_file(ogden_tca_file) %}
+
+.. _elastic-materials:
+
+Elastic materials
+=================
+
+This page summarizes the elastic materials that are assigned directly to 2D and 3D solid
```
lauraengelhardt:
```suggestion
This page summarizes the elastic materials that can be used for 2D and 3D solid
```

doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
```diff
@@ -0,0 +1,170 @@
...
+``KINEM: nonlinear`` it retains the same constitutive law but includes geometric nonlinearity;
+the model is not suitable for arbitrary large strains because its energy loses physical
+realism under strong compression.
+
+``YOUNG``, ``NUE``, and ``DENS`` define the material. See
+:ref:`MAT_Struct_StVenantKirchhoff in the Input Parameter Reference
+<MATERIALS_MAT_Struct_StVenantKirchhoff>`. The following material section is taken from
+``{{ stvenant_file }}``:
+
+{{ section_dump(stvenant, "MATERIALS") }}
```
lauraengelhardt: This shows an input example of the orthotropic SVK.

doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
```diff
@@ -0,0 +1,170 @@
...
+
+The vectors ``YOUNG``, ``SHEAR``, and ``NUE`` contain the directional elastic constants;
+``DENS`` sets density. See
+:ref:`MAT_Struct_StVenantKirchhoffOrthotropic in the Input Parameter Reference
+<MATERIALS_MAT_Struct_StVenantKirchhoffOrthotropic>`. Its definition is included in
+``{{ stvenant_file }}`` above.
+
+The material does not register model-specific Gauss-point output.
+
+.. _elastic-thermo-stvenant:
```
lauraengelhardt, line 91: @maxfirmbach I think you implemented this and could check? :)

doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
```diff
@@ -0,0 +1,170 @@
...
+
+This nonlinear isotropic material implements the arterial-wall model of Raghavan and Vorp. It
+requires ``KINEM: nonlinear`` and is intended for nearly incompressible soft-tissue response.
+
+``YOUNG`` and ``BETA`` are its constitutive parameters; ``NUE`` and ``DENS`` define Poisson's
+ratio and density. See :ref:`MAT_Struct_AAANeoHooke in the Input Parameter Reference
+<MATERIALS_MAT_Struct_AAANeoHooke>`. The following material section is taken from
+``{{ aaa_file }}``:
+
+{{ section_dump(aaa, "MATERIALS") }}
```
lauraengelhardt, line 119: Regarding the examples: I think it is not necessary to put the full material section with other materials defined as well in there. Do you do that because like this you can directly pull this from the input file such that the info doesnt become outdated when, e.g. the parameter names would be changed?

bgoderbauer (dismissed):
Thanks a lot for improving the documentation! I've mainly looked into viscoelastic materials, which I think provide just the right amount of user-facing information.

doc/documentation/src/analysis_guide_templates/materials.rst.j2
lauraengelhardt: Upon looking  further I would move these explicit syntax definitions to the respective frameworks e.g. Elast_Hyper. Idk if its possible to even do that automatically (s.t. e.g. if NUMMAT is removed for Elast_Hyper the docu would just update automatically).

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+
+One material is the *parent material*, which is assigned to the element.
+This material entry does not define its properties (except the density of the material), but contains
+pointers to the energy summands.
+Its children, denoted by the parameter ``MATIDS``, refer to the material IDs of these summands (see example below).
+
+Choosing summands
+-----------------
+
+There are two common ways to construct an isotropic response:
```
lauraengelhardt: I suppose you mean a nearly incompressible response?

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+
+- A **coupled** contribution depends on the full deformation and already contains distortional
+  and volumetric response.
+- A **split** formulation usually combines one or more isochoric ``ELAST_Iso*`` contributions
+  with one ``ELAST_Vol*`` contribution.
+
+Do not add an extra volumetric penalty to a coupled compressible law unless that additional
+volumetric stiffness is intentional. Anisotropic summands are normally added to an isotropic
+matrix and may reference fiber directions or an ``ELAST_StructuralTensor`` material.
```
lauraengelhardt:
```suggestion
volumetric stiffness is intentional.
```

Coupled vs. split only relates to incompressibility, not isotropy. I would only keep this paragraph about that. If the material is iso- vs. anisotropic is then a different thing.

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+  several terms unintentionally over-stiffens or otherwise changes the volumetric response.
+- Isochoric and coupled summands may be combined. The coupled summand already contributes both
+  distortional and volumetric response, while the isochoric term adds further distortional
+  stiffness. The result is a valid additive energy, but no longer a conventional
+  isochoric--volumetric split and should be calibrated as a combined model.
+- No isochoric summand is required. A coupled isotropic law alone is already a complete
+  compressible material, and it may also serve as the matrix for anisotropic fiber terms. A model
+  containing only volumetric terms has no shear stiffness; similarly, an anisotropic-only model
+  may leave deformation modes unsupported. Such incomplete combinations are accepted by the
+  framework but are normally not suitable as standalone solid materials.
```
lauraengelhardt: I think this is a lot of words for simply saying the user needs to take care by him/herself to either stick to a split formulation or a coupled formulation (and the framework will not prevent him/her from doing unreasonable combinations ;))

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+.. note::
+
+    A sufficiently large penalty parameter or bulk modulus keeps :math:`J` close to one, but does not
+    enforce exact incompressibility. Increasing it excessively can lead to poor conditioning and
+    volumetric locking. Suitable EAS or F-bar element technology can alleviate locking for nearly
+    incompressible displacement-based solids, but does not itself impose :math:`J=1`.
+
+The plain ``MAT_ElastHyper`` framework accepts elastic summands only. Use
+:doc:`viscoelastic_materials` when adding time-dependent branches.
+
```
lauraengelhardt, line 125:
```suggestion
```

Has nothing to do here I think.

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+.. math::
+
+   U(J)=\epsilon\left(J^\gamma+J^{-\gamma}-2\right).
+
+.. note::
+
+    A sufficiently large penalty parameter or bulk modulus keeps :math:`J` close to one, but does not
+    enforce exact incompressibility. Increasing it excessively can lead to poor conditioning and
+    volumetric locking. Suitable EAS or F-bar element technology can alleviate locking for nearly
+    incompressible displacement-based solids, but does not itself impose :math:`J=1`.
```
lauraengelhardt, line 121: I think there are some mistakes in here (e.g. it should be `\Psi=\Psi (\boldsymbol C)+p(J-1) (non-modified C for a fully incompressible formulation (see e.g. Holzapfel Eq 6.57)). 

But I think this is a nice start for a short explanation of the two ways to achieve near-incompressibility. 
- Starting from a fully incompressible formulation `\Psi=\Psi(\boldsymbol C)+p(J-1)`
- Computing `p` is computationally cumbersome, so in FE codes we usually use near incompressible formations 
  - **Option 1**: volumetric-isochoric split of C = J^2/3 \bar{C} → Formulate two dedicated contriubutions such that \Psi(\boldsymbol C) =\Psi_\mathrm{iso}(\bar {\boldsymbol C})+\Psi_\mathrm{vol}(J) 
  - **Option 2**: coupled formulation \Psi(\boldsymbol C) that already includes the volumetric penalty term


Maybe this could then even be moved further up as this is somewhat the basics for the summands.

rjoussen (changes requested):
I really like this direction and the initiative to add more thorough documentation for the materials! We definitely need more documentation on them, and especially putting them into the different categories that you defined makes a lot of sense and really helps with finding/selecting the right material.

The main problem I am seeing with this approach, though, is that it adds a giant second source of truth. The material descriptions and categories might be correct and up to date right now, but they will inevitably drift out of sync with the codebase over time. With this, almost every change to the material code will then also require changes to the documentation (e.g., deleting or adding a material, changing parameter names, enabling a material for different kinematics, adding coupling, ...).

On top of that, I think we already have a place for material documentation in the schema descriptions, which are also directly useful when writing YAML input files. I think much of the information added here could instead live there (because these descriptions are often way too short)

Regarding the categorization of materials, I think a safer approach would be to either put them in subfolders in the codebase and use that structure directly in the documentation (I never understood why all ~270 materials are in the same folder anyway), or to use some sort of a tagging system (like tag materials with their supported kinematics, what category they are, for what physics, etc.) and then use these tags to generate the documentation. Such tags could live closer to the actual implementation and are hence easier to maintain.

So, to sum up, I definitely agree that this additional documentation would be a massive improvement right now, but I also think it has the potential to become outdated very quickly and, as a result, become counterproductive.

Of course, I am also open to discussing this, those are just my personal thoughts on it :)

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+
+.. math::
+
+   \boldsymbol C &= \boldsymbol F^\mathrm{T}\boldsymbol F,
+   &\boldsymbol E &= \tfrac12(\boldsymbol C-\boldsymbol I),
+   &J &= \det\boldsymbol F,\\
+   I_1 &= \operatorname{tr}\boldsymbol C,
+   &I_2 &= \tfrac12\left[(\operatorname{tr}\boldsymbol C)^2
+     -\operatorname{tr}(\boldsymbol C^2)\right],
+   &I_3 &= \det\boldsymbol C=J^2.
```
lauraengelhardt: The formating seems a bit off here.

Also can we maybe introduce C, then the invariants and then the principal stretches? Its a bit mixed here.

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+
+The principal stretches :math:`\lambda_i` are the positive square roots of the eigenvalues of
+:math:`\boldsymbol C`.
+
+Isochoric quantities carry an overbar:
+
+.. math::
+
+   \bar{\boldsymbol C}=J^{-2/3}\boldsymbol C,\qquad
```
lauraengelhardt: I know this as "modified" quantities, but maybe I'm wrong ;)

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+   I_4 &= \boldsymbol a\mathbin{\cdot}\boldsymbol C\boldsymbol a,
+   &I_5 &= \boldsymbol a\mathbin{\cdot}\boldsymbol C^2\boldsymbol a,\\
+   I_6 &= \boldsymbol a_2\mathbin{\cdot}\boldsymbol C\boldsymbol a_2,
+   &I_8 &= \tfrac12(\boldsymbol a_1\otimes\boldsymbol a_2+
+     \boldsymbol a_2\otimes\boldsymbol a_1):\boldsymbol C .
+
+The symbols in the equations correspond to the similarly named input parameters; for example,
+:math:`c_1`, :math:`k_1`, and :math:`\kappa` denote ``C1``, ``K1``, and ``KAPPA``. Additive
+constants are retained where they are part of the implementation, although they do not affect
+stress.
```
lauraengelhardt:
```suggestion

```

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+
+    A sufficiently large penalty parameter or bulk modulus keeps :math:`J` close to one, but does not
+    enforce exact incompressibility. Increasing it excessively can lead to poor conditioning and
+    volumetric locking. Suitable EAS or F-bar element technology can alleviate locking for nearly
+    incompressible displacement-based solids, but does not itself impose :math:`J=1`.
+
+The plain ``MAT_ElastHyper`` framework accepts elastic summands only. Use
+:doc:`viscoelastic_materials` when adding time-dependent branches.
+
+Notation for the potentials
```
lauraengelhardt: Maybe just make this one consistent section with the split/coupled above "Basics of hyperelastic continuum mechanical material modeling" or so?

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
```diff
@@ -0,0 +1,157 @@
...
+   hyperelastic_anisotropic
+
+Example
+---------
+
+Here, the parent materials, which consist of two coupled material parts each, are assigned to the element.
+They combine a compressible isotropic matrix with an anisotropic fiber contribution and its structural-tensor definition.
+Their ``MATIDS``, that is, the material laws of the summation parts, refer to the material IDs of the energy summands,
+which are given as separate material IDs in the ``MATERIALS`` section.
+The example is taken from ``{{ framework_file }}``.
```
lauraengelhardt, line 143: Also here a very long example... Can we just use the MAT 1 part?

doc/documentation/src/analysis_guide_templates/hyperelastic_isotropic_coupled.rst
```diff
@@ -0,0 +1,79 @@
+.. _hyperelastic-isotropic-coupled:
+
+Coupled isotropic hyperelastic summands
+=======================================
+
+These summands define compressible isotropic energies that depend on the full deformation. They
+are referenced through ``MAT_ElastHyper.MATIDS`` and generally do not require a separate
```
lauraengelhardt: ``MAT_ElastHyper.MATIDS`` ?

doc/documentation/src/analysis_guide_templates/materials.rst.j2
```diff
@@ -51,9 +51,134 @@ Structural Material Models
...
+     - None
+     - Its three orthotropic axes are fixed to the Cartesian reference axes and cannot be rotated
+       through material or fiber input.
+
+``MAT_ElastHyper`` and ``MAT_ViscoElastHyper`` do not impose a direction count themselves; their
+requirements are determined by the selected summands. A ``STR_TENS_ID`` controls the
+structural-tensor strategy for several anisotropic summands but does not replace the underlying
+mean fiber direction.
+
 If you wish to implement a new material model, this is of course also possible, it is an in-house code after all.
```
lauraengelhardt:
```suggestion

```

doc/documentation/src/analysis_guide_templates/hyperelastic_isotropic_coupled.rst
```diff
@@ -0,0 +1,79 @@
...
+volumetric summand.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 35 65
+
+   * - Summand
```
lauraengelhardt, line 15: I did not check all equations in detail yet. What I am wondering: Is there any way how we can ensure that this does not become outdated?

lauraengelhardt (changes requested):
Thanks a lot, Ingo, for adding the documentation. I took a look at the hyperelastic materials. Someone else will have to check plasticity and viscoelasticity ;)

doc/documentation/src/analysis_guide_templates/hyperelastic_isotropic_coupled.rst
lauraengelhardt: @rjoussen I had the same thoughts as you had here as well!
I think introducing the general material subclasses is a good idea as they will not become outdated so fast (as I mentioned above, assuming the examples are taken from the codebase anyhow, such that changes in the parameter names etc. are automatically included in the doc). 
If we want to include the actual materials we need a way to do this automatically.

I also agree with you @rjoussen: Moving the strain energy functions to the input parameter descriptions (as you recommended me to do for the TCA model) would be a good idea. In the documentation we could just link that? 

+ I also agree regarding moving materials to subfolders.

doc/documentation/src/analysis_guide_templates/materials.rst.j2
ischeider: It is still a subset of the `So3Material`, since I did not include the MIXTURE materials nor Muscle materials (maybe others also left out). Let say, these are (most of) the general purpose materials. But you are right, since the title is **Solid Material Models**, it should contain all materials derived from So3Material, or one has to scope the section.

ischeider: Here, I have one question: I would like to remove the NUMMAT parameter immediately, and leave it (as an optional parameter that is just used for backward compatibility), but not being used. I have the feeling that this is not the way 4C handles changes, but in this way, we don't have to touch many files, and it wouldn't be a "breaking change".

doc/documentation/src/analysis_guide_templates/materials.rst.j2
ischeider: If you give FIBERi and RAD/AXI/CIR, then the values for RAD|AXI|CIR are the relevant ones, the value for FIBERi are ignored. I tested it with an example where I entered fiber directions for one element by FIBER1/FIBER2 in addition to the existing RAD... values =>  it's true, the results don't change (and no error is thrown).

lauraengelhardt: Uff ok. IMO, this is rather a bug than intended behavior. I cannot imagine any scenario where defining both would make sense. I would rather open an issue and fix the problem than reporting it here. What do you think?

ischeider: Behavior changed in PR #2168 .

doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
ischeider: The problem here is that the `f3_womersley.4C.yaml` is the only example, where this material is used, and it's not possible to get a single material out of the list of materials (except by using a line-number-based excerpt of the file, which can get even more easily outdated).

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
ischeider: Actually I mean the full (not necessarily isotropic) behavior...

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
ischeider: I reduced this part significantly. However, I found it important to note that a material without a volumetric part does not produce an isochoric response; so I left this one in.

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
ischeider: No, unfortunately we cannot. But I have chosen another example.

doc/documentation/src/analysis_guide_templates/materials.rst.j2
ischeider: Bug fixed by PR #2168

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
ischeider: Here I don't agree completely. For incompressible materials of course C = \bar{C} should always hold. However, for compressible cases, it is rather what I have written, see also Holzapfel  eq. 6.85. 

Also, the coupled hyperelastic models in general don't include such penalty terms. Some of them are split strictly with separate invariants and separate parameters, so that the parameter for J can be used as a penalty term, but I wouldn't promote this method, but rather the explicit penalty form.

doc/documentation/src/analysis_guide_templates/hyperelastic_isotropic_coupled.rst
ischeider: I move all equations and detailed information to the respective materials in the reference guide. The material names in the analysis guide may be incomplete, but at least the names remain correct, since they have links to their respective reference (and the links break when the name changes).

doc/documentation/src/analysis_guide_templates/materials.rst.j2
ischeider: completely rephrased. However, the muscle etc. materials are still missing ;-)

doc/documentation/src/analysis_guide_templates/hyperelastic_framework.rst.j2
ischeider: Reordered the section. Maybe I could also remove the whole subseciton title "Notation for the potentials".

doc/documentation/src/analysis_guide_templates/plasticity_materials.rst.j2
```diff
@@ -0,0 +1,685 @@
...
+This small-strain material combines isotropic linear elasticity with associative von Mises
+plasticity. It supports linear isotropic and linear kinematic hardening. Use it for metals when
+geometric nonlinearities and temperature effects are negligible.
+
+Input
+~~~~~
+
+The material requires ``KINEM: linear``. ``YOUNG``, ``NUE``, and ``DENS`` define elasticity and
+density. ``YIELD`` is the initial yield stress, ``ISOHARD`` and ``KINHARD`` are the isotropic and
+kinematic hardening moduli, and ``TOL`` controls the local constitutive iteration.
```
lauraengelhardt: This, e.g. will become outdated/differ from the documentation in the schema file rather quickly.

doc/documentation/src/analysis_guide_templates/plasticity_materials.rst.j2
```diff
@@ -0,0 +1,685 @@
...
+by plastic dissipation. The isotropic hardening curve can be linear or piecewise linear.
+
+Input
+~~~~~
+
+The material requires ``KINEM: linear``. In addition to ``YOUNG``, ``NUE``, and ``DENS``, specify
+the thermal expansion coefficient ``THEXPANS`` and reference temperature ``INITTEMP``.
+``YIELD``, ``ISOHARD``, and ``KINHARD`` define linear hardening. For piecewise-linear isotropic
+hardening, ``SAMPLENUM`` sets the number of entries in ``SIGMA_Y`` and ``EPSBAR_P``. ``TOL``
+controls the local iteration.
```
lauraengelhardt: Same for this.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1816,7 +1855,14 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
         },
-        {.description = "Saint-Venant-Kirchhoff as elastic summand"});
+        {.description = "Saint-Venant-Kirchhoff as elastic summand. The strain energy is computed "
+                        "as\n\n"
+                        "$$\n"
+                        "\\Psi = \\mu\\,\\mathrm{tr}(\\mathbf{E}^2) + "
+                        "\\frac{\\lambda}{2}(\\mathrm{tr}\\,\\mathbf{E})^2\n"
+                        "$$\n\n"
+                        "with $\\mathbf{E}$ the Green-Lagrange strain tensor and $\\mu$, "
+                        "$\\lambda$ the Lame constants."});
```
lauraengelhardt, line 1901: Adding documentation to the description is the way to go, IMO.
Could we adapt this approach for the other materials where you currently added documentation to read the docs?

doc/documentation/src/analysis_guide_templates/elastic_materials.rst.j2
```diff
@@ -0,0 +1,163 @@
...
+- :math:`q>0.5` makes the material stiffer in tension than in compression, and
+- :math:`q<0.5` makes the material stiffer in compression than in tension.
+
+``KAPPA`` (:math:`\kappa>0`) is the volumetric penalty parameter. Larger values produce a more
+nearly incompressible response but do not enforce exact incompressibility and may cause poor
+conditioning or volumetric locking.
+
+See :ref:`MAT_Ogden_TCA in the Input Parameter Reference <MATERIALS_MAT_Ogden_TCA>`. The two
+materials in the following excerpt from ``{{ ogden_tca_file }}`` differ only in ``Q`` and
+therefore demonstrate opposite asymmetry:
```
lauraengelhardt: I really appreciate the effort of adding the docs here - but this is just redundant info from the schema file that is very prone to be outdated and adds an additional place where we need to manage/update things. I am very much with @rjoussen in terms of not adding duplicate information to keep stuff maintainable.

doc/documentation/src/analysis_guide_templates/hyperelastic_anisotropic.rst.j2
```diff
@@ -0,0 +1,65 @@
...
+     - Neo-Hookean reinforcement for one fiber.
+   * - :ref:`ELAST_CoupAnisoNeoHooke_VarProp <MATERIALS_ELAST_CoupAnisoNeoHooke_VarProp>`
+     - Neo-Hookean response for one fiber with variable properties.
+   * - :ref:`ELAST_CoupAnisoPow <MATERIALS_ELAST_CoupAnisoPow>`
+     - Power-law energy for one fiber.
+   * - :ref:`ELAST_CoupAnisoExpoActive <MATERIALS_ELAST_CoupAnisoExpoActive>`
+     - Exponential law for one fiber with active response.
+   * - :ref:`ELAST_AnisoActiveStress_Evolution
+       <MATERIALS_ELAST_AnisoActiveStress_Evolution>`
+     - Active-stress evolution law for one fiber.
```
lauraengelhardt: I am not much familiar with the technicalities of read the docs - but this (as you also commented) will become outdated rather quickly. 
Would it be possible to restructure the materials into different  folders in src (as @rjoussen) and then just add a reference to the respective subclass folder here (with the note that all materials belonging to that subclass can be found there)?

lauraengelhardt (changes requested):
Thanks a lot for incorporating suggestions and improving the previous version! 
I still have several concerns regarding the duplication of information from the schema file to here (as initially noted by @rjoussen), maintainability, etc.
I marked a few places with some thoughts on how this might be handled better.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
ischeider: I hope that I now got rid of all equations in the analysis guide that can go to the material reference.

doc/documentation/src/analysis_guide_templates/hyperelastic_anisotropic.rst.j2
ischeider: I totally agree. However, there is no  stringent formalism in the material definitions, such as a subclass for particular materials (such as `so3materials`. I would love to sort the materials into a structure, where they could be divided into physics modules, but at this time it's not easily possible.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1156,7 +1165,9 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
                                        "[nu12, nu23, nu13]."}),
             parameter<double>("DENS", {.description = "mass density"}),
         },
-        {.description = "St.Venant--Kirchhoff material with orthotropy"});
+        {.description = "St.Venant--Kirchhoff material with orthotropy. Direction requirements: "
```
rjoussen, line 1168: I find "Direction Requirements" hard to understand and do not really get what is meant by it.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1263,22 +1281,52 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                        "$$\n"
+                        "\\dot{f} = \\dot{f}_\\text{growth} + \\dot{f}_\\text{nucl}\n"
+                        "$$\n\n"
+                        "with\n\n"
+                        "$$\n"
+                        "\\dot{f}_\\text{nucl} = \\frac{f_N}{s_N \\sqrt{2\\pi}} "
+                        "\\exp \\left\\{ -\\frac{1}{2} "
+                        "\\left[ \\frac{\\overline{\\varepsilon}^{pl} - \\epsilon_N}{s_N} "
+                        "\\right]^2 \\right\\} "
+                        "\\dot{\\overline{\\varepsilon}}^{pl}\n"
```
rjoussen, line 1328:
```suggestion
                        "\\dot{\\overline{\\varepsilon}}^\\text{pl}\n"
```

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1263,22 +1281,52 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+            parameter<double>("FC", {.description = "GTN constant $f_C$: "
+                                                    "void volume fraction at damage coalescence"}),
+            parameter<double>("KAPPA",
+                {.description = "GTN constant $\\kappa$: Increased damage rate after coalescence"}),
             parameter<double>(
                 "EF", {.description = "GTN stabilization parameter ef for damage coalescence",
                           .default_value = 0.0}),
         },
-        {.description = "elastic St.Venant Kirchhoff / plastic GTN"});
+        {.description = "elastic St.Venant Kirchhoff / plastic GTN for porous metal plasticity."
```
rjoussen, line 1302:
```suggestion
        {.description = "elastic St.Venant Kirchhoff / plastic Gurson-Tvergaard-Needleman (GTN) for porous metal plasticity."
```
I did not check all the equations.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1805,7 +1873,14 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
         },
-        {.description = "logarithmic neo-Hooke material acc. to Bonet and Wood"});
+        {.description = "Logarithmic neo-Hooke material acc. to Bonet and Wood. The strain energy "
+                        "is computed as\n\n"
+                        "$$\n"
+                        "\\Psi = \\frac{\\mu}{2}(I_1-3) - \\mu \\ln(J) + "
+                        "\\frac{\\lambda}{2}(\\ln J)^2\n"
+                        "$$\n\n"
+                        "with $I_1$ the first invariant of the right Cauchy-Green deformation "
+                        "tensor and $J$ the determinant of the deformation gradient."});
```
rjoussen, line 1883: $I_1$ could just be spelled out for clarity. The desciption should also provide the conversion from $E,\nu$ to the lame parameters.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
rjoussen: this should also include the conversion from E and nu to the lame parameters, otherwise the description does not connect with the input.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1873,7 +1978,13 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
             parameter<double>("C2", {.description = "material constant", .default_value = 0.0}),
             parameter<double>("C3", {.description = "material constant", .default_value = 0.0}),
```
rjoussen, line 1979:
```suggestion
            parameter<double>("C2", {.description = "material constant $c_2$", .default_value = 0.0}),
            parameter<double>("C3", {.description = "material constant $c_3$", .default_value = 0.0}),
```
this way it connects better to the equation

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1885,7 +1996,15 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
             parameter<double>("NUE", {.description = "Poisson's ratio"}),
```
rjoussen, line 1996:
```suggestion
            parameter<double>("NUE", {.description = "Poisson's ratio $\nu$"}),
```
same above

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -1895,7 +2014,13 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
         {
             input_field<double>("MUE", {.description = "Shear modulus"}),
```
rjoussen, line 2015:
```suggestion
            input_field<double>("MUE", {.description = "Shear modulus $\mu$"}),
```

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2140,7 +2358,13 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
             parameter<int>(
                 "STR_TENS_ID", {.description = "MAT ID for definition of Structural Tensor"}),
             parameter<int>("INIT",
-                {.description = "initialization modus for fiber alignment", .default_value = 1}),
+                {.description =
+                        "Initialization mode for fiber alignment: "
+                        "0 - Fibers defined by material parameters on element basis;"
+                        "1 - Fibers defined in input file on element basis;"
+                        "4 - Fibers defined in material on gauss point basis, i.e., by nodes;"
+                        "3 - Fibers defined in input file on gauss point basis, i.e., by nodes",
```
rjoussen, line 2366: I would rather recfactor this to an enum directly instead of documenting this.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2167,14 +2399,29 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
+                    .default_value = 1}),
```
rjoussen, line 2408: same here and below.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2379,7 +2715,16 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
             parameter<double>("C4",
                 {.description = "constant 4 for distribution function", .default_value = 1e16}),
         },
-        {.description = "Structural tensor strategy in anisotropic materials"});
+        {.description =
+                "Structural tensor strategy in anisotropic materials. No potential is "
+                "evaluated; it supplies the structural tensor $\\mathbf{A}$, for example\n\n"
```
rjoussen, line 2720:
```suggestion
                "evaluated; it supplies the second-order tensor $\\mathbf{A}$, for example\n\n"
```

doc/documentation/src/analysis_guide_templates/material_directions.rst.j2
```diff
@@ -0,0 +1,59 @@
...
+As an alternative, solid elements accept cylindrical coordinate-system vectors ``RAD``, ``AXI``,
+and ``CIR``. All three must be given together.
+Some specialized materials only support one of these input methods; the direction requirements
+and restrictions for each anisotropic material or summand are documented alongside its
+parameters in the Input Parameter Reference.
+
+``MAT_ElastHyper`` and ``MAT_ViscoElastHyper`` do not impose a direction count themselves; their
+requirements are determined by the selected summands. A ``STR_TENS_ID`` controls the
+structural-tensor strategy for several anisotropic summands but does not replace the underlying
+mean fiber direction.
```
rjoussen, line 59:
```suggestion
```
thats specialized material knowledge that does not belong here, in my opinion.

rjoussen (changes requested):
I left some comments, mainly on the material descriptions. I like the extensions a lot, and my comments are mainly about making the connection between input parameter names and their descriptions/equations even clearer. E.g., even if a parameter is named `C1`, it's nice if its description says "parameter $c_1$" if that symbol is used in an equation in the description. The added equations look reasonable, but I didn't check them all with the code, tbh.

I like the addition of the explanation of the material directions, which is vital information for the documentation and does not duplicate knowledge. 

Regarding the other rst files, I still have concerns about duplication and future syncing issues. I still believe lists like this would need to be retrieved automatically from typed material description members directly in the material code. These could also use tags (=typed enums) for allowed physics, role (parent, child_only, law, helper,...), behaviour (plastic, elastic, damage, viscous,...), kinematics,... If that works well, that could even be used directly in the code for checks (e.g., is material a fluid material?). Additionally, the input specs could move into the material files as well to have everything in one place. This is probably more important, or at least independent, than putting materials in sub-folders, however a bigger refactor. Hence, I would be in favour of first merging the new description texts in validmaterials independently.

PR #2147, opened 2026-07-30
[bugfix] Fix two flaws regarding strain energy output in the hyperelastic framework

ischeider:
## Description and Context

When creating the documentation of the hyperelastic framework, see PR #2146 , the LLM noted two flaws in the strain energy output of hyperelastic parts, namely in

```
ELAST_CoupBlatzKo
ELAST_VolPow
```

This PR is to fix those two.

PR: #2148

### DIFF
--- a/unittests/mat/elast/4C_elast_strain_energy_test.cpp
+++ b/unittests/mat/elast/4C_elast_strain_energy_test.cpp
@@ -1,0 +1,116 @@
+// This file is part of 4C multiphysics licensed under the
+// GNU Lesser General Public License v3.0 or later.
+//
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#include <gtest/gtest.h>
+
+#include "4C_mat_elast_coupblatzko.hpp"
+#include "4C_mat_elast_volpow.hpp"
+#include "4C_material_parameter_base.hpp"
+
+#include <cmath>
+
+namespace
+{
+  using namespace FourC;
+
+  double evaluate_blatz_ko_energy(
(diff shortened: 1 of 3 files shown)


unittests/mat/elast/4C_elast_strain_energy_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+    Core::LinAlg::Matrix<3, 1> dpsi(Core::LinAlg::Initialization::zero);
+    Core::LinAlg::Matrix<6, 1> ddpsi(Core::LinAlg::Initialization::zero);
+    summand.add_derivatives_principal(dpsi, ddpsi, prinv, 0, 0);
+
+    constexpr double step = 1.0e-6;
+    auto prinv_minus = prinv;
+    auto prinv_plus = prinv;
+    prinv_minus(2) -= step;
+    prinv_plus(2) += step;
+    const double numerical_derivative = (evaluate_blatz_ko_energy(summand, prinv_plus) -
```
c-p-schmidt, line 54: Thank you for the fix! But would it make sense to use FAD instead of an FD approximation?

c-p-schmidt (approved):

unittests/mat/elast/4C_elast_strain_energy_test.cpp
ischeider: We had to transform the variables first, thus the finite difference is much easier to implement, and 1e-9 is reasonably low, isn't it?

georghammerl (approved):

unittests/mat/elast/4C_elast_strain_energy_test.cpp
c-p-schmidt: My experience with FD checks is that it is hard to tell how well the approximation is. But that was only a suggestion; I'm also fine with the FD check.

PR #2157, opened 2026-08-05
Rebuild docker image to update the dependency hash for PR #1795

vryy:
## Description and Context

This PR is to update the dependency hash for #1795 following the instructions from https://github.com/4C-multiphysics/4C/tree/main/docker#how-to-update-the-docker-image-for-the-github-actions

#1795


presets/docker/CMakePresets.json
```diff
@@ -11,6 +11,7 @@
         "CMAKE_C_COMPILER": "/usr/bin/mpicc",
         "CMAKE_INSTALL_PREFIX": "/home/user/4C",
         "FOUR_C_ENABLE_WARNINGS_AS_ERRORS": "ON",
+        "FOUR_C_ENABLE_FE_TRAPPING": "OFF",
```
rjoussen: Why is this needed now?

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -533,6 +561,21 @@ function(__four_c_test_restart)
...
+    set(skip_var "ON")
+  endif()
+
+  # check if the platform is matched
+  # By default we enable the test, but if EXCLUDE_PLATFORM is defined, e.g., Linux|Darwin|Windows,
+  # then the test is ignored if the platform is matched.
+  if(CMAKE_SYSTEM_NAME IN_LIST _parsed_EXCLUDE_PLATFORM)
+    set(skip_var "ON")
+  endif()
+
```
rjoussen, line 578: I this needed? Dependent tests will not run anyway if the main test did not run

presets/docker/CMakePresets.json
vryy: the mechanism to trap the overflow error is different on macos, it's needed to be done separately

cmake/functions/four_c_testing_functions.cmake
vryy: it happens that the main test passed, but the following ones not

ppraegla (changes requested):
It should not be necessary to update the docker image when only adding the macOS workflow. See my review of #1795

vryy:
This PR is no longer needed because the docker dependencies are untouched

PR #2161, opened 2026-08-05
Fix cut kernel comparison

vryy:
## Description and Context

This PR is to check which tests failed after the comparison in cut kernel is fixed. Eventually, the check does no harm to the tests.

#2121

### DIFF
--- a/src/cut/4C_cut_kernel.hpp
+++ b/src/cut/4C_cut_kernel.hpp
@@ -2200,19 +2200,21 @@ namespace Cut::Kernel
         if (det < 1.0e-16) std::cout << "!!! determinant of jacobian is smaller than 1.0e-16 !!!\n";
       }
 
-      if (det < 1.0e-16 && det > 1.0e-16)
+      if (det < 1.0e-16)
+      {
 #if EXTENDED_CUT_DEBUG_OUTPUT
         std::cout << "Determinant in  compute position is very close to zero" << std::endl;
 #endif
-      if (Core::MathOperations<FloatType>::abs(det) == 0.0)  // here might lie problem for the cln
-      {
-        /* then calculation of a normal to a line like this makes no sense at
-         * all! */
+        if (Core::MathOperations<FloatType>::abs(det) == 0.0)  // here might lie problem for the cln
+        {
+          /* then calculation of a normal to a line like this makes no sense at
+           * all! */
 #if EXTENDED_CUT_DEBUG_OUTPUT
-        std::cout << "Absolute value of the determinant is zero " << std::endl;
+          std::cout << "Absolute value of the determinant is zero " << std::endl;
 #endif
-        zeroarea_ = true;
-        return false;
+          zeroarea_ = true;
+          return false;
+        }
       }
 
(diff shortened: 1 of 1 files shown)


src/cut/4C_cut_kernel.hpp
```diff
@@ -2200,7 +2200,7 @@ namespace Cut::Kernel
         if (det < 1.0e-16) std::cout << "!!! determinant of jacobian is smaller than 1.0e-16 !!!\n";
       }
 
-      if (det < 1.0e-16 && det > 1.0e-16)
+      if (det < 1.0e-16)
```
maxfirmbach, line 2203: I think we should use `std::numeric_limits` here.

maxfirmbach (commented):
@vryy I just have a minor remark.

src/cut/4C_cut_kernel.hpp
ppraegla:
```suggestion
if (Core::MathOperations<FloatType>::abs(det) < 1.0e-16)  // here might lie problem for the cln
{
   #if EXTENDED_CUT_DEBUG_OUTPUT
     std::cout << "Determinant in  compute position is very close to zero" << std::endl;
   #endif
```

Wouldn't this be the actual fix, if we assume from the cout that `if (det < 1.0e-16 && det > 1.0e-16)` should have checked if the determinant is close to zero?

However, this changes the code path. The `if (Core::MathOperations...` was never reached before since it was inside the if above that was always false. Depending on whether EXTENDED_CUT_DEBUG_OUTPUT was set, the second if was either inside or after the first if.

vryy: From the physical standpoint, it is obvious that `if (det < 1.0e-16)` is needed because zero determinant can happen with small cut case (sliver element). The comment `Determinant in  compute position is very close to zero` also clearly stated the intention. I don't know why it is conflicted with the check. In fact, this code is obscure and confusing, in the sense that the block after the `if` must be wrapped in `{}` except it is a single-line command.

src/cut/4C_cut_kernel.hpp
vryy: Hi @maxfirmbach, I agree. But this PR is meant to fix the check only. For changing the check value, I'll leave it for #2121.

maxfirmbach (approved):

src/cut/4C_cut_kernel.hpp
```diff
@@ -2200,19 +2200,21 @@ namespace Cut::Kernel
...
```
jeremylt, line 2203: Q for my understanding - would any of the tolerances in `4C_cut_tolerance.hpp` be appropriate here?

vryy: @jeremylt the tolerance value should be adapted, but I leave it for #2121

jeremylt: Makes sense! I'm eager to see this go in, as its a patch I've had to apply manually.

jeremylt (approved):

PR #2012, opened 2026-05-04
Enable using Kokkos with CUDA backend for MIRCO and in general

PhilipOesterlePekrun:
@mayrmt 

## Description and Context

This PR adds build-system support for compiling 4C with CUDA-enabled Kokkos. For this purpose, this PR introduces a compiler wrapper, `clangcuda++`, which should be used as the `CMAKE_CXX_COMPILER` (or, in the case of MPI, the `OMPI_CXX` backend), as well as some CMake additions which are optionally enabled.

This change is based on my attempts to get 4C to compile while using MIRCO (library) with Kokkos' CUDA backend for GPU offloading. Several 4C translation units fail to compile with the `kokkos_launch_compiler` or `nvcc_wrapper` provided by Kokkos because NVCC does not seem to be able to compile some more "advanced" C++ code, which we have a fair bit of in 4C since it is a large project. Clang happens to compile CUDA or CUDA-related code much better than Nvidia's own compiler (don't ask me why), but still has some issues by itself. Using the `clangcuda++` wrapper as the `CMAKE_CXX_COMPILER` (or, in the case of MPI, the `OMPI_CXX` backend), along with the corresponding target properties and compile definitions by setting `FOUR_C_CLANGCUDA=ON`, allows compiling all of 4C with CUDA-enabled Kokkos.

The current implementation distinguishes between CUDA host-side compilation and CUDA device compilation. At the moment, 4C itself only requires CUDA host-side compilation, since it does not yet contain raw Kokkos device kernels such as `Kokkos::parallel_for`, `KOKKOS_LAMBDA`, etc. in its own sources. But, this is now possible by simply marking a target or source with `FOUR_C_CLANGCUDA_DEVICE_COMPILE`, allowing GPU offloading anywhere in 4C. As long as Trilinos is built with `TPETRA_INST_CUDA=OFF`, this also does not conflict with any existing MPI parallelism (`NO` will be KokkosSerial).

The changes were verified by compiling 4C successfully with the relevant Kokkos CUDA-enabled setup and I also tested small Kokkos kernel examples to confirm that host-side and device CUDA compilation are distinguished correctly. I have documented how Trilinos, MIRCO, and 4C should be compiled for this combination to work. I'll just attach that to this PR for now: [DocumentKokkosCuda_4C_Trilinos_MIRCO.zip](https://github.com/user-attachments/files/27351494/DocumentKokkosCuda_4C_Trilinos_MIRCO.zip)

[...]

PR #2162, opened 2026-08-06
Add kokkosparallel docker and workflow

PhilipOesterlePekrun:
This PR is related to and should precede #2012. @ppraegla suggested that the docker and related workflow are pushed in a separate PR first (see https://github.com/4C-multiphysics/4C/pull/2012#discussion_r3719322270). Therefore, this PR just adds the new Trilinos installation scripts, dependencies hash computation, dockerfile, and docker_kokkosparallel.yml workflow file.

### DIFF
--- a/.github/workflows/docker_kokkosparallel.yml
+++ b/.github/workflows/docker_kokkosparallel.yml
@@ -1,0 +1,96 @@
+name: Docker Kokkos parallel
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - '.github/workflows/docker_kokkosparallel.yml'
+      - 'dependencies/current/backtrace'
+      - 'dependencies/current/cmake'
+      - 'dependencies/current/dealii'
+      - 'dependencies/current/gmsh'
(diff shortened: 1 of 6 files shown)


mayrmt (approved):
I can't speak to the GitHub workflows, but installation scripts look fine.

PhilipOesterlePekrun:
@mayrmt The workflow is basically a copy of the standard docker workflow with some small changes. So, it will allow us to use the workflow/action to build and push the docker, as Patrick suggested.

ppraegla (approved):
Thanks for splitting the PR.

ppraegla:
As a note: Once this PR is merged, the `tag-images-as-main` job in the workflow will fail on main because the image does not yet exist. And you can only build the image once the workflow is on main. This means there will be no `main` tag until the docker image is update once more. But for me, it's okay not to have the `main` tag until the next real update of the docker image.

ppraegla:
After checking the workflow file again, you can just trigger the manual build again once the docker build is finished. The workflow checks if the image already exists, which is it does. So, the build is not run again. But the docker image is tagged as main.

PR #2168, opened 2026-08-07
Detect inconsistent fiber entry and throw an error

ischeider:
## Description and Context
When I asked the LLM to help me with the material documentation (PR #2146 ), it wrote "For materials using the default anisotropy framework, this coordinate system [RAD|AXI|CIR] takes precedence over explicit ``FIBER*`` vectors. @lauraengelhardt insisted that this is rather a bug than a feature. So I added a check for giving both `FIBER` and `RAD|AXI|CIR` entries. Since a check for adding `FIBER2` without `FIBER1` (or `FIBER3` without any of them) did not exist as well, I added this, too.

I found that this behavior is similar for element based and node based directions. So I implemented the check for both elements and nodes. For node based fiber directions, they may be defined explicitly or by CIR|TAN|HELIX|ANGLE for cardiac monodomains. An explanation for those parameters would definitely be useful in the documentation ;-)
The parameter RAD is never read, but always calculated from CIR x TAN. Due to the comments I removed this parameter completely.

PR #2146

AI was used for code generation (including the unittests)
I understand what *chatGPT 5.6 sol* did, that is, what the check does and how, but I could barely code it in the same smart way. Particularly, since the legacy dat format style of the element and node definitions include also nonexistent optional parameters as entered but empty (I didn't expect that), which made the if clauses  a bit more complex.

### DIFF
--- a/doc/documentation/src/analysis_guide_templates/geometrydefinition.rst.j2
+++ b/doc/documentation/src/analysis_guide_templates/geometrydefinition.rst.j2
@@ -53,9 +53,9 @@ The nodes are read in by the section ``NODE COORDS``.
 
 Within this section, the nodes are read one per line. There are different types of nodes:
 
-- General nodes (NODE)
-- Control points for nurbs geometry (CP)
-- Nodes with addition fiber information (FNODE)
+- General nodes (`NODE`)
+- Control points for nurbs geometry (`CP`)
+- Nodes with addition fiber information  (`FNODE`)
 
 All node coordinates must be given with three coordinates, even if the problem is 1D or 2D:
 
@@ -64,13 +64,14 @@ All node coordinates must be given with three coordinates, even if the problem i
    NODE COORDS:
    - "NODE <ID> COORD <coord-x> <coord-y> <coord-z> [ROTANGLE <coord-yz> <coord-xz> <coord-xy>]"
    - "CP <ID> COORD <coord-x> <coord-y> <coord-z> <weight>"
-   - "FNODE <ID> COORD <coord-x> <coord-y> <coord-z> [FIBER1|FIBER2|FIBER3|CIR|TAN|RAD|HELIX|TRANS] <further parameters>"
+   - "FNODE <ID> COORD <coord-x> <coord-y> <coord-z> FIBER1 <x y z> [FIBER2 <x y z> [FIBER3 <x y z>]]"
+   - "FNODE <ID> "COORD <coord-x> <coord-y> <coord-z> CIR <x y z> TAN <x y z> HELIX <angle> TRANS <angle>"
 
-For ``FNODE``, explicit fiber directions must be numbered consecutively starting with ``FIBER1``.
-They cannot be combined with cardiac fiber directions. Cardiac fiber directions require ``CIR``,
-``TAN``, ``HELIX``, and ``TRANS``; ``RAD`` is optional. All nodes of an element must provide the
-same set of parameters, either explicit or cardiac fiber directions. Their direction
-vectors and angle values may vary between nodes.
+For ``FNODE`` in combination with fiber directions, the directions must be numbered consecutively starting with ``FIBER1``.
+For the problemtype `Cardiac_Monodomain`, the fiber directions can be entered alternatively
+using ``CIR``, ``TAN``, ``HELIX``, and ``TRANS``.
+All nodes of an element must provide the same set of parameters, either explicit or cardiac fiber directions.
+Their direction vectors and angle values may vary between nodes.
 
 .. _geometrysets:
 
--- a/doc/documentation/src/analysis_guide_templates/geometrydefinition.rst.j2
+++ b/doc/documentation/src/analysis_guide_templates/geometrydefinition.rst.j2
@@ -66,6 +66,11 @@ All node coordinates must be given with three coordinates, even if the problem i
    - "CP <ID> COORD <coord-x> <coord-y> <coord-z> <weight>"
    - "FNODE <ID> COORD <coord-x> <coord-y> <coord-z> [FIBER1|FIBER2|FIBER3|CIR|TAN|RAD|HELIX|TRANS] <further parameters>"
 
+For ``FNODE``, explicit fiber directions must be numbered consecutively starting with ``FIBER1``.
+They cannot be combined with cardiac fiber directions. Cardiac fiber directions require ``CIR``,
+``TAN``, ``HELIX``, and ``TRANS``; ``RAD`` is optional. All nodes of an element must provide the
+same set of parameters, either explicit or cardiac fiber directions. Their direction
+vectors and angle values may vary between nodes.
 
 .. _geometrysets:
 
@@ -177,4 +182,3 @@ while the parameter :math:`\sigma` is defined by the parameter ``RADIUSDISTRIBUT
     PARTICLE DYNAMIC/DEM:
       INITIAL_RADIUS: RadiusFromParticleInput|NormalRadiusDistribution|LogNormalRadiusDistribution
       RADIUSDISTRIBUTION_SIGMA: <variation>   # variation sigma for a radius distribution
-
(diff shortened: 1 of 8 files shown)


src/core/fem/src/general/node/4C_fem_general_fiber_node_holder.cpp
```diff
@@ -60,7 +61,13 @@ void Core::Nodes::NodalFiberHolder::set_angle(AngleType type, const std::vector<
 const std::vector<double>& Core::Nodes::NodalFiberHolder::get_angle(
     Core::Nodes::AngleType type) const
 {
-  return angles_.at(type);
+  const auto angle = angles_.find(type);
+  if (angle == angles_.end())
+  {
+    FOUR_C_THROW("Nodal fiber data does not contain the requested {} angle.",
+        type == AngleType::Helix ? "HELIX" : "TRANS");
```
jeremylt, line 68: Question: Is there an enum to string helper for the `AngleType` enum? This sort of helpful printing statement ternary is the sort of thing that ends up getting repeated a lot in debugging and error output. I noticed that the enums in the particle code have such helpers for exactly that reason.

jeremylt:
> The parameter RAD is never read, but always calculated from CIR x TAN

For what its worth, I think that sort of thing can be a trap for users because it creates as situation where they believe they specified something when they in fact did not. Speaking from past experience with users rather than with any expertise in the codebase

src/core/fem/src/general/node/4C_fem_general_fiber_node_holder.cpp
ischeider: For now, it's the one conversion for AngleType, but I see the point. However, here we need to print the parameter entry which is in capital letters, so the created string "Helix" wouldn't be correct.

jeremylt: ahh, good point about wanting the parameter entry over the usual human readable name

src/solid_ele/4C_solid_ele_fiber.cpp
```diff
@@ -38,26 +73,28 @@ Discret::Elements::Fibers Discret::Elements::read_fibers(
 
     fibers.element_fibers.emplace_back(tensor);
   }
+  std::cout << "Number of fibers: " << fibers.element_fibers.size() << std::endl;
```
tuchpaul:
```suggestion
```
This needs to be removed.

doc/documentation/src/analysis_guide_templates/geometrydefinition.rst.j2
```diff
@@ -66,6 +66,11 @@ All node coordinates must be given with three coordinates, even if the problem i
    - "CP <ID> COORD <coord-x> <coord-y> <coord-z> <weight>"
    - "FNODE <ID> COORD <coord-x> <coord-y> <coord-z> [FIBER1|FIBER2|FIBER3|CIR|TAN|RAD|HELIX|TRANS] <further parameters>"
 
+For ``FNODE``, explicit fiber directions must be numbered consecutively starting with ``FIBER1``.
+They cannot be combined with cardiac fiber directions. Cardiac fiber directions require ``CIR``,
+``TAN``, ``HELIX``, and ``TRANS``; ``RAD`` is optional. All nodes of an element must provide the
+same set of parameters, either explicit or cardiac fiber directions. Their direction
+vectors and angle values may vary between nodes.
```
tuchpaul: Here one could explicitly state that one refers to the fiber definition of the cardiac monodomain. 
Since, you also have cardiac fibers when you simulate cardiac growth and remodeling using the mixture framework.

tuchpaul (commented):
Thanks for taking care! I left some additional comments.

tuchpaul:
> > The parameter RAD is never read, but always calculated from CIR x TAN
> 
> For what its worth, I think that sort of thing can be a trap for users because it creates as situation where they believe they specified something when they in fact did not. Speaking from past experience with users rather than with any expertise in the codebase

I am also in favor of removing `RAD` from the `FNODE`-input.

doc/documentation/src/analysis_guide_templates/geometrydefinition.rst.j2
ischeider: The `FIBER` definition works for general anisotropic materials, only the CIR|TAN|HELIX|TRANS works for cardiac monodomains. I rephrased the sentence.

tuchpaul (approved):

src/core/fem/src/general/node/4C_fem_general_fiber_node_utils.cpp
```diff
@@ -36,6 +36,43 @@ void Core::Nodes::project_fibers_to_gauss_points(const Core::Nodes::Node* const*
...
+      const auto& reference_angles = fiberNodes[0]->angles();
+      const auto& node_angles = fiberNodes[inode]->angles();
+      const bool same_angles =
+          reference_angles.size() == node_angles.size() &&
+          std::all_of(reference_angles.begin(), reference_angles.end(),
+              [&node_angles](const auto& angle) { return node_angles.contains(angle.first); });
+      if (!same_angles)
+      {
+        FOUR_C_THROW("All nodes of an element must define the same angle types.");
+      }
```
jeremylt, line 73:
```suggestion
      FOUR_C_ASSERT_ALWAYS(same_angles, "All nodes of an element must define the same angle types.");
```
Question - is `FOUR_C_ASSERT_ALWAYS` or `FOUR_C_THROW` preferred in this case?

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -66,6 +66,74 @@ namespace Core::IO::Internal
...
+    {
+      FOUR_C_THROW(
+          "Fiber directions in FNODE must be defined either by CIR/TAN/HELIX/TRANS or by FIBER1, "
+          "FIBER2, etc., but not by both.");
+    }
+
+    if (has_cardiac_directions && !(has_circular && has_tangential && has_helix && has_transverse))
+    {
+      FOUR_C_THROW("Cardiac fiber directions in FNODE require all of CIR, TAN, HELIX, and TRANS.");
+    }
```
jeremylt, line 133:
```suggestion
    FOUR_C_ASSERT_ALWAYS(!has_cardiac_directions || (has_circular && has_tangential && has_helix && has_transverse),
      "Cardiac fiber directions in FNODE require all of CIR, TAN, HELIX, and TRANS.");
```
Same question here, I don't know the norms

src/core/io/src/4C_io_meshreader.cpp
```diff
@@ -66,6 +66,74 @@ namespace Core::IO::Internal
...
+    const bool has_transverse = data.angles.contains(Core::Nodes::AngleType::Transverse);
+    const bool has_cardiac_directions =
+        has_circular || has_tangential || has_helix || has_transverse;
+
+    if (!data.fibers.empty() && has_cardiac_directions)
+    {
+      FOUR_C_THROW(
+          "Fiber directions in FNODE must be defined either by CIR/TAN/HELIX/TRANS or by FIBER1, "
+          "FIBER2, etc., but not by both.");
+    }
```
jeremylt, line 128:
```suggestion
    FOUR_C_ASSERT_ALWAYS(data.fibers.empty() || !has_cardiac_directions,
        "Fiber directions in FNODE must be defined either by CIR/TAN/HELIX/TRANS or by FIBER1, "
        "FIBER2, etc., but not by both.");
```
Same question here, I don't know the norms

jeremylt (approved):
I just have a question about norms around `FOUR_C_ASSERT_ALWAYS` vs `FOUR_C_THROW`

src/core/fem/src/general/node/4C_fem_general_fiber_node_utils.cpp
rjoussen: As far as I know `FOUR_C_ASSERT_ALWAYS` is preferred since it usually leads to a slightly more expressive error message (and looks cleaner in the code).

PR #2163, opened 2026-08-06
Preparations for fluid output to vtk based

c-p-schmidt:
## Description and Context
I recently worked on moving the fluid output to a vtk-based output. To do so, I have two preparatory commits to start with.
The first commit is just applying some minor improvements on several fluid files, like:
    - remove unnecessary comments from cpp functions
    - move docu from cpp to hpp where appropriate
    - remove unnecessary header inclusions
    - remove unnecessary returns
    - add nodiscard where reasonable

The second commit then mainly structures the current output into output meant for visualization (which is replaced in the upcoming PRs) and into restart output, which is kept.

If you are interested in the overall work (which I intentionally split to enable a reasonable review), you can have a look at the changes in a PR on my fork https://github.com/c-p-schmidt/4C/pull/17

## Related Issues and Pull Requests
part of #211

## Disclosure of AI assistance
In this upcoming series of PRs, I mainly used AI to figure out why some quantities cannot be written in the new format straightforwardly due to map mismatches.

PR #2172, opened 2026-08-10
Migrate output of FluidImplicitTimeInt to vtk-based output

c-p-schmidt:
## Description and Context
As a final step in migrating the fluid output to a vtk-based format, the output of FluidImplicitTimeInt is addressed in this PR. 
**Note:** 
- There is still conventional fluid-related output originating, e.g., from FSI, or called from adapters, but the folder `fluid` is clean now.
- I decided to remove the output of **fld_growth_displ** as it also did not work before. The problem is that the output is written only for parts of the discretization, which still enabled output using **write_vector**, and post-processing also seemed to work (it at least did not throw an error); however, it could not be displayed, e.g., in Paraview, due to the mismatch in discretization. Some more info you can read [here](https://github.com/c-p-schmidt/4C/tasks/44b2329d-ded9-4a64-a02f-bf6b3d76ddf2). So basically, there is no change in functionality.

part of #211
follows #2163 #2164 #2165 #2167 #2170

In this series of PRs, I mainly used AI to figure out why some quantities cannot be written in the new format straightforwardly due to map mismatches.


lkoeglmeier (approved):
Nice - thanks a lot for your efforts!

engstlerL (approved):

PR #2175, opened 2026-08-10
Docker update more robust dependencies hash

ppraegla:
The dependencies hash is computed from the SHA1 of every file in the specified folders. Hidden files (like the macOS .DS_Store) can unintentionally change the computed hash and will lead to a failing pipeline. Hidden files should not be used for the dependencies build. If at some point hidden files are needed for the build, e.g., a .dockerignore, the check needs to be adapted.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,7 +1,12 @@
 # Docker
 
 This folder contains Dockerfiles for different configurations. The `Dockerfile` in the folder `dependencies` is the Dockerfile to
-build the project with the current dependencies. Other configurations reside in different subfolders.
+build the project with the current dependencies. Other configurations reside in different subfolders. The other configurations are:
+- oldest supported: dependencies image with the oldest supported dependencies (currently only Trilinos).
+- prebuilt 4C: docker image with ready-to-use 4C executable
+- minimal prebuilt 4C: prebuilt 4C docker image with everything removed not need to execute 4C
+- trilinos develop: dependencies image with Trilinos develop installation. Currently not working due to Epetra removal.
+- kokkos parallel: dependencies image with Trilinos installation with Kokkos OpenMP and Cuda.
 
 ## Build docker images locally
 
@@ -16,7 +21,7 @@ docker build --tag 4c-dependencies --file docker/dependencies/Dockerfile .
 ## How to update the docker images for the Github Actions
 
 1. Make your changes to the Dockerfile or dependencies
-2. Compute the sha1 of the docker and dependencies folder.
+2. Compute the sha1 of the docker and dependencies folder. Make sure that no unwanted files are in these folders.
 
 ```bash
 cd <project_root>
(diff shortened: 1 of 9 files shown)


docker/README.md
```diff
@@ -1,7 +1,12 @@
 # Docker
 
 This folder contains Dockerfiles for different configurations. The `Dockerfile` in the folder `dependencies` is the Dockerfile to
-build the project with the current dependencies. Other configurations reside in different subfolders.
+build the project with the current dependencies. Other configurations reside in different subfolders. The other configurations are:
+- oldest supported: dependencies image with the oldest supported dependencies (currently only Trilinos).
+- prebuilt 4C: docker image with ready-to-use 4C executable
+- minimal prebuilt 4C: prebuilt 4C docker image with everything removed not need to execute 4C
+- trilinos develop: dependencies image with Trilinos develop installation. Currently not working due to Epetra removal.
+- kokkos parallel: dependencies image with Trilinos installation with Kokkos OpenMP and Cuda.
```
jeremylt, line 9:
```suggestion
- `oldest_supported`: dependencies image with the oldest supported dependencies (currently only Trilinos).
- `prebuilt_4C`: docker image with ready-to-use 4C executable
- `prebuilt_4C_minimal`: prebuilt 4C docker image with everything removed not need to execute 4C
- `trilinos_develop`: dependencies image with Trilinos develop installation. Currently not working due to Epetra removal.
- `kokkosparallel`: dependencies image with Trilinos installation with Kokkos OpenMP and Cuda.
```
Minor - should we use folder names exactly since that's what's given above? Also, should `kokkosparallel` folder be renamed `kokkos_parallel` to match the pattern of the rest?

davidrudlstorfer (approved):

docker/README.md
ppraegla: My idea was to use names as I would speak about the configurations, e.g., 'the oldest support configuration'. But I can also use the folder names, if that makes more sense to you.

You are right, I will rename the folder to `kokkos_parallel` to be consistent with the other folders.

jeremylt: I could go either way with the names in the list; it just stood out to me that above we used the folder name

rjoussen (approved):

docker/README.md
PhilipOesterlePekrun: We can rename kokkosparallel to kokkos_parallel. My initial thought was just to group them like "kokkos with parallel backend" rather than "kokkos and parallel" or something. But we might want to change other occurrences then to make it clear.

PhilipOesterlePekrun: Also, I think the CUDA build looks to be taking a lot longer because the ccache is somehow not working for it. Quit an inconvenience of course, if it has to compile fully every time. I am looking into it.

ppraegla: I realized that consistently renaming the `kokkosparallel` folder introduces too many changes for the scope of this PR. I would leave the renaming to @PhilipOesterlePekrun . He is already looking into the Cuda build.

PR #2185, opened 2026-08-14
Remove redundant definition from input file

c-p-schmidt:
## Description and Context
While working on coating simulations, I realized that there is redundant information in one of our input files. As the input file states, the initial field is defined by condition and not by function. Nevertheless, a function to define this initial field was also defined, which is now removed.

none

none


tuchpaul (approved):

ischeider (approved):
My personal view is rather to make a comment behind INTIFUNCNO that it is not used here, so that one can switch easily to another simulation feature. But I know this is not the 4C policy ;-)

PR #2190, opened 2026-08-17
Add porosity output for StructPoro* materials

lkoeglmeier:
The first commit adds the option to output the `porosity` for StructPoro* materials. 
The second commit renames the quantity `porosity `to `volfrac_multiphase` in the porofluid_pressure_based* framework, as this quantity only describes the volume fraction of the multiphase porespace - the porosity is the sum of the volume fraction in the multiphase porespace and possible additional porous networks. Due to the renaming, the order of the output quantities in the VTK output has changed.

AI was used for a first implementation to output the porosity field (first commit), for test migration after renaming (second commit) and review. All AI-assisted code has been reviewed and validated by the author, who takes responsibility for the contribution.


src/inpar/4C_inpar_IO_runtime_vtk_output_structure.cpp
```diff
@@ -82,6 +82,7 @@ namespace Inpar
                 deprecated_selection<Solid::OptQuantityType>("OPTIONAL_QUANTITY",
                     {
                         {"No", Solid::optquantity_none},
+                        {"porosity", Solid::optquantity_porosity},
```
davidrudlstorfer: This is an extension to an already deprecated selection. Isn't there some poro IO section where this could be added? In #59 it is stated that this file needs to be reworked anyways. So it would be nice to not add more deprecated code

lkoeglmeier:
As discussed with @c-p-schmidt (team-output), the quantity `porosity` was output in the old output format via the `vis_names`/`vis_data` functions in the material. It is currently not possible to output these quantities using the new vtk-based output framework. Since other materials also store material-specific output quantities in these functions, we agreed to open a separate PR to migrate these material quantities to the new output framework.
This PR will therefore be reduced to the second commit only: renaming the quantity` porosity` to `volfrac_multiphase`.

davidrudlstorfer (approved):
Is there any formatting change in the vtu files or is the git diff broken that much here?

c-p-schmidt:
> Is there any formatting change in the vtu files or is the git diff broken that much here?

Not sure if I understood you correctly, but renaming the quantity causes a reordering within the file.

c-p-schmidt (approved):

davidrudlstorfer:
> Not sure if I understood you correctly, but renaming the quantity causes a reordering within the file.

Aaah ups. Didn't see that

PR #2193, opened 2026-08-17
Add brew versions to the key hash for caching of macOS workflow

vryy:
## Description and Context

Add the version of all libraries installed with brew to the hash code of dependency and build step. This is necessary since any upstream update from brew can invalidate the build cache.

#1795 

The hash script is written based on suggestion of Gemini.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,23 +1,27 @@
 name: macOS built test
-
 on:
   schedule:
     - cron: 3 4 * * 0
       timezone: "Europe/Berlin"
   workflow_dispatch:
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-
 jobs:
   build:
     name: Build 4C on macOS
     runs-on: macos-26
+    outputs:
+      brew-version-hash: ${{ steps.brew-versions.outputs.hash }}
     steps:
       - uses: actions/checkout@v7
       - name: System preparation
         run: ./dependencies/darwin/prepare_system.sh
+      - name: Get Homebrew versions
+        id: brew-versions
+        run: |
+          # Generate a key of all brew libraries
+          echo "hash=$(brew list --versions | sha256sum | awk '{print $1}')" >> $GITHUB_OUTPUT
       - name: Restore dependencies cache
         id: cache-deps
         uses: actions/cache/restore@v6
@@ -37,13 +41,13 @@ jobs:
         with:
           path: ${{ env.HOME }}/opt
           key: ${{ runner.os }}-deps-${{ hashFiles('./dependencies/darwin/compile_dependencies.sh', 'dependencies/darwin/**/install.sh',
-            'dependencies/current/**/install.sh') }}
+            'dependencies/current/**/install.sh') }}-brew-${{ steps.brew-versions.outputs.hash }}
       - name: Restore build cache
         id: cache-build
         uses: actions/cache/restore@v6
         with:
           path: build/
-          key: ${{ runner.os }}-build-${{ github.sha }}
+          key: ${{ runner.os }}-build-${{ github.sha }}-brew-${{ steps.brew-versions.outputs.hash }}
       - name: Build 4C
         if: steps.cache-build.outputs.cache-hit != 'true'
         run: |
@@ -56,8 +60,7 @@ jobs:
         uses: actions/cache/save@v6
         with:
           path: build/
-          key: ${{ runner.os }}-build-${{ github.sha }}
-
+          key: ${{ runner.os }}-build-${{ github.sha }}-brew-${{ steps.brew-versions.outputs.hash }}
   test:
     name: Test 4C on macOS
     needs: build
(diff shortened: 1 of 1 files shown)


ppraegla (commented):
I think the proposed solution does not really solve the problem of the failing test job. If the system libraries have changed between the build and test job, the cache does not exist. So, restore will silently fail (fail-on-cache-miss is false by default), and running the tests will fail because of the missing build and dependencies folders.

I suggest computing the hash from `brew list --version` as you do now, and save it as output. In the test job check if the saved hash from the build job matches the hash in the test job. If not, exit with a clear error message that the system libraries changed. If we see that this happens often, we can look for other solutions (cache the homebrew output as well, or try to pin the versions). What do you think?

c-p-schmidt:
> I think the proposed solution does not really solve the problem of the failing test job. If the system libraries have changed between the build and test job, the cache does not exist. So, restore will silently fail (fail-on-cache-miss is false by default), and running the tests will fail because of the missing build and dependencies folders.
> 
> I suggest computing the hash from `brew list --version` as you do now, and save it as output. In the test job check if the saved hash from the build job matches the hash in the test job. If not, exit with a clear error message that the system libraries changed. If we see that this happens often, we can look for other solutions (cache the homebrew output as well, or try to pin the versions). What do you think?

That sounds like a reasonable suggestion to me. 👍🏼

vryy:
> I think the proposed solution does not really solve the problem of the failing test job. If the system libraries have changed between the build and test job, the cache does not exist. So, restore will silently fail (fail-on-cache-miss is false by default), and running the tests will fail because of the missing build and dependencies folders.
> 
> I suggest computing the hash from `brew list --version` as you do now, and save it as output. In the test job check if the saved hash from the build job matches the hash in the test job. If not, exit with a clear error message that the system libraries changed. If we see that this happens often, we can look for other solutions (cache the homebrew output as well, or try to pin the versions). What do you think?

I guess that was the case in the commit. The test job calculates the hash again, and it must match the hash in the build job to proceed. If it does not match, it still proceeds but likely fails because system libraries change (that was the case of run [#2](https://github.com/4C-multiphysics/4C/actions/runs/31922983047)). Btw, the hash generated from `brew list --version` is a bit fragile because it is just a list of {library-name}-{version-number}, but I think it is reliable enough because it's fast to calculate, rather than calculating the hash from all files in the brew directory. I checked this in my fork and this works:

https://github.com/vryy/4C/actions/runs/32039011290

Pinning brew version could be a long-term solution, we should think about it in case it fails so often. But macos is updated frequently anyway, I assume everyone like to enjoy recent updates, right :)))

.github/workflows/macos.yml
ppraegla, line 13:
```suggestion
    runs-on: macos-26
    outputs:
      brew-version-hash: ${{ steps.brew-versions.outputs.hash }}
```
Save the hash in the build job to be used in the test job.

.github/workflows/macos.yml
```diff
@@ -18,13 +18,18 @@ jobs:
...
+          # Generate a key of all brew libraries
+          echo "hash=$(brew list --versions | sha256sum | awk '{print $1}')" >> $GITHUB_OUTPUT
       - name: Restore dependencies cache
         id: cache-deps
         uses: actions/cache/restore@v4
         with:
           path: ${{ env.HOME }}/opt
           key: ${{ runner.os }}-deps-${{ hashFiles('./dependencies/darwin/compile_dependencies.sh', 'dependencies/darwin/**/install.sh',
-            'dependencies/current/**/install.sh') }}
+            'dependencies/current/**/install.sh') }}-brew-${{ steps.brew-versions.outputs.hash }}
```
ppraegla:
```suggestion
            'dependencies/current/**/install.sh') }}
```

The hash is no longer required in the cache name. Also remove it in all other cache names.

.github/workflows/macos.yml
```diff
@@ -66,19 +71,24 @@ jobs:
       - uses: actions/checkout@v6
       - name: System preparation
         run: ./dependencies/darwin/prepare_system.sh
+      - name: Get Homebrew versions
+        id: brew-versions
+        run: |
+          # Generate a key of all brew libraries
+          echo "hash=$(brew list --versions | sha256sum | awk '{print $1}')" >> $GITHUB_OUTPUT
       - name: Restore dependencies cache
```
ppraegla, line 89:
```suggestion
      - name: Check if the hash differs from the hash in the build job
        if: steps.brew-versions.outputs.hash != needs.build.outputs.brew-version-hash
        run: |
          {
            echo "### Changed system libraries!"
            echo ""
            echo "The test job aborted because the brew system libraries changed after the build job completed."
            echo "The system libraries installed in the test job have a different version than the ones used in the build job to build the dependencies and 4C."
            echo ""
            echo "This should rarely happen. A re-run of the workflow should solve this problem."
          } >> $GITHUB_STEP_SUMMARY
          exit 1
      - name: Restore dependencies cache
```

Before restoring the caches, check if the hash in the test and build job match. If not, use the [GitHub job summary](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-of-adding-a-job-summary) to format a message that will be shown in the workflow summary and abort the test job.

I use [command grouping](https://www.gnu.org/software/bash/manual/html_node/Command-Grouping.html) to avoid `>> $GITHUB_STEP_SUMMARY` after every echo.

ppraegla (commented):
@vryy As discussed, I added the changes from my suggested solution. This way, it should be clearer why the test job failed. We still use the hash to determine whether the system libraries have changed, but we explicitly check that both hashes match. Using the hash in the cache key hides the intention of the hash and you still have to look into the log why the test job failed.

I hope my suggested solution is clearer now.

vryy:
Thanks for the suggestion, the changes seem to be working fine:

https://github.com/vryy/4C/actions/runs/32826550590

I'll squash the commit upon merge.

ppraegla (dismissed):
I would still remove the brew hash from the cache key because it unnecessarily complicates the key. But it is not wrong.

c-p-schmidt (dismissed):

c-p-schmidt (dismissed):

vryy:
> I would still remove the brew hash from the cache key because it unnecessarily complicates the key. But it is not wrong.

I still need the brew hash in cache key to trigger new cache build. The current one in workflow associated with old brew version and does not work. Furthermore, it provides more guard for the caching.

ppraegla (approved):

c-p-schmidt (approved):

c-p-schmidt:
I guess you need to rebase against the current main, since the **ensure_all_tests_pass_kokkos** was introduced in https://github.com/4C-multiphysics/4C/pull/2225 yesterday evening.

PR #2195, opened 2026-08-18
Clean-up 1D pipeflow

engstlerL:
## Description and Context
Some cleanup of the 1D pipeflow implementation, divided into two commits:
- The first one is a pure cleanup: removes unused include, dead code, and modernizes syntax. Also the visualization section is refactored to only run the computation when output is requested.
- The second one changes the input schema: The former used reference_area_A0 is replaced by reference_radius_r0 (as used in the 0D formulation), cardiac period is moved to the structured_tree_model group, and unused input fields are removed. Accordingly, the test files are updated.

AI (opencode) has been used to find unused code and improve syntax.


bgoderbauer (approved):

src/reduced_lung/src/1d_pipe_flow/4C_reduced_lung_1d_pipe_flow_main.cpp
```diff
@@ -1194,6 +1162,8 @@ namespace ReducedLung1dPipeFlow
             pressure_solution, Core::IO::OutputEntity::node, {"p"});
         visualization_writer.append_result_data_vector_with_context(
             flow_solution, Core::IO::OutputEntity::node, {"Q"});
+        visualization_writer.append_result_data_vector_with_context(
```
c-p-schmidt, line 1165: Do you already have a vtk-output test in the repo? Probably not, because otherwise it should fail now with this added quantity.

c-p-schmidt (approved):

PR #2197, opened 2026-08-18
Make NUMMAT param in material parameter input optional

ischeider:
## Description and Context

With the new input format it is not necessary to define the vector length before the values of a vector are given in a parameter field such as `NUMMAT` in a material list. Some materials like `MAT_Solid_Superposition` or `MAT_mixture` omit this length parameter already, but many older materials have it.

I now made all instances of NUMMAT and similar length parameters in the material parameter definitions optional. Note that the parameter has not been removed but is just not read. This is for backward compatibility reason. I know that this will be a point of discussion, but I thought, the optional parameter does not hurt. If desired I can code a warning that this deprecated parameter is contained in the input file.

In this commit I did not change any input file, so they still have the NUMMAT parameter  (and the like). I would like to remove the parameter there as well, but this touches 886 input files (44%):

Parameter | Files | Lines
-- | -- | --
NUMMAT | 795 | 1051
NUMMATEL | 68 | 132
NUMFACINEL | 68 | 132
NUMREAC | 64 | 100
NUMBRANCH | 8 | 8
NUMMATRF | 6 | 6
NUMMATEL2D | 6 | 6
NUMMATEL3D | 4 | 4
NUMELAST | 2 | 6
NUMVISCO | 2 | 6
NUMPAIR | 2 | 2
MEMBRANENUMMAT | 1 | 1
SAMPLENUM | 6 | 6
YOUNGNUM | 47 | 47
total | 886 | 1507

Some more length parameters exist: `NUMSCAL, NUMDOF, NUMVOLFRAC, NUMSLIPSYS, TOTALNUMDOF`, but with most of them I am not sure whether they denote a physically meaningful parameter that is worth to be mentioned.

I made the first change in 4C_global_legacy_module_validmaterials.cpp` for the ElastHyper material (and the corresponding `4C_mat_elasthyper.cpp`), then let Claude Opus 5 do it for the other occurrences. Several times I mentioned distinct material parameter definitions (see the table above), where the LLM should remove the fixed vector size as well. 

After reviewing the code by me, the LLM did the compiling and testing as well. The table above was also generated by LLM.

### DIFF
--- a/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
+++ b/src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
@@ -39,12 +39,6 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
   using namespace Core::IO::InputSpecBuilders;
   std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> known_materials;
 
-  const auto size_from_optional_count = [](const std::string& count_parameter_name)
-  {
-    return [count_parameter_name](const Core::IO::InputParameterContainer& container)
-    { return container.get<std::optional<int>>(count_parameter_name).value_or(0); };
-  };
-
   /*----------------------------------------------------------------------*/
   // Newtonian fluid
   {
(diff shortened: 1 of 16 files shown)


davidrudlstorfer:
Really nice! In my opinion we can directly delete it, we already had a few breaking changes in 4C and this one is rather easy to resolve.

ischeider:
I just noticed that there was a very similar PR, #2080 , recently, where NUMMAT was deleted rigorously for a MIX_... material part. So, it seems deleting the parameter is not such a severe change, but here we have many more affected problems (and people).

rjoussen:
A breaking change is better than being left with a dead input parameter in my opinion, so I am also for removing it completely.

c-p-schmidt:
I'm also fine with this breaking change.

Another option would be to establish the concept of deprecation within 4C, meaning that you would need to check at any location in the code where those parameters are accessed that the default value is obtained, and if not add a screen output that this set parameter is ignored in the code and should be removed and that in an upcoming release an error will be thrown if the parameter is still available.

However, I'm not sure how helpful that really would be.

lauraengelhardt:
Same from my side, I would be fine with just deleting.

ischeider:
Okay, seems that there is a strong opinion on this. Since there are several parameters involved, we need an upgrade mechanism for this. I have to think about it.

lauraengelhardt:
I just want to mention #74 again in this context. We had a meeting about versioning/upgrade mechanisms last October, with the idea of implementing a prototype, but unfortunately not much has happened since then (from my side, due to a lack of time - I’m still very much in favor of this feature). I’d be happy to revisit this :smiley:

PR #2199, opened 2026-08-19
Trilinos 16.2.2

jeremylt:
## Description and Context

Trilinos 16.2.2 released with some fixes for c++-23
See https://github.com/trilinos/Trilinos/pull/15571

None

None


c-p-schmidt (approved):
Since we are using C++20 at the moment, I would consider this optional, but I also don't think it's harmful. So feel free to proceed from my perspective.

ppraegla (approved):
I see the update similarly to @c-p-schmidt. I think we do not need to update Trilinos just because there is a new version available. In the past, we updated Trilinos only when we needed new features for our development, because an update means we need to install the new version on our clusters, which someone has to do. Also, we try to keep the list of supported Trilinos versions small because we only test the newest and the oldest version.

PR #2204, opened 2026-08-20
Minor rework of the ssi framework

c-p-schmidt:
## Description and Context
The cleanup mainly consists of:
- add consts and nodiscard where appropriate
- prefer passing by const reference over using std::shared_ptr
- rework of safety checks

None

Some autocomplete within the IDE for trivial changes.

### DIFF
--- a/src/ssi/4C_ssi_base.cpp
+++ b/src/ssi/4C_ssi_base.cpp
@@ -45,7 +45,7 @@ FOUR_C_NAMESPACE_OPEN
 
 /*----------------------------------------------------------------------*/
 /*----------------------------------------------------------------------*/
-SSI::SSIBase::SSIBase(MPI_Comm comm, const Teuchos::ParameterList& globaltimeparams)
+SSI::SSIBase::SSIBase(const MPI_Comm comm, const Teuchos::ParameterList& globaltimeparams)
     : AlgorithmBase(*SSI::Utils::problem_from_instance(), comm, globaltimeparams),
       diff_time_step_size_(globaltimeparams.get<bool>("DIFFTIMESTEPSIZE")),
       fieldcoupling_(Teuchos::getIntegralValue<SSI::FieldCoupling>(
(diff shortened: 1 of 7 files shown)


bgoderbauer (dismissed):

src/ssi/4C_ssi_base.cpp
```diff
@@ -76,29 +76,24 @@ SSI::SSIBase::SSIBase(MPI_Comm comm, const Teuchos::ParameterList& globaltimepar
...
-        "Please set 'CONVFORM' to 'convective' in the SCALAR TRANSPORT DYNAMIC section.");
-  }
+  FOUR_C_ASSERT_ALWAYS(convform == ScaTra::convform_conservative or isintensive,
+      "Inconsistent scalar transport formulation on a deforming domain: "
+      "The scalar is defined as volume-referenced (IS_INTENSIVE_SCALAR = false), "
+      "therefore the conservative form is required to account for volume changes. "
+      "Please set 'CONVFORM' to 'conservative' in the SCALAR TRANSPORT DYNAMIC section.");
+
+  FOUR_C_ASSERT_ALWAYS(convform != ScaTra::convform_conservative or !isintensive,
+      "Inconsistent scalar transport formulation on a deforming domain: "
```
dragos-ana: My personal opinion, but I would find these checks slightly easier to understand if they were in the form:

```
if (isintensive){
// check for conservative form
...
}
else {
// check for convective form
...
}

```

src/ssi/4C_ssi_base.cpp
```diff
@@ -585,21 +580,21 @@ void SSI::SSIBase::evaluate_and_set_temperature_field()
 
 /*----------------------------------------------------------------------*/
 /*----------------------------------------------------------------------*/
-void SSI::SSIBase::set_velocity_fields(std::shared_ptr<const Core::LinAlg::Vector<double>> vel)
+void SSI::SSIBase::set_velocity_fields(const Core::LinAlg::Vector<double>& vel) const
```
dragos-ana, line 586: If I understand this correctly, it's not really const because the internal state of `ssicoupling_` is modified through the setter, which is allowed here because `ssicoupling_` is a shared_ptr.

I think that it might be more "honest" if this setter, and other setters e.g., for the mesh displacement, remain non-const, but this would require reverting a lot of the const-setting you did. You can better decide how the API should look like, I just wanted to note that.

dragos-ana (commented):
Looks good to me. Added two non-blocking comments.

dragos-ana (dismissed):

src/ssi/4C_ssi_base.cpp
c-p-schmidt: As discussed offline, we will keep it as is and rework object ownership within SSI in follow-up PRs.

dragos-ana (approved):

lkoeglmeier (approved):

PR #2205, opened 2026-08-20
Implement L1 time integration scheme for FSLS

bgoderbauer:
## Description and Context

Adds a second time-integration scheme for the fractional derivative in the FSLS model (fractional standard linear solid, see [A. Birzle, W. A. Wall, A viscoelastic nonlinear compressible material model of lung parenchyma – Experiments and numerical identification](https://doi.org/10.1016/j.jmbbm.2019.02.024))

It is  selected via a new `SOLVE` parameter on the existing `VISCO_FSLS` material.  `SOLVE` defaults to `GrunwaldLetnikov` (the current behavior), and the new option is `L1` (uniform-grid L1 scheme), which has a theoretical order of $2-\alpha$. It is a classical solution scheme for fractional derivatives, which assumes step-wise linear interpolation of the function under the fractional derivative and directly integrates the history convolution. Therefore, it still needs the full history of the target function.

### Formulas

The FSLS evolution equation being discretized:

```math
\tau^\alpha \, {}^{C}\!D_t^\alpha \hat{Q} + \hat{Q} = \beta S^\infty,
\qquad
S^{\mathrm{eff}} = (1+\beta) S^\infty - \hat{Q}
```

where $S^\infty$ is the equilibrium (hyperelastic) stress, $\hat{Q}$ is the FSLS internal stress-like variable, $\tau$ the relaxation time, $\alpha \in [0,1)$ the fractional order, $\beta \ge 0$ the viscous-to-elastic weighting, and $S^{\mathrm{eff}}$ is the effective stress including the viscous contributions.

**L1 weights:** On a uniform time grid of step $h$:

```math
K_k^h = \frac{(\tau/h)^\alpha}{\Gamma(2-\alpha)}
        \left[(k+1)^{1-\alpha} - k^{1-\alpha}\right], \qquad k \ge 0
```

**History remainder:** With committed increments

```math
\Delta\hat{Q}_j=\hat{Q}_{j+1}-\hat{Q}_j:
```
```math
r_n^Q = -K_0^h \hat{Q}_n + \sum_{k=1}^{n} K_k^h \, \Delta \hat{Q}_{n-k}
```

**Closed-form update:** 

```math
\hat{Q}_{n+1} = \frac{\beta S^\infty_{n+1} - r_n^Q}{1 + K_0^h},
\qquad
\mathbb{C}^{\mathrm{eff}} = \left[(1+\beta) - \frac{\beta}{1+K_0^h}\right] \mathbb{C}^\infty
```

**Numerical robustness:**
- The bracket $(k+1)^{1-\alpha} - k^{1-\alpha}$ is evaluated via an `expm1`/`log1p` reformulation instead of direct subtraction, avoiding catastrophic cancellation for large $k$ (verified against naive subtraction up to $k = 10^8$).
- The prefactor $(\tau/h)^\alpha / \Gamma(2-\alpha)$ is evaluated in log-space to avoid overflow/underflow for extreme $\tau/h$ ratios.
- $\alpha = 0$ is special-cased to the exact value $1$, so a degenerate $\tau/h$ can never produce a $0 \times \log(\infty)$ NaN.
- The L1 scheme requires a uniform time step; a mid-simulation change of $h$ is detected and rejected with a clear error instead of silently mixing weights from two different grids.

### History handling

The committed $\hat{Q}$ history lives in the existing `ViscoElastState`/`FslsState` infrastructure (the same per-Gauss-point,
full, untruncated history container already used by the Grünwald-Letnikov scheme), since it is needed for restart and MPI-communication support.

Helpful caches specific to the L1 scheme, which are useful to avoid recomputations: 

- **Weight cache** (`FslsL1WeightCache`): caches $K_k^h$ per element/contribution, since the weights depend only on $(\alpha, \tau, h)$, not on Gauss point or history content. Extended lazily as $k$ grows with the simulation.
- **History-remainder cache** (`FslsL1StepCache`): caches $r_n^Q$ per Gauss point. Since $r_n^Q$ is constant across all Newton iterations of one time step, it's recomputed only when the committed history has actually grown (detected by a change in history length, which increases by exactly one entry per converged step) rather than on every Newton iteration.

### Testing

- New unittests are added for the L1 scheme, which test its kernels and numerical robustness.
- One small regression test for the new discretization scheme mirroring the existing FSLS test with Grünwald-Letnikov discretization.

Claude Sonnet 5 was used for architecture design, refactoring, documentation, and test design of the implementation.


src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
```diff
@@ -24,13 +25,17 @@ namespace Mat::ViscoElast
...
+        solve_(matdata.parameters.get<std::string>("SOLVE"))
   {
     FOUR_C_ASSERT_ALWAYS(tau_ > 0.0,
         "Invalid TAU={} in VISCO_FSLS (MAT {}). TAU has to be positive.", tau_, matdata.id);
 
     FOUR_C_ASSERT_ALWAYS(alpha_ >= 0.0 && alpha_ < 1.0,
         "Invalid ALPHA={} in VISCO_FSLS (MAT {}). Expected 0 <= ALPHA < 1.", alpha_, matdata.id);
+
+    FOUR_C_ASSERT_ALWAYS(beta_ >= 0.0,
+        "Invalid BETA={} in VISCO_FSLS (MAT {}). Expected BETA >= 0.", beta_, matdata.id);
```
lauraengelhardt: You can use the validators directly in the definition of the parameters.

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2452,6 +2452,11 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
             parameter<double>("TAU", {.description = "relaxation parameter"}),
             parameter<double>("ALPHA", {.description = "fractional order derivative"}),
             parameter<double>("BETA", {.description = "emphasis of viscous to elastic part"}),
+            deprecated_selection<std::string>("SOLVE", {"GrunwaldLetnikov", "L1"},
```
lauraengelhardt: I would recommend to use an enum parameter here instead of the deprecated_selection. We would like to get rid of that in the long run ;)

src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
```diff
@@ -161,10 +176,28 @@ namespace Mat::ViscoElast
...
+  {
+    if (solve == "GrunwaldLetnikov") return FslsSolveKind::grunwald_letnikov;
+    if (solve == "L1") return FslsSolveKind::l1;
+
+    FOUR_C_THROW(
+        "Invalid SOLVE='{}' in VISCO_FSLS for MAT_ViscoElastHyper (MAT {}, GP {}, ELE {}). Use "
+        "GrunwaldLetnikov or L1.",
+        solve, visco_mat_id, gp, ele_gid);
+  }
+
```
lauraengelhardt: If you use an enum parameter you can get rid of all these unnecessary lines.

src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
```diff
@@ -161,10 +176,28 @@ namespace Mat::ViscoElast
         "{}, ELE {}). Expected 0 <= ALPHA < 1.",
         metadata.alpha, metadata.summand_mat_id, point.visco_mat_id, point.gp, point.ele_gid);
 
+    FOUR_C_ASSERT_ALWAYS(metadata.beta >= 0.0,
+        "Invalid BETA={} in VISCO_FSLS (MAT {}, referenced by MAT_ViscoElastHyper MAT {}, GP {}, "
+        "ELE {}). Expected BETA >= 0.",
+        metadata.beta, metadata.summand_mat_id, point.visco_mat_id, point.gp, point.ele_gid);
+
```
lauraengelhardt: I think this is not necessary. You validate it once in the beginning, why would it change?

src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
```diff
@@ -186,80 +219,272 @@ namespace Mat::ViscoElast
...
+          input.dt, input.visco_mat_id, input.gp, input.ele_gid);
+
+      FOUR_C_ASSERT_ALWAYS(input.tau > 0.0,
+          "Invalid TAU={} in FSLS kernel evaluation (MAT {}, GP {}, ELE {}). Expected TAU > 0.",
+          input.tau, input.visco_mat_id, input.gp, input.ele_gid);
+
+      FOUR_C_ASSERT_ALWAYS(input.alpha >= 0.0 && input.alpha < 1.0,
+          "Invalid ALPHA={} in FSLS kernel evaluation (MAT {}, GP {}, ELE {}). Expected 0 <= "
+          "ALPHA < 1.",
+          input.alpha, input.visco_mat_id, input.gp, input.ele_gid);
```
lauraengelhardt: Wouldn't this also be covered by the validators?

src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
```diff
@@ -186,80 +219,272 @@ namespace Mat::ViscoElast
...
+      if (alpha == 0.0) return 1.0;
 
-    q_current_for_history.update(lambda_1 * input.beta, stress, 0.);
-    q_current_for_history.update(lambda_2, q_history_sum, 1.);
+      const double log_c = alpha * std::log(tau / dt) - std::lgamma(2.0 - alpha);
+      return std::exp(log_c);
+    }
+
+
+    /// Returns K_k^h from the contribution-wide cache, extending it as needed. The L1 scheme is
```
lauraengelhardt: What do you mean with "extending it as needed"?

src/mat/viscoelast/4C_mat_viscoelast_fsls.hpp
```diff
@@ -16,18 +16,99 @@
...
+   *   \hat Q_{n+1} = \frac{\beta S^\infty_{n+1} - r_n^Q}{1 + K_0^h}, \qquad
+   *   \mathbb{C}^{\mathrm{eff}} = \left[(1+\beta) - \frac{\beta}{1+K_0^h}\right] \mathbb{C}^\infty.
+   * \f]
+   * (See Mat::ViscoElast::Kernels::evaluate_fsls_kernel for where these are actually assembled.)
+   *
+   * \f$r_n^Q\f$ depends only on committed history, never on the current Newton trial, so it needs
+   * to be recomputed only once per converged time step and can be reused across every Newton
+   * iteration within that step. Committed history strictly grows by exactly one entry per
+   * converged step, so comparing the stored history size against the current one is a sufficient
+   * staleness check; no separate step counter is needed.
```
lauraengelhardt: Is this part really still needed? I don't think you (or the AI) should explain what is not implemented.

src/mat/viscoelast/4C_mat_viscoelast_fsls.hpp
```diff
@@ -16,18 +16,99 @@
...
+  };
+
+  /// Time-discretization scheme for the FSLS fractional derivative, selected by the material's
+  /// SOLVE parameter.
+  enum class FslsSolveKind
+  {
+    grunwald_letnikov,
+    l1
+  };
+
```
lauraengelhardt, line 99: See - you have the enums already here, you can directly use those for you input.

tests/input_files/visco_fsls_l1_history_h8_small.4C.yaml
lauraengelhardt, line 1: Could you remove all unused sections (e.g. the DISCRETISATION part)?

tests/list_of_tests.cmake
```diff
@@ -2429,6 +2429,7 @@ __four_c_test_restart(BASED_ON ${current} SAME_FILE NP 2 RESTART_STEP 50)
 four_c_test(TEST_FILE viscoelasthyper_split_precedence.4C.yaml NP 2 RETURN_AS current)
 __four_c_test_restart(BASED_ON ${current} SAME_FILE NP 2 RESTART_STEP 50)
 four_c_test(TEST_FILE visco_fsls_history_h8_small.4C.yaml NP 2 RETURN_AS current)
```
lauraengelhardt: Do you want to rename this other test, so its directly clear this uses the other scheme?

unittests/mat/viscoelast/4C_mat_viscoelast_fsls_l1_test.cpp
```diff
@@ -0,0 +1,519 @@
...
+  using Mat::ViscoElast::FslsL1StepCache;
+  using Mat::ViscoElast::FslsL1WeightCache;
+  namespace Kernels = Mat::ViscoElast::Kernels;
+  using Kernels::FslsHistory;
+  using Kernels::FslsKernelInput;
+  using Kernels::FslsStressVector;
+  using Kernels::FslsTangentMatrix;
+  using Mat::ViscoElast::FslsSolveKind;
+
+  /// Birzle/Wall median FSLS parameters for viable lung parenchyma, used throughout as a
```
lauraengelhardt: Could you add the citation or of not relevant then just remove here? I don't think that for the test is actually matters, or what do you think? Like this it is obvious to us, but confusing to other people ;)

unittests/mat/viscoelast/4C_mat_viscoelast_fsls_l1_test.cpp
```diff
@@ -0,0 +1,519 @@
...
+        lgammal(2.0L - alpha);
+    const long double c = expl(log_c);
+    if (k == 0) return c;
+
+    const long double kd = static_cast<long double>(k);
+    return c * (powl(kd + 1.0L, p) - powl(kd, p));
+  }
+
+  /// Naive double-precision bracket evaluation (no expm1/log1p robustification), used to
+  /// demonstrate that the production formula is measurably more accurate for large k.
```
lauraengelhardt: What is the "production formula"?

unittests/mat/viscoelast/4C_mat_viscoelast_fsls_l1_test.cpp
```diff
@@ -0,0 +1,519 @@
...
+      input.solve_kind = FslsSolveKind::l1;
+      input.previous_history = &history;
+      input.l1 = {&cache, &step_cache};
+
+      FslsStressVector q_current(Core::LinAlg::Initialization::zero);
+      FslsStressVector q_additive(Core::LinAlg::Initialization::zero);
+      FslsTangentMatrix cmatq_additive(Core::LinAlg::Initialization::zero);
+      Kernels::evaluate_fsls_kernel(stress, cmat, q_current, q_additive, cmatq_additive, input);
+      return q_additive;
+    };
```
lauraengelhardt, line 334: If I am not mistaken many things are similar between the tests. Could you move setup to e..g a common `FslsKernelInput` construction?

lauraengelhardt (changes requested):
I cannot comment on the method itself ofc, but have some general comments to the code. Let me know what you think! :)

src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
bgoderbauer: Thanks, I forgot about the validators! I removed the asserts inside the code in favor of them

src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
bgoderbauer: This means that once per time step, the cache with the history weights is extended by the newly needed weight. The weights aren't computed once in advance, and also not every time they are needed. I've updated the description.

src/mat/viscoelast/4C_mat_viscoelast_fsls.hpp
bgoderbauer: I've changed it to only use this enum class.

unittests/mat/viscoelast/4C_mat_viscoelast_fsls_l1_test.cpp
bgoderbauer: I've added the paper info.

unittests/mat/viscoelast/4C_mat_viscoelast_fsls_l1_test.cpp
bgoderbauer: It means just the way it is implemented in the source code. There's a different arithmetic used for evaluation, so this is tested here for its numerical robustness.

unittests/mat/viscoelast/4C_mat_viscoelast_fsls_l1_test.cpp
bgoderbauer: you're right, I've extracted the relevant setup details from the different tests.

bgoderbauer:
Thanks for the review @lauraengelhardt :smile:. The input parameter handling looks much better now!

lauraengelhardt (dismissed):
Nice, thanks for incorporating the changes!

src/global_legacy_module/4C_global_legacy_module_validmaterials.cpp
```diff
@@ -2447,11 +2448,21 @@ std::unordered_map<Core::Materials::MaterialType, Core::IO::InputSpec> Global::v
...
     known_materials[Core::Materials::mes_fsls] = group("VISCO_FSLS",
         {
-            parameter<double>("TAU", {.description = "relaxation parameter"}),
-            parameter<double>("ALPHA", {.description = "fractional order derivative"}),
-            parameter<double>("BETA", {.description = "emphasis of viscous to elastic part"}),
+            parameter<double>(
+                "TAU", {.description = "relaxation parameter", .validator = positive<double>()}),
+            parameter<double>("ALPHA", {.description = "fractional order derivative",
+                                           .validator = in_range(0.0, excl(1.0))}),
+            parameter<double>("BETA", {.description = "emphasis of viscous to elastic part",
```
dragos-ana, line 2459: It was there in the old version too, but wouldn't `weighting` be a more descriptive term instead of `emphasis`?

src/mat/viscoelast/4C_mat_viscoelast_fsls.hpp
```diff
@@ -22,12 +22,92 @@ FOUR_C_NAMESPACE_OPEN
...
+  struct FslsL1State
+  {
+    FslsL1WeightCache* weights = nullptr;
+    FslsL1StepCache* step = nullptr;
+  };
+
+  /// Time-discretization scheme for the FSLS fractional derivative, selected by the material's
+  /// SOLVE parameter.
+  enum class FslsSolveKind
+  {
```
dragos-ana, line 95: Maybe `FslsSolutionScheme`? And the parameter could accordingly be called `SOLUTION_SCHEME`?

src/mat/viscoelast/4C_mat_viscoelast_fsls.hpp
```diff
@@ -22,12 +22,92 @@ FOUR_C_NAMESPACE_OPEN
...
+  /**
+   * \brief Cache of L1-scheme fractional-derivative weights shared by all Gauss points of one
+   * FSLS contribution instance.
+   *
+   * The weights \f$K_k^h\f$ depend only on (alpha, tau, dt), not on Gauss point or history
+   * content, so they are computed once per contribution and reused/extended as the history grows.
+   * The L1 scheme requires a uniform time step; if dt changes between evaluations the cache is
+   * invalidated and rebuilt from scratch rather than silently mixing weights from two different
+   * grids. See FslsL1StepCache for the L1 discretization this cache feeds into, and
+   * Kernels::evaluate_l1_weight for the formula and its numerically robust evaluation.
```
dragos-ana, line 34: I think that referencing works this way in Doxygen, although I am also not entirely sure.

```suggestion
   * \ref Kernels::evaluate_l1_weight for the formula and its numerically robust evaluation.
```

src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
```diff
@@ -186,80 +182,267 @@ namespace Mat::ViscoElast
...
+      else
+      {
+        constexpr double relative_grid_tolerance = 1.0e-12;
+        const bool grid_matches =
+            cache.alpha == alpha && cache.tau == tau &&
+            std::abs(cache.dt - dt) <= relative_grid_tolerance * std::max(cache.dt, dt);
+        FOUR_C_ASSERT_ALWAYS(grid_matches,
+            "FSLS L1 scheme requires a uniform time step, but the time step size changed from "
+            "{} to {} (MAT {}, GP {}, ELE {}). A variable-step L1 scheme is a different, "
+            "unimplemented method; keep TIMESTEP constant while VISCO_FSLS uses SOLVE l1.",
```
dragos-ana, line 327: I am not acquainted with the exact method, but this does not tell the full story - the grids could be non-matching due to different values of alpha and tau as well, at least in this check.

src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
```diff
@@ -186,80 +182,267 @@ namespace Mat::ViscoElast
...
+
+  double Kernels::evaluate_l1_weight(
+      const std::size_t k, const double alpha, const double tau, const double dt)
+  {
+    FOUR_C_ASSERT_ALWAYS(dt > 0.0,
+        "Invalid time step size dt={} in FSLS L1 weight evaluation. Expected dt > 0.", dt);
+    FOUR_C_ASSERT_ALWAYS(
+        tau > 0.0, "Invalid TAU={} in FSLS L1 weight evaluation. Expected TAU > 0.", tau);
+    FOUR_C_ASSERT_ALWAYS(alpha >= 0.0 && alpha < 1.0,
+        "Invalid ALPHA={} in FSLS L1 weight evaluation. Expected 0 <= ALPHA < 1.", alpha);
```
dragos-ana, line 443: Here you have again the checks for the parameter bounds although you know, at least from the input, that tau and alpha will be within the set bounds.

Maybe having a dedicated struct for these parameters (time integration parameters?) and passing them to the various functions would be a good idea to avoid this uncertainty? Something along the lines of:


```
struct FslsTimIntParams
{
  FslsTimIntParams(const double input_alpha, const double input_tau, const double input_beta)
      : alpha(input_alpha), tau(input_tau), beta(input_beta)
  {
    FOUR_C_ASSERT_ALWAYS(alpha >= 0.0 && alpha < 1.0, ...);
    FOUR_C_ASSERT_ALWAYS(tau > 0.0, ...);
    FOUR_C_ASSERT_ALWAYS(beta >= 0.0, ...);
  }

  const double alpha;
  const double tau;
  const double beta;
};
```

unittests/mat/viscoelast/4C_mat_viscoelast_fsls_l1_test.cpp
```diff
@@ -0,0 +1,475 @@
...
+    }
+    return history;
+  }
+
+  TEST(FslsL1KernelTest, ClosedFormRoundTripsExactTargetArtificialStress)
+  {
+    constexpr std::size_t n = 5;
+    constexpr double dt = 0.01;
+    FslsHistory history = make_synthetic_history(n);
+    const auto& at_gp = history[0];
```
dragos-ana, line 183:
```suggestion
    const auto& history_at_gp = history[0];
```

unittests/mat/viscoelast/4C_mat_viscoelast_fsls_l1_test.cpp
```diff
@@ -0,0 +1,475 @@
...
+            (q_additive_plus(j) - q_additive_minus(j)) / (2.0 * step);
+        EXPECT_NEAR(numerical_derivative, cmatq_additive(j, i), 1.0e-8)
+            << "row=" << j << " col=" << i;
+      }
+    }
+  }
+
+  /// Analytic Mittag-Leffler relaxation solution for a constant equilibrium stress applied at
+  /// t=0+ with zero initial memory: Qhat(t) = beta*S0*(1 - E_alpha(-(t/tau)^alpha)).
+  double mittag_leffler_e_alpha(const double x, const double alpha)
```
dragos-ana, line 366: I think it might be helpful to document what x is, since it does not (directly) appear in the formula you added.

dragos-ana (commented):
Looks good to me! I added a few cosmetic suggestions.

src/mat/viscoelast/4C_mat_viscoelast_fsls.hpp
bgoderbauer: I wanted to keep the same naming as in the `GeneralizedMaxwell` model. If we change it, we should change it there as well for consistency, but I'd rather not add it to this PR.

src/mat/viscoelast/4C_mat_viscoelast_fsls.cpp
bgoderbauer: This is a free function for accessing the internals, used only for unit testing. It is not used in the regular input file workflow, which is why the safeguards are added here. I've added a comment :smiley:

bgoderbauer:
Thanks for the suggestions @dragos-ana, I've added them to the code!

dragos-ana (approved):

lauraengelhardt (approved):

PR #2206, opened 2026-08-20
Reduced lung: Volume dependent pleural pressure bc 

maxiludwig:
## Description and Context

Adds a boundary condition that prescribes the pleural pressure of a boundary node from the total volume of all terminal units, instead of requiring it to be prescribed by hand as a function of time.

The new `volume_dependent_pleural_pressure` input list constrains the pressure dof of every node carrying its `id`, following a normalized linear-exponential pressure-volume curve. Only the `Frozen` coupling is implemented (and is the default): it uses the volume of the last converged timestep, so the value is constant during a Newton solve and the equation stays a Dirichlet row.

Two supporting changes: the evaluators now take an `AssemblyContext` instead of a bare time, so a condition can depend on more state without touching every signature; and how a value is produced is modelled by a `ValueModel` variant on `BoundaryConditionModel`, replacing the old `Type` enum that conflated the constrained variable with the source of the value.

Verified by unit tests for the assembly, the curve, and the new input checks, plus an input test on 1 and 3 ranks that covers parsing and the collective volume reduction.

Implemented with Claude Code (Claude Opus 5): code generation, refactoring, tests, and commit messages. All changes were reviewed and validated by me.

### DIFF
--- a/src/reduced_lung/src/4C_reduced_lung_boundary_conditions.cpp
+++ b/src/reduced_lung/src/4C_reduced_lung_boundary_conditions.cpp
@@ -16,6 +16,7 @@
 #include "4C_utils_function_manager.hpp"
 #include "4C_utils_function_of_time.hpp"
 
+#include <cstdint>
 #include <map>
 #include <ranges>
 #include <tuple>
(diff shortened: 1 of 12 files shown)


bgoderbauer (dismissed):
Thanks for implementing this feature! I'm not quite sure, but maybe some of the abstract definition and indirection layers in the boundary_conditions.cpp file could be simplified. It is a little hard to follow life-cycle and workflow of the variables and how boundary conditions are defined and reached in a simulation run

maxiludwig:
I cleaned it up a bit following you comment. Nothing substantial changed.

bgoderbauer (approved):

lkoeglmeier (approved):

Issue #2207, opened 2026-08-21
[beaminteraction] Allow for multiple assembly managers in saddle-point formulation and make routines safer


maxfirmbach:
## Description
Currently we restrict the saddlepoint formulation for beam contact to one assembly manager only. Using another one in addition, e.g. due to beam coupling conditions, bad things happen due to the following two functions which are not safe:
```cpp
bool BeamInteraction::SubmodelEvaluator::BeamContact::have_lagrange_dofs() const
{
  if (assembly_managers_.size() == 0) return false;

  const auto indirect_assembly_manager =
      std::dynamic_pointer_cast<BeamContactAssemblyManagerInDirect>(assembly_managers_[0]);
  if (indirect_assembly_manager)
  {
    const auto mortar_manager = indirect_assembly_manager->get_mortar_manager();
    if (mortar_manager)
    {
      const bool have_lagrange_dofs = mortar_manager->have_lagrange_dofs();
      if (have_lagrange_dofs && assembly_managers_.size() != 1)
        FOUR_C_THROW("Only working for single assembly manager");
      return have_lagrange_dofs;
    }
  }
  return false;
}
```
and
```cpp
std::shared_ptr<const BeamInteraction::SubmodelEvaluator::BeamContactAssemblyManagerInDirect>
BeamInteraction::SubmodelEvaluator::BeamContact::get_lagrange_multiplier_assembly_manager() const
{
  if (!std::dynamic_pointer_cast<BeamContactAssemblyManagerInDirect>(assembly_managers_[0]))
    FOUR_C_THROW("Expected a BeamContact Assembly Manager");
  if (assembly_managers_.size() != 1) FOUR_C_THROW("Only working for single assembly manager");
  return std::dynamic_pointer_cast<BeamContactAssemblyManagerInDirect>(assembly_managers_[0]);
}
```
.
Both are using hard coded arrays ``assembly_managers_[0]``, which can result in undefined data structures later one. We should fix this and find a way to remove this behavior and the hard coded zero entry.

## Possible Solution and Definition of Done
Provide a way to handle multiple assembly managers.

## Interested Parties

@isteinbrecher 

## Related to

https://github.com/4C-multiphysics/4C/issues/1141

isteinbrecher:
This is related to https://github.com/4C-multiphysics/4C/issues/1141 (I added it to the description)

Issue #1141, opened 2025-08-21
Improve Lagrange Multiplier implementation to deal with multiple interaction conditions


dharinib98:
## Description

Currently the use of Lagrange Multipliers for beaminteraction problems are is restricted to allow only one type of interaction(see #1004  and #637 ). The aim of this issue is to extend it to handle both multiple interaction types and conditions. 
## Possible Solution and Definition of Done

- [ ]  Find a suitable place to store the Lagrange maps
- [ ] Modify current to loop over all available Assembly Managers and conditions. 
## Interested Parties

@isteinbrecher @mayrmt @maxfirmbach

maxfirmbach:
@mayrmt @dharinib98 @isteinbrecher I have a case, where I want to apply two type of constraints, one from `beaminteraction` + one from the `constraint_framework`. Currently this is not possible as we always calculate the rowmap of the constraint block as `discret()->dof_row_map()->max_all_gid() + 1`. Thus we would have non-unique GIDs during assembly as both constraint blocks would start at the same GID.

Is this issue meant to solve such type of things?

dharinib98:
> [@mayrmt](https://github.com/mayrmt) [@dharinib98](https://github.com/dharinib98) [@isteinbrecher](https://github.com/isteinbrecher) I have a case, where I want to apply two type of constraints, one from `beaminteraction` + one from the `constraint_framework`. Currently this is not possible as we always calculate the rowmap of the constraint block as `discret()->dof_row_map()->max_all_gid() + 1`. Thus we would have non-unique GIDs during assembly as both constraint blocks would start at the same GID.
> 
> Is this issue meant to solve such type of things?

@maxfirmbach I think this only tackles the different interaction conditions within `beaminteraction` itself or the conditions handled by the mortar manager, and I think that is not the case with the `constraint_framework`. Also I am not quite sure about the `constraint_framework` part of the code, so I am not sure if this implementation can in someway be extended to help solve your case.

isteinbrecher:
[...]

Issue #1004, opened 2025-07-24
Beaminteraction using Lagrange Multipliers


dharinib98:
## Description

For beaminteraction problems, an additional feature needs to be implemented that enforces constraints with the use of lagrange multipliers as an alternative to the penatly approach. With the major changes already on the way, this issue serves to track the post implementation work and some major issues that need to be fixed as part of this implementation.

## Possible Solution and Definition of Done

- [ ] Make this work in parallel
- [ ] Deal with the case of multiple beam interaction coupling conditions
- [ ] Design input parameters to select between penalty and Lagrange multipliers for constraint enforcement and the type of formulation (condensed, block matrix).

## Interested Parties

@isteinbrecher @mayrmt @maxfirmbach

maxfirmbach:
@dharinib98 It seems like a lot of things have already been finished here 😃. What is missing to get this issue done?

dharinib98:
@maxfirmbach with #2209 done, we can close this. that was the last thing missing. Thanks for taking care of that:)

isteinbrecher:
> @maxfirmbach with https://github.com/4C-multiphysics/4C/pull/2209 done, we can close this. that was the last thing missing. Thanks for taking care of that:)

As far as I can see, #2209 does not resolve this issue, it just enables one Lagrange multiplier assembly manager in combination with multiple penalty ones. For this issue, we have to figure out how to take care of multiple Lagrange multiplier assembly managers.

PR #2209, opened 2026-08-24
[beaminteraction] Allow for several assembly managers in saddle-point formulation

maxfirmbach:
## Description and Context

This Draft-PR exists to gather ideas on how to make the beam-solid saddlepoint formulation work with several assembly managers. In the current state, it just comments out error messages and introduces loops over the assembly managers in the respective function calls that matter. Currently there is also no testing included.

Related to #1141, #2207


isteinbrecher:
@maxfirmbach thanks for the changes, nice that this is now working. A complete resolution of this issue (arbitrary number of assembly managers with either Lagrange or penalty approach) will likely be more complicated, as we have to figure out how to correctly assign the Lagrange multiplier DOFs in that case.

If I understand your case correctly (one Lagrange multiplier assembly manager and some penalty ones) this should work fine. However, we should still leave in an error, if multiple Lagrange multiplier assembly managers are present.

maxfirmbach:
@isteinbrecher Is there a way to distinguish assembly managers based on penalty or lagrange multiplier constraint enforcement (e.g. direct vs indirect?)?

isteinbrecher:
> @isteinbrecher Is there a way to distinguish assembly managers based on penalty or lagrange multiplier constraint enforcement (e.g. direct vs indirect?)?

It should be pretty easy. Take this snippet:
```c++
const auto mortar_manager = indirect_assembly_manager->get_mortar_manager();
if (mortar_manager)
{
  const bool have_lagrange_dofs = mortar_manager->have_lagrange_dofs();
}
```
If `have_lagrange_dofs` is true, then the assembly manager holds Lagrange Multipliers that will end up in the final system of equations.

maxfirmbach:
Putting out of draft-mode. Should we include a small test, e.g. our minimal example of a beam lattice embedded in a solid cube?

isteinbrecher (commented):
We should still raise an error if we have multiple Lagrange multiplier assembly managers, as this is not covered by the other code.

And a test would also be nice, combining penalty point couplings (indirect?) with Lagrange multiplier beam-to-solid coupling.

src/beaminteraction/src/contact/4C_beaminteraction_contact_submodel_evaluator.cpp
```diff
@@ -244,33 +244,46 @@ void BeamInteraction::SubmodelEvaluator::BeamContact::setup()
...
-  if (assembly_managers_.size() != 1) FOUR_C_THROW("Only working for single assembly manager");
-  return std::dynamic_pointer_cast<BeamContactAssemblyManagerInDirect>(assembly_managers_[0]);
+  for (const auto& assembly_manager : assembly_managers_)
+  {
+    auto direct_manager =
+        std::dynamic_pointer_cast<BeamContactAssemblyManagerInDirect>(assembly_manager);
+    if (direct_manager) return direct_manager;
+  }
+
+  return nullptr;
```
isteinbrecher: We should also have a similar check as in `have_lagrange_dofs` here.

maxfirmbach: I adapted the method. In the current state it does not throw any errors as `have_lagrange_dofs` takes care of that already, still, the current version returns the assembly manager for lagrange multipliers if exactly one exists, otherwise it throws due to `have_lagrange_dofs`.

maxfirmbach:
I included two simple test cases, one with beam-beam point coupling and one with the penalty-based variant.

dharinib98:
As far I understood, this one helps us solve more than one assembly managers while using penalty and not the saddle point. So the actual requirement in the issue #1004 still needs to be worked on right ? (For example this will still not work if I use the saddle point formulation to run a model with two fully embedded beams because that is the current state of the saddle point approach  now)

maxfirmbach:
@dharinib98 Yep, this PR's main goal is to make one "lagrange-multiplier"-based assembly manager work with other penalty-based one. It is still only allowed to have one "lagrange-multiplier"-based one though. Prior to this PR, the saddle point system was only allowed to be build with one assembly manager at all.

src/beaminteraction/src/contact/4C_beaminteraction_contact_submodel_evaluator.cpp
```diff
@@ -244,33 +244,56 @@ void BeamInteraction::SubmodelEvaluator::BeamContact::setup()
...
 /*----------------------------------------------------------------------*
  *----------------------------------------------------------------------*/
 std::shared_ptr<const BeamInteraction::SubmodelEvaluator::BeamContactAssemblyManagerInDirect>
 BeamInteraction::SubmodelEvaluator::BeamContact::get_lagrange_multiplier_assembly_manager() const
 {
-  if (!std::dynamic_pointer_cast<BeamContactAssemblyManagerInDirect>(assembly_managers_[0]))
-    FOUR_C_THROW("Expected a BeamContact Assembly Manager");
-  if (assembly_managers_.size() != 1) FOUR_C_THROW("Only working for single assembly manager");
-  return std::dynamic_pointer_cast<BeamContactAssemblyManagerInDirect>(assembly_managers_[0]);
+  if (!have_lagrange_dofs()) return nullptr;
```
isteinbrecher: A simple call to `have_lagrange_dofs` should be fine, as we did not return a null pointer before.
```suggestion
// Check that there is exactly one assembly manager carrying Lagrange multipliers.
have_lagrange_dofs();
```

src/beaminteraction/src/contact/4C_beaminteraction_contact_submodel_evaluator.cpp
```diff
@@ -244,33 +244,56 @@ void BeamInteraction::SubmodelEvaluator::BeamContact::setup()
...
+
+    const auto mortar_manager = indirect_manager->get_mortar_manager();
+
+    if (mortar_manager && mortar_manager->have_lagrange_dofs())
+    {
+      return indirect_manager;
+    }
+  }
+
+  return nullptr;
```
isteinbrecher: Lets raise an error here, if no assembly manager with Lagrange Multipliers could be found.

tests/list_of_tests.cmake
```diff
@@ -2941,6 +2937,8 @@ __four_c_test_vtk(BASED_ON ${current} PVD_RESULT xxx-beam-to-solid-volume-segmen
 four_c_test(TEST_FILE beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_penalty_line4.4C.yaml NP 3 REQUIRED_DEPENDENCIES ArborX)
 four_c_test(TEST_FILE beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_restart_first.4C.yaml NP 2 RETURN_AS current REQUIRED_DEPENDENCIES ArborX)
 __four_c_test_restart(BASED_ON ${current} TEST_FILE beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_restart_second.4C.yaml NP 3 RESTART_STEP 1 REQUIRED_DEPENDENCIES ArborX)
+four_c_test(TEST_FILE beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_beam_to_beam_coupling.4C.yaml NP 1 REQUIRED_DEPENDENCIES ArborX)
+four_c_test(TEST_FILE beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_beam_to_beam_penalty_coupling.4C.yaml NP 1 REQUIRED_DEPENDENCIES ArborX)
```
isteinbrecher: Should we maybe run them with more than one processor?

tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_beam_to_beam_coupling.4C.yaml
isteinbrecher, line 1: Should we really add this file? Wouldn't that produce a non-invertible beam block, although we expect it to be invertible?

I would skip this test answer add it after we have investigated what is going wrong there.

tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_beam_to_beam_penalty_coupling.4C.yaml
```diff
@@ -0,0 +1,937 @@
...
+      DENS: 0.0
+      CROSSAREA: 0.0003141592653589793
+      SHEARCORR: 1.0
+      MOMINPOL: 1.5707963267948965e-08
+      MOMIN2: 7.853981633974483e-09
+      MOMIN3: 7.853981633974483e-09
+DESIGN POINT PENALTY COUPLING CONDITIONS:
+  - E: 4
+    POSITIONAL_PENALTY_PARAMETER: 1000.0
+    ROTATIONAL_PENALTY_PARAMETER: 100.0
```
isteinbrecher: We should replace a few of these conditions with `BEAM INTERACTION/BEAM TO BEAM POINT COUPLING CONDITIONS` - this will add indirect assembly managers and improve the coverage of this test.

We can talk about how to do that in person.

isteinbrecher:
```yaml
BEAM INTERACTION/BEAM TO BEAM POINT COUPLING CONDITIONS:
  - E: 1
    COUPLING_ID: 1
    PARAMETERS:
      POSITIONAL_PENALTY_PARAMETER: 1.2
      ROTATIONAL_PENALTY_PARAMETER: 1.3
      PROJECTION_VALID_FACTOR: 2.0
      CONSTRAINT_ENFORCEMENT: penalty_indirect
  - E: 2
    COUPLING_ID: 1
```
The sets `E` are element sets.

maxfirmbach:
@isteinbrecher I reworked the assembly manager and lagrange multiplier manager logic and refactored the testing.

tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_beam_to_beam_penalty_coupling.4C.yaml
maxfirmbach: For documentation purpose: Currently we hit a throw with using `penalty_indirect`.

isteinbrecher: Let's keep that in mind (can you open an issue and post the failing input file?), but we can go ahead with this PR, as it at least allows to use the direct point coupling pairs.

isteinbrecher (dismissed):

mayrmt (dismissed):

maxfirmbach:
@isteinbrecher It seems like the parameter `MAX_NUMBER_OF_PAIRS_PER_ELEMENT` has quite an influence. If I don't set it, it is initialized with zero and therefore the following loop is just not run through:
```cpp
    for (unsigned int i_dof = 0; i_dof < parameters_.n_lambda_node_translational; i_dof++)
      my_lambda_gid_translational.push_back(
          my_lambda_gid_start_value + i_dof + i_node * parameters_.n_lambda_node());
```

If I set `MAX_NUMBER_OF_PAIRS_PER_ELEMENT: 1` in the input file, things work as expected ... weird, maybe we should include checks, throws and asserts.

tests/input_files/beam3r_herm2line3_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_beam_to_beam_penalty_coupling.4C.yaml
maxfirmbach: Done. Setting `MAX_NUMBER_OF_PAIRS_PER_ELEMENT: 1` did the job.

isteinbrecher:
That makes sense! I thought `MAX_NUMBER_OF_PAIRS_PER_ELEMENT` was at default `1`, but it is 0. Yes there should be a check on this. I think we can even simply require `MAX_NUMBER_OF_PAIRS_PER_ELEMENT` to be larger or equal to 1.

isteinbrecher (approved):

dharinib98 (approved):

PR #2210, opened 2026-08-24
Fix normal pressure Neumann conditions on 2D solid lines

ischeider:
## Description and Context
Currently ` DESIGN LINE NEUMANN CONDITIONS` does not consider the `TYPE` parameter, particularly `TYPE=[pseudo_]orthopressure`. 
This issue is fixed here, i.e.,  orthopressure and pseudo_orthopressure are honored on plane solid lines; a test is added.
The pressure is scaled by thickness and add follower-load linearization.

GitHub Copilot CLI assisted with implementing the 2D solid-line `orthopressure` handling, follower-load linearization, thickness scaling, and associated unit and regression tests. 

The resulting code was compiled and validated using the solid-element unit suite and the QUAD9 quarter-cylinder regression test.

Reviewed and tested by me (to my best knowledge)


src/solid_ele/4C_solid_ele_neumann_evaluator.hpp
```diff
@@ -23,6 +27,60 @@ namespace Core::FE
...
+    double axial_derivative = 0.0;
+    for (int node = 0; node < Core::FE::num_nodes(celltype); ++node)
+    {
+      radial_derivative += shape.derivatives(0, node) * current_coordinates(node, 0);
+      axial_derivative += shape.derivatives(0, node) * current_coordinates(node, 1);
+    }
+
+    const std::array<double, 2> normal_measure{
+        reference_thickness * axial_derivative, -reference_thickness * radial_derivative};
+    for (int node = 0; node < Core::FE::num_nodes(celltype); ++node)
```
dragos-ana, line 55: The function implementation could be moved to the .cpp. Also, I think that it would benefit a lot from formulae added as comments for each step of the computation (along with doxygen for the parameters and so on ;) ). 

Also: does it need to be exposed in the hpp, or could it be simply implemented as a helper function in an anonymous namespace in the corresponding cpp? As far as I see, it's only used within `evaluate_normal_pressure`, and within the unit test you added.

UPDATE: After looking over the unit test it's fine for me to keep it here, as it makes testing the functionality a bit more accessible as described in the respective comment.

src/solid_ele/4C_solid_ele_neumann_evaluator.cpp
```diff
@@ -21,6 +22,99 @@
...
+    constexpr int num_nodes = Core::FE::num_nodes(celltype);
+    const int num_dof_per_node = element.num_dof_per_node(*element.nodes()[0]);
+    const auto& onoff = condition.parameters().get<std::vector<int>>("ONOFF");
+    const auto& values = condition.parameters().get<std::vector<double>>("VAL");
+    const auto& function_ids = condition.parameters().get<std::vector<std::optional<int>>>("FUNCT");
+    FOUR_C_ASSERT_ALWAYS(onoff.size() >= 2 && values.size() >= 2 && function_ids.size() >= 2,
+        "A two-dimensional pressure condition requires at least two entries in ONOFF, VAL, and "
+        "FUNCT.");
+    FOUR_C_ASSERT_ALWAYS(
+        onoff[0] == 1 && onoff[1] == 0, "Normal pressure must be activated on the first dof only.");
```
dragos-ana, line 43: I am not really acquainted with the orthopressure conditions, but this is the current convention right? This would mean that the function id and values should also be posed on the first dof, would it make sense to add these consistency conditions here as well?

src/solid_ele/4C_solid_ele_neumann_evaluator.cpp
```diff
@@ -21,6 +22,99 @@
...
+        current_coordinates(node, component) =
+            element.nodes()[node]->x()[component] +
+            local_displacement[node * num_dof_per_node + component];
+
+    const auto integration = Core::FE::create_gauss_integration<celltype>(
+        Discret::Elements::get_gauss_rule_stiffness_matrix<celltype>());
+    for (int gp = 0; gp < integration.num_points(); ++gp)
+    {
+      const auto xi = Core::Elements::evaluate_parameter_coordinate<celltype>(integration, gp);
+      Core::Elements::ElementNodes<celltype, 2> nodes{.coordinates = current_coordinates};
```
dragos-ana, line 63: This could be moved out of the GP loop.

src/solid_ele/4C_solid_ele_neumann_evaluator.cpp
```diff
@@ -21,6 +22,99 @@
...
+      const auto shape = Core::Elements::evaluate_shape_functions_and_derivs<celltype>(xi, nodes);
+
+      Core::LinAlg::Matrix<2, 1> current_coordinate;
```
dragos-ana, line 66: Same here, you could declare it out of the loop, and then update it via multiply_tn:


```suggestion
      current_coordinate.multiply_tn(1.0,current_coordinates, shape.values,0.0);
```

unittests/solid_ele/4C_solid_ele_neumann_evaluator_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+    auto evaluate = [&](const Core::LinAlg::Matrix<2, 2>& evaluated_coordinates,
+                        Core::LinAlg::SerialDenseMatrix* load_linearization)
+    {
+      Core::LinAlg::SerialDenseVector force(4);
+      const auto integration = Core::FE::create_gauss_integration<celltype>(
+          Discret::Elements::get_gauss_rule_stiffness_matrix<celltype>());
+      for (int gp = 0; gp < integration.num_points(); ++gp)
+      {
+        const auto xi = Core::Elements::evaluate_parameter_coordinate<celltype>(integration, gp);
+        const Core::Elements::ElementNodes<celltype, 2> nodes{.coordinates = evaluated_coordinates};
```
dragos-ana, line 39: Same here, could be moved out of the loop

dragos-ana, line 39: This could be moved out of the Gauss point loop

src/solid_ele/4C_solid_ele_neumann_evaluator.hpp
```diff
@@ -23,6 +27,60 @@ namespace Core::FE
...
+    for (int force_node = 0; force_node < Core::FE::num_nodes(celltype); ++force_node)
+      for (int coordinate_node = 0; coordinate_node < Core::FE::num_nodes(celltype);
+          ++coordinate_node)
+      {
+        const double factor = shape.values(force_node) * pressure_times_weight *
+                              reference_thickness * shape.derivatives(0, coordinate_node);
+        (*load_linearization)(
+            force_node* num_dof_per_node, coordinate_node* num_dof_per_node + 1) -= factor;
+        (*load_linearization)(
+            force_node* num_dof_per_node + 1, coordinate_node * num_dof_per_node) += factor;
```
dragos-ana, line 71: Maybe I'm missing something, but doesn't the current computation assume that the applied pressure is spatially homogeneous? Otherwise we would have a contribution from $\partial{p(\boldsymbol{x},t)} / \partial{\boldsymbol{x}}$, in addition to the current one from the changing normal $\partial{\boldsymbol{n}(\boldsymbol{x},t)} / \partial{\boldsymbol{x}}$? It's fine for me (if it's also included in the documentation of the function), but then it may not be entirely consistent to apply functions of space as normal pressure...

unittests/solid_ele/4C_solid_ele_neumann_evaluator_test.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "4C_solid_ele_neumann_evaluator.hpp"
+
+#include "4C_fem_general_element_integration.hpp"
+#include "4C_solid_ele_calc_lib_integration.hpp"
+
+namespace
+{
+  using namespace FourC;
+
+  TEST(PlaneNormalPressure, ReferenceThicknessAndConsistentLinearization)
```
dragos-ana, line 19: As far as I see, the unit test addresses specifically the "helper" function `add_normal_pressure_load`. 
It may be more practical to test the function that really evaluates the Neumann condition (`Discret::Elements::evaluate_normal_pressure_by_element`), but I understand that it is more tedious because one would have to construct an element, a discretization, and so on for such a test. 

I don't have a strong opinion on this and it's fine for me as it is - I just wanted to note that this current approach isolates a function that is ultimately an internal step of the Neumann condition evaluation.

dragos-ana (commented):
Thanks for implementing this! I added some comments / suggestions

src/solid_ele/4C_solid_ele_neumann_evaluator.hpp
ischeider: You're right, but the implementation is done as for the 3D case, where also no spatial pressure derivative is used. It would not be very much effort to implement it, but there would only be it would rather be for completeness. A dam with height-dependent water pressure might be an application, but nobody missed it so far. We might keep it in mind for a future implementation; I'll document it.

unittests/solid_ele/4C_solid_ele_neumann_evaluator_test.cpp
ischeider: Constructing an element inside a discretization is exactly what the input file does and we have an input file for testing the procedure as a whole. Thus, I hope it is sufficient as it is.

src/solid_ele/4C_solid_ele_neumann_evaluator.hpp
ischeider: I left it as it is...

src/solid_ele/4C_solid_ele_neumann_evaluator.cpp
ischeider: Actually, the strange thing is rather that entering arrays of more than 1 value is necessary at all. I would rather remove the check before on how many entries are given. Alternatively, one should rather enter NUMDOF=1 (if we don't remove NUMDOF completely, which makes even more sense) for a [pseudo_]orthopressure neumann load.
Removing the check makes the orthopressure much more intuitive, so I delete the first check and just check `onoff[0]==1`.

src/solid_ele/4C_solid_ele_neumann_evaluator.hpp
dragos-ana: That's fine. I still think that formulae would be nice here to better follow the implementation, but I leave that up to you.

src/solid_ele/4C_solid_ele_neumann_evaluator.cpp
dragos-ana: True, wouldn't it make sense to directly enforce that the size of onoff is 1?

PR #2105, opened 2026-06-30
Enable adaptive time stepping for contact and periodic rve boundary condtions 

m-frey:
## Description

This PR enables adaptive time stepping for contact and the constraint framework by implementing the missing `reset_step_state()` functionality.

### Changes
- Implement `reset_step_state()` for the contact model evaluator.
- Implement `reset_step_state()` for the constraint framework.
- Restore the last converged state before retrying a failed adaptive step.

Issue #64, opened 2024-11-29
Wrong log output during passive time step adaption


davidrudlstorfer:
_Ported from Gitlab. Original issue [here](https://gitlab.lrz.de/baci/baci/-/issues/1216)_

## Bug Description: What's going wrong?

After further investigating https://gitlab.lrz.de/baci/baci/-/issues/1212 and https://gitlab.lrz.de/baci/baci/-/merge_requests/2113 the overall time step adaption (halve time step size when no convergence is reached and double time step size once 4 subsequent steps converged) seems to be bug free.

Unfortunately, the log output from BACI/4C is wrong once a time step is halved or doubled. One exemplary input file is:

[4_beams_input.dat.txt](https://github.com/user-attachments/files/17960194/4_beams_input.dat.txt)
[collagen_solver_settings.xml.txt](https://github.com/user-attachments/files/17960195/collagen_solver_settings.xml.txt)

which produces the following log output:

[log.txt](https://github.com/user-attachments/files/17960203/log.txt)

with the exemplary section:

<details><summary>Click to expand</summary>

```
 #It      ||F||_2         step    ||dx||_2
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   0  6.77186e-02  0.00000e+00  0.00000e+00
   1  2.38705e-03  1.00000e+00  1.89177e-02
   2  1.24825e-05  1.00000e+00  1.47900e-04
   3  3.38640e-10  1.00000e+00  4.25099e-07 (Converged!)
************************************************************************
-- Final Status Test Results --
Converged....OR Combination -> 
  Converged....AND Combination -> 
    Converged....Structure-F-Norm = 3.386e-10 < 1.000e-05
                 (Unscaled 2-Norm, Absolute)
    Converged....Structure-Update-Norm = 4.251e-07 < 1e-05
                 (Unscaled 2-Norm, Absolute)
                 (Min Step Size:  1.000e+00 >= 1)
                 (Max Lin Solv Tol:  0.000e+00 < 0.5)
  **...........Number of Iterations = 3 < 5
************************************************************************
Nonlinear solver successful. Double timestep size!
PID  0 currently monitors   624 beam potential pairs


--- Monitor Dirichlet boundary condition 101 
          step          time      ref_area     curr_area           f_x           f_y           f_z           m_x           m_y          m_z
[...]

PR #2211, opened 2026-08-24
Tighten solid `DIVERCONT` tests

rjoussen:
## Description and Context

Preparation of #2216 

This PR is just some preparatory work to tighten the existing `DIVERCONT` test cases:
 
- `DIVERCONT != stop` is removed from all tests (and the preconditioner tutorial) that do not need it because they never encounter non-convergence.
- Add last `time_step_size`, `total_time` and`total_num_steps` to result quantities for testing. Appropriate tests are added to all tests that use non-constant time-step sizes. (might be interesting for #2105 as well that still needs this result check)
- Some tested occurrences of `adapt_step` are temporarily (until `DIVERCONT` is refactored) replaced by `halve_step`. This is mainly because the currently possible time-step size recovery (that is enabled with this option) has two problems:
    1. It only actually increases the timestep size if the remaining number of steps is even.
    2. It increases the time-step too early, i.e. before update and output and the step summary is written to screen. It does so even in the very last step, which would mean that the `time_step_size` result tests would need to test for a time-step size that never happened, if time-step recovery happens to take place in the last step.
  
  Due to this change, the brownian dynamic test that used `adapt_step` has a small result drift, since stochastic forces are sensitive to the size and number of time-steps.

@isteinbrecher and @davidrudlstorfer for the changed beam test, especially the brownian dynamic one with changed result values.

codex only showed me where the special result quantities are registered.


src/structure_new/src/utils/4C_structure_new_resulttest.cpp
davidrudlstorfer, line 1: Nice that we can test this now!

tests/input_files/beam3r_line5_static_test1_adaptstep.4C.yaml
davidrudlstorfer, line 1: Why is here the number of steps changed?

tests/input_files/beam3sk_line3rot_static_test1_adaptstep.4C.yaml
davidrudlstorfer, line 1: Same here?

tests/input_files/beam3wk_line3rot_static_test1_adaptstep.4C.yaml
davidrudlstorfer, line 1: And here

tests/input_files/beam3r_herm2line3_backweuler_browndyn_vanderWaals_doublelengthspec_smallsepapprox_twocrossedbeams.4C.yaml
davidrudlstorfer, line 1: I am not 100% sure that I can follow your argumentation here.

Why do you need to change this test to halve_step? Cannot we just keep the old option? Once this is fixed the test results can be updated anyways, why now?

davidrudlstorfer (commented):
Thanks @rjoussen for the pull request and for working on this!

I did not have the time to further write in #2189 but I greatly appreciate your work on this topic.

A few points from my side:

1. I am using the adapt_step option in all of my simulations. Background is that I have a lot of snap-through actions happening which only converge with a very small time step. It is unfortunately not possible to predict when these snap-through dynamics occur. So the time step is just halved during these high speed maneuvers and then increased again. Therefore, I need this option on a daily basis

2. I did not yet follow all details of all of your PRs and discussion so I need to ask: is this closely related to #64? I never worked on that again but I think this is closely related do what you are doing

3. The remarks to this PR: I really like the new introduction of test quantities, these seem super helpful. Unfortunately, even with your description I do not understand why the one beam test needs to be changed from adapt to halve step? The other tests also still keep it? (Just FYI, all the tests including Brownian Dyn, LJ or vanderWaals are in my area of expertise) @isteinbrecher needs to have a look on Nurbs and regular beams

All in all, really nice progress but I do not fully understand a few details

rjoussen:
@davidrudlstorfer I hope #2216 addresses some of your general remarks

tests/input_files/beam3r_line5_static_test1_adaptstep.4C.yaml
rjoussen: As mentioned in #2216, the current adapt_step logic internally changes the endstep. When I refactor the adapt step behaviour as proposed, this wont happen anymore, so we must ensure here that there are enough steps to actually reach the max time specified at 1. I thought I change this now already, because I want to avoid changing the tests too much when I actually introduce behaviour changes at the same time.

tests/input_files/beam3r_herm2line3_backweuler_browndyn_vanderWaals_doublelengthspec_smallsepapprox_twocrossedbeams.4C.yaml
rjoussen: For the NUMSTEP change see my other comment. 

I change it to halve step, because my planned version of `adapt_step` will increase the time-step reliably after a certain number of steps. This test is rather long and makes use of the (anti-)feature that the time-step increases sometimes after 4 and sometimes after 5 steps. My planned version of adapt_step will not be able to reproduce this exactly anymore, but it will be able to reproduce the halve_step behaviour. That's why I wanted to change it now already, then change it again to adapt_step later but with settings in place that mimic the current halve_step behaviour. I could leave it at adapt_step, but then the result changes that are necessary now will be necessary when I do the bigger refactor.

tests/input_files/beam3wk_line3rot_static_test1_adaptstep.4C.yaml
rjoussen: Here the reason to switch (temporarily) to halve_step is actually a little different: This case never actually increases the time-step size, it instead runs 4 steps at dt=0.25. After this 4th step, it runs into the problem that it incrreases to 0.5, then the simulation ends, which means the result test only sees the last 0.5, which I dont want to test for. After the refactor, it will be changed back to adapt_step.

tests/input_files/beam3sk_line3rot_static_test1_adaptstep.4C.yaml
rjoussen: see my other comment

tests/input_files/beam3r_line5_static_test1_adaptstep.4C.yaml
rjoussen: The reason why I dont switch to halve_step here is that this test has the following time-step history (if I remember correctly): 6 steps tota, first decrease twice, then 0.125, 0.125, 0.125, 0.125, recover after exactly 4 steps, then 0.5, 0.5. This behaviour will be exactly reproducible by the refactor.

isteinbrecher: > 6 steps tota, first decrease twice, then 0.125, 0.125, 0.125, 0.125

I think that is correct, I know this test quite well 😂

Looking at this, should we maybe allow `NUMSTEP` to be empty in the future and only specify this in the `TIME STEP CONTROL` section?

isteinbrecher (dismissed):
Thanks for looking into this @rjoussen.

Just one small question, are there still some `continue` tests left that would fail otherwise?

rjoussen:
> Just one small question, are there still some continue tests left that would fail otherwise?

I will open a seperate issue listing them

tests/input_files/beam3r_herm2line3_backweuler_browndyn_vanderWaals_doublelengthspec_smallsepapprox_twocrossedbeams.4C.yaml
davidrudlstorfer: I would honestly do the changes later, because then we can really compare how large the deviations are. I would be in favor of that approach rather than changing it twice

tests/input_files/beam3r_line5_static_test1_adaptstep.4C.yaml
rjoussen: We could, but removing `NUMSTEP` as a main input parameter is a bigger refactor. It currently defaults to 200 for some reason. I also think it has equal "rights" to be in the main STRUCTURE section, since people might want to control their simulation by steps rather than physical time (e.g., if computation wall time is a concern). I think the cleanest would be to allow both, but also accept only one and set the other two infinity.

tests/input_files/beam3r_herm2line3_backweuler_browndyn_vanderWaals_doublelengthspec_smallsepapprox_twocrossedbeams.4C.yaml
rjoussen: okay, I will keep it unchanged here and then do result changes together with the refactor.

davidrudlstorfer: Thank you 🙏🏻

tests/input_files/beam3wk_line3rot_static_test1_adaptstep.4C.yaml
rjoussen: ~Here I would like to keep halve_step in temporarily though, otherwise I have to change the value for last time-step-size test later.~
Actually I will keep it unchanged here as well.

davidrudlstorfer (approved):
Thanks for the changes :)

isteinbrecher (approved):
> I will open a seperate issue listing them

Thanks! I agree that we should avoid this in standard tests and aim for one (or a few) test that actually requires this feature.

PR #2214, opened 2026-08-24
Remove `ele_eval_error_flag`

rjoussen:
## Description and Context

Preparation for #2216 

This proposes to remove the existing `ele_eval_error_flag`. I believe that it was originally introduced to _somehow_ communicate local failure on material level upwards, but it is a bit of a strange design and does not work in its current form:
1. Materials rely on the possibility that the `Solid::Elements::ParamsInterface` pointer is passed to them via the `Teuchos::ParameterList` (which only ever happens in pure solid problems).
2. They can use this pointer to set the mentioned flag, which then hopefully will be read by the structure model evaluator (only if the model evaluator is the current evaluation path, which might not be the case for coupled problems.)
3. Even if the flag is set and read, the model evaluator will return false, which results in an untyped exception in nox, so no action is taken from it. The more dangerous case is if the flag is set but not processed correctly.

On top of that, the only material that sets this flag is `couptransverselyisotropic`, at a place where I believe it can only be reached due to a programming error, because it asserts that the determinant of the right Cauchy Green tensor is positive. 

 

none

### DIFF
--- a/src/mat/elast/4C_mat_elast_couptransverselyisotropic.cpp
+++ b/src/mat/elast/4C_mat_elast_couptransverselyisotropic.cpp
@@ -15,7 +15,6 @@
 #include "4C_mat_elast_aniso_structuraltensor_strategy.hpp"
 #include "4C_mat_service.hpp"
 #include "4C_material_parameter_base.hpp"
-#include "4C_structure_new_elements_paramsinterface.hpp"
 
 #include <Teuchos_ParameterList.hpp>
 
@@ -176,7 +175,7 @@ void Mat::Elastic::CoupTransverselyIsotropic::add_stress_aniso_principal(
     const int gp, const int eleGID)
 {
   // direct return if an error occurred
-  if (reset_invariants(rcg, &params)) return;
+  reset_invariants(rcg);
 
   // switch to stress notation
   Core::LinAlg::SymmetricTensor<double, 3, 3> rcg_inv_s;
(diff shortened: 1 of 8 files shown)


src/mat/elast/4C_mat_elast_couptransverselyisotropic.cpp
```diff
@@ -176,7 +175,7 @@ void Mat::Elastic::CoupTransverselyIsotropic::add_stress_aniso_principal(
     const int gp, const int eleGID)
 {
   // direct return if an error occurred
-  if (reset_invariants(rcg, &params)) return;
+  reset_invariants(rcg);
```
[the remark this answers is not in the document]
rjoussen: It is still used by other implementations of `add_stress_aniso_principal`

src/mat/elast/4C_mat_elast_couptransverselyisotropic.cpp
```diff
@@ -249,19 +248,14 @@ void Mat::Elastic::CoupTransverselyIsotropic::update_elasticity_tensor(
...
   const double I3 = Core::LinAlg::det(rcg);
-  if (I3 < 0.0)
-  {
-    std::stringstream msg;
-    msg << __LINE__ << " -- " << __PRETTY_FUNCTION__ << "I3 is negative!";
-    error_handling(params, msg);
-    return -1;
-  }
+  FOUR_C_ASSERT_ALWAYS(
+      I3 >= 0.0, "Determinant of right Cauchy-Green tensor is negative (value: {}).", I3);
```
davidrudlstorfer: Nice that you updated this :)

davidrudlstorfer (dismissed):
LGTM 🚀

src/mat/elast/4C_mat_elast_couptransverselyisotropic.cpp
c-p-schmidt: Shouldn't we also exclude a 0 determinant?
```suggestion
      I3 > 0.0, "Determinant of right Cauchy-Green tensor is not positive (value: {}).", I3);
```

c-p-schmidt (dismissed):

src/mat/elast/4C_mat_elast_couptransverselyisotropic.cpp
rjoussen: fair

davidrudlstorfer (approved):

c-p-schmidt (approved):

PR #2215, opened 2026-08-24
Improve solid time-step status handling

rjoussen:
Preparation for #2216 

This PR improves how statuses/return codes are currently handled in the solid time integration. There are no behavioral changes (I hope!), except that some errors that may have previously gone undetected now fail.

The changes are:
- Change the return type of `integrate()` and `integrate_step()` from `int` to `void`. Their `int` return codes were either unused (i.e., unconditionally 0) or (worse) ignored by callers.
- The base implementation of `perform_error_action()` (which only asserts success) is moved to `adapter_str_structure`. This prepares it to be called directly from the timeloop (for both old and new time integration), which is where it belongs, in my opinion. Its current place inside solve is very dangerous for partitioned solvers that call solve: Perform error action might change the timestep and reset a step, for which the coupled partitioned solver has to be responsible (because they have synchronized it with the other field).
- The return enum of the `solve()` method, `ConvergenceStatus`, is renamed to `StepStatus`, such that it can be used for other routines in the timeloop as well that can alter the current solid time step status (I mainly want to use it for the predictor).  It is also converted to an `enum class` (from a pure `enum`) and moved to the `adapter_str_stucture` header. (t used to be in input?). Finally, the enum field names are renamed to be a little more precise
- `Structure::solve()` is marked as `[[nodiscard]]` to enforce that callers are actually handling the returned status. Many (maybe all) partitioned solvers ignored the return of `solve()` so far, which is potentially very dangerous. All of them assert success now.

AI mainly reviewed my code or did very scoped changes to single files/functions.


src/structure/4C_structure_timint.cpp
```diff
@@ -2445,9 +2442,8 @@ Solid::ConvergenceStatus Solid::TimInt::perform_error_action(Solid::ConvergenceS
       // reset step (e.g. quantities on element level)
       reset_step();
 
-      return Solid::conv_success;
+      return Solid::StepStatus::no_errors;
```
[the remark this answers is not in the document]
rjoussen: The old return status was inconsistent already. In the layout of the old time integration it actually doesnt matter, perform error action is only called _instead_ and not before/after calling update/output, and any status no_errors/fail_repeat will allow the timeloop to continue.

I am still changing it here already, as suggested. Note: It does not happen in line 2485.

src/adapter/4C_adapter_str_structure.hpp
davidrudlstorfer, line 1: Stupid question; is this file meaning "4C_adapter_structure_structure"? Or is the str meaning something else? If yes, would it make sense to rename this to something more clear?

src/porofluid_pressure_based_elast/4C_porofluid_pressure_based_elast_partitioned.cpp
```diff
@@ -297,7 +297,9 @@ void PoroPressureBased::PorofluidElastPartitionedAlgorithm::do_struct_step()
   }
 
   // Newton-Raphson iteration
-  structure_algo()->solve();
+  const auto structure_status = structure_algo()->solve();
+  FOUR_C_ASSERT_ALWAYS(
+      structure_status == Solid::StepStatus::no_errors, "Structural solve failed.");
```
davidrudlstorfer, line 302: This is the change to a lot of files, would it not be possible to directly assert this in the solve call?

src/structure/4C_structure_timint_centrdiff.cpp
```diff
@@ -232,9 +232,6 @@ int Solid::TimIntCentrDiff::integrate_step()
 
   // update of end-velocities \f$V_{n+1}\f$
   veln_->update(dthalf, *accn_, 1.0);
-
-  // wassup?
```
davidrudlstorfer, line 236: Amazing comment is going haha

davidrudlstorfer (commented):
LGTM, just a few minor details/questions

src/adapter/4C_adapter_str_structure.hpp
rjoussen: That is a really tough question, actually. I believe the logic is roughly like this: `adapter_<field that is wrapped>_<field this adapter is used for>`. So this would be the wrapper for the structure (solid) field for use in the structure algorithms, whereas `adapter_str_structure_new` is for use specifically in the new time integration. But then `adapter_str_wrapper` wraps that whole adapter object again.  But idk, it's really hard to understand. I wouldn't rename it, to be honest. I think the goal should be to get rid of these global adapters. I don't really see how they provide any benefit in this global adapter folder.

src/porofluid_pressure_based_elast/4C_porofluid_pressure_based_elast_partitioned.cpp
rjoussen: No, since `solve()` legitimately can return a non-successful status. And in fact, if you want to retry the time-step based on that status, it must return a status. It was returning a status before already, but it was discarded in most places. There could be a solve wrapping function that is void, asserting success of solve internally, but I don't really see the advantage of that (there is already integrate_step in the new time integration only, which kind of does that). Also, maybe in the future, partitioned solvers want to make use of the status and restart both fields in a synchronized way.

src/structure/4C_structure_timint_centrdiff.cpp
rjoussen: Introduced by @poppmuc 14 years ago ;)

src/adapter/4C_adapter_str_structure.hpp
davidrudlstorfer: Sounds good:)

davidrudlstorfer (approved):

lauraengelhardt (approved):

Issue #2216, opened 2026-08-25
Refactor solid time-step-adaptivity and `DIVERCONT` behavior


rjoussen:
## Description

>Not to be confused with the time-step adaptivity through `STRUCTURAL DYNAMIC/TIMEADAPTIVITY`, which uses an a priori error estimation and performs its own time-step adaptation based on that.

As discussed in part in #2189, the only way to trigger time-step size adaptivity is to use the `DIVERCONT` parameter. The current behavior, though, is suboptimal and could have more functionality/flexibility. It also would be nice to have more reasons to adapt the time-step size than just `DIVERCONT`.

Here is a summary of what the options of `DIVERCONT` currently are, what behavior they have (in the new integration only), and what my plan is with them:

`DIVERCONT` option | behavior |Plan
-|-|-
`stop` (default)  | stops simulation, potentially writing a final output. | no change
`continue`| continues simulation without any changes, ignoring divergence | I think test coverage should be heavily reduced. I will probably rename it to `ignore` to be able to use it in an input enum directly (continue is a keyword).
`repeat_step` | resets the step and repeats it without any changes.| remove
`halve_step` | halves step size. Also adapts `NUMSTEP`/`step_end` parameter. | remove in favor of `adapt_step`
`adapt_step` | same as `halve_step`, but also checks for possible increase (see below) | This is widely used and will stay, but how the timestep size is changed will be controlled by the new `TIME STEP CONTROL` input group (see below)
`rand_adapt_step` | first computes an average random number across processors (???) and then either increases or decreases the timestep, depending on what it did last time.  | remove
`rand_adapt_step_ele_err` | exactly the same behavior as `rand_adapt_step` | remove
`repeat_simulation` | same as `stop`, but doesn't throw immediately. Instead, delegates abortion to the timeloop.  | remove
`adapt_penaltycontact` | not implemented error | remove since not implemented.

> Remark on old-time integration: Mostly similar behavior, except:
> - `continue` will actually lead to  `stop`, but without final output writing.
> - `adapt_penaltycontact` is implemented, but it does not seem to reset the step and is untested, so I don't know if it works as expected. If it works it alters the penalty parameter by a random factor between 0.95 and 1.

#### If the solver did not converge:
- After non-convergence, it halves the step size. How often it does that can be controlled by `MAXDIVCONREFINEMENTLEVEL`
[...]

PR #2217, opened 2026-08-25
Remove old integration eval-error-flag

rjoussen:
## Description and Context

Preparation for #2216

In #2214, I missed that there is also a similar mechanism in the old time integration. I am removing it because this will make it easier to remove some of the divercon enum options.

none


c-p-schmidt (approved):
Looks good to me.

bgoderbauer (approved):

Issue #2218, opened 2026-08-25
`DIVERCONT: continue` usage in test files


rjoussen:
The following test files use the solid version of `DIVERCONT: continue` (maybe `ignore` in the future) and actually need this parameter, i.e. their nonlinear solvers do not converge. All of them except the beam test also use `MAXITER: 1`, so hopefully non-convergence can be attributed to that.

- sohex8_easfull_patchtest.4C.yaml
- sohex8_disp_patchtest_varytimecurve.4C.yaml
- meshtying2D_patch_linstatic_2_new_struct.4C.yaml
- contact2D_patch_linstatic.4C.yaml
- w1_w1nurbs.4C.yaml
- meshtying2D_inf_plate.4C.yaml
- ~beam3r_herm2line3_beam_to_beam_point_coupling_elbow_offset_indirect_global_assembly_lagrange_projection_on_nodes.4C.yaml~
- sotet10_patchtest.4C.yaml
- sohex8_disp_patchtest.4C.yaml
- meshtying2D_patch_linstatic.4C.yaml
- contact2D_hertz_linstatic_new_struct.4C.yaml
- sohex8_incompr_block.4C.yaml
- sotet4_patchtest.4C.yaml
- sohex8_disp_patchtest_varytimecurve_new_struc.4C.yaml
- contact2D_patch_bound_new_struct.4C.yaml
- meshtying2D_patch_linstatic_2.4C.yaml
- contact3D_patch_linstatic_new_struct.4C.yaml
- contact2D_patch_bound.4C.yaml
- contact2D_hertz_linstatic.4C.yaml
- meshtying2D_inf_plate_new_struct.4C.yaml
- sosh8_lin_bending_patchtest.4C.yaml
- contact3D_patch_linstatic.4C.yaml
- contact2D_patch_linstatic_new_struct.4C.json

isteinbrecher:
Thanks for the list @rjoussen !

The test `beam3r_herm2line3_beam_to_beam_point_coupling_elbow_offset_indirect_global_assembly_lagrange_projection_on_nodes.4C.yaml` should not need `DIVERCONT: continue`. I will fix this.

PR #2219, opened 2026-08-25
Fix beam test time step parameters

isteinbrecher:
Remove a not needed parameter from a beam test.

#2218


lauraengelhardt (approved):

rjoussen (approved):
I probably missed it when trying to remove continue everywhere because it doesn't use quotation marks around continue.

Issue #2220, opened 2026-08-25
Add vtk tests for element material quantities


lkoeglmeier:
## Description
There is the possibility to output element material quantities stored in the material via `vis_names`/`vis_data` which is currently not tested.
The material files that list/store such data are listed below:

- [ ] 4C_mat_aaaneohooke.cpp
- [ ] 4C_mat_constraintmixture.cpp
- [ ] 4C_mat_crystal_plasticity.cpp
- [ ] 4C_mat_damage.cpp
- [ ] 4C_mat_elasthyper.cpp
- [ ] 4C_mat_growthremodel_elasthyper.cpp
- [ ] 4C_mat_maxwell_0d_acinus_Ogden.cpp
- [ ] 4C_mat_membrane_active_strain.cpp
- [ ] 4C_mat_plastic_VarConstUpdate.cpp
- [ ] 4C_mat_plasticelasthyper.cpp
- [ ] 4C_mat_plasticlinelast.cpp
- [ ] 4C_mat_plasticnlnlogneohooke.cpp
- [ ] 4C_mat_scalardepinterp.cpp
- [ ] 4C_mat_structporo_reaction.cpp
- [ ] 4C_mat_structporo_reaction_ecm.cpp
- [ ] 4C_mat_superelastic_sma.cpp
- [ ] 4C_mat_thermoplastichyperelast.cpp
- [ ] 4C_mat_viscoanisotropic.cpp
- [ ] 4C_mat_elast_remodelfiber.cpp
- [ ] 4C_membrane_evaluate.cpp

## Possible Solution and Definition of Done

- [ ]  Add vtk tests for all of these output quantities. 
- [ ] Unify functions `vis_names`/`vis_data`  and `register_output_data_names`/`evaluate_output_data` if possible or rename them to clearly distinguish their purposes.

ischeider:
Interesting topic! Maybe it's worth another issue, but let's post it here first: It would be really great to have an auto-generated list of the material output quantities for the reference documentation. Since we have the toll create_rtd, one could simply add this extraction there.

PR #2221, opened 2026-08-25
Add output of material element data and Nodeowner

lkoeglmeier:
Adds the possibility to output `Nodeowner` and element material quantities stored in the material via `vis_names`/`vis_data` with the new vtk-based output framework.

One vtk test for `porosity` in the StructPoro material and `Nodeowner` is added.

Issue  #2220 will list all materials that store such data, so that additional tests can be added here as a follow-up task.

#2190

AI was used for a first implementation and review.  All AI-assisted code has been reviewed and validated by the author, who takes responsibility for the contribution.


src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
```diff
@@ -435,6 +436,76 @@ namespace Core::IO
...
+    }
+
+    append_result_data_vector_with_context(sysdata, Core::IO::OutputEntity::node, {"Nodeowner"});
+  }
+
+  /*-----------------------------------------------------------------------------------------------*
+   *-----------------------------------------------------------------------------------------------*/
+  void DiscretizationVisualizationWriterMesh::append_material_element_data()
+  {
+    std::map<std::string, int> names;  // contains name and dimension of data
```
c-p-schmidt: Maybe this would be a more suitable name?
```suggestion
    std::map<std::string, int> names_and_size;  // contains name and size of data
```

src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
```diff
@@ -435,6 +436,76 @@ namespace Core::IO
...
+      // get names and dimensions from every element
+      ele.user_element()->vis_names(names);
+    }
+
+    // By applying gather_all we get the combined map including all elemental values
+    // which were found by vis_names
+    Core::LinAlg::gather_all(names, discretization_->get_comm());
+
+    FOUR_C_ASSERT_ALWAYS(
+        std::all_of(names.begin(), names.end(), [](const auto& pair) { return pair.second >= 1; }),
```
c-p-schmidt: You could also use the std::ranges version.
```suggestion
        std::ranges::all_of(names, [](const auto& pair) { return pair.second >= 1; }),
```

src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
```diff
@@ -435,6 +436,76 @@ namespace Core::IO
...
+
+      // MultiVector stuff from the elements is put in
+      auto sysdata =
+          Core::LinAlg::MultiVector<double>(*discretization_->element_row_map(), dimension, true);
+
+      for (auto ele : discretization_->my_row_element_range())
+      {
+        if (!element_filter_(ele.user_element())) continue;
+
+        std::fill(eledata.begin(), eledata.end(), 0.0);
```
c-p-schmidt: Also here you could use the std::ranges version.
```suggestion
        std::ranges::fill(eledata, 0.0);
```

src/core/io/src/4C_io_discretization_visualization_writer_mesh.cpp
```diff
@@ -435,6 +436,76 @@ namespace Core::IO
...
+
+        // get data for a given name from element & put in sysdata
+        ele.user_element()->vis_data(name, eledata);
+
+        const int lid = discretization_->element_row_map()->lid(ele.user_element()->id());
+        FOUR_C_ASSERT_ALWAYS(lid >= 0, "Internal error: element not found in element_row_map()");
+
+        FOUR_C_ASSERT_ALWAYS(
+            (int)eledata.size() == dimension, "element manipulated size of visualization data");
```
c-p-schmidt: Here, you could avoid the C-style cast.
```suggestion
            static_cast<int>(eledata.size()) == dimension, "element manipulated size of visualization data");
```

c-p-schmidt (dismissed):
Looks good to me, only some minor commands.

Thank you also for documenting the missing output checks in the issue you opened.

maxiludwig (approved):

c-p-schmidt (approved):

rjoussen:
I think this is at least somewhat related to #1754, right? I think `MATERIAL_ELEMENT_QUANTITIES` could maybe document a bit more what kind of material element quantities it turns on, because it doesnt turn on gauss point quantities for instance (if I am not mistaken)

c-p-schmidt:
> I think this is at least somewhat related to #1754, right? I think `MATERIAL_ELEMENT_QUANTITIES` could maybe document a bit more what kind of material element quantities it turns on, because it doesnt turn on gauss point quantities for instance (if I am not mistaken)

If this is desired, a completely different approach is necessary, as it depends on the material being activated, which properties are written to output. I don't see that a top-level flag will reasonably document that. If we would like to have such a feature, I think the option should be split in follow-up work, and activation should then maybe move to the material itself, as only there is it clear which parameters can be activated for output.

rjoussen:
Yes, definitely, I didn't mean that it should block this PR in any way. I mainly wanted to link it such that it shows up in the other discussion :)

Issue #2208, opened 2026-08-22
[beaminteraction] Enable `Beam3rLine2Line2` for beam-solid volume coupling scenarios


maxfirmbach:
## Description

Currently only the `Beam3rHerm2Line3` Simo-Reissner beam formulation is enabled for beam-solid volume coupling scenarios (?). It would be nice to enable the formulation based on `Beam3rLine2Line2` as it features the same amount of DOFs per node, which for some scenarios enables more efficient solution techniques, e.g. algebraic multigrid might work out-of-the-box for such beam elements.

For lattice structures with beam-beam point coupling constraints, the approximation quality of `Beam3rLine2Line2` is usually sufficient, especially in the case of straight fibers.

## Possible Solution and Definition of Done

- [x] Extend the templating in beaminteraction for shape functions used by `Beam3rLine2Line2`.
- [x] Do testing of the respective functionality.
- [ ] Make nullspace work properly.

## Interested Parties

@isteinbrecher @mayrmt

isteinbrecher:
This should be possible straight forwardly, by simply enabling the templates. However, some testing is still needed to ensure that this works as expected.

One more thing, while we could enable all beams with all type of couplings, this will dramatically increase compile time. So we should find a good common ground of pairs that are actually used in simulations. In the good old days we could have managed this with some compile time flags, but I don't know the modern approach for this.

mayrmt:
We could still manage this with compile time flags, though they need to be configurable through CMake.

PR #2222, opened 2026-08-26
[beaminteraction] Template beam shape function type for beam-solid volume condition

maxfirmbach:
## Description and Context

This if a first step to make #2208 work. In this PR the beam shape function type is templated such that different types than `t_hermite` can be used in the near future for centerline displacement interpolation.

The current implementation only considers the mortar-based coupling approach.

#2208

### DIFF
--- a/src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_conditions.cpp
+++ b/src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_conditions.cpp
@@ -428,6 +428,10 @@ BeamInteraction::BeamToSolidConditionVolumeMeshtying::create_contact_pair_intern
     const std::vector<Core::Elements::Element const*>& ele_ptrs)
 {
   const auto* beam_element = dynamic_cast<const Discret::Elements::Beam3Base*>(ele_ptrs[0]);
+
+  FOUR_C_ASSERT_ALWAYS(
+      beam_element != nullptr, "Could not cast ele_ptrs[0] to Discret::Elements::Beam3Base.");
+
   const Core::FE::CellType shape_beam = beam_element->shape();
   const bool beam_is_hermite = beam_element->hermite_centerline_interpolation();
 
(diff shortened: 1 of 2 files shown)


src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_conditions.hpp
```diff
@@ -341,7 +344,7 @@ namespace BeamInteraction
   template <template <typename...> class BtsClass, typename... BtsMortarTemplateArguments,
       typename... BtsMortarShape>
   std::shared_ptr<BeamInteraction::BeamContactPair> create_beam_to_solid_volume_pair_mortar(
-      const Core::FE::CellType shape,
+      const Core::FE::CellType shape, const bool beam_is_hermite,
```
isteinbrecher: We will also need a `shape_beam` here (the shape of the centreline). I think a suitable interface would be
```c++
const Core::FE::CellType shape_solid, const Core::FE::CellType shape_beam, const bool beam_is_hermite, ...
```

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_conditions.cpp
```diff
@@ -389,9 +390,19 @@ BeamInteraction::create_beam_to_solid_volume_pair_mortar(const Core::FE::CellTyp
  */
 template <template <typename...> class BtsClass, typename... BtsMortarTemplateArguments>
 std::shared_ptr<BeamInteraction::BeamContactPair>
-BeamInteraction::create_beam_to_solid_volume_pair_mortar(const Core::FE::CellType shape)
+BeamInteraction::create_beam_to_solid_volume_pair_mortar(
+    const Core::FE::CellType shape, const bool beam_is_hermite)
 {
-  return create_beam_to_solid_volume_pair_shape<BtsClass, BtsMortarTemplateArguments...>(shape);
+  if (beam_is_hermite)
```
isteinbrecher, line 399:
```suggestion
  if (shape_beam == line2 && beam_is_hermite)
```

maxfirmbach: My initial plan was to remove the error message and place the following piece of code there:
```cpp
    return create_beam_to_solid_volume_pair_shape<GeometryPair::t_line2, BtsClass, BtsMortarTemplateArguments...>(shape);
```

src/beaminteraction/src/contact/beam_to_solid/4C_beaminteraction_contact_beam_to_solid_conditions.hpp
maxfirmbach: I added `shape_beam` to all relevant functions and introduced a switch-case for the Lagrange centerline interpolation cases. Currently, all of them throw, as the templates are not initialized with the respective cell types yet.

isteinbrecher (dismissed):

dharinib98 (dismissed):

maxfirmbach:
I added another assert after casting one member of the geometry pair.

isteinbrecher (approved):

mayrmt (approved):

PR #2223, opened 2026-08-26
[particle] Do not use ParticleObjShrdPtr for ghost particles data transfer

vovannikov:
## Description and Context

This is continuation of the previous optimizations and #2117 in particular. The current implementation already follows the pattern established in that PR after the comments of the others. This PR:

- introduces `ParticleGhostEntry` (mirrors `ParticleRefreshEntry`) to pack/unpack `[type][globalid][bingid][ownedindex][states]` directly into send buffers, bypassing `ParticleObject` creation when creating ghost particles.
- Cache `cached_procs_send_ghost_particles_to_` / `cached_procs_receive_ghost_particles_from_` from the bin ghosting topology in `determine_ghosting_dependent_maps_and_sets()`, avoiding an `MPI_Allreduce` in `ghost_particles()`.
- Reuse `ParticleUtils::immediate_send_recv_known_procs()` for both the new ghost particle exchange (`communicate_and_insert_ghost_particles()`) and the direct ghosting map handshake (`communicate_direct_ghosting_map()`), instead of duplicating MPI `Isend`/`Irecv` logic.

This optimization gets more efficient as the number of ranks grows (not surprising).
image

#2117, #2145, #2074

Claude Sonnet 5 was used to "smart" rebase the modifications I had originally prepared earlier 2 months ago https://github.com/4C-multiphysics/4C/compare/main...vovannikov:4C:particle_remove_particleobj_ghosts_orig onto the actual `main` in order to accommodate immediately the code structure suggested in #2117.

### DIFF
--- a/src/particle/src/engine/4C_particle_engine.cpp
+++ b/src/particle/src/engine/4C_particle_engine.cpp
@@ -332,8 +332,6 @@ void Particle::ParticleEngine::ghost_particles()
 {
   TEUCHOS_FUNC_TIME_MONITOR("Particle::ParticleEngine::GhostParticles");
 
-  std::map<int, std::map<Particle::Type, std::map<int, std::pair<int, int>>>> directghosting;
-
   // clear all containers of ghosted particles
   particlecontainerbundle_->clear_all_containers_of_specific_status(Status::Ghosted);
 
@@ -341,7 +339,7 @@ void Particle::ParticleEngine::ghost_particles()
   const auto sdata = pack_particles_to_be_ghosted();
 
   // communicate ghost data using cached comm graph and insert into containers (no Allreduce)
-  communicate_and_insert_ghost_particles(sdata, directghosting);
+  const auto directghosting = communicate_and_insert_ghost_particles(sdata);
 
   // communicate and build map for direct ghosting
   communicate_direct_ghosting_map(directghosting);
@@ -1920,10 +1918,12 @@ void Particle::ParticleEngine::insert_owned_particles(
   invalidate_particle_safety_flags();
 }
 
-void Particle::ParticleEngine::communicate_and_insert_ghost_particles(
-    const std::map<int, std::vector<char>>& sdata,
-    std::map<int, std::map<Particle::Type, std::map<int, std::pair<int, int>>>>& directghosting)
+std::map<int, std::map<Particle::Type, std::map<int, std::pair<int, int>>>>
+Particle::ParticleEngine::communicate_and_insert_ghost_particles(
+    const std::map<int, std::vector<char>>& sdata)
 {
+  std::map<int, std::map<Particle::Type, std::map<int, std::pair<int, int>>>> directghosting;
+
   // communicate ghost particle data using the cached comm graph derived from bin ghosting
   // topology (no MPI_Allreduce)
   const std::map<int, std::vector<char>> rdata =
@@ -1950,6 +1950,8 @@ void Particle::ParticleEngine::communicate_and_insert_ghost_particles(
       container->add_particle(ghostedindex, entry.globalid, entry.states);
 
       // add index relating (owned and ghosted) particles to col bins
+      FOUR_C_ASSERT_ALWAYS(entry.bingid >= 0,
+          "received ghosted particle contains no information about its bin gid!");
       particlestobins_[binning_->bincolmap_->lid(entry.bingid)].push_back(
           std::make_pair(entry.type, ghostedindex));
 
@@ -1966,6 +1968,8 @@ void Particle::ParticleEngine::communicate_and_insert_ghost_particles(
   validparticleneighbors_ = false;
   validglobalidtolocalindex_ = false;
   validdirectghosting_ = false;
+
+  return directghosting;
 }
 
 void Particle::ParticleEngine::remove_particles_from_containers(
--- a/src/particle/src/engine/4C_particle_engine.cpp
+++ b/src/particle/src/engine/4C_particle_engine.cpp
@@ -20,6 +20,7 @@
 #include "4C_particle_engine_communication_utils.hpp"
 #include "4C_particle_engine_container.hpp"
 #include "4C_particle_engine_container_bundle.hpp"
+#include "4C_particle_engine_ghost_entry.hpp"
 #include "4C_particle_engine_object.hpp"
 #include "4C_particle_engine_refresh_entry.hpp"
 #include "4C_particle_engine_runtime_vtp_writer.hpp"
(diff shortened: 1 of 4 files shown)


src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1200,6 +1195,17 @@ void Particle::ParticleEngine::determine_ghosting_dependent_maps_and_sets()
...
+  // cache procs to send/receive ghost particle data to/from, derived directly from the bin
+  // ghosting topology determined above (avoids MPI_Allreduce during ghost_particles())
+  cached_procs_send_ghost_particles_to_.clear();
+  for (const auto& it : binning_->thisbinsghostedby_)
+    for (int proc : it.second) cached_procs_send_ghost_particles_to_.insert(proc);
+
+  cached_procs_receive_ghost_particles_from_.clear();
+  for (int gid : binning_->ghostedbins_)
+    cached_procs_receive_ghost_particles_from_.insert(
+        binning_->binstrategy_->bin_discret()->g_element(gid)->owner());
```
jeremylt, line 1206: to clarify for my understanding - using a set here because these loops may (likely do?) contain duplication?

vovannikov: Exactly, because one rank can own multiple bins.

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1907,59 +1920,45 @@ void Particle::ParticleEngine::insert_owned_particles(
...
-      if (gidofbin < 0)
-        FOUR_C_THROW("received ghosted particle contains no information about its bin gid!");
+      // get container of ghosted particles of current particle type
+      ParticleContainer* container =
+          particlecontainerbundle_->get_specific_container(entry.type, Status::Ghosted);
 
       // add particle to container of ghosted particles
       int ghostedindex(0);
-      container->add_particle(ghostedindex, globalid, states);
+      container->add_particle(ghostedindex, entry.globalid, entry.states);
```
jeremylt, line 1950: I wonder, is there a way to add all of the particles to a container at once? I've generally had better perf from fewer, bigger memory activities than more, smaller ones

vovannikov: Unfortunately, this does not seem to be the case. The main difficulty comes from that how the particle states are stored in a `std::vector<std::vector<double>>`. However, I guess, this can be optimized for the cases of particle initialization or trivial states, at least.

jeremylt: That's a shame. We may want to revisit that in the future, with Kokkos DualViews memory manipulation speed being a potential bottleneck

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1907,59 +1920,45 @@ void Particle::ParticleEngine::insert_owned_particles(
   invalidate_particle_safety_flags();
 }
 
-void Particle::ParticleEngine::insert_ghosted_particles(
-    std::vector<std::vector<std::pair<int, ParticleObjShrdPtr>>>& particlestoinsert,
+void Particle::ParticleEngine::communicate_and_insert_ghost_particles(
+    const std::map<int, std::vector<char>>& sdata,
     std::map<int, std::map<Particle::Type, std::map<int, std::pair<int, int>>>>& directghosting)
```
ppraegla: Should `directghosting` be returned instead of being an argument?

src/particle/src/engine/4C_particle_engine.cpp
```diff
@@ -1907,59 +1920,45 @@ void Particle::ParticleEngine::insert_owned_particles(
...
+      container->add_particle(ghostedindex, entry.globalid, entry.states);
 
       // add index relating (owned and ghosted) particles to col bins
-      particlestobins_[binning_->bincolmap_->lid(gidofbin)].push_back(
-          std::make_pair(type, ghostedindex));
-
-      // get local index of particle in container of owned particles of sending processor
-      int ownedindex = particleobject->return_container_index();
+      particlestobins_[binning_->bincolmap_->lid(entry.bingid)].push_back(
+          std::make_pair(entry.type, ghostedindex));
```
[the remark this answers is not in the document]
ppraegla: I think it makes sense to keep the check of the previous version for the gid
```
FOUR_C_ASSERT_ALWAYS(entry.bingid >= 0, "received ghosted particle contains no information about its bin gid!");
```
Or even adapt it to `binning_->bincolmap_->lid(entry.bingid) >= 0` as Copilot suggested.

src/particle/src/engine/4C_particle_engine_ghost_entry.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include "4C_particle_engine_ghost_entry.hpp"
+
+#include "4C_comm_pack_helpers.hpp"
+
+FOUR_C_NAMESPACE_OPEN
+
+/*---------------------------------------------------------------------------*
+ | definitions                                                               |
+ *---------------------------------------------------------------------------*/
```
ppraegla:
```suggestion
```

We should try to reduce these types of comments.

src/particle/src/engine/4C_particle_engine_ghost_entry.cpp
```diff
@@ -0,0 +1,54 @@
...
+void Particle::ParticleGhostEntry::pack(Particle::Type type, int globalid, int bingid,
```
ppraegla:
```suggestion
void Particle::ParticleGhostEntry::pack(ParticleType type, int globalid, int bingid,
```

I would use the enum class name and not the typedef. The typedef is probably more useful when you also need to define the Particle namespace, e.g., in the tests to avoid `Particle::ParticleType`.

Same for other occurrences of `Particle::Type` in the cpp and hpp.

src/particle/src/engine/4C_particle_engine_ghost_entry.hpp
```diff
@@ -0,0 +1,95 @@
...
+// See the LICENSE.md file in the top-level for license information.
+//
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
+#ifndef FOUR_C_PARTICLE_ENGINE_GHOST_ENTRY_HPP
+#define FOUR_C_PARTICLE_ENGINE_GHOST_ENTRY_HPP
+
+/*---------------------------------------------------------------------------*
+ | headers                                                                   |
+ *---------------------------------------------------------------------------*/
```
ppraegla:
```suggestion
```

Same as in the cpp. These comments should be removed.

ppraegla (commented):
Thanks for this PR, @vovannikov. Apart from the Assert, the comments are optional.

You need to rebase the PR due to changes to GitHub Actions.

src/particle/src/engine/4C_particle_engine.cpp
vovannikov: Yes, good idea!

vovannikov:
> Thanks for this PR, @vovannikov. Apart from the Assert, the comments are optional.
> 
> You need to rebase the PR due to changes to GitHub Actions.

Yep, done. All the comments also have been addressed. Thank you for the review!

ppraegla (approved):

georghammerl (approved):

PR #2227, opened 2026-08-26
Refactor `DIVERCONT` and add time-step control to solid time integration

### LINKED ISSUE #64: Wrong log output during passive time step adaption

### LINKED ISSUE #2216: Refactor solid time-step-adaptivity and `DIVERCONT` behavior

rjoussen:
### First commit:

`DiverContAct` is refactored to an enum class and is now directly used for the input of `DIVERCONT`. For this, I renamed the `continue` option to `ignore`, because `continue` is a keyword. This required changing all tests that used `continue`.

### Second commit:

This is the actual refactor, largely described in #2216 already. In short:

#### Behavioral changes

- `DIVERCONT: adapt_step` is still available, but the time step is now adapted according to the `TIMESTEP CONTROL` input section, that looks like this:
```yaml
  TIMESTEP CONTROL:
    DECREASE_FACTOR: 0.5
    MIN_TIMESTEP_RATIO: 0.001
    STEPS_TO_INCREASE: 10
    MAX_TIMESTEP: null # or any number. Independent of the initial time step
    INCREASE_FACTOR: null # or any number >= 1. Defaults to 1/DECREASE_FACTOR
    MAX_AVERAGE_NONLINEAR_ITERATIONS: null # or any positive number. Defaults to MAXITER 
    REDUCE_TO_MAXTIME: false
```
I hope the effect of each parameter is clear from the descriptions.

- The end step of the simulation is no longer automatically adapted. This was quite the discussion in #2216. I am definitely not strictly against it, however, when I tried to implement it I encountered the problem that it is unclear what to do if `1/DECREASE_FACTOR` is not a whole number. You would need to round/ceil it, and the same thing when the time step is increased, which could lead to weird changes in the end step. @lauraengelhardt 

- This is implemented for the old and new time integration, mainly because I didn't want to document the difference between both.
- All `DIVERCONT` options except `ignore`, `stop`, `adapt_step` and `adapt_penaltycontact` are removed. `adapt_penaltycontact` still throws with not implemented in the new time integration. The previous `halbe_step` behaviour can bes restored for example by setting `MAX_TIMESTEP` to 0 or `STEPS_TO_INCREASE` to a large number.

#### Code
- `perform_error_action` is now called unconditionally in the time-loop, which makes the control flow a lot easier to understand and doesn't mix `solve()` responsibility with time-loop level decisions. 
- All post-output tasks are moved to the new adapter-level `finalize_step()` function. The base implementation just prints the step, derived methods add their functionality to it, e.g. `Solid::TimeInt::Base` adds the dynamic rebalance hook, `Solid::TimeInt::Implicit` adds the check for time-step recovery.
- `StepStatus` is split into actual status and a new `StepAction`, because the previous `fail_repeat` is not a status. This new enum is the return type of `perform_error_action`

The input spec and all associated math for the time-step size computation is in the neutral `TimeStepping` module, such that other algorithms can use it as well.

#### Testing

- The time step control mechanism is unittested (I hope every branch)
- 3 regression tests, check their descriptions for details.
- Some of the beam tests needed to be updated, see #2211 for details why.

Closes #2216 
Closes #64

#2189 

All layout decisions are from me, also the implementation for the new time integration. Ai assited with writing/setting up the unittests as well as copying the behaviour to the old time integration (though that didnt fully work unfortunately)


src/structure_new/src/nonlinear_solver/4C_structure_new_nln_solver_nox.cpp
```diff
@@ -203,7 +203,7 @@ Solid::StepStatus Solid::Nln::SOLVER::Nox::solve()
 
   // Check if we do something special if the non-linear solver fails,
   // otherwise an error is thrown.
-  if (data_sdyn().get_divergence_action() == Solid::divcont_stop)
+  if (data_sdyn().get_divergence_action() == Solid::DivContAct::stop)
     problem_->check_final_status(finalstatus);
```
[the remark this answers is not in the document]
rjoussen: Removed

src/structure/4C_structure_timint_impl.cpp
```diff
@@ -110,6 +119,135 @@ Solid::TimIntImpl::TimIntImpl(const Teuchos::ParameterList& timeparams,
...
+    case StepStatus::nonlinear_solver_failed:
+    {
+      if (write_cout) std::cout << "Nonlinear solver did not converge!\n";
+      const auto divergence_action = divcontype_;
+      switch (divergence_action)
+      {
+        case DivContAct::adapt_step:
+        {
+          prepare_retry_with_reduced_time_step();
+          return StepAction::retry_step;
```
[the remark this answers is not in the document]
rjoussen: I am unsure if we should test that, since the old time integration strategy is deprecated anyway and I mainly implemented it out of convenience/to be able to reduce the divercontact enum. I did try it out locally though and all the test work, except that the actual result tests did not because they don't exist in the old strategy.

src/structure_new/src/input/4C_structure_new_input.cpp
```diff
@@ -276,26 +279,12 @@ namespace Solid
...
-            parameter<int>("MAXDIVCONREFINEMENTLEVEL",
-                {.description =
-                        "number of times timestep is halved in case nonlinear solver diverges",
-                    .default_value = 10}),
+            parameter<DivContAct>("DIVERCONT",
+                {.description = "Action to be taken in case the nonlinear solver does not "
+                                "converge. If `adapt_step`, the time-step size is adapted "
+                                "according to the `TIMESTEP CONTROL` settings. `ignore` ignores "
+                                "the non-convergence and continues anyway.",
+                    .default_value = DivContAct::stop}),
```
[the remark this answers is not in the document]
rjoussen: I will add a commit for the documentation once the actual shape of the refactor is final

src/timestepping/4C_timestepping_time_step_control.cpp
```diff
@@ -0,0 +1,156 @@
...
+        reduce_to_max_time(input.reduce_to_max_time)
+  {
+  }
+
+  [[nodiscard]] Core::IO::InputSpec TimeStepControlSettings::input_spec()
+  {
+    using namespace Core::IO::InputSpecBuilders;
+    using namespace Core::IO::InputSpecBuilders::Validators;
+
+    return group<InputParameters>("TIMESTEP CONTROL",
```
davidrudlstorfer, line 42: Just a general remark: I would prefer if we could just use snake case for all new parameters.

Some poro modules and some beam modules already use them and with every new "module" it would be nice to just go to an easier readable format. This was also very briefly talked about in the mysterious issue #74 

Especially I want to highlight that for example your top section key contains space whereas the sub keys are with underscores.

davidrudlstorfer (commented):
Just a minor remark that is probably still up for discussion

src/timestepping/4C_timestepping_time_step_control.cpp
rjoussen: I agree that snake case is nicer to read than all-caps. Here, I followed the convention within the `STRUCTURAL DYNAMIC` section. Though, since the intention of this input section is to be used directly for other algorithms as well, I will refactor it. Although I don't fully understand why snake case, and not the directly grammatically correct version with spaces and everything. But that's more a question of #74, which I don't think this PR is linked to ;)

rjoussen: done.

src/structure_new/src/input/4C_structure_new_input.cpp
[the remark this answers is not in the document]
rjoussen: I ended up removing the duplicated information in the structural analysis guide. The input file already documents this option, any additional information here is hard to keep up to date.

src/structure/4C_structure_timint_impl.cpp
```diff
@@ -110,6 +119,135 @@ Solid::TimIntImpl::TimIntImpl(const Teuchos::ParameterList& timeparams,
...
+      number_of_nonlinear_iterations_per_step_, timemax_ - (*time_)[0], stepmax_ - step_);
+
+  if (std::abs(new_dt - (*dt_)[0]) > std::numeric_limits<double>::epsilon())
+  {
+    if (myrank_ == 0)
+    {
+      std::cout << "Updating the time-step size for the next step:\n";
+    }
+    set_new_time_step_size(new_dt);
+  }
```
c-p-schmidt, line 219: Maybe you want to inform the user in an else branch that a new time step size has been calculated, but ignored because it does not deviate much from the original?

src/structure/4C_structure_timint_impl.cpp
```diff
@@ -110,6 +119,135 @@ Solid::TimIntImpl::TimIntImpl(const Teuchos::ParameterList& timeparams,
...
+  set_new_time_step_size(new_dt);
+
+  reset_step();
+}
+
+void Solid::TimIntImpl::set_new_time_step_size(const double new_dt)
+{
+  const double old_dt = (*dt_)[0];
+
+  FOUR_C_ASSERT(new_dt > 0.0, "Time-step size must be positive.");
```
c-p-schmidt: Don't you always want to do these two checks? They should not be expensive, but the behavior might be pretty bad in simulation runs.

src/structure_new/src/implicit/4C_structure_new_timint_implicit.cpp
```diff
@@ -222,232 +223,142 @@ const ::NOX::Abstract::Group& Solid::TimeInt::Implicit::get_solution_group() con
...
 
-  const int maxnumfinestep = 4;
+  const double old_dt = get_delta_time();
 
-  if (get_divergence_action() != Solid::divcont_adapt_step)
-    return;
-  else if (status == Solid::StepStatus::no_errors and get_div_con_refine_level() != 0)
-  {
-    set_div_con_num_fine_step(get_div_con_num_fine_step() + 1);
+  FOUR_C_ASSERT(new_dt > 0.0, "Time-step size must be positive.");
```
c-p-schmidt: Same here, as it is the corresponding method in the new time integration

src/structure/4C_structure_timint_impl.cpp
rjoussen: There is currently no code path in which this could happen, but I agree that if this happens in the future and remains undetected its quite bad. So I am changing it to always.

src/structure/4C_structure_timint_impl.cpp
rjoussen: The problem is that there are at least 5 different possibilities why this might happen:
1. no remaining time/steps
2. the current dt is already larger than the maximum dt.
3. increase factor is 1
4. dt was adjusted but then the reduce to max time options reduces it back to exactly the old step size.
5. the number of succesfull steps at current dt is too small

So just informing about it is of little value, I think. In fact, in most of these cases a "did not adjust time-step" log would be strange. At the same time, I also dont want to log the entire math leading to the new time-step size, that is why logging happens here.

c-p-schmidt (dismissed):

src/structure/4C_structure_timint_impl.cpp
```diff
@@ -110,6 +119,135 @@ Solid::TimIntImpl::TimIntImpl(const Teuchos::ParameterList& timeparams,
...
+        }
+        case DivContAct::ignore:
+        {
+          if (write_cout) std::cout << "WARNING: Nonlinear solver non-convergence is ignored.\n";
+          return StepAction::accept_step;
+        }
+        case DivContAct::stop:
+        {
+          output(true);
+          FOUR_C_THROW("Abort due to DIVERCONT: {}.", divergence_action);
```
lauraengelhardt:
```suggestion
          FOUR_C_THROW("Nonlinear solver did not converge. Aborting since DIVERCONT is set to {}.", divergence_action);
```

src/structure/4C_structure_timint_impl.cpp
```diff
@@ -110,6 +119,135 @@ Solid::TimIntImpl::TimIntImpl(const Teuchos::ParameterList& timeparams,
...
+        }
+        case DivContAct::adapt_penaltycontact:
+        {
+          if (have_contact_meshtying())
+          {
+            if (write_cout) std::cout << "Adapting penalty parameter.\n";
```
lauraengelhardt:
```suggestion
            if (write_cout) std::cout << ("Nonlinear solver did not converge. DIVERCONT is set to {}. Adapting the contact penalty parameter.\n", divergence_action);
```

lauraengelhardt: And for the actions below as well? I think its helpful to first know what is the problem and then why what is adapted :)

src/structure/4C_structure_timint_impl.cpp
```diff
@@ -110,6 +119,135 @@ Solid::TimIntImpl::TimIntImpl(const Teuchos::ParameterList& timeparams,
...
+    case StepStatus::linear_solver_failed:
+    {
+      FOUR_C_THROW("Linear solver did not converge!");
+    }
+    case StepStatus::evaluation_failed:
+    {
+      FOUR_C_THROW("Evaluation of the residual or jacobian failed!");
+    }
+    default:
+      FOUR_C_THROW("Inconsistent step status.");
```
lauraengelhardt: I know we need a default action but should the user end up here? Something like: "Unexpected step status. A default action is required, but reaching this branch indicates an unexpected state."

src/structure_new/src/implicit/4C_structure_new_timint_implicit.cpp
```diff
@@ -222,232 +223,142 @@ const ::NOX::Abstract::Group& Solid::TimeInt::Implicit::get_solution_group() con
...
+        }
+        case DivContAct::ignore:
+        {
+          if (write_cout) std::cout << "WARNING: Nonlinear solver non-convergence is ignored.\n";
+          return StepAction::accept_step;
+        }
+        case DivContAct::stop:
+        {
+          output(true);
+          FOUR_C_THROW("Abort due to DIVERCONT: {}.", divergence_action);
```
lauraengelhardt: Same for these error messages as my comment above.

src/timestepping/4C_timestepping_time_step_control.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+    /// Return the input specification for the time_step_control parameter group.
+    [[nodiscard]] static Core::IO::InputSpec input_spec();
+
+    /// Factor used to reduce the time-step size after a failed step.
+    double decrease_factor;
+    /// Minimum allowed absolute time-step size.
+    double min_time_step;
+    /// Number of accepted steps before a time-step increase may be attempted.
+    size_t steps_to_increase;
```
lauraengelhardt: Why is steps_to_increase here of size_t and in the InputParameters struct of  type int?

lauraengelhardt: Also couldn't these parameters be const?

src/timestepping/4C_timestepping_time_step_control.cpp
```diff
@@ -0,0 +1,155 @@
...
+                        "iterations over the last step_to_increase steps is below this value. "
+                        "If omitted, time-step increase is attempted regardless of the "
+                        "number of Newton iterations.",
+                    .validator = Validators::null_or(Validators::positive<double>()),
+                    .store = in_struct(&TimeStepControlSettings::InputParameters::
+                            max_average_nonlinear_iterations)}),
+            parameter<bool>("reduce_to_max_time",
+                {.description = "If true, the last time-step size of the simulation is reduced to "
+                                "exactly reach the "
+                                "specified maximum simulation time.",
```
lauraengelhardt: Just formatting here.

src/timestepping/4C_timestepping_time_step_control.cpp
```diff
@@ -0,0 +1,155 @@
...
+                        "never been reduced.If this is smaller than the initial time-step size, "
+                        "time-step increase is only attempted once the current time-step size has "
+                        "been reduced below this value.",
+                    .validator = Validators::null_or(Validators::positive<double>()),
+                    .store = in_struct(&TimeStepControlSettings::InputParameters::max_time_step)}),
+            parameter<std::optional<double>>("increase_factor",
+                {.description = "Factor applied when increasing the global time-step size "
+                                "during time-step increase. If omitted, "
+                                "increase_factor = 1/decrease_factor. If 1.0, the time step is "
+                                "never increased.",
```
lauraengelhardt: And here.

lauraengelhardt (commented):
I left some comments regarding the error messages - the rest looks good imo :) Thank you!

src/structure/4C_structure_timint_impl.cpp
rjoussen: The information that the nonlinear solver did not converge is printed immediately when we enter the case nonlinear_solver_failed (line 132). So the information is there already. It's of course not harmful to print it here again though. Do you want me to do that?

src/timestepping/4C_timestepping_time_step_control.hpp
rjoussen: > Why is steps_to_increase here of size_t and in the InputParameters struct of type int?

It is handy to have it as size_t because we later compare it to size() things. In the input it can't be size_t because that is not a supported type of input_spec currently.

rjoussen: > Also couldn't these parameters be const?

I changed it, but it is a little awkward because basedatasdyn can only construct it in the init() function, not its constructor. So the settings in basedatasdyn are now std::optional, to make this possible.

src/structure/4C_structure_timint_impl.cpp
rjoussen: changed here and in the new time integration.

src/structure/4C_structure_timint_impl.cpp
rjoussen: I adapted the logging slightly now. "nonelinear solver failed" and the divercont option is always printed if nonlinear solver failed. Afterwards, normal exits only print what they are doing. FOUR_C_THROW exits print the whole info again (nonlinear solver failed, divercont is set to ..., aborting)

lauraengelhardt: I wasn't aware, I'm fine with that solution :)

lauraengelhardt (dismissed):
Looks good to me, thanks for incorporating the suggestions!
If you could only squash/organize the commits before merging, that would be great.

rjoussen:
I needed to rebase, now ready.

c-p-schmidt (approved):

lauraengelhardt (approved):

PR #2231, opened 2026-08-27
[beaminteraction] Add line2 templating for beam solid volume coupling

maxfirmbach:
## Description and Context

This PR enables the use of `beam3rline2line2` elements for beam-solid volume coupling. Therefore, mainly the respective templates are expanded.

Related to #2208


isteinbrecher (approved):
Looks good to me. Did you check if the test produces similar results to the original test?

maxfirmbach:
@isteinbrecher I did, there is a difference of about 10^-3 for all quantities.

tests/input_files/beam3r_line2line2_static_beam_to_solid_volume_meshtying_mortar_lagrange_saddlepoint_beam_to_beam_penalty_coupling.4C.yaml
isteinbrecher, line 1: Do we really need this test? What does it test?

maxfirmbach: Good point. I added this test mainly to see if the results are actually similar to the one using a hermite centerline, also in the case of beam-beam point couplings. I can delete this one though if not wanted 👍.

mayrmt (approved):

PR #2236, opened 2026-08-28
Add output of artery element length

lkoeglmeier:
This pr adds the output of the artery element length. Due to the coupling with the `porofluid_pressure_based*` framework the elements can follow the deformation of the tissue and change their length.


src/art_net/4C_art_net_artery_ele_calc_pres_based.cpp
```diff
@@ -216,12 +217,14 @@ void Discret::Elements::ArteryEleCalcPresBased<distype>::evaluate_flow(Artery* e
 
   // calculate the element length
   const double L = calculate_ele_length(ele, discretization, la);
+  FOUR_C_ASSERT(ele_length.length() == 1, "ele_length must be of size 1");
+  ele_length(0) = L;
```
davidrudlstorfer, line 221: if this vector must be size 1, cannot you change this to a float directly?

Seems a bit weird to me that you initially check that the vector has length one and then you use it as a regular float by writing something into the first entry

tests/input_files/ref/porofluid_pressure_based_elast_scatra_2D_quad4_nodetopoint_artery_airway_coupling_mono-vtk-files/artery-00024-0.vtu
davidrudlstorfer, line 1: Please change these results to actually readable files, aka, ASCII encoding

src/art_net/4C_art_net_impl_stationary.cpp
```diff
@@ -575,7 +583,7 @@ void Arteries::ArtNetImplStationary::get_radius()
 //<><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>//
 //<><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>//
 //<><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>//
-void Arteries::ArtNetImplStationary::reconstruct_flow()
+void Arteries::ArtNetImplStationary::get_element_flow_and_length()
```
maxiludwig: Just a suggestion: I was wandering about the naming: is it really a `get_` or would maybe `compute_` be clearer?

maxiludwig (approved):

davidrudlstorfer (approved):

PR #2237, opened 2026-08-31
Unify reduced lung test mesh construction

maxiludwig:
## Description and Context

The junction and boundary condition tests built their meshes from `Discret::Elements::RedAirway`, which they only used as a generic two-node line element.
They now share the mesh helpers of `4C_reduced_lung_helpers_test.cpp`, which move into the shared test utils and build via `build_discretization_from_nodes_and_elements` like the production code does.
This removes an undeclared dependency of the reduced lung tests on `red_airways`, along with three near-duplicate mesh helpers.

The changes in this PR were made by Claude Opus 5.

### DIFF
--- a/src/reduced_lung/tests/4C_reduced_lung_test_utils_test.hpp
+++ b/src/reduced_lung/tests/4C_reduced_lung_test_utils_test.hpp
@@ -13,12 +13,19 @@
 
 #include "4C_config.hpp"
 
+#include "4C_fem_discretization.hpp"
 #include "4C_linalg_map.hpp"
 #include "4C_linalg_sparsematrix.hpp"
 #include "4C_linalg_vector.hpp"
+#include "4C_rebalance.hpp"
 #include "4C_reduced_lung_airways.hpp"
+#include "4C_reduced_lung_helpers.hpp"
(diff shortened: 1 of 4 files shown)


src/reduced_lung/tests/4C_reduced_lung_test_utils_test.hpp
```diff
@@ -13,19 +13,76 @@
...
 #include "4C_linalg_vector.hpp"
+#include "4C_rebalance.hpp"
 #include "4C_reduced_lung_airways.hpp"
+#include "4C_reduced_lung_helpers.hpp"
 #include "4C_reduced_lung_terminal_unit.hpp"
 
+#include <mpi.h>
+
+#include <array>
+#include <memory>
```
davidrudlstorfer, line 28: Do we really need these imports here?

davidrudlstorfer (approved):

src/reduced_lung/tests/4C_reduced_lung_test_utils_test.hpp
maxiludwig: They all are used, but they would also come with the included header files.

lkoeglmeier (approved):

PR #2241, opened 2026-09-04
Alveolar recruitment derecruitment

maxiludwig:
## Description and Context

This PR makes the reference volume a state variable driven by a piecewise-linear pressure law on an opening/closing hysteresis loop, with an optional exponential relaxation of the response, and advances it at the end of each time step. The related paper is Geitner et al. 2024 (https://doi.org/10.1002/cnm.3787) 
Recruitment is modelled as a third constitutive family alongside rheology and elasticity: it is a variant, its type is part of the terminal unit model block key, and a tree that mixes recruiting and rigid terminal units simply splits into one block per recruitment model. Rheology and elasticity reach the reference volume through a new reference volume context so neither constitutive law had to be reimplemented for the recruiting case.

The reference volume can optionally be linearized in the Jacobian (`reference_volume_linearization`, `Frozen` by default), which keeps Newton convergence when the reference volume follows the pressure within a step. Existing input files are unaffected: both the pressure law and the time law default to `None`, and a terminal unit without recruitment carries no element-wise recruitment data at all.

Verified with 18 new unit tests — the piecewise law on both hysteresis branches, the time law, the state advance, path switching in both directions, block splitting, input validation, output, and four finite-difference checks of the analytic Jacobian against the coupled linearization.

### Commit structure

| Commit | |
|---|---|
| `Add a terminal unit reference volume context` | Pure refactor: compute the reference volume once per element instead of dividing by it at seven call sites. No behaviour change. |
| `Add terminal unit recruitment input parameters` | The user-facing input contract only; nothing consumes it yet. |
| `Make the reference volume a recruitment state` | The model itself: pressure law, time law, hysteresis, state advance, output, and the model block key. |
| `Linearize the recruitment reference volume` | Carry dV0/dp into the Jacobian and split the elastic pressure gradient into its partials. Opt-in. |
| `Test terminal unit recruitment end to end` | Three input files: pressure law alone, with relaxation, and a mixed tree. |

Written with Claude Code (Claude Opus 5), used for implementation, refactoring, tests and commit history. All commits carry a `Co-Authored-By` trailer. I reviewed and validated the result and take responsibility for it.


src/reduced_lung/src/4C_reduced_lung_input.cpp
```diff
@@ -257,6 +257,100 @@ Core::IO::InputSpec ReducedLung::valid_parameters()
...
+                                 "None keeps it at the geometry-derived constant.",
+                  .store = in_struct(&ReducedLungParameters::LungTree::TerminalUnits::
+                          RecruitmentModel::pressure_law_type),
+              }),
+          input_field<
+              ReducedLungParameters::LungTree::TerminalUnits::RecruitmentModel::TimeLawType>(
+              "time_law_type",
+              {
+                  .description = "Time law for recruitment reference-volume evolution.",
+                  .store = in_struct(&ReducedLungParameters::LungTree::TerminalUnits::
```
bgoderbauer: Maybe add a little more information on the time law here?

src/reduced_lung/src/4C_reduced_lung_input.cpp
```diff
@@ -257,6 +257,100 @@ Core::IO::InputSpec ReducedLung::valid_parameters()
...
+                          RecruitmentModel::time_law_type),
+              }),
+          input_field<ReducedLungParameters::LungTree::TerminalUnits::RecruitmentModel::
+                  ReferenceVolumeLinearization>("reference_volume_linearization",
+              {
+                  .description = "Linearization mode for recruitment reference-volume "
+                                 "sensitivities.",
```
bgoderbauer: Also here

src/reduced_lung/src/4C_reduced_lung_input.cpp
```diff
@@ -257,6 +257,100 @@ Core::IO::InputSpec ReducedLung::valid_parameters()
...
+                  input_field<double>("p_opening_min",
+                      {.description = "Minimal critical pressure on opening path.",
+                          .store = in_struct(&ReducedLungParameters::LungTree::TerminalUnits::
+                                  RecruitmentModel::LinearPressure::p_opening_min)}),
+                  input_field<double>("delta_p_minmax",
+                      {.description = "Pressure span between min and max critical pressure.",
+                          .store = in_struct(&ReducedLungParameters::LungTree::TerminalUnits::
+                                  RecruitmentModel::LinearPressure::delta_p_minmax)}),
+                  input_field<double>("epsilon_v0_switch",
+                      {.description = "Tolerance for full open/closed state switching.",
```
bgoderbauer: Is this a tolerance for numerical stability? If yes, I would add this info and maybe a default

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit_recruitment.hpp
```diff
@@ -0,0 +1,187 @@
...
+#include "4C_utils_exceptions.hpp"
+
+#include <utility>
+#include <variant>
+#include <vector>
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace ReducedLung::TerminalUnits::Recruitment
+{
```
bgoderbauer: We want to avoid overly nested namespaces, so I wouldn't add this one. The other models also lie directly in `TerminalUnits` as the bottom layer

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit.cpp
```diff
@@ -85,6 +87,31 @@ namespace ReducedLung
...
+       * all others it is produced by the configured recruitment law. Elasticity and rheology
+       * consume the result through TerminalUnitData::reference_volume_context without needing to
+       * know which of the two cases applies.
+       */
+      void refresh_reference_volume_contexts(TerminalUnitData& data,
+          const Recruitment::RecruitmentModel& recruitment_model,
+          const Core::LinAlg::Vector<double>& locally_relevant_dofs, const double dt)
+      {
+        FOUR_C_ASSERT(data.reference_volume_context.size() == data.number_of_elements(),
+            "Reference volume context vector is not sized with the terminal-unit model block.");
```
bgoderbauer: I would remove the assert inside this function. It compares two things, both of which lie inside `data`, so I would usually expect this structure to be correct before it is given to the function.

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit_common.hpp
```diff
@@ -26,6 +27,33 @@ namespace ReducedLung
...
+   * units without recruitment and for the frozen linearization.
+   */
+  struct ReferenceVolumeContext
+  {
+    ///< Effective reference volume.
+    double v0_eff = std::numeric_limits<double>::quiet_NaN();
+    ///< Reciprocal of the effective reference volume, cached because assembly mostly needs it.
+    double inv_v0_eff = std::numeric_limits<double>::quiet_NaN();
+    ///< Derivative of the effective reference volume w.r.t. the transmural pressure.
+    double dv0_dp = std::numeric_limits<double>::quiet_NaN();
```
bgoderbauer: I don't quite understand why we need quiet_NaNs here?

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit_recruitment.hpp
```diff
@@ -0,0 +1,187 @@
...
+        return;
+    }
+    FOUR_C_THROW("Unknown recruitment pressure-law type enum value.");
+  }
+
+  /**
+   * @brief Reference volume of one element at the last converged time step.
+   *
+   * The `_n` suffix matches the state vectors this reads. Blocks without a recruitment law
+   * report their constant geometry-derived value.
```
bgoderbauer: This reads a little weird.

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit_recruitment.cpp
```diff
@@ -0,0 +1,413 @@
...
+            "Terminal unit {}: recruitment parameter 'epsilon_v0_switch' must be in [0, 0.5).",
+            global_element_id);
+      }
+
+      if (time_law_type == TimeLawType::ExponentialRelaxation && !(tau > 0.0))
+      {
+        FOUR_C_THROW(
+            "Terminal unit {}: recruitment parameter 'tau' must be > 0 for exponential "
+            "relaxation time law.",
+            global_element_id);
```
bgoderbauer: You can use `FOUR_C_ASSERT_ALWAYS` here. Saves a few lines 😄

bgoderbauer (changes requested):
I've looked into the implementation details and have a few comments. Overall, it is quite easy to understand, that's very nice! I have one major concern at the moment: the way you implemented the recruitment kernels is not really in line with the existing kernels for the elastic and rheological model evaluators. At the moment, there is no recruitment model "creator"; instead, the recruitment model dispatch is repeated for every element and iteration. This makes the kernels harder to read and likely slower in performance. I'd propose setting the dispatch again up front during creation and storing the callables directly in the model blocks. This way, we can again loop over all elements in a single model block in the innermost kernel, making it easier to read and faster to execute, and we don't need to call `std::visit` repeatedly everywhere.

src/reduced_lung/src/4C_reduced_lung_input.cpp
maxiludwig: No, it's not for numerical stability. It comes from the fact, that the time relaxation curve would never actually reach its target, as it only converges towards it. So this tolerance considers when to assume that the relaxation has reached the target.

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit_common.hpp
maxiludwig: They were values for the gap between block creation and the first state sync. The registry now seeds the context with the initial reference volume, so there's no gap and no need for the initializers.

maxiludwig:
Thanks for the valuable review, @bgoderbauer! Recruitment now has `make_internal_state_updater` and `make_end_of_timestep_routine`, both dispatching once at creation and looping over the block inside the callable. I named it internal_state_updater so it matches the rheology and airway wall/flow state updaters.

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit_common.hpp
```diff
@@ -26,6 +26,34 @@ namespace ReducedLung
 
 namespace ReducedLung::TerminalUnits
 {
+  /**
+   * @brief Effective reference volume of one terminal-unit element and its pressure derivatives.
+   *
+   * Recruitment makes the reference volume depend on the transmural pressure. The derivatives
```
lkoeglmeier:
```suggestion
   * Recruitment makes the reference volume depend on the transpulmonary pressure. The derivatives
```

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit_common.hpp
```diff
@@ -26,6 +26,34 @@ namespace ReducedLung
...
+   * units without recruitment and for the frozen linearization. Built through
+   * make_reference_volume_context().
+   */
+  struct ReferenceVolumeContext
+  {
+    ///< Effective reference volume.
+    double v0_eff;
+    ///< Reciprocal of the effective reference volume, cached because assembly mostly needs it.
+    double inv_v0_eff;
+    ///< Derivative of the effective reference volume w.r.t. the transmural pressure.
```
lkoeglmeier: also here transpulmonary?

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit_elasticity.cpp
```diff
@@ -58,40 +60,54 @@ namespace ReducedLung::TerminalUnits::Elasticity
     }
 
     /**
-     * Compute d(p_el)/dq for the linear elasticity model.
+     * Compute d(p_el)/dq and d(p_el)/dV0 for the linear elasticity model.
```
lkoeglmeier:
```suggestion
     * Compute d(p_el)/dq and d(p_el)/dv0 for the linear elasticity model.
```

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit.cpp
```diff
@@ -140,28 +141,60 @@ namespace ReducedLung
...
+          // guaranteed to have happened at this dof vector, and both history updates below must
+          // see the reference volume of the time step that just converged.
+          recruitment_state_updater(data, locally_relevant_dofs, dt);
+          rheology_end_of_timestep_routine(data, locally_relevant_dofs, dt);
+          recruitment_end_of_timestep_routine(data, locally_relevant_dofs, dt);
+        };
         auto elasticity_output_evaluator =
             Elasticity::make_output_evaluator(model.elasticity_model);
         auto rheology_output_evaluator = Rheology::make_output_evaluator(model.rheological_model);
+        auto recruitment_output_evaluator =
```
lkoeglmeier, line 186: Is it on purpose that you only name it recruitment and don't include derecruitment in the names?

tests/input_files/reduced_lung_terminal_unit_recruitment_mixed.4C.yaml
```diff
@@ -0,0 +1,82 @@
+TITLE:
+  - "3 terminal units with mixed recruitment models"
```
lkoeglmeier, line 2: at first sight, it is not super clear to me what "mixed" means here. Maybe you could add a more detailed explanation if you think it makes sense here.

lkoeglmeier (commented):
I just have a few small comments. It looks good to me.

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit.cpp
maxiludwig: Yes, otherwise every name would get super long. I would argue that recruitment and derecruitment go hand in hand and someone reading the code will quickly understand that it refers to both. In a paper I would still speak of recruitment-derecruitment

tests/input_files/reduced_lung_terminal_unit_recruitment_mixed.4C.yaml
maxiludwig: You're right. The test has three TUs with three models for RD. 
1. no RD
2. Pressure-dependent RD
3. Pressure- and time-dependent RD

src/reduced_lung/src/terminal_units/4C_reduced_lung_terminal_unit.cpp
lkoeglmeier: fine for me. Maybe you could add this information to a comment, unless I missed it.

lkoeglmeier (dismissed):

lkoeglmeier (approved):

bgoderbauer:
Thanks @maxiludwig! I think this is in a good state now

bgoderbauer (approved):

bgoderbauer:
I'm curious, did you already check the timings of your implementation compared to a non-recruitment case?

maxiludwig:
> I'm curious, did you already check the timings of your implementation compared to a non-recruitment case?

Not really as this will be strongly dependent on the parameters of the RD. So i will first have to construct a distribution of critical pressures that is somewhat realistic to test this.

PR #2243, opened 2026-09-07
Allow local Newton helper to return status

rjoussen:
## Description and Context

Adds a local newton helper that can return a more detailed status, including wether it converged or not. This is helpful if non-convergence should not lead to throws immediately. Existing local newton APIs remain unchanged.

assisted by gpt 5.5

### DIFF
--- a/src/core/utils/src/numerics/4C_utils_local_newton.hpp
+++ b/src/core/utils/src/numerics/4C_utils_local_newton.hpp
@@ -18,7 +18,10 @@
 #include "4C_linalg_tensor.hpp"
 #include "4C_utils_fad.hpp"
 
+#include <algorithm>
+#include <cmath>
 #include <functional>
+#include <limits>
 
 FOUR_C_NAMESPACE_OPEN
 
@@ -26,6 +29,27 @@ namespace Core::Utils
 {
   constexpr double LOCAL_NEWTON_DEFAULT_TOLERANCE = 1e-12;
   constexpr unsigned LOCAL_NEWTON_DEFAULT_MAXIMUM_ITERATIONS = 50;
+
+  /// Result of a local Newton solve that may end without convergence.
+  template <typename VectorType, typename JacobianType>
+  struct LocalNewtonResult
+  {
+    /// Final Newton iterate. If \c is_converged is true, this is the converged solution. Otherwise,
+    /// this is the last computed iterate.
+    VectorType x;
+
+    /// Jacobian evaluated at x.
+    JacobianType jacobian;
+
+    /// L2 norm of the residual evaluated at x.
+    double residuum_norm;
+
+    /// Number of Newton iterations performed.
+    unsigned iteration_count;
+
+    /// Whether the Newton method converged within the maximum number of iterations.
+    bool is_converged;
+  };
 
   /// @brief Free functions defining a Newton iterations for different scalar, vector and jacobian
   /// types.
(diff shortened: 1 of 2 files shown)


src/core/utils/src/numerics/4C_utils_local_newton.hpp
```diff
@@ -110,6 +131,48 @@ namespace Core::Utils
...
+   * @brief Runs the local Newton method and returns a LocalNewtonResult.
+   *
+   * @tparam ScalarType The type of the scalar used within method (type of tolerance or norm of
+   * residuum).
+   * @tparam VectorType The type of the residuum and the unknowns.
+   * @tparam ResiduumAndJacobianEvaluator A class that defines the operator() with the signature
+   * std::tuple<VectorType, JacobianType>(VectorType) that evaluates the residuum and it's
+   * jacobian at a specific point.
+   * @param residuum_and_jacobian_evaluator A function object that evaluates the residuum and it's
+   * jacobian at a specific point.
```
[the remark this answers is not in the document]
rjoussen: changed it here and in the other places that already used this.

src/core/utils/src/numerics/4C_utils_local_newton.hpp
```diff
@@ -110,6 +131,48 @@ namespace Core::Utils
...
+      local_newton_iteration(x_0, residuum, std::move(jacobian));
+
+      std::tie(residuum, jacobian) = residuum_and_jacobian_evaluator(x_0);
+      residuum_norm = l2_norm(residuum);
+
+      ++iteration;
+    }
+
+    const bool converged = !(residuum_norm > tolerance);
+    return {x_0, jacobian, FADUtils::cast_to_double(residuum_norm), iteration, converged};
```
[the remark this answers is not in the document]
rjoussen: done.

maxfirmbach:
@rjoussen 

> This is helpful if non-convergence should not lead to throws immediately.

I'm not convinced by this statement 😉. How is this justified? Maybe you can explain why this is necessary.

rjoussen:
> I'm not convinced by this statement 😉. How is this justified? Maybe you can explain why this is necessary.

@maxfirmbach: A material performs local Newton, but does not reach convergence. Instead of throwing, I might want to try other recovery strategies, e.g., changing my initial guess, or (that's why I want it here) request a smaller global time step. Only if such recovery fails, I want to throw. If local Newton unconditionally throws, I don't have this possibility.

maxfirmbach:
@rjoussen Alright 👍.

src/core/utils/src/numerics/4C_utils_local_newton.hpp
```diff
@@ -27,6 +28,27 @@ namespace Core::Utils
...
+    JacobianType jacobian;
+
+    /// L2 norm of the residual evaluated at x.
+    double residuum_norm;
+
+    /// Number of Newton iterations performed.
+    unsigned iterations;
+
+    /// Whether the Newton method converged within the maximum number of iterations.
+    bool converged;
```
jeremylt: minor - personally I think this is a bit clearer and is consistent with other bools I've seen in the codebase
```suggestion
    bool is_converged;
```

src/core/utils/src/numerics/4C_utils_local_newton.hpp
```diff
@@ -27,6 +28,27 @@ namespace Core::Utils
   constexpr double LOCAL_NEWTON_DEFAULT_TOLERANCE = 1e-12;
   constexpr unsigned LOCAL_NEWTON_DEFAULT_MAXIMUM_ITERATIONS = 50;
 
+  /// Result of a local Newton solve that may end without convergence.
+  template <typename VectorType, typename JacobianType>
+  struct LocalNewtonResult
+  {
+    /// Final Newton iterate. If converged is true, this is the converged solution. Otherwise, this
```
jeremylt:
```suggestion
    /// Final Newton iterate. If `is_converged` is true, this is the converged solution. Otherwise, this
```
Does 4C use code markup in documentation? The sentence is harder to parse when it's less obvious that this is a variable name, in my opinion

jeremylt:
seems like a helpful feature to add that matches what I've seen in most solvers I'm familiar with in other codebases

src/core/utils/src/numerics/4C_utils_local_newton.hpp
```diff
@@ -27,6 +28,27 @@ namespace Core::Utils
...
+    /// is the last computed iterate.
+    VectorType x;
+
+    /// Jacobian evaluated at x.
+    JacobianType jacobian;
+
+    /// L2 norm of the residual evaluated at x.
+    double residuum_norm;
+
```
dragos-ana, line 46: An increment norm would be nice as well, but I've seen that the entire framework only checks the residuals and not the increments. Maybe something similar to what is done in `InelasticDefgradTransvIsotropElastViscoplastUtils` with different convergence checks would also be interesting here. 
But of course, not part of this PR ;)

src/core/utils/src/numerics/4C_utils_local_newton.hpp
```diff
@@ -27,6 +28,27 @@ namespace Core::Utils
...
+    /// Number of Newton iterations performed.
+    unsigned iterations;
```
dragos-ana:
```suggestion
    unsigned iteration_count;
```
Maybe just a tiny bit more expressive?

src/core/utils/src/numerics/4C_utils_local_newton.hpp
```diff
@@ -110,6 +132,49 @@ namespace Core::Utils
   inline double l2_norm(const Core::LinAlg::SerialDenseVector& x) { return Core::LinAlg::norm2(x); }
   /// @}
 
+  /*!
+   * @brief Runs the local Newton method and returns a LocalNewtonResult.
+   *
+   * @tparam ScalarType The type of the scalar used within method (type of tolerance or norm of
```
dragos-ana:
```suggestion
   * @tparam ScalarType The type of the scalar used within the method (type of tolerance / type of residuum norm).
```

src/core/utils/src/numerics/4C_utils_local_newton.hpp
```diff
@@ -110,6 +132,49 @@ namespace Core::Utils
...
+
+      std::tie(residuum, jacobian) = residuum_and_jacobian_evaluator(x_0);
+      residuum_norm = l2_norm(residuum);
+
+      ++iteration;
+    }
+
+    const bool converged = !(residuum_norm > tolerance);
+    return {std::move(x_0), std::move(jacobian), FADUtils::cast_to_double(residuum_norm), iteration,
+        converged};
```
dragos-ana: Using designated initializers might be better instead of relying on the order.

src/core/utils/src/numerics/4C_utils_local_newton.hpp
```diff
@@ -110,6 +132,49 @@ namespace Core::Utils
...
+    {
+      local_newton_iteration(x_0, residuum, std::move(jacobian));
+
+      std::tie(residuum, jacobian) = residuum_and_jacobian_evaluator(x_0);
+      residuum_norm = l2_norm(residuum);
+
+      ++iteration;
+    }
+
+    const bool converged = !(residuum_norm > tolerance);
```
dragos-ana: This could lead to faulty behavior, I think. If the residual norm is not evaluable, e.g., it returns NaN, then I think that `NaN > tolerance `will evaluate the `false`, and therefore `converged = true`.

Maybe some dedicated treatment for NaN would be important here to make this bullet-proof.

dragos-ana (commented):
Nice enhancement! I've added some comments

src/core/utils/src/numerics/4C_utils_local_newton.hpp
rjoussen: I am not sure, but this should work.

src/core/utils/src/numerics/4C_utils_local_newton.hpp
rjoussen: I think a lot of the Viscoplastic local newton loop's advanced functionality could move here actually.

src/core/utils/src/numerics/4C_utils_local_newton.hpp
rjoussen: strange that the compiler did not complain about it.

src/core/utils/src/numerics/4C_utils_local_newton.hpp
rjoussen: good catch. I changed it now to the safer and easier to read <=, which means that custom scalar types need to implement this operator. See corresponding test.

src/core/utils/src/numerics/4C_utils_local_newton.hpp
rjoussen: changed here and in two other places.

dragos-ana (approved):

jeremylt (approved):
LGTM

PR #2245, opened 2026-09-08
Handle evaluation failure in NOX wrappers

rjoussen:
## Description and Context

Since I need a way to communicate/signal evaluation failures through nox to the time integrator, I thought I could use the the `false` channel of the nox callbacks. However, the handling of `false` evaluations is quite inconsistent at the moment in our nox wrappers, hence I did the following cleanup first:

- Add `[[nodiscard]]` around most relevant methods that can return non-successful statuses
- Allow some methods to actually return a non-successful status instead of throwing raw char exceptions, namely in
  - `NOX::Nln::Direction::Newton::compute`
  - `NOX::Nln::GroupBase::computeGradient`
  - `NOX::Nln::GroupBase::computeJacobian`
  - `NOX::Nln::GroupBase::computeF`
  - `NOX::Nln::Group::computeF`
  - `NOX::Nln::Group::compute_f_and_jacobian`
- `LineSearch::Backtrack::compute` previously implemented a recovery mechanism for throwing evaluations. Since such evaluations now return false, this needed to be adapted. The previous safeguard check
 `outer_tests_ptr_->checkStatus` is removed, since the function call itself led to a char exception, which was then incorrectly assumed to be caused by a too-long step. ~Otherwise, the behaviour was uncovered before, and is uncovered still (It is pretty hard to trigger `false` evaluations currently)~

I am not entirely sure yet if I will actually use this mechanism to signal evaluation failure, since nox is unfortunately pretty bad at preserving the reason for nonlinear solver failure. In particular, it does not indicate that the solver failed because the evaluation failed. This is already incorrectly assumed this way in the `structure_new` nox interface and may be part of a follow-up. Because of this, typed exceptions might be better suited to signal evaluation failure. Nevertheless, I think this provides valuable cleanup.

**Disclaimer:** I am not very familiar with NOX.

reviewed by codex.

### DIFF
--- a/src/fsi/src/partitioned/model_evaluator/4C_fsi_str_model_evaluator_partitioned.cpp
+++ b/src/fsi/src/partitioned/model_evaluator/4C_fsi_str_model_evaluator_partitioned.cpp
@@ -17,6 +17,7 @@
 #include "4C_structure_new_nln_solver_generic.hpp"
 #include "4C_structure_new_timint_basedataglobalstate.hpp"
 #include "4C_structure_new_timint_implicit.hpp"
+#include "4C_utils_exceptions.hpp"
 
 #include <Teuchos_ParameterList.hpp>
 
@@ -140,14 +141,18 @@ Solid::ModelEvaluator::PartitionedFSI::solve_relaxation_linear(
   // ---------------------------------------------------------------------------
   // Compute F and jacobian
   // ---------------------------------------------------------------------------
-  grp_ptr->computeJacobian();
+  const auto jacobian_status = grp_ptr->computeJacobian();
+  FOUR_C_ASSERT_ALWAYS(jacobian_status == ::NOX::Abstract::Group::ReturnType::Ok,
+      "Failed to compute Jacobian for partitioned FSI structure model evaluator.");
 
   // overwrite F with boundary force
   interface_force_np_ptr_->scale(-(ti_impl->tim_int_param()));
   ti_impl->dbc_ptr()->apply_dirichlet_to_rhs(*interface_force_np_ptr_);
   Teuchos::RCP<NOX::Nln::Vector> nox_force =
       Teuchos::make_rcp<NOX::Nln::Vector>(interface_force_np_ptr_);
-  grp_ptr->set_f(nox_force);
+  const auto residual_status = grp_ptr->set_f(nox_force);
+  FOUR_C_ASSERT_ALWAYS(residual_status == ::NOX::Abstract::Group::ReturnType::Ok,
+      "Failed to set residual for partitioned FSI structure model evaluator.");
 
   // ---------------------------------------------------------------------------
   // Check if we are using a Newton direction
@@ -171,7 +176,9 @@ Solid::ModelEvaluator::PartitionedFSI::solve_relaxation_linear(
   // solve the linear system of equations and update the current state
   // ---------------------------------------------------------------------------
   // compute the Newton direction
-  grp_ptr->computeNewton(p);
+  const auto newton_status = grp_ptr->computeNewton(p);
+  FOUR_C_ASSERT_ALWAYS(newton_status == ::NOX::Abstract::Group::ReturnType::Ok,
+      "Failed to compute Newton direction for partitioned FSI structure model evaluator.");
 
   // get the increment from the previous solution step
   const auto& increment = dynamic_cast<const NOX::Nln::Vector&>(grp_ptr->getNewton());
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
@@ -129,13 +129,10 @@ void NOX::Nln::Group::set_skip_update_x(bool skipUpdateX) { skipUpdateX_ = skipU
 
   if (isF()) return ::NOX::Abstract::Group::Ok;
 
-  const bool success = userInterfacePtr->compute_f(
+  isValidRHS = userInterfacePtr->compute_f(
       xVector.get_linalg_vector(), RHSVector.get_linalg_vector(), NOX::Nln::FillType::Residual);
 
-  if (not success)
-  {
-    throw "NOX::Nln::Group::computeF() - fill failed";
-  }
+  if (not isValidRHS) return ::NOX::Abstract::Group::ReturnType::Failed;
 
   isValidRHS = true;
 
@@ -162,18 +159,16 @@ void NOX::Nln::Group::set_skip_update_x(bool skipUpdateX) { skipUpdateX_ = skipU
     isValidRHS = false;
     prePostOperatorPtr_->run_pre_compute_f(RHSVector.get_linalg_vector(), *this);
 
-    bool status = false;
     Teuchos::RCP<NOX::Nln::LinearSystem> nlnSharedLinearSystem =
         Teuchos::rcp_dynamic_cast<NOX::Nln::LinearSystem>(linearSystemPtr);
 
     if (nlnSharedLinearSystem.is_null())
       throw_error("compute_f_and_jacobian", "Dynamic cast of the shared linear system failed!");
 
-    status = nlnSharedLinearSystem->compute_f_and_jacobian(xVector, RHSVector);
-    if (!status) throw_error("compute_f_and_jacobian", "evaluation failed!");
+    isValidRHS = isValidJacobian =
+        nlnSharedLinearSystem->compute_f_and_jacobian(xVector, RHSVector);
 
-    isValidRHS = true;
-    isValidJacobian = true;
+    if (not isValidRHS or not isValidJacobian) return ::NOX::Abstract::Group::ReturnType::Failed;
 
     ret = ::NOX::Abstract::Group::Ok;
 
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_group.hpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_group.hpp
@@ -76,18 +76,19 @@ namespace NOX
       void computeX(const ::NOX::Abstract::Group& grp, const ::NOX::Abstract::Vector& d,
           double step) override;
 
-      ::NOX::Abstract::Group::ReturnType computeF() override;
+      [[nodiscard]] ::NOX::Abstract::Group::ReturnType computeF() override;
 
-      ::NOX::Abstract::Group::ReturnType applyJacobianInverse(Teuchos::ParameterList& p,
-          const ::NOX::Abstract::Vector& input, ::NOX::Abstract::Vector& result) const override;
+      [[nodiscard]] ::NOX::Abstract::Group::ReturnType applyJacobianInverse(
+          Teuchos::ParameterList& p, const ::NOX::Abstract::Vector& input,
+          ::NOX::Abstract::Vector& result) const override;
 
       //! Compute and store \f$F(x)\f$ and the jacobian \f$\frac{\partial F(x)}{\partial x}\f$ at
       //! the same time. This can result in a huge performance gain in some special cases, e.g.
       //! contact problems.
-      virtual ::NOX::Abstract::Group::ReturnType compute_f_and_jacobian();
+      [[nodiscard]] virtual ::NOX::Abstract::Group::ReturnType compute_f_and_jacobian();
 
       //! set right hand side
-      ::NOX::Abstract::Group::ReturnType set_f(Teuchos::RCP<NOX::Nln::Vector> Fptr);
+      [[nodiscard]] ::NOX::Abstract::Group::ReturnType set_f(Teuchos::RCP<NOX::Nln::Vector> Fptr);
 
 #if !(FOUR_C_TRILINOS_INTERNAL_VERSION_GE(2025, 4))
       //! set the solution vector to zero
--- a/src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
+++ b/src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
@@ -36,7 +36,8 @@ NOX::Nln::LineSearch::Backtrack::Backtrack(const Teuchos::RCP<::NOX::GlobalData>
       outer_tests_ptr_(outerTests),
       inner_tests_ptr_(innerTests)
 {
-  reset(gd, params);
+  const bool success = reset(gd, params);
+  FOUR_C_ASSERT_ALWAYS(success, "Reset of Backtrack line search failed!");
 }
 
 /*----------------------------------------------------------------------------*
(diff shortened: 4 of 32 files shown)


rjoussen:
I looked into the failing tests a bit more and tried to understand the recovery mechanism of line search backtracking. Turns out there are two tests that cover that behaviour, but you only see that if `FOUR_C_ENABLE_FE_TRAPPING` is set. The "feature" used via `Allow Exceptions` in input works as follows:
- If the evaluation returns false, the step is marked as too long. This is the same behavior as before, except before this was done via a very(!) brittle try-catch block. 
- Additionally (and here it gets wild imo), if `Allow Exceptions` is set and `FOUR_C_ENABLE_FE_TRAPPING` is set, backtracking temporarily disables FE trapping (via `fedisableexcept`). However, backtracking itself does not check for floating-point exceptions. Instead, this enables the evaluator to detect them and potentially return false, which then triggers recovery in backtracking. Afterward, backtracking clears the fe flags and re-enables FE trapping. If `FOUR_C_ENABLE_FE_TRAPPING` is not set, floating point exceptions remain undetected entirely. This is what happens in the two currently failing test cases. The new model evaluator error check can turn floating-point exceptions into a `false` evaluation. 

I will think about how to clean this up as well.

src/fsi/src/partitioned/model_evaluator/4C_fsi_str_model_evaluator_partitioned.cpp
```diff
@@ -140,14 +141,18 @@ Solid::ModelEvaluator::PartitionedFSI::solve_relaxation_linear(
   // ---------------------------------------------------------------------------
   // Compute F and jacobian
   // ---------------------------------------------------------------------------
-  grp_ptr->computeJacobian();
+  const auto jacobian_status = grp_ptr->computeJacobian();
+  if (jacobian_status != ::NOX::Abstract::Group::ReturnType::Ok)
+    FOUR_C_THROW("Failed to compute Jacobian for partitioned FSI structure model evaluator.");
```
vovannikov: You use `FOUR_C_ASSERT_ALWAYS` in most of the other similar cases down below. I would use it here too for consistency.

src/fsi/src/partitioned/model_evaluator/4C_fsi_str_model_evaluator_partitioned.cpp
```diff
@@ -140,14 +141,18 @@ Solid::ModelEvaluator::PartitionedFSI::solve_relaxation_linear(
...
 
   // overwrite F with boundary force
   interface_force_np_ptr_->scale(-(ti_impl->tim_int_param()));
   ti_impl->dbc_ptr()->apply_dirichlet_to_rhs(*interface_force_np_ptr_);
   Teuchos::RCP<NOX::Nln::Vector> nox_force =
       Teuchos::make_rcp<NOX::Nln::Vector>(interface_force_np_ptr_);
-  grp_ptr->set_f(nox_force);
+  const auto residual_status = grp_ptr->set_f(nox_force);
+  if (residual_status != ::NOX::Abstract::Group::ReturnType::Ok)
+    FOUR_C_THROW("Failed to set residual for partitioned FSI structure model evaluator.");
```
vovannikov: Here too.

src/fsi/src/partitioned/model_evaluator/4C_fsi_str_model_evaluator_partitioned.cpp
```diff
@@ -171,7 +176,10 @@ Solid::ModelEvaluator::PartitionedFSI::solve_relaxation_linear(
   // solve the linear system of equations and update the current state
   // ---------------------------------------------------------------------------
   // compute the Newton direction
-  grp_ptr->computeNewton(p);
+  const auto newton_status = grp_ptr->computeNewton(p);
+  if (newton_status != ::NOX::Abstract::Group::ReturnType::Ok)
+    FOUR_C_THROW(
+        "Failed to compute Newton direction for partitioned FSI structure model evaluator.");
```
vovannikov: And here.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.hpp
```diff
@@ -76,18 +76,19 @@ namespace NOX
...
 
       //! Compute and store \f$F(x)\f$ and the jacobian \f$\frac{\partial F(x)}{\partial x}\f$ at
       //! the same time. This can result in a huge performance gain in some special cases, e.g.
       //! contact problems.
-      virtual ::NOX::Abstract::Group::ReturnType compute_f_and_jacobian();
+      [[nodiscard]] virtual ::NOX::Abstract::Group::ReturnType compute_f_and_jacobian();
 
       //! set right hand side
-      ::NOX::Abstract::Group::ReturnType set_f(Teuchos::RCP<NOX::Nln::Vector> Fptr);
+      [[nodiscard]] ::NOX::Abstract::Group::ReturnType set_f(Teuchos::RCP<NOX::Nln::Vector> Fptr);
```
vovannikov, line 91: I understand why you added `[[nodiscard]]` here and in general this is a good idea. There is only one issue with this: it creates interface inconsistency, since `::NOX::Abstract::Group` has not these attributes, so if any of these methods are called via the `::NOX::Abstract::Group` pointer, no warnings are emitted. I don't have a strong opinion on this. Let's see how this draft evolves.

vovannikov:
> I looked into the failing tests a bit more and tried to understand the recovery mechanism of line search backtracking. Turns out there are two tests that cover that behaviour, but you only see that if `FOUR_C_ENABLE_FE_TRAPPING` is set. The "feature" used via `Allow Exceptions` in input works as follows:
> 
> * If the evaluation returns false, the step is marked as too long. This is the same behavior as before, except before this was done via a very(!) brittle try-catch block.
> * Additionally (and here it gets wild imo), if `Allow Exceptions` is set and `FOUR_C_ENABLE_FE_TRAPPING` is set, backtracking temporarily disables FE trapping (via `fedisableexcept`). However, backtracking itself does not check for floating-point exceptions. Instead, this enables the evaluator to detect them and potentially return false, which then triggers recovery in backtracking. Afterward, backtracking clears the fe flags and re-enables FE trapping. If `FOUR_C_ENABLE_FE_TRAPPING` is not set, floating point exceptions remain undetected entirely. This is what happens in the two currently failing test cases. The new model evaluator error check can turn floating-point exceptions into a `false` evaluation.
> 
> I will think about how to clean this up as well.

I will have a look into it. I guess the best way would be actually to remove our own `Backtrack` implementation completely and simply reuse the one from Trilinos.

src/fsi/src/partitioned/model_evaluator/4C_fsi_str_model_evaluator_partitioned.cpp
rjoussen: must have missed this file, thanks!

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.hpp
rjoussen: I see what you mean and admittedly overlooked that. However, the only reason why we wrap/implement/override these functions is that we want to use them ourselves, and if we do so, we have to make sure that we use them correctly, which of course means not ignoring the return value. If NOX internally uses their functions without [[nodiscard]], it is NOX's responsibility to still process the return (which I hope NOX does). That being said, I can also remove the [[nodiscard]] flags from all functions that directly override NOX functions. It still makes sense to take care of the return value everywhere we use it I assume.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
```diff
@@ -198,25 +172,21 @@ bool NOX::Nln::LineSearch::Backtrack::compute(::NOX::Abstract::Group& grp, doubl
...
-      print_update(utils_->out(::NOX::Utils::InnerIteration));
-    }
-    // catch error of the computeF method
-    catch (const char* e)
+    rtype = grp.computeF();
+    if (rtype != ::NOX::Abstract::Group::ReturnType::Ok)
     {
-      if (not fp_except_.shall_be_caught_) FOUR_C_THROW("An exception occurred: {}", e);
+      utils_->out(::NOX::Utils::Warning)
+          << "Evaluation was not successful. The line search step length will be reduced!\n";
```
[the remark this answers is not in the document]
rjoussen: It was like this before already. Another argument not to have our own brittle backtracking implementation, I assume.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.cpp
```diff
@@ -129,13 +129,10 @@ ::NOX::Abstract::Group::ReturnType NOX::Nln::Group::computeF()
...
 
-  const bool success = userInterfacePtr->compute_f(
+  isValidRHS = userInterfacePtr->compute_f(
       xVector.get_linalg_vector(), RHSVector.get_linalg_vector(), NOX::Nln::FillType::Residual);
 
-  if (not success)
-  {
-    throw "NOX::Nln::Group::computeF() - fill failed";
-  }
+  if (not isValidRHS) return ::NOX::Abstract::Group::ReturnType::Failed;
```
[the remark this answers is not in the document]
rjoussen: leaving the isValidRHS as false is sufficient here. compute_f is already responsible for clearing the right hand side anyway, and run_pre_compute_f is already inert for invalid RHS. Same below.

rjoussen:
>  will have a look into it. I guess the best way would be actually to remove our own Backtrack implementation completely and simply reuse the one from Trilinos.

Probably. I looked into how to communicate floating-point exceptions safely. It appears that returning `false` from an evaluation to signal recoverable failure is not the way to go. Even if our wrappers can handle that more safely or transform into a `ReturnType::Failed`, the Trilinos implementations likely can not. Additionally, a bool flag obviously doesn't carry which exact problem occurred, so blind recovery is dangerous anyway. I think the safest way would be to allow evaluation to throw typed exceptions, something like:
```cpp
const bool exception_occurred = FloatingPointException::catch_evaluation_failure([&]{compute_f_and_jacobian();})
```
which expands to
```cpp
template <typename Function>
bool catch_evaluation_failure(Function&& f)
{
    // set private global flag that safe evaluation is active
    safe_evaluation_environment = true;
    try
    {
        std::forward<F>(f)();
        return false;
    }
    catch (TypedFloatingPointException)
    {
        return true;
    }
}
```
But that should be done independently of this PR.

vovannikov:
I have looked in the current `Backtrack` algorithm and it seems it is not so easy to remove it immediately since it relies on the inner tests, which are not used by `Backtrack` from Trilinos. However, I would like to propose a few immediate modifications:
- Remove completely option "Allow Exceptions" since it is very confusing. Just consider it is set to true by default.
- Refactor `FloatingPointException` to remove `shall_be_caught_` completely, this is another very confusing parameter that overcomplicates things. If a user uses `FloatingPointException` in the first place, then he/she wants this special handling of exceptions.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.hpp
vovannikov: After all, I think it is better to remove those `[[nodiscard]]` attributes on the overloaded members for consistency. Sure, keep taking care of the return values. Just the attributes. And you can keep `[[nodiscard]]` on the methods we defined ourselves.

As a sidenote, I think this has to be refactored. Probably, I would split the internal NOX `Group` interfaces from ours, but this is not so easy to do.

rjoussen:
@vovannikov  I am not sure actually. If a user configured with `FE_TRAPPING` enabled, i think they expect that FPEs throw exceptions unconditionally, at least that seems to be the normal convention in 4C, exceptions are enabled in global_full_main. On top of that, cutting back the step on every floating-point exception in the evaluation seems really dangerous. Especially, it only makes sense if this FPE is actually caused by a too-large step, which is unclear and can only be determined at the point where it occurs. It might also be a programming error for example. So I think I would rather remove this feature entirely in its current shape, to be honest. Also, there is still the problem that the catch block does not even run the inner status tests currently (as copilot pointed out), so FPEs might cut back the step forever.

Alternatively, I am currently working on rank-local time-step reduction requests, which allow materials to request a smaller global time-step from the calling algorithm. I could imagine this feature being fairly easily reused to reduce the line-search step as well.

vovannikov:
> @vovannikov I am not sure actually. If a user configured with FE_TRAPPING enabled, i think they expect that FPEs throw exceptions unconditionally, at least that seems to be the normal convention in 4C, exceptions are enabled in global_full_main. On top of that, cutting back the step on every floating-point exception in the evaluation seems really dangerous. Especially, it only makes sense if this FPE is actually caused by a too-large step, which is unclear and can only be determined at the point where it occurs. It might also be a programming error for example. So I think I would rather remove this feature entirely in its current shape, to be honest. Also, there is still the problem that the catch block does not even run the inner status tests currently (as copilot pointed out), so FPEs might cut back the step forever.

I am not sure that my suggestions should overlap or be orthogonal to your observations. For instance, currently we have 4 tests with `Allow Exceptions` option: 2 have it enabled and 2 disabled. Why? I mean can one answer this question without looking into the code? Just from the physics or some other features of the problem? If the answer is "No", then such an option should not exist.

Removal of `shall_be_caught_` in that context seems to be not doing any harm either.

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.hpp
rjoussen: I totally agree that this needs to be refactored, it is super confusing how it is right now.

rjoussen:
> After all, I think it is better to remove those [[nodiscard]] attributes on the overloaded members for consistency. Sure, keep taking care of the return values. Just the attributes. And you can keep [[nodiscard]] on the methods we defined ourselves.

I understand that concern, but I think returning non-successful statuses (`false` or `NOX::Abstract::Group::ReturnType::Failed`) from functions that previously threw an error, without marking these functions as `[[nodiscard]]` at the same time, is too dangerous because it allows ignoring such statuses. Since this was the initial goal of this PR, I am closing it. Let me know if you disagree @vovannikov :)

Instead, I will open a new PR suggesting removing the `Allow Exceptions` parameter in line search, this is independent anyway.

vovannikov:
Closing the PR is a bit too radical, in my opinion :) By itself, adding `[[nodiscard]]` is not a bad idea, especially if we a talking about our own members whose declarations we fully control. By `[[nodiscard]]` by its design can not operate as a replacement of exceptions, since it emits a warning by default and not an error and it can be suppressed by a preceding `(void)...` construction. But probably, the prior refactoring of `Group`s and maybe some other parts of the code which are inherited from the vanilla `NOX` objects is not a bad idea.

rjoussen:
@vovannikov

Okay, I reopened this, but with a lot narrower changes:
- all methods we introduce get a `[[nodsicard]]` flag.
- Put `FOUR_C_ASSERT_ALWAYS` to some call sites that were previously ignored. Not all of them are directly related to the added nodiscards, but we should also check the result of functions that are directly inherited from nox.
- No changes to the actual return values of any functions. If we cant guarantee nodiscard, those changes are too risky, in my opinion.

I will address the `Allow Exceptions` flag seperately, no changes required for this PR.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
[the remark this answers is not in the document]
rjoussen: oudated

src/solver_nonlin_nox/4C_solver_nonlin_nox_group.hpp
rjoussen: nodiscards are removed in the latest commit

vovannikov:
@rjoussen, re-check this PR after #2253 gets merged. I guess that some part of it got covered by that PR and also by #2252.

rjoussen:
@vovannikov absolutely, I am aware that there is some overlap. I will tackle the merge conflicts with #2252 first.

Issue #481, opened 2025-03-04
[BUG] Restarts with Penalty Contact Formulation


dragos-ana:
## Bug Description: What's going wrong?

In the last couple of weeks, it was reported to me that utilizing restarts in contact simulations with the penalty method is dangerous, as the results with and without restart may deviate from each other. I observed such phenomena myself during the course of my Master's Thesis, 8-9 months ago, when utilizing Coulomb friction. 
To study whether this is really the case and to identify the origin of these issues, I considered our current framework tests `contact3D_slidingblock_penalty.dat` and `contact3D_slidingblock_coulomb.dat`, and created similar input files for both simulations with penalty contact and with Lagrange Multiplier contact. To quantify the observations, I looked at the element-averaged Cauchy stress component $\sigma_{xz}$, specified at a cross-section of the upper, smaller block in the contact problem. Specifically, the cross-section is a slice in $z-$ direction at $z=0.25$. 
For the given Coulomb friction coefficient $\mu_{\text{fric}}=0.1$, no significant deviations in the considered stress were observed when employing different restarts at $t=\{0.5~\mathrm{s}, ~2~\mathrm{s},~ 3.5~\mathrm{s}, ~7~\mathrm{s} \}$. However, when raising the Coulomb friction coefficient to $\mu_{\text{fric}}=10.0$, the deviations when employing restarts for the penalty contact simulations are quite significant, as can be observed in the relative errors below.  

![Image](https://github.com/user-attachments/assets/107ebeb9-d5f0-4bf8-bd65-f9a06f57d1cf)

Interestingly, applying the Lagrange Multiplier method for the same contact problem does not reveal any restart-based deviations, see below:

![Image](https://github.com/user-attachments/assets/82955249-a13e-4976-87f0-8b9ead413308)

The same holds for the case when no friction is simulated for penalty contact:

![Image](https://github.com/user-attachments/assets/97e95487-5566-4b7b-b96a-33a3511a32e5)

## Expectations

Even though the penalty contact seems to not have converged to the LM solution yet (which can be controlled with the normal and tangential penalty parameters), I would still expect that the solution remains the same if one uses one or more restarts along the way. 

## Your Environment

- **Relevant repo SHA1s:**
- **Relevant configure flags or configure script:**
- **Operating system and version:**
- **Compiler and TPL versions:**

## Related Issues and Pull Requests

- Blocks
- Follows
- Related to
- Part of
- Composed of

isteinbrecher:
[...]

PR #2247, opened 2026-09-10
Simplify vtk testing framework to also enable testing against other fixtures (e.g., against restarted runs)

dragos-ana:
Simplifies the vtk testing framework but also enhances its functionality, e.g., by enabling tests against other fixtures (useful for testing restarts).
The explicit vtk-restart testing function is not necessary anymore and was removed as a result.

Addendum to: #1138, so it might be interesting for @knarfnitram @davidrudlstorfer @isteinbrecher @c-p-schmidt


davidrudlstorfer (changes requested):
From a modularity perspective this is super radical and not reusable at all.

Wouldn't it be possible to also build this up in a modular way and maybe even reuse existing stuff?

For example run an existing base test, then introduce a new "subtest" that only does the restart from a certain point (which we also already have?) and then run a regular vtk comparison test?

This is now 150 lines of code that have to be maintained separately and cannot be used anywhere else at all 😅

dragos-ana:
@davidrudlstorfer Both current vtk comparison tests `__four_c_test_vtk` and `four_c_test_restarted_vtk` explicitly compare against reference pvd files that already exist under `tests/input_files/...`. 
This is a completely different test here, since we compare between runs and not with already saved reference files. 
As far as "cannot be used anywhere else" goes, this is actually extremely necessary in my opinion - I am doing this in preparation to solving #481 where restarted penalty contact simulations are not consistent with full ones. And I am very sure that there are a lot of frameworks where restarts do not necessarily work as consistently as expected.

If anything, then this should actually replace `four_c_test_restarted_vtk` which, as far as I see, does the exact same thing as `__four_c_test_vtk` but for the restarted simulation, in a two-stage process.

davidrudlstorfer:
> This is a completely different test here, since we compare between runs and not with already saved reference files.

But then I guess the other function could just be extended to use the saved reference file as default and if the argument is handed over it uses the path of currently executed simulation? Please do not get me wrong here, the functionality is nice but the implementation adds another function rather than unifying behaviour

dragos-ana:
If we want to unify this completely, it would mean that `__four_c_test_vtk` becomes a very general vtk comparison test, where one can specify both files, and everything else vtk related has to be done only using this test (both the test I implemented and `four_c_test_restarted_vtk` would have to be rewritten / removed). 

I'll try this out here.

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -1189,14 +1189,14 @@ function(
...
+# Central function to define a vtk test for a previous input file test: either against a reference pvd file, or against another input file test (e.g., corresponding restart test)
 #
 # required parameters:
-#   BASED_ON:                     Reference to previous test
-#   PVD_RESULT:                   name of the .pvd file in the output directory of the previous test that references the .vtk files to compare
-#   PVD_REFERENCE:                name of the .pvd file in tests/input_files that
-#                                 references the .vtk files to compare against
-#   TOLERANCE:                    difference the values may have
+#   BASED_ON:                      Reference to previous test for which the comparison is run
+#   BASED_ON_PVD:                  Name of the .pvd file in the output directory of the previous test that references the .vtk files to compare
```
davidrudlstorfer: I am not yet super happy with the name of this variable, maybe we find something better but my imagination is currently not super good 😆 maybe @c-p-schmidt has an idea.

Maybe something along the lines

BASED_ON_REFERENCE
and
BASED_ON_TEST?

It should be more straight forward to understand what is going on in the list of tests when one reads the vtk comparison

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -1189,14 +1189,14 @@ function(
...
+#   PVD_REFERENCE or COMPARE_WITH: Either PVD_REFERENCE, i.e., the name of the .pvd file in tests/input_files that references the .vtk files to compare against, or COMPARE_WITH: similarly references a previous test to take as a reference for comparison
```
davidrudlstorfer: Same goes for this

PVD_REFERENCE is a path, so maybe we could include this in the variable naming? 

PVD_REFERENCE_PATH? I am also not yet super convinced of the naming

tests/list_of_tests.cmake
```diff
@@ -3033,13 +3038,6 @@ four_c_test_tutorial(TEST_FILE preconditioner/tutorial_prec_fsi_muelu.4C.yaml NP
...
-four_c_test_restarted_vtk(TESTNAME correct-restarted-vtk-files-for-solid_runtime_tet4 FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory PVD_RESULTFILENAME xxx-structure.pvd PVD_REFERENCEFILENAME ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-06 TIMEOUT 10 TIME_STEPS 0.5)
-four_c_test_restarted_vtk(TESTNAME correct-restarted-vtk-files-for-solid_runtime_tet4 FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory_implicit PVD_RESULTFILENAME xxx-1-structure.pvd PVD_REFERENCEFILENAME ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-06 TIMEOUT 10 TIME_STEPS 0.5)
-four_c_test_restarted_vtk(TESTNAME thermo3D_FBC_ost.4C.yaml-restarted-vtk-with-initial-state FILE thermo3D_FBC_ost.4C.yaml RESTART_STEP 3 RESTARTFROM_PATHTYPE relative_from_parent PVD_RESULTFILENAME xxx-thermo.pvd PVD_REFERENCEFILENAME ref/thermo3D_FBC_ost-thermo.pvd TOLERANCE 1e-06)
-
```
davidrudlstorfer, line 3042: Amazing that these are now gone 🤩

davidrudlstorfer (commented):
@dragos-ana I am 100% honest when saying that these changes are just amazing!

Now the functionality of the cmake test functions is even extended while still removing more than 170 lines of code. This really shows how unnecessary the old duplicated test definition already was.

Big thank you, that you invested the time into further looking into this. The functionality is now really nice and easily understandable.

The only thing I am not yet really sold is the variable naming, IMHO the variables should be a little bit more self explanatory. Maybe someone else has some ideas how to improve the naming, I will also give it a thought.

All in all, big thank you :)

tests/list_of_tests.cmake
rjoussen: @dragos-ana @davidrudlstorfer The only problem is that these tests were testing restart behaviour that is now untested. See #1482 and following fixes in #1844 and #1871. Those were mainly testing that the pvd collection files are written correctly for different `restart_from` path types. There were many problems with this. If we want to continue testing this, the better place is probably to extend the normal `four_c_test_restart` by such a parameter.

cmake/functions/four_c_testing_functions.cmake
rjoussen: Idea for naming:
```cmake
PVD_RESULT_FILE     # the name of the pvd file in this test to compare
REFERENCE_TEST      # optional: the name of the dependent test
PVD_REFERENCE_FILE  # if REFERENCE_TEST is provided: the name of the pvd reference file
                    # in the output directory of the reference test.
                    # Otherwise: the path to the pvd reference file in tests/input_files
```
I think that would fit the current design quite well.

davidrudlstorfer: I like that @rjoussen 👍

c-p-schmidt: I also like the suggestion 👍🏼

tests/list_of_tests.cmake
dragos-ana: Re-enabled now within the newest commit. They are now run with different options for `solid_runtime_tet4.4C.yaml` for both SAME_FILE and TEST_FILE specifications.

tests/list_of_tests.cmake
```diff
@@ -2026,67 +2027,80 @@ four_c_test(TEST_FILE solid_ele_wedge6_Standard_stressout.4C.yaml NP 2 RETURN_AS
...
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATHTYPE relative_from_parent)
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory)
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory_implicit)
+__four_c_test_vtk(BASED_ON ${current_restart} PVD_RESULT_FILE xxx-1-structure.pvd PVD_REFERENCE_FILE ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-06 TIMEOUT 10 TIME_STEPS 0.5)
+# TEST_FILE
+__four_c_test_restart(BASED_ON ${current} TEST_FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE absolute)
+__four_c_test_restart(BASED_ON ${current} TEST_FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE relative)
+__four_c_test_restart(BASED_ON ${current} TEST_FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE relative_from_parent)
+__four_c_test_restart(BASED_ON ${current} TEST_FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory)
+__four_c_test_restart(BASED_ON ${current} TEST_FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory_implicit)
```
rjoussen: To test this properly, every restart should compare its vtk output against the reference, or the result of the test without the restart. Otherwise, we just test that every restart_from version can write pvds, but not if they are actually correct.

dragos-corneliu-ana: Done for the TEST FILE cases, which tests the required pvd file validity for different restartfrom options. 

For the SAME FILE cases, parallel test runs do not work in the current framework, because the comparison is run for restart pvds `xxx-<count>-structure.pvd`, and `<count>` is only really determined if you run them sequentially (race conditions). One could extend the restart test functionality to also incorporate a specified optional name for the restart files, but I see this as a separate PR

tests/list_of_tests.cmake
```diff
@@ -1647,7 +1648,7 @@ four_c_test(TEST_FILE scatra_3D_tet4_hex8_s2i_constperm_mortar_saddlepoint_petro
 __four_c_test_restart(BASED_ON ${current} SAME_FILE NP 3 RESTART_STEP 10 REQUIRED_DEPENDENCIES TRILINOS_MUMPS)
 four_c_test(TEST_FILE scatra_3D_tet4_hex8_s2i_constperm_mortar_standard.4C.yaml NP 3 RETURN_AS current REQUIRED_DEPENDENCIES TRILINOS_MUMPS)
 __four_c_test_restart(BASED_ON ${current} SAME_FILE NP 3 RESTART_STEP 10 ASSERT_RESTART_STEP 10 REQUIRED_DEPENDENCIES TRILINOS_MUMPS)
-__four_c_test_restart(BASED_ON ${current} TEST_FILE scatra_3D_tet4_hex8_s2i_constperm_mortar_standard_restart.4C.yaml NP 3 RESTART_STEP last_possible ASSERT_RESTART_STEP 20 REQUIRED_DEPENDENCIES TRILINOS_MUMPS)
+#__four_c_test_restart(BASED_ON ${current} TEST_FILE scatra_3D_tet4_hex8_s2i_constperm_mortar_standard_restart.4C.yaml NP 3 RESTART_STEP last_possible ASSERT_RESTART_STEP 20 REQUIRED_DEPENDENCIES TRILINOS_MUMPS)
```
rjoussen: This should not be commented out.

tests/list_of_tests.cmake
```diff
@@ -2026,67 +2027,85 @@ four_c_test(TEST_FILE solid_ele_wedge6_Standard_stressout.4C.yaml NP 2 RETURN_AS
...
-__four_c_test_vtk(BASED_ON ${current} PVD_RESULT xxx-structure.pvd PVD_REFERENCE ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-08)
+__four_c_test_vtk(BASED_ON ${current} PVD_RESULT_FILE xxx-structure.pvd PVD_REFERENCE_FILE ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-08)
+# test restarts using SAME_FILE / TEST_FILE settings and different --restartfrom options
+# SAME_FILE
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATHTYPE absolute RETURN_AS current_restart)
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATHTYPE relative)
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATHTYPE relative_from_parent)
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory)
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory_implicit)
+__four_c_test_vtk(BASED_ON ${current_restart} PVD_RESULT_FILE xxx-1-structure.pvd PVD_REFERENCE_FILE ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-06 TIMEOUT 10 TIME_STEPS 0.5)
```
rjoussen: This is also already weird from a race-condition point of view: There is nothing that says that another restart than current_restart could be first and hence produce the xxx-1-structure pvd file that you want to compare here.

tests/list_of_tests.cmake
```diff
@@ -2353,12 +2372,15 @@ __four_c_test_restart(BASED_ON ${current} TEST_FILE tfsi_plate.4C.yaml NP 3 REST
 four_c_test(TEST_FILE thermo_bar_DBC_genalpha.4C.yaml NP 2)
 four_c_test(TEST_FILE thermo-line.4C.yaml NP 2)
 four_c_test(TEST_FILE thermo3D_FBC_ost.4C.yaml NP 2)
+four_c_test(TEST_FILE thermo3D_FBC_ost.4C.yaml NP 1 RETURN_AS current)
```
rjoussen: there is no reason this test should run twice with different NP. Just use the first and return as current.

cmake/functions/four_c_testing_functions.cmake
rjoussen:
```suggestion
  if(_parsed_SAME_FILE AND DEFINED _parsed_TEST_FILE)
```
Certainly not your fault, but I spotted this mistake which should be corrected.

tests/list_of_tests.cmake
```diff
@@ -2353,12 +2372,15 @@ __four_c_test_restart(BASED_ON ${current} TEST_FILE tfsi_plate.4C.yaml NP 3 REST
...
+__four_c_test_restart(BASED_ON ${current} SAME_FILE NP 1 RESTART_STEP 3 RETURN_AS current_restart)
+__four_c_test_vtk(BASED_ON ${current_restart} PVD_RESULT_FILE xxx-1-thermo.pvd PVD_REFERENCE_FILE ref/thermo3D_FBC_ost-thermo.pvd TOLERANCE 1e-06)
```
rjoussen:
```suggestion
__four_c_test_vtk(BASED_ON ${current_restart} PVD_RESULT_FILE xxx-1-thermo.pvd REFERENCE_TEST $(current) PVD_REFERENCE_FILE xxx-thermo.pvd TOLERANCE 1e-06)
```
this would now also work. When I introdcued this test, that is all what I wanted to test at least. Would test your new functionality again also and remove a reference file. But not urgently necessary to change that right now.

tests/list_of_tests.cmake
rjoussen: make sense, thanks for the explanation. As pointed out in one review comment, I think there is already an inconsitency with that unfortunately.

tests/list_of_tests.cmake
rjoussen: Looks very nice!

cmake/functions/four_c_testing_functions.cmake
dragos-ana: Why is this a mistake? This is a check that both cannot be specified at the same time...

rjoussen: `parsed_SAME_FILE` is missing the initial underscore, which, I believe, means that this it is not set, and this check always results in false, because `parsed_SAME_FILE` is not set, only `_parsed_SAME_FILE` is.

rjoussen: So the check is completely valid and should be there, but its not doing what we think currently.

dragos-ana: You are completely right, thanks

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -444,15 +444,17 @@ function(four_c_test)
...
-# Define a restart test that restarts from a previous test
+# Define a restart test that restarts from a previous test.
 #
 # required parameters:
 #   BASED_ON:                name of the base test that created the restart files
 #   RESTART_STEP:            number of the restart step to restart from or last_possible
 #   SAME_FILE or TEST_FILE:  either SAME_FILE to indicate that the restart should be done from the same input file
-#                            as the base test, or TEST_FILE to indicate that a different input file should be used for the restart
+#                            as the base test, or TEST_FILE to indicate that a different input file should be used for the restart.
+#                            Note that in the case of SAME_FILE, the restart output is also saved within the same directory as the base test; in the case of TEST_FILE, a new test directory is created (except when RESTARTFROM_PATHTYPE is same_directory_implicit!).
```
davidrudlstorfer: I know that the formatting does not work super well in the cmake files but maybe you could manually add a few lines breaks in some places.

cmake/functions/four_c_testing_functions.cmake
```diff
@@ -444,15 +444,17 @@ function(four_c_test)
...
 #
 # optional parameters:
+#   RESTARTFROM_PATHTYPE     <absolute|relative|relative_from_parent|same_directory|same_directory_implicit>. Defaults to absolute if not specified.
```
davidrudlstorfer: Would this make more sense as

RESTART_FROM_PATH_TYPE?

The variable is somehow hard to read, missing a few underscores and only containing one

tests/list_of_tests.cmake
```diff
@@ -2026,67 +2027,85 @@ four_c_test(TEST_FILE solid_ele_wedge6_Standard_stressout.4C.yaml NP 2 RETURN_AS
...
+__four_c_test_restart(BASED_ON ${current} TEST_FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE relative_from_parent RETURN_AS current_restart)
+__four_c_test_vtk(BASED_ON ${current_restart} PVD_RESULT_FILE xxx-structure.pvd PVD_REFERENCE_FILE ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-06 TIMEOUT 10 TIME_STEPS 0.5)
+__four_c_test_restart(BASED_ON ${current} TEST_FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory RETURN_AS current_restart)
+__four_c_test_vtk(BASED_ON ${current_restart} PVD_RESULT_FILE xxx-structure.pvd PVD_REFERENCE_FILE ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-06 TIMEOUT 10 TIME_STEPS 0.5)
+__four_c_test_restart(BASED_ON ${current} TEST_FILE solid_runtime_tet4.4C.yaml RESTART_STEP 1 RESTARTFROM_PATHTYPE same_directory_implicit RETURN_AS current_restart)
+__four_c_test_vtk(BASED_ON ${current_restart} PVD_RESULT_FILE xxx-1-structure.pvd PVD_REFERENCE_FILE ref/solid_runtime_tet4-vtk/structure.pvd TOLERANCE 1e-06 TIMEOUT 10 TIME_STEPS 0.5)
 four_c_test(TEST_FILE solid_runtime_velocity_output.4C.yaml NP 2 RETURN_AS current)
```
davidrudlstorfer: Is this really needed? To me this looks like a little bit too much for every pipeline. Wouldn't it make sense to for example execute such tests in a local pipeline for testing?
It's weird to me to execute the same test nearly 10 times

davidrudlstorfer (commented):
Looks good! Only a few minor formatting / variable comments

cmake/functions/four_c_testing_functions.cmake
dragos-ana: I agree, but I think that `RESTARTFROM_PATH_TYPE` would be slightly better because --restartfrom is also the option used in the input files.

tests/list_of_tests.cmake
dragos-ana: I did it now for two separate BASED_ON (mainly because only one `same_directory_implicit` restart is allowed now for each BASED_ON). 
I'm fine either way - although I think that it's also valuable, at least in theory, to try to restart using all restartfrom options from a file differing from the original test input file (although it's also not really done so now, because we use the same file in TEST_FILE).

dragos-ana:
@rjoussen @davidrudlstorfer Since this race condition problem seems very strange to me, I decided to add a restart output prefix in the newest commit. Specifically, the restart framework now explicitly requires unique restart output prefixes: this means that two SAME_FILE restarts from the same BASED_ON require at least one of the output prefixes to be specified (while the other one can be the default xxx). For TEST_FILE restarts, it's required that the test names (which now also contain the restart output prefix) are unique. One sensible point is the `same_directory_implicit` option for `--restartfrom`: since both TEST_FILE and SAME_FILE write the restart files to the same directory, which could again lead to this race condition problem, I decided to only allow one `same_directory_implicit` test for each BASED_ON (either SAME_FILE or TEST_FILE). To me it seems that the --restartfrom options should only be tested once, and hence, the specification of --restartfrom is rather uncommon for restart tests. Therefore, I think that this restriction in regards to `same_directory_implicit` will not be problematic.

tests/list_of_tests.cmake
dragos-ana: I agree, see my comment below:

https://github.com/4C-multiphysics/4C/pull/2247#issuecomment-5677731253

tests/list_of_tests.cmake
rjoussen: This is important behavior and needs to be tested, in my opinion. As long as we "offer" the implicit naming and restarts (=not specifying restartfrom), we need to ensure that everything works, including pvd file continuation. Also, this test is very cheap.

tests/list_of_tests.cmake
```diff
@@ -2026,67 +2027,77 @@ four_c_test(TEST_FILE solid_ele_wedge6_Standard_stressout.4C.yaml NP 2 RETURN_AS
...
+__four_c_test_vtk(BASED_ON ${current_restart} REFERENCE_TEST ${current} PVD_RESULT_FILE restart_relative_from_parent-structure.pvd PVD_REFERENCE_FILE xxx-structure.pvd TOLERANCE 1e-15 TIMEOUT 10 TIME_STEPS 0.5)
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATH_TYPE same_directory RESTART_OUTPUT_PREFIX "restart_same_directory" RETURN_AS current_restart)
+__four_c_test_vtk(BASED_ON ${current_restart} REFERENCE_TEST ${current} PVD_RESULT_FILE restart_same_directory-structure.pvd PVD_REFERENCE_FILE xxx-structure.pvd TOLERANCE 1e-15 TIMEOUT 10 TIME_STEPS 0.5)
+__four_c_test_restart(BASED_ON ${current} SAME_FILE RESTART_STEP 1 RESTARTFROM_PATH_TYPE same_directory_implicit RETURN_AS current_restart)
+__four_c_test_vtk(BASED_ON ${current_restart} REFERENCE_TEST ${current} PVD_RESULT_FILE xxx-1-structure.pvd PVD_REFERENCE_FILE xxx-structure.pvd TOLERANCE 1e-15 TIMEOUT 10 TIME_STEPS 0.5)
 four_c_test(TEST_FILE solid_runtime_velocity_output.4C.yaml NP 2 RETURN_AS current)
```
rjoussen, line 2098: Looks good now!

rjoussen (approved):
Thanks!

tests/list_of_tests.cmake
davidrudlstorfer: I am still not 100% sure with this here, so the functionality is now just there without actually being needed?

> This is important behavior and needs to be tested, in my opinion. As long as we "offer" the implicit naming and restarts (=not specifying restartfrom), we need to ensure that everything works, including pvd file continuation. Also, this test is very cheap.

In my opinion we should only offer functionality that is also needed and not just because we can 😅

Maybe I am not understanding something here 😅

And if the functionality is already tested in other tests we should not test it here again?

davidrudlstorfer (approved):
First of all, thanks a lot @dragos-ana for the work you put into this! It looks really really good now and a lot of duplicated code was able to be removed. Thanks!

There is only the one discussion/thread open which I am not yet fully understanding.

I do not wanna block this any longer so feel free to merge, it would still be nice if this could be resolved now or soonish

tests/list_of_tests.cmake
rjoussen: > In my opinion we should only offer functionality that is also needed and not just because we can 😅

Agreed, but as long as restarts without restartfrom are possible, it should be tested that that works. There was a discussion about it recently if that's needed: #1991. 

> And if the functionality is already tested in other tests we should not test it here again?

It is not tested other than these tests, as far as I know. Normal restart of course are, but these tests specifically test pvd file continuation across restarts.

dragos-ana: I think we have to separate here two aspects: first, we can use various path specifications for `--restartfrom`, such as absolute, relative, ... Whether all available possibilities work should be tested in my opinion as well: I also had problems with restarts a couple of months if I was not in the same directory during restarts (but could have also been a mistake from my side). Anyway, I think that the various possibilities should be tested at least once. So that would mean running 5 different restarts tests from an input file.  
The second aspect is the input file specification for these different restarts: here, I am unsure whether we also need to re-test all 5 possibilities using the `TEST_FILE` option, in addition to the `SAME_FILE` restarts. We could remove this, since this specification does not really affect the behavior that we want to test. Or we can keep it for completeness. I'm fine with either variant.

dragos-ana:
@davidrudlstorfer @rjoussen I'll merge this now. If the remaining thread should be further discussed, we can do it in private and come up with an alternative solution in a subsequent PR.

PR #2248, opened 2026-09-10
Change InitialField to enum class

c-p-schmidt:
## Description and Context
This PR changes the InitialField from an enum to an enum class and removes all unused members. As a result, quite some code could be deleted.
Moreover, the input definition is changed from **deprecated_selection** to **parameter**.

Copilot integration within the IDE was used for some restructuring.


c-p-schmidt:
One input parameter has changed (one that is probably not used anymore), and some that were unused have been removed, thus technically a breaking change.

davidrudlstorfer (approved):
Nice cleanup :)

engstlerL (approved):

PR #2250, opened 2026-09-11
Change VelocityField to enum class

c-p-schmidt:
## Description and Context
This PR changes the VelocityField from an enum to an enum class and renames the member **Navier_Stokes** -> **from_other_field**. I know this is another breaking change, but **Navier_Stokes** has been wrong in most use cases, so it is an utterly bad name. And even when it was not wrong, **from_other_field** is a better name as it comes from the fluid field. In a lot of other applications, it is the deformation velocity of the solid, so **from_other_field** sounds reasonable to me.

Changed from **deprecated_selection** to **parameter** for the input definition and improved some checks along the way.

Follows #2248

Copilot integration within the IDE was used for some restructuring.


davidrudlstorfer (approved):
thanks for the cleanup :)

tuchpaul (approved):
Thanks for solving Navier Stokes! 🥳

PR #2251, opened 2026-09-14
Cache the full performance test meshes in nightly pipeline

maxiludwig:
# Description

GitHub LFS bandwidth for the repo is nearing its quota, perhaps largely from `nightly_tests.yml` re-downloading the same 4 performance-test meshes (~28MB) every night. This caches the LFS objects across nightly runs so they're only downloaded when the mesh files actually change.

# Changes
- `actions/checkout` no longer fetches LFS content directly (`lfs: false`) — it only checks out the small pointer files.
- Added `actions/cache@v6`, keyed on a hash of the mesh pointer files, caching `.git/lfs`.
- Added an explicit `git lfs pull` step, which resolves objects from the cache on a hit, or downloads from LFS on a miss (first run, or whenever a mesh changes).


davidrudlstorfer (approved):
looks good

bgoderbauer (approved):

c-p-schmidt (approved):
Looks good, let's give it a try!

PR #2252, opened 2026-09-14
Introduce `NOX::Nln::LineSearch::Controller` to simplify and separate interfaces

vovannikov:
When reviewing #2245 by @rjoussen I noticed a few suboptimal design decisions in the class hierarchy:
1. `NOX::Nln::LineSearch::Generic` uses multiple inheritance and as result, it mixes our interfaces and those of `NOX`. It is better to minimize such situations if possible.
2. `NOX::Nln::Inner::StatusTest::Interface::Required` is basically obsolete and does not add any value but unnecessary complexity.
3. Since `NOX::Nln::Inner::StatusTest::Generic` is our own interface, there is no need to use the base classes of `NOX` if those lead to unnecessary `dynamic_cast`s. I restructured its interface accordingly.

#2245

AI tools were not used in this contribution


mayrmt (dismissed):

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_controller.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+FOUR_C_NAMESPACE_OPEN
+
+namespace NOX
+{
+  namespace Nln
+  {
+    namespace LineSearch
+    {
+      class Controller
```
rjoussen, line 35: Could you document, at least briefly, what the intent of this class is?

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.hpp
```diff
@@ -32,55 +33,65 @@ namespace NOX
...
-            ::NOX::StatusTest::CheckType checkType) const override;
+          //! get the current search direction
+          const ::NOX::Abstract::Vector& get_search_direction() const override;
 
-        //! @name Access functionality
-        //@{
-        //! get the number of line search iterations
-        int get_num_iterations() const override;
+          //! get current step length
+          double get_step_length() const override;
```
rjoussen, line 61: all of these should be marked `[[nodsicard]]`

rjoussen (commented):
Nice, that already cleans up a lot! I left some minor comments.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.hpp
vovannikov: In fact `[[nodiscard]]` is optional here, since once the virtual method was declared `[[nodiscard]]` it remains so for the overrides too. Though `[[nodiscard]]` is handy, I find its syntax quite ugly and I prefer not to pollute the code with it when this is unnecessary.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_controller.hpp
vovannikov: Sure, done!

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.hpp
rjoussen: I dont think function attribrutes are inherited unfortunetely: https://stackoverflow.com/questions/49576298/are-function-attributes-inherited

vovannikov: Yes, you are right. For some reason I thought the behavior was different. Fixed!

rjoussen (approved):
Looks good now!

mayrmt (approved):
Great!

PR #2253, opened 2026-09-14
Refactor floating point exception handling in line search ("Allow Exceptions")

rjoussen:
## Description and Context

Came up while working on #2245 

Our own Backtrack line search implementation had the following feature, enabled by setting `Allow Exceptions: true`:
- every `char` exception (probably, but not necessarily raised by an evaluation returning `false`. It could just as well be a programming error or anything else) during evaluation got caught and remapped to `step_too_long`
- If 4C was configured with `FOUR_C_ENABLE_FE_TRAPPING`, it would temporarily disable floating point exceptions (`fedisableexcept`) around the evaluation, which basically allows floating point exceptions to occur silently. The new structure model evaluator was the only place that checked for FE events that had occurred and returned false if one had occurred, which eventually escalated to a char exception. 

This has many issues some of which are:
- A char exception should not mean step too long unconditionally. It is not even guaranteed that the char exception means that evaluation failed.
- In almost the entire code, it seems to be the convention that `FOUR_C_ENABLE_FE_TRAPPING` means that floating point exceptions cause program termination. The current backtrack violates that.
- Any floating point exceptions/events not inside the structure model evaluator remain undetected.
- backtrack was reducing the step due to caught char exceptions without re-running the inner status tests, which could lead to an infinite loop

That's why this PR refactors this feature in a couple of steps (same as commits)
1. Remove `Allow Exceptions` from two tests that did not actually need it
2. Run the evaluation operations (`computeF`) in `Backtrack::compute` through the newly introduced `run_and_detect_floating_point_exceptions()` wrapper. This wrapper internally disables FP exceptions, runs the evaluation, checks and reduces any FP events that occur, and returns a global decision. It also has an RAII guard to make it exception-safe and exactly restore the previous FPE environment. 
3. Remove nox floating point exception entirely, because FPEs are now handled in backtrack completely. Also, it seems to be the wrong mechanism to detect a too long line search step, especially in combination with the necessity to catch a char exception. The structure model evaluator's eval error check function is then a no-op and is deleted.

Note: the last commit makes the entire model evaluator returning bools obsolete, because there is no false return path anymore. This is a bigger refactor and will be done in a follow-up.

none


src/mat/elast/4C_mat_elast_coupneohooke.cpp
```diff
@@ -56,14 +56,15 @@ void Mat::Elastic::CoupNeoHooke::add_derivatives_principal(Core::LinAlg::Matrix<
...
-    dPI(2) = ddPII(2) = std::numeric_limits<double>::quiet_NaN();
+  const double prinv2_to_beta_m1 = std::exp(std::log(prinv(2)) * (-beta - 1.));
+  dPI(2) -= c * prinv2_to_beta_m1;
+  ddPII(2) += c * (beta + 1.) * prinv2_to_beta_m1 / prinv(2);
+  FOUR_C_ASSERT_ALWAYS(std::isfinite(dPI(2)),
+      "Error in principal derivative computation. Derivative of strain energy wrt. I3 = {}",
+      dPI(2));
+  FOUR_C_ASSERT_ALWAYS(std::isfinite(ddPII(2)),
+      "Error in principal derivative computation. Second derivative of strain energy wrt. I3 = {}",
+      ddPII(2));
```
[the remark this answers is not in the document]
rjoussen: Tests added.

src/mat/4C_mat_elasthyper_service.cpp
```diff
@@ -35,6 +36,11 @@ void Mat::elast_hyper_evaluate(const Core::LinAlg::Tensor<double, 3, 3>& defgrd,
   static Core::LinAlg::Matrix<6, 1> ddPII(Core::LinAlg::Initialization::uninitialized);
   ddPII.clear();
 
+  // check consistency
+  const double detF = det(defgrd);
+  FOUR_C_ASSERT_ALWAYS(
+      detF > 0, "Negative determinant of the deformation gradient. det(F) = {}.", detF);
```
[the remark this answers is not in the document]
rjoussen: This assertion was removed

jeremylt (dismissed):
This looks like a good cleanup to me. Am I understanding correctly that this doesn't allow for new failure modes that were previously caught?

rjoussen:
> Am I understanding correctly that this doesn't allow for new failure modes that were previously caught?

Yes. In a way, this is removing functionality, but the functionality was so brittle that I think removing is worth it. If a local-overflow-to-line-search-step-reduction mechanism is desired in the future, this needs to be done differently.

src/mat/elast/4C_mat_elast_coupneohooke.cpp
vovannikov, line 67: I guess the reason for making this modification is `std::numeric_limits<double>::quiet_NaN()`? Did you check that there is not any other part of the code that exploits `dPI(2)` or `dPII(2)` to be that value?

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
vovannikov, line 168: If I tracked the code properly, now there is not way to get into this loop, since there is no any point in the code that code set `status_ = NOX::Nln::Inner::StatusTest::status_step_too_long`. Apparently, the behavior of the solver has changed, most probably this was not properly covered by the tests.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
```diff
@@ -340,7 +340,7 @@ bool Solid::ModelEvaluator::Structure::initialize_inertia_and_damping()
   // assemble the rayleigh damping matrix
   rayleigh_damping_matrix();
 
-  return eval_error_check();
+  return true;
```
vovannikov, line 343: Are you sure that blindly returning `true` here no matter what is safe? If am a bit confused. The same applies to the other similar modifications.

vovannikov (commented):
You have moved even further than I initially proposed in that PR :) I actually was also thinking that the complete removal of `FloatingPointException` is the way we need to proceed, but I was not sure how this can be safely done. `FloatingPointException` introduces a lot of unnecessary complexity of the code, but prior to its complete removal we need to ensure that the error handling scenarios are preserved. Currently, the main things that bug me are:
1. One of the control loops in the `Backtrack` linesearch algorithm became unreachable.
2. The default return of `true` in some of the methods looks a bit too radical for me.

Both these modifications seem to modify the original control flow quite significantly.

src/mat/elast/4C_mat_elast_coupneohooke.cpp
rjoussen: prinv(2) is here the determinant of the right cauchy-green tensor, which is positive by construction. The only reason the else block could be entered here is due to prinv(2) being nan. With the new asserts, such values are caught anyway.

> Did you check that there is not any other part of the code that exploits dPI(2) or dPII(2) to be that value?

I would say if that happens, that is a bug. The more likely scenario is that non-finite values here fail in elast_hyper_add_isotropic_stress_cmat and then in calculate_gamma_delta because of invalid operations.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
rjoussen: Yes, this is safe. eval_error_check was the only remaining function that could modify the return value, and it only checked for nox floating point exceptions. If this is gone, we don't need this anymore.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
rjoussen: The `NOX::Nln::Inner::StatusTest::Armijo::check_status` implementation can return `step_too_long`, just based on the merit function. So I think this loop is still entered, but not for floating point exceptions anymore.

src/mat/elast/4C_mat_elast_coupneohooke.cpp
vovannikov: I agree with this, one should not rely on such things.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
vovannikov: Indeed looks that way. Anyways, since previously the failures here could have appeared from the floating point exceptions, this is redundant.

vovannikov (dismissed):
Yep, after a more careful look at the code, it seems fine to remove `FloatingPointException` this way.

Another thing to consider, is to revert to the default C++ behavior for which the floating point error do not really trigger exceptions.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
vovannikov: Just a small thing from my side. It is worth adding a specific comment around these `return true;` to comment why it is was made and why returning `true` is safe and there is not a better way (currently) to do it. A better approach actually would be to make the functions called inside each of this methods more verbose returning `true` or `false` and then pass this data to the caller. But surely not within this PR.

rjoussen: I really dislike the `return true` here too. However, I don't think the solution is to make some of the functions called herein return a bool. That's because returning false from these evaluation functions is not really useful: Firstly, it doesn't tell you anything about what went wrong. Secondly, a false return here would propagate to the nox callback at some point, leading to either an internal nox exception or one of our own char exceptions, called from our nox implementation/wrappers. In that case, it is more useful to refactor this function to void and just throw our own exceptions inside the evaluation (then you know at least what went wrong).
That being said, I will still leave a brief comment here for now to explain the ugly `return true`; it doesn't hurt if it's there, even temporarily.

jeremylt: Could also explicitly open an issue and cite it in the comment

rjoussen: @jeremylt My follow-up PR on this is ready, and I will open it as soon as this is merged, so it's not needed, I think.

rjoussen:
> Another thing to consider, is to revert to the default C++ behavior for which the floating point error do not really trigger exceptions.

What do you mean by that?

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_structure.cpp
rjoussen: comment is added now

tests/input_files/sohex8fbar_cooks_nl_line_search_new_struc.4C.yaml
```diff
@@ -12,8 +12,8 @@ PROBLEM TYPE:
 STRUCTURAL DYNAMIC:
   DYNAMICTYPE: "Statics"
   RESTARTEVERY: 200
-  TIMESTEP: 50
-  NUMSTEP: 2
+  TIMESTEP: 6.25
+  NUMSTEP: 32
```
vovannikov: And here?

tests/input_files/sohex8_easfull_cooks_nl_line_search_new_struc.4C.yaml
```diff
@@ -10,8 +10,8 @@ PROBLEM TYPE:
 STRUCTURAL DYNAMIC:
   DYNAMICTYPE: "Statics"
   RESTARTEVERY: 0
-  TIMESTEP: 12.5
-  NUMSTEP: 8
+  TIMESTEP: 6.25
+  NUMSTEP: 16
```
vovannikov: Was there any specific reason to change these settings?

vovannikov (dismissed):

vovannikov:
> > Another thing to consider, is to revert to the default C++ behavior for which the floating point error do not really trigger exceptions.
> 
> What do you mean by that?

We have `FOUR_C_ENABLE_FE_TRAPPING` set by default, whereas the default C++ standard option is the opposite. In fact, `-ftrapping-math` is a very tricky thing and it does not guaranty that an FP exception gets triggered in the case of such an operation. But the tricky thing is that `-ftrapping-math` is the default option in GCC but not in clang, where it actually also depends on the optimization level.

jeremylt (dismissed):

tests/input_files/sohex8_easfull_cooks_nl_line_search_new_struc.4C.yaml
rjoussen: As explained in the PR description, these two tests would fail now because of the new assertions in the elastic neo hooke material, which doesn't lead to step too long anymore. These tests are consequently fixed by reducing the time-step size.

rjoussen: Lmk if that works for you, then I'll merge

rjoussen:
> We have `FOUR_C_ENABLE_FE_TRAPPING` set by default, whereas the default C++ standard option is the opposite. In fact, `-ftrapping-math` is a very tricky thing and it does not guaranty that an FP exception gets triggered in the case of such an operation. But the tricky thing is that `-ftrapping-math` is the default option in GCC but not in clang, where it actually also depends on the optimization level.

I agree. I'll open an issue for that.

vovannikov:
> > We have `FOUR_C_ENABLE_FE_TRAPPING` set by default, whereas the default C++ standard option is the opposite. In fact, `-ftrapping-math` is a very tricky thing and it does not guaranty that an FP exception gets triggered in the case of such an operation. But the tricky thing is that `-ftrapping-math` is the default option in GCC but not in clang, where it actually also depends on the optimization level.
> 
> I agree. I'll open an issue for that.

I would not rush with this, this is something to consider. I would think carefully on that prior to implementation. Anyways, this is somewhat a low priority question.

tests/input_files/sohex8_easfull_cooks_nl_line_search_new_struc.4C.yaml
vovannikov: Ah, ok, then this has to be resolved in another way. If the solver crashes with the old settings, reducing the timestep size is not a proper solution, since this reduction should happen automatically.

src/mat/elast/4C_mat_elast_coupneohooke.cpp
vovannikov: Do you mean that somewhere here a value goes to infinity due to large nonlinearity? Then I guess the issue is not the new assertions, but rather that `try... catch` block that used to catch `char *` and reducing the timestep subsequently fot removed. Probably, the right solution then is actually to bring that block back, with `fp_except_` removed of course. I am not absolutely happy handling all type of exceptions as recoverable here, but that seemed to be the initial intent of `Backtrack`.

rjoussen: > Do you mean that somewhere here a value goes to infinity due to large nonlinearity?

Yes, that is what is happening. One goes to plus inf, the other one to minus inf, to be precise, which later leads to a nan value in the stress calculation.

> Probably, the right solution then is actually to bring that block back, with fp_except_ removed of course. I am not absolutely happy handling all type of exceptions as recoverable here

A "simple" `try catch(Core::Exception)` doesn't work here, because the non-finite value here might only occur on one rank, but not on others. Hence, if there is any MPI collective between the try-catch block in backtrack and the raised exception in this material, this will lead to an MPI deadlock. Such events would need to be reduced first at a meaningful MPI boundary.

> but that seemed to be the initial intent of `Backtrack`

Agreed, and that's the problem in my opinion. That is faulty by design and should be removed. I would even argue that the two existing tests that need this do not ensure the feature works in general; they are testing a corner case where it happens to work. The corner case is: you have a `FOUR_C_ENABLE_FE_TRAPPING` build; the char exception was actually due to an evaluation failure, which in turn came from an FPE, which can be meaningfully mitigated by reducing the step size. 

Whether an evaluation failure occurs because the current step is too large can only be determined at the place where the failure actually happens. That decision should not be in `Backtrack`, I think. 

All in all, I think it's worth removing this feature in its current form, even though it happened to be helpful for two tests.

Also, I am working on (safe) material-triggered step reduction requests. My main goal was to reduce the time step in such cases, but it could also be used easily for line-search backtracking. But this is a somewhat larger change/feature and cannot be done here as a quick fix.

tests/input_files/sohex8_easfull_cooks_nl_line_search_new_struc.4C.yaml
jeremylt: I'm generally a fan of breaking things up, but maybe this should be combined with the larger fix PR?

src/mat/elast/4C_mat_elast_coupneohooke.cpp
vovannikov: I would not say that the stepsize handling in `Backtrack` and inner tests is non-natural. These classes have all the necessary infrastructure, so the intent was there (I was not the one who wrote them, of course, so may only reason about).

My concern is not even about the tests failing, it is about the solver robustness, the tests have just revealed the problem (that's their job). Getting `inf` is by itself not a dramatic failure and should be handled in one way or another. The issue is if a user specifies in the input a too large step, a solver then should be able to reduce it untill reaching the predefied minimum limit and only then it is allowed to crash, if a solver has the step reduction functionality, and apparently `Backtrack` has it. This behavior seems to be not preserved within this PR.

rjoussen:
Okay, next try: Handling of FP exceptions in Backtrack is restored, but is no longer routed through the nox floating point exception object. Instead, Backtrack is the only place where the FPE environment is altered, and FPE events are checked for. This removes the try-catch blocks and makes it much clearer that Backtrack owns allowing FPEs here. This is also much safer in many regards than the previous version with try-catch blocks.

Especially, the changes to the coup neo hooke material have been reverted and the two affected tests don't need to change. The Allow Exceptions input option is also kept, because I think this is still somewhat sketchy behavior that shouldn't be enabled by default (that's also how it was before, btw)

For details, see the updated PR description and code.

With these changes, this PR is no longer breaking, hence removing the label.

tests/input_files/sohex8_easfull_cooks_nl_line_search_new_struc.4C.yaml
rjoussen: The time step is no longer changed in the recent push.

rjoussen: > I'm generally a fan of breaking things up

same :)

src/mat/elast/4C_mat_elast_coupneohooke.cpp
rjoussen: I believe my last push restored the intended behavior almost entirely, except for previous char exceptions unrelated to FPEs, but it seemed not the intent of Backtrack to catch those anyway. I still think it's somewhat risky to backtrack on every FP event, but it's significantly safer now than it used to be.

src/solver_nonlin_nox/4C_solver_nonlin_nox_input.cpp
```diff
@@ -170,8 +170,8 @@ std::vector<Core::IO::InputSpec> NOX::valid_parameters()
                   .default_value = 0.5}),
           parameter<bool>("Allow Exceptions",
               {.description =
-                      "Set to true, if exceptions during the force evaluation and backtracking "
-                      "routine should be allowed.",
+                      "If set to true and if FOUR_C_ENABLE_FE_TRAPPING is enabled, floating point "
+                      "exceptions during trial evaluations will lead to a step reduction.",
                   .default_value = false})},
```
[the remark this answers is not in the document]
rjoussen: PR description changed.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
```diff
@@ -7,20 +7,105 @@
...
+      {
+        os << "WARNING: Floating point exception occurred on rank "
+           << Core::Communication::my_mpi_rank(comm);
+      }
+
+      return Core::Communication::sum_all(static_cast<int>(local_fe_except), comm) > 0;
+    }
+#endif
+
+    // if FOUR_C_ENABLE_FE_TRAPPING is not defined or if allow_floating_point_exceptions is false,
```
jeremylt: comment is confusing - all code above is guarded by FOUR_C_ENABLE_FE_TRAPPING

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
```diff
@@ -127,13 +213,19 @@ bool NOX::Nln::LineSearch::Backtrack::compute(::NOX::Abstract::Group& grp, doubl
...
-    if (rtype != ::NOX::Abstract::Group::Ok) throw_error("compute", "Unable to compute F!");
+  bool fpe_occurred = false;
+
+  fpe_occurred = run_and_detect_floating_point_exceptions([&]() { rtype = grp.computeF(); },
+      allow_floating_point_exceptions_, comm, utils_->out(::NOX::Utils::Warning));
+  if (rtype != ::NOX::Abstract::Group::Ok) throw_error("compute", "Unable to compute F!");
 
+  if (fpe_occurred)
+  {
+    utils_->out() << "Last step caused a floating point exception. Reducing step length...\n";
```
jeremylt, line 220: Is it useful to print the rank as before?

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
rjoussen: I removed the comment.

src/solver_nonlin_nox/4C_solver_nonlin_nox_linesearch_backtrack.cpp
rjoussen: The rank is printed directly by the `run_and_detect_floating_point_exceptions` function into the warning stream. If rank-local things are printed, they should include the rank-number, in my opinion, otherwise it looks strange why something is printed multiple times.

jeremylt: ah, now I see that above

jeremylt (approved):

src/mat/elast/4C_mat_elast_coupneohooke.cpp
vovannikov: I like the way it looks now. I also agree that this is still not the best design, I would keep it for the being at least. It is good that the hacks with the FPE treatment are now localized within a single class and not exposed to any other class.

vovannikov (approved):
@rjoussen thank you for this effort! Excellent result within the existing constraints posed by the legacy code.

rjoussen:
Thanks for reviewing me into it @vovannikov :)

PR #2258, opened 2026-09-15
Clean up predictor state handling

rjoussen:
### Commit 1-5

These are pure cleanup for various predictor related stuff, see individual commit messages.

### Commit 6

This adds an explicit `reset_state()` method to the predictor, which resets all flags the predictor sets during prediction.
This is not strictly necessary right now, but it will be useful once evaluation failure during prediction can be recovered from (which I am working on). It is hence called from the `structure_new_timeint_implicit` existing `reset_step()` function.

none


src/structure_new/src/predict/4C_structure_new_predict_generic.cpp
```diff
@@ -77,7 +76,16 @@ void Solid::Predict::Generic::predict(::NOX::Abstract::Group& grp)
   // post-process the prediction step
   post_predict(grp);
 
-  ispredict = false;
+  reset_state();
+}
+
+/*----------------------------------------------------------------------------*
+ *----------------------------------------------------------------------------*/
```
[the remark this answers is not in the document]
rjoussen: The functionality is implicitly tested: It gets called inside in the end of every prediction (inside predict()) and again by every existing regression test that calls `reset_step`. This shows that the new functionbis not resetting too much state. Explicitly testing state reset for potentially unfinished predictor is very hard at the moment because there is currently no non-fatal predictor failure. Unit tests are also infeasible because the predictor relies on too many objects, including a fully set-up time integrator.

src/structure_new/src/implicit/4C_structure_new_timint_implicit.cpp
```diff
@@ -221,6 +221,12 @@ const ::NOX::Abstract::Group& Solid::TimeInt::Implicit::get_solution_group() con
   return nln_solver().get_solution_group();
 }
 
+void Solid::TimeInt::Implicit::reset_step()
+{
+  ImplicitBase::reset_step();
+  predictor().reset_state();
+}
```
vovannikov, line 228: If I understood correctly, this change is just for consistency, isn't it? I mean there was no big need to reset the state of the predictor, but it indeed a reasonable thing to do here.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_data.hpp
```diff
@@ -152,18 +152,6 @@ namespace Solid
...
-      [[nodiscard]] Solid::PredEnum get_predictor_type() const override
-      {
-        check_init_setup();
-        return predict_type_;
-      }
-
-
-
-      //! is the current state the predictor state?
-      bool is_predictor_state() const;
```
vovannikov, line 165: Why is this `is_predictor_state()` member and many other similar ones get removed? Just to ensure that everything works fine. I actually fine this `is_predictor*` members very confusing and that infrastructure should be refactored in general.

vovannikov (commented):
This is a very good clean up! I have just a couple of questions to clarify understanding before merging this.

src/structure_new/src/model_evaluator/4C_structure_new_model_evaluator_data.hpp
rjoussen: `is_predictor_state` specifically provided knowledge about the implicit time integrator's ispredictor_state_ flag. According to its documentation, this flag is only indicating that the last state has been set by the predictor and should disallow the `set_state` routines to set a new state. This is exactly how it is used, but I think this also means it does not need to be available outside the implicit time integrator. The particular function here is unused, same applies to the other removed predictor-related getter methods.

src/structure_new/src/implicit/4C_structure_new_timint_implicit.cpp
rjoussen: At the moment, yes. That's because currently, this method is called only after a failed solve and ensures resetting such that the current step can be restarted from a fresh (resetted) state. During/after solution, the predictor state should have been reset already anyway, so there is currently no explicit need for it.

However, once a time step should be able to be reset and retried also due to a failed predictor, it is important to keep track of the predictor state as well. This makes this function a correct reset routine during the whole solid time step, not only after a failed/incomplete solve.

vovannikov (approved):

PR #2260, opened 2026-09-16
Skip performance test on forks

maxiludwig:
# Description
Switch off the nightly performance-test job on forks. Forks without their own pushed LFS content were pulling the performance-test meshes from this repo's LFS storage on every scheduled run, consuming a large share of our LFS bandwidth quota. This change makes performance testing run only on `4C-multiphysics/4C`.


davidrudlstorfer (approved):
Do you have an exemplary fork where this was running?

Normally people have the testing disabled for forks anyways. I don't know if this is the real problem

maxiludwig:
I did not check all forks, but at least @c-p-schmidt  and me dont have testing disabled. :upside_down_face:

davidrudlstorfer:
And then one more question, if the repo is forked, isn't then the LFS also "forked" and therefore pulling for the fork user? This then should not go towards the @4C-multiphysics quota?

maxiludwig:
Apparently the forks still pull from the parent directory LFS 
https://docs.github.com/en/repositories/working-with-files/managing-large-files/collaboration-with-git-large-file-storage

c-p-schmidt:
> Do you have an exemplary fork where this was running?
> 
> Normally people have the testing disabled for forks anyways. I don't know if this is the real problem

I would say, at least currently, it is a problem, as we have run into our quota limit. Nevertheless, it is a good point. We can remind people (including me 😉 ) that this can be done. However, since we cannot enforce it (as far as I know), this is still a valid remedy for this problem imo.

c-p-schmidt (approved):

davidrudlstorfer:
@maxiludwig thanks for the explanation and the link :)

maxiludwig:
@4C-multiphysics/developers 
Please sync your personal fork for this change to take action.

c-p-schmidt:
> @4C-multiphysics/developers Please sync your personal fork for this change to take action.

Alternatively, you can deactivate the Actions (i.e., testing) on your fork. For that do the following on your 4C fork:
Settings -> Actions -> General (Click **Disable actions**)

PR #2263, opened 2026-09-17
Make pleural pressure offset spatially varying

maxiludwig:
# Description 
Make the pleural pressure offset spatially varying by turning `pressure_offset` into a `Core::IO::InputField<double>`. Residual assembly now looks the offset up per boundary node via its attached `global_element_id`. `TimeFunction`-based conditions are unaffected. Added a unit test covering two nodes on different elements getting different offsets from the same definition.

# Disclosure of AI assistance
Written with Claude Code (Claude Opus 5)


bgoderbauer:
How does the pressure offset translate to a value in Pascal?

maxiludwig:
This depends on the unit system which is used for the simulation.  The pressure offset has the units of a pressure.

src/reduced_lung/tests/4C_reduced_lung_boundary_conditions_test.cpp
```diff
@@ -59,11 +59,11 @@ namespace
...
-            PleuralPressureDefinition::NormalizedLinearExponential{.pressure_offset = 0.5,
-                .linear_coefficient = 2.0,
-                .exponential_coefficient = 0.25,
-                .exponential_rate = 0.1}};
+        .normalized_linear_exponential = PleuralPressureDefinition::NormalizedLinearExponential{
+            .pressure_offset = Core::IO::InputField<double>(0.5),
+            .linear_coefficient = 2.0,
+            .exponential_coefficient = 0.25,
+            .exponential_rate = 0.1}};
   }
```
bgoderbauer, line 67: Ah okay, makes sense. What is the unit system used for these default values? Should we maybe document this in the code?

maxiludwig: I am not sure if this makes sense. Our unittests are unit-agnostic so far and I am not sure if it makes sense to change that.

bgoderbauer (approved):

engstlerL (approved):

PR #2264, opened 2026-09-18
Refactoring: Contact "stresses" are correctly renamed to normal / tangential tractions

dragos-ana:
## Description and Context

Previously, contact tractions were inconsistently called "stresses" in the contact framework; this PR refactors this aspect to correctly reflect the difference. Additionally, the difference to the contact forces also tracked in the contact strategies is better documented, and the computation of contact tractions in the Lagrange strategy is refactored.


bgoderbauer (approved):

apps/post_processor/4C_post_processor_single_field_writers.cpp
rjoussen, line 1: Are the post-processors still needed for contact? They can be output in vtk now.

c-p-schmidt: In theory, you are right, but I think there are still contact capabilities that might only be available in the old time integration, so we might consider keeping it.

c-p-schmidt: I briefly double-checked. The changes to the contact manager are actually for the old time integration and use the output that needs to be post-processed.

c-p-schmidt (approved):

apps/post_processor/4C_post_processor_single_field_writers.cpp
rjoussen: Okay, thanks!

PR #2267, opened 2026-09-18
[beaminteraction, solver] Add scaling to param list and get in `Teko`

maxfirmbach:
## Description and Context

This is a prep-step for a future block preconditioner. I attach the interface scaling vector for the beam-solid contact case to the Teko solver parameterlist and retrieve it in the respective preconditioning adapter. In addition the Lagrange multiplier map is attached to the reordering maps such that we can split into a 3x3 matrix later on inside the preconditioner.

Currently this is just a transfer of information ... I tested everything by making sure that for the relevant cases printing out the vector contains the correct values.

The actual use of the scaling vector inside a preconditioner will come in a later PR. In beaminteraction I had to move one method from `protected` to `public`.


mayrmt (approved):
LGTM.

src/beaminteraction/src/4C_beaminteraction_str_model_evaluator.cpp
```diff
@@ -878,6 +879,33 @@ bool Solid::ModelEvaluator::BeamInteractionModelEvaluator::assemble_jacobian(
             (*me_vec_ptr_)[0]);
 
     beam_contact_model->assemble_stiff(jac);
+
+    auto kappa_inv = beam_contact_model->get_penalty_kappa_inverse();
+    auto structure_solver = tim_int().get_data_sdyn().get_lin_solvers().at(Solid::model_structure);
+
+    auto& teko_parameters = structure_solver->params().sublist("Teko Parameters");
```
maxfirmbach, line 886: Seems like all the following code needs to be inside an if-statement checking if `Teko` is actually used as preconditioner.

isteinbrecher (approved):

PR #2269, opened 2026-09-21
Change evaluate methods of model evaluator to `void` (instead of `bool`)

rjoussen:
## Description and Context

This changes the evaluation functions (i.e. the remaining `evaluate_*` and `assemble_*` methods) of the `ModelEvaluator::Generic` class to `void`. This is a follow-up to #2253.

The reasons for this are:
1. After #2253, none of these methods or their implementations return false in any code path. The only remaining false path was due to floating point exception detection in the structure model evaluator, which got removed in #2253.
2. The second reason is that returning `false` in these functions is pretty useless, because these functions eventually participate in the callbacks for the nox solver, and false returns leads to non-recoverable failures anyway. This is why I think the model evaluators should not be able to return non-successful statuses in the first place. For details see below.
---
### Details
The call chain of the model evaluator's evaluate functions is roughly as follows:

#### Combined force and stiffness evluation):

- `bool Solid::ModelEvaluatorManager::apply_force_stiff`
  - `bool Solid::IMPLICIT::Statics::apply_force_stiff` (false is preserved)
    - `bool Solid::TimeInt::NoxInterface::compute_f_and_jacobian` (false is preserved)
      - `bool NOX::Nln::LinearSystem::compute_f_and_jacobian` (false is preserved)
        - `::NOX::Abstract::Group::ReturnType NOX::Nln::Group::compute_f_and_jacobian` (false leads to `FOUR_C_THROW`, **not** `ReturnType::Failed`)
          - `bool NOX::Nln::Direction::Newton::compute` (non-successful status leads to char exception) <- direct overwrite of `bool ::NOX::Direction::Newton::compute`, hence called from nox) **or**
          - `Solid::TimeInt::Implicit::evaluate()` (return value is currently ignored, also see #2245 
 
#### Just force evaluation:
-  `bool Solid::ModelEvaluatorManager::apply_force`
   - `bool Solid::IMPLICIT::Statics::apply_force` (false is preserved)
     - ` Solid::TimeInt::NoxInterface::compute_f` (false is preserved)
       - `::NOX::Abstract::Group::ReturnType NOX::Nln::GroupBase::computeF` (true is asserted via `FOUR_C_ASSERT`, **not** `FOUR_C_ASSERT_ALWAYS`, then `ReturnType::Ok` is returned) **or**
       - `::NOX::Abstract::Group::ReturnType NOX::Nln::Group::computeF` (false leads to char exception)
       - both of these are overwrites of `::NOX::Abstract::Group::ReturnType ::NOX::Abstract::Group::computeF()`, so they might be called directly from nox.

Other call chains mostly follow this pattern. This means that returning false in the model evaluator could lead to:
- `FOUR_C_THROW`, but at a much less descriptive place compared to throwing where the failure actually happened.
- Our own char exception (ugly)
- non-success being ignored in `Solid::TimeInt::Implicit::evaluate()` (dangerous. As far as I can see, this currently doesn't happen, but it makes it really dangerous to change any of the return values to non-successful statuses, see again #2245)
- a `false` return value from `bool ::NOX::LineSearch::Generic::compute`
- a non-successful return value from `::NOX::Abstract::Group::ReturnType ::NOX::Abstract::Group::computeF` or `::NOX::Abstract::Group::ReturnType ::NOX::Abstract::Group::computeJacobian`, handled by nox (see below)
- Direct callers of NOX::Nln::GroupBase::computeF() could ignore a non-successful status. This currently might happen here: https://github.com/4C-multiphysics/4C/blob/main/src/fsi/src/monolithic/nonlinear_solver/4C_fsi_nox_group.cpp#L41-L53

Even if non-successful statuses would/could be propagated to the actual nox callbacks, nox will most likely throw its own internal error, see for example
- https://github.com/trilinos/Trilinos/blob/develop/packages/nox/src/NOX_Solver_LineSearchBased.C?utm_source=chatgpt.com#L123-L128
- https://github.com/trilinos/Trilinos/blob/develop/packages/nox/src/NOX_Direction_Newton.C?utm_source=chatgpt.com#L96-L98
- https://github.com/trilinos/Trilinos/blob/develop/packages/nox/src/NOX_Direction_Newton.C?utm_source=chatgpt.com#L114-L116

#2245 
follows #2253 

none


c-p-schmidt (approved):
The description sounds convincing

